PR #1343, opened 2015-01-07
MULE-8206 responseStreamingMode="NEVER" isn't always respected

dfeist:


dfeist:
Don't merge, this is still under discussion..

PR #1345, opened 2015-01-07
MULE-8169: Refactor transport discovery for OSGi

pablokraan:
_ Initial change to enable using a different transport descriptor service factory in an OSGi container

### DIFF
--- a/core/src/main/java/org/mule/api/registry/MuleTransportDescriptorService.java
+++ b/core/src/main/java/org/mule/api/registry/MuleTransportDescriptorService.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+
+package org.mule.api.registry;
+
+import org.mule.api.MuleContext;
+import org.mule.config.i18n.MessageFactory;
+import org.mule.util.SpiUtils;
+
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/mule/api/registry/MuleTransportDescriptorService.java
```diff
@@ -0,0 +1,64 @@
...
+    private final Map<String, TransportServiceDescriptorFactory> serviceDescriptorFactories = new ConcurrentHashMap<>();
+
+    @Override
+    public ServiceDescriptor getDescriptor(String transport, MuleContext muleContext, Properties overrides) throws ServiceException
+    {
+        TransportServiceDescriptorFactory transportServiceDescriptorFactory = serviceDescriptorFactories.get(transport);
+
+        if (transportServiceDescriptorFactory == null)
+        {
+            synchronized (serviceDescriptorFactories)
```
pabad: Change serviceDescriptorFactories to be a Cache and create it using a CacheBuild with a CacheLoader implementation

core/src/main/java/org/mule/api/registry/MuleTransportDescriptorService.java
```diff
@@ -0,0 +1,64 @@
...
+                }
+            }
+        }
+
+        return transportServiceDescriptorFactory.create(muleContext, overrides);
+    }
+
+    private TransportServiceDescriptorFactory createServiceDescriptorFactory(String transport) throws ServiceException
+    {
+        //Stripe off and use the meta-scheme if present
```
pabad: stripe -> strip

PR #1348, opened 2015-01-08
MULE-7875: Adding domain option to FunctionalTestCase and HTTP module ex...

afelisatti:
...ample.

### DIFF
--- a/tests/functional/src/main/java/org/mule/tck/junit4/FunctionalTestCase.java
+++ b/tests/functional/src/main/java/org/mule/tck/junit4/FunctionalTestCase.java
@@ -105,10 +105,18 @@ public abstract class FunctionalTestCase extends AbstractMuleContextTestCase
     protected void doTearDownAfterMuleContextDispose() throws Exception
     {
         super.doTearDownAfterMuleContextDispose();
-        if (domainContext != null)
+        try
         {
-            domainContext.dispose();
+            if (domainContext != null && !(domainContext.isDisposed() || !domainContext.isDisposing()))
+            {
+                domainContext.dispose();
+            }
         }
+        finally
+        {
+            domainContext = null;
+        }
+
     }
 
     /**
--- a/tests/functional/src/main/java/org/mule/tck/junit4/FunctionalTestCase.java
+++ b/tests/functional/src/main/java/org/mule/tck/junit4/FunctionalTestCase.java
@@ -7,8 +7,8 @@
 package org.mule.tck.junit4;
 
 import static org.junit.Assert.fail;
-
 import org.mule.MessageExchangePattern;
+import org.mule.api.MuleContext;
 import org.mule.api.MuleEvent;
 import org.mule.api.MuleException;
 import org.mule.api.component.Component;
@@ -47,6 +47,8 @@ import org.junit.Assert;
  */
 public abstract class FunctionalTestCase extends AbstractMuleContextTestCase
 {
+    protected static MuleContext domainContext;
+
     public FunctionalTestCase()
     {
         super();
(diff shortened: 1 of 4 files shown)


tests/functional/src/main/java/org/mule/tck/junit4/FunctionalTestCase.java
```diff
@@ -47,6 +47,8 @@
  */
 public abstract class FunctionalTestCase extends AbstractMuleContextTestCase
 {
+    protected static MuleContext domainContext;
```
asequeira: This will make two tests fails if run concurrently with different domains.

afelisatti: We need to have a getter for this, just like in DomainFunctionalTestCase.

tests/functional/src/main/java/org/mule/tck/junit4/FunctionalTestCase.java
```diff
@@ -64,26 +66,49 @@ protected String getConfigResources()
...
         }
-        configResources = getConfigFile();
-        if (configResources != null)
+        else
         {
-            if (configResources.contains(","))
+            configResources = getConfigFile();
+            if (configResources != null)
+            {
+                if (configResources.contains(","))
```
asequeira: ',' is a valid filename character, we shouldn't rely on this for deciding whether the string contains more than one filename.

tests/functional/src/main/java/org/mule/tck/junit4/FunctionalTestCase.java
```diff
@@ -64,26 +66,49 @@ protected String getConfigResources()
...
+        return configurationBuilder;
+    }
+
+    @Override
+    protected void doTearDownAfterMuleContextDispose() throws Exception
+    {
+        super.doTearDownAfterMuleContextDispose();
+        if (domainContext != null)
+        {
+            domainContext.dispose();
```
asequeira: In AbstractMuleContextTestCase muleContext is disposed only if this condition applies:

muleContext != null && !(muleContext.isDisposed() || muleContext.isDisposing()))

I don't know if there is a reason for this, but if there is we should consider it for domainContext.

tests/functional/src/main/java/org/mule/tck/junit4/FunctionalTestCase.java
```diff
@@ -64,26 +66,49 @@ protected String getConfigResources()
...
+        {
+            configurationBuilder.setDomainContext(domainContext);
+        }
+        return configurationBuilder;
+    }
+
+    @Override
+    protected void doTearDownAfterMuleContextDispose() throws Exception
+    {
+        super.doTearDownAfterMuleContextDispose();
```
asequeira: in AbstractMuleTestCase this method deletes the .mule directory, is it possible that this would cause problems when doing it before domain dispose?

tests/functional/src/main/java/org/mule/tck/junit4/FunctionalTestCase.java
afelisatti: Agreed, but this is how the method currently works. That line was just moved.

tests/functional/src/main/java/org/mule/tck/junit4/FunctionalTestCase.java
afelisatti: I can't find doTearDownAfterMuleContextDispose() in AbstractMuleTestCase. It is defined in AbstractMuleContextTestCase where it is empty.

tests/functional/src/main/java/org/mule/tck/junit4/FunctionalTestCase.java
afelisatti: I'll add the checks, thanks.

afelisatti: I think the check is made since the method is run in another annotated with "After" but itself is annotated with "AfterClass", so it runs twice and the second time the context is most likely supposed to have been disposed already. I added the check though and made sure the domainContext is set to null at the end too.

tests/functional/src/main/java/org/mule/tck/junit4/FunctionalTestCase.java
asequeira: Sorry, it's disposeContext the one that deletes .mule directory and then calls doTearDownAfterMuleContextDispose

PR #1353, opened 2015-01-09
mule-3.6.0

brs-mule:


pablokraan:
I don't know what are you trying to do, but this is not the way to do it

PR #1354, opened 2015-01-09
Fixes MULE-8217 and MULE-6427

ryandcarter:
Updates MessageVariableResolverFactory to resolve “payload” expression
via the MessageContext so that “payload” and “message.payload” are
treated equally.


vromero:
Merged, thanks for your contributions Ryan.

PR #1356, opened 2015-01-12
MULE-8213: Move packages on annotations module from org.mule to org.mule...

pablokraan:
....module.annotations


marianogonzalez:
Mule 4 is supposed to be built on the ext-api which by then should be on all its glory. Does it make sense to move this package or to completely remove it in favor of the ext-api?

pablokraan:
This change is needed to run mule inside OSGi. Probably this module will be removed in the final mule 4, but I need it now and can't wait for extensions API

PR #1357, opened 2015-01-12
MULE-8221 JmsConnector fails to start when username is configured

dfeist:


transports/jms/src/test/java/org/mule/transport/jms/vendors/ActiveMQJmsConnectorTestCase.java
```diff
@@ -105,4 +109,23 @@ public void testCustomActiveMqConnectorConfig() throws Exception
...
+
+    /**
+     * See MULE-8221
+     */
+    @Test
+    public void testActiveMqConnectorWithUsernameAndPassword() throws Exception
+    {
+        JmsConnector c = (JmsConnector) muleContext.getRegistry().lookupConnector("activeMqJmsConnectorWithUsernameAndPassword");
+
+        assertTrue(c instanceof ActiveMQJmsConnector);
```
marianogonzalez: use hamcrest

marianogonzalez: won't this fail on WMQ? This tests are imported by it and this assertion will fail

transports/jms/src/main/java/org/mule/transport/jms/JmsConnector.java
```diff
@@ -432,7 +432,7 @@ protected Connection createConnection() throws MuleException, JMSException
 
         Connection connection;
 
-        if (username != null)
+        if (!cacheJmsSessions && username != null)
```
marianogonzalez: what happens if I want to use session caching and still have authentication? Can't I cache the sessions for this user?

transports/jms/src/test/java/org/mule/transport/jms/vendors/ActiveMQJmsConnectorTestCase.java
dfeist: This test is ActiveMQ specific so this isn't an issue.

transports/jms/src/main/java/org/mule/transport/jms/JmsConnector.java
marianogonzalez: how do you prevent wmq from importing this test?

dfeist: Yes you can, but user/password are set on CachingConnectionFactory and not passed when requesting connection.  This is the cause of this bug.

transports/jms/src/test/java/org/mule/transport/jms/vendors/ActiveMQJmsConnectorTestCase.java
marianogonzalez: how do you prevent wmq from importing this test?

PR #1358, opened 2015-01-12
MULE-8220

asequeira:
Build number Maven plugin should be activated always


pablokraan:
Remember to add the jira description in the commit message

PR #1362, opened 2015-01-14
MULE-8226 JMS Durable Subscriber - unexpected behaviour in Mule HA Clust...

vromero:
...er - no failover


transports/jms/src/test/java/org/mule/transport/jms/JmsConnectorTestCase.java
```diff
@@ -199,9 +206,51 @@ public void createConnectionFactoryWrapperWhenUsingTransactionManager() throws E
         JmsConnector connector = createConnectionFactoryWhenGettingConnection(mock(TestXAConnectionFactory.class));
         assertThat(connector.getConnectionFactory(), instanceOf(DefaultXAConnectionFactoryWrapper.class));
     }
+    
+    @Test
+    public void doesNotDirtyNotificationThreadClassLoaderOnNotification() throws Exception
```
pablokraan: This tests passes with any changes in the production code. I think we need two tests: 
1) Test checking that the notification is call using the app classloader
2) Test checking that classloader is restored after notification processing

PR #1377, opened 2015-01-20
MULE-8245 Make parsing expression optional in set-payload.

dfeist:


vromero:
Expressions are pervasive and therefore this should be taken into consideration as a global thing, with special emphasis con extensions api.
Personally I would rather prefer to have something like there used to be, value for strings and expression for expressions. Anyhow, I would rather fix the specific usecase with a java transformer and or an append-string-transformer, and then contact @marianogonzalez to find out how this would fit into ext api.

PR #1378, opened 2015-01-21
MULE-8247: Including check for endpoint's autoDelete attribute before tu...

afelisatti:
...rning to connector's.


transports/sftp/src/test/java/org/mule/transport/sftp/SftpAutoDeleteTestCase.java
```diff
@@ -0,0 +1,118 @@
...
+    {
+        testDirectory(AUTO_DELETE_OFF);
+        assertTrue(Arrays.asList(sftpClient.listFiles()).contains(FILENAME));
+    }
+
+    @Test
+    public void endpointAutoDeleteTrue() throws Exception
+    {
+        testDirectory(AUTO_DELETE_ON);
+        assertFalse(Arrays.asList(sftpClient.listFiles()).contains(FILENAME));
```
marianogonzalez: hamcrest

transports/sftp/src/test/java/org/mule/transport/sftp/SftpAutoDeleteTestCase.java
```diff
@@ -0,0 +1,118 @@
...
+        latch = new Latch();
+        sftpClient.changeWorkingDirectory("..");
+        sftpClient.mkdir(AUTO_DELETE_OFF);
+        sftpClient.mkdir(AUTO_DELETE_ON);
+    }
+
+    @After
+    public void tearDownData() throws IOException
+    {
+        sftpClient.changeWorkingDirectory("..");
```
marianogonzalez: here you're assuming that the client succesfully changed directory on the first line of testDirectory(String)

transports/sftp/src/test/java/org/mule/transport/sftp/SftpAutoDeleteTestCase.java
```diff
@@ -0,0 +1,118 @@
...
+        assertFalse(Arrays.asList(sftpClient.listFiles()).contains(FILENAME));
+    }
+
+    private void testDirectory(String directory) throws Exception
+    {
+        sftpClient.changeWorkingDirectory(directory);
+        sftpClient.storeFile(FILENAME, new ByteArrayInputStream(FILE_CONTENT.getBytes()));
+        MuleClient muleClient = muleContext.getClient();
+        muleClient.dispatch(getUrl(directory), new DefaultMuleMessage(TEST_MESSAGE, muleContext));
+        latch.await(1500, TimeUnit.MILLISECONDS);
```
marianogonzalez: 1500 -> constant

transports/sftp/src/test/resources/mule-sftp-auto-delete-implicit-connector.xml
```diff
@@ -0,0 +1,27 @@
+<mule xmlns="http://www.mulesoft.org/schema/mule/core" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
marianogonzalez: this file is almost identical to the other one. Use a spring profile or at the very least an import

transports/sftp/src/main/java/org/mule/transport/sftp/SftpReceiverRequesterUtil.java
```diff
@@ -213,7 +213,17 @@ public InputStream retrieveFile(String fileName, SftpNotifier notifier) throws E
         // This special InputStream closes the SftpClient when the stream is closed.
         // The stream will be materialized in a Message Dispatcher or Service
         // Component
-        return new SftpInputStream(client, fileInputStream, fileName, connector.isAutoDelete(), endpoint);
+        return new SftpInputStream(client, fileInputStream, fileName, determineAutoDelete(), endpoint);
+    }
+
+    private boolean determineAutoDelete()
+    {
+        Boolean autoDelete = Boolean.valueOf((String) endpoint.getProperty("autoDelete"));
```
marianogonzalez: this is wrong. valueOf() will never return null. On the contrary, it will return false if the input is null.

The null check you actually have to do is against the endpoint property

PR #1393, opened 2015-01-27
MULE-8270: HttpRequester not configuring TLS cipher suites and protocols

andresgregoire:


modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerTlsRestrictedProtocolsAndCiphers.java
```diff
@@ -0,0 +1,127 @@
...
+import javax.net.ssl.SSLException;
+import javax.net.ssl.SSLSocket;
+import javax.net.ssl.SSLSocketFactory;
+
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class HttpListenerTlsRestrictedProtocolsAndCiphers extends FunctionalTestCase
```
dfeist: Test case class name should end with "TestCase"

modules/http/src/test/java/org/mule/module/http/functional/requester/HttpRequesterTlsRestrictedProtocolsAndCiphers.java
```diff
@@ -0,0 +1,167 @@
...
+import org.eclipse.jetty.server.ServerConnector;
+import org.eclipse.jetty.server.handler.AbstractHandler;
+import org.eclipse.jetty.util.ssl.SslContextFactory;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
+
+
+public class HttpRequesterTlsRestrictedProtocolsAndCiphers extends FunctionalTestCase
```
dfeist: Same again.

modules/http/src/main/java/org/mule/module/http/internal/request/grizzly/TlsTransportCustomizer.java
```diff
@@ -0,0 +1,53 @@
...
+import com.ning.http.client.providers.grizzly.TransportCustomizer;
+
+import org.glassfish.grizzly.filterchain.FilterChainBuilder;
+import org.glassfish.grizzly.nio.transport.TCPNIOTransport;
+import org.glassfish.grizzly.ssl.SSLFilter;
+
+/**
+ * Transport customizer that sets enabled cipher suites and protocols in the SSL filter.
+ */
+public class TlsTransportCustomizer implements TransportCustomizer
```
dfeist: I don't like this approach, but it's either this or extend (and duplicate a lot of code) from com.ning.http.client.providers.grizzly.GrizzlyAsyncHttpProvider.

pablolagreca: I discussed this with Andres. I don't see any approach being better. We have now automated test cases that verify the customization done by TlsTransportCustomizer so anyway, if AHC change something those test cases are going to fail.

PR #1398, opened 2015-01-30
MULE-8278 HTTP Listener does not handle "Expect: 100-Continue" (RFC 2616...

dfeist:
...) header correctly

### DIFF
--- a/modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/GrizzlyRequestDispatcherFilter.java
+++ b/modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/GrizzlyRequestDispatcherFilter.java
@@ -8,8 +8,6 @@ package org.mule.module.http.internal.listener.grizzly;
 
 import static org.mule.module.http.api.HttpConstants.Protocols.HTTP;
 import static org.mule.module.http.api.HttpConstants.Protocols.HTTPS;
-
-import org.mule.module.http.api.HttpConstants;
 import org.mule.module.http.internal.domain.InputStreamHttpEntity;
 import org.mule.module.http.internal.domain.request.HttpRequestContext;
 import org.mule.module.http.internal.domain.response.HttpResponse;
@@ -26,6 +24,9 @@ import org.glassfish.grizzly.filterchain.FilterChainContext;
 import org.glassfish.grizzly.filterchain.NextAction;
 import org.glassfish.grizzly.http.HttpContent;
 import org.glassfish.grizzly.http.HttpRequestPacket;
+import org.glassfish.grizzly.http.HttpResponsePacket;
+import org.glassfish.grizzly.http.util.Header;
+import org.glassfish.grizzly.http.util.HttpStatus;
 
 /**
  * Grizzly filter that dispatches the request to the right request handler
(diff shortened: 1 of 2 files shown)


modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/GrizzlyRequestDispatcherFilter.java
```diff
@@ -49,6 +50,26 @@ public NextAction handleRead(final FilterChainContext ctx) throws IOException
         final HttpContent httpContent = ctx.getMessage();
         final HttpRequestPacket request = (HttpRequestPacket) httpContent.getHttpHeader();
 
+        // Handle Expect Continue
+        if (request.requiresAcknowledgement())
+        {
+            final HttpResponsePacket.Builder responsePacketBuilder = HttpResponsePacket.builder(request);
+            if ("100-continue".equalsIgnoreCase(request.getHeader(Header.Expect)))
```
andresgregoire: Use constant defined in HttpHeaders.Values for the value of the header. Maybe we should also use the Expect header name defined in HttpHeaders.Names? Use static imports.

modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerExceptContinueTestCase.java
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+
+public class HttpListenerExceptContinueTestCase extends FunctionalTestCase
+{
+
+    @Rule
+    public DynamicPort listenPort = new DynamicPort("port");
+    private String flowReceivedMessage;
+    private String expectHeader;
```
andresgregoire: These two instance variables are not used

modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerExceptContinueTestCase.java
```diff
@@ -0,0 +1,60 @@
...
+            @Override
+            public void eventReceived(MuleEventContext context, Object component) throws Exception
+            {
+                assertThat(context.getMessageAsString(), is(TEST_MESSAGE));
+            }
+        });
+
+        final HttpRequestOptions requestOptions = newOptions().method(POST.name()).build();
+        MuleMessage message = new DefaultMuleMessage(new ByteArrayInputStream(TEST_MESSAGE.getBytes()), muleContext);
+        message.setOutboundProperty(Header.Expect.name(), "100-Continue");
```
andresgregoire: The Header class is not imported in this test. Use constant from HttpConstants.Values for the value (and again, maybe use constant from HttpConstants.Names instead of the constant from Grizzly).

modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerExceptContinueTestCase.java
```diff
@@ -0,0 +1,60 @@
...
+import org.mule.tck.functional.EventCallback;
+import org.mule.tck.junit4.FunctionalTestCase;
+import org.mule.tck.junit4.rule.DynamicPort;
+
+import java.io.ByteArrayInputStream;
+
+import org.junit.Rule;
+import org.junit.Test;
+
+public class HttpListenerExceptContinueTestCase extends FunctionalTestCase
```
andresgregoire: Typo in class name Except -> Expect

PR #1408, opened 2015-02-03
MULE-8290 - adding logic to read http listener request path property

pablolagreca:


transports/http/src/main/java/org/mule/transport/http/components/StaticResourceMessageProcessor.java
```diff
@@ -61,7 +62,24 @@ public MuleEvent process(MuleEvent event) throws MuleException
         }
 
         String path = event.getMessage().getInboundProperty(HttpConnector.HTTP_REQUEST_PATH_PROPERTY);
-        String contextPath = event.getMessage().getInboundProperty(HttpConnector.HTTP_CONTEXT_PATH_PROPERTY);
+        String contextPath = event.getMessage().getInboundProperty(org.mule.module.http.api.HttpConstants.RequestProperties.HTTP_LISTENER_PATH);
```
dfeist: Is http transprot still supported? We should continue to support it until it's removed but I'm unsure if this change does this.

pablolagreca: Yes Dan, we do. Now I check first for the http listener property, and if it doesn't exists I use the property from the old transport.

There's a test case for the transport functionality validating that it continues to work as before.

PR #1412, opened 2015-02-05
MULE-8296: Undeployment does NOT release pooled database connections

pablokraan:


modules/db/src/main/java/org/mule/module/db/internal/domain/database/GenericDbConfig.java
```diff
@@ -46,9 +48,11 @@
  * Defines a database configuration that is not customized for any particular
  * database vendor
  */
-public class GenericDbConfig implements DbConfig, Initialisable
+public class GenericDbConfig implements DbConfig, Initialisable, Disposable
 {
 
+    protected static Log logger = LogFactory.getLog(GenericDbConfig.class);
```
marianogonzalez: use slf4j

modules/db/src/test/java/org/mule/module/db/internal/config/domain/database/DbConfigFactoryBeanTestCase.java
```diff
@@ -0,0 +1,48 @@
...
+import org.junit.Test;
+
+@SmallTest
+public class DbConfigFactoryBeanTestCase extends AbstractMuleTestCase
+{
+
+    @Test
+    public void disposesDbConfig() throws Exception
+    {
+        final GenericDbConfig dbConfig = spy(new GenericDbConfig(null, "test", null));
```
marianogonzalez: why a spy? if all you are verifying is that dbConfig.dispose() is being invoked, you could use a simple mock

PR #1416, opened 2015-02-06
MULE-8292: HTTP Listener doesn't validate clients

andresgregoire:


modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/GrizzlyServerManager.java
```diff
@@ -227,7 +225,8 @@ private SSLFilter createSslFilter(final TlsContextFactory tlsContextFactory)
     {
         try
         {
-            final SSLEngineConfigurator serverConfig = new SSLEngineConfigurator(tlsContextFactory.createSslContext(), false, false, false);
+            boolean clientAuth = tlsContextFactory.isTrustStoreConfigured();
+            final SSLEngineConfigurator serverConfig = new SSLEngineConfigurator(tlsContextFactory.createSslContext(), false, clientAuth, false);
```
andresgregoire: We need to define if we keep wantClientAuth=false (the last argument of the SSLEngineConstructor), or we use clientAuth.

PR #1420, opened 2015-02-09
MULE-8304: It is not possible to define more than one keystore/truststore

andresgregoire:


modules/http/src/test/java/org/mule/module/http/functional/requester/AbstractHttpRequestTestCase.java
```diff
@@ -98,7 +98,7 @@ private void enableHttpsServer(Server server)
 
         try
         {
-            sslContextFactory.setKeyStorePath(getKeyStorePath());
+            sslContextFactory.setKeyStorePath(FileUtils.getResourcePath("serverKeystore", getClass()));
```
dfeist: static?

PR #1424, opened 2015-02-10
MULE-8306: Add a LeakCleaner interface in order to allow the MuleApplica...

marcosnc:
...tionClassLoader to perform cleanup tasks, in particular to deregister DB drivers from DriverManager when MuleApplicationClassloader is disposed.

### DIFF
--- a/modules/launcher/src/main/java/org/mule/module/launcher/LeakCleaner.java
+++ b/modules/launcher/src/main/java/org/mule/module/launcher/LeakCleaner.java
@@ -1,0 +1,16 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.module.launcher;
+
+/**
+ * Implementations of this class should take care about objects that may leak memory after application undeployment.
+ */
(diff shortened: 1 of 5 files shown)


modules/launcher/src/main/java/org/mule/module/launcher/LeakCleaner.java
```diff
@@ -0,0 +1,16 @@
...
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.module.launcher;
+
+/**
+ * Implementations of this class should take care about objects that may leak memory after application undeployment.
+ */
+public interface LeakCleaner
```
dfeist: LeakCleaner isn't a good name.  You are cleaning something, so it can't become a leak.

modules/launcher/src/main/java/org/mule/module/launcher/MuleApplicationClassLoader.java
```diff
@@ -46,6 +49,8 @@
 
     protected static final URL[] CLASSPATH_EMPTY = new URL[0];
 
+    private static final String DEFAULT_LEAK_CLEANER_CLASS = "DefaultLeakCleaner.class";
```
dfeist: Why? Can't you either specify implementation directly or if that doen't make sense use Service Provider approach.

PR #1427, opened 2015-02-10
MULE-8306: Add a LeakCleaner interface in order to allow the artifacts (...

marcosnc:
...applications and domains) to perform cleanup tasks, in particular to deregister DB drivers from DriverManager when the artifact is disposed.

### DIFF
--- a/modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
+++ b/modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
@@ -9,7 +9,10 @@ package org.mule.module.launcher.artifact;
 import org.mule.module.launcher.DirectoryResourceLocator;
 import org.mule.module.launcher.FineGrainedControlClassLoader;
 import org.mule.module.launcher.LocalResourceLocator;
+import org.mule.util.IOUtils;
+import org.mule.util.StringUtils;
 
+import java.io.InputStream;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -24,11 +27,17 @@ import org.apache.commons.logging.LogFactory;
  */
 public abstract class AbstractArtifactClassLoader extends FineGrainedControlClassLoader implements ArtifactClassLoader
 {
+    private static final String DEFAULT_LEAK_CLEANER_CLASS = "DefaultLeakCleaner.class";
+
     protected Log logger = LogFactory.getLog(getClass());
 
     protected List<ShutdownListener> shutdownListeners = new ArrayList<ShutdownListener>();
 
     private LocalResourceLocator localResourceLocator;
+
+    private LeakCleaner leakCleaner;
+
+    private String leakCleanerClassName;
 
     public AbstractArtifactClassLoader(URL[] urls, ClassLoader parent)
     {
@@ -60,7 +69,60 @@ public abstract class AbstractArtifactClassLoader extends FineGrainedControlClas
                 logger.error(e);
             }
         }
+
+        LeakCleaner leakCleaner = getLeakCleanerInstance();
+        if (leakCleaner != null)
+        {
+            leakCleaner.clean();
+        }
+
         super.dispose();
+    }
+
+    public String getLeakCleanerClassName()
+    {
+        return StringUtils.defaultString(leakCleanerClassName, DEFAULT_LEAK_CLEANER_CLASS);
+    }
+
+    public void setLeakCleanerClassName(String leakCleanerClass)
+    {
+        this.leakCleanerClassName = leakCleanerClass;
+        leakCleaner = null;
+    }
+
+    public LeakCleaner getLeakCleanerInstance()
+    {
+        if(leakCleaner==null)
+        {
+            leakCleaner = createLeakCleanerInstance(getLeakCleanerClassName());
+        }
+        return leakCleaner;
+    }
+
+    private LeakCleaner createLeakCleanerInstance(String className)
+    {
+        InputStream classStream = this.getClass().getResourceAsStream(className);
+        if (classStream == null)
+        {
+            classStream = AbstractArtifactClassLoader.class.getResourceAsStream(className);
+        }
+        if (classStream == null)
+        {
+            logger.warn("Could not find leak cleaner class with name: " + className);
+            return null;
+        }
+
+        byte[] classBytes = IOUtils.toByteArray(classStream);
+        Class clazz = this.defineClass(null, classBytes, 0, classBytes.length);
+        try
+        {
+            return (LeakCleaner) clazz.newInstance();
+        }
+        catch (Exception e)
+        {
+            logger.warn("Could not instantiate leak cleaner instance of type: " + className, e);
+        }
+        return null;
     }
 
     public URL findLocalResource(String resourceName)
--- a/modules/launcher/src/main/java/org/mule/module/launcher/artifact/DefaultLeakCleaner.java
+++ b/modules/launcher/src/main/java/org/mule/module/launcher/artifact/DefaultLeakCleaner.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.module.launcher.artifact;
+
+import java.sql.Driver;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.util.Enumeration;
+
+public class DefaultLeakCleaner implements LeakCleaner
+{
+
+    @Override
+    public void clean() {
+        unregisterDbDrivers();
+    }
+
+    private void unregisterDbDrivers()
+    {
+        Enumeration<Driver> drivers = DriverManager.getDrivers();
+        while (drivers.hasMoreElements()) {
+            Driver driver = drivers.nextElement();
+            try {
+                DriverManager.deregisterDriver(driver);
+            }
+            catch (SQLException e)
+            {
+                throw new RuntimeException("Can not deregister driver: " + driver.getClass(), e);
+            }
+        }
+    }
+
+}
--- a/modules/launcher/src/main/java/org/mule/module/launcher/artifact/LeakCleaner.java
+++ b/modules/launcher/src/main/java/org/mule/module/launcher/artifact/LeakCleaner.java
@@ -1,0 +1,16 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.module.launcher.artifact;
+
+/**
+ * Implementations of this class should take care about objects that may leak memory after application undeployment.
+ */
+public interface LeakCleaner
+{
+
+    void clean();
+}
--- a/modules/launcher/src/test/java/org/mule/module/launcher/MuleApplicationClassLoaderTestCase.java
+++ b/modules/launcher/src/test/java/org/mule/module/launcher/MuleApplicationClassLoaderTestCase.java
@@ -6,17 +6,21 @@
  */
 package org.mule.module.launcher;
 
+import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 import org.mule.api.config.MuleProperties;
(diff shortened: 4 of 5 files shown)


modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
```diff
@@ -24,12 +27,18 @@
  */
 public abstract class AbstractArtifactClassLoader extends FineGrainedControlClassLoader implements ArtifactClassLoader
 {
+    private static final String DEFAULT_LEAK_CLEANER_CLASS = "DefaultLeakCleaner.class";
```
pablokraan: Can you use a full qualified class name instead? (ie: org.mule.module.launcher.artifact.DefaultLeakCleaner)

modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
```diff
@@ -60,9 +69,62 @@ public void dispose()
...
+        {
+            leakCleaner.clean();
+        }
+
         super.dispose();
     }
 
+    public String getLeakCleanerClassName()
+    {
+        return StringUtils.defaultString(leakCleanerClassName, DEFAULT_LEAK_CLEANER_CLASS);
```
pablokraan: Set the default value in leakCleanerClassName

modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
```diff
@@ -24,12 +27,18 @@
...
 {
+    private static final String DEFAULT_LEAK_CLEANER_CLASS = "DefaultLeakCleaner.class";
+
     protected Log logger = LogFactory.getLog(getClass());
 
     protected List<ShutdownListener> shutdownListeners = new ArrayList<ShutdownListener>();
 
     private LocalResourceLocator localResourceLocator;
 
+    private LeakCleaner leakCleaner;
```
pablokraan: I don't see any reason why there should be a LeakCleaner field because that instance is used only during classlaoder disponsig

modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
```diff
@@ -60,9 +69,62 @@ public void dispose()
...
+    public LeakCleaner getLeakCleanerInstance()
+    {
+        if(leakCleaner==null)
+        {
+            leakCleaner = createLeakCleanerInstance(getLeakCleanerClassName());
+        }
+        return leakCleaner;
+    }
+
+    private LeakCleaner createLeakCleanerInstance(String className)
```
pablokraan: This method can be protected instead, so you can inject a mock LeakCleaner for testing purposes

modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
```diff
@@ -60,9 +69,62 @@ public void dispose()
...
+    {
+        InputStream classStream = this.getClass().getResourceAsStream(className);
+        if (classStream == null)
+        {
+            classStream = AbstractArtifactClassLoader.class.getResourceAsStream(className);
+        }
+        if (classStream == null)
+        {
+            logger.warn("Could not find leak cleaner class with name: " + className);
```
pablokraan: I think this is an error that we should throw, not just a warning in the log

modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
```diff
@@ -60,9 +69,62 @@ public void dispose()
...
+
+        byte[] classBytes = IOUtils.toByteArray(classStream);
+        Class clazz = this.defineClass(null, classBytes, 0, classBytes.length);
+        try
+        {
+            return (LeakCleaner) clazz.newInstance();
+        }
+        catch (Exception e)
+        {
+            logger.warn("Could not instantiate leak cleaner instance of type: " + className, e);
```
pablokraan: Idem

modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
```diff
@@ -60,9 +69,62 @@ public void dispose()
...
+
+    public void setLeakCleanerClassName(String leakCleanerClass)
+    {
+        this.leakCleanerClassName = leakCleanerClass;
+        leakCleaner = null;
+    }
+
+    public LeakCleaner getLeakCleanerInstance()
+    {
+        if(leakCleaner==null)
```
pablokraan: Fix code style

modules/launcher/src/main/java/org/mule/module/launcher/artifact/DefaultLeakCleaner.java
```diff
@@ -0,0 +1,37 @@
...
+
+    @Override
+    public void clean() {
+        unregisterDbDrivers();
+    }
+
+    private void unregisterDbDrivers()
+    {
+        Enumeration<Driver> drivers = DriverManager.getDrivers();
+        while (drivers.hasMoreElements()) {
```
pablokraan: Fix code style

modules/launcher/src/main/java/org/mule/module/launcher/artifact/DefaultLeakCleaner.java
```diff
@@ -0,0 +1,37 @@
...
+            Driver driver = drivers.nextElement();
+            try {
+                DriverManager.deregisterDriver(driver);
+            }
+            catch (SQLException e)
+            {
+                throw new RuntimeException("Can not deregister driver: " + driver.getClass(), e);
```
pablokraan: This should be a warning in the logs, not an error. Something like "Unable to dispose driver..... This can cause a memory leak"

modules/launcher/src/main/java/org/mule/module/launcher/artifact/LeakCleaner.java
```diff
@@ -0,0 +1,16 @@
...
+ */
+package org.mule.module.launcher.artifact;
+
+/**
+ * Implementations of this class should take care about objects that may leak memory after application undeployment.
+ */
+public interface LeakCleaner
+{
+
+    void clean();
```
pablokraan: I think this method should not throw any exception because if there is an error during memory leak cleaning, that is not an application error per se, and that must be clear in the method's javadoc

modules/launcher/src/main/java/org/mule/module/launcher/artifact/LeakCleaner.java
```diff
@@ -0,0 +1,16 @@
...
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.module.launcher.artifact;
+
+/**
+ * Implementations of this class should take care about objects that may leak memory after application undeployment.
+ */
+public interface LeakCleaner
```
pablokraan: Not sure that LeakCleaner is the right name because is not going to clean every memory leak. Is something like memory leak prevention.

modules/launcher/src/test/java/org/mule/module/launcher/MuleApplicationClassLoaderTestCase.java
```diff
@@ -98,6 +102,32 @@ public void cleanUp()
     }
 
     @Test
+    public void leakCleanerIsCreatedByCorrectClassLoaderApp() throws Exception
```
pablokraan: Can you move these test into a new test class? (current test class has a setup method that builds a complete classloader seting up folders and stuff)

modules/launcher/src/main/java/org/mule/module/launcher/artifact/LeakCleaner.java
dfeist: Agree, you prevent leaks not clean them.  Something like ResourceCloser, or ResourceReleaser or something..

modules/launcher/src/test/java/org/mule/module/launcher/MuleApplicationClassLoaderTestCase.java
```diff
@@ -98,6 +102,32 @@ public void cleanUp()
...
+    private void ensureLeakCleanerIsCreatedByCorrectClassLoader(AbstractArtifactClassLoader artifactClassLoader) throws Exception
+    {
+        assertThat(artifactClassLoader.getClass().getClassLoader(), is(Thread.currentThread().getContextClassLoader()));
+        artifactClassLoader.setLeakCleanerClassName("TestLeakCleaner.class");
+        artifactClassLoader.dispose();
+
+        // We must call the getClassLoaderClassName method from TestLeakCleaner dynamically in order to not load the
+        // class by the current class loader if not a java.lang.LinkageError is raised.
+        Method getClassLoaderMethod = artifactClassLoader.getLeakCleanerInstance().getClass().getMethod("getClassLoader");
+        ClassLoader classLoader = (ClassLoader) getClassLoaderMethod.invoke(artifactClassLoader.getLeakCleanerInstance());
```
pablokraan: Instead of defining getLeakCleanerInstance in the AbstractArtifactClasLoader, you can create a subclass of it and put that method there and use that instance on the test

modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
marcosnc: I tried to use it, but it didn't work. Here the class is just a resource, that's why I need to put the suffix .class.

modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
marcosnc: Using it this way you ensure that setting leakCleanerClassName to null with the setter won't remove the leak cleaning, but i'm not sure if we have to take care of that, so I think it's ok to set the default value to the field definition.

modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
marcosnc: I did it in this way for testing purposes. Other way can't verify how that instance was created.

modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
marcosnc: The method can be protected, but I can not inject the LeakCleaner, that instance must be created in this class loader in this strange way. If I inject it from outside the whole trick won't work.

modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
marcosnc: Could be, but if we raise an error the entire dispose mechanism will fail, and here we want to improve the cleaning. If we raise an error the leak will continue and other resources could remains in memory too.

modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
marcosnc: Uhh... missed this! Thanks!

modules/launcher/src/main/java/org/mule/module/launcher/artifact/DefaultLeakCleaner.java
marcosnc: Again...

modules/launcher/src/main/java/org/mule/module/launcher/artifact/DefaultLeakCleaner.java
marcosnc: You are right but... I don't know If we can use logs here... This class is injected into the class loader and I'm not sure if the log mechanism will work. Will try and tell you.

modules/launcher/src/main/java/org/mule/module/launcher/artifact/LeakCleaner.java
marcosnc: Agree, may be I could pass the logger to this method if it can not be instantiated in the concrete class.

modules/launcher/src/main/java/org/mule/module/launcher/artifact/LeakCleaner.java
marcosnc: Ok, sounds good.

modules/launcher/src/test/java/org/mule/module/launcher/MuleApplicationClassLoaderTestCase.java
marcosnc: Ok.

PR #1432, opened 2015-02-11
MULE-8306: Add a ResourceReleaser interface in order to allow the artifa...

marcosnc:
...cts (applications and domains) to perform cleanup tasks, in particular to deregister DB drivers from DriverManager when the artifact is disposed.

### DIFF
--- a/modules/launcher/src/main/java/org/mule/module/launcher/MuleSharedDomainClassLoader.java
+++ b/modules/launcher/src/main/java/org/mule/module/launcher/MuleSharedDomainClassLoader.java
@@ -84,6 +84,15 @@ public class MuleSharedDomainClassLoader extends AbstractArtifactClassLoader imp
         }
     }
 
+    /**
+     * For testing purposes
+     */
+    protected MuleSharedDomainClassLoader(ClassLoader parent)
+    {
+        super(new URL[0], parent);
+        domain = null;
+    }
+
     @Override
     public String toString()
     {
--- a/modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
+++ b/modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
@@ -9,7 +9,9 @@ package org.mule.module.launcher.artifact;
 import org.mule.module.launcher.DirectoryResourceLocator;
 import org.mule.module.launcher.FineGrainedControlClassLoader;
 import org.mule.module.launcher.LocalResourceLocator;
+import org.mule.util.IOUtils;
 
+import java.io.InputStream;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -24,11 +26,16 @@ import org.apache.commons.logging.LogFactory;
  */
 public abstract class AbstractArtifactClassLoader extends FineGrainedControlClassLoader implements ArtifactClassLoader
 {
+
+    private static final String DEFAULT_RESOURCE_RELEASER_CLASS_LOCATION = "/org/mule/module/launcher/artifact/DefaultResourceReleaser.class";
+
     protected Log logger = LogFactory.getLog(getClass());
 
     protected List<ShutdownListener> shutdownListeners = new ArrayList<ShutdownListener>();
 
     private LocalResourceLocator localResourceLocator;
+
+    private String resourceReleaserClassLocation = DEFAULT_RESOURCE_RELEASER_CLASS_LOCATION;
 
     public AbstractArtifactClassLoader(URL[] urls, ClassLoader parent)
     {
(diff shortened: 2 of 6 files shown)


modules/launcher/src/main/java/org/mule/module/launcher/MuleSharedDomainClassLoader.java
```diff
@@ -84,6 +84,15 @@ public MuleSharedDomainClassLoader(String domain, ClassLoader parent)
         }
     }
 
+    /**
+     * For testing purposes
+     */
+    protected MuleSharedDomainClassLoader(ClassLoader parent)
```
pablokraan: Adding things "for testing purposes" is not a good idea, even worst when you add a constructor that creates an incomplete/invalid instance

modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
```diff
@@ -60,9 +67,32 @@ public void dispose()
...
+        try
+        {
+            InputStream classStream = this.getClass().getResourceAsStream(resourceReleaserClassLocation);
+            byte[] classBytes = IOUtils.toByteArray(classStream);
+            Class clazz = this.defineClass(null, classBytes, 0, classBytes.length);
+            return (ResourceReleaser) clazz.newInstance();
+        }
+        catch (Exception e)
+        {
+            throw new RuntimeException("Can not create resource releaser instance of type: " + resourceReleaserClassLocation, e);
```
pablokraan: "instance of type" -> "instance from resource"

modules/launcher/src/main/java/org/mule/module/launcher/artifact/DefaultResourceReleaser.java
```diff
@@ -0,0 +1,45 @@
...
+
+    private Logger logger = LoggerFactory.getLogger(getClass());
+
+    @Override
+    public void release()
+    {
+        deregisterDrivers();
+    }
+
+    private void deregisterDrivers()
```
pablokraan: Rename to deregisterJdbcDrivers

modules/launcher/src/main/java/org/mule/module/launcher/artifact/DefaultResourceReleaser.java
```diff
@@ -0,0 +1,45 @@
...
+        {
+            Driver driver = drivers.nextElement();
+            try
+            {
+                logger.debug("Deregistering driver: " + driver.getClass());
+                DriverManager.deregisterDriver(driver);
+            }
+            catch (Exception e)
+            {
+                logger.warn("Can not deregister driver: " + driver.getClass(), e);
```
pablokraan: Showing the exception here will add noise to the log because user will be probably unable to change anything to fix it. Maybe the exception could be added when DEBUG level is enabled only.

modules/launcher/src/main/java/org/mule/module/launcher/artifact/ResourceReleaser.java
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Implementations of this class should take care about resources that may leak memory after application undeployment.
+ * Mule ensures to create an instance of this class with the same class loader that loaded the application resources
+ * in order to ensure the access to them.
+ */
+public interface ResourceReleaser
+{
+
+    /**
+     * This method will be called as part of the application undeployment.
```
pablokraan: "This method will be called as part of the application undeployment." -> "Attempts to release, during a {@link Artifact} undeployment, resouces that were not explicilty released and could cause a memory leak"

modules/launcher/src/main/java/org/mule/module/launcher/artifact/ResourceReleaser.java
```diff
@@ -0,0 +1,23 @@
...
+     * No exceptions should be thrown due to any issues while cleaning resources are not related to the application
+     * itself but with resource housekeeping.
+     */
+    void release();
```
pablokraan: It could be better to wrap calls to release method inside a try/catch instead of relying on the javadoc

modules/launcher/src/test/java/org/mule/module/launcher/ResourceReleaserTestCase.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private void ensureResourceReleaserIsCreatedByCorrectClassLoader(AbstractArtifactClassLoader classLoader) throws Exception
+    {
+        assertThat(classLoader.getClass().getClassLoader(), is(Thread.currentThread().getContextClassLoader()));
+        classLoader.setResourceReleaserClassLocation(TEST_RESOURCE_RELEASER_CLASS_LOCATION);
+        classLoader.dispose();
+
+        // We must call the getClassLoader method from TestResourceReleaser dynamically in order to not load the
+        // class by the current class loader, if not a java.lang.LinkageError is raised.
+        ResourceReleaser resourceReleaserInstance = ((KeepResourceReleaserInstance)classLoader).getResourceReleaserInstance();
```
pablokraan: fix code style

PR #1433, opened 2015-02-12
MULE-8306: Add a ResourceReleaser interface in order to allow the artifa...

marcosnc:
...cts (applications and domains) to perform cleanup tasks, in particular to deregister DB drivers from DriverManager when the artifact is disposed.

### DIFF
--- a/modules/launcher/src/main/java/org/mule/module/launcher/MuleSharedDomainClassLoader.java
+++ b/modules/launcher/src/main/java/org/mule/module/launcher/MuleSharedDomainClassLoader.java
@@ -44,35 +44,7 @@ public class MuleSharedDomainClassLoader extends AbstractArtifactClassLoader imp
 
             validateAndGetDomainFolders();
 
-            addURL(domainLibraryFolder.getParentFile().toURI().toURL());
-
-            if (domainLibraryFolder.exists())
-            {
-                Collection<File> jars = FileUtils.listFiles(domainLibraryFolder, new String[] {"jar"}, false);
-
-                if (logger.isDebugEnabled())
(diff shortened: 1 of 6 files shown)


modules/launcher/src/main/java/org/mule/module/launcher/MuleSharedDomainClassLoader.java
```diff
@@ -158,6 +130,39 @@ private void validateAndGetDomainFolders() throws Exception
...
+    {
+        addURL(domainLibraryFolder.getParentFile().toURI().toURL());
+
+        if (domainLibraryFolder.exists())
+        {
+            Collection<File> jars = FileUtils.listFiles(domainLibraryFolder, new String[] {"jar"}, false);
+
+            if (logger.isDebugEnabled())
+            {
+                {
```
pablokraan: Remove this redundant block

modules/launcher/src/main/java/org/mule/module/launcher/artifact/DefaultResourceReleaser.java
```diff
@@ -0,0 +1,45 @@
...
+        {
+            Driver driver = drivers.nextElement();
+            try
+            {
+                logger.debug("Deregistering driver: " + driver.getClass());
+                DriverManager.deregisterDriver(driver);
+            }
+            catch (Exception e)
+            {
+                logger.warn("Can not deregister driver: " + driver.getClass(), e);
```
pablokraan: Can you change it to something like "Can not deregister driver %s. This can cause a memory leak"

PR #1441, opened 2015-02-18
MULE-8310: FTP client not timing out

pablokraan:


transports/ftp/src/main/resources/META-INF/mule-ftp.xsd
```diff
@@ -153,6 +153,13 @@
                 </xsd:documentation>
             </xsd:annotation>
         </xsd:attribute>
+        <xsd:attribute name="connectionTimeout" type="mule:substitutableInt">
```
dfeist: By having this attirbute here it's available on endpoints too.

PR #1442, opened 2015-02-18
MULE-8061: Adding message to create MuleMessages for testing and replaci...

afelisatti:
...ng repeated uses with it.


core/src/test/java/org/mule/tck/junit4/AbstractMuleContextTestCase.java
```diff
@@ -406,6 +406,15 @@ protected MuleMessage getTestMuleMessage()
         return new DefaultMuleMessage(TEST_PAYLOAD, muleContext);
     }
 
+    /**
+     * @param message
+     * @return creates a new {@link org.mule.api.MuleMessage} with message as payload
+     */
+    protected MuleMessage getTestMuleMessage(Object message)
+    {
+        return new DefaultMuleMessage(message, muleContext);
```
pablokraan: There is already a getTestMuleMessage method that does the same using a default message. That method should invoke the new one passing that default value

PR #1462, opened 2015-03-04
[DO NOT MERGE - FOR REVIEW ONLY] Spring lifecycle

marianogonzalez:
IF YOU MERGE -> YOU BRING PASTRIES


core/src/main/java/org/mule/api/registry/InitialisingRegistry.java
```diff
@@ -0,0 +1,50 @@
...
+
+import org.mule.api.MuleException;
+
+/**
+ * A {@link Registry} which not only registers and fetches objects,
+ * but it's also capable of applying lifecycle and injects dependencies
+ *
+ * @since 3.7.0
+ */
+public interface InitialisingRegistry extends Registry
```
dfeist: Improve name, because it's doesn't just to initialise, it does all lifecycle phases. MagicBucketRegistry ? ;-)

core/src/main/java/org/mule/api/registry/ObjectLimbo.java
```diff
@@ -0,0 +1,55 @@
...
+ * Those key/value pairs are held into this limbo until one
+ * registry is initialised and is kind enough to take ownership
+ * over these objects. Limbo objects should not be added into more than
+ * one registry since that may lead to inconsistencies. It is up to the
+ * registry to invoke {@link #clear()} upon registration to make sure
+ * that doesn't happen
+ *
+ * @since 3.7.0
+ */
+public interface ObjectLimbo
```
dfeist: I'm unsure why you can't use a temporary SimpleRegistry implementation whose contents then get copied to SpringRegistry before it's removed, rather than needing to introduce this concept and associated classes.

core/src/main/java/org/mule/context/DefaultMuleContextBuilder.java
```diff
@@ -92,6 +95,25 @@
...
+    /**
+     * if {@code createSimpleRegistry} is {@code true}, then a {@link SimpleRegistry}
+     * will automatically be added to the created context. That only makes sense
+     * when trying to start a light weight context which doesn't require starting a heavier
+     * registry such as Spring or Guice or when testing. This is not something we would
+     * advice on production
+     *
+     * @param createSimpleRegistry whether or not to add a {@link SimpleRegistry} to the created context
+     */
+    public DefaultMuleContextBuilder(boolean createSimpleRegistry)
```
dfeist: Don't like this boolean here, feel it should be transparent.  MuleContext shoud have a SimpleRegisty/Limbo by default when it's created which gets replaces by SpringRegistry if it happens to be configured with Spring maybe?  Obisouly multiple registries made this easier, but that has it's own problems.

core/src/main/java/org/mule/registry/DefaultRegistryBroker.java
```diff
@@ -7,41 +7,63 @@
...
+/**
+ * @deprecated as of 3.7.0. This will be removed in Mule 4.0
+ */
+@Deprecated
 public class DefaultRegistryBroker extends AbstractRegistryBroker
 {
-    private TransientRegistry transientRegistry;
-    private List<Registry> registries = new CopyOnWriteArrayList<Registry>();
+    private final List<Registry> registries = new CopyOnWriteArrayList<>();
+    private final AtomicReference<InitialisingRegistry> initialisingRegistry = new AtomicReference<>(null);
```
dfeist: Beffer field name "single" or "defaut" registry, this is what it is isn't it?

core/src/main/java/org/mule/registry/MuleRegistryHelper.java
```diff
@@ -146,35 +149,13 @@ public Connector lookupConnector(String name)
...
-     * should be avoided and the intelligent, role specific endpoint lookup methods
-     * should be used instead.<br/><br/>
-     *
-     * @param name the idendtifer/name used to register endpoint in registry
-     * @return foo
+     * {@inheritDoc}
      */
-    /*public ImmutableEndpoint lookupEndpoint(String name)
+    @Override
+    public ObjectLimbo getLimbo()
```
dfeist: See other comment about using a simple registry rather than limbo.

core/src/main/java/org/mule/registry/DefaultRegistryBroker.java
```diff
@@ -7,41 +7,63 @@
...
+            initialisingRegistry.compareAndSet((InitialisingRegistry) registry, null);
+        }
     }
 
     protected Collection<Registry> getRegistries()
     {
         return registries;
     }
+
+    protected InitialisingRegistry getInitialisingRegistry()
```
dfeist: Not convinced about this name, for future state.

core/src/main/java/org/mule/registry/TransientRegistry.java
```diff
@@ -66,7 +67,7 @@ public TransientRegistry(String id, MuleContext muleContext)
 
     private void putDefaultEntriesIntoRegistry()
     {
-        Map<String, Object> processors = new HashMap<String, Object>();
+        Map<String, Object> processors = new HashMap<>();
```
dfeist: Why remove generics?

modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/GrizzlyServerManager.java
```diff
@@ -186,7 +189,7 @@ public Server createSslServerFor(TlsContextFactory tlsContextFactory, WorkManage
         sslFilterDelegate.addFilterForAddress(serverAddress, createSslFilter(tlsContextFactory));
         httpServerFilterDelegate.addFilterForAddress(serverAddress, createHttpServerFilter(usePersistentConnections, connectionIdleTimeout));
         executorProvider.addExecutor(serverAddress, workManagerSource);
-        final GrizzlyServer grizzlyServer = new GrizzlyServer(serverAddress, transport, httpListenerRegistry);
+        final GrizzlyServer grizzlyServer = wrap(new GrizzlyServer(serverAddress, transport, httpListenerRegistry), serverAddress);
```
dfeist: Review these changes with @pablolagreca to see is there is a simpler alternative approach with some changes to http connector.

modules/scripting/src/main/java/org/mule/module/scripting/builders/ScriptConfigurationBuilder.java
```diff
@@ -66,6 +67,9 @@ public ScriptConfigurationBuilder(String scriptEngineName, String[] configResour
     protected void doConfigure(MuleContext muleContext) throws Exception
     {
         this.muleContext = muleContext;
+        SimpleRegistry registry = new SimpleRegistry(muleContext);
```
dfeist: This related to my previous comments.  IMO, when spring-config isn't on the classpath muleContext should still be ready to use without having to manually create, initialise and add a SimpleRegisty.

modules/spring-config/src/main/java/org/mule/config/spring/SpringConfigurationBuilder.java
```diff
@@ -58,7 +59,10 @@ protected void doConfigure(MuleContext muleContext) throws Exception
         // Note: The SpringRegistry must be created before applicationContext.refresh() gets called because
         // some beans may try to look up other beans via the Registry during preInstantiateSingletons().
         muleContext.addRegistry(registry);
-        registry.initialise();
+        if (muleContext.getLifecycleManager().isPhaseComplete(Initialisable.PHASE_NAME))
```
dfeist: Why is the configuration builder doing lifecycle?  This should be in addRegisty() IMO.  Also once addRegisty() is gone, wondering what happending is you use multiple SpringConfigurationBuilders? How does this work, because each ConfigurationBuilder is creating and adding a new registry?  If it doesn't make sense to use mutiple ConfigurationBuilders any more, we need to give this some thought and work out if there are any API changes that can be made to reflect this.  The original ConfigruationBuilder interface was put together like this to specifically support/allow multiple configuration sources.

modules/spring-config/src/main/java/org/mule/config/spring/SpringRegistryBootstrap.java
```diff
@@ -0,0 +1,489 @@
...
+ * Note that the key used for transformers must be 'transformer.x' where 'x' is a sequential number.  The transformer name will be
+ * automatically generated as JMSMessageToXXX where XXX is the return class name i.e. JMSMessageToString unless a 'name'
+ * parameter is specified. If no 'returnClass' is specified the default in the transformer will be used.
+ * <p/>
+ * Note that all objects defined have to have a default constructor. They can implement injection interfaces such as
+ * {@link MuleContextAware} and lifecycle interfaces such as {@link Initialisable}.
+ *
+ * @since 3.7.0
+ */
+public class SpringRegistryBootstrap implements MuleContextAware
```
dfeist: Worth adding RegistyBootstrap interface in mule-core registry package?  I would.

modules/spring-config/src/main/java/org/mule/config/spring/SpringRegistryLifecycleManager.java
```diff
@@ -7,20 +7,41 @@
...
 import org.mule.lifecycle.phases.MuleContextStopPhase;
 import org.mule.lifecycle.phases.NotInLifecyclePhase;
+import org.mule.processor.AbstractMessageProcessorOwner;
 import org.mule.registry.AbstractRegistryBroker;
+import org.mule.routing.requestreply.AbstractAsyncRequestReplyRequester;
+
+import java.util.LinkedHashSet;
+import java.util.Set;
 
 public class SpringRegistryLifecycleManager extends RegistryLifecycleManager
```
dfeist: I'm unsure, with the rest of the changes, why a spring specific lifecycle manager is needed at all now?

modules/spring-config/src/main/java/org/mule/config/spring/SpringXmlConfigurationBuilder.java
```diff
@@ -132,6 +128,7 @@ else if (domainContext != null)
         // some beans may try to look up other beans via the Registry during
         // preInstantiateSingletons().
         muleContext.addRegistry(registry);
```
dfeist: Same comment as before.  We're deprecating the multiple registries, but this configurtion builder uses it.

core/src/main/java/org/mule/api/registry/InitialisingRegistry.java
marianogonzalez: How about LifecycleRegistry?

core/src/main/java/org/mule/api/registry/ObjectLimbo.java
marianogonzalez: That's the first thing we tried and wasn't good. You run into situations in which you need that registry in points where it should no longer be available, you start needing to add listeners to remove it... it's a mess

core/src/main/java/org/mule/registry/TransientRegistry.java
marianogonzalez: it's a java 7 diamond

modules/spring-config/src/main/java/org/mule/config/spring/SpringConfigurationBuilder.java
marianogonzalez: will add a warning on the registry broker if more than one registry is added

PR #1465, opened 2015-03-04
MULE-8297: Updated assembly-whitelist.txt

marcosnc:


marianogonzalez:
I think there're version upgrades here that are not listed in the migration guide

PR #1470, opened 2015-03-06
MULE-8362: Update spring-aop to version 3.2.12 or greater

pablokraan:


marianogonzalez:
let's discuss on post scrum if this should be included in the migration guide

PR #1473, opened 2015-03-09
MULE-8367: Adding relative path inbound property to reflect the request ...

afelisatti:
...path without the base path part.


pablolagreca:
Please add a migration guide entry with the new property and create a JIRA for documentation to document the new property.

PR #1475, opened 2015-03-09
MULE-8368: Modify substitutableInt restriction to allow APIGateway exten...

marianogonzalez:
...sion language

### DIFF
--- a/modules/spring-config/src/main/resources/META-INF/mule.xsd
+++ b/modules/spring-config/src/main/resources/META-INF/mule.xsd
@@ -7969,7 +7969,7 @@
             </xsd:documentation>
         </xsd:annotation>
         <xsd:restriction base="xsd:string">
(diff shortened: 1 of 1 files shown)


modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -7969,7 +7969,7 @@
             </xsd:documentation>
         </xsd:annotation>
         <xsd:restriction base="xsd:string">
-            <xsd:pattern value="(\#\[.+\]|\$\{[^\}]+\})"/>
+            <xsd:pattern value="(\#\[.+\]|\$\{[^\}]+\}|\!\[.+\])"/>
```
marcosnc: Could you use ((\#|\!)\[.+\]|\$\{[^\}]+\}) instead

PR #1477, opened 2015-03-09
MULE-8370: ObjectAlreadyExistsException when using splitter and until-su...

marianogonzalez:
...ccessful

### DIFF
--- a/core/src/main/java/org/mule/routing/AsynchronousUntilSuccessfulProcessingStrategy.java
+++ b/core/src/main/java/org/mule/routing/AsynchronousUntilSuccessfulProcessingStrategy.java
@@ -31,6 +31,7 @@ import org.mule.util.store.QueuePersistenceObjectStore;
 
 import java.io.NotSerializableException;
 import java.io.Serializable;
+import java.util.Random;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
@@ -52,6 +53,7 @@ import org.apache.commons.logging.LogFactory;
 public class AsynchronousUntilSuccessfulProcessingStrategy extends AbstractUntilSuccessfulProcessingStrategy implements Initialisable, Startable, Stoppable, MessagingExceptionHandlerAware
 {
 
+    private static final Random random = new Random();
     protected transient Log logger = LogFactory.getLog(getClass());
     private MessagingExceptionHandler messagingExceptionHandler;
     private ScheduledExecutorService scheduledPool;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mule/routing/AsynchronousUntilSuccessfulProcessingStrategy.java
```diff
@@ -215,8 +217,11 @@ public static Serializable buildQueueKey(final MuleEvent muleEvent)
         // the key is built in way to prevent UntilSuccessful workers across a
         // cluster to compete for the same
         // events over a shared object store
-        String key = muleEvent.getFlowConstruct() + "-" + muleEvent.getMuleContext().getClusterId() + "-"
-                     + muleEvent.getId();
+        // it also adds a random trailer to support events which have been
+        // splitted and thus have the same id. Random number was chosen over
+        // UUID for performance reasons
+        String key = String.format("%s-%s-%s-%d", muleEvent.getFlowConstruct(), muleEvent.getMuleContext().getClusterId(), muleEvent.getId(), random.nextInt());
```
pablokraan: I'm not sure if the key can be random. @pablolagreca can you confirm if this is OK?

marianogonzalez: just to clarify.. the key is not random... I added a random trailer to the key we already have just to support the case of splitted events with the same id

tests/integration/src/test/java/org/mule/test/routing/UntilSuccessfulWithSplitterTestCase.java
```diff
@@ -0,0 +1,73 @@
...
+    public static class FailAtFirstAttempt implements MessageProcessor
+    {
+
+        @Override
+        public MuleEvent process(MuleEvent event) throws MuleException
+        {
+            final String payload = event.getMessageAsString();
+            seenPayloads.add(payload);
+
+            if (seenPayloads.count(payload) == 1)
```
pablokraan: This should use a counter instead of checking the payload, otherwise the class name is incorrect

marianogonzalez: the idea is to fail the first time, once per message. Counter doesn't do that trick

core/src/main/java/org/mule/routing/AsynchronousUntilSuccessfulProcessingStrategy.java
pablolagreca: No problem with making the key random. Despite it exceeds the purpose of this PR, looking at the reason to build the message key that way, I see an existing issue when a server is shutdown gracefully or crashes.

Currently each node schedules for processing the events that it receives. So every node in a cluster would process the message in which the until-successful was executed. 
If a new node is added to the cluster then it would execute the start phase of the until-successful component which would schedule EVERY message in the object store to be process in that new node which may lead to race conditions between the new node and the node that store the event the first time.

Also it doesn't seem to manage well failover. When a node goes done all the pending tasks of the until-successful will not be executed until the node gets restarted.

marianogonzalez: Created MULE-8378 and MULE-8379 to track the concerns expressed by @pablolagreca.

@pablokraan  can we proceed with merging?

PR #1499, opened 2015-03-13
MULE-8384: Removing new properties map creation in session merge, updati...

afelisatti:
...ng existing map with new properties instead.


dfeist:
Have you fun all CE tests locally? Be good to do that before merge, givin it's hard to know if this will have any side-effects.

afelisatti:
This branch is tracked in Cloudbees (Mule-3.x-CE-MULE-8384). Two tests have failed but it seems unrelated (I have seen the same tests fail locally numerous times before).

PR #1501, opened 2015-03-13
MULE-7588: LifeCycle is inconsistent across registries

marianogonzalez:
* Fixing issue with optional registry bootstrap objects
* FIxing issue with wrong attempt to rename an agent

### DIFF
--- a/core/src/main/java/org/mule/config/bootstrap/AbstractRegistryBootstrap.java
+++ b/core/src/main/java/org/mule/config/bootstrap/AbstractRegistryBootstrap.java
@@ -79,9 +79,9 @@ public abstract class AbstractRegistryBootstrap implements Initialisable, MuleCo
 
     protected final transient Logger logger = LoggerFactory.getLogger(getClass());
 
-    public String TRANSFORMER_KEY = ".transformer.";
-    public String OBJECT_KEY = ".object.";
-    public String SINGLE_TX = ".singletx.";
+    public static final String TRANSFORMER_KEY = ".transformer.";
+    public static final String OBJECT_KEY = ".object.";
+    public static final String SINGLE_TX = ".singletx.";
 
     protected ArtifactType supportedArtifactType = ArtifactType.APP;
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/mule/config/bootstrap/AbstractRegistryBootstrap.java
```diff
@@ -253,6 +253,10 @@ private void registerObject(String key, String value) throws Exception
         {
             throwExceptionIfNotOptional(optional, cnfe, "Ignoring optional object: " + className);
         }
+        catch (NoSuchMethodException nsme)
```
pablokraan: This exception was never thrown before. What is the root cause of this problem?

marianogonzalez: just somehting that i found during testing... That exception could also happen with abstract classes or classes without a default constructor... not directly related but worth accounting for

modules/spring-config/src/main/java/org/mule/config/spring/OptionalObjectsController.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Set;
+
+/**
+ * Keeps track of keys that have been marked as optional
+ * on a registry-bootstrap file. It also tracks the optional keys
+ * that couldn't in fact be instantiated and are discarded
+ *
+ * @since 3.7.0
+ */
+public class OptionalObjectsController
```
pablokraan: Dont understand the idea. Optional objects are still in the registry? If that is true another bean could reference an optional bean that is going to be remove before intialization, which would be an error

marianogonzalez: if it's optional then you shouldn't depend on it directly since that would fail pseudo-randomly. That behavior exists already in all previous versions. If you read further, there's a been processor that will remove the bean definition from the registry as soon as somebody requests it

PR #1504, opened 2015-03-16
MULE-8392: Update developers list in Mule pom

pablokraan:


asequeira:
Alejandro Nosenzo is not in the list.

PR #1506, opened 2015-03-19
MULE-8411: XmlToXMLStreamReader does not support OutputHandler as source type

andresgregoire:


modules/xml/src/main/java/org/mule/module/xml/util/XMLUtils.java
```diff
@@ -333,6 +333,19 @@ else if (obj instanceof byte[])
             // TODO Handle encoding/charset?
             return factory.createXMLStreamReader(new ByteArrayInputStream((byte[]) obj));
         }
+        else if (obj instanceof OutputHandler)
+        {
+            ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+            try
+            {
+                ((OutputHandler)obj).write(RequestContext.getEvent(), outputStream);
```
marianogonzalez: RequestContext is deprecated. Is there no other way of getting the event?

modules/xml/src/main/java/org/mule/module/xml/util/XMLUtils.java
```diff
@@ -333,6 +333,19 @@ else if (obj instanceof byte[])
             // TODO Handle encoding/charset?
             return factory.createXMLStreamReader(new ByteArrayInputStream((byte[]) obj));
         }
+        else if (obj instanceof OutputHandler)
```
marianogonzalez: this solution looks tied to the implementation of the current XML transformers which use this util class. Can be make it so that this is resolved by the generic transformer resolution algorithm?

modules/xml/src/test/java/org/mule/transformers/xml/XmlToXMLStreamReaderTestCase.java
```diff
@@ -0,0 +1,104 @@
...
+
+import javax.xml.stream.XMLInputFactory;
+import javax.xml.stream.XMLStreamException;
+import javax.xml.stream.XMLStreamReader;
+
+import org.custommonkey.xmlunit.XMLUnit;
+import org.junit.Test;
+import org.w3c.dom.Document;
+
+public class XmlToXMLStreamReaderTestCase extends AbstractTransformerTestCase
```
marianogonzalez: The name of this test doesn't represent what you're doing. This test can only be applied to test the output handler case. Reflect that on the name

modules/xml/src/main/java/org/mule/module/xml/util/XMLUtils.java
```diff
@@ -333,6 +333,19 @@ else if (obj instanceof byte[])
...
+        {
+            ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
```
marianogonzalez: Use a PipedOutputStream instead

PR #1509, opened 2015-03-19
MULE-8417: Domain deployment fails on path with spaces

pablokraan:


modules/launcher/src/main/java/org/mule/module/launcher/domain/DefaultMuleDomain.java
```diff
@@ -66,7 +68,14 @@ public DefaultMuleDomain(DomainClassLoaderRepository domainClassLoaderRepository
         URL resource = deploymentClassLoader.findLocalResource(this.DOMAIN_CONFIG_FILE_LOCATION);
         if (resource != null)
         {
-            this.configResourceFile = new File(resource.getFile());
+            try
+            {
+                this.configResourceFile = new File(URLDecoder.decode(resource.getFile(), "UTF-8"));
```
marianogonzalez: if you do this here , shouldn't the same be done for applciations?

pablokraan: Applications do not find resources in this way

PR #1513, opened 2015-03-20

marcosnc:


modules/xml/src/test/java/org/mule/transformers/xml/xslt/XsltResultDocumentTestCase.java
```diff
@@ -0,0 +1,71 @@
...
+        File outputFile = temporaryFolder.newFile(UUID.getUUID());
+
+        try
+        {
+            executeFlowAndValidateOutput(cities, outputFile);
+            executeFlowAndValidateOutput(cities, outputFile);
+        }
+        finally
+        {
+            outputFile.delete();
```
pablokraan: Do you need to delete this file that is inside the temporary folder?

PR #1514, opened 2015-03-20

marcosnc:


modules/xml/src/main/java/org/mule/module/xml/transformer/XsltTransformer.java
```diff
@@ -245,10 +245,12 @@ protected void doTransform(MuleMessage message, String outputEncoding, Source so
         {
             if (transformer != null)
             {
-                // clear transformation parameters before returning transformer to the
-                // pool
+                // clear transformation parameters before returning transformer to the pool
                 transformer.clearParameters();
 
+                // Clean up transformer before return it to the pool
+                transformer.reset();
```
marianogonzalez: this should be done on the passivateObject() method of PooledXsltTransformerFactory

PR #1515, opened 2015-03-20

marcosnc:


modules/xml/src/test/java/org/mule/transformers/xml/xslt/XsltResultDocumentTestCase.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Test
+    public void writeOutputSeveralTimes() throws Exception
+    {
+        String cities = IOUtils.getResourceAsString(INPUT_FILE, getClass());
+
+        File outputFile = temporaryFolder.newFile(UUID.getUUID());
+
+        executeFlowAndValidateOutput(cities, outputFile);
+        executeFlowAndValidateOutput(cities, outputFile);
```
marianogonzalez: delete the output file before writing to it for the second time. Otherwise, you're not sure if the second operation succeded or failed silently since you could be reading the data generated by the first execution

PR #1519, opened 2015-03-25
MULE-8430: Domain creates .mule folder on CWD instead of MULE_HOME

pablokraan:


modules/launcher/src/test/java/org/mule/module/launcher/domain/DomainMuleContextBuilderTestCase.java
```diff
@@ -0,0 +1,31 @@
...
+
+@SmallTest
+public class DomainMuleContextBuilderTestCase extends AbstractMuleTestCase
+{
+
+    @Test
+    public void createsContainerConfiguration() throws Exception
+    {
+        DomainMuleContextBuilder builder = new DomainMuleContextBuilder("test");
+
```
marianogonzalez: remove empty spaces

modules/launcher/src/main/java/org/mule/module/launcher/domain/DomainMuleContextBuilder.java
```diff
@@ -41,7 +41,7 @@ public DomainMuleContextBuilder(String domainId)
     @Override
     protected MuleConfiguration getMuleConfiguration()
     {
-        DefaultMuleConfiguration defaultMuleConfiguration = new DefaultMuleConfiguration();
+        DefaultMuleConfiguration defaultMuleConfiguration = new DefaultMuleConfiguration(true);
```
marianogonzalez: if I migrate to a version containing this fix, does it mean that a new .mule will be created on MULE_HOME and whatever data stored in CWD will no longer be accessible? If so, shouldn't we include a migration guide entry saying that and instructing how to move the data over?

modules/launcher/src/test/java/org/mule/module/launcher/domain/DomainMuleContextBuilderTestCase.java
pablokraan: what empty spaces?

marianogonzalez: meant empty lines between each test line

pablokraan: That empty lines are intended to separate the three conceptual parts of a test: setup, execute, assert

modules/launcher/src/main/java/org/mule/module/launcher/domain/DomainMuleContextBuilder.java
pablokraan: Is a good point that this should be documented

PR #1524, opened 2015-03-28
MULE-8436 - TestCase: Dates should be independent of locale settings

basmussen:
TestCase breaks down if the default locale is not en_US. Issue affects all mule branches. JIRA issue MULE-8436.


marianogonzalez:
Thank you for this contribution Ben!

PR #1525, opened 2015-03-30
MULE-8410: Leverage MULE-7588 in extensions api

marianogonzalez:


core/src/main/java/org/mule/DefaultMuleContext.java
```diff
@@ -1046,5 +1046,9 @@ public void setLocalMuleClient(DefaultLocalMuleClient localMuleContext)
     public void setExtensionManager(ExtensionManager extensionManager)
     {
         this.extensionManager = extensionManager;
+        if (extensionManager instanceof MuleContextAware)
```
pablokraan: This is wrong to me, a dependency should not be updated just because is being used by another object. Somebody else should create the relation between these objects

PR #1528, opened 2015-03-31
MULE-8441: Add a way to inject all available core extensions in a core e...

pablokraan:
...xtension


modules/launcher/src/main/java/org/mule/module/launcher/coreextension/IllegalDependencyException.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+package org.mule.module.launcher.coreextension;
+
+import org.mule.MuleCoreExtension;
+
+/**
+ * Thrown to indicate that a mule core extension has an illegal dependency
+ * against another {@link MuleCoreExtension}.
+ */public class IllegalDependencyException extends RuntimeException
```
marianogonzalez: this exception is generic and can potentially be used to other uses cases, which is good IMO. Why not moving this to core and removing the reference to core extensions from the javadoc?

PR #1530, opened 2015-03-31
MULE-8442: app.home property is no longer available

marianogonzalez:


modules/launcher/src/test/java/org/mule/module/launcher/DeploymentServiceTestCase.java
```diff
@@ -176,6 +177,25 @@ public void deploysAppZipOnStartup() throws Exception
     }
 
     @Test
+    public void appHomePropertyIsPresent() throws Exception
+    {
+        addExplodedAppFromResource(dummyAppDescriptor.zipPath);
+        doRedeployAppByChangingConfigFile(dummyAppDescriptor.path, PROPERTIES_APP_CONFIG_XML);
```
pablokraan: Why redeploy if deployment service was not started yet?

marianogonzalez: to avoid unnecesary creation of an appDescriptor... I just use the dummy one and change the mule-config.xml file for the one which does what I need

PR #1544, opened 2015-04-07
MULE-8452: Introduce type level aliases for extensions API

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/module/extension/internal/util/NameUtils.java
```diff
@@ -217,7 +218,11 @@ public static boolean isUncountable(String word)
 
     public static String getGlobalPojoTypeName(DataType type)
     {
-        return hyphenize(type.getRawType().getSimpleName());
+        Alias alias = type.getRawType().getAnnotation(Alias.class);
+        String name = alias != null ? alias.value() : type.getRawType().getSimpleName();
+        return hyphenize(name);
+
+        //return hyphenize(type.getRawType().getSimpleName());
```
pablokraan: remove commented code

PR #1546, opened 2015-04-08
MULE-8421: Improve MuleMessage's data type consistency

pablokraan:


core/src/main/java/org/mule/DefaultMuleMessage.java
```diff
@@ -224,6 +246,18 @@ public DefaultMuleMessage(Object message, MuleMessage previous, MuleContext mule
         resetAccessControl();
     }
 
+    private MuleMessageDataTypeWrapper<?> createDataTypeWrapper(DataType<?> dataType)
+    {
+        if (dataType instanceof MuleMessageDataTypeWrapper)
+        {
+            return new MuleMessageDataTypeWrapper<>(this, ((MuleMessageDataTypeWrapper) dataType).getDelegate());
```
dfeist: Does this happen somewhere, or it's just incase?

I assume you can't just return 'dataType' because it may have a diferent message instance?

core/src/main/java/org/mule/DefaultMuleMessage.java
```diff
@@ -1336,6 +1368,24 @@ public Object getPayload()
     @Override
     public synchronized void setPayload(Object payload)
     {
+        DataType  newDataType;
```
dfeist: Do we need any special handling for NullPayload.class here?

core/src/main/java/org/mule/MuleMessageDataTypeWrapper.java
```diff
@@ -0,0 +1,101 @@
...
+    @Override
+    public void setEncoding(String encoding)
+    {
+        dataType.setEncoding(encoding);
+
+        if (encoding != null)
+        {
+            muleMessage.setOutboundProperty(MuleProperties.MULE_ENCODING_PROPERTY, encoding);
+        }
+    }
```
dfeist: Doesn't the mimeType outbound property also need updating if the encoding is updated?

core/src/main/java/org/mule/DefaultMuleMessage.java
```diff
@@ -196,12 +206,24 @@ public DefaultMuleMessage(Object message, MuleMessage previous, MuleContext mule
...
             MuleMessage payloadMessage = (MuleMessage) message;
+            dataType = createDataTypeWrapper(payloadMessage.getDataType());
             setPayload(payloadMessage.getPayload());
             copyMessageProperties(payloadMessage);
         }
         else
         {
-            setPayload(message);
+            DataType newDataType;
+            if (isCloningMessage)
```
dfeist: Wondering if there is another way to do this to avoid flag and if/else.  Could this constrcutor recieve the dataType maybe?

core/src/main/java/org/mule/DefaultMuleMessage.java
pablokraan: This happens every time a message is created from another message

core/src/main/java/org/mule/DefaultMuleMessage.java
```diff
@@ -485,6 +524,15 @@ public void setProperty(String key, Object value, PropertyScope scope)
         }
     }
 
+    private void updateProperties(String key, Object value)
```
dfeist: This method name is unclear I think.  It's not updating properties, it's updating the dataType, and it's only doing it for a specific property key.

core/src/main/java/org/mule/DefaultMuleMessage.java
pablokraan: I'll add it

core/src/main/java/org/mule/MuleMessageDataTypeWrapper.java
pablokraan: Yes. I already found that problem. I was planning to include that as a separated issue as that problem already exist and is a bit more complex that just updating that property. (need to update mimeType/encoding when content-type property is set on the message)

core/src/main/java/org/mule/MuleMessageDataTypeWrapper.java
```diff
@@ -0,0 +1,101 @@
...
+import org.mule.api.transformer.DataType;
+import org.mule.transformer.types.MimeTypes;
+
+/**
+ * Wraps the @{link DataType} associated to a {@link MuleMessage} in order to
+ * update message properties when data type is modified and vice versa.
+ *
+ * @since 3.7.0
+ */
+public class MuleMessageDataTypeWrapper<T> implements DataType<T>
```
dfeist: Can't visability be reduced, or even private in DefaultMuleMessage? Or, does it need to be public for any reason?

core/src/main/java/org/mule/DefaultMuleMessage.java
pablokraan: I'll check if that can be done

core/src/main/java/org/mule/DefaultMuleMessage.java
pablokraan: The idea is to update content-type property too

core/src/main/java/org/mule/MuleMessageDataTypeWrapper.java
pablokraan: It can be reduced

PR #1548, opened 2015-04-08
MULE-8453: Add data type to message properties

pablokraan:


core/src/main/java/org/mule/transformer/types/TypedValue.java
```diff
@@ -0,0 +1,39 @@
...
+        this.value = value;
+        this.dataType = dataType;
+    }
+
+    public DataType<?> getDataType()
+    {
+        return dataType;
+    }
+
+    public Object getValue()
```
dfeist: Is it not possible to use generics here to return the type?

PR #1549, opened 2015-04-08
MULE-8455 Upgrade license checker

vromero:


pom.xml
```diff
@@ -1697,10 +1696,14 @@
                 <plugin>
                     <groupId>com.mycila</groupId>
                     <artifactId>license-maven-plugin</artifactId>
-                    <version>2.6</version>
+                    <version>2.8</version>
                     <configuration>
                         <skip>${skipVerifications}</skip>
-                        <header>${licensePath}</header>
+                        <header>com/mycila/maven/plugin/license/templates/CPAL.txt</header>
```
marianogonzalez: are you willing to bet the little hair you have left that this generates the exact same header as we have now?

marianogonzalez: how about EE? Do you have to configure the plugin again? you will need a file there because that license is custom

vromero: I can bet it is the same file as I personally did the pull request to the mycila project. And of course I tested locally before commiting.
For EE it has to remain as it is, the license is not included (obviously) in the license checker distribution.

marianogonzalez: what do you mean that EE has to remain as it is? EE inherits the plugin definition from CE, so at the very least you have to override it

vromero: you are right about EE. I've created a parallel PR https://github.com/mulesoft/mule-ee/pull/733

PR #1552, opened 2015-04-09
MULE-8463: Remove code duplication of classes implementing AnnotatedObje...

pablokraan:
...ct interface


core/src/main/java/org/mule/AbstractAnnotatedObject.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Collections;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import javax.xml.namespace.QName;
+
+/**
+ * Base implementation for {@link AnnotatedObject}
+ */
+public class AbstractAnnotatedObject implements AnnotatedObject
```
marianogonzalez: abstract modified missing?

PR #1560, opened 2015-04-09
MULE-8216: Implement a validations extensions

marianogonzalez:


core/src/main/java/org/mule/util/ArrayUtils.java
```diff
@@ -138,5 +141,57 @@ else if (componentType.equals(Double.TYPE))
...
+    public static int getLength(Object array)
+    {
+        if (array == null)
+        {
+            return 0;
+        }
+
+        checkArgument(array.getClass().isArray(), String.format("Object of type %s is not an array", array.getClass().getName()));
+
+        Class<?> componentType = array.getClass().getComponentType();
```
vromero: I have no idea if it's necessary but just my 2 cents, lists and maps could be measured too: http://grepcode.com/file/repo1.maven.org/maven2/com.rabbitmq/amqp-client/2.3.0/com/rabbitmq/client/impl/Frame.java#284

extensions/pom.xml
```diff
@@ -19,7 +19,7 @@
     <description>Mule extensions that augment the core Mule functionality through the Mule Extension API</description>
 
     <modules>
-        <!-- <module>validation</module> -->
+         <module>validation</module>
```
vromero: Should we start with the right first step by placing this extensions on a different git repo?

marianogonzalez: I don't see why that's a right first step. This API is not only meant for external extensions or connectors like devkit was. This is meant as the very foundation of future mule.

Thus, this is unrelated to the decision of keeping all components which are part of the distro together vs splitting each component to its own.

core/src/main/java/org/mule/util/ArrayUtils.java
marianogonzalez: I don't understand how that method works, thus I don't want to repackage it Mule. I don't want to depend on that library either. Thanks for the tip though, I thought it was not possible to do that in Java

extensions/validation/src/main/java/org/mule/extension/validation/internal/CommonValidationOperations.java
```diff
@@ -0,0 +1,519 @@
...
+
+/**
+ * Defines the operations of {@link ValidationExtension}
+ * which executes the {@link Validator}s that the extension
+ * provides out of the box
+ *
+ * @see ValidationExtension
+ * @since 3.7.0
+ */
+@ImplementationOf(ValidationExtension.class)
```
marcosnc: Why do you need this two way reference? 
Here, pointing to ValidationExtension.class, and in ValidationExtension using @Operations({CommonValidationOperations.class, CustomValidatorOperation.class})

extensions/validation/src/main/java/org/mule/extension/validation/internal/CommonValidationOperations.java
```diff
@@ -0,0 +1,519 @@
...
+     * substitution group.
+     *
+     * @param validations     the nested validation operations
+     * @param throwsException if {@code true}, then a {@link ValidationException} is thrown if any validator fails.
+     * @param muleEvent       the current {@link MuleEvent}
+     * @return the same {@code muleEvent} that was passed as argument
+     * @throws ValidationException if at least one validator fails and {@code throwsException} is {@code true}
+     */
+    @Operation
+    public MuleEvent all(@RestrictedTo(ValidationExtension.class) List<NestedProcessor> validations,
```
pablolagreca: This method should be moved to a different class. It's not a type of validation but a validation execution strategy.

extensions/validation/src/main/java/org/mule/extension/validation/internal/CommonValidationOperations.java
marianogonzalez: We discussed this personally. This is not a two way reference but an improvement will be made in the ext-api so that operations of an extensible extension do not need to explicitelly set this. That will be however a different issue

PR #1564, opened 2015-04-12
MULE-8457: Replace SetPayloadTransformer with a plain message processor

pablokraan:


tests/integration/src/test/resources/set-payload-data-type-config.xml
```diff
@@ -11,7 +11,7 @@
     <flow name="setPayload">
         <vm:inbound-endpoint path="setPayloadLocal" exchange-pattern="request-response"/>
 
-        <set-payload name ="localSetPayload" value="#['test']" encoding="#['UTF-16']" mimeType="#['text/xml']" returnClass="java.lang.Long" ignoreBadInput="false" />
+        <set-payload name ="localSetPayload" value="#['test']" encoding="UTF-16" mimeType="text/xml" returnClass="java.lang.Long" ignoreBadInput="false" />
```
marianogonzalez: isn't this a functionality regression? doesn't it break backwards? If that's the case and we're fine with that, then please add a migration guide entry

pablokraan: No. This feature was added in a previous PR and after discussing it with some people I decided to removed it

PR #1566, opened 2015-04-13
MULE-8465: Improve set properties transformers to set property datatype

pablokraan:


core/src/test/java/org/mule/transformer/simple/AddVariablePropertyTransformerTestCase.java
```diff
@@ -41,13 +51,14 @@
     public static final String EXPRESSION = "#[string:someValue]";
     public static final String EXPRESSION_VALUE = "expressionValueResult";
     public static final String NULL_EXPRESSION = "#[string:someValueNull]";
-    public static final String NULL_EXPRESSION_VALUE = null;
+    public static final String CUSTOM_ENCODING = "UFT-8";
```
aiannucci: Encoding should be UTF-8

PR #1567, opened 2015-04-13
MULE-8373 Upgrade ActiveMQ to version 5.11.0

vromero:


modules/atom/pom.xml
```diff
@@ -206,7 +206,7 @@
         </dependency>
         <dependency>
             <groupId>org.apache.activemq</groupId>
-            <artifactId>activemq-core</artifactId>
+            <artifactId>activemq-broker</artifactId>
```
marianogonzalez: what's the difference between broker and core? is it possible that some user might be depending on some funcionality that is on core and now is no longer provided by the broker? if so, at the very minimum there should be a migration guide entry

vromero: It is just a change coming forced by activemq 5.8: http://activemq.apache.org/58-migration-guide.html

Now its divided on broker, broker-persistence-* and client. However -all is still containing everything. Assuming AFAIK we recommended -all so far. There is no migration guide needed. Anyhow I'll put it just in case.

marianogonzalez: I understand this is forced by activemq, but I think that a migration guide entry saying so and warning the user that he has to look into this reduces risk on our side. Also, AFAIK, using -all is not a recommendation we give (might be wrong though)

vromero: This is a test dependency. It won't be included on distros.

PR #1578, opened 2015-04-15
MULE-8469: Message transformation must no update payloads data type when...

pablokraan:
... payload was not transformed


transports/vm/src/test/resources/org/mule/test/config/content-type-setting-transform-configs-service.xml
```diff
@@ -32,7 +32,7 @@
             <outbound>
                 <pass-through-router>
                     <outbound-endpoint ref="outbound1" >
-                        <test:no-action-transformer mimeType="text/xml"/>
```
dfeist: Have you deprecated/removed this element? I think we should!

PR #1582, opened 2015-04-16
MULE-8498: Message transformation must maintain original mimeType and en...

pablokraan:
...coding hen transformer does not provide them

### DIFF
--- a/distributions/standalone/src/main/resources/MIGRATION.txt
+++ b/distributions/standalone/src/main/resources/MIGRATION.txt
@@ -71,6 +71,9 @@ MULE-8469: applying a message transformer does not changes message's data type i
            </message-properties-transformer>
            or like this:
            <set-property propertyName="Content-Type" value="text/baz;charset=UTF-16BE"/>
+MULE-8498: applying a message transformer that changes message's payload will update the message data type, but instead of using transformer's output data type it will use a merge between payload's and transformer data types.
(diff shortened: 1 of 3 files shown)


distributions/standalone/src/main/resources/MIGRATION.txt
```diff
@@ -71,6 +71,9 @@ MULE-8469: applying a message transformer does not changes message's data type i
            </message-properties-transformer>
            or like this:
            <set-property propertyName="Content-Type" value="text/baz;charset=UTF-16BE"/>
+MULE-8498: applying a message transformer that changes message's payload will update the message data type, but instead of using transformer's output data type it will use a merge between payload's and transformer data types.
+           If transformer's output data type does not provides mime type and/or encoding, then the original payload data type mime type and/or encoding will be used.
+           This can cause that different transformers are applied on an application after the upgrade. In case there is a failure use <set-payload> to set encoding and mime type while maintaining the same payload.
```
dfeist: Don't you mean "use <set-property>"?

pablokraan: No. Set-payload enables you to set payload data type using mimeType and encoding attributes and then set payload to #[payload]. You can also use set-property with propertyName="Content-Type" and value like "text/baz;charset=UTF-16BE"

PR #1599, opened 2015-04-20
MULE-8510: Setting a NullPayload in a property must behave as setting nu...

pablokraan:
...ll


core/src/main/java/org/mule/transformer/simple/AbstractAddVariablePropertyTransformer.java
```diff
@@ -48,7 +49,7 @@ public Object transformMessage(MuleMessage message, String outputEncoding) throw
         else
         {
             Object value = valueEvaluator.resolveValue(message);
-            if (value == null)
+            if (value == null || value instanceof NullPayload)
```
dfeist: Does it make sense adding a static method NullPayload.isNull() which does null check too, or something similar?  Or is this check only going to happen like this on specific occasions?

PR #1613, opened 2015-04-21
MULE-8511 Add support for non blocking processing of flows

dfeist:


dfeist:
DO NOT MERGE.  Still 1 failing test.

core/src/main/java/org/mule/DefaultMuleEvent.java
```diff
@@ -83,7 +86,7 @@
     /** Mutable MuleEvent state **/
     private boolean stopFurtherProcessing = false;
     private int timeout = TIMEOUT_NOT_SET_VALUE;
-    private transient ResponseOutputStream outputStream;
```
pablokraan: Can you remove all the references to ResponseOutputStream from this class and deprecate it?

core/src/main/java/org/mule/DefaultMuleEvent.java
```diff
@@ -272,6 +299,16 @@ public DefaultMuleEvent(MuleEvent rewriteEvent, FlowConstruct flowConstruct)
             rewriteEvent.isSynchronous());
     }
 
+    /**
+     * Copy constructor used when ReplyToHandler instance needs switching out
+     * @param rewriteEvent
+     * @param replyToHandler
+     */
+    public DefaultMuleEvent(MuleEvent rewriteEvent, ReplyToHandler replyToHandler)
```
pablokraan: The variation of constructors in this class is a mess... would be good to introduce a builder in 3.7

dfeist: For sure, but that is out of scope for this issue.  Let's consider later how important this is for 3.7 or not.

core/src/main/java/org/mule/DefaultMuleEvent.java
```diff
@@ -400,16 +438,35 @@ public DefaultMuleEvent(MuleMessage message,
 
     protected boolean resolveEventSynchronicity()
```
pablokraan: Need unit test to check all these variations

core/src/main/java/org/mule/DefaultMuleEvent.java
dfeist: I'm not sure we can/should deprecate ResponseOutputStream, I haven't analysed where it is used.  But we can use OutputStream everywhere in MuleEvent I think, I don't see any issue with that. You?

core/src/main/java/org/mule/DefaultMuleEvent.java
```diff
@@ -400,16 +438,35 @@ public DefaultMuleEvent(MuleMessage message,
...
+        return transacted
+               || isSynchronousProcessingStrategy()
+               || exchangePattern.hasResponse() && !isNonBlockingProcessingStrategy()
+               || message.getProperty(MuleProperties.MULE_FORCE_SYNC_PROPERTY,
+                                      PropertyScope.INBOUND, Boolean.FALSE);
+    }
+
+    private boolean isSynchronousProcessingStrategy()
+    {
         if (flowConstruct != null && flowConstruct instanceof ProcessingDescriptor)
```
pablokraan: This condition can be simplified to return (flowConstruct instanceof ProcessingDescriptor) && ((ProcessingDescriptor) flowConstruct).isSynchronous()

core/src/main/java/org/mule/DefaultMuleEvent.java
```diff
@@ -400,16 +438,35 @@ public DefaultMuleEvent(MuleMessage message,
...
+        {
+            return false;
+        }
+    }
+
+    private boolean isNonBlockingProcessingStrategy()
+    {
+        if (flowConstruct != null && flowConstruct instanceof Pipeline)
+        {
+            return ((Pipeline) flowConstruct).getProcessingStrategy() instanceof NonBlockingProcessingStrategy;
```
pablokraan: simplify this condition

core/src/main/java/org/mule/NonBlockingVoidMuleEvent.java
```diff
@@ -0,0 +1,29 @@
...
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule;
+
+/**
+ * A {@link NonBlockingVoidMuleEvent} represents a void return, but in contrast to {@link org.mule.VoidMuleEvent}
+ * while the return up the stack is void, a asynchronous response is expected.  Because Mule support both blocking and
```
pablokraan: a asynchronous -> an asynchronous

core/src/main/java/org/mule/NonBlockingVoidMuleEvent.java
```diff
@@ -0,0 +1,29 @@
...
+ * non-blocking the it is the return of this event instance that signals that a callback is to be expected.
```
pablokraan: Clarify "non-blocking the it is the return of this event"

core/src/main/java/org/mule/NonBlockingVoidMuleEvent.java
```diff
@@ -0,0 +1,29 @@
...
+{
+
+    private static final NonBlockingVoidMuleEvent instance = new NonBlockingVoidMuleEvent();
+
+    public static NonBlockingVoidMuleEvent getInstance()
+    {
+        return instance;
+    }
+
+    protected NonBlockingVoidMuleEvent()
```
pablokraan: This should be private

core/src/main/java/org/mule/VoidMuleEvent.java
```diff
@@ -40,7 +40,12 @@ public static VoidMuleEvent getInstance()
         return instance;
     }
 
-    private VoidMuleEvent()
+    public static boolean isVoid(MuleEvent event)
```
pablokraan: I guess you added this method to avoid repeating the "instance of VoidMuleEvent" code, but I don't think this is the right place to put it. Also, if VoidMuleEvent can't be instantiated, this can be replaced with event == VoidMuleEvent.getInstance()

core/src/main/java/org/mule/api/NonBlockingSupported.java
```diff
@@ -0,0 +1,20 @@
...
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.api;
+
+import org.mule.api.processor.MessageProcessor;
+
+/**
+ * Marker interface used to mark the {@link org.mule.api.processor.InterceptingMessageProcessor} and
```
pablokraan: "Marker interface used to mark the..." is a bit redundant.
Add a separated paragraph for the "Note:" section

core/src/main/java/org/mule/api/NonBlockingSupported.java
```diff
@@ -0,0 +1,20 @@
...
+ * {@link org.mule.api.processor.MessageRouter} that support non-blocking processing.  Unsupported implementations will
+ * cause Mule to fall back to synchronous blocking processing.  Note: All standard {@link org.mule.api.processor.MessageProcessor}
+ * implementations that aren't one of the two types mentioned above are implicitly supported.
+ */
+public interface NonBlockingSupported
```
pablokraan: not sure if there is a better name... but a name starting with a negative prefix seems counter intuitive

core/src/main/java/org/mule/api/processor/CompletionHandler.java
```diff
@@ -0,0 +1,21 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.api.processor;
+
+import org.mule.api.MessagingException;
+
+/**
+ * Interfaces used to perform callback in the case of completion of processing or failure.
+ */
+public interface CompletionHandler<T, E>
```
pablokraan: "Interfaces used to perform.. " -> "Performs a callback"
If this is used only on non blocking, then that should be added into the javadoc

core/src/main/java/org/mule/VoidMuleEvent.java
dfeist: The reason I added this wasn't to avoid == check, but rather because there are cases I need to check if the event is VoidMuleEvent *or* NonBlockingVoidMuleEvent and didn't really want to do if(event == VoidMuleEvent.getInstance() || event == NonBlockingVoidMuleEvent.getInstance()) or event instance VoidMuleEvent.  I can switch back to instanceof test if you think it is preferred though.

core/src/main/java/org/mule/api/processor/CompletionHandler.java
pablokraan: Is there a need for using generics here?

core/src/main/java/org/mule/api/processor/CompletionHandler.java
```diff
@@ -0,0 +1,21 @@
...
+{
+
+    public void onCompletion(T t);
```
pablokraan: Add javadocs

core/src/main/java/org/mule/api/source/NonBlockingMessageSource.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package org.mule.api.source;
+
+/**
+ * NonBlockingMessageSource's create instances of MuleEvent that contiain a {@link org.mule.api.transport.NonBlockingResponseReplyToHandler}
+ * allowing for a response or error to be returned asynchronously by a different thread freeing up the request thread.
+ *
+ * Implementations must support both blocking and non-blocking and therefore need to test the response {@link org.mule.api.MuleEvent}
+ * returned when invoking {@link org.mule.api.processor.MessageProcessor#process(org.mule.api.MuleEvent)} on the listener.
+ * If the event returned is an instaance of {@link org.mule.NonBlockingVoidMuleEvent} then the MessageSource should wait for the
```
pablokraan: instaance -> instance

core/src/main/java/org/mule/api/NonBlockingSupported.java
dfeist: It's a bit strange yes, but it's " 'NonBlocking' supported", not "Non 'Blocking' Supported".  I'd leave it as is, as other interfaces added in this changeset have same prefix.

core/src/main/java/org/mule/api/transport/ErrorHandlingNonBlockingResponseReplyToHandler.java
```diff
@@ -0,0 +1,51 @@
...
+import org.mule.api.processor.CompletionHandler;
+
+/**
+ * Specialized version of {@link org.mule.api.transport.NonBlockingResponseReplyToHandler} that uses a {@link org.mule.api.MessagingException}
+ * to handle exceptions.
+ */
+public class ErrorHandlingNonBlockingResponseReplyToHandler extends NonBlockingResponseReplyToHandler
+{
+
+    private MessagingExceptionHandler messagingExceptionHandler;
```
pablokraan: final

core/src/main/java/org/mule/api/transport/ErrorHandlingNonBlockingResponseReplyToHandler.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public ErrorHandlingNonBlockingResponseReplyToHandler(CompletionHandler completionHandler, MessagingExceptionHandler exceptionHandler)
+    {
+        super(completionHandler);
+        this.messagingExceptionHandler = messagingExceptionHandler;
```
pablokraan: wrong assignment

core/src/main/java/org/mule/api/processor/CompletionHandler.java
dfeist: This interface doesn't "perform the callback", it's whoever invokes it that performs callback, i'll update to make a bit clearer though.

This is a generic interface, so I disagree non-blocking should be mentioned here.

core/src/main/java/org/mule/DefaultMuleEvent.java
dfeist: Ack.

core/src/main/java/org/mule/NonBlockingVoidMuleEvent.java
dfeist: Ack.

core/src/main/java/org/mule/api/transport/ErrorHandlingNonBlockingResponseReplyToHandler.java
```diff
@@ -0,0 +1,51 @@
...
+import org.mule.api.MessagingException;
+import org.mule.api.MuleEvent;
+import org.mule.api.exception.MessagingExceptionHandler;
+import org.mule.api.processor.CompletionHandler;
+
+/**
+ * Specialized version of {@link org.mule.api.transport.NonBlockingResponseReplyToHandler} that uses a {@link org.mule.api.MessagingException}
+ * to handle exceptions.
+ */
+public class ErrorHandlingNonBlockingResponseReplyToHandler extends NonBlockingResponseReplyToHandler
```
pablokraan: Add unit test for this class

core/src/main/java/org/mule/api/transport/NonBlockingResponseReplyToHandler.java
```diff
@@ -0,0 +1,110 @@
...
+import org.mule.api.processor.CompletionHandler;
+import org.mule.processor.chain.NonBlockingProcessorIterator;
+import org.mule.processor.chain.ProcessorIterator;
+
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+
+/**
+ * {@link org.mule.api.transport.NonBlockingResponseReplyToHandler} is used to process the Flow response phase when the
```
pablokraan: "{@link org.mule.api.transport.NonBlockingResponseReplyToHandler} is used to process " -> Processes

core/src/main/java/org/mule/api/transport/NonBlockingResponseReplyToHandler.java
```diff
@@ -0,0 +1,110 @@
...
+ * have been added to be executed as part of the response phase before invoking the {@link org.mule.api.processor.CompletionHandler}
+ * provided.  In the case of an exception, response processors will not be invoked and instead {@link org.mule.api.processor.CompletionHandler#onFailure(Object)}
+ * invoked directly.
+ */
+public class NonBlockingResponseReplyToHandler implements ReplyToHandler
+{
+
+    protected final CompletionHandler completionHandler;
+
+    private MessageProcessorExecutionTemplate messageProcessorExecutionTemplate = MessageProcessorExecutionTemplate
```
pablokraan: final?

core/src/main/java/org/mule/api/transport/NonBlockingResponseReplyToHandler.java
```diff
@@ -0,0 +1,110 @@
...
+            @Override
+            public MuleEvent process(MuleEvent event) throws MuleException
+            {
+                completionHandler.onCompletion(event);
+                return null;
+            }
+        });
+    }
+
+    public synchronized void processReplyTo(MuleEvent event, MuleMessage returnMessage, Object replyTo) throws
```
pablokraan: @Override

core/src/main/java/org/mule/api/transport/NonBlockingResponseReplyToHandler.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    public void processExceptionReplyTo(MuleEvent event, MessagingException exception, Object replyTo)
+    {
+        completionHandler.onFailure(exception);
+    }
+
+    public void addResponseMessageProcessor(MessageProcessor mp)
+    {
+        responseProcessors.add(0, mp);
```
pablokraan: mp -> messageProcessor

core/src/main/java/org/mule/api/transport/NonBlockingResponseReplyToHandler.java
```diff
@@ -0,0 +1,110 @@
...
+        {
+            iterator.process();
+        }
+        catch (MessagingException e)
+        {
+            completionHandler.onFailure(e);
+        }
+    }
+
+    public void processExceptionReplyTo(MuleEvent event, MessagingException exception, Object replyTo)
```
pablokraan: Javadocs?

core/src/main/java/org/mule/api/transport/NonBlockingResponseReplyToHandler.java
dfeist: Meaning is identical :-)

core/src/main/java/org/mule/api/transport/NonBlockingResponseReplyToHandler.java
pablokraan: @Override

core/src/main/java/org/mule/construct/AbstractPipeline.java
```diff
@@ -199,7 +201,7 @@ protected void doInitialise() throws MuleException
         if (messageSource != null)
         {
             // Wrap chain to decouple lifecycle
-            messageSource.setListener(new AbstractInterceptingMessageProcessor()
+            messageSource.setListener(new MessageProcessor()
```
pablokraan: Why was this change required?

dfeist: AbstractInterceptingMessageProcessor is not needed here as far as i can see.  If we change to MessageProcessor then non blocking is simplicity supported, ese it would need to implement NonBlockingSupported.

core/src/main/java/org/mule/execution/AsyncResponseFlowProcessingPhase.java
```diff
@@ -60,12 +65,46 @@ public MuleEvent process() throws Exception
                     {
                         MuleEvent muleEvent = template.getMuleEvent();
                         fireNotification(muleEvent, MESSAGE_RECEIVED);
-                        muleEvent = template.routeEvent(muleEvent);
-                        return muleEvent;
+                        if (muleEvent.getExchangePattern().hasResponse() && !muleEvent.isSynchronous())
```
pablokraan: Can you extract this into a new method?

core/src/main/java/org/mule/api/NonBlockingSupported.java
pablokraan: What does means "supports non blocking" here? Is there any condition that must be required?

core/src/main/java/org/mule/processor/AbstractNonBlockingMessageProcessor.java
```diff
@@ -0,0 +1,75 @@
...
+ */
+public abstract class AbstractNonBlockingMessageProcessor implements NonBlockingMessageProcessor
+{
+
+    @Override
+    public MuleEvent process(MuleEvent event) throws MuleException
+    {
+        if (isNonBlocking(event))
+        {
+            processNonBlocking(event, new NonBlockingCompletionHandler(event));
```
pablokraan: Ca n you add a factory method for the completion handler?

core/src/main/java/org/mule/processor/AbstractNonBlockingMessageProcessor.java
```diff
@@ -0,0 +1,75 @@
...
+            return NonBlockingVoidMuleEvent.getInstance();
+        }
+        else
+        {
+            return processBlocking(event);
+        }
+    }
+
+    public boolean isNonBlocking(MuleEvent event)
```
pablokraan: Should not this method be on MuleEvent?

core/src/main/java/org/mule/execution/AsyncResponseFlowProcessingPhase.java
dfeist: Sure.  I could alternatively add a isNonBlocking() method to MuleEvent, because I use this else where. WDYT?

core/src/main/java/org/mule/processor/AbstractNonBlockingMessageProcessor.java
```diff
@@ -0,0 +1,75 @@
...
+        private MuleEvent event;
+        private ReplyToHandler replyToHandler;
+
+        public NonBlockingCompletionHandler(MuleEvent event)
+        {
+            this.event = event;
+            this.replyToHandler = event.getReplyToHandler();
+        }
+
+        public void onFailure(final MessagingException exception)
```
pablokraan: Add @Override

core/src/main/java/org/mule/processor/AbstractNonBlockingMessageProcessor.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    abstract protected void processNonBlocking(MuleEvent event, NonBlockingCompletionHandler completionHandler) throws MuleException;
+
+    abstract protected MuleEvent processBlocking(MuleEvent event) throws MuleException;
+
+    protected class NonBlockingCompletionHandler implements CompletionHandler<MuleEvent, MessagingException>
+    {
+
+        private MuleEvent event;
```
pablokraan: Finals

core/src/main/java/org/mule/processor/AbstractNonBlockingMessageProcessor.java
dfeist: Could be, I just made same comment on your previous comment :-)

core/src/main/java/org/mule/processor/AbstractNonBlockingMessageProcessor.java
```diff
@@ -0,0 +1,75 @@
...
+
+        public void onCompletion(MuleEvent event)
+        {
+            try
+            {
+                replyToHandler.processReplyTo(event, null, null);
+            }
+            catch (Exception e)
+            {
+                this.event.getFlowConstruct().getExceptionListener().handleException(e, event);
```
pablokraan: Potencial trainwreck :)

core/src/main/java/org/mule/processor/AbstractResponseMessageProcessor.java
```diff
@@ -6,15 +6,53 @@
...
+import org.mule.VoidMuleEvent;
 import org.mule.api.MuleEvent;
 import org.mule.api.MuleException;
+import org.mule.api.processor.MessageProcessor;
+import org.mule.api.transport.NonBlockingResponseReplyToHandler;
+import org.mule.processor.chain.DefaultMessageProcessorChain;
 
-public abstract class AbstractResponseMessageProcessor extends AbstractInterceptingMessageProcessor
+/**
+ * Asbtract implementation of a {@link org.mule.api.processor.MessageProcessor} that performs processing during the
```
pablokraan: Asbtract -> Abstract

core/src/main/java/org/mule/processor/AbstractResponseMessageProcessor.java
```diff
@@ -6,15 +6,53 @@
...
+ * response procesing phase.  Non-blocking processing is supported by implementors seperating the response processing
```
pablokraan: procesing > processing

core/src/main/java/org/mule/processor/AbstractResponseMessageProcessor.java
```diff
@@ -6,15 +6,53 @@
...
+ * into a seperate {@link org.mule.api.processor.MessageProcessor} returned by {@link #getResponseProcessor()}.
```
pablokraan: seperate > separate

core/src/main/java/org/mule/processor/AbstractResponseMessageProcessor.java
```diff
@@ -6,15 +6,53 @@
...
+ */
+public abstract class AbstractResponseMessageProcessor extends AbstractInterceptingMessageProcessor implements
+        NonBlockingSupported
 {
 
+    private DefaultMessageProcessorChain chain;
+
     public MuleEvent process(MuleEvent event) throws MuleException
```
pablokraan: Can you add missing @Override?

core/src/main/java/org/mule/processor/chain/NonBlockingProcessorIterator.java
```diff
@@ -0,0 +1,98 @@
...
+    {
+        if (event.getExchangePattern().hasResponse() && !event.isSynchronous())
+        {
+            if ((processor instanceof MessageRouter || processor instanceof InterceptingMessageProcessor) && !
+                    (processor instanceof NonBlockingSupported))
+            {
+                event = new DefaultMuleEvent(event.getMessage(), event, true);
+            }
+
+            if (processor instanceof NonBlockingMessageProcessor)
```
pablokraan: This seems to be the only place NonBlockingMessageProcessor is used besides the classes that implement it. Is the same if event.isNonBlocking is used if it exists?

core/src/main/java/org/mule/processor/chain/AbstractMessageProcessorChain.java
```diff
@@ -230,4 +232,16 @@ public void setMuleContext(MuleContext context)
             }
         }
     }
+
+    @Override
+    public boolean isNonBlocking(MuleEvent event)
+    {
+        return false;
```
pablokraan: Do this means that adding any message processor chain will convert the processing to blocking?

core/src/main/java/org/mule/processor/chain/AbstractMessageProcessorChain.java
```diff
@@ -55,7 +58,7 @@ public AbstractMessageProcessorChain(String name, List<MessageProcessor> process
         this.processors = processors;
     }
 
-    public MuleEvent process(MuleEvent event) throws MuleException
+    public MuleEvent processBlocking(MuleEvent event) throws MuleException
```
pablokraan: Missing @Override?

core/src/main/java/org/mule/processor/chain/NonBlockingProcessorIterator.java
```diff
@@ -0,0 +1,98 @@
...
+            {
+                event = new DefaultMuleEvent(event, new NonBlockingResponseReplyToHandler(new CompletionHandler<MuleEvent, MessagingException>()
```
pablokraan: Can be this a inner or top level class in order to make it unit testable?

core/src/main/java/org/mule/processor/chain/NonBlockingProcessorIterator.java
```diff
@@ -0,0 +1,98 @@
...
+                    public void onFailure(MessagingException exception)
+                    {
+                        replyToHandler.processExceptionReplyTo(event, exception, null);
+                    }
+                }));
+            }
+        }
+    }
+
+    private void resume() throws MessagingException
```
pablokraan: Seems like this method can be moved into the NonBlockingResponseReplyToHandler anonymous class

core/src/main/java/org/mule/processor/chain/ProcessorIterator.java
```diff
@@ -0,0 +1,140 @@
...
+ * of the first processor to invoke the second and so on.  MessageProcessor implementations aside from simply iterating
+ * over processors implement rules regarding if and when iteration should stop early or even stop temporarily and be
+ * continued later.
+ * <p/>
+ * This implementation processes each {@link org.mule.api.processor.MessageProcessor} in sucession in the same thread
+ * until or processors have been invoked or one of the following is returned by a processor:
+ * <li>{@link org.mule.VoidMuleEvent}</li>
+ * <li><code>null</code></li>
+ */
+public class ProcessorIterator
```
pablokraan: Name is misleading to me, I would expect a ProcessIterator to implement Iterator, and that is not the case

core/src/main/java/org/mule/processor/chain/ProcessorIterator.java
```diff
@@ -0,0 +1,140 @@
...
+{
+
+    final protected MessageProcessorExecutionTemplate messageProcessorExecutionTemplate;
```
pablokraan: protected final

core/src/main/java/org/mule/processor/chain/ProcessorIterator.java
```diff
@@ -0,0 +1,140 @@
...
+    public ProcessorIterator(MuleEvent event, List<MessageProcessor> processors,
+                             MessageProcessorExecutionTemplate messageProcessorExecutionTemplate, boolean copyOnVoidEvent)
+    {
+        this.event = event;
+        this.processors = processors;
+        this.copyOnVoidEvent = copyOnVoidEvent;
+        this.messageProcessorExecutionTemplate = messageProcessorExecutionTemplate;
+    }
+
+    public MuleEvent process() throws MessagingException
```
pablokraan: Should not be a MessageProcessor?

core/src/main/java/org/mule/processor/AbstractNonBlockingMessageProcessor.java
dfeist: Will fix making this MessagingExceptionHandlerAware, this is also required so this will behave correctly inside an exception strategy and not go into a loop..

core/src/main/java/org/mule/processor/chain/ProcessorIterator.java
```diff
@@ -0,0 +1,140 @@
...
+        }
+        else
+        {
+            return messageProcessorExecutionTemplate.execute(processor, event);
+        }
+    }
+
+    protected void preProcess(MessageProcessor processor)
+    {
+        // template method.
```
pablokraan: The template method seems to be processNext, not this one

core/src/main/java/org/mule/processor/chain/ProcessorIterator.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    protected MessageProcessor nextProcessor()
+    {
+        return processors.get(index++);
+    }
+
+    private boolean processorMayReturnVoidEvent(MessageProcessor processor)
```
pablokraan: processorMayReturnVoidEvent > mayProcessorReturnVoidEvent?

core/src/main/java/org/mule/processor/chain/ProcessorIterator.java
```diff
@@ -0,0 +1,140 @@
...
+        {
+            return false;
+        }
+        else
+        {
+            return true;
+        }
+    }
+
+    public static ProcessorIterator getProcessorIterator(MuleEvent event,
```
pablokraan: getProcessorIterator > createProcessorIterator?

core/src/main/java/org/mule/processor/chain/ProcessorIterator.java
pablokraan: Add unit tests for this class

core/src/main/java/org/mule/processor/chain/ServiceProcessorIterator.java
```diff
@@ -0,0 +1,44 @@
...
+import org.mule.api.MuleEvent;
+import org.mule.api.processor.MessageProcessor;
+import org.mule.execution.MessageProcessorExecutionTemplate;
+
+import java.util.List;
+
+/**
+ * ProcessorIterator specifically for use with instances of {@link org.mule.api.service.Service}
+ */
+public class ServiceProcessorIterator extends ProcessorIterator
```
pablokraan: Add unit test for this class

core/src/main/java/org/mule/processor/strategy/NonBlockingProcessingStrategy.java
```diff
@@ -0,0 +1,35 @@
...
+import org.mule.util.concurrent.ThreadNameHelper;
+
+import java.util.List;
+
+/**
+ * This strategy uses a {@link javax.resource.spi.work.WorkManager} to schedule the processing of the pipeline of
+ * message processors
+ * in a single worker thread.
+ */
+public class NonBlockingProcessingStrategy extends SynchronousProcessingStrategy
```
pablokraan: What is the purpose of having this marker class?

core/src/test/java/org/mule/construct/FlowValidationTestCase.java
```diff
@@ -54,6 +58,34 @@ public void testProcessingStrategyCantBeAsyncWithRedelivery() throws Exception
     }
 
     @Test
+    public void testProcessingStrategyNonBlockingSupported() throws Exception
+    {
+        flow.setProcessingStrategy(new NonBlockingProcessingStrategy());
+        flow.setMessageSource(new NonBlockingMessageSource()
```
pablokraan: Can you use a mock here?

core/src/test/java/org/mule/construct/FlowValidationTestCase.java
```diff
@@ -54,6 +58,34 @@ public void testProcessingStrategyCantBeAsyncWithRedelivery() throws Exception
...
+            {
+            }
+        });
+        flow.validateConstruct();
+    }
+
+    @Test(expected = FlowConstructInvalidException.class)
+    public void testProcessingStrategyNonBlockingNotSupported() throws Exception
+    {
+        flow.setProcessingStrategy(new NonBlockingProcessingStrategy());
```
pablokraan: Same

core/src/main/java/org/mule/processor/chain/NonBlockingProcessorIterator.java
dfeist: It could, but why, or what would be the advantage? My feeling is that it belongs in NonBlockingProcessorIterator.

core/src/test/java/org/mule/processor/chain/DefaultMessageProcessorChainTestCase.java
```diff
@@ -60,12 +83,40 @@
...
+        this.nonBlocking = nonBlocking;
+        this.synchronous = synchronous;
+    }
+
+    @Before
+    public void before()
+    {
+        muleContext = mock(MuleContext.class);
+        MuleConfiguration muleConfiguration = mock(MuleConfiguration.class);
+        Mockito.when(muleConfiguration.isContainerMode()).thenReturn(false);
```
pablokraan: Use static imports

core/src/main/java/org/mule/processor/chain/ProcessorIterator.java
dfeist: No, it doesn't have a MuleEvent param.

core/src/main/java/org/mule/processor/chain/ProcessorIterator.java
dfeist: This is also a template method...

core/src/test/java/org/mule/processor/chain/DefaultMessageProcessorChainTestCase.java
```diff
@@ -253,13 +328,15 @@ public void testMixedMPChainWithNullReturn2() throws MuleException, Exception
...
-        builder.chain(new AppendingInterceptingMP("1"), new AppendingMP("2"),
-            new ReturnVoidMPInterceptongMP(), new AppendingMP("3"), new AppendingInterceptingMP("4"),
-            new AppendingMP("5"));
-        assertEquals("0before12after1", builder.build()
-            .process(getTestEventUsingFlow("0"))
-            .getMessage()
-            .getPayload());
+        builder.chain(new AppendingInterceptingMP("1"), getAppendingMP("2"),
+                      new ReturnVoidMPInterceptongMP(), getAppendingMP("3"), new AppendingInterceptingMP("4"),
+                      getAppendingMP("5"));
```
pablokraan: Can you use constants for this test values?

core/src/main/java/org/mule/processor/chain/ProcessorIterator.java
dfeist: This is existing method name, this method was just moved from DefaultMuleProcessorChain, until you think this is major i'd prefer to leave it as it was.

core/src/main/java/org/mule/processor/chain/ProcessorIterator.java
dfeist: Moving this to factory anyway..

core/src/test/java/org/mule/processor/chain/DefaultMessageProcessorChainTestCase.java
```diff
@@ -60,12 +83,40 @@
     @Mock(answer = RETURNS_DEEP_STUBS)
     protected MuleContext muleContext;
 
+    protected MessageExchangePattern exchangePattern;
+    protected boolean nonBlocking;
+    protected boolean synchronous;
+    private volatile int threads = 1;
```
pablokraan: why volatile?

core/src/main/java/org/mule/processor/strategy/NonBlockingProcessingStrategy.java
dfeist: It's not a marker because it's not used to mark anything.  Rather it's a processing strategy implementation that extends SynchronousProcessingStrategy that doesn't currently have any implementation.  Can you think of a better way to allow non-blocking to be configured via a processingStrategy, I think we'd need to redefine interface/job of processingStrategy ideally, which we can't really do in 3.x.

core/src/test/java/org/mule/processor/chain/DefaultMessageProcessorChainTestCase.java
```diff
@@ -851,24 +1241,43 @@ public MuleEvent process(MuleEvent event) throws MuleException
...
         Mockito.when(event.getMuleContext()).thenReturn(muleContext);
         Mockito.when(event.getFlowConstruct()).thenReturn(mock(Flow.class));
-        Mockito.when(muleContext.getConfiguration()).thenReturn(mock(MuleConfiguration.class));
+        Mockito.when(event.getSession()).thenReturn(mock(MuleSession.class));
+        Mockito.when(event.isSynchronous()).thenReturn(synchronous);
         return event;
     }
 
+    @Parameterized.Parameters
+    public static Collection<Object[]> parameters()
```
pablokraan: Move this method to the top of the class after field definitions

core/src/test/java/org/mule/processor/chain/DefaultMessageProcessorChainTestCase.java
dfeist: Could do but I'd prefer not to because I think it makes assertions harder to read without requiring concatenations and I think thats more important in this case.  Anything major that constants would add, because it's really not important if one test method uses "1" and another one uses "one".

core/src/test/java/org/mule/tck/SensingNullMessageProcessor.java
```diff
@@ -6,34 +6,53 @@
...
     protected InternalMessageSource source = new InternalMessageSource();
     private long waitTime = 0;
     public Latch latch = new Latch();
 
-    public MuleEvent process(MuleEvent event) throws MuleException
+    @Override
+    protected void processNonBlocking(final MuleEvent event, NonBlockingCompletionHandler completionHandler) throws MuleException
     {
-        if (waitTime > 0)
+        sleepIfNeeded();
```
pablokraan: Can you avoid a thread.sleep?

core/src/test/java/org/mule/tck/junit4/AbstractMuleTestCase.java
```diff
@@ -55,6 +55,11 @@
     public static final String TEST_TIMEOUT_SYSTEM_PROPERTY = "mule.test.timeoutSecs";
 
     /**
+     * Default timeout for waiting for responses
+     */
+    public static final int RECEIVE_TIMEOUT = 5000;
```
pablokraan: A unit test should not use a timeout of 5 seconds

dfeist: I only moved this up from AbstractMuleContextTestCase, AbstractMuleContextTestCase is also a base unit test case...

core/src/test/java/org/mule/tck/SensingNullMessageProcessor.java
dfeist: This is used in AsyncRequestReplyRequesterTestCase I can't remember exactly why but reworking AsyncRequestReplyRequesterTestCase to not use a sleep is out of scope of this PR.

core/src/main/java/org/mule/processor/chain/NonBlockingProcessorIterator.java
dfeist: Can you expand, I'm not sure I follow this comment..

core/src/main/java/org/mule/processor/chain/AbstractMessageProcessorChain.java
dfeist: No, I'll clean this up, this isn't required.

modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -2477,7 +2477,7 @@
                  <xsd:attribute name="processingStrategy" use="optional" default="default">
                     <xsd:annotation>
                         <xsd:documentation>
-                            Strategy used to process the configured message processors when exchange pattern is 'one-way'.  
+                            Strategy used to process the configured message processors when exchange pattern is 'one-way'.
```
pablokraan: Create a jira in Studio in order to update these attribute descriptions

core/src/main/java/org/mule/DefaultMuleEvent.java
pablokraan: There is only one field defined as ResponseOutputStream in all mule code. So I guess, the rest of the usages are only in order to fill that field. Now that you are removing that usage, I guess is safe to deprecate it, but it could be in a separated issue

core/src/main/java/org/mule/DefaultMuleEvent.java
pablokraan: It was not my intention to add that as part of this PR but just to make a point. It would be good to do it on 3.7

core/src/main/java/org/mule/VoidMuleEvent.java
pablokraan: Not sure about it. We discussed a similar issue the other day related to NullPayload checks.

core/src/main/java/org/mule/api/transport/NonBlockingResponseReplyToHandler.java
pablokraan: But less verbose :)

core/src/main/java/org/mule/execution/AsyncResponseFlowProcessingPhase.java
pablokraan: sure

core/src/main/java/org/mule/processor/chain/NonBlockingProcessorIterator.java
pablokraan: NonBlockingMessageProcessor is used to mark many classes, but this is the only line of code (at least the only one I noted) when NonBlockingMessageProcessor is considered. So, I started to think why this interface is needed and if it is not enough to know that the event being processed is non-blocking in order to execute or not the logic inside that if block.

core/src/main/java/org/mule/processor/chain/NonBlockingProcessorIterator.java
pablokraan: The only place where its called is in the anonymous class and its private. The anonymous class is already using the replyToHandler that is also used in this method

core/src/main/java/org/mule/processor/chain/ProcessorIterator.java
pablokraan: a template method is a method where you define the steps and let subclasses to define the behaviour of that steps. There a no steps here

core/src/main/java/org/mule/processor/strategy/NonBlockingProcessingStrategy.java
pablokraan: Right, is not a marker interface because there is behaviour defined in the parent interface. What I was trying to say is why we need to define a new concept that does exactly the same that its parent.
We can talk about this to see if we find a better solution

modules/spring-config/src/main/resources/META-INF/mule.xsd
dfeist: STUDIO-6341 STUDIO-6342

core/src/main/java/org/mule/processor/chain/NonBlockingProcessorIterator.java
dfeist: Let's talk about this one..

core/src/main/java/org/mule/processor/chain/NonBlockingProcessorIterator.java
dfeist: Let's discuss this one if you think it's importante.  I think it's harder to read putting it anonymous class and also the it's the iterator thats being resumed, not the anonymous class.  So it it was to go in callback then it should be renamed to resumeIterator or something.  I'm not convinced though.

core/src/test/java/org/mule/tck/junit4/AbstractMuleTestCase.java
pablokraan: AbstractMuleTestCase is supposed to be for really unit tests (the small ones), so a timeout of 5 seconds is not appropriate

dfeist: Agree, but 5000 is also far to long for AbstractMuleContextTestCase IMO. Anyway i'll revert this and use a local latch timeout for this instead.

PR #1617, opened 2015-04-21
MULE-8512: Allow default processing strategy to be configured per app

marianogonzalez:


core/src/main/java/org/mule/construct/AbstractPipeline.java
```diff
@@ -91,6 +90,24 @@ protected MessageProcessor createPipeline() throws MuleException
         return builder.build();
     }
 
+    protected ProcessingStrategy createDefaultProcessingStrategy()
```
dfeist: Default can't only go in MuleConfiguration?  Or is this because of need to test without MueContext etc.?

modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -1366,7 +1366,7 @@
                         </xsd:documentation>
                     </xsd:annotation>
                 </xsd:attribute>
-                <xsd:attribute name="flowEndingWithOneWayEndpointReturnsNull" type="substitutableBoolean" default="false">
+                <xsd:attribute name="flowEndingWithOneWayEndpointReturnsNull" type="substitutableBoolean" use="optional">
```
dfeist: Is this change related to this PR?

modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -2474,74 +2543,13 @@
                         </xsd:documentation>
                     </xsd:annotation>
                 </xsd:attribute>
-                 <xsd:attribute name="processingStrategy" use="optional" default="default">
```
dfeist: Add simple comment, saying where default is defined now.

modules/spring-config/src/main/resources/META-INF/mule.xsd
marianogonzalez: no, this looks like a mistake of some sort

core/src/main/java/org/mule/construct/AbstractPipeline.java
marianogonzalez: this method is a fallback for creating the processing strategy you'll use if the user hasn't either set something on the flow, nor the configuration, nor the system property. Will add javadoc to better explain

PR #1618, opened 2015-04-21
MULE-8512: Allow default processing strategy to be configured per app

marianogonzalez:


core/src/main/java/org/mule/construct/AbstractPipeline.java
```diff
@@ -91,6 +92,34 @@ protected MessageProcessor createPipeline() throws MuleException
...
+     * {@link #setProcessingStrategy(ProcessingStrategy)},
+     * through {@link MuleConfiguration#getDefaultProcessingStrategy()}
+     * or the {@link MuleProperties#MULE_DEFAULT_PROCESSING_STRATEGY}
+     * system property
+     *
+     * @return a {@link SynchronousProcessingStrategy}
+     */
+    protected ProcessingStrategy createDefaultProcessingStrategy()
+    {
+        return new SynchronousProcessingStrategy();
```
dfeist: In the end is it no possible to only define processing strategy default gloablly? Is this because of need to test without muleContext?

modules/spring-config/src/main/java/org/mule/config/spring/MuleConfigurationConfigurator.java
```diff
@@ -84,6 +89,22 @@ public Object getObject() throws Exception
...
 
+    private void determineDefaultProcessingStrategy(DefaultMuleConfiguration defaultConfig)
+    {
+        if (config.getDefaultProcessingStrategy() != null)
+        {
+            defaultConfig.setDefaultProcessingStrategy(config.getDefaultProcessingStrategy());
+        }
+        else
+        {
+            String processingStrategyFromSystemProperty = System.getProperty(MuleProperties.MULE_DEFAULT_PROCESSING_STRATEGY);
```
dfeist: In all other cases we read/use the system property in DefaultMuleConfiguration, any reason we can't do same in this case?

core/src/main/java/org/mule/construct/AbstractPipeline.java
marianogonzalez: this method only gets invoked if the mule configuration doesn't specify a default. Look at the initialiseProcessingStrategy() method

modules/spring-config/src/main/java/org/mule/config/spring/MuleConfigurationConfigurator.java
marianogonzalez: simply wasn't aware of that. will change it

dfeist: Look at responseTimeout default.  That can be configured via system property and spring.

MuleProperties.SYSTEM_PROPERTY_PREFIX + "timeout.synchronous"

marianogonzalez: on second thoughts, I think that the implementation I did here is better than that.

First of all, applySystemProperties() is already quite ugly as it is like to add yet another if to it. Secondly, if I do it there, I would have to give DefaultMuleConfiguration some level of knowledge about how to transform the system property string into a ProcessingStrategy (even if that's by delegating into the ProcessingStrategyUtils). On the other hand, as this is implemented now, I can use a PropertyEditor to decouple that.

I'm sticking to the current implementation

core/src/main/java/org/mule/construct/AbstractPipeline.java
marianogonzalez: this method is needed because Flow and AbstractPipeline have defaults of different types

PR #1641, opened 2015-04-27
MULE-8545 Update spring schemas to 3.7

vromero:


vromero:
close as one change that sholdnt be the file slipped in

PR #1646, opened 2015-04-27
MULE-8524 Add direct threading profile which can be used HTTP listener t...

dfeist:
...o allow selectors to be used without workers

### DIFF
--- a/core/src/main/java/org/mule/config/ChainedThreadingProfile.java
+++ b/core/src/main/java/org/mule/config/ChainedThreadingProfile.java
@@ -83,7 +83,7 @@ public class ChainedThreadingProfile implements ThreadingProfile
      */
     public ChainedThreadingProfile(ThreadingProfile delegate, boolean dynamic)
     {
-        if (!dynamic)
+        if (!(dynamic || delegate instanceof DirectThreadingProfile))
         {
             // for static dependencies, we delegate to a fixed copy
             delegate = new ImmutableThreadingProfile(delegate);
@@ -184,7 +184,7 @@ public class ChainedThreadingProfile implements ThreadingProfile
     public WorkManager createWorkManager(String name, int shutdownTimeout)
     {
         // we deliberately don't instantiate the chained profile as we just want a cloned copy, not recursion
-        return workManagerFactory.createWorkManager(new ImmutableThreadingProfile(this), name, shutdownTimeout);
+        return workManagerFactory.createWorkManager(this, name, shutdownTimeout);
     }
 
     public ExecutorService createPool()
@@ -195,7 +195,7 @@ public class ChainedThreadingProfile implements ThreadingProfile
     public ExecutorService createPool(String name)
     {
         // we deliberately don't instantiate the chained profile as we just want a cloned copy, not recursion
-        return poolFactory.createPool(name, new ImmutableThreadingProfile(this));
+        return poolFactory.createPool(name, this);
     }
 
     public boolean isDoThreading()
@@ -216,7 +216,7 @@ public class ChainedThreadingProfile implements ThreadingProfile
     @Override
     public ScheduledExecutorService createScheduledPool(String name)
     {
-        return poolFactory.createScheduledPool(name, new ImmutableThreadingProfile(this));
+        return poolFactory.createScheduledPool(name, this);
     }
 
     public MuleContext getMuleContext()
--- a/core/src/main/java/org/mule/config/ChainedThreadingProfile.java
+++ b/core/src/main/java/org/mule/config/ChainedThreadingProfile.java
@@ -76,14 +76,18 @@ public class ChainedThreadingProfile implements ThreadingProfile
     /**
      * Generate a mutable threading profile.  Default values are taken from the "delegate"
      * argument.  If dynamic is true then changes in the delegate instance are reflected in
-     * this instance.
+     * this instance.  DirectThreadingProfile instances cannot be chained.
      * 
      * @param delegate Source of default values.
      * @param dynamic If true, changes in delegate are reflected in this instance
      */
     public ChainedThreadingProfile(ThreadingProfile delegate, boolean dynamic)
     {
-        if (!(dynamic || delegate instanceof DirectThreadingProfile))
+        if (delegate instanceof DirectThreadingProfile)
+        {
+            throw new UnsupportedOperationException("DirectThreadingProfile instances cannot be chained.");
+        }
+        if (!dynamic)
         {
             // for static dependencies, we delegate to a fixed copy
             delegate = new ImmutableThreadingProfile(delegate);
--- a/core/src/main/java/org/mule/config/ChainedThreadingProfile.java
+++ b/core/src/main/java/org/mule/config/ChainedThreadingProfile.java
@@ -183,7 +183,6 @@ public class ChainedThreadingProfile implements ThreadingProfile
 
     public WorkManager createWorkManager(String name, int shutdownTimeout)
     {
-        // we deliberately don't instantiate the chained profile as we just want a cloned copy, not recursion
         return workManagerFactory.createWorkManager(this, name, shutdownTimeout);
     }
 
@@ -194,7 +193,6 @@ public class ChainedThreadingProfile implements ThreadingProfile
 
     public ExecutorService createPool(String name)
     {
-        // we deliberately don't instantiate the chained profile as we just want a cloned copy, not recursion
         return poolFactory.createPool(name, this);
     }
 
--- a/core/src/main/java/org/mule/config/DirectThreadingProfile.java
+++ b/core/src/main/java/org/mule/config/DirectThreadingProfile.java
@@ -24,7 +24,7 @@ import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 
 /**
- * {@link org.mule.api.config.ThreadingProfile} implementation that executes work in the callee thread.
+ * {@link org.mule.api.config.ThreadingProfile} implementation that executes work in the caller thread.
  */
 public class DirectThreadingProfile extends ImmutableThreadingProfile
 {
@@ -36,7 +36,7 @@ public class DirectThreadingProfile extends ImmutableThreadingProfile
 
     public DirectThreadingProfile(ThreadingProfile delegate)
     {
-        // Ignore delegate threading profie
+        // Ignore delegate threading profie.  This constructor is used by Spring.
         this();
     }
 
--- a/core/src/main/java/org/mule/config/DirectThreadingProfile.java
+++ b/core/src/main/java/org/mule/config/DirectThreadingProfile.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.config;
+
+import org.mule.api.MuleContext;
+import org.mule.api.config.ThreadingProfile;
+import org.mule.api.context.MuleContextAware;
+import org.mule.api.context.WorkManager;
+import org.mule.config.pool.DefaultThreadPoolFactory;
+import org.mule.config.pool.ThreadPoolFactory;
+import org.mule.util.concurrent.NamedThreadFactory;
+
+import com.google.common.util.concurrent.MoreExecutors;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.RejectedExecutionHandler;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+
+/**
+ * {@link org.mule.api.config.ThreadingProfile} implementation that executes work in the callee thread.
+ */
+public class DirectThreadingProfile extends ImmutableThreadingProfile
+{
+
+    public DirectThreadingProfile()
+    {
+        super(0, 0, 0, 0, 0, 0, false, new ThreadPoolExecutor.CallerRunsPolicy(), new NullThreadFactory());
+    }
+
+    public DirectThreadingProfile(ThreadingProfile delegate)
+    {
+        // Ignore delegate threading profie
+        this();
+    }
+
+    @Override
+    public ExecutorService createPool(String name)
+    {
+        return MoreExecutors.newDirectExecutorService();
+    }
+
+    @Override
+    public ScheduledExecutorService createScheduledPool(String name)
+    {
+        throw new UnsupportedOperationException("This implementation does not support the creation of scheduled pools");
+    }
+
+    private static class NullThreadFactory implements ThreadFactory
+    {
+
+        @Override
+        public Thread newThread(Runnable r)
+        {
+            throw new UnsupportedOperationException();
+        }
+    }
+}
--- a/core/src/main/java/org/mule/config/ImmutableThreadingProfile.java
+++ b/core/src/main/java/org/mule/config/ImmutableThreadingProfile.java
@@ -162,7 +162,7 @@ public class ImmutableThreadingProfile implements ThreadingProfile
 
     public WorkManager createWorkManager(String name, int shutdownTimeout)
     {
-        return workManagerFactory.createWorkManager(new ImmutableThreadingProfile(this), name, shutdownTimeout);
+        return workManagerFactory.createWorkManager(this, name, shutdownTimeout);
     }
 
     public ExecutorService createPool()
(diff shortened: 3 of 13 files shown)


core/src/main/java/org/mule/config/ChainedThreadingProfile.java
```diff
@@ -83,7 +83,7 @@ public ChainedThreadingProfile(ThreadingProfile delegate)
      */
     public ChainedThreadingProfile(ThreadingProfile delegate, boolean dynamic)
     {
-        if (!dynamic)
+        if (!(dynamic || delegate instanceof DirectThreadingProfile))
```
pablokraan: Can not "delegate instanceof DirectThreadingProfile"  be replaced with "!delegate.isDoThreading()"?

core/src/main/java/org/mule/config/ChainedThreadingProfile.java
```diff
@@ -184,7 +184,7 @@ public void setWorkManagerFactory(WorkManagerFactory workManagerFactory)
     public WorkManager createWorkManager(String name, int shutdownTimeout)
     {
         // we deliberately don't instantiate the chained profile as we just want a cloned copy, not recursion
-        return workManagerFactory.createWorkManager(new ImmutableThreadingProfile(this), name, shutdownTimeout);
+        return workManagerFactory.createWorkManager(this, name, shutdownTimeout);
```
pablokraan: Comment is previous line seems wrong now

pablokraan: I don't know this class in detail, but javadoc says that we should only use ImmutableThreadingProfile instances

core/src/main/java/org/mule/config/ChainedThreadingProfile.java
dfeist: I'm not sure there is a direct relationship between the two.  The issue here is that currently threadProfile is always copied, get the copy mechanism (copy to immutableThreadingProfile) over-rides all implementation behaviour because it does not delegate.  So what i have had to do (to avoid bigger changes) is simply not copy in the case of the direct implementation but continue copying as today in the other cases.

core/src/main/java/org/mule/config/DirectThreadingProfile.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.RejectedExecutionHandler;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+
+/**
+ * {@link org.mule.api.config.ThreadingProfile} implementation that executes work in the callee thread.
```
pablokraan: callee -> caller?

core/src/main/java/org/mule/config/DirectThreadingProfile.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public DirectThreadingProfile()
+    {
+        super(0, 0, 0, 0, 0, 0, false, new ThreadPoolExecutor.CallerRunsPolicy(), new NullThreadFactory());
+    }
+
+    public DirectThreadingProfile(ThreadingProfile delegate)
+    {
+        // Ignore delegate threading profie
+        this();
```
pablokraan: Why is this constructor required?

core/src/main/java/org/mule/config/ChainedThreadingProfile.java
dfeist: I've moved the responsibility for copied down to the workManagerFactory itself. Will update javadoc.

core/src/main/java/org/mule/config/DirectThreadingProfile.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public ExecutorService createPool(String name)
+    {
+        return MoreExecutors.newDirectExecutorService();
```
pablokraan: What about the pool name?

core/src/main/java/org/mule/config/ImmutableThreadingProfile.java
```diff
@@ -162,7 +162,7 @@ public void setWorkManagerFactory(WorkManagerFactory workManagerFactory)
 
     public WorkManager createWorkManager(String name, int shutdownTimeout)
     {
-        return workManagerFactory.createWorkManager(new ImmutableThreadingProfile(this), name, shutdownTimeout);
+        return workManagerFactory.createWorkManager(this, name, shutdownTimeout);
```
pablokraan: Are you sure this change does not break anything?

core/src/main/java/org/mule/config/pool/DefaultThreadPoolFactory.java
```diff
@@ -28,19 +30,20 @@
 
     public ThreadPoolExecutor createPool(String name, ThreadingProfile tp)
     {
+        ThreadingProfile threadingProfileCopy = tp instanceof DirectThreadingProfile ? tp : new ImmutableThreadingProfile(tp);
```
pablokraan: Is not good to have "instanceof DirectThreadingProfile" in different places. I'm not asking for a shared method to check that, is just that we need to remind to add this check when we add new threading profile related code

core/src/main/java/org/mule/config/DirectThreadingProfile.java
dfeist: Spring uses it.  Will add comment.

core/src/main/java/org/mule/config/DirectThreadingProfile.java
dfeist: Pool name is only normally used for configuring the NamedThreadFactory.  Here we have no threadFactory and no threads, so it's not required.

core/src/main/java/org/mule/config/pool/DefaultThreadPoolFactory.java
pablokraan: Not sure I'm following this change: did you removed the creation of an ImmutableThreadingProfile from all the other places and just added it here?

core/src/main/java/org/mule/config/ImmutableThreadingProfile.java
dfeist: The clone still happens, just further down instead of being repeated in every threadProfile implementation.

core/src/main/java/org/mule/config/pool/DefaultThreadPoolFactory.java
dfeist: This is required because the existing code is so bad, i hate it too.  I suggest we create an issue to improve this code, as there is not time to do this in scope of this issue.

core/src/main/java/org/mule/config/pool/DefaultThreadPoolFactory.java
```diff
@@ -81,8 +84,9 @@ private void configureThreadPoolExecutor(String name, ThreadingProfile tp, Threa
     @Override
     public ScheduledThreadPoolExecutor createScheduledPool(String name, ThreadingProfile tp)
     {
-        ScheduledThreadPoolExecutor pool = internalCreateScheduledPool(tp);
-        configureThreadPoolExecutor(name, tp, pool);
+        ThreadingProfile threadingProfileCopy = tp instanceof DirectThreadingProfile ? tp : new ImmutableThreadingProfile(tp);
```
pablokraan: Why can't you create an ImmutableThreadingProfile from a DirectThreadingProfile in order to avoid all these instanceof checks?

core/src/main/java/org/mule/work/MuleWorkManager.java
```diff
@@ -81,7 +83,7 @@ public MuleWorkManager(ThreadingProfile profile, String name, int shutdownTimeou
             name = "WorkManager#" + hashCode();
         }
 
-        this.threadingProfile = new ImmutableThreadingProfile(profile);
+        this.threadingProfile = profile instanceof DirectThreadingProfile ? profile : new ImmutableThreadingProfile(profile);
```
pablokraan: same ugly check again

core/src/test/java/org/mule/config/DirectThreadingProfileTestCase.java
```diff
@@ -0,0 +1,56 @@
...
+
+import junit.framework.Assert;
+import org.hamcrest.CoreMatchers;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+public class DirectThreadingProfileTestCase extends AbstractMuleContextTestCase
+{
+
+    private Thread workThread;
```
pablokraan: Convert this field to variable

core/src/test/java/org/mule/config/DirectThreadingProfileTestCase.java
```diff
@@ -0,0 +1,56 @@
...
+import org.mule.tck.junit4.AbstractMuleContextTestCase;
+
+import java.util.concurrent.Callable;
+
+import junit.framework.Assert;
+import org.hamcrest.CoreMatchers;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+public class DirectThreadingProfileTestCase extends AbstractMuleContextTestCase
```
pablokraan: Why not an AbstractMuleTestCase?

modules/http/src/main/java/org/mule/module/http/internal/config/HttpNamespaceHandler.java
```diff
@@ -66,6 +67,8 @@ public void init()
 
         registerBeanDefinitionParser("raml-api-configuration", new ChildDefinitionParser("apiConfiguration", RamlApiConfiguration.class));
         registerBeanDefinitionParser("worker-threading-profile", new ThreadingProfileDefinitionParser("workerThreadingProfile",  MuleProperties.OBJECT_DEFAULT_MESSAGE_RECEIVER_THREADING_PROFILE));
+        registerBeanDefinitionParser("direct-threading-profile", new ThreadingProfileDefinitionParser("workerThreadingProfile",  MuleProperties.OBJECT_DEFAULT_MESSAGE_RECEIVER_THREADING_PROFILE,
```
pablokraan: "workerThreadingProfile" -> "directThreadingProfile"?

modules/http/src/main/resources/META-INF/mule-httpn.xsd
```diff
@@ -117,7 +117,16 @@
             <xsd:extension base="mule:abstractExtensionType">
                 <xsd:sequence>
                     <xsd:element ref="tls:context" minOccurs="0" maxOccurs="1" />
-                    <xsd:element name="worker-threading-profile" type="mule:asynchronousThreadingProfileType" minOccurs="0" maxOccurs="1"/>
+                    <xsd:choice>
+                        <xsd:element name="worker-threading-profile" type="mule:asynchronousThreadingProfileType" minOccurs="0" maxOccurs="1"/>
+                        <xsd:element name="direct-threading-profile" minOccurs="0" maxOccurs="1">
+                            <xsd:annotation>
+                                <xsd:documentation>
+                                    Use this threading profile to disable the use of a worker thread pool for processing requests and instead process requests using the selector threads.
```
pablokraan: process _> processing?

modules/http/src/test/java/org/mule/module/http/functional/proxy/HttpProxyTemplateTestCase.java
```diff
@@ -49,11 +59,25 @@
 
     private RequestHandlerExtender handlerExtender;
     private boolean consumeAllRequest = true;
+    private String configFile;
+    private String requestThreadContains;
+
+    @Parameterized.Parameters
+    public static Collection<Object[]> parameters()
+    {
+        return Arrays.asList(new Object[][] {{"http-proxy-template-config.xml","worker"}, {"http-proxy-template-selectors-config.xml", "SelectorRunner"}});
```
pablokraan: fix code style

modules/http/src/test/java/org/mule/module/http/functional/proxy/HttpProxyTemplateTestCase.java
```diff
@@ -49,11 +59,25 @@
...
+    }
+
+    public HttpProxyTemplateTestCase(String configFile, String requestThreadContains)
```
pablokraan: requestThreadContains > expectedThreadName?

modules/http/src/test/resources/http-proxy-template-selectors-config.xml
```diff
@@ -0,0 +1,42 @@
...
+               http://www.mulesoft.org/schema/mule/http http://www.mulesoft.org/schema/mule/http/current/mule-http.xsd
+               http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-current.xsd">
+
+
+    <spring:bean name="sensingMessageProcessor" class="org.mule.tck.SensingNullMessageProcessor"/>
+
+    <http:listener-config name="proxyConfig" host="localhost" port="${proxyPort}" >
+        <http:direct-threading-profile />
+    </http:listener-config>
+    <http:request-config name="requestConfig" host="localhost" port="${httpPort}" />
```
pablokraan: Can you add some empty line to make this XML more readable?

modules/http/src/test/java/org/mule/module/http/functional/proxy/HttpProxyTemplateTestCase.java
dfeist: I don't know the thread name.  This is primarily to determine if it is a selector or worker thread.

modules/http/src/main/resources/META-INF/mule-httpn.xsd
dfeist: Do you mean "for processing requests and instead process requests"?  This is fine.

modules/http/src/main/java/org/mule/module/http/internal/config/HttpNamespaceHandler.java
dfeist: I hadn't change this to not change method on HttpListenerConfig.  But i've just look at it's not in interface, so will update.

core/src/test/java/org/mule/config/DirectThreadingProfileTestCase.java
dfeist: Fixed.

core/src/test/java/org/mule/config/DirectThreadingProfileTestCase.java
dfeist: Can't without creating some sort of holder for it, as it needs to be final, yet if it's final it can't be set from in anonymous class.  Ok to leave as it is?

core/src/main/java/org/mule/config/pool/DefaultThreadPoolFactory.java
dfeist: No, thats whats so horrible about ImmutableThreadingProfie, it doesn't wrap/delegate, rather it copies fields and then provides it's own implementation for createPool, createWorkManager etc.

core/src/main/java/org/mule/config/ChainedThreadingProfile.java
```diff
@@ -83,7 +83,7 @@ public ChainedThreadingProfile(ThreadingProfile delegate)
      */
     public ChainedThreadingProfile(ThreadingProfile delegate, boolean dynamic)
```
pablokraan: Update class javadoc

core/src/main/java/org/mule/config/pool/DefaultThreadPoolFactory.java
pablokraan: Is not better to fix ImmutableThreadingProfile instead of adding more noise?

core/src/test/java/org/mule/config/DirectThreadingProfileTestCase.java
pablokraan: what about an final AtomicReference?

core/src/main/java/org/mule/config/ChainedThreadingProfile.java
dfeist: I changed implementation in the end, and updated method javadoc.

core/src/main/java/org/mule/config/pool/DefaultThreadPoolFactory.java
dfeist: Yes, but not in scope of this PR. And given this PR has config changes that we need to make today..

core/src/test/java/org/mule/config/DirectThreadingProfileTestCase.java
dfeist: That would work if you think it's worth to making that change..

modules/http/src/test/java/org/mule/module/http/functional/proxy/HttpProxyTemplateTestCase.java
pablokraan: The variable name is not clear to me, you need to read to code to understand what its used for.

core/src/main/java/org/mule/config/DirectThreadingProfile.java
pablokraan: Is not a bug that spring invokes this constructor?

dfeist: No, it's explicit in code.  Buy what i'll do is create a custom ThreadingProfileDefinitionParser for this case.  This will be a better solution.

PR #1648, opened 2015-04-27
MULE-8513 Add support to HTTP listener/requester to leverage non-blockin...

dfeist:
...g processing strategy


dfeist:
REVIEW ONLY. DO NOT MERGE.

pablolagreca:
There are no functional test cases for error handling scenarios. Please add some.

dfeist:
The error case is currently covered with this test method: org.mule.module.http.functional.proxy.HttpProxyTemplateErrorTestCase#failIfTargetServiceIsDown.  I look to see if there are any other existing error test cases which I can paramatize to extend for the non-blocking case.

PR #1650, opened 2015-04-28
MULE-8535: Ensure that transformers use a proper return datatype

pablokraan:


core/src/main/java/org/mule/transformer/simple/ObjectToString.java
```diff
@@ -39,9 +39,7 @@ public ObjectToString()
         registerSourceType(DataTypeFactory.BYTE_ARRAY);
         registerSourceType(DataTypeFactory.INPUT_STREAM);
         registerSourceType(DataTypeFactory.create(OutputHandler.class));
-        //deliberately set the mime for this transformer to text plain so that other transformers
-        //that serialize string types such as XML or JSON will not match this
-        setReturnDataType(DataTypeFactory.TEXT_STRING);
+        setReturnDataType(DataTypeFactory.STRING);
```
marianogonzalez: no test cases? what's the impact of this? sounds deliberate

pablokraan: This changes is needed to better propagate the message's datatype. There were no tests for this and I'm not sure that adding tests for this will add value

PR #1662, opened 2015-04-29
MULE-8513 Add support to HTTP listener/requester to leverage non-blockin...

dfeist:
...g processing strategy. Fix failing test.


core/src/main/java/org/mule/exception/TemplateMessagingExceptionStrategy.java
```diff
@@ -45,7 +47,7 @@ final public MuleEvent handleException(Exception exception, MuleEvent event)
             event.getMessage().setExceptionPayload(new DefaultExceptionPayload(exception));
 
             // MULE-8551 Still need to add support for non-blocking components in excepton strategies.
-            if(event.isAllowNonBlocking())
+            if(nonBlocking)
```
pablokraan: Please add a unit test for this as there is no test in CE failing because of this change

PR #1669, opened 2015-04-29
MULE-8550: Transformers are registered twice

marianogonzalez:


pablolagreca:
Seems to me that this is not fixing the root cause of the problem.

PR #1670, opened 2015-04-29
MULE-8519: Add a trace/debug log when a Validation was done successfully

andresgregoire:


extensions/validation/src/main/java/org/mule/extension/validation/internal/ValidationSupport.java
```diff
@@ -24,6 +27,7 @@
  */
 abstract class ValidationSupport
 {
+    protected final Logger logger = LoggerFactory.getLogger(getClass());
```
marianogonzalez: static loggers perform better

PR #1677, opened 2015-04-30
MULE-8556: SpringRegistry is not thread-safe

marianogonzalez:


tests/integration/src/test/java/org/mule/config/spring/SpringRegistryConcurrencyTestCase.java
```diff
@@ -0,0 +1,161 @@
...
+    {
+        return "spring-registry-concurrency-config.xml";
+    }
+
+    @Override
+    protected void doSetUp() throws Exception
+    {
+        latch = new Latch();
+        auxiliaryLatch = new Latch();
+        lookupObjects = null;
```
pablokraan: This line should not be required

tests/integration/src/test/java/org/mule/config/spring/SpringRegistryConcurrencyTestCase.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    @Test
+    public void concurrentReadAndWrite() throws Exception
+    {
+        final Thread thread = newReaderThread();
+
+        Object object = new Startable()
```
pablokraan: object -> startableObject?

tests/integration/src/test/java/org/mule/config/spring/SpringRegistryConcurrencyTestCase.java
marianogonzalez: might not be strictly required. But it's a good practice to clear state completely since the tests do write to that field

tests/integration/src/test/java/org/mule/config/spring/SpringRegistryConcurrencyTestCase.java
marianogonzalez: yes my patroncito

PR #1682, opened 2015-05-04
MULE-8555: Promote to public api DefaultHttpListenerConfig.resolvePath

svacas:


aiannucci:
Does this change affect 3.7? If so, the PR should be to 3.x branch and then backported to 3.6.x

modules/http/src/main/java/org/mule/module/http/api/listener/HttpListenerConfig.java
```diff
@@ -34,6 +34,12 @@
     boolean hasTlsConfig();
 
     /**
+     * @param listenerPath path of the listener to be resolved
```
marianogonzalez: add a better javadoc

PR #1693, opened 2015-05-06
MULE-8216: Implement a validations extensions [For Mariano Gonzalez]

asequeira:
Adding more tests

### DIFF
--- a/extensions/validation/src/test/java/org/mule/extension/validation/BasicValidationTestCase.java
+++ b/extensions/validation/src/test/java/org/mule/extension/validation/BasicValidationTestCase.java
@@ -14,6 +14,7 @@ import static org.junit.Assert.assertThat;
 import static org.junit.Assert.fail;
 import static org.mule.extension.validation.internal.ImmutableValidationResult.error;
 import static org.mule.extension.validation.internal.ValidationExtension.DEFAULT_LOCALE;
+
 import org.mule.api.MuleEvent;
 import org.mule.extension.validation.api.MultipleValidationException;
 import org.mule.extension.validation.api.MultipleValidationResult;
@@ -24,6 +25,7 @@ import org.mule.util.ExceptionUtils;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 
 import java.util.Arrays;
 import java.util.HashMap;
@@ -47,7 +49,6 @@ public class BasicValidationTestCase extends ValidationTestCase
     public void email() throws Exception
     {
         assertValid(EMAIL_VALIDATION_FLOW, getTestEvent(VALID_EMAIL));
-
         assertInvalidEmail(INVALID_EMAIL);
         assertInvalidEmail(" " + VALID_EMAIL);
         assertInvalidEmail(VALID_EMAIL + " ");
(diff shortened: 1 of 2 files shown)


extensions/validation/src/test/java/org/mule/extension/validation/BasicValidationTestCase.java
```diff
@@ -149,22 +156,17 @@ public void isFalse() throws Exception
     public void notEmpty() throws Exception
     {
         final String flowName = "notEmpty";
-
```
marianogonzalez: restore this empty line

extensions/validation/src/test/java/org/mule/extension/validation/BasicValidationTestCase.java
```diff
@@ -79,21 +88,19 @@ public void url() throws Exception
...
-    private MuleEvent getTimeEvent(String time, String pattern) throws Exception {
+    private MuleEvent getTimeEvent(String time, String pattern) throws Exception
+    {
         MuleEvent event = getTestEvent(time);
         event.setFlowVariable("pattern", pattern);
-
         return event;
     }
 
+
```
marianogonzalez: remove line

extensions/validation/src/test/java/org/mule/extension/validation/BasicValidationTestCase.java
```diff
@@ -14,6 +14,7 @@
 import static org.junit.Assert.fail;
 import static org.mule.extension.validation.internal.ImmutableValidationResult.error;
 import static org.mule.extension.validation.internal.ValidationExtension.DEFAULT_LOCALE;
+
```
marianogonzalez: remove empty line

extensions/validation/src/test/java/org/mule/extension/validation/BasicValidationTestCase.java
```diff
@@ -193,8 +191,6 @@ public void successfulAll() throws Exception
     {
         MuleEvent event = getAllEvent(VALID_EMAIL, VALID_URL);
         MuleEvent responseEvent = runFlow("all", event);
-
-        // validation was successful and payload was untouched
```
marianogonzalez: restore the comment

extensions/validation/src/test/java/org/mule/extension/validation/ValidationTestCase.java
```diff
@@ -26,7 +26,6 @@
 
     static final String VALID_URL = "http://localhost:8080";
     static final String INVALID_URL = "here";
-
```
marianogonzalez: restore line

extensions/validation/src/test/java/org/mule/extension/validation/BasicValidationTestCase.java
asequeira: I don't think is a good practice to add empty lines in our code.

extensions/validation/src/test/java/org/mule/extension/validation/BasicValidationTestCase.java
asequeira: We shouldn't add comments to the test, perhaps a better solution would be to change the method name to reflect what are we trying to prove.

extensions/validation/src/test/java/org/mule/extension/validation/BasicValidationTestCase.java
asequeira: Is there a policy about which import statements allow an empty line?

extensions/validation/src/test/java/org/mule/extension/validation/BasicValidationTestCase.java
marianogonzalez: we use empty lines to separate sections on our tests and make it more readable all the time. Actually, the original PR passed code review in that form. If you check other test cases of other modules you'll find that all over the place

extensions/validation/src/test/java/org/mule/extension/validation/BasicValidationTestCase.java
marianogonzalez: intellij code stylesheet has the policy embedded. if it was intellij who inserted that line then fine

extensions/validation/src/test/java/org/mule/extension/validation/BasicValidationTestCase.java
marianogonzalez: agreed. change the name

extensions/validation/src/test/java/org/mule/extension/validation/BasicValidationTestCase.java
asequeira: We should agree on a way for doing this and stick to it always or leave the decision to the committer.

marianogonzalez: AFAIK you're the first to disagree. This is a common practice

extensions/validation/src/test/java/org/mule/extension/validation/ValidationTestCase.java
asequeira: I don't see why we should have lines between constants, a formatter tool will never understand that.

marianogonzalez: the goal is to make it more readable to the human. No formatter tool covers 100% of the cases

extensions/validation/src/test/java/org/mule/extension/validation/BasicValidationTestCase.java
asequeira: I understand that for splitting setup, execution and assertions but not for declaring a variable. Anyway, we should agree to split these sections always or never do it.
For this particular case I'm creating a constant, which seems to be the correct solution. Is that OK for you?

marianogonzalez: true that we should agree on something and always do it, although at least on the PR's I've review we always do it. If you want bring this up on post-scrum and we can hold the merge. 'Til then, I want my empty line :)

PR #1694, opened 2015-05-06
Fix MULE-8563. Add the posibility to define a name different from the fi...

pablocabrera85:
Add @JsonProperty handling to allow developer to use a different property names when adding extra authorization parameters

### DIFF
--- a/modules/devkit-support/src/main/java/org/mule/security/oauth/processor/BaseOAuth2AuthorizeMessageProcessor.java
+++ b/modules/devkit-support/src/main/java/org/mule/security/oauth/processor/BaseOAuth2AuthorizeMessageProcessor.java
@@ -7,6 +7,7 @@
 
 package org.mule.security.oauth.processor;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import org.mule.api.MessagingException;
 import org.mule.api.MuleEvent;
 import org.mule.api.MuleException;
@@ -118,15 +119,16 @@ public abstract class BaseOAuth2AuthorizeMessageProcessor<T extends OAuth2Manage
         {
             for (AuthorizationParameter<?> parameter : params)
             {
+                String parameterName = parameter.getName();
                 Field field = null;
                 try
                 {
-                    field = this.getClass().getDeclaredField(parameter.getName());
+                    field = this.getClass().getDeclaredField(parameterName);
                 }
                 catch (NoSuchFieldException e)
                 {
                     throw new MessagingException(CoreMessages.createStaticMessage(String.format(
-                        "Code generation error. Field %s should be present in class", parameter.getName())),
+                        "Code generation error. Field %s should be present in class", parameterName)),
                         event, e);
                 }
 
@@ -134,12 +136,16 @@ public abstract class BaseOAuth2AuthorizeMessageProcessor<T extends OAuth2Manage
 
                 try
                 {
+                    JsonProperty property=field.getAnnotation(JsonProperty.class);
+                    if(property!=null && StringUtils.isNotEmpty(property.value())){
+                        parameterName = property.value();
+                    }
                     Object value = field.get(this);
                     if (value != null)
                     {
                         Object transformed = this.evaluateAndTransform(getMuleContext(), event,
                             parameter.getType(), null, value);
-                        extraParameters.put(parameter.getName(), this.toString(event, transformed)
+                        extraParameters.put(parameterName, this.toString(event, transformed)
                             .toLowerCase());
                     }
                 }
--- a/modules/devkit-support/src/test/java/org/mule/security/oauth/processor/OAuth2AuthorizeMessageProcessorTestCase.java
+++ b/modules/devkit-support/src/test/java/org/mule/security/oauth/processor/OAuth2AuthorizeMessageProcessorTestCase.java
@@ -6,6 +6,7 @@
  */
 package org.mule.security.oauth.processor;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import org.mule.api.MuleContext;
 import org.mule.api.MuleEvent;
 import org.mule.api.MuleException;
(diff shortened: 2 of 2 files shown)


modules/devkit-support/src/main/java/org/mule/security/oauth/processor/BaseOAuth2AuthorizeMessageProcessor.java
```diff
@@ -134,12 +136,16 @@ private void setState(Map<String, String> extraParameters, MuleEvent event)
 
                 try
                 {
+                    JsonProperty property=field.getAnnotation(JsonProperty.class);
```
marianogonzalez: do not couple to Jackson. Define your own annotations for this

modules/devkit-support/src/main/java/org/mule/security/oauth/processor/BaseOAuth2AuthorizeMessageProcessor.java
```diff
@@ -134,12 +136,16 @@ private void setState(Map<String, String> extraParameters, MuleEvent event)
...
+                    if(property!=null && StringUtils.isNotEmpty(property.value())){
```
marianogonzalez: apply correct code formatting

modules/devkit-support/src/test/java/org/mule/security/oauth/processor/OAuth2AuthorizeMessageProcessorTestCase.java
```diff
@@ -215,6 +222,9 @@ public String answer(InvocationOnMock invocation) throws Throwable
         private String customField;
         private String anotherCustomField;
 
+        @JsonProperty(annotatedPropertyName)
```
marianogonzalez: is devkit supposed to generate such a field on the MP if the connector is annotated somehow?

modules/devkit-support/src/main/java/org/mule/security/oauth/processor/BaseOAuth2AuthorizeMessageProcessor.java
marianogonzalez: I don't think this code belongs here. devkit-support module is annotations agnostic (actually, devkit annotations have compile retention and mule does not and will not depend on them).

The AuthorizationParameter are obtained by calling moduleObject.getDefaultUnauthorizedConnector().getAuthorizationParameters();

So, if what you want to specify aliases for them, then you should add code on the connector itself so that parameter.getName() returns the alias instead of the real name. This provides better compatibility between different connector and mule versions.

Notice that if you build a connector in which you use the alias functionality and deploy it on a mule version which doesn't have this improvement, then your alias would not work. If you do it on the connector, you'll have better portability.

In summary, I think that this is a devkit concern, not a mule one. If you agree, I'll close both this PR and the Jira issue

modules/devkit-support/src/main/java/org/mule/security/oauth/processor/BaseOAuth2AuthorizeMessageProcessor.java
```diff
@@ -118,15 +119,16 @@ private void setState(Map<String, String> extraParameters, MuleEvent event)
         {
             for (AuthorizationParameter<?> parameter : params)
             {
+                String parameterName = parameter.getName();
                 Field field = null;
                 try
                 {
-                    field = this.getClass().getDeclaredField(parameter.getName());
+                    field = this.getClass().getDeclaredField(parameterName);
```
pablocabrera85: This line prevents us from specifying an alias directly on DevKit. If the field I need to add on the map needs the name "Api-Key", java doesn't allow you to have an "Api-Key" field, since it is not a valid name. Get declared field will crash.

modules/devkit-support/src/main/java/org/mule/security/oauth/processor/BaseOAuth2AuthorizeMessageProcessor.java
pablocabrera85: We cannot use an alias on DevKit because this is using reflection to add the field from the class.

I'll add a new annotation to decouple if from Jackson. 

Using an annotation was one possible way of avoiding modifying org.mule.common.security.oauth.AuthorizationParameter class to add an alias/propertyName field

modules/devkit-support/src/test/java/org/mule/security/oauth/processor/OAuth2AuthorizeMessageProcessorTestCase.java
pablocabrera85: Yes, there will be some changes required on DevKit, a branch with this changes was already coded.

modules/devkit-support/src/main/java/org/mule/security/oauth/processor/BaseOAuth2AuthorizeMessageProcessor.java
marianogonzalez: What you're proposing still doesn't seem like the best solution to me.

AuthorizationParameter is a DTO used to bridge the Devkit way of representing things (which are mainly annotations) against Mule's. Additionally, If you define a new annotation then it belongs on Devkit's codebase not mule's. If it's on devkit's, then mule cannot access it because there're dependency issues which prevents Mule from knowing devkit's jars.

I would simply add a getAlias() method to AuthorizationParameter. So, when the field is fetch you use the name, but when you build the extra parameters map which is used to build the request you use the alias. In this way the design remains consistent, in the sense that AuthorizationParameter remains a DTO and the two worlds remain separated.

marianogonzalez:
Hello Pablo,

Any updates on this discussion? WIll you be sending other PR or should be close this one?

Thank you

pablokraan:
Closing this PR as there were no updates in a long time

PR #1695, opened 2015-05-06
MULE-8565: Add support for preemptive basic authentication in the HTTP module

andresgregoire:


modules/http/src/main/resources/META-INF/mule-httpn.xsd
```diff
@@ -774,6 +774,14 @@
                         </xsd:documentation>
                     </xsd:annotation>
                 </xsd:attribute>
+                <xsd:attribute name="preemptive" type="mule:substitutableBoolean" use="optional">
```
dfeist: Whats the default?

andresgregoire: It's false, to keep the current behavior. I will add it to the schema.

PR #1697, opened 2015-05-06
MULE-8216: Implement a validations extensions

asequeira:
Adding more tests


marianogonzalez:
invalid. you only picked and choose the comments you liked

extensions/validation/src/test/java/org/mule/extension/validation/BasicValidationTestCase.java
```diff
@@ -36,6 +38,7 @@
 
     private static final String CUSTOM_VALIDATOR_MESSAGE = "Do you wanna build a snowman?";
     private static final String EMAIL_VALIDATION_FLOW = "email";
+    public static final String NOT_EMPTY_FLOW = "notEmpty";
```
marianogonzalez: All these static constants are used by several tests. It doesn't make sense to put notEmpty here since it's meaningful to one and only one test.  Additionally, if we were to put it here, it should be private, not public

PR #1708, opened 2015-05-07
MULE-8575 - Set transaction timeout when creating a transaction objec…

pablolagreca:
…t - Using default tx timeout when creating delegate transaction

### DIFF
--- a/core/src/test/java/org/mule/processor/DelegateTransactionTestCase.java
+++ b/core/src/test/java/org/mule/processor/DelegateTransactionTestCase.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.processor;
+
+import static org.hamcrest.core.Is.is;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/mule/processor/DelegateTransactionTestCase.java
```diff
@@ -0,0 +1,43 @@
...
+import org.mule.tck.junit4.AbstractMuleTestCase;
+
+import org.junit.Test;
+import org.mockito.Mockito;
+
+
+public class DelegateTransactionTestCase extends AbstractMuleTestCase
+{
+
+    public static final int DEFAULT_TX_TIMEOUT = 2;
```
marianogonzalez: why public?

core/src/test/java/org/mule/processor/DelegateTransactionTestCase.java
```diff
@@ -0,0 +1,43 @@
...
+    private MuleContext mockMuleContext = Mockito.mock(MuleContext.class, RETURNS_DEEP_STUBS.get());
```
marianogonzalez: this should be initialised on a before() method. Each test should have a brand new mock instance. Additionally, use static imports

pablolagreca: Each test has a new brand instance. The test class is created per test method.

PR #1720, opened 2015-05-11
MULE-8553: Adding max-send-buffer-size entry in wrapper.conf

afelisatti:


distributions/standalone/src/main/resources/conf/wrapper.conf
```diff
@@ -22,6 +22,10 @@ wrapper.java.additional.5=-Dorg.glassfish.grizzly.nio.transport.TCPNIOTransport.
 # Increase this value if you get "Java.lang.OutOfMemoryError: PermGen space error"
 wrapper.java.additional.6=-XX:MaxPermSize=128m
 
+# Limit HTTP module send buffer size to 1MB by default to avoid running out of Direct Memory.  To optimize for
```
dfeist: Can we put this property alongside the other one and have them share the same comment?  This would be better wouldn't it?

PR #1737, opened 2015-05-15
MULE-8609: Lifecycle should not be applied on all prototype message p…

marianogonzalez:
…rocessors upon lookup

### DIFF
--- a/modules/spring-config/src/main/java/org/mule/config/spring/SpringRegistry.java
+++ b/modules/spring-config/src/main/java/org/mule/config/spring/SpringRegistry.java
@@ -14,14 +14,18 @@ import org.mule.api.MuleException;
 import org.mule.api.MuleRuntimeException;
 import org.mule.api.lifecycle.InitialisationException;
 import org.mule.api.lifecycle.LifecycleException;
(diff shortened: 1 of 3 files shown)


modules/spring-config/src/main/java/org/mule/config/spring/SpringRegistry.java
```diff
@@ -199,6 +203,24 @@ public Object lookupObject(String key)
...
     }
 
+    private boolean shouldApplyPrototypeLifecycle(String key, Object object)
+    {
+        if (applicationContext.isSingleton(key))
+        {
+            return false;
+        }
+
+        if (object instanceof MessageProcessor)
```
dfeist: Why these types and not others?  At minimum this needs explaining in a comment, but isn't the a bigger issue here?

marianogonzalez: The reason is explained on the issue but I'll add a comment. Basically, prototype types other than these end up being used in a dynamic flow/pipeline which applies lifecycle, resulting on the MP's LC being applied twice.

PR #1748, opened 2015-05-18
MULE-8609: Should be able to fetch prototype by name without lifecycl…

marianogonzalez:
…e applied


core/src/main/java/org/mule/api/registry/LifecycleRegistry.java
```diff
@@ -40,4 +40,22 @@
      */
     Object applyLifecycle(Object object, String phase) throws MuleException;
 
+    /**
+     * Look up a single object by name.
+     * <p/>
+     * Because {@link #lookupObject(String)} might
```
dfeist: Isn't it the case (given this comment is in LifecyceRegistry and not Registry) that with lookupObject the returned object will have had lifecycle applied?

PR #1757, opened 2015-05-19
Mule 8514

marcosnc:
FOR REVIEW ONLY - DO NOT MERGE

### DIFF
--- a/core/src/test/java/org/mule/tck/SensingNullRequestResponseMessageProcessor.java
+++ b/core/src/test/java/org/mule/tck/SensingNullRequestResponseMessageProcessor.java
@@ -6,12 +6,15 @@
  */
 package org.mule.tck;
 
-import org.mule.api.CompletionHandler;
-import org.mule.api.MessagingException;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.not;
+import static org.hamcrest.Matchers.sameInstance;
+import static org.junit.Assert.assertThat;
+import org.mule.api.MuleContext;
 import org.mule.api.MuleEvent;
 import org.mule.api.MuleException;
+import org.mule.api.registry.RegistrationException;
 import org.mule.processor.AbstractRequestResponseMessageProcessor;
-import org.mule.util.concurrent.Latch;
 
 /**
  * Can be used to sense request and response threads used during processing.
@@ -33,6 +36,23 @@ public class SensingNullRequestResponseMessageProcessor extends AbstractRequestR
     {
         responseThread = Thread.currentThread();
         return super.processRequest(event);
+    }
+
+    public static void assertSameThread(MuleContext muleContext, String beanName) throws Exception
+    {
+        SensingNullRequestResponseMessageProcessor instance = getInstance(muleContext, beanName);
+        assertThat(instance.requestThread, is(sameInstance(instance.responseThread)));
+    }
+
+    public static void assertDifferentThreads(MuleContext muleContext, String beanName) throws Exception
+    {
+        SensingNullRequestResponseMessageProcessor instance = getInstance(muleContext, beanName);
+        assertThat(instance.requestThread, not(sameInstance(instance.responseThread)));
+    }
+
+    private static SensingNullRequestResponseMessageProcessor getInstance(MuleContext muleContext, String beanName) throws RegistrationException
+    {
+        return muleContext.getRegistry().lookupObject(beanName);
     }
 
 }
--- a/modules/cxf/src/main/java/org/mule/module/cxf/CxfInboundMessageProcessor.java
+++ b/modules/cxf/src/main/java/org/mule/module/cxf/CxfInboundMessageProcessor.java
@@ -9,9 +9,12 @@ package org.mule.module.cxf;
 import static org.mule.module.cxf.HttpRequestPropertyManager.getBasePath;
 import static org.mule.module.cxf.HttpRequestPropertyManager.getRequestPath;
 import static org.mule.module.cxf.HttpRequestPropertyManager.getScheme;
+import org.mule.DefaultMuleEvent;
+import org.mule.NonBlockingVoidMuleEvent;
 import org.mule.VoidMuleEvent;
 import org.mule.api.DefaultMuleException;
 import org.mule.api.ExceptionPayload;
+import org.mule.api.MessagingException;
 import org.mule.api.MuleEvent;
 import org.mule.api.MuleException;
 import org.mule.api.MuleMessage;
@@ -20,11 +23,13 @@ import org.mule.api.lifecycle.InitialisationException;
 import org.mule.api.lifecycle.Lifecycle;
 import org.mule.api.transformer.TransformerException;
 import org.mule.api.transport.OutputHandler;
+import org.mule.api.transport.ReplyToHandler;
 import org.mule.config.i18n.MessageFactory;
 import org.mule.message.DefaultExceptionPayload;
 import org.mule.module.cxf.support.DelegatingOutputStream;
 import org.mule.module.xml.stax.StaxSource;
 import org.mule.processor.AbstractInterceptingMessageProcessor;
+import org.mule.processor.NonBlockingMessageProcessor;
 import org.mule.transformer.types.DataTypeFactory;
 import org.mule.transport.http.HttpConnector;
 import org.mule.transport.http.HttpConstants;
@@ -52,6 +57,7 @@ import org.apache.cxf.binding.soap.SoapBindingConstants;
 import org.apache.cxf.binding.soap.jms.interceptor.SoapJMSConstants;
 import org.apache.cxf.endpoint.Server;
 import org.apache.cxf.interceptor.StaxInEndingInterceptor;
+import org.apache.cxf.message.Exchange;
 import org.apache.cxf.message.ExchangeImpl;
 import org.apache.cxf.message.Message;
 import org.apache.cxf.message.MessageImpl;
@@ -71,7 +77,7 @@ import org.w3c.dom.Node;
  * built by a MessageProcessorBuilder which is responsible for configuring it and the
  * {@link Server} that it dispatches to.
  */
-public class CxfInboundMessageProcessor extends AbstractInterceptingMessageProcessor implements Lifecycle
+public class CxfInboundMessageProcessor extends AbstractInterceptingMessageProcessor implements Lifecycle, NonBlockingMessageProcessor
 {
 
     public static final String JMS_TRANSPORT = "jms";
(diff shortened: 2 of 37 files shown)


core/src/test/java/org/mule/tck/SensingNullRequestResponseMessageProcessor.java
```diff
@@ -35,4 +38,21 @@ protected MuleEvent processResponse(MuleEvent event) throws MuleException
         return super.processRequest(event);
     }
 
+    public static void assertSameThread(MuleContext muleContext, String beanName) throws Exception
```
dfeist: These methods don't belong here.  It's a good idea to have an assertResponeRequestThreadsDifferent/assertResponeRequestThreadsSame here, but it' should take no parameters and rather do lookup etc.

modules/cxf/src/main/java/org/mule/module/cxf/CxfInboundMessageProcessor.java
```diff
@@ -229,150 +235,192 @@ protected MuleEvent sendToDestination(MuleEvent event) throws MuleException, IOE
...
+                        originalReplyToHandler.processExceptionReplyTo(exception, replyTo);
+                    }
+                });
             }
 
-            String path = muleReqMsg.getInboundProperty(HttpConnector.HTTP_REQUEST_PATH_PROPERTY);
-            if (path == null)
+            MuleEvent responseEvent = sendThroughCxf(event, exchange);
+
+            if (responseEvent==null || !responseEvent.equals(NonBlockingVoidMuleEvent.getInstance()))
```
dfeist: Formatting.

modules/cxf/src/main/java/org/mule/module/cxf/CxfInboundMessageProcessor.java
```diff
@@ -229,150 +235,192 @@ protected MuleEvent sendToDestination(MuleEvent event) throws MuleException, IOE
...
+                return processResponse(event, exchange, responseEvent);
             }
 
-            if (method != null)
-            {
-                m.put(Message.HTTP_REQUEST_METHOD, method);
-                m.put(Message.PATH_INFO, path);
-                String basePath = getBasePath(muleReqMsg);
-                m.put(Message.BASE_PATH, basePath);
+            return NonBlockingVoidMuleEvent.getInstance();
```
dfeist: I think you can just return "responseEvent" here can't you?

modules/cxf/src/main/java/org/mule/module/cxf/CxfInboundMessageProcessor.java
```diff
@@ -229,150 +235,192 @@ protected MuleEvent sendToDestination(MuleEvent event) throws MuleException, IOE
...
+                final ReplyToHandler originalReplyToHandler = event.getReplyToHandler();
+
+                event = new DefaultMuleEvent(event, new ReplyToHandler()
+                {
+                    @Override
+                    public void processReplyTo(MuleEvent event, MuleMessage returnMessage, Object replyTo) throws MuleException
+                    {
+                        try
+                        {
+                            originalReplyToHandler.processReplyTo(event, returnMessage, replyTo);
```
dfeist: Where are you creating OutputHandler in non-blocking case?

modules/cxf/src/main/java/org/mule/module/cxf/CxfInboundMessageProcessor.java
```diff
@@ -229,150 +235,192 @@ protected MuleEvent sendToDestination(MuleEvent event) throws MuleException, IOE
...
+                        }
+                        catch (Exception e)
+                        {
+                            processExceptionReplyTo(new MessagingException(event, e), replyTo);
```
dfeist: \What does exception payload get created in non-blocking case? Also where is 500 status code set?

modules/cxf/src/main/java/org/mule/module/cxf/CxfOutboundMessageProcessor.java
```diff
@@ -52,7 +55,7 @@
  * The CxfOutboundMessageProcessor performs outbound CXF processing, sending an event
  * through the CXF client, then on to the next MessageProcessor.
  */
-public class CxfOutboundMessageProcessor extends AbstractInterceptingMessageProcessor implements CloneableMessageProcessor
+public class CxfOutboundMessageProcessor extends AbstractInterceptingMessageProcessor implements CloneableMessageProcessor, NonBlockingSupported
```
dfeist: Inbound implements NonBlockingMP, but outbound implements NonBlockingSupported, this is somewhat inconsistent.

modules/cxf/src/main/java/org/mule/module/cxf/CxfOutboundMessageProcessor.java
```diff
@@ -235,9 +250,37 @@ protected MuleEvent doSendWithClient(MuleEvent event) throws Exception
...
+                @Override
+                public void handleResponse(Map<String, Object> ctx, Object[] res)
+                {
+                    try
+                    {
+                        event.getReplyToHandler().processReplyTo(buildResponseMessage(event, responseHolder.value, res), null, null);
+                    }
+                    catch (MuleException ex)
+                    {
+                        event.getReplyToHandler().processExceptionReplyTo(wrapException(responseHolder.value, ex), null);
```
dfeist: Invoke handleException() here instead of repeating wrapException?

modules/cxf/src/main/java/org/mule/module/cxf/MuleInvoker.java
```diff
@@ -125,11 +126,17 @@ else if(e instanceof ComponentException) {
             responseEvent = null;
         }
 
+        if( responseEvent instanceof NonBlockingVoidMuleEvent)
```
dfeist: Formatting

modules/cxf/src/main/java/org/mule/module/cxf/MuleInvoker.java
```diff
@@ -125,11 +126,17 @@ else if(e instanceof ComponentException) {
...
+        {
+            exchange.put(CxfConstants.MULE_EVENT, responseEvent);
+            return new Object();
```
dfeist: Why new Object() and not null or something else? Add comment?

modules/cxf/src/test/java/org/mule/module/cxf/CxfBasicNonBlockingTestCase.java
```diff
@@ -0,0 +1,91 @@
...
+            XMLUnit.getTransformerFactory();
+        }
+        catch (TransformerFactoryConfigurationError e)
+        {
+            XMLUnit.setTransformerFactory(XMLUtils.TRANSFORMER_FACTORY_JDK5);
+        }
+    }
+
+    @Test
+    public void testEchoService() throws Exception
```
dfeist: Not clear what you are testing/asserting here because i) there isn't a non-blocking component in the flow ii) you are asserting threads are same and not different.

Also, is it possible to paramatize an existing CXF test case?

modules/cxf/src/test/resources/holder-conf-httpn-nb.xml
```diff
@@ -0,0 +1,82 @@
...
+
+    <spring:bean name="sensingRequestResponseProcessorEcho" class="org.mule.tck.SensingNullRequestResponseMessageProcessor" scope="singleton" />
+    <flow name="echoServiceClient" processingStrategy="non-blocking">
+        <http:listener path="/echoClient" config-ref="listenerConfig" allowedMethods="POST" />
+        <transformer ref="holderTransformer"/>
+        <processor ref="sensingRequestResponseProcessorEcho" />
+        <cxf:jaxws-client serviceClass="org.mule.module.cxf.HolderService" operation="echo"/>
+        <http:request path="echo" method="POST" config-ref="requestConfig"/>
+    </flow>
+    <spring:bean name="sensingRequestResponseProcessorEchoProxy" class="org.mule.tck.SensingNullRequestResponseMessageProcessor" scope="singleton" />
```
dfeist: Spaces between flows for readability.

modules/cxf/src/test/resources/proxy-conf-flow-httpn-nb.xml
```diff
@@ -0,0 +1,60 @@
...
+    <!-- Echo Service Start -->
+    <flow name="echo-service">
+        <http:listener path="/services/echo" config-ref="listenerConfig" allowedMethods="POST" />
+        <cxf:proxy-service />
+        <vm:outbound-endpoint path="echo-service-vm" exchange-pattern="request-response" />
+    </flow>
+    <flow name="echo-service-vm">
+        <inbound-endpoint address="vm://echo-service-vm" exchange-pattern="request-response" />
+        <logger level="WARN" message="********** -> ECHO IMPLEMENTATION"/>
+        <!--<logger level="WARN" message="********** -> ECHO IMPLEMENTATION #[java.lang.Thread.sleep(5000); 'OK']"/>-->
```
dfeist: Remove commented out line?

modules/cxf/src/main/java/org/mule/module/cxf/CxfInboundMessageProcessor.java
marcosnc: the first part could be removed by using:
if (!NonBlockingVoidMuleEvent.getInstance().equals(responseEvent))
but I think that making explicit the null comparison is better here.

modules/cxf/src/main/java/org/mule/module/cxf/CxfInboundMessageProcessor.java
marcosnc: Yes, but here I also think that this makes more explicit what's happening.

modules/cxf/src/main/java/org/mule/module/cxf/CxfInboundMessageProcessor.java
marcosnc: It's not being created. The output handler expects a DelegatingOutputStream to be the content of the CXF output message received in the exchange. In the NB case, we already have the returning mule event, which have no output stream set.

modules/cxf/src/main/java/org/mule/module/cxf/CxfInboundMessageProcessor.java
marcosnc: You're right.

modules/cxf/src/main/java/org/mule/module/cxf/CxfOutboundMessageProcessor.java
marcosnc: Yes, will change that in CxfOutboundMP.

modules/cxf/src/main/java/org/mule/module/cxf/CxfOutboundMessageProcessor.java
marcosnc: Perfect!

modules/cxf/src/main/java/org/mule/module/cxf/MuleInvoker.java
marcosnc: Yes, just something different from null is required here as output value, due to the important value is stored in the exchange.

modules/cxf/src/test/java/org/mule/module/cxf/CxfBasicNonBlockingTestCase.java
marcosnc: In fact I was checking that using a blocking component inside a non blocking flow will fall back to blocking processing. Will change the test name to reflect this.

PR #1766, opened 2015-05-21
MULE-8602: Make ObservableList having a size >= 18 work with Collections#reverse

yyamano:


modules/launcher/src/test/java/org/mule/module/launcher/util/ObservableListTestCase.java
```diff
@@ -0,0 +1,46 @@
...
+public class ObservableListTestCase
+{
+    @Test
+    public void testReverseObservableListWith18Elements()
+    {
+        // See java.util.Collections.REVERSE_THRESHOLD and Collections#reverse() for the magic number
+        ObservableList<Integer> list = createObservableList(18);
+        Collections.reverse(list);
+
+        assertEquals((Integer)17, list.get(0));
```
marianogonzalez: please use hamcrest to perform the assertions

modules/launcher/src/test/java/org/mule/module/launcher/util/ObservableListTestCase.java
```diff
@@ -0,0 +1,46 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.module.launcher.util;
+
+import static org.junit.Assert.assertEquals;
+import org.junit.Test;
+
+import java.util.Collections;
+
+public class ObservableListTestCase
```
marianogonzalez: this class should extends AbstractMuleTestCase

modules/launcher/src/main/java/org/mule/module/launcher/util/ObservableList.java
```diff
@@ -303,6 +303,7 @@ public boolean add(E o)
 
         private Iterator<E> iterDelegate;
         protected int cursor = 0;
+	protected int lastRet = -1;
```
marianogonzalez: please fix code formatting

modules/launcher/src/main/java/org/mule/module/launcher/util/ObservableList.java
```diff
@@ -323,14 +324,16 @@ public boolean hasNext()
...
+            E next = iterDelegate.next();
+            lastRet = cursor++;
+            return next;
         }
 
         @Override
         public void remove()
         {
             ObservableList.this.remove(cursor--);
+	    lastRet = -1;
```
marianogonzalez: please fix code formatting

modules/launcher/src/test/java/org/mule/module/launcher/util/ObservableListTestCase.java
```diff
@@ -0,0 +1,46 @@
...
+        Collections.reverse(list);
+
+        assertEquals((Integer)16, list.get(0));
+        assertEquals((Integer)0, list.get(16));
+    }
+
+    private ObservableList<Integer> createObservableList(int n)
+    {
+        ObservableList<Integer> list = new ObservableList<Integer>();
+        for (int i = 0; i < n; i++) {
```
marianogonzalez: please fix code formatting

modules/launcher/src/test/java/org/mule/module/launcher/util/ObservableListTestCase.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    @Test
+    public void testReverseObservableListWith17Elements()
+    {
+        // See java.util.Collections.REVERSE_THRESHOLD and Collections#reverse() for the magic number
+        ObservableList<Integer> list = createObservableList(17);
+        Collections.reverse(list);
+
+        assertEquals((Integer)16, list.get(0));
```
marianogonzalez: again, please use hamcrest

modules/launcher/src/main/java/org/mule/module/launcher/util/ObservableList.java
marianogonzalez: please replace all occurences of -1 with a constant. That constant should be private and final

modules/launcher/src/test/java/org/mule/module/launcher/util/ObservableListTestCase.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.Collections;
+
+public class ObservableListTestCase
+{
+    @Test
+    public void testReverseObservableListWith18Elements()
+    {
+        // See java.util.Collections.REVERSE_THRESHOLD and Collections#reverse() for the magic number
+        ObservableList<Integer> list = createObservableList(18);
```
marianogonzalez: please replace 18 with a constant

modules/launcher/src/test/java/org/mule/module/launcher/util/ObservableListTestCase.java
marianogonzalez: please replace 17 with the constant defined above minus 1

modules/launcher/src/test/java/org/mule/module/launcher/util/ObservableListTestCase.java
marianogonzalez: use constants here as well

modules/launcher/src/main/java/org/mule/module/launcher/util/ObservableList.java
```diff
@@ -353,6 +356,7 @@ public void add(E o)
         {
             ObservableList.this.add(o);
             cursor++;
+            lastRet = -1;
```
marianogonzalez: why setting this to -1 again if this method doesn't return anything?

modules/launcher/src/main/java/org/mule/module/launcher/util/ObservableList.java
```diff
@@ -323,14 +324,16 @@ public boolean hasNext()
         @Override
         public E next()
         {
-            cursor++;
-            return iterDelegate.next();
+            E next = iterDelegate.next();
+            lastRet = cursor++;
```
marianogonzalez: could you elaborate a little bit on what was the root cause of the issue and why you consider that this lastRet is the best solution for the issue?

marianogonzalez:
Hello Yuji,

Thank you for this contribution. I added some comments requesting some changes and also requesting for some clarification on the solution.

Once the changes are applied, please squash the commits and let us know to review it again.

Thanks!

modules/launcher/src/main/java/org/mule/module/launcher/util/ObservableList.java
yyamano: to follow the spec at http://docs.oracle.com/javase/7/docs/api/java/util/ListIterator.html#set(E)

```
void set(E e)

Replaces the last element returned by next() or previous() with the specified element 
(optional operation). This call can be made only if neither remove() nor add(E) have 
been called after the last call to next or previous.
```

modules/launcher/src/main/java/org/mule/module/launcher/util/ObservableList.java
yyamano: The original implementation of set() is	broken. For example, the following simple test code throws IndexOutOfBoundsException. I will add it to ObservableListTestCase.

```
@Test
public void testListIteratorSetAfterNext()
{
    ObservableList<String> list = new ObservableList<String>();                                                                                               
    list.add("a");                                                                                                                                            
    ListIterator<String> it = list.listIterator();                                                                                                            
    String s = it.next();                                                                                                                                     
    it.set("A");                                                                                                                                              

    assertThat("a", equalTo(s));                                                                                                                              
    assertThat("A", equalTo(list.get(0)));                                                                                                                    
}
```

I introduced lastRet to make set() work with both next() and previous().
java.util.Collections#reverse() uses both as below:

```
ListIterator fwd = list.listIterator();                                                                                                                       
ListIterator rev = list.listIterator(size);                                                                                                                   
for (int i=0, mid=list.size()>>1; i<mid; i++) {                                                                                                               
    Object tmp = fwd.next();                                                                                                                                  
    fwd.set(rev.previous());                                                                                                                                  
    rev.set(tmp);                                                                                                                                             
}
```

yyamano:
@marianogonzalez Thank you for the review. 

I just push the new version. Also please see my comments above.

modules/launcher/src/main/java/org/mule/module/launcher/util/ObservableList.java
```diff
@@ -301,8 +301,11 @@ public boolean add(E o)
     private class ObservableIterator implements Iterator<E>
     {
 
+        protected static final int LAST_RET_NONE = -1;
```
marianogonzalez: looks like this should be private

modules/launcher/src/main/java/org/mule/module/launcher/util/ObservableList.java
```diff
@@ -301,8 +301,11 @@ public boolean add(E o)
...
+
```
marianogonzalez: please remove this empty line

modules/launcher/src/main/java/org/mule/module/launcher/util/ObservableList.java
```diff
@@ -301,8 +301,11 @@ public boolean add(E o)
...
         private Iterator<E> iterDelegate;
         protected int cursor = 0;
+        protected int lastRet = LAST_RET_NONE;
```
marianogonzalez: why protected? Looks like it could be private

modules/launcher/src/main/java/org/mule/module/launcher/util/ObservableList.java
```diff
@@ -382,7 +390,11 @@ public int previousIndex()
         @Override
         public void set(E e)
         {
-            ObservableList.this.set(cursor, e);
+            if (lastRet < 0)
+            {
+                throw new IllegalStateException();
```
marianogonzalez: please add a meaningful message to this exception

modules/launcher/src/test/java/org/mule/module/launcher/util/ObservableListTestCase.java
marianogonzalez: code formatting is still missing in this block

marianogonzalez:
Hello @yyamano,

Thanks for applying the changes. I added a few more comments and hopefully this can be merged afterwards. We're hitting code freeze for Mule 3.7.0 on June 2nd so if you could have these changes by then, we can include them in the next release.

Thank you

yyamano:
@marianogonzalez The changes include

* Fix a bug in ObservableIterator#remove() and ObservableListIterator#add().  
  The delegate's iterators, actual classes are ArrayList.{Itr,ListItr}, have an internal state called "modCount" to avoid concurrent modification.  
  We need to update ObservableList via the delegate. To know the exact problem, you can run testListIteratorRemove() that I added with other tests.
* Add a message to IllegalStateException.
* Replace -1 in ObservableListIterator#add() with LAST_RET_NONE.   
  Thus, the constant has to be protected.
* Remove a blank line.
* Fix code formatting.

marianogonzalez:
Thank you for the contribution @yyamano

PR #1774, opened 2015-05-22
MULE-8580: Fixed HttpSharePortTestCase.

marcosnc:


marianogonzalez:
why does this fixes the problem? Old Http transport has issues with JDK8?

marcosnc:
It seems that.. I'm trying if this works on cloudbees and then see that.

PR #1785, opened 2015-05-26
MULE-8645: Remove jasper-jdt-6.0.29.jar from Mule distributions

pablokraan:


distributions/standalone/src/main/resources/MIGRATION.txt
```diff
@@ -125,6 +125,7 @@ MULE-8569: For those with custom implementation of class org.mule.config.spring.
     * The method createApplicationContext(MuleContext, ConfigResource[]) is now private. If you want to overwrite it, use doCreateApplicationContext(MuleContext, ConfigResource[], OptionalObjectsController) instead.
     * If you want to intercept and change the list of resources to be loaded, override the new addResources(List<ConfigResource>) method
 MULE-8639: Grizzly was upgraded to 2.3.21.
+MULE-8645: jasper-jdt-6.0.29 is not included anymore on Mule distributions because of some detected vulnerabilities. In case this artifact is needed, when using drools for example, then manually add it on <MULE_HOME>/lib/opt
```
marianogonzalez: wouldn't this also work if added to the app? If so, we should probably say so since that's the easiest way for the raw Studio user

PR #1792, opened 2015-05-27
MULE-8649: Resolve data type from simple MEL expressions

pablokraan:


core/src/main/java/org/mule/el/mvel/MVELExpressionExecutor.java
```diff
@@ -39,9 +39,9 @@
 
     protected static final int COMPILED_EXPRESSION_MAX_CACHE_SIZE = 1000;
 
-    protected ParserConfiguration parserConfiguration;
+    public ParserConfiguration parserConfiguration;
 
-    protected LoadingCache<String, Serializable> compiledExpressionsCache;
+    public LoadingCache<String, Serializable> compiledExpressionsCache;
```
marianogonzalez: why public?

core/src/main/java/org/mule/api/el/ExpressionLanguage.java
```diff
@@ -109,4 +110,5 @@
     @Deprecated
     <T> T evaluate(String expression, MuleMessage message, Map<String, Object> vars);
 
+    TypedValue evaluateTyped(String expression, MuleMessage message);
```
marianogonzalez: I think this should receive the event, not the message. I know that might lead to inconsistency with other similar classes and that it might even be impossible due to the contract that collaborators of this class might have. But I'd like to raise that point anyhow because MEL currently has issues because of this duality. I've seen some functions failed because the _muleEvent context variable not always exists.

gdssrao: I am not sure but want to add some point here with my experience. When ever i use Z_ or C_ in expression then the evaluation not happening as expected is this relevant here? If not please Ignore

gdssrao:
I am not sure but want to add some point here with my experience. When ever i use Z_ or C_ in expression then the evaluation not happening as expected is this relevant here? If not please Ignore

core/src/main/java/org/mule/el/mvel/MVELExpressionExecutor.java
pablokraan: I forgot to revert that line after hacking :)

core/src/main/java/org/mule/api/el/ExpressionLanguage.java
pablokraan: The issue here is that this logic is needed on some transformers, and they work on the message, not the event.

PR #1797, opened 2015-05-28
MULE-8612: Fixed watermark MIN and MAX selectors when compareTo returns values that are not 1, 0 or -1.

jsaliba:


core/src/test/java/org/mule/transport/polling/watermark/WatermarkSelectorTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+import org.mule.transport.polling.watermark.selector.MaxValueWatermarkSelector;
+import org.mule.transport.polling.watermark.selector.MinValueWatermarkSelector;
+
+import java.text.SimpleDateFormat;
+import java.util.Calendar;
+
+import org.junit.Test;
+
+public class WatermarkSelectorTest
```
pablokraan: Classes must be named with "TestCase"suffix and must extend from AbstractMuleTestCase

PR #1807, opened 2015-05-29
MULE-8654: Setting muleContext to internal transformers.

afelisatti:


core/src/test/java/org/mule/processor/IdempotentRedeliveryPolicyTestCase.java
```diff
@@ -90,6 +86,7 @@ public MuleEvent answer(InvocationOnMock invocationOnMock) throws Throwable
         muleLockFactory.initialise();
         when(mockMuleContext.getLockFactory()).thenReturn(muleLockFactory);
         when(mockMuleContext.getRegistry().get(MuleProperties.OBJECT_STORE_MANAGER)).thenReturn(mockObjectStoreManager);
+        when(mockMuleContext.getConfiguration().getDefaultEncoding()).thenReturn(UTF_8);
```
marianogonzalez: how is this change related to the issue?

core/src/main/java/org/mule/processor/IdempotentRedeliveryPolicy.java
```diff
@@ -102,6 +103,8 @@ public void initialise() throws InitialisationException
         idrId = String.format("%s-%s-%s",appName,flowName,"idr");
         lockFactory = muleContext.getLockFactory();
         store = createStore();
+        LifecycleUtils.initialiseIfNeeded(objectToByteArray, muleContext);
```
marianogonzalez: static imports

transports/ftp/src/test/java/org/mule/transport/ftp/reliability/FtpRedeliveryPolicyTestCase.java
```diff
@@ -0,0 +1,50 @@
...
+    public static Collection<Object[]> parameters()
+    {
+        return Arrays.asList(new Object[][] {
+                {ConfigVariant.FLOW, "reliability/ftp-redelivery-policy-config.xml"}
+        });
+    }
+    @Test
+    public void testRedeliveryPolicyDLQConsumesMessage() throws Exception
+    {
+        ExceptionListener exceptionListener = new ExceptionListener(muleContext).setNumberOfExecutionsRequired(3);
```
marianogonzalez: aawwweeessooomeeeeee.... didn't know we had this class :)

marianogonzalez: I'd be great if instead of hardcoding a 3 here and a 2 in the flow, you could define a MAX_REDELIVERY_ATTEMPTS constant and expose it as a system property. So, in here you would use MAX_REDELIVERY_ATTEMPS +1 and in the flow ${MAX_REDELIVERY_ATTEMPS}. In this way, you don't spread magic numbers across both the test class and the configuration.

Something like this is implemented in AbstractSftpRedeliveryTestCase

core/src/test/java/org/mule/processor/IdempotentRedeliveryPolicyTestCase.java
afelisatti: When the transform method is executed, the encoding is determined. Since muleContext was null before, it was determined by the mule.encoding system property instead (see SystemUtils getDefaultEncoding method).

transports/ftp/src/test/java/org/mule/transport/ftp/reliability/FtpRedeliveryPolicyTestCase.java
afelisatti: Nice. Thanks.

core/src/test/java/org/mule/processor/IdempotentRedeliveryPolicyTestCase.java
marianogonzalez: perfect. I didn't see that one coming. Thanks

PR #1815, opened 2015-06-01
MULE-8630 Add further non-blocking functional test case for foreach, …

dfeist:
…choice, wire-tap, enricher etc.


core/src/main/java/org/mule/enricher/MessageEnricher.java
```diff
@@ -63,9 +67,13 @@ public MuleEvent process(MuleEvent event) throws MuleException
...
         ExpressionManager expressionManager = event.getMuleContext().getExpressionManager();
 
-        MuleEvent enricherEvent;
         //TODO: change DefaultMuleEvent.copy to DefaultMuleEvent.copyPreservingSession
-        enricherEvent = DefaultMuleEvent.copy(event);
+        MuleEvent enricherEvent = DefaultMuleEvent.copy(event);
+        if (event.isAllowNonBlocking() && event.getReplyToHandler() != null)
+        {
+            // TODO: MULE-8662 Support non-blocking execution within message enricher
+            enricherEvent = new DefaultMuleEvent(event.getMessage(), event, true);
```
marcosnc: Two copies are created here, could we perform only one copy instead?

core/src/main/java/org/mule/routing/WireTap.java
```diff
@@ -54,6 +55,8 @@ public MuleEvent process(MuleEvent event) throws MuleException
         try
         {
             MuleEvent tapEvent = DefaultMuleEvent.copy(event);
+            // Tap should not respond to reply to handler
+            tapEvent = new DefaultMuleEvent(tapEvent, (ReplyToHandler) null);
```
marcosnc: Two copies are created here, could we perform only one copy instead?

core/src/main/java/org/mule/enricher/MessageEnricher.java
dfeist: The thing is that DefaultMuleEvent.copy(event); has a fair amount of logic that i don't want to skip or duplicate.

tests/integration/src/test/java/org/mule/NonBlockingFunctionalTestCase.java
```diff
@@ -0,0 +1,204 @@
...
+import org.mule.tck.functional.FunctionalTestComponent;
+import org.mule.tck.junit4.FunctionalTestCase;
+import org.mule.util.concurrent.Latch;
+
+import java.util.concurrent.TimeUnit;
+
+import org.junit.Test;
+
+public class NonBlockingFunctionalTestCase extends FunctionalTestCase
+{
```
marcosnc: The problem with this kind of tests is that you're loading the entire config file (with several forms) in each test, but each test only uses just one flow. This will impact in the build time. How about to separate this case into several cases using a smaller config for each one?

tests/integration/src/test/java/org/mule/test/config/SecurityFilterNonBlockingTestCase.java
```diff
@@ -14,7 +14,7 @@
 /**
  * Test configuration of security filters
  */
-public class SecurityFilterNonBlockingTestCase extends FunctionalTestCase
+    public class SecurityFilterNonBlockingTestCase extends FunctionalTestCase
```
marcosnc: Tab added?

core/src/main/java/org/mule/enricher/MessageEnricher.java
marcosnc: Then add a TODO comment to fix this when we have a builder for MuleEvent. I think the could be avoided now by adding an overload to DefaultMuleEvent.copy but is an ugly solution...

tests/integration/src/test/java/org/mule/test/tck/AssertionMessageProcessorTestCase.java
```diff
@@ -104,13 +113,13 @@ public void processZeroEvents() throws Exception
...
+        AssertionMessageProcessor asp = createAssertionMessageProcessor();
         asp.setFlowConstruct(flowConstruct);
         asp.setExpression(TRUE_EXPRESSION);
         asp.start();
-        asp.process(mock(MuleEvent.class));
-        asp.process(mock(MuleEvent.class));
-        asp.process(mock(MuleEvent.class));
+        asp.process(mockEvent);
+        asp.process(mockEvent);
+        asp.process(mockEvent);
```
marcosnc: Here three different events where processed before your change, now the same event is processed three times, is it ok?

dfeist: Yes, I can't see how this affects the test method, can you?

PR #1816, opened 2015-06-01
MULE-8658: XML header is not removed when proxying, using non blockin…

marcosnc:
…g cxf:proxy-service and cxf:proxy-client.

Inbound interceptors were not correctly applied on proxy response.

### DIFF
--- a/modules/cxf/src/main/java/org/mule/module/cxf/CxfInboundMessageProcessor.java
+++ b/modules/cxf/src/main/java/org/mule/module/cxf/CxfInboundMessageProcessor.java
@@ -56,6 +56,7 @@ import org.apache.commons.logging.LogFactory;
 import org.apache.cxf.Bus;
 import org.apache.cxf.binding.soap.SoapBindingConstants;
 import org.apache.cxf.binding.soap.jms.interceptor.SoapJMSConstants;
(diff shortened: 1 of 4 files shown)


modules/cxf/src/main/java/org/mule/module/cxf/CxfInboundMessageProcessor.java
```diff
@@ -422,12 +405,51 @@ private MuleEvent sendThroughCxf(MuleEvent event, Exchange exchange) throws Tran
...
-        d.getMessageObserver().onMessage(m);
+        try
+        {
+            d.getMessageObserver().onMessage(m);
+        }
+        catch (SuspendedInvocationException e)
+        {
+            MuleEvent responseEvent = (MuleEvent) exchange.get(CxfConstants.MULE_EVENT);
+
+            if (responseEvent == null || !responseEvent.equals(NonBlockingVoidMuleEvent.getInstance()))
```
dfeist: Why is exception swallowed if event is NonBlockingVoidMuleEvent?  This doesn't happen per-request does it? And if it doesn't happen per request, surely it's an error case and the exception needs handling, logging or propagating?

PR #1826, opened 2015-06-03
MULE-8668 MessageProcessorContainers should not by default support no…

dfeist:
…n-blocking

### DIFF
--- a/core/src/main/java/org/mule/processor/NonBlockingProcessorExecutor.java
+++ b/core/src/main/java/org/mule/processor/NonBlockingProcessorExecutor.java
@@ -17,6 +17,7 @@ import org.mule.api.NonBlockingSupported;
 import org.mule.api.processor.InterceptingMessageProcessor;
 import org.mule.api.processor.MessageProcessor;
 import org.mule.api.processor.MessageProcessorChain;
+import org.mule.api.processor.MessageProcessorContainer;
 import org.mule.api.processor.MessageRouter;
 import org.mule.api.transport.ReplyToHandler;
 import org.mule.execution.MessageProcessorExecutionTemplate;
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/mule/processor/NonBlockingProcessorExecutor.java
```diff
@@ -51,8 +52,8 @@ protected void preProcess(MessageProcessor processor)
     {
         if (event.isAllowNonBlocking() && replyToHandler != null)
         {
-            if ((processor instanceof MessageRouter || processor instanceof InterceptingMessageProcessor) && !
-                    (processor instanceof NonBlockingSupported))
+            if ((processor instanceof MessageProcessorContainer || processor instanceof InterceptingMessageProcessor)
+                && !(processor instanceof NonBlockingSupported))
```
marcosnc: I think a function here would be nice :)

core/src/test/java/org/mule/processor/chain/NonBlockingProcessorExecutorTestCase.java
```diff
@@ -53,18 +58,68 @@ public void executeRequestResponseNonBlocking() throws MuleException, Interrupte
...
+        processors.add(processor3);
+        assertBlockingExecution(processors);
+    }
+
+    @Test
+    public void executeRequestResponseWithMPContainerBlocking() throws MuleException, InterruptedException
+    {
+        when(event.getExchangePattern()).thenReturn(MessageExchangePattern.REQUEST_RESPONSE);
+        when(event.isSynchronous()).thenReturn(false);
+        when(event.isAllowNonBlocking()).thenReturn(true);
```
marcosnc: Use a configuring function to reuse this code.

core/src/main/java/org/mule/processor/NonBlockingProcessorExecutor.java
```diff
@@ -51,8 +52,7 @@ protected void preProcess(MessageProcessor processor)
     {
         if (event.isAllowNonBlocking() && replyToHandler != null)
         {
-            if ((processor instanceof MessageRouter || processor instanceof InterceptingMessageProcessor) && !
-                    (processor instanceof NonBlockingSupported))
+            if (processorSupportsNonBlocking(processor))
```
marcosnc: you should negate this function, right?

PR #1830, opened 2015-06-03
MULE-8660 - Commit failure does not trigger exception strategy

pablolagreca:


distributions/standalone/src/main/resources/MIGRATION.txt
```diff
@@ -131,6 +131,7 @@ MULE-8628: The HTTP connector will now ignore it's own custom properties (http.*
     * properties generated by a listener won't affect a subsequent request
     * properties generated by that request won't affect the listener response
     * if such properties are desired, they should be explicitly added as headers using a response/request builder
+MULE-8660: AbstractMessageReceiver.routeMessage(..) no longer return nulls if the endpoint exchange pattern is one-way. It will always return the result of the flow so if a transaction commit fails the exception strategy is executed using the message result of the flow execution. Custom message receivers implementations may need to be updated.
```
pablokraan: A better way to reference the method would be AbstractMessageReceiver#routeMessage(Object) as in javadoc

PR #1838, opened 2015-06-04
MULE-8679, fix the way we create the path

damiansima:


core/src/test/java/org/mule/util/TestsLogConfigurationHelper.java
```diff
@@ -63,7 +65,17 @@ public static void clearLoggingConfig()
 
     private static String findLogConfigurationPath(Class<?> testClass)
     {
-        File folder = new File(testClass.getClassLoader().getResource("").getPath().toString());
+        String encodedFolder = testClass.getClassLoader().getResource("").getPath().toString();
+
+        String folderPath ;
+        try {
```
marianogonzalez: wrong code formatting

core/src/test/java/org/mule/util/TestsLogConfigurationHelper.java
```diff
@@ -63,7 +65,17 @@ public static void clearLoggingConfig()
...
+            folderPath = URLDecoder.decode(encodedFolder, "UTF-8");
+        } catch (UnsupportedEncodingException e) {
+            folderPath = encodedFolder;
```
marianogonzalez: this is wrong... is you can't decode it you can't just return a path which might potentially have wrong characters. If this case you should return NULL_CONFIG_FILE

core/src/test/java/org/mule/util/TestsLogConfigurationHelper.java
```diff
@@ -63,7 +65,17 @@ public static void clearLoggingConfig()
 
     private static String findLogConfigurationPath(Class<?> testClass)
     {
-        File folder = new File(testClass.getClassLoader().getResource("").getPath().toString());
+        String encodedFolder = testClass.getClassLoader().getResource("").getPath().toString();
+
+        String folderPath ;
+        try {
+            folderPath = URLDecoder.decode(encodedFolder, "UTF-8");
```
marianogonzalez: why did you hardcoded UTF-8? Shouldn't you use the system encoding?

PR #1847, opened 2015-06-05
MULE-8638: [Regression] NPE when filter fails.

marcosnc:


core/src/main/java/org/mule/execution/AsyncResponseFlowProcessingPhase.java
```diff
@@ -98,6 +98,18 @@ private void fireNotification(MuleEvent event, int action)
     {
         try
         {
+            if (event == null)
```
dfeist: Why does it have to be the last event? I think with the transport in this scenario it used the request event (this may have been incorrect, but it's worth comparing with transport).

marcosnc: We are trying to retrieve the last available event here, because filters set nullify the event and the notification needs an event to being fired.

PR #1851, opened 2015-06-08
MULE-8626: Connection and Keep-Alive message properties should not affect Listener/Requestor connection reuse behaviour.

afelisatti:


dfeist:
We need to revisit this, even though this is the behaviour we discussed it conflicts directly with the approach take when a "Transfer-Encoding: chunked" header is set as a header and then ignored if responeStreamingMode="never".  In one case the header wins and in the other case the connector config wins, this is confusing and inconsistent/

PR #1874, opened 2015-06-15
MULE-8676: HTTP listener should ignore 'Transfer-Encoding' property a…

afelisatti:
…s it is a hop-by-hop header

### DIFF
--- a/distributions/standalone/src/main/resources/MIGRATION.txt
+++ b/distributions/standalone/src/main/resources/MIGRATION.txt
@@ -7,7 +7,7 @@ Please check http://www.mulesoft.org/documentation/display/current/Mule+Release+
 Migration changes from 3.7.x to 3.8.x
 MULE-8626: The HTTP Connector will now ignore a "Connection" outbound property when responding to a request (listener) or making one (request), instead of transforming it to a header. This means that: if such property is desired, it should be explicitly added as a header using a response/request builder.
 MULE-8678: The HTTP Connector will now ignore a "Host" outbound property when making a request, instead of transforming it to a header. This means that: if such property is desired, it should be explicitly added as a header using a request builder.
-MULE-8676: The HTTP Connector will now ignore a "Transfer-Encoding" outbound property when responding a request, instead of transforming it to a header. This means that: if such property is desired, it should be explicitly added as a header using a response builder.
+MULE-8676: The HTTP Connector will now ignore a "Transfer-Encoding" outbound property when sending a response, instead of transforming it to a header. This means that: if such property is desired, it should be explicitly added as a header using a response builder.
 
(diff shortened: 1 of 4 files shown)


afelisatti:
@pablolagreca , @dfeist : We need to decide whether the inconsistency between requestor and listener regarding the presence of both Transfer-Encoding and Content-Length should be solved on the listener side with this PR. The problem can be seen in tests stringWithTransferEncodingAndContentLengthHeader and inputStreamWithTransferEncodingAndContentLengthHeader. Without my changes, the first results in streaming but the second does not. Furthermore, HttpRequestStreamingTestCase's doesNotStreamWithContentLengthTransferEncodingHeadersAndStreamingModeAuto means the requestor does not stream when both headers are present.

dfeist:
IMO the requestor behaviour is correct here, and the listener behaviour inconsistent.

distributions/standalone/src/main/resources/MIGRATION.txt
```diff
@@ -7,6 +7,7 @@ Please check http://www.mulesoft.org/documentation/display/current/Mule+Release+
 Migration changes from 3.7.x to 3.8.x
 MULE-8626: The HTTP Connector will now ignore a "Connection" outbound property when responding to a request (listener) or making one (request), instead of transforming it to a header. This means that: if such property is desired, it should be explicitly added as a header using a response/request builder.
 MULE-8678: The HTTP Connector will now ignore a "Host" outbound property when making a request, instead of transforming it to a header. This means that: if such property is desired, it should be explicitly added as a header using a request builder.
+MULE-8676: The HTTP Connector will now ignore a "Transfer-Encoding" outbound property when responding a request, instead of transforming it to a header. This means that: if such property is desired, it should be explicitly added as a header using a response builder.
```
dfeist: ...when sending a response.., ... respond to a request...

modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseBuilder.java
```diff
@@ -174,7 +174,7 @@ else if (payload instanceof InputStream)
                 try
                 {
                     ByteArrayHttpEntity byteArrayHttpEntity = new ByteArrayHttpEntity(event.getMessage().getPayloadAsBytes());
-                    if (responseStreaming == ALWAYS || (responseStreaming == AUTO && CHUNKED.equals(existingTransferEncoding)))
+                    if (responseStreaming == ALWAYS || (responseStreaming == AUTO && existingContentLength == null && CHUNKED.equals(existingTransferEncoding)))
```
dfeist: Shouldn't we log in the case when both explicit headers are set.

PR #1878, opened 2015-06-17
MULE-8707: Classloader leak using Oracle JDBC Driver

pablokraan:


modules/launcher/src/main/java/org/mule/module/launcher/artifact/DefaultResourceReleaser.java
```diff
@@ -34,6 +40,11 @@ private void deregisterJdbcDrivers()
             {
                 logger.debug("Deregistering driver: {}", driver.getClass());
                 DriverManager.deregisterDriver(driver);
+
+                if (isOracleDriver(driver))
+                {
+                    deregisterOracleDiagnosabilityMBean();
```
marcosnc: I think it should be done just once in case of an oracle driver was deregistered.
What would happen if there are two different oracle driver versions loaded? (for example an app that inserts into two databases or that reads from one and writes to another)

pablokraan: I don't think you can have 2 oracle drivers in the same app. If that happens, there would be only a mbean and it would be registered, second attempt will fail silently, but because this is on application disposal there is no harm. But again, I don't think this is a vlid scenario

PR #1890, opened 2015-06-23
MULE-8723: Mime Type for .CSV files is not being auto detected and set

pablokraan:


core/src/main/resources/META-INF/mime.types
```diff
@@ -2,6 +2,7 @@ text/plain                   txt cgi java
 text/json                    json
 text/xml                     xml
 text/html                    htm html shtml
+text/csv				     csv
```
marcosnc: Indentation

PR #1892, opened 2015-06-23
MULE-8728: Apply minor miscellaneous fixes per the ext-api review

marianogonzalez:


core/src/main/java/org/mule/api/lifecycle/LifecycleUtils.java
```diff
@@ -22,11 +22,28 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(LifecycleUtils.class);
 
+    /**
+     * if {@code object} implements the {@link Initialisable} interface,
```
pablokraan: Start javadocs with upper case

core/src/main/java/org/mule/api/lifecycle/LifecycleUtils.java
```diff
@@ -60,31 +93,81 @@ public static void startIfNeeded(Object object) throws MuleException
...
 
+    /**
+     * if {@code object} implements the {@link Disposable} interface,
+     * then it invokes the {@link Disposable#dispose()} method on it.
+     * If the dispose operation fails, then the exception will be silently
+     * logged using this class {@link #LOGGER}
+     *
+     * @param object the object you're trying to dispose
+     */
     public static void disposeIfNeeded(Object object)
```
pablokraan: Is not clear to me why there are scenarios where I would like to log into this class logger instead of the app logger

core/src/main/java/org/mule/api/registry/ServiceRegistry.java
```diff
@@ -31,11 +27,9 @@
...
      *                      class or interface of the service providers being detected.
      * @param loader        the class loader to be used to load
-     *                      provider-configuration files and instantiate provider classes,
-     *                      or <code>null</code> if the system class loader (or, failing that
-     *                      the bootstrap class loader) is to be used.
-     * @return A {@link Collection} that yields provider objects
-     * for the given service, in some arbitrary order.
+     *                      provider/configuration files and instantiate provider instances.
+     *                      If {@code null}, it will be up to the implementation to choose a {@link ClassLoader}
+     * @return A {@link Collection} that yields provider objects for the given service, in some arbitrary order.
```
pablokraan: Does "Arbitrary order" mean that the order will be always the same or not?

core/src/main/java/org/mule/api/lifecycle/LifecycleUtils.java
pablokraan: I would change the javadoc to something like "Invokes {@link Initialisable#initialise()} on a given object if it implements {@link Initialisable}"

core/src/test/java/org/mule/registry/SpiServiceRegistryTestCase.java
```diff
@@ -0,0 +1,49 @@
...
+    {
+        ClassLoader classLoader = mock(ClassLoader.class, RETURNS_DEEP_STUBS);
+        serviceRegistry.lookupProviders(TestDiscoverableObject.class, classLoader);
+
+        verify(classLoader).getResources("META-INF/services/" + TestDiscoverableObject.class.getName());
+    }
+
+    private void assertProvidersFound(Collection<TestDiscoverableObject> discoveredObjects)
+    {
+        assertThat(discoveredObjects, hasSize(1));
```
pablokraan: Check also that the provider is the desired one

core/src/test/java/org/mule/registry/TestDiscoverableObject.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.registry;
+
+public interface TestDiscoverableObject
```
pablokraan: Can be TestDiscoverableObject and TestDiscoverableObjectImpl inner classes of SpiServiceRegistryTestCase?

core/src/main/java/org/mule/api/registry/ServiceRegistry.java
marianogonzalez: order is not guaranteed

core/src/main/java/org/mule/api/lifecycle/LifecycleUtils.java
marianogonzalez: fixed

core/src/main/java/org/mule/api/lifecycle/LifecycleUtils.java
marianogonzalez: removed the method all together

core/src/test/java/org/mule/registry/TestDiscoverableObject.java
marianogonzalez: nope

core/src/test/java/org/mule/registry/SpiServiceRegistryTestCase.java
marianogonzalez: done

core/src/main/java/org/mule/api/registry/ServiceRegistry.java
pablokraan: Could not be a problem if depending on the environment, providers are returned in different order?

marianogonzalez: it should not. This is just a facade for ServiceLoader which doesn't guarantee it either. Good practice when discovering objects is not to depend  on the order but to weight it some how.

PR #1895, opened 2015-06-24
MULE-8730: Update LICENSE.txt to fix year and url for third-party sof…

anosenzo:
…tware used by Mule

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,7 +1,7 @@
 Mule ESB is distributed under the terms of the Common Public Attribution 
 License (CPAL), whose definition follows below.  For information on the license
 type for each third-party software product used by Mule, refer to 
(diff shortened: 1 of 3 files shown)


LICENSE.txt
```diff
@@ -1,7 +1,7 @@
 Mule ESB is distributed under the terms of the Common Public Attribution 
 License (CPAL), whose definition follows below.  For information on the license
 type for each third-party software product used by Mule, refer to 
-http://www.mulesoft.org/documentation/display/MULE3INTRO/Third-party+Software+in+Mule 
+https://developer.mulesoft.com/docs/display/current/Third-Party+Software+In+Mule
```
marianogonzalez: to only change this file accomplishes nothing. You also have to run mvn license:format on the project root so that this header is updated on all the source files.

PR #1908, opened 2015-06-30
MULE-8728: Apply minor miscellaneous fixes per the ext-api review

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/module/extension/internal/util/CapabilityUtils.java
```diff
@@ -20,7 +20,7 @@
  *
  * @since 3.7.0
  */
-public final class CapabilityUtils
+public abstract class CapabilityUtils
```
pablokraan: Remove abstract and add a private constructor

PR #1931, opened 2015-07-13
MULE-8479: Operations of an Extensible extensions should not require …

afelisatti:
…to be annotated with @ImplementationOf

### DIFF
--- a/extensions/validation/src/main/java/org/mule/extension/validation/internal/NumberValidationOperation.java
+++ b/extensions/validation/src/main/java/org/mule/extension/validation/internal/NumberValidationOperation.java
@@ -23,6 +23,7 @@ import org.apache.commons.lang.StringUtils;
  *
  * @since 3.7.0
  */
+
 public class NumberValidationOperation extends ValidationSupport
 {
 
--- a/extensions/validation/src/main/java/org/mule/extension/validation/internal/ValidationStrategies.java
+++ b/extensions/validation/src/main/java/org/mule/extension/validation/internal/ValidationStrategies.java
@@ -9,7 +9,6 @@ package org.mule.extension.validation.internal;
 import static org.mule.extension.validation.internal.ImmutableValidationResult.error;
 import org.mule.api.MuleEvent;
 import org.mule.api.NestedProcessor;
-import org.mule.extension.annotations.ExtensionOf;
 import org.mule.extension.annotations.Operation;
 import org.mule.extension.annotations.RestrictedTo;
 import org.mule.extension.validation.api.MultipleValidationException;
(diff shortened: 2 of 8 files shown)


extensions/validation/src/main/java/org/mule/extension/validation/internal/NumberValidationOperation.java
```diff
@@ -23,6 +23,7 @@
  *
  * @since 3.7.0
  */
+
```
marianogonzalez: remove

extensions/validation/src/main/java/org/mule/extension/validation/internal/ValidationStrategies.java
```diff
@@ -27,7 +26,7 @@
  *
  * @since 3.7.0
  */
-@ExtensionOf(ValidationExtension.class)
+
```
marianogonzalez: remove

modules/extensions-support/src/main/java/org/mule/module/extension/internal/introspection/AnnotationsBasedDescriber.java
```diff
@@ -244,9 +245,9 @@ private void calculateExtendedTypes(Class<?> actingClass, Method method, Operati
             extensionOf = actingClass.getAnnotation(ExtensionOf.class);
         }
 
-        if (extensionOf != null)
+        if (extensionOf != null || extensionType.getAnnotation(Extensible.class) != null)
         {
-            operation.withCapability(new ExtendingOperationCapability(extensionOf.value()));
+            operation.withCapability(new ExtendingOperationCapability(extensionOf != null ? extensionOf.value() : extensionType));
```
marianogonzalez: this if block is kinda hard to follow, not to mention that per each operation you'll be evaluation extensionOf != null twice (once in the if and again in the ternary).

Do you think it'd be more clear if:

1) Check for extensionType to be extensible in the describer's constructor and store it on a private final boolean field
2) break this down in if (extensionOf != null ) {..} else if(this.isExtensible) {... }

?

modules/extensions-support/src/test/java/org/mule/module/extension/internal/AnnotationsBasedDescriberTestCase.java
```diff
@@ -361,4 +377,57 @@ public void setExtendedProperty(String extendedProperty)
...
+    {
+        @ExtensionOf(ExtensibleExtension.class)
+        @Operation
+        public String sayBye()
+        {
+            return "Bye!";
+        }
+    }
+
+    private static class ExtensibleExtensionOperation
```
marianogonzalez: this class got way too big... Can you split it out to another test which specifically focuses on extensible/extensionOf concerns?

PR #1934, opened 2015-07-14
MULE-8479: Operations of an Extensible extensions should not require …

afelisatti:
…to be annotated with @ImplementationOf

### DIFF
--- a/modules/extensions-support/src/test/java/org/mule/module/extension/internal/AbstractAnnotationsBasedDescriberTestCase.java
+++ b/modules/extensions-support/src/test/java/org/mule/module/extension/internal/AbstractAnnotationsBasedDescriberTestCase.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.module.extension.internal;
+
+import org.mule.extension.introspection.declaration.Describer;
+import org.mule.extension.introspection.declaration.fluent.Declaration;
+import org.mule.extension.introspection.declaration.fluent.OperationDeclaration;
+import org.mule.module.extension.internal.introspection.AnnotationsBasedDescriber;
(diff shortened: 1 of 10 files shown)


modules/extensions-support/src/test/java/org/mule/module/extension/internal/ExtensionOfTestCase.java
```diff
@@ -0,0 +1,122 @@
...
+import org.mule.extension.annotations.Operations;
+import org.mule.extension.introspection.declaration.fluent.Declaration;
+import org.mule.extension.introspection.declaration.fluent.OperationDeclaration;
+import org.mule.module.extension.internal.capability.metadata.ExtendingOperationCapability;
+
+import java.util.Set;
+
+import org.junit.Test;
+
+public class ExtensionOfTestCase extends AbstractAnnotationsBasedDescriberTestCase
```
marianogonzalez: This name seems very weird. I would change it for something like ExtensibleExtensionTestCase but better

modules/extensions-support/src/test/java/org/mule/module/extension/internal/AbstractAnnotationsBasedDescriberTestCase.java
```diff
@@ -0,0 +1,48 @@
...
+import org.mule.tck.junit4.AbstractMuleTestCase;
+import org.mule.util.CollectionUtils;
+
+import org.apache.commons.collections.Predicate;
+
+public abstract class AbstractAnnotationsBasedDescriberTestCase extends AbstractMuleTestCase
+{
+    private Describer describer;
+
+    public Describer getDescriber()
```
marianogonzalez: isn't protected enough?

modules/extensions-support/src/test/java/org/mule/module/extension/internal/AbstractAnnotationsBasedDescriberTestCase.java
```diff
@@ -0,0 +1,48 @@
...
+    {
+        return describer;
+    }
+
+    public void setDescriber(Describer describer)
```
marianogonzalez: same as above

PR #1938, opened 2015-07-17
MULE-8769: Loggers memory leak after fixing MULE-8635

marianogonzalez:


modules/launcher/src/test/java/org/mule/module/launcher/log4j2/LoggerContextReaperThreadFactoryTestCase.java
```diff
@@ -0,0 +1,57 @@
...
+    public void testProperties() throws Exception
+    {
+        assertThat(thread.getName(), is(THREAD_NAME));
+        assertThat(thread.getContextClassLoader(), is(sameInstance(contextClassLoader)));
+    }
+
+    @Test
+    public void runDelegatesIntoTheRightRunnable()
+    {
+        thread.run();
```
marcosnc: You should kill this thread.

marianogonzalez: this is run(), not start(). All this thus is forward to the runnable, it won't actually spawn a thread in the OS

marcosnc: you're right

PR #1941, opened 2015-07-21
MULE-8788: Polling message receiver should unregister scheduler on di…

marcosnc:
…spose.

CronScheduler should not fail if it receives two stop messages, for example when the flow is programmatically stopped and then the application is undeployed.


core/src/main/java/org/mule/transport/polling/MessageProcessorPollingMessageReceiver.java
```diff
@@ -218,7 +219,7 @@ protected void doInitialise() throws InitialisationException
             ((Initialisable) override).initialise();
         }
 
-        getSchedulerFactory().create(schedulerNameOf(this), createWork());
+        createAndRegisterScheduler();
```
pablokraan: createAndRegisterScheduler > createScheduler

core/src/main/java/org/mule/transport/polling/MessageProcessorPollingMessageReceiver.java
```diff
@@ -240,31 +241,62 @@ protected void doStart() throws MuleException
...
+                    ((Disposable) override).dispose();
+                }
+                catch (Exception e)
+                {
+                    logger.warn(String.format("Could not dispose polling override of class %s. Message receiver will continue to dispose", override.getClass().getCanonicalName()), e);
+                }
             }
-            catch (Exception e)
+
+            unregisterAndDisposeScheduler();
```
pablokraan: unregisterAndDisposeScheduler > disposeScheduler

core/src/test/java/org/mule/transport/polling/MessageProcessorPollingMessageReceiverTestCase.java
```diff
@@ -109,6 +117,42 @@ public void testNestedOneWayEndpoint() throws Exception
...
+        receiver.stop();
+        receiver.dispose();
+
+        assertThat(getAllSchedulers().size(), is(0));
+        verify(scheduler).dispose();
+    }
+
+    private Collection<Scheduler> getAllSchedulers()
+    {
+        return muleContext.getRegistry().lookupScheduler(new Predicate<String>()
```
pablokraan: Why don't you lookupUp using lookupObjects(Scheduler.class)?

modules/schedulers/src/main/java/org/mule/modules/schedulers/cron/CronScheduler.java
```diff
@@ -170,7 +174,10 @@ public void stop() throws MuleException
     {
         try
         {
-            quartzScheduler.pauseAll();
+            if (quartzScheduler != null)
```
pablokraan: Is this change needed? the scheduler is nullified at dispose, which is executed after the stop.

PR #1949, opened 2015-07-23
MULE-8796: Unnecessary tracking of overridden non disposable objects …

pablokraan:
…in TransientRegistry

### DIFF
--- a/core/src/test/java/org/mule/registry/TransientRegistryTestCase.java
+++ b/core/src/test/java/org/mule/registry/TransientRegistryTestCase.java
@@ -6,8 +6,12 @@
  */
 package org.mule.registry;
 
+import static org.hamcrest.Matchers.empty;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/mule/registry/TransientRegistryTestCase.java
```diff
@@ -60,12 +66,26 @@ public void testObjectLifecycleDoubleRegistration() throws Exception
     }
 
     @Test
+    public void doesNotTracksNonDisposableOverriddenObjects() throws Exception
+    {
+        final Logger log = mock(Logger.class);
+        final TransientRegistry.RegistryMap registryMap = new TransientRegistry.RegistryMap(log);
+        Object value1 = new Object();
+        Object value2 = new Object();
```
marcosnc: You don't need to add two objects for this test.

PR #1951, opened 2015-07-27
MULE-8798: Message mime type/encoding must be reset when payload is s…

pablokraan:
…et without a datatype

_ Keeping datatype when message is transformed when getPayloadAsX is used and original payload is consumable

### DIFF
--- a/core/src/test/java/org/mule/MuleMessageDataTypePropagationTestCase.java
+++ b/core/src/test/java/org/mule/MuleMessageDataTypePropagationTestCase.java
@@ -8,9 +8,13 @@
 package org.mule;
 
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.any;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.hasItem;
 import static org.hamcrest.core.IsNot.not;
+import static org.mockito.Matchers.anyObject;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Matchers.argThat;
 import static org.mockito.Mockito.RETURNS_DEEP_STUBS;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/mule/MuleMessageDataTypePropagationTestCase.java
```diff
@@ -267,6 +271,26 @@ public void setsDataType() throws Exception
...
+    {
+        InputStream payload = mock(InputStream.class);
+        DefaultMuleMessage muleMessage = new DefaultMuleMessage(payload, muleContext);
+        muleMessage.setMimeType(APPLICATION_XML);
+        muleMessage.setEncoding(CUSTOM_ENCODING);
+
+        MuleRegistry muleRegistry = mock(MuleRegistry.class);
+        when(muleContext.getRegistry()).thenReturn(muleRegistry);
+        Transformer transformer = mock(Transformer.class);
+        when(transformer.transform(org.mockito.Matchers.any(), org.mockito.Matchers.anyString())).thenReturn("TEST");
```
marianogonzalez: why not static imports?

core/src/test/java/org/mule/MuleMessageDataTypePropagationTestCase.java
```diff
@@ -267,6 +271,26 @@ public void setsDataType() throws Exception
...
+        when(muleRegistry.lookupTransformer(argThat(any(DataType.class)), argThat(any(DataType.class)))).thenReturn(transformer);
+
+        muleMessage.getPayloadAsString();
+
+        assertThat(muleMessage.getDataType().getMimeType(),equalTo(APPLICATION_XML));
```
marianogonzalez: format code

PR #1965, opened 2015-08-05
Mule 3.7.x

sumangalipelli:


pablokraan:
Not sure what are you trying to do, but this is not the right way to do it. Closing

PR #1992, opened 2015-08-20
MULE-8725: Implememnt multitenant configuration eviction policy

marianogonzalez:


core/src/main/java/org/mule/time/Time.java
```diff
@@ -0,0 +1,32 @@
...
+package org.mule.time;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Represents a scalar amount of time expressed in a certain {@link TimeUnit}
+ *
+ * @since 4.0
+ */
+public interface Time
```
pablokraan: This interface is redundant. Rename ImmutableTime to Time and use that.

core/src/main/java/org/mule/time/TimeSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.time;
+
+import java.util.function.Supplier;
+
+/**
+ * A {@link Supplier} which provides the current system time
+ * in milliseconds. This is a singleton class, use {@link #INSTANCE}
```
pablokraan: Why a singleton?

core/src/main/java/org/mule/util/MapUtils.java
```diff
@@ -138,4 +140,27 @@ public static String toString(Map props, boolean newline)
         buf.append('}');
         return buf.toString();
     }
+
+    /**
+     * Puts the {@code key}/{@code value} pair into the given {@code map} only as long
+     * as the {@code key} is not already present on the {@code map}. If the {@code key}
```
pablokraan: Remove the "If the ..." part that is already documented in the @throws section

core/src/main/java/org/mule/util/MapUtils.java
```diff
@@ -138,4 +140,27 @@ public static String toString(Map props, boolean newline)
...
+     * @param map   a {@link Map}
+     * @param key   the key
+     * @param value the value
+     * @param <K>   the generic type of the key
+     * @param <V>   the generic type of the value
+     * @throws IllegalStateException if the {@code map} already contains the {@code key}
+     */
+    public static <K, V> void idempotentPut(Map<K, V> map, K key, V value)
+    {
+        checkArgument(value != null, "value cannot be null");
```
pablokraan: Why the value can't be null but the key can?

core/src/main/java/org/mule/util/MapUtils.java
```diff
@@ -138,4 +140,27 @@ public static String toString(Map props, boolean newline)
...
+        if (map.containsKey(key))
```
pablokraan: This is not thread safe

core/src/main/java/org/mule/util/MapUtils.java
```diff
@@ -138,4 +140,27 @@ public static String toString(Map props, boolean newline)
...
+        {
+            throw new IllegalStateException(String.format("A %s is already registered for the name '%s'", value.getClass().getSimpleName(), key));
```
pablokraan: A %s is already registered for the name '%s' > Key '%s' is already registered with value %s

core/src/test/java/org/mule/tck/util/TestTimeSupplier.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+/**
+ * A test {@link Supplier} for externalizing the system time. It is initialised at a given
+ * initial {@link #timeInMillis} using the {@link #TestTimeSupplier(long)} constructor. The value can be
+ * retrieved through the {@link #get()} method.
+ * <p/>
+ * The usefulness of ths class comes from the behavior provided by the {@link #move(long, TimeUnit)}
+ * method which allows changing that value per the tests needs. Again, the modified value is to be retrieved
```
pablokraan: "Again, the modified value..." this part is a bit confusing. The original value is there before you call to #move, so you can get the original value with #get before changing it.

core/src/test/java/org/mule/tck/util/TestTimeSupplier.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+public class TestTimeSupplier implements Supplier<Long>
+{
+
+    private long timeInMillis;
+
+    /**
+     * Creates a new instance
+     *
+     * @param timeInMillis the initial value in milliseconds
```
pablokraan: timeInMillis: time in milliseconds to be supplied until its modified by #move(..)

core/src/test/java/org/mule/tck/util/TestTimeSupplier.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    @Override
+    public Long get()
+    {
+        return timeInMillis;
+    }
+
+    /**
+     * Moves the current {@link #timeInMillis} by the given {@code time} which is expressed
+     * in the given {@code unit}. Note that {@code time} can be a negavite number in which case
```
pablokraan: Add \<p/> before "Note that..."

pablokraan: negavite > negative

core/src/test/java/org/mule/tck/util/TestTimeSupplier.java
```diff
@@ -0,0 +1,60 @@
...
+     * you would be "going back in time"
+     *
+     * @param timeOffset the offset to be applied on {@link #timeInMillis}
+     * @param unit       a {@link TimeUnit} which qualifies the {@code timeOffset}
+     * @return the updated {@link #timeInMillis}
```
pablokraan: Is needed to return the updated time when you can get that value invoking the #get method?

modules/extensions-spring-support/src/main/java/org/mule/module/extension/internal/config/ConfigurationInstanceProviderFactoryBean.java
```diff
@@ -89,4 +85,30 @@ public boolean isSingleton()
...
+
+        return new ImmutableDynamicConfigPolicy(getExpirationPolicy(policyElement));
+    }
+
+    private ExpirationPolicy getExpirationPolicy(ElementDescriptor dynamicConfigPolicyElement)
+    {
+        ElementDescriptor expirationPolicyElement = dynamicConfigPolicyElement.getChildByName("expiration-policy");
+        if (expirationPolicyElement == null)
+        {
+            return new ImmutableExpirationPolicy(5, TimeUnit.MINUTES, TimeSupplier.INSTANCE);
```
pablokraan: How the time supplier will help to test our code if the supplier instance is hardcoded here?

modules/extensions-spring-support/src/main/java/org/mule/module/extension/internal/config/ExtensionConfigNamespaceHandler.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.springframework.beans.factory.xml.NamespaceHandler;
+
+/**
+ * A {@link NamespaceHandler} for elements that configure how the platform
+ * handles extensions
+ *
+ * @since 4.0
+ */
+public class ExtensionConfigNamespaceHandler extends AbstractMuleNamespaceHandler
```
pablokraan: Why do you need to put theses bean definition parsers in a handler other than ExtensionNamespaceHandler?

modules/extensions-spring-support/src/main/resources/META-INF/mule-extension.xsd
```diff
@@ -0,0 +1,113 @@
...
+<xsd:schema xmlns="http://www.mulesoft.org/schema/mule/extension"
+            xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+            xmlns:mule="http://www.mulesoft.org/schema/mule/core"
+            targetNamespace="http://www.mulesoft.org/schema/mule/extension"
+            elementFormDefault="qualified"
+            attributeFormDefault="unqualified">
+
+    <xsd:import namespace="http://www.w3.org/XML/1998/namespace"/>
+    <xsd:import namespace="http://www.mulesoft.org/schema/mule/core"
+                schemaLocation="http://www.mulesoft.org/schema/mule/core/3.7/mule.xsd"/>
```
pablokraan: 3.7 > 4.0?

modules/extensions-spring-support/src/main/resources/META-INF/mule-extension.xsd
```diff
@@ -0,0 +1,113 @@
...
+
+    <xsd:annotation>
+        <xsd:documentation>
+            Configuration elements common to all Mule Extensions
```
pablokraan: If this is for application's extensions only, should not make that explicit on the file name? (thinking about creating server extensions in the future)

modules/extensions-spring-support/src/main/resources/META-INF/mule-extension.xsd
```diff
@@ -0,0 +1,113 @@
...
+
+    <xsd:complexType name="dynamicConfigurationExpirationType">
+        <xsd:attribute name="frequency" type="mule:substitutableLong" use="required">
+            <xsd:annotation>
+                <xsd:documentation>
+                    A scalar time value for how often should the platform check for expirable dynamic configs
+                </xsd:documentation>
+            </xsd:annotation>
+        </xsd:attribute>
+        <xsd:attribute name="timeUnit" type="mule:timeUnitType" use="required">
```
pablokraan: Can't use milliseconds as default value?

modules/extensions-spring-support/src/main/resources/META-INF/mule-extension.xsd
```diff
@@ -0,0 +1,113 @@
...
+            <xsd:documentation>
+                Settings which define how the platform should manage dynamic instances of a configuration element.
+            </xsd:documentation>
+        </xsd:annotation>
+        <xsd:sequence>
+            <xsd:choice>
+                <xsd:element name="expiration-policy" type="expirationPolicyType" minOccurs="0" maxOccurs="1">
+                    <xsd:annotation>
+                        <xsd:documentation>
+                            Configures the maximum amount of time that a dynamic configuration instance can remain idle
```
pablokraan: Configures the maximum amount of time > is not the minimum amount of time?

modules/extensions-spring-support/src/test/java/org/mule/module/extension/internal/DynamicConfigExpirationTestCase.java
```diff
@@ -0,0 +1,67 @@
...
+        final MuleEvent event = getTestEvent(myName);
+        String returnedName = runFlow("dynamic", event).getMessage().getPayloadAsString();
+
+
+        HeisenbergExtension config = muleContext.getRegistry().lookupObject(HeisenbergExtension.class);
+
+        // validate we actually hit the correct dynamic config
+        assertThat(returnedName, is(myName));
+        assertThat(config.getPersonalInfo().getName(), is(myName));
+
```
pablokraan: This test should use the test time supplier

modules/extensions-support/src/main/java/org/mule/module/extension/internal/config/ExtensionConfig.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.mule.time.Time;
+
+/**
+ * A config object which gathers all the configuration properties about how
+ * the platform should handle extensions
+ *
+ * @since 4.0
+ */
+public interface ExtensionConfig
```
pablokraan: Why do you need an interface for this?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ConfigurationExpirationMonitor.java
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Monitors configuration instances looking for those which should be expired.
+ *
+ * It only exposes {@link #start()} and {@link #stop()} method to signal
+ * when monitoring should start/stop.
+ *
+ * @since 4.0
+ */
+public interface ConfigurationExpirationMonitor extends Startable, Stoppable
```
pablokraan: This interface does not make any sense to me, at least no with this name, because it does not has anything to do with expiration monitoring.

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ConfigurationInstanceProviderWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+import org.mule.extension.runtime.ConfigurationInstanceProvider;
+import org.mule.module.extension.internal.runtime.ExpirableContainer;
+
+import com.google.common.collect.ImmutableMap;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * A wrapper class which decorates a {@link #configurationInstanceProvider} with extra
```
pablokraan: A wrapper class which decorates a > Decorates a

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ConfigurationInstanceProviderWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @since 4.0
+ */
+final class ConfigurationInstanceProviderWrapper implements ExpirableContainer<Object>
+{
+
+    private final ConfigurationInstanceProvider<?> configurationInstanceProvider;
+    private final Map<String, Object> configurationInstances = new ConcurrentHashMap<>();
+
+    ConfigurationInstanceProviderWrapper(ConfigurationInstanceProvider<?> configurationInstanceProvider)
```
pablokraan: Add javadocs

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ConfigurationInstanceProviderWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+    {
+        this.configurationInstanceProvider = configurationInstanceProvider;
+    }
+
+    ConfigurationInstanceProvider<?> getConfigurationInstanceProvider()
```
pablokraan: why default visibility?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ConfigurationInstanceProviderWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+        return configurationInstanceProvider;
+    }
+
+    void addConfigurationInstance(String registrationName, Object configurationInstance)
+    {
+        idempotentPut(configurationInstances, registrationName, configurationInstance);
+    }
+
+    /**
+     * if {@link #configurationInstanceProvider} implements the {@link ExpirableContainer} interface
```
pablokraan: This javadoc is not clear at all

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultConfigurationExpirationMonitor.java
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        private DefaultConfigurationExpirationMonitor manager = new DefaultConfigurationExpirationMonitor();
+
+        private Builder()
+        {
+        }
+
+        /**
+         * Specifies how ofter should this instance check for expired config instances
```
pablokraan: ofter > often

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultConfigurationExpirationMonitor.java
```diff
@@ -0,0 +1,201 @@
...
+    /**
+     * Starts a scheduler which fires expiration tasks on the given {@link #frequency} and executes
+     * the {@link #expirationHandler} on each matching configuration instance
+     *
+     * @throws MuleException if the manager could not be started
+     */
+    @Override
+    public void start() throws MuleException
+    {
+        executor = Executors.newSingleThreadScheduledExecutor(r -> new Thread(r, getPrefix(muleContext) + "extension.expiration.manager"));
```
pablokraan: Is there any way you can provide a scheduler that uses the TimeSupplier instead of depending on the real time?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultConfigurationExpirationMonitor.java
```diff
@@ -0,0 +1,201 @@
...
+        executor.scheduleWithFixedDelay(() -> expire(), frequency, frequency, timeUnit);
+    }
+
+    private void expire()
+    {
+        if (muleContext.isStopping() || muleContext.isStopped())
```
pablokraan: Can you execute this check every time an entry is checked for expiration instead of checking just once at the beginning of the method?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultConfigurationExpirationMonitor.java
```diff
@@ -0,0 +1,201 @@
...
+        final String key = config.getKey();
+        LOGGER.debug("Expiring configuration of key {}", key);
+        try
+        {
+            expirationHandler.accept(key, config.getValue());
+            LOGGER.debug("Configuration of key {} was expired", key);
+        }
+        catch (Exception e)
+        {
+            LOGGER.error(String.format("Could not process expiration for dynamic config '%s' of type '%s'. Will try again on next cycle",
```
pablokraan: There are log messages created using templates like "{}" and others using "%s". Can we use only "{}"?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultConfigurationExpirationMonitor.java
```diff
@@ -0,0 +1,201 @@
...
+                                       key, config.getValue().getClass().getName()), e);
+        }
+    }
+
+    /**
+     * Orderly shutdowns the scheduler that executes the expiration tasks. It waits up to 30 seconds
```
pablokraan: Orderly shutdowns > Shutdowns

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultExtensionManager.java
```diff
@@ -50,6 +58,7 @@
     private ObjectNameHelper objectNameHelper;
     private ExtensionDiscoverer extensionDiscoverer = new DefaultExtensionDiscoverer(new DefaultExtensionFactory(serviceRegistry), serviceRegistry);
     private ImplicitConfigurationFactory implicitConfigurationFactory;
+    private ConfigurationExpirationMonitor configurationExpirationMonitor;
```
pablokraan: Why there is only a ConfigurationExpirationMonitor for all the different types of available extensions?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultExtensionManager.java
```diff
@@ -59,6 +68,29 @@ public void initialise() throws InitialisationException
     }
 
     /**
+     * It starts the {@link #configurationExpirationMonitor}
```
pablokraan: It starts > Starts

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultExtensionManager.java
```diff
@@ -262,6 +294,41 @@ private void logExtensionHotUpdate(Extension extension, Extension actual)
         }
     }
 
+    private ConfigurationExpirationMonitor newConfigurationExpirationManager()
```
pablokraan: can you inject the ConfigurationExpirationMonitor instead of having this dependency?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultExtensionManager.java
```diff
@@ -262,6 +294,41 @@ private void logExtensionHotUpdate(Extension extension, Extension actual)
...
+        {
+            LOGGER.error(String.format("Could not unregister expired dynamic config of key '%s' and type %s",
+                                       key, object.getClass().getName()), e);
+        }
+    }
+
+    private Time getConfigurationExpirationFrequency()
+    {
+        ExtensionConfig extensionConfig = muleContext.getConfiguration().getExtension(ExtensionConfig.class);
+        if (extensionConfig != null)
```
pablokraan: Should not every extension have a ExtensionConfig?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ExtensionRegistry.java
```diff
@@ -103,18 +104,26 @@ ExtensionStateTracker getExtensionState(Extension extension)
...
+            getExtensions().stream().filter(extension -> extension.isCapableOf(capabilityType)).forEach(capables::add);
             cachedCapables = capables.build();
             capabilityToExtension.put(capabilityType, cachedCapables);
         }
 
         return cachedCapables;
     }
+
+    /**
+     * Returns a {@link Map} which values are the configuration instances which should be expired
```
pablokraan: Would be more readable if you explain keys and then values

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ExtensionRegistry.java
```diff
@@ -103,18 +104,26 @@ ExtensionStateTracker getExtensionState(Extension extension)
...
+     * and the keys are they registration keys
+     *
+     * @return a immutable {@link Map}
```
pablokraan: a immutable > an immutable

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ConfigurationInstanceProviderWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+ * management information like the configuration instances it has generated and their
+ * registration keys.
+ * <p/>
+ * It also implements the {@link ExpirableContainer} interface which means that it can
+ * determine which configuration instances should be expired (only as long as the
+ * {@link #configurationInstanceProvider} itself implements the same interface
+ *
+ * @since 4.0
+ */
+final class ConfigurationInstanceProviderWrapper implements ExpirableContainer<Object>
```
pablokraan: After reviewing the usage, this seems to be more a ExpirableConfigurationInstanceProviderContainer than a  ConfigurationInstanceProviderWrapper

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ExtensionStateTracker.java
```diff
@@ -74,29 +70,45 @@
...
      * Registers the creation of a new configuration instance
      *
-     * @param instanceKey           the key under which the instance was registered in the {@link MuleRegistry}
+     * @param providerName          the name of a registered {@link ConfigurationInstanceProvider}
+     * @param registrationName      the name on which the {@code configurationInstance} has been registered on the {@link MuleRegistry}
      * @param configurationInstance the configuration instance. Cannot be {@code null}
      * @param <C>                   the generic type for the {@code configurationInstance}
      */
-    <C> void registerConfigurationInstance(String instanceKey, C configurationInstance)
+    <C> void registerConfigurationInstance(String providerName, String registrationName, C configurationInstance)
```
pablokraan: registrationName > instanceName?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/AbstractOperationExecutor.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+package org.mule.module.extension.internal.runtime;
+
+import org.mule.extension.runtime.OperationContext;
+import org.mule.extension.runtime.OperationExecutor;
+
+import java.util.function.Consumer;
+
+/**
+ * Base class for {@link OperationExecutor}s which handle common behaviors such as notifications, etc.
```
pablokraan: which handle > which handles

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/AbstractOperationExecutor.java
```diff
@@ -0,0 +1,68 @@
...
+        {
+            notify(operationContext, adapter -> adapter.notifyFailedOperation(e));
+            throw e;
+        }
+
+        return result;
+    }
+
+    /**
+     * Implements this method with the actual logic of how to execute the operation
```
pablokraan: Implements this method with the actual logic of how to execute the operation > Executes the actual operation logic

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/BaseObjectBuilder.java
```diff
@@ -25,7 +25,7 @@
  *
  * @since 3.7.0
  */
-abstract class BaseObjectBuilder<T> implements ObjectBuilder<T>
+public abstract class BaseObjectBuilder<T> implements ObjectBuilder<T>
```
pablokraan: Why is this change required?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/ImmutableDynamicConfigPolicy.java
```diff
@@ -0,0 +1,31 @@
...
+package org.mule.module.extension.internal.runtime;
+
+import org.mule.extension.runtime.ExpirationPolicy;
+
+/**
+ * An immutable implementation of {@link DynamicConfigPolicy}
+ *
+ * @since 4.0
+ */
+public final class ImmutableDynamicConfigPolicy implements DynamicConfigPolicy
```
pablokraan: This separation on class/interface is redundant.

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/ImmutableDynamicConfigPolicy.java
```diff
@@ -0,0 +1,31 @@
...
+{
+
+    private final ExpirationPolicy expirationPolicy;
+
+    public ImmutableDynamicConfigPolicy(ExpirationPolicy expirationPolicy)
```
pablokraan: Add javadocs

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/NullDynamicConfigPolicy.java
```diff
@@ -0,0 +1,35 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.module.extension.internal.runtime;
+
+import org.mule.extension.runtime.ExpirationPolicy;
+
+/**
+ * Null implementation of {@link DynamicConfigPolicy}.
+ * <p/>
+ * This is a singleton class. Instances should be get through the {@link #INSTANCE}
```
pablokraan: Why a singleton?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/NullExpirationPolicy.java
```diff
@@ -0,0 +1,55 @@
...
+package org.mule.module.extension.internal.runtime;
+
+import org.mule.extension.runtime.ExpirationPolicy;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Null implementation of {@link ExpirationPolicy}.
+ * <p/>
+ * This is a singleton class. Instances should be get through the {@link #INSTANCE}
```
pablokraan: Why?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/NullExpirationPolicy.java
```diff
@@ -0,0 +1,55 @@
...
+        return false;
+    }
+
+    /**
+     * Always returns {@code 0}
+     */
+    @Override
+    public long getMaxIdleTime()
+    {
+        return 0;
```
pablokraan: Maybe Integer.MAX_VALUE is better to avoid being all the time checking for expired elements

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/ConfigurationInstanceProviderFactory.java
```diff
@@ -0,0 +1,47 @@
...
+import org.mule.api.MuleContext;
+import org.mule.extension.introspection.Configuration;
+import org.mule.extension.introspection.Extension;
+import org.mule.extension.runtime.ConfigurationInstanceProvider;
+import org.mule.module.extension.internal.manager.ExtensionManagerAdapter;
+import org.mule.module.extension.internal.runtime.DynamicConfigPolicy;
+import org.mule.module.extension.internal.runtime.resolver.ResolverSet;
+
+/**
+ * A factory class which creates instances of {@link ConfigurationInstanceProvider}
```
pablokraan: A factory class which > A factory which

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DefaultConfigurationInstanceProviderFactory.java
```diff
@@ -0,0 +1,69 @@
...
+ * @since 4.0
+ */
+public final class DefaultConfigurationInstanceProviderFactory implements ConfigurationInstanceProviderFactory
+{
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public <T> ConfigurationInstanceProvider<T> createConfigurationInstanceProvider(
```
pablokraan: This method is doing two different things that, in my opinion, should be two separated methods. (ie: dynamic/static providers)

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceHolder.java
```diff
@@ -0,0 +1,108 @@
...
+package org.mule.module.extension.internal.runtime.config;
+
+import org.mule.extension.runtime.Expirable;
+import org.mule.extension.runtime.ExpirationPolicy;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * A holder class which decorates a {@link #configurationInstance} with extra
```
pablokraan: A holder class which decorates > Decorates

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceHolder.java
```diff
@@ -0,0 +1,108 @@
...
+final class DynamicConfigurationInstanceHolder implements Expirable
+{
+
+    private final String registrationName;
+    private final Object configurationInstance;
+
+    private final AtomicInteger usageCount = new AtomicInteger(0);
+    private long lastUsedMillis = now();
+
+    public DynamicConfigurationInstanceHolder(String registrationName, Object configurationInstance)
```
pablokraan: Add javadoc

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceHolder.java
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * A holder class which decorates a {@link #configurationInstance} with extra
+ * management information such as registration name, usage tracking, etc.
+ * <p/>
+ * This class is thread safe.
+ *
+ * @since 4.0
+ */
+final class DynamicConfigurationInstanceHolder implements Expirable
```
pablokraan: DynamicConfigurationInstanceHolder > ExpirableDynamicConfigurationInstance?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceHolder.java
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * Returns the held configuration instance
+     */
+    public Object getConfigurationInstance()
+    {
+        return configurationInstance;
+    }
+
+    /**
+     * acknowledges and tracks the fact that there's an ongoing operation execution
```
pablokraan: Is this javadoc in the right method?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceHolder.java
```diff
@@ -0,0 +1,108 @@
...
+     * using the {@link #configurationInstance}. This method is thread safe.
+     *
+     * @return the updated usage count.
+     */
+    public int accountUsage()
```
pablokraan: AccountUsage suggest to me that this will return the current number of usages, not that it will also increment it.
Can't make this void and rename to startUsage/stopUsage or similar?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceHolder.java
```diff
@@ -0,0 +1,108 @@
...
+     */
+    public String getRegistrationName()
+    {
+        return registrationName;
+    }
+
+    /**
+     * @return {@code true} if there's no ongoing operation using the {@link #configurationInstance}. {@code false} otherwise
+     */
+    private boolean isNotInUse()
```
pablokraan: Change it to isInUse

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceProvider.java
```diff
@@ -0,0 +1,180 @@
...
+ * A {@link ConfigurationInstanceProvider} which continuously evaluates the same
+ * {@link ResolverSet} and then uses the resulting {@link ResolverSetResult}
+ * to build an instance of a given type.
+ * <p/>
+ * Although each invocation to {@link #get(OperationContext)} is guaranteed to end up in an invocation
+ * to {@link #resolverSet#resolve(MuleEvent)}, the resulting {@link ResolverSetResult} might not end up
+ * generating a new instance. This is so because {@link ResolverSetResult} instances are put in a cache to
+ * guarantee that equivalent evaluations of the {@code resolverSet} return the same instance.
+ *
+ * @since 3.7.0
```
pablokraan: Is correct to maintain 3.7.0 here when the class was move to a different package?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceProvider.java
```diff
@@ -0,0 +1,180 @@
...
+     */
+    @Override
+    public T get(OperationContext operationContext)
+    {
+        try
+        {
+            ResolverSetResult result = resolverSet.resolve(asOperationContextAdapter(operationContext).getEvent());
+            DynamicConfigurationInstanceHolder configurationInstanceHolder = getHolder(result);
+
+            operationContext.onOperationSuccessful(signal -> discountUsage(configurationInstanceHolder));
```
pablokraan: Can reuse the same lamba for both success/failure?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceProvider.java
```diff
@@ -0,0 +1,180 @@
...
+            operationContext.onOperationFailed(signal -> discountUsage(configurationInstanceHolder));
+
+            configurationInstanceHolder.accountUsage();
```
pablokraan: This must be execute in the method that creates the configurationInstanceHolder, otherwise there is a race condition

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceProvider.java
```diff
@@ -0,0 +1,180 @@
...
+            return (T) configurationInstanceHolder.getConfigurationInstance();
+        }
+        catch (Exception e)
+        {
+            throw new MuleRuntimeException(e);
+        }
+    }
+
+    private DynamicConfigurationInstanceHolder getHolder(ResolverSetResult result) throws Exception
```
pablokraan: Rename result to resolverSetResult

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceProvider.java
```diff
@@ -0,0 +1,180 @@
...
+                    // re-check in case some other thread beat us to it...
+                    configurationInstance = cache.get(result);
+                    if (configurationInstance == null)
+                    {
+                        configurationInstance = createAndRegisterConfigurationInstance(result);
+                        cache.put(result, configurationInstance);
+                    }
+
+                    // downgrade
+                    cacheReadLock.lock();
```
pablokraan: This downgrade seems required only because the outer block of code is surrounded by a finally that is doing cacheReadLock.unlock(). I think the code would be more readable if it has tow parts, one using only the read lock to check for existence and another one using only the write lock

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceProvider.java
```diff
@@ -0,0 +1,180 @@
...
+        }
+        finally
+        {
+            cacheReadLock.unlock();
+        }
+
+        return configurationInstance;
+    }
+
+    private DynamicConfigurationInstanceHolder createAndRegisterConfigurationInstance(ResolverSetResult result) throws MuleException
```
pablokraan: Find a better name for createAndRegisterConfigurationInstance

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceProvider.java
```diff
@@ -0,0 +1,180 @@
...
+        }
+        finally
+        {
+            cacheWriteLock.unlock();
+        }
+
+        return expiredConfigs.build();
+    }
+
+    private void discountUsage(DynamicConfigurationInstanceHolder configurationInstanceHolder)
```
pablokraan: Inline this method

modules/extensions-support/src/main/java/org/mule/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -51,7 +47,7 @@
  *
  * @since 3.7.0
  */
-public class MuleExtensionUtils
+public final class MuleExtensionUtils
```
pablokraan: Why do you need the final keyword here?

modules/extensions-support/src/test/java/org/mule/module/extension/internal/runtime/DefaultOperationContextTestCase.java
```diff
@@ -0,0 +1,120 @@
...
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@SmallTest
+@RunWith(MockitoJUnitRunner.class)
+public class DefaultOperationContextTestCase extends AbstractMuleTestCase
+{
+
+    private static final String PARAM_NAME = "param1";
+    public static final String VALUE = "Do you want to build a snowman?";
```
pablokraan: private

modules/spring-config/src/main/java/org/mule/config/spring/parsers/specific/TimeFactoryBean.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.mule.time.ImmutableTime;
+import org.mule.time.Time;
+
+import java.util.concurrent.TimeUnit;
+
+import org.springframework.beans.factory.FactoryBean;
+
+/**
+ * A {@link FactoryBean} which crates instances of {@link ImmutableTime}
```
pablokraan: crates > creates

modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -10,9 +10,9 @@
 
     <xsd:import namespace="http://www.w3.org/XML/1998/namespace"/>
     <xsd:import namespace="http://www.springframework.org/schema/beans"
-                schemaLocation="http://www.springframework.org/schema/beans/spring-beans-3.1.xsd"/>
+                schemaLocation="http://www.springframework.org/schema/beans/spring-beans-4.1.xsd"/>
     <xsd:import namespace="http://www.springframework.org/schema/context"
-                schemaLocation="http://www.springframework.org/schema/context/spring-context-3.1.xsd"/>
+                schemaLocation="http://www.springframework.org/schema/context/spring-context-4.1.xsd"/>
```
pablokraan: Is not this something to fix on every schema file?

core/src/main/java/org/mule/time/Time.java
marianogonzalez: done

core/src/main/java/org/mule/time/TimeSupplier.java
marianogonzalez: This class simply returns System.currentTimeInMillis(). It's completely stateless and all instances will behave the same. Having multiple instances of it doesn't make much sense memory wise.

core/src/main/java/org/mule/util/MapUtils.java
marianogonzalez: done

core/src/main/java/org/mule/util/MapUtils.java
marianogonzalez: this is a mistake. The key should be not null. It should not be up to this method to validate the value

core/src/main/java/org/mule/util/MapUtils.java
marianogonzalez: fixed

core/src/main/java/org/mule/util/MapUtils.java
marianogonzalez: this method is not intended to be thread safe. I'll add that to the javadoc

core/src/test/java/org/mule/tck/util/TestTimeSupplier.java
marianogonzalez: changed it to be more clear. What I meant is that after calling move(), you don't get to retrieve the original value again

core/src/test/java/org/mule/tck/util/TestTimeSupplier.java
marianogonzalez: both fixed

core/src/test/java/org/mule/tck/util/TestTimeSupplier.java
marianogonzalez: fixed

core/src/test/java/org/mule/tck/util/TestTimeSupplier.java
marianogonzalez: it can be handy in some occasions and it doesn't hurt at all.

modules/extensions-spring-support/src/main/java/org/mule/module/extension/internal/config/ConfigurationInstanceProviderFactoryBean.java
marianogonzalez: There are unit tests in which ImmutableExpirationPolicy is used with a TestTimeSupplier.

For the real production code, we use the real one and test it with a FunctionalTestCase

modules/extensions-spring-support/src/main/java/org/mule/module/extension/internal/config/ExtensionConfigNamespaceHandler.java
marianogonzalez: ExtensionNamespaceHandler is the generic and dynamic one which can handle any extension. There's an instance of it per each discovered extension. Each instance is bound to each extension's namespace

This one is unique and is bounded uniquely to the 'extension' namespace. This one is static and serves one purpose only which is configuring platform wide extensions properties.

It makes sense to have them separate because they have different concerns and purposes

modules/extensions-spring-support/src/main/resources/META-INF/mule-extension.xsd
marianogonzalez: yes, but that has to happen when we upgrade all schemas. If i do that right now chances are it won't work

modules/extensions-spring-support/src/main/resources/META-INF/mule-extension.xsd
marianogonzalez: I would do that change when/if that actually happens. I don't think we're sure how such a feature would look like anyway

modules/extensions-spring-support/src/main/resources/META-INF/mule-extension.xsd
marianogonzalez: I don't think it makes sense to default MILLISECONDS when the default value itself is in minutes. Also, having a default for the timeUnit forces the user to go see what's the default. By requesting both values, the IDE guides the user into a more clear experience, since the user naturally thinks of time in units anyway

modules/extensions-spring-support/src/main/resources/META-INF/mule-extension.xsd
marianogonzalez: nop.. It's up to how much time will you allow config instances to slack before being purge. It's a maximum not a minimum. The minimum will always be zero

modules/extensions-spring-support/src/test/java/org/mule/module/extension/internal/DynamicConfigExpirationTestCase.java
marianogonzalez: No. What I'm testing here is that the expiration configuration in dynamic-config-expiration.xml is applied. I need to use whatever TimeSupplier the platform uses. 

The TestTimeSupplier is used in unit tests below

modules/extensions-support/src/main/java/org/mule/module/extension/internal/config/ExtensionConfig.java
marianogonzalez: The actual instance of extension config is parsed using mule's legacy parsers which require the setter. I don't want to expose the setter so I use the interface

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ConfigurationExpirationMonitor.java
marianogonzalez: how would you do it? You need a component with that responsibility and you need to be able to start and stop it

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ConfigurationInstanceProviderWrapper.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ConfigurationInstanceProviderWrapper.java
marianogonzalez: renamed

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ConfigurationInstanceProviderWrapper.java
marianogonzalez: to match the visibility of the class itself

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ConfigurationInstanceProviderWrapper.java
marianogonzalez: how come?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultConfigurationExpirationMonitor.java
marianogonzalez: fixed

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultConfigurationExpirationMonitor.java
marianogonzalez: I'm sure there's a way. I think however that should be a separate story, out of the scope of this PR

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultConfigurationExpirationMonitor.java
marianogonzalez: The end result would be the same because this is done in a way that if the mule context starts stopping while this task is running, it will wait for it to finish.

However, checking on each iteration is a good optimization. I'll do that

marianogonzalez: actually, the best is to check at both places,  at the start plus on each iteration

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultConfigurationExpirationMonitor.java
marianogonzalez: no. As convenient as it is, the {} thing only works when you want to format a message without an exception. If you want the logger to also log the exception, then you need to build it yourself

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultConfigurationExpirationMonitor.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultExtensionManager.java
marianogonzalez: we discussed this with LG and found no use case which justifies having N amount of expiry threads running. Let's discuss this offline if you do have a case

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultExtensionManager.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultExtensionManager.java
marianogonzalez: this is not a dependency, this is a factory method. Also notice that the returned monitor requires this mananger's extension registry and has a BiConsumer which executes private methods of this manager. I don't think this is a dependency you can inject

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultExtensionManager.java
marianogonzalez: no. Every config can have settings of its own, such as the expiration-policy. But we also need application wide configurations for application wide concerns

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ExtensionRegistry.java
marianogonzalez: I disagree. The registration keys qualify the values. It's more natural to first enunciate a concept and then list the things that qualify it than the other way around

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ExtensionRegistry.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ExtensionStateTracker.java
marianogonzalez: nop. Registration name is more clear. Instance name is more open to interpretation and confusion between that name and the name of the provider, etc

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/AbstractOperationExecutor.java
marianogonzalez: fixed

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/BaseObjectBuilder.java
marianogonzalez: to better separate implementations of this class in more meaningful packages instead of having them all together in a big cat bag

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/ImmutableDynamicConfigPolicy.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/ImmutableDynamicConfigPolicy.java
marianogonzalez: fixed

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/NullDynamicConfigPolicy.java
marianogonzalez: There's no point in doing it differently. It's the same as with the VoidEvent or the NullPayload

core/src/main/java/org/mule/time/TimeSupplier.java
marianogonzalez: It's the same as the VoidEvent or Nullpayload

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ConfigurationInstanceProviderWrapper.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/NullExpirationPolicy.java
marianogonzalez: same answer as before

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/NullExpirationPolicy.java
marianogonzalez: good one

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/ConfigurationInstanceProviderFactory.java
marianogonzalez: fixed

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DefaultConfigurationInstanceProviderFactory.java
marianogonzalez: that depends on your point of view. From the caller's point of view, I just want a ConfigurationInstanceProvider and I'd like to be decoupled from the fact that those can be static or dynamic. To return one or other type can be though of as part of the creation logic that this factory is trying to extract.

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceHolder.java
marianogonzalez: fixed... you're a man of few words I can tell..

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceHolder.java
marianogonzalez: That sounds unnecessary on my opinion. There's no need to reference every single implemented interface on a class name

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceHolder.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceHolder.java
marianogonzalez: don't agree with this one. It doesn't start with the get prefix, so that should be hint enough that this isn't just a getter.
Also, the account verb and the javadoc are pretty explicit on this method actually modifying state in order to track the usage.

Finally, this method does return the current number of usages, so what the method suggests you is what it actually does, which is good

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceHolder.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceProvider.java
marianogonzalez: in my opinion yes, but I'm open to discussion

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceProvider.java
marianogonzalez: no, because the signal is of different types. I though of giving both signals a common ancestor but it didn't make much sense

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceProvider.java
marianogonzalez: no. The same holder can (and hopefully is) used many times. You have to account usage each time you return this instance and discount each time the operation finishes

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceProvider.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceProvider.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceProvider.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceProvider.java
marianogonzalez: that's only a good idea if the two lambdas from your previous comment could be reused. Since that's not the case, I'd rather not inline this

modules/extensions-support/src/main/java/org/mule/module/extension/internal/util/MuleExtensionUtils.java
marianogonzalez: everybody is final until probed otherwise

modules/extensions-support/src/test/java/org/mule/module/extension/internal/runtime/DefaultOperationContextTestCase.java
marianogonzalez: done

modules/spring-config/src/main/resources/META-INF/mule.xsd
marianogonzalez: yes, but that should be a different story to be done once we finish removing components on the branch

core/src/main/java/org/mule/util/MapUtils.java
pablokraan: Is a bit too much to call the method "idempotentPut" when it is not thread safe

modules/extensions-spring-support/src/main/java/org/mule/module/extension/internal/config/ConfigurationInstanceProviderFactoryBean.java
pablokraan: Being able to use a testTimeSupplier is not useful only for unit tests. During a functional/integration test using this supplier could make more easy to write tests that could require complex synchronization using latches otherwise

modules/extensions-spring-support/src/main/resources/META-INF/mule-extension.xsd
pablokraan: My comment was related to the fact that the configuration is not expired when this max is reached, because it depends on when the reaper thread is executed. That is why to me this is a minimum :)

modules/extensions-spring-support/src/test/java/org/mule/module/extension/internal/DynamicConfigExpirationTestCase.java
pablokraan: Why do you need to use "whatever TimeSupplier the platform uses"? You are not testing the time supplier

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ConfigurationInstanceProviderWrapper.java
pablokraan: This javadoc is a direct translation of java to English :). I would say: Returns the expired configuration instances or an empty map if the provider is not a {@link ExpirableContainer}

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultConfigurationExpirationMonitor.java
pablokraan: Why is out of the scope of this PR is you are introducing and using this concept here?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultConfigurationExpirationMonitor.java
```diff
@@ -0,0 +1,207 @@
...
+        catch (Exception e)
+        {
+            LOGGER.error("Found exception trying to expire idle configurations. Will try again on next cycle", e);
+        }
+
+    }
+
+    private void handleExpiration(Map.Entry<String, Object> config)
+    {
+        if (muleContext.isStopping() || muleContext.isStopped())
```
pablokraan: Use isMuleContextStopped here

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultConfigurationExpirationMonitor.java
```diff
@@ -0,0 +1,207 @@
...
+        {
+            executor.awaitTermination(30, TimeUnit.SECONDS);
+        }
+        catch (InterruptedException e)
+        {
+            throw new MuleRuntimeException(createStaticMessage("Exception found while waiting for expiration thread to finish"), e);
+        }
+    }
+
+    private boolean isMuleContextStopped()
```
pablokraan: Is not this something like "stopChecking"?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultExtensionManager.java
pablokraan: Is not clear from your comment if there is or not a dependency here

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ExtensionRegistry.java
pablokraan: A map is "An object that maps keys to values". That is what the method returns. The user sees a map key/value, not "values that are qualified with a name"

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DefaultConfigurationInstanceProviderFactory.java
pablokraan: From the user point of view, having parameters that are used only in one scenario is confusing (muleContext and dynamicConfigPolicy)

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceHolder.java
pablokraan: Is not about referencing every implemented interface. Saying that something is a "holder" is pretty vague.

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceProvider.java
pablokraan: To me is incorrect, because a class importing it won't compile against 3.7.0

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceProvider.java
pablokraan: I understand that the idea is to reuse the holder. The problem is that there is a race condition between getHolder returns the holder and configurationInstanceHolder.accountUsage() is executed, because the getExpired can be executed there.

modules/extensions-support/src/main/java/org/mule/module/extension/internal/util/MuleExtensionUtils.java
pablokraan: Is redundant. Removed it

modules/extensions-spring-support/src/main/resources/META-INF/mule-extension.xsd
marianogonzalez: changed

modules/extensions-spring-support/src/test/java/org/mule/module/extension/internal/DynamicConfigExpirationTestCase.java
marianogonzalez: this is an integration test. I use the TestTimeSupplier to test the things work on its own, then I use this test to verify that they all work together using the same setup I'll get in production.

What's the alternative? If i want to test this XML config with a TestTimeSupplier, I would need a mechanism to provide the supplier and I don't want users tampering with that in real life

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultConfigurationExpirationMonitor.java
marianogonzalez: silly me. Fixed

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultConfigurationExpirationMonitor.java
marianogonzalez: renamed

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ExtensionRegistry.java
marianogonzalez: still disagree. But changed it anyway

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DefaultConfigurationInstanceProviderFactory.java
marianogonzalez: this class and its interface are internal so there's no user, but I see your point. Changed it

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceHolder.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceProvider.java
marianogonzalez: changed

modules/extensions-support/src/main/java/org/mule/module/extension/internal/util/MuleExtensionUtils.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/config/DynamicConfigurationInstanceProvider.java
marianogonzalez: really good catch. Fixed

PR #1996, opened 2015-08-23
MULE-8789: Socket buffer sizes in the HTTP transport for outbound connections not set correctly

andresgregoire:
Changed default memory buffer sizes when the socket buffer sizes are not set in the connector.

### DIFF
--- a/transports/http/src/main/java/org/apache/commons/httpclient/HttpConnection.java
+++ b/transports/http/src/main/java/org/apache/commons/httpclient/HttpConnection.java
@@ -739,11 +739,11 @@ public class HttpConnection {
                 socket.setReceiveBufferSize(rcvBufSize);
             }
             int outbuffersize = socket.getSendBufferSize();
-            if (outbuffersize <= 0) {
+            if (outbuffersize <= 0 || sndBufSize < 0) {
                 outbuffersize = 2048;
             }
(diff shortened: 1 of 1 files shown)


transports/http/src/main/java/org/apache/commons/httpclient/HttpConnection.java
```diff
@@ -739,11 +739,11 @@ public void open() throws IOException {
                 socket.setReceiveBufferSize(rcvBufSize);
             }
             int outbuffersize = socket.getSendBufferSize();
-            if (outbuffersize <= 0) {
+            if (outbuffersize <= 0 || sndBufSize < 0) {
                 outbuffersize = 2048;
```
marianogonzalez: extract constant

marianogonzalez:
wrong code format... wrong code format everywhere

andresgregoire:
This is a class from commons-httpclient, that's why the code format doesn't follow Mule conventions

transports/http/src/main/java/org/apache/commons/httpclient/HttpConnection.java
andresgregoire: Same as the other comment, this code is from commons-httpclient.

marianogonzalez:
fair enough. This is good for now, but we need to set a policy around overriden classes moving forward

PR #1998, opened 2015-08-24
MULE-6417: / by zero in ComponentStatistics

andyhoffy:
Included changes:

* Eliminated '/ by zero' by skipping branch data updates while executedEvent(Count) is zero.
* Added synchronization around clear() - now consistent with other usage of synchronization and fix other potential bugs/race conditions.
* Added ComponentStatisticsTestCase to validate existing behavior and capture functional changes.
* Added javadoc to describe current behavior and known weaknesses/limitations.
    
A better fix may be to introduce an explicit Event object that can accumulate branch events and then be committed as an atomic update. However, that would break existing APIs.

### DIFF
--- a/core/src/main/java/org/mule/management/stats/ComponentStatistics.java
+++ b/core/src/main/java/org/mule/management/stats/ComponentStatistics.java
@@ -16,7 +16,10 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 /**
- * 
+ * ComponentStatistics is a basic metrics aggregation class that is accessible
+ * via the JMX api. This class is not thread-safe - occasional errors in 
+ * reported statistics should be expected, especially when the {@code clear()} 
+ * method is used.
  */
 public class ComponentStatistics implements Statistics
 {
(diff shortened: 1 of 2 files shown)


elrodro83:
Hi Andrew,

We will require that the PR contains only one commit for it to be merged into the repo.

I see that you have merged stuff from some other branches as well, those should be out of the PR. I suggest that you create a new one with only your changes.

Thanks!

andyhoffy:
Ok, thanks for looking at my request, and for the info/feedback - I was trying to follow the directions in the CONTRIBUTING doc, and it sounds like I borked it up. 

For my own future reference - there was a suggestion in the doc to merge the latest updates from the master before submitting a pull request - does it look to you like I messed up the rebasing?

Thanks!
-Andrew

elrodro83:
The branch should have your commit right after the version you get from the rebase. Then you can create a pul request just for that commit.

For this particular case, I'd suggest that you create a new fork, apply your changes and create a new PR.

Thanks,
Rodro

andyhoffy:
All set - thanks for the suggestion. I created a new pull request and I am closing out this one. The new one can be found here:

https://github.com/mulesoft/mule/pull/2043

PR #2007, opened 2015-08-27
Deprecating methods not used anymore on mule 4 branch

pablokraan:


core/src/main/java/org/mule/api/MuleEvent.java
```diff
@@ -322,6 +322,10 @@
 
     void clearFlowVariables();
 
+    /**
+     * @deprecated use {@code getSession()} to manipulate session state
+     */
+    @Deprecated
```
marianogonzalez: shouldn't we wait to get sign off on the message changes spec before doing this?

marianogonzalez:
Sounds to me that we shouldn't do this until the new message structure is merged in 4.x

core/src/main/java/org/mule/api/MuleEvent.java
pablokraan: These changes are not related to the changes we discussed about changing the message/event. This is just cleaning things that are not used anymore or must not be used

PR #2010, opened 2015-08-27
MULE-8476: Fixing tests by adding further fallback. Removing unnecess…

afelisatti:
…ary bits now.


modules/extensions-support/src/test/java/org/mule/module/extension/internal/introspection/ManifestBasedVersionResolverTestCase.java
```diff
@@ -52,6 +52,7 @@ public void worksWithManifest() throws Exception
             {
                 FileUtils.deleteQuietly(manifest);
             }
+            testManifestInputStream.close();
```
marianogonzalez: use a try resource block (preferred) or use IOUtils.closeQuietly

PR #2016, opened 2015-08-31
MULE-8205 Fix of persistent connection and streaming scenarios for HTTP 1.0 clients

elrodro83:
Added corresponding TCs also.

### DIFF
--- a/modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseBuilder.java
+++ b/modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseBuilder.java
@@ -52,6 +52,7 @@ import java.util.Set;
 
 import javax.activation.DataHandler;
 
+import org.glassfish.grizzly.http.Protocol;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -69,6 +70,7 @@ public class HttpResponseBuilder extends HttpMessageBuilder implements Initialis
     private AttributeEvaluator reasonPhraseEvaluator;
     private MuleContext muleContext;
 
+    @Override
     public void initialise() throws InitialisationException
     {
         super.initialise();
@@ -113,7 +115,14 @@ public class HttpResponseBuilder extends HttpMessageBuilder implements Initialis
             final Collection<String> paramValues = resolvedHeaders.getAll(name);
             for (String value : paramValues)
             {
-                httpResponseHeaderBuilder.addHeader(name, value);
+                if (HttpHeaders.Names.TRANSFER_ENCODING.equals(name) && !supportsTransferEncoding(event))
+                {
+                    logger.debug("CLient HTTP version is lower than 1.1 so the unsupported 'Transfer-Encoding' header has been removed and 'Content-Length' will be sent instead.");
+                }
+                else
+                {
+                    httpResponseHeaderBuilder.addHeader(name, value);
+                }
             }
         }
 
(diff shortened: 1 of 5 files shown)


modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerResponseStreamingTestCase.java
```diff
@@ -46,63 +47,126 @@ protected String getConfigFile()
     public void string() throws Exception
     {
         final String url = getUrl("string");
-        testResponseIsContentLengthEncoding(url);
+        testResponseIsContentLengthEncoding(url, HttpVersion.HTTP_1_1);
```
afelisatti: This class has become way too crowded. I think it would be best to add an abstract class and then have one for each HTTP version. Parameterising it with the two versions would be even better but I'm not sure whether that's possible.

modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseBuilder.java
```diff
@@ -113,7 +115,14 @@ public HttpResponse build(org.mule.module.http.internal.domain.response.HttpResp
             final Collection<String> paramValues = resolvedHeaders.getAll(name);
             for (String value : paramValues)
             {
-                httpResponseHeaderBuilder.addHeader(name, value);
+                if (HttpHeaders.Names.TRANSFER_ENCODING.equals(name) && !supportsTransferEncoding(event))
+                {
+                    logger.debug("CLient HTTP version is lower than 1.1 so the unsupported 'Transfer-Encoding' header has been removed and 'Content-Length' will be sent instead.");
```
afelisatti: Client.

modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseBuilder.java
```diff
@@ -208,11 +220,24 @@ else if (payload instanceof InputStream)
         return httpResponseBuilder.build();
     }
 
-    private void resolveEncoding(HttpResponseHeaderBuilder httpResponseHeaderBuilder, String existingTransferEncoding, String existingContentLength, ByteArrayHttpEntity byteArrayHttpEntity)
+    private boolean supportsTransferEncoding(MuleEvent event)
+    {
+        String httpVersion = event.getMessage().<String> getInboundProperty(HttpConstants.RequestProperties.HTTP_VERSION_PROPERTY);
```
afelisatti: Static imports.

modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/ResponseStreamingCompletionHandler.java
```diff
@@ -109,6 +118,14 @@ public void completed(WriteResult result)
         }
     }
 
+    private void doComplete()
+    {
+        close();
+        responseStatusCallback.responseSendSuccessfully();
+        ctx.notifyDownstream(HttpServerFilter.RESPONSE_COMPLETE_EVENT);
```
afelisatti: Static imports.

modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerPersistentConnectionsTestCase.java
```diff
@@ -53,127 +61,203 @@ protected String getConfigFile()
     @Test
     public void nonPersistentCheckHeader() throws Exception
     {
-        assertThat(performRequest(nonPersistentPort.getNumber()), is(CLOSE));
+        assertThat(performRequest(nonPersistentPort.getNumber(), HttpVersion.HTTP_1_1, false), is(CLOSE));
```
afelisatti: Same as below. I think dividing this tests makes sense.

PR #2023, opened 2015-08-31
MULE-8205 Fix persistent connection and streaming for HTTP 1.0 clients

elrodro83:
When the payload was an InputStream and the client HTTP version was 1.0, the payload stream was not being closed. Thas is also fixed now.

### DIFF
--- a/modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseBuilder.java
+++ b/modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseBuilder.java
@@ -7,6 +7,8 @@
 package org.mule.module.http.internal.listener;
 
 import static org.mule.module.http.api.HttpConstants.RequestProperties.HTTP_PREFIX;
+import static org.mule.module.http.api.HttpConstants.RequestProperties.HTTP_STATUS_PROPERTY;
+import static org.mule.module.http.api.HttpConstants.RequestProperties.HTTP_VERSION_PROPERTY;
 import static org.mule.module.http.api.HttpHeaders.Names.CONNECTION;
 import static org.mule.module.http.api.HttpHeaders.Names.CONTENT_LENGTH;
 import static org.mule.module.http.api.HttpHeaders.Names.TRANSFER_ENCODING;
@@ -21,7 +23,6 @@ import org.mule.api.context.MuleContextAware;
 import org.mule.api.lifecycle.Initialisable;
 import org.mule.api.lifecycle.InitialisationException;
 import org.mule.api.transformer.DataType;
-import org.mule.module.http.api.HttpConstants;
 import org.mule.module.http.api.HttpHeaders;
 import org.mule.module.http.api.requester.HttpStreamingType;
 import org.mule.module.http.internal.HttpMessageBuilder;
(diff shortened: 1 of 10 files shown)


modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseBuilder.java
```diff
@@ -208,11 +221,24 @@ else if (payload instanceof InputStream)
         return httpResponseBuilder.build();
     }
 
-    private void resolveEncoding(HttpResponseHeaderBuilder httpResponseHeaderBuilder, String existingTransferEncoding, String existingContentLength, ByteArrayHttpEntity byteArrayHttpEntity)
+    private boolean supportsTransferEncoding(MuleEvent event)
+    {
+        String httpVersion = event.getMessage().<String> getInboundProperty(HTTP_VERSION_PROPERTY);
+        return !(Protocol.HTTP_0_9.getProtocolString().equals(httpVersion) || Protocol.HTTP_1_0.getProtocolString().equals(httpVersion));
```
andresgregoire: Use org.mule.module.http.internal.domain.HttpProtocol enum instead of a class from Grizzly (if necessary, add 0.9 version to this enum)

modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseBuilder.java
```diff
@@ -113,7 +116,14 @@ public HttpResponse build(org.mule.module.http.internal.domain.response.HttpResp
             final Collection<String> paramValues = resolvedHeaders.getAll(name);
             for (String value : paramValues)
             {
-                httpResponseHeaderBuilder.addHeader(name, value);
+                if (TRANSFER_ENCODING.equals(name) && !supportsTransferEncoding(event))
+                {
+                    logger.debug("Client HTTP version is lower than 1.1 so the unsupported 'Transfer-Encoding' header has been removed and 'Content-Length' will be sent instead.");
```
andresgregoire: Add "if (logger.isDebugEnabled) ". (I know there are other calls to logger.debug in this class without this check, but they are wrong)

modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerPersistentConnectionsTestCase1_0.java
```diff
@@ -0,0 +1,81 @@
...
+import static org.junit.Assert.assertThat;
+import static org.mule.module.http.api.HttpHeaders.Values.CLOSE;
+import static org.mule.module.http.api.HttpHeaders.Values.KEEP_ALIVE;
+
+import java.io.IOException;
+
+import org.apache.http.HttpVersion;
+import org.junit.Test;
+
+public class HttpListenerPersistentConnectionsTestCase1_0 extends HttpListenerPersistentConnectionsTestCase
```
andresgregoire: Don't use an underscore in the class name, and put the protocol version before "TestCase".

modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerPersistentConnectionsTestCase1_1.java
```diff
@@ -0,0 +1,94 @@
...
+import static org.hamcrest.Matchers.nullValue;
+import static org.junit.Assert.assertThat;
+import static org.mule.module.http.api.HttpHeaders.Values.CLOSE;
+
+import java.io.IOException;
+
+import org.apache.http.HttpVersion;
+import org.junit.Test;
+
+public class HttpListenerPersistentConnectionsTestCase1_1 extends HttpListenerPersistentConnectionsTestCase
```
andresgregoire: Same as the other test class.

PR #2024, opened 2015-09-01
MULE-8807 Added the ability to configure the time between evictions for XsltTranformers

tyrulersh:
Added the ability to configure the time between evictions for the pool of transformers internally maintained by the XsltTransformer.

### DIFF
--- a/modules/xml/src/main/java/org/mule/module/xml/transformer/XsltTransformer.java
+++ b/modules/xml/src/main/java/org/mule/module/xml/transformer/XsltTransformer.java
@@ -94,6 +94,11 @@ public class XsltTransformer extends AbstractXmlTransformer
     private static final int MAX_IDLE_TRANSFORMERS = 32;
     // MAX_IDLE is also the total limit
     private static final int MAX_ACTIVE_TRANSFORMERS = MAX_IDLE_TRANSFORMERS;
+    /**
+     * A default value of -1 disables eviction functionality.
+     * @see org.apache.commons.pool.impl.GenericObjectPool#setTimeBetweenEvictionRunsMillis(long)
+     */
+    private static final long DEFAULT_TIME_BETWEEN_EVICTIONS_IN_MILLISECONDS = -1;
 
     //Saxon shipped with Mule
     public static final String PREFERRED_TRANSFORMER_FACTORY = "net.sf.saxon.TransformerFactoryImpl";
@@ -117,6 +122,8 @@ public class XsltTransformer extends AbstractXmlTransformer
         transformerPool.setMinIdle(MIN_IDLE_TRANSFORMERS);
         transformerPool.setMaxIdle(MAX_IDLE_TRANSFORMERS);
         transformerPool.setMaxActive(MAX_ACTIVE_TRANSFORMERS);
+        transformerPool.setTimeBetweenEvictionRunsMillis(
+                DEFAULT_TIME_BETWEEN_EVICTIONS_IN_MILLISECONDS);
         contextProperties = new HashMap<String, Object>();
     }
 
@@ -143,6 +150,16 @@ public class XsltTransformer extends AbstractXmlTransformer
         {
             throw new InitialisationException(te, this);
         }
+    }
+
+    /**
+     * Eviction is enabled when minEvictableIdleTimeMillis is greater than zero.
+     * @see GenericObjectPool
+     * @see GenericObjectPool#setMinEvictableIdleTimeMillis(long)
+     */
+    private boolean isTransformerPoolEvictionEnabled()
+    {
+        return getTimeBetweenTransformerEvictions() > 0;
     }
 
     /**
@@ -448,6 +465,14 @@ public class XsltTransformer extends AbstractXmlTransformer
     }
 
     /**
+     * @return The current minimum number of allowable idle transformer objects in the pool.
+     */
+    public int getMinIdleTransformers()
+    {
+        return transformerPool.getMinIdle();
+    }
+
+    /**
      * @return The current maximum number of allowable idle transformer objects in the
      *         pool
      */
@@ -464,6 +489,16 @@ public class XsltTransformer extends AbstractXmlTransformer
     public void setMaxIdleTransformers(int maxIdleTransformers)
     {
         transformerPool.setMaxIdle(maxIdleTransformers);
+    }
+
+    public long getTimeBetweenTransformerEvictions()
+    {
+        return transformerPool.getTimeBetweenEvictionRunsMillis();
+    }
+
+    public void setTimeBetweenTransformerEvictions(long timeBetweenTransformerEvictions)
+    {
+        transformerPool.setTimeBetweenEvictionRunsMillis(timeBetweenTransformerEvictions);
     }
 
     /**
--- a/modules/xml/src/main/resources/META-INF/mule-xml.xsd
+++ b/modules/xml/src/main/resources/META-INF/mule-xml.xsd
@@ -116,7 +116,7 @@ The Jaxen filter allows you to route messages based on XPath expressions. The Ja
                     </xsd:element>
                     <xsd:element name="context-property" type="spring:entryType" minOccurs="0" maxOccurs="unbounded">
                         <xsd:annotation>
-                            <xsd:documentation>A property that wil be made available to the filter context. Expression Evaluators can be used to grab these properties from the message at runtime.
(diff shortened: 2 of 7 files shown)


modules/xml/src/main/resources/META-INF/mule-xml.xsd
```diff
@@ -514,14 +514,25 @@ The Jaxen filter allows you to route messages based on XPath expressions. The Ja
...
-                            <xsd:documentation>A property that wil be made available to the transform context. Expression Evaluators can be used to grab these properties from the message at runtime.
+                            <xsd:documentation>A property that will be made available to the transform context. Expression Evaluators can be used to grab these properties from the message at runtime.
                             </xsd:documentation>
                         </xsd:annotation>
                     </xsd:element>
                 </xsd:sequence>
+                <xsd:attribute name="timeBetweenTransformerEvictions" type="mule:substitutableInt">
+                    <xsd:annotation>
+                        <xsd:documentation>
+                          Transformers are pooled for better throughput, since performing an XSL
```
elrodro83: The pool is prepopulated with the transformers so they are not constructed during message processing, but the presence of the pool does not affect the time of the actual transformation. What is the relationship of that with eviction?

elrodro83:
What is the use case for evicting transformers?

modules/xml/src/main/java/org/mule/module/xml/transformer/XsltTransformer.java
```diff
@@ -92,6 +92,13 @@
     private static final int MAX_IDLE_TRANSFORMERS = 32;
     // MAX_IDLE is also the total limit
     private static final int MAX_ACTIVE_TRANSFORMERS = MAX_IDLE_TRANSFORMERS;
+    /**
+     * A default value of -1 disables eviction functionality.
+     * @see org.apache.commons.pool.impl.GenericObjectPool#setTimeBetweenEvictionRunsMillis(long)
+     */
+    private static final long DEFAULT_TIME_BETWEEN_EVICTIONS_IN_MILLISECONDS = -1;
+
+    private static final long TRANSFORMER_POOL_EVICTION_DISABLEMENT_UPPER_BOUND = 0;
```
elrodro83: This constant is not being used

modules/xml/src/main/java/org/mule/module/xml/transformer/XsltTransformer.java
```diff
@@ -141,6 +150,23 @@ protected void doInitialise() throws InitialisationException
         {
             throw new InitialisationException(te, this);
         }
+
+        if (logger.isWarnEnabled() && !isTransformerPoolEvictionEnabled())
+        {
+            logger.warn("Transformer pool eviction is disabled.");
```
elrodro83: Why is this a warning?

modules/xml/src/main/java/org/mule/module/xml/transformer/XsltTransformer.java
```diff
@@ -141,6 +150,23 @@ protected void doInitialise() throws InitialisationException
...
+    /**
+     * Eviction is enabled when minEvictableIdleTimeMillis is greater than zero.
+     * @see GenericObjectPool
+     * @see GenericObjectPool#setMinEvictableIdleTimeMillis(long)
+     */
+    private boolean isTransformerPoolEvictionEnabled()
+    {
+        long timeBetweenTransformerEvictions = getTimeBetweenTransformerEvictions();
+        boolean transformerPoolEvictionIsEnabled = timeBetweenTransformerEvictions > 0;
+        return transformerPoolEvictionIsEnabled;
```
elrodro83: This should be a single line. The variable names are the same as the methods that are called.

modules/xml/src/main/java/org/mule/module/xml/transformer/XsltTransformer.java
```diff
@@ -456,6 +482,17 @@ public void setMaxIdleTransformers(int maxIdleTransformers)
         transformerPool.setMaxIdle(maxIdleTransformers);
     }
 
+    public long getTimeBetweenTransformerEvictions()
+    {
+        long evictionTime = transformerPool.getTimeBetweenEvictionRunsMillis();
+        return evictionTime;
```
elrodro83: This should be a single line. The variable names are the same as the methods that are called. evictionTime alone is confusing also.

modules/xml/src/main/resources/META-INF/mule-xml.xsd
tyrulersh: Where is the prepopulation taking place? The only thing that I can find that is close to prepopulation is the single call to transformerPool.addObject() in XsltTransformer.doInitialise(), which only adds a single object to the pool, but no more. Any additional objects are added *during* transformation in XsltTransformer.doTransform() upon borrowing objects from the pool.

Am I missing something?

tyrulersh:
Use case: An application with burstable traffic through various endpoints that make use of transformers. If transformers are never evicted, they remain through the life of the application in which they're used, consuming memory potentially unnecessarily, or even worse, to the point of OOM exceptions.

Note: My description above is my recollection, of my situation almost a year ago, from the top of my head. If you'd like more details, I can work on getting that.

modules/xml/src/main/java/org/mule/module/xml/transformer/XsltTransformer.java
tyrulersh: My thinking was that since transformer pools have no eviction, by deafult, users should be warned about this. However, I see now that this warning would be outputted independent of whether eviction was purposely disabled by a user or just as part of the default configuration.

modules/xml/src/main/resources/META-INF/mule-xml.xsd
elrodro83: That's partially correct. Since the maxIdle attribute of the pool is equal to its max size, the objects will be created (and added to the pool) on demand, to support up until 32 concurrent consumers. After that, those objects will remain in the pool with no need to be rebuilt.

What this approach will do, it will create a new thread which has some unwanted implications (as per https://commons.apache.org/proper/commons-pool/api-1.6/org/apache/commons/pool/impl/GenericObjectPool.html) that will create the objects up to its max size, but will then keep running periodically.

A better solution IMO for having the prepopulation as you want is to add a initialTransformersCount property to XsltTransformer and update its doInitialise method to take that into account.

tyrulersh: What are some of the unwanted implications of a thread that evicts objects?

Regarding the Pre-population comment: Yes, your stated solution for pre-population does seem better, but the subject matter of this changeset was not transformer pre-population.

elrodro83: From what i understand you want to achieve, you want two things:
1. Pre-population of the pool at startup. That's already discussed.
2. Eviction of unused objects during low-traffic moments. That's not happening even with this change, since the maxIdle is the same as maxActive by default, so even if the eviction thread is running, it will never deallocate objects from the pool. To do this you will have to also change maxIdle. Anyway, since there is a maxActive in place, the pool will not grow indefinitely and the memory consumption will be limited.

tyrulersh: The only thing I want is #2, not #1. I don't think that I've stated a desire for prepopulation, but please point it out to me if I mistakenly gave that impression.

Nevertheless, the question of whether or not eviction is happening with this changeset needs to be ironed out and that can and should be done by a test. I will write a test for eviction and either (1) add it to the changeset if it proves that the changeset allows for eviction or (2) fix the changeset to allow for eviction if the test proves the changeset doesn't do so already. This is my mistake for not including such a test from the get-go and I will rectify it. Once complete I will push the changes for further review.

tyrulersh: To the end of my previous comment, I've added `XsltTransformerPoolEvictionTestCase`. The single test in that class ensures proper eviction functionality. There was no need to change production code to get idle object eviction.

Note that I greatly despise the state of `XsltTransformerPoolEvictionTestCase` since it's difficult to digest. I've attempted to make it as straight-forward as possible using appropriate method factorization and explanatory comments. I believe it's ugliness is due mainly to the fact that it tests a necessarily mutithreaded scenario, but any and all improvements are very welcome.

tyrulersh:
I've addressed the current code review comments, so these changes are now ready for another round of review.

aiannucci:
I'm closing this PR due to inactivity and conflicts.
Please reopen if still needed.
Thank you!

PR #2027, opened 2015-09-02
MULE-8920: Remove hot swap of extension models

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/module/extension/internal/config/DeclaredConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.module.extension.internal.config;
+
+import org.mule.extension.introspection.ConfigurationModel;
+
+/**
+ * Component to hold a configuration's instance and its name.
```
pablokraan: I would say holds/groups configuration's instance related information (ie: not just the instance and the name)

modules/extensions-support/src/main/java/org/mule/module/extension/internal/introspection/AnnotationsBasedDescriber.java
```diff
@@ -306,8 +304,35 @@ private void addTypeRestrictions(ParameterDescriptor parameter, org.mule.module.
...
+
+    private void declareCapabilities(DescribingContext context, HasCapabilities<?> declaration, Function<CapabilityExtractor, Object> extractFunction)
+    {
+        List<CapabilityExtractor> extractors = context.getParameter(CAPABILITY_EXTRACTORS, List.class);
+        if (CollectionUtils.isEmpty(extractors))
+        {
+            return;
+        }
+
+        extractors.forEach(extractor -> {
```
pablokraan: We need to review how the lambdas are formatted, because this seems to be the only place where the brackets are in the same line instead of in a new one

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DescribingContextFactory.java
```diff
@@ -0,0 +1,37 @@
...
+import org.mule.extension.introspection.declaration.DescribingContext;
+import org.mule.extension.introspection.declaration.fluent.DeclarationDescriptor;
+import org.mule.module.extension.internal.DefaultDescribingContext;
+import org.mule.module.extension.spi.CapabilityExtractor;
+
+import com.google.common.collect.ImmutableList;
+
+import java.util.List;
+
+public final class DescribingContextFactory
```
pablokraan: Add javadoc

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/resolver/ConnectionArgumentResolver.java
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.module.extension.internal.runtime.resolver;
+
+import org.mule.extension.runtime.OperationContext;
+
+public class ConnectionArgumentResolver implements ArgumentResolver<Object>
```
pablokraan: Is this class needed? If yes, add javadoc and maybe could be renamed, as this looks as a NullArgumentResolver

modules/extensions-support/src/main/java/org/mule/module/extension/internal/spi/BaseCapabilityExtractor.java
```diff
@@ -0,0 +1,36 @@
...
+package org.mule.module.extension.internal.spi;
+
+import org.mule.extension.introspection.declaration.fluent.ConfigurationDescriptor;
+import org.mule.extension.introspection.declaration.fluent.DeclarationDescriptor;
+import org.mule.extension.introspection.declaration.fluent.OperationDescriptor;
+import org.mule.module.extension.spi.CapabilityExtractor;
+
+import java.lang.reflect.Method;
+
+public abstract class BaseCapabilityExtractor implements CapabilityExtractor
```
pablokraan: Javadoc

modules/extensions-support/src/main/java/org/mule/module/extension/spi/CapabilityExtractor.java
```diff
@@ -31,8 +35,12 @@
...
      *
      * @param declarationDescriptor the descriptor that is being built
-     * @param capableType the type of the {@link Capable} class
+     * @param extensionType the type of the {@link Capable} class
      * @return the extracted capability or {@code null} if none found
      */
-    Object extractCapability(DeclarationDescriptor declarationDescriptor, Class<?> capableType);
+    Object extractExtensionCapability(DeclarationDescriptor declarationDescriptor, Class<?> extensionType);
+
+    Object extractConfigCapability(ConfigurationDescriptor configurationDescriptor, Class<?> configType);
```
pablokraan: javadocs

modules/extensions-support/src/main/java/org/mule/module/extension/internal/config/DeclaredConfiguration.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/introspection/AnnotationsBasedDescriber.java
marianogonzalez: my thoughts exactly, but I think that should be part of the general reformatting of the code. I don't think there's much value in tackling this problem on its own.

Meanwhile, what I can tell you is that these lambdas are intellij formatted so at the very least we know they will be consistent across team members

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DescribingContextFactory.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/resolver/ConnectionArgumentResolver.java
marianogonzalez: This is an unfinished part of the next PR that I forgot to remove. Deleting...

modules/extensions-support/src/main/java/org/mule/module/extension/internal/spi/BaseCapabilityExtractor.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/spi/CapabilityExtractor.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/spi/BaseCapabilityExtractor.java
```diff
@@ -0,0 +1,54 @@
...
+import org.mule.extension.introspection.declaration.fluent.ConfigurationDescriptor;
+import org.mule.extension.introspection.declaration.fluent.DeclarationDescriptor;
+import org.mule.extension.introspection.declaration.fluent.OperationDescriptor;
+import org.mule.module.extension.spi.CapabilityExtractor;
+
+import java.lang.reflect.Method;
+
+/**
+ * Base implementation of {@link CapabilityExtractor} which provides null implementations for all its
+ * extract methods.
```
pablokraan: I guess this should be "abstract methods", in which case the class should not be abstract

marianogonzalez: not really. I want this class to be abstract because it makes no sense to create instances of it. The methods should not be abstract either, because what I want is for extended types to only implement the levels they need. Makes sense??

PR #2028, opened 2015-09-03
MULE-8922: Enricher broken when target expression references a key in…

pablokraan:
…side a map

### DIFF
--- a/core/src/test/java/org/mule/el/mvel/datatype/AbstractVarAssignmentDataTypePropagatorTestCase.java
+++ b/core/src/test/java/org/mule/el/mvel/datatype/AbstractVarAssignmentDataTypePropagatorTestCase.java
@@ -11,6 +11,7 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.mule.mvel2.MVEL.compileExpression;
 import static org.mule.tck.junit4.matcher.DataTypeMatcher.like;
 import static org.mule.transformer.types.MimeTypes.JSON;
+import static org.mule.transformer.types.MimeTypes.UNKNOWN;
 import org.mule.api.MuleEvent;
 import org.mule.api.transformer.DataType;
 import org.mule.api.transport.PropertyScope;
@@ -32,12 +33,15 @@ import org.mule.transformer.types.TypedValue;
 
 import java.nio.charset.StandardCharsets;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
 
 public abstract class AbstractVarAssignmentDataTypePropagatorTestCase extends AbstractMuleContextTestCase
 {
 
     public static final String CUSTOM_ENCODING = StandardCharsets.UTF_16.name();
     public static final String PROPERTY_NAME = "foo";
+    public static final String INNER_PROPERTY_NAME = "bar";
 
     private final EnricherDataTypePropagator dataTypePropagator;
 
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/mule/el/mvel/datatype/AbstractVarAssignmentDataTypePropagatorTestCase.java
```diff
@@ -53,16 +57,42 @@ protected void doAssignmentDataTypePropagationTest(PropertyScope scope, String e
...
+        assertThat(testEvent.getMessage().getPropertyDataType(PROPERTY_NAME, scope), like(Map.class, UNKNOWN, CUSTOM_ENCODING));
+    }
+
+    private CompiledExpression compileMelExpression(String expression, MuleEvent testEvent)
+    {
         final ParserConfiguration parserConfiguration = MVELExpressionLanguage.createParserConfiguration(Collections.EMPTY_MAP);
         final MVELExpressionLanguageContext context = createMvelExpressionLanguageContext(testEvent, parserConfiguration);
 
         CompiledExpression compiledExpression = (CompiledExpression) compileExpression(expression, new ParserContext(parserConfiguration));
+
```
marianogonzalez: remove unnecessary line

core/src/test/java/org/mule/el/mvel/datatype/AbstractVarAssignmentDataTypePropagatorTestCase.java
```diff
@@ -53,16 +57,42 @@ protected void doAssignmentDataTypePropagationTest(PropertyScope scope, String e
...
+        MuleEvent testEvent = getTestEvent(TEST_MESSAGE);
+        final Map<String, String> propertyValue = new HashMap<>();
+        propertyValue.put(INNER_PROPERTY_NAME, TEST_MESSAGE);
+        testEvent.getMessage().setProperty(PROPERTY_NAME, propertyValue, scope, expectedDataType);
+
+        CompiledExpression compiledExpression = compileMelExpression(expression, testEvent);
+
+        // Attempts to propagate a different dataType, which should be ignored
+        dataTypePropagator.propagate(testEvent.getMessage(), new TypedValue(propertyValue, DataType.STRING_DATA_TYPE), compiledExpression);
```
marianogonzalez: I might be wrong but this looks weird to me:

* is it ok to ignore a bad datatype? should it fail instead?
* I don't know all the "functionals" about this, but it seems to me like you're testing several things on the same method. Maybe this should be splitted into several tests?

pablokraan: There are two datatypes involved. One is the map dataType which is type of the property in the message. Another one is the String dataType which is the dataType of the value which is enriched with the map expression.
This should not fail at all, is just checking that the propagation didn't affect the dataType in the property

core/src/test/java/org/mule/el/mvel/datatype/AbstractVarAssignmentDataTypePropagatorTestCase.java
pablokraan: All empty lines are unnecessary from the code point of view, is really interesting to me to know what makes this line "unnecesary" for you. For me, this line separates parts of the test setup code to make it more readeable (for me at least).

marianogonzalez: github's diff view gave me the impression there already was an empty line. Disregard

PR #2035, opened 2015-09-07
MULE-8927 - Properly encode tokens of endpoint uris

elrodro83:
The attributes that end in and endpoint URI (for instance: smtp user and passwords) previously had to be encoded as a workaround for this issue. That workaround has to be removed when this fix is in place. 

Keep in mind that this also affects the result of MEL expressions (for instance, if an expression evaluated to a username with an @ char, it had to be encoded before using it), so a check has to be done during migration to remove all the additional encoding that was placed for workarounding this.

### DIFF
--- a/core/src/main/java/org/mule/endpoint/URIBuilder.java
+++ b/core/src/main/java/org/mule/endpoint/URIBuilder.java
@@ -6,6 +6,10 @@
  */
 package org.mule.endpoint;
 
+import static java.net.URLEncoder.encode;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.apache.commons.collections.TransformerUtils.nopTransformer;
+
 import org.mule.AbstractAnnotatedObject;
 import org.mule.api.MuleContext;
 import org.mule.api.endpoint.EndpointException;
@@ -13,8 +17,8 @@ import org.mule.api.endpoint.EndpointURI;
 import org.mule.util.ClassUtils;
 
 import java.io.File;
+import java.io.UnsupportedEncodingException;
 import java.net.URI;
-import java.net.URLDecoder;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
@@ -22,6 +26,8 @@ import java.util.Map;
 import java.util.StringTokenizer;
 import java.util.TreeMap;
 import java.util.concurrent.atomic.AtomicReference;
+
+import org.apache.commons.collections.Transformer;
 
 /**
  * This has the following logic:
@@ -68,6 +74,22 @@ public class URIBuilder extends AbstractAnnotatedObject
     public static final String[] USERHOST_ATTRIBUTES = new String[]{USER, HOST};
     // this doesn't include address, since that is handled separately (and is exclusive with these)
     public static final String[] ALL_TRANSPORT_ATTRIBUTES = new String[]{USER, PASSWORD, HOST, PORT, PATH};
+
+    protected static final Transformer URL_ENCODER = new Transformer()
+    {
+        @Override
+        public Object transform(Object input)
+        {
+            try
+            {
+                return encode((String) input, UTF_8.name());
+            }
+            catch (UnsupportedEncodingException e)
+            {
+                throw new AssertionError("UTF-8 is unknown");
+            }
+        }
+    };
 
     private String address;
     private String meta;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/mule/endpoint/URIBuilder.java
```diff
@@ -69,6 +73,22 @@
...
     public static final String[] ALL_TRANSPORT_ATTRIBUTES = new String[]{USER, PASSWORD, HOST, PORT, PATH};
 
+    protected static final Transformer URL_ENCODER = new Transformer()
+    {
+        @Override
+        public Object transform(Object input)
+        {
+            try
+            {
+                return URLEncoder.encode((String) input, "UTF-8");
```
aiannucci: You can get the UTF-8 literal from StandardCharsets class

core/src/main/java/org/mule/endpoint/URIBuilder.java
```diff
@@ -255,25 +286,25 @@ private void appendMeta(StringBuilder buffer)
         }
     }
 
-    private OrderedQueryParameters appendAddress(StringBuilder buffer)
+    private OrderedQueryParameters appendAddress(StringBuilder buffer, Transformer tokenProcessor, Transformer tokenEncoder)
```
afelisatti: Can the nopTransformer use all over be avoided somehow? Maybe by having another method that doesn't need another transformer?

core/src/test/java/org/mule/endpoint/URIBuilderTestCase.java
```diff
@@ -167,6 +170,47 @@ public void testConstructAddressWithParamsInRootPath() {
...
+
+    @Test
+    public void testConstructAddressWithSpecialURICharsInPath() throws UnsupportedEncodingException
+    {
+        URIBuilder uri = new URIBuilder();
+        uri.setProtocol("http");
+        uri.setHost("my-domain.com");
+        uri.setPath("folder%a/resource%b");
+        String result = uri.getEncodedConstructor();
+        assertEquals("http://my-domain.com/folder%a/resource%b", result);
```
afelisatti: Use hamcrest.

core/src/test/java/org/mule/endpoint/URIBuilderTestCase.java
```diff
@@ -167,6 +170,47 @@ public void testConstructAddressWithParamsInRootPath() {
...
 
+    /**
+     * MULE-6279
+     * @throws UnsupportedEncodingException 
+     */
+    @Test
+    public void testSetAddressWithSpecialURIChars() throws UnsupportedEncodingException
+    {
+        URIBuilder uri = new URIBuilder();
+        uri.setAddress("smtp://user%40my-domain.com:" + URLEncoder.encode("! \"#$%&'()*+,-./:;<=>?@[\\]_`{|}~", "UTF-8") + "@smtp.my-domain.com:25");
```
afelisatti: Use constants.

core/src/test/java/org/mule/endpoint/URIBuilderTestCase.java
```diff
@@ -167,6 +170,47 @@ public void testConstructAddressWithParamsInRootPath() {
...
+        String result = uri.getEncodedConstructor();
+        assertEquals("smtp://user%40my-domain.com:" + URLEncoder.encode("! \"#$%&'()*+,-./:;<=>?@[\\]_`{|}~", "UTF-8") + "@smtp.my-domain.com:25", result);
```
afelisatti: Isn't this the same as above? Can't you create it just once?

core/src/test/java/org/mule/endpoint/URIBuilderTestCase.java
```diff
@@ -167,6 +170,47 @@ public void testConstructAddressWithParamsInRootPath() {
...
+    public void testConstructAddressWithSpecialURIChars() throws UnsupportedEncodingException
+    {
+        URIBuilder uri = new URIBuilder();
+        uri.setProtocol("smtp");
+        uri.setUser("user@my-domain.com");
+        uri.setPassword("! \"#$%&'()*+,-./:;<=>?@[\\]_`{|}~");
+        uri.setHost("smtp.my-domain.com");
+        uri.setPort("25");
+        String result = uri.getEncodedConstructor();
+        assertEquals("smtp://user%40my-domain.com:" + URLEncoder.encode("! \"#$%&'()*+,-./:;<=>?@[\\]_`{|}~", "UTF-8") + "@smtp.my-domain.com:25", result);
```
afelisatti: Again, if this is the same you could make the whole thing a constant.

core/src/main/java/org/mule/endpoint/URIBuilder.java
elrodro83: The complexity of the transformations, having the expression evaluator in the middle, would make such thing viable only by duplicating the methods that use the transformer (by adding a version that doesn't have the transformer).

PR #2036, opened 2015-09-07
MULE-8929: Proxy validation looses XML tag when payload is body.

afelisatti:


modules/cxf/src/main/java/org/mule/module/cxf/support/ReversibleValidatingInterceptor.java
```diff
@@ -20,7 +20,7 @@ public ReversibleValidatingInterceptor()
     {
         super(Phase.READ);
         addAfter(ProxySchemaValidationInInterceptor.class.getName());
-        addAfter(StartBodyInterceptor.class.getName());
+        addBefore(StartBodyInterceptor.class.getName());
```
marcosnc: Add some comment to clarify why is this required.

modules/cxf/src/test/java/org/mule/module/cxf/ProxyValidationComparisonTestCase.java
```diff
@@ -0,0 +1,85 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+@RunWith(Parameterized.class)
+public class ProxyValidationComparisonTestCase extends FunctionalTestCase
+{
+    private static final HttpRequestOptions HTTP_REQUEST_OPTIONS = newOptions().method(POST.name()).disableStatusCodeValidation().build();
+
+    private static final String SAMPLE_REQUEST = "<soap:Envelope xmlns:soap=\"http://schemas.xmlsoap.org/soap/envelope/\">"
```
marcosnc: Make it clear that the test needs the xml to be one liner.

PR #2041, opened 2015-09-08
MULE-8931: Add the possibility of activate the profile release with t…

anosenzo:
…he property 'buildDistributions=true'


asequeira:
Why not use -Prelease?

PR #2045, opened 2015-09-11
MULE-8831 compiled expressions setters support method overloading

elrodro83:
This depends on a release of MVEL with this change: https://github.com/mulesoft/mvel/pull/14

### DIFF
--- a/tests/performance/src/test/java/org/mule/el/mvel/MVELDeepAssignPerformanceTestCase.java
+++ b/tests/performance/src/test/java/org/mule/el/mvel/MVELDeepAssignPerformanceTestCase.java
@@ -1,0 +1,132 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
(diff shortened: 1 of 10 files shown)


tests/performance/src/test/java/org/mule/el/mvel/MVELDeepAssignPerformanceTestCase.java
```diff
@@ -0,0 +1,132 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+
+public class MVELDeepAssignPerformanceTestCase extends AbstractMuleContextTestCase
+{
+    @Rule
+    public ContiPerfRule rule = new ContiPerfRule();
+
+    @Override
+    public int getTestTimeoutSecs()
```
dfeist: Need 180 seconds, this is a lot!

elrodro83: Got that value from MVELPerformanceTestCase.
Should we reduce it in all these tests?

dfeist: Don't worry it's in  tests/performance and doesn't run in normal build

PR #2047, opened 2015-09-16
MULE-8945 Exceptions in scatter-gather routes are difficult to find

elrodro83:


core/src/main/java/org/mule/work/ProcessingMuleEventWork.java
```diff
@@ -55,7 +59,7 @@ protected void doRun()
     {
         try
         {
-            this.resultEvent = this.messageProcessor.process(event);
+            this.resultEvent = messageProcessorExecutionTemplate.execute(messageProcessor, event);
```
marianogonzalez: how is this related to this PR purpose? This probably impacts every single mule task.

elrodro83: Only MuleExceptions are stored for this work instance. For others it is left null.
When getResult checks for exceptions, if it was any other than a MessageException, it would assume it run successfully and return null (both exception and resultEvent are null for this work instance) effectively swallowing the exception.

What the template does, is wrap any exception in a MessagingException (a subclass of MuleException) if necessary, so it is properly handled.

core/src/main/java/org/mule/component/DefaultComponentLifecycleAdapter.java
```diff
@@ -345,7 +345,7 @@ public Object invoke(MuleEvent event) throws MuleException
         }
         catch (Exception e)
         {
-            throw new ComponentException(RequestContext.getEvent(), component, e);
+            throw new ComponentException(MessageFactory.createStaticMessage("%s: %s", e.getClass().getName(), e.getMessage()), RequestContext.getEvent(), component, e);
```
marianogonzalez: static import

PR #2048, opened 2015-09-16
MULE-8948: CE Build fails when using JDK8 without adding the property…

anosenzo:
… -DxDocLint='-Xdoclint:none'

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2139,6 +2139,15 @@
                 <vmtype>org.eclipse.jdt.internal.launching.macosx.MacOSXType</vmtype>
             </properties>
         </profile>
+        <profile>
+            <id>jdk8+</id>
+            <activation>
+                <jdk>[1.8,)</jdk>
+            </activation>
+            <properties>
+                <xDocLint>'-Xdoclint:none'</xDocLint>
+            </properties>
+        </profile>
 
         <!-- C.I. staged build profiles -->
         <profile>


pom.xml
```diff
@@ -2139,6 +2139,15 @@
                 <vmtype>org.eclipse.jdt.internal.launching.macosx.MacOSXType</vmtype>
             </properties>
         </profile>
+        <profile>
+            <id>jdk8+</id>
+            <activation>
+                <jdk>[1.8,)</jdk>
+            </activation>
+            <properties>
+                <xDocLint>'-Xdoclint:none'</xDocLint>
```
marianogonzalez: i don't think this is the correct way to do it. If tomorrow we want to support jdk9 you need to copy this very same thing into that profile too.

This should be a compiler argument. Check the 4.x branch, we already did this

anosenzo: Maybe I am missing something but as in the activation part we use  <jdk>[1.8,)</jdk> it should work with a range from jdk8 to any newer jdk used. That is why I named the profile jdk8+. Always taking account that the newer jdks support the use of the argument -Xdoclint. 

Again maybe I am missing something but as I see in the branch 4.x we set directly the xDocLint property with the value -Xdoclint:none in the properties section. So the property will be always setted at least you pass a different value in the command line. 
That is working in mule 4.x because you are supporting only the use of jdk8 (source=1.8 in the compiler plugin). But in the case of 3.x that we support jdk7 and jdk8 we can't do this because using jdk7 with the argument -Xdoclint will throw an error.

PR #2050, opened 2015-09-17
MULE-8947 Handle unpadded base64 strings when unencoding

elrodro83:


tests/performance/src/test/java/org/mule/Base64EncodingPerformanceTestCase.java
```diff
@@ -0,0 +1,55 @@
+package org.mule;
```
marianogonzalez: missing license header. You should always run a maven build before submitting PR

tests/performance/src/test/java/org/mule/Base64EncodingPerformanceTestCase.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.io.IOException;
+
+import org.databene.contiperf.PerfTest;
+import org.databene.contiperf.Required;
+import org.databene.contiperf.junit.ContiPerfRule;
+import org.junit.Rule;
+import org.junit.Test;
+
+public class Base64EncodingPerformanceTestCase extends AbstractMuleContextTestCase
```
marianogonzalez: why is this necessary? It doesn't look like your fix would affect performance that much. The problem with this type of tests is that given our "heavy loaded" CI environment it's very likely to get flacky

elrodro83:
Removed the performance test, created https://www.mulesoft.org/jira/browse/MULE-8953 for tracking the change of the Base64 implementation.

PR #2051, opened 2015-09-17
MULE-8951: SFTP connection leak when the user doesn't have permissions.

afelisatti:


transports/sftp/src/test/java/org/mule/transport/sftp/SftpReceiverRequesterUtilTestCase.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class SftpReceiverRequesterUtilTestCase extends AbstractMuleContextTestCase
```
marianogonzalez: the prefix mock on every field is redundant

transports/sftp/src/main/java/org/mule/transport/sftp/SftpMessageReceiver.java
```diff
@@ -223,6 +223,7 @@ public MuleEvent process() throws Exception
             {
                 sftpStream.setErrorOccurred();
             }
+            throw e;
```
marianogonzalez: really? I understand that it's "common sense" for this exception to be thrown here but I'm surprised that it isn't and still works since this transport is used a lot.

Are you sure about the need/impact of doing this?

PR #2053, opened 2015-09-17
MULE-8906: Implement simple connection support in Ext-Api

marianogonzalez:


core/src/main/java/org/mule/registry/SimpleRegistry.java
```diff
@@ -77,13 +77,16 @@ protected void doRegisterObject(String key, Object object, Object metadata) thro
 
         doPut(key, object);
 
-        try
-        {
-            getLifecycleManager().applyCompletedPhases(object);
-        }
-        catch (MuleException e)
+        if (!(object instanceof MuleContext))
```
pablokraan: Why is this check required?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/ExtensionProperties.java
```diff
@@ -0,0 +1,28 @@
...
+package org.mule.module.extension.internal;
+
+import static org.mule.api.config.MuleProperties.PROPERTY_PREFIX;
+
+/**
+ * Constants for the Extensions Framework
+ *
+ * @since 4.0
+ */
+public final class ExtensionProperties
```
pablokraan: Final modifier is redundant here

modules/extensions-support/src/main/java/org/mule/module/extension/internal/introspection/ParameterDescriptor.java
```diff
@@ -65,16 +64,6 @@ void setDefaultValue(Object defaultValue)
         this.defaultValue = defaultValue;
     }
 
-    boolean isHidden()
```
pablokraan: Why is this feature removed?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/model/AbstractAnnotatedModelEnricher.java
```diff
@@ -0,0 +1,58 @@
...
+ * for enriching {@link BaseDeclaration declarations} constructed from an annotated model
+ *
+ * @see AnnotationsBasedDescriber
+ * @since 4.0
+ */
+public abstract class AbstractAnnotatedModelEnricher implements ModelEnricher
+{
+
+    /**
+     * Extracts (if present) an {@link Annotation} from class backing the {@code declaration}.
```
pablokraan: Remove the "(if present)" part as that is already documented in the returns section

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/connector/ConnectionInterceptor.java
```diff
@@ -0,0 +1,239 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Implements simple connection management by using the {@link #before(OperationContext)} phase to
+ * set a connection as parameter value of key {@link ExtensionProperties#CONNECTION_PARAM} into an
+ * {@link OperationContext}.
+ * <p/>
+ * The connection is not actually established until the first invokation to the {@link #before(OperationContext)}
```
pablokraan: invokation > invocation (is legal, but we use it with c)

core/src/main/java/org/mule/registry/SimpleRegistry.java
marianogonzalez: if this registry has completed phases is because the muleContext is already started. Don't start it again.

modules/extensions-support/src/main/java/org/mule/module/extension/internal/ExtensionProperties.java
marianogonzalez: why? I don't want anybody with the brillant idea of extending this class. What's the harm?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/introspection/ParameterDescriptor.java
marianogonzalez: not used anymore. Haven't for a while but only now I noticed. I didn't see the point at removing it in a separate PR, create an issue, etc

modules/extensions-support/src/main/java/org/mule/module/extension/internal/model/AbstractAnnotatedModelEnricher.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/connector/ConnectionInterceptor.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/connector/ConnectionInterceptor.java
```diff
@@ -0,0 +1,239 @@
...
+                }
+            }
+            finally
+            {
+                writeLock.unlock();
+            }
+        }
+    }
+
+    private void doStopHandler(boolean fail) throws MuleException
```
pablokraan: Split this method in two, stopHandler and silentlyStopHandler

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/resolver/ConfigurationArgumentResolver.java
```diff
@@ -13,19 +13,14 @@
...
- * it is exposed as a singleton
+ * it is exposed as a singleton through the {@link #INSTANCE} field
  *
  * @since 3.7.1
  */
 public final class ConfigurationArgumentResolver implements ArgumentResolver<Object>
 {
 
-    private static final ConfigurationArgumentResolver INSTANCE = new ConfigurationArgumentResolver();
-
```
pablokraan: This singleton should go away

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/resolver/ConnectorArgumentResolver.java
```diff
@@ -0,0 +1,48 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.module.extension.internal.runtime.resolver;
+
+import static org.mule.util.Preconditions.checkArgument;
+import org.mule.extension.runtime.OperationContext;
+import org.mule.module.extension.internal.ExtensionProperties;
+
+/**
+ * An implementation of {@link ArgumentResolver} which returns the value of
```
pablokraan: An implementation of {@link ArgumentResolver} which returns > Returns

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/resolver/ConnectorArgumentResolver.java
```diff
@@ -0,0 +1,48 @@
...
+ * the {@link ExtensionProperties#CONNECTION_PARAM} parameter, which is expected
+ * to have been previously set on the supplied {@link OperationContext}
+ * <p/>
+ * Because this {@link ArgumentResolver} is stateless and thread-safe,
```
pablokraan: To me these conditions does not mean that the class must be a singleton. If there is only one instance used on MethodArgumentResolverDelegate, then that class should maintain the instance instead of making this class a singleton.

modules/extensions-support/src/test/java/org/mule/module/extension/internal/runtime/connector/ConnectionInterceptorTestCase.java
```diff
@@ -0,0 +1,244 @@
...
+
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@SmallTest
```
pablokraan: In theory, a test using a real muleContext (ie: not a mock) is not a smallTest

modules/extensions-support/src/test/java/org/mule/module/extension/internal/runtime/connector/petstore/PetStoreClient.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public int getDisconnectCount()
+    {
+        return disconnectCount;
+    }
+
+    public boolean isConnected()
+    {
+        checkState(disconnectCount >= 0, "negative disconnectCount. Something is wrong");
```
pablokraan: Remove "Something is wrong", problem is already described with "negative disconnectCount"

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/connector/ConnectionInterceptor.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/resolver/ConnectorArgumentResolver.java
marianogonzalez: done

modules/extensions-support/src/test/java/org/mule/module/extension/internal/runtime/connector/ConnectionInterceptorTestCase.java
marianogonzalez: at one point I heard that it was a smalltest if it lasted less than a second which it does. But I agree with your concerns so removed.

modules/extensions-support/src/test/java/org/mule/module/extension/internal/runtime/connector/petstore/PetStoreClient.java
marianogonzalez: done

core/src/main/java/org/mule/config/builders/DefaultsConfigurationBuilder.java
```diff
@@ -106,6 +106,7 @@ protected void doConfigure(MuleContext muleContext) throws Exception
         registry.registerObject(MuleProperties.OBJECT_EXPRESSION_LANGUAGE, new MVELExpressionLanguageWrapper(muleContext));
         registry.registerObject(MuleProperties.OBJECT_CONNECTOR_MESSAGE_PROCESSOR_LOCATOR, new MuleConnectorOperationLocator());
         registry.registerObject(MuleProperties.OBJECT_TIME_SUPPLIER, new TimeSupplier());
+        registry.registerObject(MuleProperties.OBJECT_MULE_CONTEXT, muleContext);
```
dfeist: Why put muleContext into the registry? Or is it already there, and just missing from this config builder?

marianogonzalez: it is there for the SpringRegistry, I'm adding it here as well so that it's available for injection on the test cases

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/connector/ConnectionInterceptor.java
```diff
@@ -0,0 +1,239 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Implements simple connection management by using the {@link #before(OperationContext)} phase to
+ * set a connection as parameter value of key {@link ExtensionProperties#CONNECTION_PARAM} into an
+ * {@link OperationContext}.
+ * <p/>
+ * The connection is not actually established until the first invocation to the {@link #before(OperationContext)}
```
dfeist: This behaves assume a lot about the reusability and thread safety of the connection, rather than delegating these concerns to a connectionFactory/connectionHandler.

modules/extensions-support/src/main/resources/META-INF/services/org.mule.extension.introspection.declaration.spi.ModelEnricher
```diff
@@ -0,0 +1 @@
+org.mule.module.extension.internal.runtime.connector.ConnectorModelEnricher
```
dfeist: Once there are more model enrichers is it the order in this text file that determines the order in which they should be applied?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/connector/ConnectionInterceptor.java
marianogonzalez: This is for the case of simple connections, those which don't need pooling and thus the assumptions are correct.

ConnectionHandler is not responsible for those things. That would be the responsibility of the ConnectionService which is TODO and referenced in this class

modules/extensions-support/src/main/resources/META-INF/services/org.mule.extension.introspection.declaration.spi.ModelEnricher
marianogonzalez: the behavior will we the one of the standard JDK ServiceLoader. The order of the enrichers should not matter. Properly sorting the interceptors is a different story

modules/extensions-support/src/test/java/org/mule/module/extension/internal/runtime/connector/petstore/PetStoreConnectorConfig.java
```diff
@@ -0,0 +1,34 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.module.extension.internal.runtime.connector.petstore;
+
+import org.mule.extension.annotations.Configuration;
+import org.mule.extension.annotations.Parameter;
+import org.mule.extension.annotations.connector.Connector;
+
+@Configuration(name = "config")
+@Connector(PetStoreClientConnectionHandler.class)
```
dfeist: This is ok today, as this is the only parameter, but might there be other connector parameters in the future?

marianogonzalez: yes. Actually there're more in the spec, but those will be implemented in separate issues

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/connector/ConnectionInterceptor.java
dfeist: Will a single ConnectionInterceptor always be used or will there be multiple implementations.  If it's just one then even if in simple cases connection isn't pooled this implementation should support all possibilities via delegation rather than do nothing in after().  This can be worked out when other pieces of this are implemented though..

modules/extensions-support/src/main/resources/META-INF/services/org.mule.extension.introspection.declaration.spi.ModelEnricher
dfeist: Agreed. Need to think about best way to do that.  IMO the "Connector Framework" needs to own this ordering extending from anything defined by the generic interceptor framework.

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/connector/ConnectionInterceptor.java
marianogonzalez: one interceptor instance per config instance. My idea is to have the after implemented even if the connection is not pooled and the ConnectionService knowing in which cases it actually needs to do nothing.

modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/OperationContextAdapter.java
```diff
@@ -27,4 +33,31 @@
...
+     * Sets a variable of the given {@code key} and {@code value}.
+     *
+     * @param key   the variable's key. Cannot be {@code null}
+     * @param value the associated value. Cannot be {@code null}
+     * @return the value previously associated with the {@code key} or {@code null} if no such association existed.
+     */
+    Object setVariable(String key, Object value);
+
+    /**
+     * Removes the variable value assosiated with {@code key}.
```
pablokraan: assosiated > associated

PR #2057, opened 2015-09-21
MULE-8956 Properly set event in the context for blocking requests

elrodro83:


modules/http/src/main/java/org/mule/module/http/internal/request/DefaultHttpRequester.java
```diff
@@ -298,6 +299,9 @@ private MuleEvent innerProcess(MuleEvent muleEvent, boolean checkRetry) throws M
         {
             validateResponse(muleEvent);
         }
+        // Set RequestContext ThreadLocal in new thread for backwards compatibility
+        OptimizedRequestContext.unsafeSetEvent(muleEvent);
```
dfeist: The issue is caused by the HttpListener not setting the event in the RequestContext.  If the event was already set there would be nothing to do here, because i) the HttpRequester mutates the event but doesn't create a new instance ii) there isn't a thread boundary

core/src/main/java/org/mule/processor/BlockingProcessorExecutor.java
```diff
@@ -81,10 +84,11 @@ protected MuleEvent executeNext() throws MessagingException
 
         if (copyOnVoidEvent && processorMayReturnVoidEvent(processor))
         {
-            MuleEvent copy = OptimizedRequestContext.criticalSetEvent(event);
+            MuleEvent copy = ((event instanceof ThreadSafeAccess) ? new DefaultMuleEvent(new DefaultMuleMessage(event.getMessage()), event) : event);
```
dfeist: Can we find a way to add a test method for this?

modules/http/src/main/java/org/mule/module/http/internal/listener/DefaultHttpListener.java
```diff
@@ -162,7 +163,10 @@ public void responseSendSuccessfully()
 
     private MuleEvent createEvent(HttpRequestContext requestContext) throws HttpRequestParsingException
     {
-        return HttpRequestToMuleEvent.transform(requestContext, muleContext, flowConstruct, parseRequest, listenerPath);
+        MuleEvent muleEvent = HttpRequestToMuleEvent.transform(requestContext, muleContext, flowConstruct, parseRequest, listenerPath);
+        // Update RequestContext ThreadLocal for backwards compatibility
+        OptimizedRequestContext.unsafeSetEvent(muleEvent);
```
dfeist: Can we find a way to add a test method/assertion for this?

elrodro83:
Added requested test cases

PR #2058, opened 2015-09-22
MULE-8956 Properly handling event in the context for synchronous flows

elrodro83:
This is the same change as in https://github.com/mulesoft/mule/pull/2057 but in the 3.6 code wich is rather different.


elrodro83:
Added requested test cases

PR #2060, opened 2015-09-25
MULE-8844 cause of exception in AsyncUntilSuccesful available for log

elrodro83:


elrodro83:
This would also fix https://www.mulesoft.org/jira/browse/MULE-6118

elrodro83:
Have to do some changes after https://github.com/mulesoft/mule/pull/2079 is merged.

PR #2072, opened 2015-10-01
MULE-8972 Include flowstack as part of an exception message

elrodro83:


elrodro83:
https://www.mulesoft.org/jira/browse/MULE-8980 was created in order to expand this base functionality.

PR #2075, opened 2015-10-02
MULE-8969: Add a way to provide more component information for debugger

pablokraan:


core/src/main/java/org/mule/api/debug/Debuggable.java
```diff
@@ -0,0 +1,29 @@
...
+public interface Debuggable
+{
+
+    /**
+     * Returns debug information about the fields of this object.
+     *
+     * @param event event used while debugging. Non null
+     * @return a non null {@link List}
+     */
+    List<FieldDebugInfo> getDebugInfo(MuleEvent event);
```
dfeist: Would a type rather than list be better if toString() is being used, or in-case we need to add some other non-field information in the future?

core/src/main/java/org/mule/api/debug/FieldDebugInfoFactory.java
```diff
@@ -0,0 +1,190 @@
...
+        attributeEvaluationExecutors.put(Integer.class, new IntegerAttributeEvaluatorExecutor());
+        attributeEvaluationExecutors.put(Object.class, defaultAttributeEvaluationExecutor);
+    }
+
+    private interface AttributeEvaluationExecutor
+    {
+        Object evaluate(MuleEvent event, AttributeEvaluator attributeEvaluator);
+    }
+
+    private static class BooleanAttributeEvaluatorExecutor implements AttributeEvaluationExecutor
```
dfeist: Is the difference in naming deliberate or accidental? 'Evaluator' vs. 'Evaluation'.

core/src/main/java/org/mule/api/debug/FieldDebugInfoFactory.java
```diff
@@ -0,0 +1,190 @@
...
+     * Creates a debug info for a field
+     *
+     * @param name      field's name. Must be a non blank {@link String}
+     * @param type      field's type
+     * @param evaluator evaluator used to obtain the field's value
+     * @param event     event used on the evaluator to obtain the field's value
+     * @return a {@link SimpleFieldDebugInfo} if the field evaluation is successful,
+     *         and a {@link ErrorFieldDebugInfo} otherwise
+     */
+    public static FieldDebugInfo<?> createFieldDebugInfo(String name, Class type, AttributeEvaluator evaluator, MuleEvent event)
```
dfeist: Can't generics be used here?

core/src/main/java/org/mule/api/debug/SimpleFieldDebugInfo.java
```diff
@@ -0,0 +1,33 @@
...
+{
+
+    /**
+     * Creates a debug information for a simple field
+     *
+     * @param name name of the created field. Must be a not blank {@link String}
+     * @param type class of the field. Cannot be null.
+     * @param value value of the field. Can be null
+     */
+    SimpleFieldDebugInfo(String name, Class type, Object value)
```
dfeist: Can't the type uses generics here?

core/src/main/java/org/mule/api/debug/ObjectFieldDebugInfo.java
```diff
@@ -0,0 +1,36 @@
...
+
+package org.mule.api.debug;
+
+import static org.mule.util.Preconditions.checkArgument;
+
+import java.util.List;
+
+/**
+ * Provides debug information for an object's field that contains a value
+ * composed of many simple values.
```
dfeist: Not 100% sure what this means, does this means this is used to provide values for all List,Collection,Array etc?  What about maps?

core/src/main/java/org/mule/api/debug/Debuggable.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+
+package org.mule.api.debug;
```
dfeist: Not sure if this is API, then again that depends on the definition of API. It's API for debugger, but not for general Mule user.

core/src/main/java/org/mule/api/debug/Debuggable.java
pablokraan: I thought about it, but right now it only contains the field information, so is just a wrapper on a list

core/src/main/java/org/mule/api/debug/FieldDebugInfoFactory.java
pablokraan: I'll rename it

core/src/main/java/org/mule/api/debug/SimpleFieldDebugInfo.java
pablokraan: nope

core/src/main/java/org/mule/api/debug/ObjectFieldDebugInfo.java
pablokraan: explained

core/src/main/java/org/mule/api/debug/Debuggable.java
pablokraan: Is API as many different projects will depend on it (mule, debugger, devkit, extensions API)

PR #2077, opened 2015-10-05
MULE-8982: Add debug information on HTTP requester

pablokraan:


modules/http/src/main/java/org/mule/module/http/internal/request/DefaultHttpRequester.java
```diff
@@ -39,16 +44,32 @@
...
 
-public class DefaultHttpRequester extends AbstractNonBlockingMessageProcessor implements Initialisable, MuleContextAware, FlowConstructAware
+public class DefaultHttpRequester extends AbstractNonBlockingMessageProcessor implements Initialisable, MuleContextAware, FlowConstructAware, Debuggable
 {
 
     public static final List<String> DEFAULT_EMPTY_BODY_METHODS = Lists.newArrayList("GET", "HEAD", "OPTIONS");
     public static final String DEFAULT_PAYLOAD_EXPRESSION = "#[payload]";
     public static final String DEFAULT_FOLLOW_REDIRECTS = "true";
 
+    static final String URI_DEBUG = "URI";
```
dfeist: Will HttpListener be debuggable, if so some of these should be defined in a more common location ideally?

PR #2083, opened 2015-10-06
MULE-8977 Remove superfluous data from exception message

elrodro83:


core/src/main/java/org/mule/api/MuleException.java
```diff
@@ -131,17 +131,22 @@ public final String getMessage()
     protected void initialise()
     {
         setExceptionCode(ExceptionHelper.getErrorCode(getClass()));
-        String javadoc = ExceptionHelper.getJavaDocUrl(getClass());
-        String doc = ExceptionHelper.getDocUrl(getClass());
-        if (javadoc != null)
+        if(DefaultMuleConfiguration.docLinks)
```
pablokraan: There is no need to make this conditional, we don't want to include this info anymore

core/src/main/java/org/mule/api/MuleException.java
```diff
@@ -171,10 +176,6 @@ public String getVerboseMessage()
         buf.append("Type                  : ")
```
pablokraan: Is not the type also redundant here?

PR #2090, opened 2015-10-07
MULE-8988 Fix uses of isConsumable

elrodro83:


pablokraan:
On the commit message use "MULE-8988: When logging a payload as part of an exception message, show its content rather than its class" and then add a comment if you want, like "fixing compilation"

core/src/main/java/org/mule/DefaultMuleMessage.java
```diff
@@ -1752,6 +1750,15 @@ protected IllegalStateException newException(String message)
         return exception;
     }
 
+    /**
+     * @deprecated since 3.8.0. Use {@link ClassUtils#isConsumable(Class)} instead.
```
pablokraan: Maintain the original javadoc and add the deprecation info

PR #2100, opened 2015-10-09
MULE-8973: NPE in scatter-gather inside dynamically referenced sub flow

elrodro83:
This affects not only scatter-gather, but any MuleContextAware processor in the subflow.


modules/spring-config/src/test/java/org/mule/config/spring/factories/FlowRefFactoryBeanTestCase.java
```diff
@@ -205,4 +243,8 @@ private void verifyProcess(FlowRefFactoryBean flowRefFactoryBean, InitializableM
     {
     }
 
+    interface MuleContextAwareProcessor extends InitializableMessageProcessor, MuleContextAware
```
marianogonzalez: this is unnecesary. you can use mockito with extraInterfaces option

marianogonzalez:
Although I understand that the bug is not directly related to scatter-gather, consider if there's value on adding a functional test case on scatter-gather. up to you

elrodro83:
Added integration test for scatter-gather and choice, where i also experieced the issue.

PR #2104, opened 2015-10-13
MULE-9001: Paths for processors in dynamically referenced subflows are null

elrodro83:


core/src/main/java/org/mule/processor/chain/DynamicMessageProcessorContainer.java
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Specific case of a {@link MessageProcessorContainer} when its inner processors are dynamic (for instance, with a dynamically referenced subflow).
+ */
+public interface DynamicMessageProcessorContainer extends MessageProcessorContainer, MessageProcessor
+{
+
+    /**
+     * Builds the flowMap for the inner processors relative to the calling flow.
+     */
+    FlowMap doAddMessageProcessorPathElements();
```
dfeist: Simpler method name? Also "do" is typically used for protected template methods in abstract classes rather then this type of use.

core/src/main/java/org/mule/construct/AbstractPipeline.java
```diff
@@ -356,7 +355,7 @@ protected void doStart() throws MuleException
 
     private void createFlowMap()
     {
-        if (!flowMap.isEmpty())
+        if (flowMap != null)
         {
             logger.warn("flow map already populated");
```
dfeist: I wouldn't agree that this is a warning from the user perspective.  Not sure a log statement is even needed here

PR #2106, opened 2015-10-14
MULE-8982: Add debug information on HTTP requester

pablokraan:
_ Adding debug info for query params
_ Improving debug info field matchers to compare using matchers instead of simple values


modules/http/src/main/java/org/mule/module/http/internal/request/DefaultHttpRequester.java
```diff
@@ -613,11 +614,32 @@ public Object evaluate() throws Exception
                 return resolveResponseTimeout(event);
             }
         }));
+        fields.add(createFieldDebugInfo("Query Params", Object.class, getQueryParamsDebugInfo(event)));
```
elrodro83: Constant for "Query Params", as for the other fields.

elrodro83: Shouldn't second parameter be List instead of Object?

PR #2121, opened 2015-10-27
MULE-9023: Scatter-gather generates wrong data type when Content-Type…

afelisatti:
… header is present

### DIFF
--- a/tests/integration/src/test/java/org/mule/test/routing/ScatterGatherRouterTestCase.java
+++ b/tests/integration/src/test/java/org/mule/test/routing/ScatterGatherRouterTestCase.java
@@ -17,15 +17,18 @@ import static org.hamcrest.CoreMatchers.startsWith;
 import static org.hamcrest.Matchers.hasSize;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.fail;
-
+import static org.mule.api.config.MuleProperties.CONTENT_TYPE_PROPERTY;
+import org.mule.DefaultMessageCollection;
 import org.mule.api.ExceptionPayload;
 import org.mule.api.MessagingException;
 import org.mule.api.MuleEvent;
 import org.mule.api.MuleException;
+import org.mule.api.MuleMessage;
 import org.mule.api.processor.MessageProcessor;
 import org.mule.api.routing.AggregationContext;
 import org.mule.api.routing.ResponseTimeoutException;
 import org.mule.api.transport.DispatchException;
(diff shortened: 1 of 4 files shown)


tests/integration/src/test/java/org/mule/test/routing/ScatterGatherRouterTestCase.java
```diff
@@ -273,6 +277,17 @@ public void setsVariablesAfterRouting() throws Exception
         FlowAssert.verify("setsVariablesAfterRouting");
     }
 
+    @Test
+    public void returnsCorrectDataType() throws Exception
+    {
+        MuleEvent event = getTestEvent(TEST_PAYLOAD);
+        event.getMessage().setProperty(CONTENT_TYPE_PROPERTY, "application/json", PropertyScope.INBOUND);
+        MuleMessage response = runFlow("dataType", event).getMessage();
+        assertThat(response, is(Matchers.instanceOf(DefaultMessageCollection.class)));
```
marianogonzalez: static import

tests/integration/src/test/java/org/mule/test/routing/ScatterGatherRouterTestCase.java
```diff
@@ -273,6 +277,17 @@ public void setsVariablesAfterRouting() throws Exception
...
+        assertThat(((DefaultMessageCollection) response).size(), is(3));
```
marianogonzalez: use Matchers.hasSize()

tests/integration/src/test/java/org/mule/test/routing/ScatterGatherRouterTestCase.java
```diff
@@ -273,6 +277,17 @@ public void setsVariablesAfterRouting() throws Exception
...
+        assertThat(((DefaultMessageCollection) response).getMessage(0).getDataType().getMimeType(), is("text/plain"));
```
marianogonzalez: shouldn't you also test that the other 2 responses have the same mime Type you set on the original?

tests/integration/src/test/java/org/mule/test/routing/ScatterGatherRouterTestCase.java
afelisatti: That would require some transformations since it's not actually a Collection.

marianogonzalez: cool. Disregard then

PR #2129, opened 2015-10-29
MULE-9029 100-Continue response is sent as two packets instead of one…

dfeist:
… when chunked transfer encoding is disabled.

### DIFF
--- a/modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/BaseResponseCompletionHandler.java
+++ b/modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/BaseResponseCompletionHandler.java
@@ -41,6 +41,13 @@ public abstract class BaseResponseCompletionHandler extends EmptyCompletionHandl
         {
             httpResponsePacket.setChunked(true);
         }
+        else
+        {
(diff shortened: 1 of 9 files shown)


afelisatti:
Do we know if the Grizzly issue has been addressed and this fix validated?

dfeist:
I created JIRA issue and sent an email to user list, but no responses yet.

modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerExpectHeaderTestCase.java
```diff
@@ -146,4 +185,15 @@ private void assertFromStream(InputStream inputStream, String expectedInput) thr
...
 
+    protected String getExpectedResponseBody()
+    {
+        return TEST_MESSAGE;
+    }
+
+    @Parameterized.Parameters
+    public static Collection<Object[]> parameters()
+    {
+        return Arrays.asList(new Object[][] {{Boolean.TRUE.toString()},{Boolean.FALSE.toString()}});
```
afelisatti: Static imports for Boolean. Also, this should be moved to the top so the parametrisation of the test is clearer.

PR #2130, opened 2015-10-30
MULE-8916: Unclear message when more than one transformer is available

afelisatti:


core/src/main/java/org/mule/registry/TypeBasedTransformerResolver.java
```diff
@@ -153,7 +153,7 @@ protected Transformer getNearestTransformerMatch(List<Transformer> trans, Class
         {
             if (logger.isDebugEnabled())
             {
-                logger.debug("Comparing transformers for best match: input = " + input + " output = " + output + " Possible transformers = " + trans);
+                logger.debug("Comparing transformers for best match: source = " + input + " target = " + output + " Possible transformers = " + trans);
```
asequeira: Won't this print the objectId for trans instead of the content?

core/src/main/resources/META-INF/services/org/mule/i18n/core-messages.properties
```diff
@@ -264,7 +264,7 @@
 253=Mule transaction is null, but enlist method is called
 254=Object "{0}" has already been registered in the Registry. Registered object is "{1}", Object being registered is "{2}"
 255=Transformer does not implement Discoverable interface, it must do so to be registered via the Registry Bootstrap. Transformer is "{0}"
-256=There are two transformers that are an exact match for input: "{0}", output: "{1}". Transformers are: "{2}" and "{3}"
+256=There are two transformers that are an exact match for source type "{0}" and target type "{1}". Transformers are: "{2}" and "{3}". Check your transformer use to avoid implicit transformations and choose between the two options.
```
asequeira: What happens if there are more than two transformers?

core/src/main/java/org/mule/registry/TypeBasedTransformerResolver.java
afelisatti: No, AbstractTransformer overrides toString() printing all relevant data.

asequeira: ok

PR #2132, opened 2015-11-02
MULE-9023: Scatter-gather generates wrong data type when Content-Type…

pablokraan:
… header is present

### DIFF
--- a/modules/http/src/main/java/org/mule/module/http/internal/listener/HttpRequestToMuleEvent.java
+++ b/modules/http/src/main/java/org/mule/module/http/internal/listener/HttpRequestToMuleEvent.java
@@ -72,6 +72,8 @@ public class HttpRequestToMuleEvent
 
         final Map<String, DataHandler> inboundAttachments = new HashMap<>();
         Object payload = NullPayload.getInstance();
+        final String contentTypeValue = request.getHeaderValue(HttpHeaders.Names.CONTENT_TYPE);
+
         if (parseRequest)
         {
             final HttpEntity entity = request.getEntity();
@@ -83,7 +85,6 @@ public class HttpRequestToMuleEvent
                 }
                 else
                 {
-                    final String contentTypeValue = request.getHeaderValue(HttpHeaders.Names.CONTENT_TYPE);
                     if (contentTypeValue != null)
                     {
                         final MediaType mediaType = MediaType.parse(contentTypeValue);
@@ -122,6 +123,8 @@ public class HttpRequestToMuleEvent
         }
 
         final DefaultMuleMessage defaultMuleMessage = new DefaultMuleMessage(payload, inboundProperties, outboundProperties, inboundAttachments, muleContext);
+        defaultMuleMessage.setMimeType(contentTypeValue);
+
         return new DefaultMuleEvent(
                 defaultMuleMessage,
                 resolveUri(requestContext),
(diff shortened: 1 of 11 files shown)


modules/http/src/main/java/org/mule/module/http/internal/listener/HttpRequestToMuleEvent.java
```diff
@@ -122,6 +122,15 @@ else if (entity instanceof InputStreamHttpEntity)
         }
 
         final DefaultMuleMessage defaultMuleMessage = new DefaultMuleMessage(payload, inboundProperties, outboundProperties, inboundAttachments, muleContext);
+
+        for (String property : inboundProperties.keySet())
+        {
+            if (property.equalsIgnoreCase(MuleProperties.CONTENT_TYPE_PROPERTY))
```
dfeist: Use org.mule.module.http.api.HttpHeaders.Names#CONTENT_TYPE instead.

modules/http/src/main/java/org/mule/module/http/internal/listener/HttpRequestToMuleEvent.java
```diff
@@ -122,6 +122,15 @@ else if (entity instanceof InputStreamHttpEntity)
         }
 
         final DefaultMuleMessage defaultMuleMessage = new DefaultMuleMessage(payload, inboundProperties, outboundProperties, inboundAttachments, muleContext);
+
+        for (String property : inboundProperties.keySet())
```
dfeist: Why not use "request.getHeaderValue(HttpHeaders.Names.CONTENT_TYPE);" as used already on line 86.  Of move line 86 up so it's in scope?

modules/http/src/main/java/org/mule/module/http/internal/request/HttpResponseToMuleEvent.java
```diff
@@ -108,9 +108,18 @@ else if (responseContentType.startsWith(APPLICATION_X_WWW_FORM_URLENCODED.toLowe
         }
 
 
-        MuleMessage message = new DefaultMuleMessage(muleEvent.getMessage().getPayload(), inboundProperties,
+        DefaultMuleMessage message = new DefaultMuleMessage(muleEvent.getMessage().getPayload(), inboundProperties,
                                                      null, inboundAttachments, muleContext, muleEvent.getMessage().getDataType());
 
+
+        for (String property : inboundProperties.keySet())
```
dfeist: This is the second time i've seen this right after DefaultMuleMessage constructor, is there no way it can be part of the constructor logic?

modules/http/src/main/java/org/mule/module/http/internal/listener/HttpRequestToMuleEvent.java
pablokraan: Yes, refactoring the code to make use of that line is out of scope (there are 4 levels of "if" in that code, I prefer to keep them untouched)

modules/http/src/main/java/org/mule/module/http/internal/request/HttpResponseToMuleEvent.java
pablokraan: Right, this code is duplicated. I can move it to some static method, but moving it to the constructor will introduce the original problem again

PR #2140, opened 2015-11-04
MULE-9035 - Add extensions parent pom to script for deploying to mave…

asequeira:
…n central repo

### DIFF
--- a/buildtools/scripts/DeployArtifacts.groovy
+++ b/buildtools/scripts/DeployArtifacts.groovy
@@ -64,6 +64,8 @@ private OptionAccessor parseOptions(arguments)
 def deployJars()
 {
     deployToRemote(ceRepoUrl, ceRepoId, 'org.mule', 'mule', version, 'pom')
+    deployToRemote(ceRepoUrl, ceRepoId, 'org.mule', 'mule-extensions', version, 'pom')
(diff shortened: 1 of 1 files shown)


buildtools/scripts/DeployArtifacts.groovy
```diff
@@ -64,6 +64,8 @@ private OptionAccessor parseOptions(arguments)
 def deployJars()
 {
     deployToRemote(ceRepoUrl, ceRepoId, 'org.mule', 'mule', version, 'pom')
+    deployToRemote(ceRepoUrl, ceRepoId, 'org.mule', 'mule-extensions', version, 'pom')
+    deployToRemote(ceRepoUrl, ceRepoId, 'org.mule.modules', 'mule-module-validation', version, 'jar')
```
aiannucci: Is this line needed, since you are deploying all org.mule.modules artifacts in the 'for each' below? Aren't you deploying twice?

asequeira: No, because although module-module-validation groupId is org.mule.modules its Maven module is a submodule of extensions, not modules. Then it is not listed in the modules pom, nor in the extensions pom. I've done this until we have a definition about how we will handle this in Mule 4.

PR #2146, opened 2015-11-09
MULE-9044: HTTP Listener returning 500 instead of 400 on invalid Cont…

afelisatti:
…ent-Type


modules/http/src/main/java/org/mule/module/http/internal/listener/DefaultHttpListener.java
```diff
@@ -126,12 +126,14 @@ public void handleRequest(HttpRequestContext requestContext, HttpResponseReadyCa
                     final HttpMessageProcessContext messageProcessContext = new HttpMessageProcessContext(DefaultHttpListener.this, flowConstruct, config.getWorkManager(), muleContext.getExecutionClassLoader());
                     messageProcessingManager.processMessage(httpMessageProcessorTemplate, messageProcessContext);
                 }
-                catch (HttpRequestParsingException e)
+                catch (HttpRequestParsingException | IllegalArgumentException e)
                 {
+                    logger.warn("Exception occurred parsing request:", e);
```
elrodro83: Can we send a "friendly" error message along with the status code?

PR #2154, opened 2015-11-13
MULE-9051: ForEach fails to provide path elements when not initialised

elrodro83:


core/src/test/java/org/mule/routing/ForeachTestCase.java
```diff
@@ -212,6 +219,42 @@ public void testNestedIteratorPayload() throws Exception
         assertNestedProcessedMessages();
     }
     
+    @Test
+    public void addProcessorPathElementsBeforeInit() throws MuleException
+    {
+        Foreach f = new Foreach();
```
pablokraan: Rename f to something more meaningful

PR #2155, opened 2015-11-13
MULE-9055:Cannot specify my own version when creating a Mule Extension

pablocabrera85:


extensions/pom.xml
```diff
@@ -20,7 +20,9 @@
     <modules>
          <module>validation</module>
     </modules>
-
+    <properties>
```
marianogonzalez: put properties before the modules

PR #2156, opened 2015-11-14
MULE-9057: Flag when Deployment Service has finished deploying all ex…

eguser:
…isting domains and applications upon Mule startup

### DIFF
--- a/modules/launcher/src/main/java/org/mule/module/launcher/MuleDeploymentService.java
+++ b/modules/launcher/src/main/java/org/mule/module/launcher/MuleDeploymentService.java
@@ -65,6 +65,8 @@ public class MuleDeploymentService implements DeploymentService
     private final DeploymentDirectoryWatcher deploymentDirectoryWatcher;
     private DefaultArchiveDeployer<Application> applicationDeployer;
 
+    private boolean isStarted = false;
+
     public MuleDeploymentService(PluginClassLoaderManager pluginClassLoaderManager)
     {
         DomainClassLoaderRepository domainClassLoaderRepository = new MuleDomainClassLoaderRepository();
@@ -98,6 +100,7 @@ public class MuleDeploymentService implements DeploymentService
         addStartupListener(summaryDeploymentListener);
 
         deploymentDirectoryWatcher.start();
+        isStarted = true;
 
         for (StartupListener listener : startupListeners)
         {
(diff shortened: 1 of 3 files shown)


modules/launcher/src/main/java/org/mule/module/launcher/MuleDeploymentService.java
```diff
@@ -104,6 +106,7 @@ public void start()
             try
             {
                 listener.onAfterStartup();
+                this.isStarted = true;
```
asequeira: I think you should set isStarted to false when stopped.

asequeira:
Please add a test case

modules/launcher/src/main/java/org/mule/module/launcher/MuleDeploymentService.java
asequeira: Shouldn't this be set before the listeners? I mean, I would expect the isStarted to be true if I'm executing onAfterStartup() method.

modules/launcher/src/main/java/org/mule/module/launcher/MuleDeploymentService.java
```diff
@@ -118,6 +121,10 @@ public void stop()
         deploymentDirectoryWatcher.stop();
     }
 
+    public boolean isStarted() {
```
asequeira: Fix identation.

modules/launcher/src/main/java/org/mule/module/launcher/MuleDeploymentService.java
```diff
@@ -118,6 +121,10 @@ public void stop()
...
+        return this.isStarted;
```
asequeira: Remove "this"

modules/launcher/src/main/java/org/mule/module/launcher/MuleDeploymentService.java
asequeira: Remove "this"

modules/launcher/src/main/java/org/mule/module/launcher/MuleDeploymentService.java
```diff
@@ -65,6 +65,8 @@
     private final DeploymentDirectoryWatcher deploymentDirectoryWatcher;
     private DefaultArchiveDeployer<Application> applicationDeployer;
 
+    private static boolean isStarted = false;
```
asequeira: Shouldn't this be an instance variable?

eguser:
Thanks Ale,

I have implemented your suggestions and have added a test case.

eguser:
Would isReady() and "boolean deploymentServiceReady" make more sense? Is there an existing way to determine that the Deployment Service has finished operating on existing artifacts in apps and domains?

PR #2167, opened 2015-11-17
MULE-9066: Adding validation to control when set-property has an empt…

FedeAmdan:
…y propertyName


tests/integration/src/test/java/org/mule/properties/InvalidSetPropertyTestCase.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.mule.test.infrastructure.deployment.AbstractFakeMuleServerTestCase;
+import org.junit.Test;
+
+public class InvalidSetPropertyTestCase extends AbstractFakeMuleServerTestCase
+{
+    @Test
+    public void emptyPropertyName() throws Exception
+    {
+       muleServer.start();
```
marianogonzalez: starting a muleServer to test this seems overkill. Use a unit test

modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -4456,7 +4456,7 @@
     <xsd:complexType name="setPropertyType">
         <xsd:complexContent>
             <xsd:extension base="abstractAddPropertyTransformerType">
-                <xsd:attribute name="propertyName" type="attributeType" use="required"/>
+                <xsd:attribute name="propertyName" type="nonEmptyAttributeType" use="required"/>
```
marianogonzalez: if you do this you should be consistent and do it on all extensions of abstractAddPropertyTransformerType

PR #2169, opened 2015-11-18
MULE-9068 Remove transformation methods from MuleMessage

dfeist:


dfeist:
REVIEW ONLY. Tests are still running.

core/src/main/java/org/mule/DefaultMessageCollection.java
```diff
@@ -231,17 +227,16 @@ public synchronized void setPayload(Object payload, DataType dataType)
...
      *
      * {@inheritDoc}
      */
-    @Override
     public Object getPayload(Class outputType) throws TransformerException
     {
         if (invalidatedPayload)
         {
-            return super.getPayload(outputType);
+            return muleContext.getTransformationService().getPayload(this, outputType);
```
pablokraan: Why not to call the super method if the DefaulMuleMessage is invoking the transformation service?

core/src/main/java/org/mule/DefaultMessageCollection.java
```diff
@@ -262,15 +257,11 @@ public int size()
         return getMessageList().size();
     }
 
-    /**
-     * {@inheritDoc}
-     */
-    @Override
     public byte[] getPayloadAsBytes() throws Exception
```
pablokraan: is this method still required?

core/src/main/java/org/mule/DefaultMessageCollection.java
```diff
@@ -278,15 +269,11 @@ public int size()
         }
     }
 
-    /**
-     * {@inheritDoc}
-     */
-    @Override
     public String getPayloadAsString(String encoding) throws Exception
```
pablokraan: Same comments for all the methods that are still in the code but have no @Override anymore

core/src/main/java/org/mule/DefaultMuleMessage.java
```diff
@@ -492,7 +406,7 @@ public void setProperty(String key, Object value, PropertyScope scope, DataType<
         updateDataTypeWithProperty(key, value);
     }
 
-    private void setPropertyInternal(String key, Object value, PropertyScope scope, DataType<?> dataType)
+    protected void setPropertyInternal(String key, Object value, PropertyScope scope, DataType<?> dataType)
```
pablokraan: Kind of ugly to have a method for internal use only that is now protected

core/src/main/java/org/mule/DefaultMuleMessage.java
```diff
@@ -1864,4 +1488,10 @@ public int hashCode()
     {
         return properties.getOrphanFlowVariables();
     }
+
+    protected void setOriginalPayload(Object originalPayload)
```
pablokraan: Add this method to MuleMessage to avoid downcasting to DefaultMuleMessage

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,486 @@
...
+import org.mule.transformer.types.MimeTypes;
+
+import java.io.InputStream;
+import java.util.Arrays;
+import java.util.List;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+public class TransformationService
```
pablokraan: Add javadocs

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,486 @@
...
+
+    public TransformationService(MuleContext muleContext)
+    {
+        this.muleContext = muleContext;
+    }
+
+    /**
+     * Will apply a list of transformers to the payload of the message. This *Will* change the payload of the
+     * message. This method provides the only way to alter the paylaod of this message without recreating a
+     * copy of the message
```
pablokraan: Add an empty line between method and param descriptions

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,486 @@
...
+    private MuleContext muleContext;
+
+    public TransformationService(MuleContext muleContext)
+    {
+        this.muleContext = muleContext;
+    }
+
+    /**
+     * Will apply a list of transformers to the payload of the message. This *Will* change the payload of the
+     * message. This method provides the only way to alter the paylaod of this message without recreating a
```
pablokraan: I don't understand this line as the message still has the setPayload method

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,486 @@
...
+
+    private MuleContext muleContext;
+
+    public TransformationService(MuleContext muleContext)
+    {
+        this.muleContext = muleContext;
+    }
+
+    /**
+     * Will apply a list of transformers to the payload of the message. This *Will* change the payload of the
```
pablokraan: Will apply < Applies

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,486 @@
...
+     * are incompatible with the message payload
+     */
+    public void applyTransformers(MuleMessage message, MuleEvent event, List<? extends Transformer> transformers) throws MuleException
+    {
+        applyTransformers(message, event, transformers, null);
+    }
+
+    /**
+     * Will apply a list of transformers to the payload of the message. This *Will* change the payload of the
+     * message. This method provides the only way to alter the paylaod of this message without recreating a
```
pablokraan: "This method provides the only way to alter the paylaod" this is clearly not true at thre is at least another method that does the same

core/src/main/java/org/mule/TransformationService.java
pablokraan: paylaod > payload

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,486 @@
...
+     * @throws TransformerException if a transformer cannot be found or there is an error during transformation of the
+     * payload
+     */
+    public <T> T getPayload(MuleMessage message, DataType<T> outputType) throws TransformerException
+    {
+        return getPayload(message, outputType, message.getEncoding());
+    }
+
+    /**
+     * Converts the message implementation into a String representation. If encoding
```
pablokraan: Converts the payload not the message

pablokraan: String > {@link String} ?

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,486 @@
...
+     * is required it will use the encoding set on the message
+     *
+     * @return String representation of the message payload
+     * @throws Exception Implementation may throw an endpoint specific exception
+     *
+     */
+    public final String getPayloadAsString(MuleMessage message) throws Exception
```
pablokraan: Why is this method final?

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,486 @@
...
+    {
+        return getPayloadAsString(message, message.getEncoding());
+    }
+
+    /**
+     * Get the message payload for logging without throwing exception
```
pablokraan: Get > Gets

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,486 @@
...
+     * Converts the message implementation into a String representation. If encoding
+     * is required it will use the encoding set on the message
+     *
+     * @return message payload as object
```
pablokraan: as object > as a String

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,486 @@
...
+     */
+    public String getPayloadForLogging(MuleMessage message, String encoding)
+    {
+        try
+        {
+            return getPayloadAsString(message, encoding);
+        }
+        catch (Exception e)
+        {
+            return  "[Message could not be converted to string]";
```
pablokraan: string > String

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,486 @@
...
+        }
+    }
+
+    /**
+     * Get the message payload for logging without throwing exception
+     * Converts the message implementation into a String representation.
+     *
+     * @return message payload as object
+     */
+    public String getPayloadForLogging(MuleMessage message)
```
pablokraan: All these methods should clarify that the message payload could be replaced with the transformed value when the payload is consumable and that the message could be in an inconsistent state if the paylod was consume but the transformed value is not of the expected type

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,486 @@
...
+        {
+            result = ((MessageTransformer) transformer).transform(message, event);
+        }
+        else
+        {
+            result = transformer.transform(message);
+        }
+
+        // Update the RequestContext with the result of the transformation.
+        RequestContext.internalRewriteEvent(message, false);
```
pablokraan: is this needed? Seems to me that the message has no changed at this point

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,486 @@
...
+        // Pass in the message itself
+        Object result = transformer.transform(message, encoding);
+
+        // Unless we disallow Object.class as a valid return type we need to do this extra check
+        if (!resultType.getType().isAssignableFrom(result.getClass()))
+        {
+            throw new TransformerException(CoreMessages.transformOnObjectNotOfSpecifiedType(resultType, result));
+        }
+
+        // If the payload is a stream and we've consumed it, then we should set the payload on the
```
pablokraan: Seems to me that if the payload is consumable, then the new value should be set on the message before throwing the previous exception

core/src/main/java/org/mule/api/MuleContext.java
```diff
@@ -379,5 +380,9 @@
      * @since 3.8.0
      */
     Collection<ExceptionContextProvider> getExceptionContextProviders();
+
+    TransformationService getTransformationService();
```
pablokraan: Add javadocs

core/src/main/java/org/mule/routing/filters/ExpressionFilter.java
```diff
@@ -86,6 +86,13 @@ public ExpressionFilter()
         this.config = new ExpressionConfig();
     }
 
+    public ExpressionFilter(String expression, MuleContext muleContext)
```
pablokraan: Add javadoc

core/src/test/java/org/mule/tck/junit4/AbstractMuleContextTestCase.java
```diff
@@ -606,4 +607,25 @@ protected File getFileInsideWorkingDirectory(String fileName)
     {
         return new File(getWorkingDirectory(), fileName);
     }
+
+    protected String getPayloadAsString(MuleMessage message) throws Exception
```
pablokraan: Add javadocs on these methods

modules/annotations/src/main/java/org/mule/config/expression/ExpressionFilterParser.java
```diff
@@ -26,6 +27,14 @@
  */
 public class ExpressionFilterParser
 {
+
+    private MuleContext muleContext;
+
+    public ExpressionFilterParser(MuleContext muleContext)
```
pablokraan: Add javadoc

tests/functional/src/main/java/org/mule/tck/junit4/DomainFunctionalTestCase.java
```diff
@@ -107,4 +108,9 @@ public ApplicationConfig(String applicationName, String[] applicationResources)
         }
     }
 
+    protected String getPayloadAsString(MuleMessage message, MuleContext muleContext) throws Exception
```
pablokraan: Add javadoc

tests/integration/src/test/java/org/mule/test/integration/transformer/response/ResponseTransformerScenariosTestCase.java
```diff
@@ -70,7 +70,7 @@ protected String getConfigFile()
     // "request",
     // null);
     // assertNotNull(message);
-    // assertEquals("request" + "customResponse", message.getPayloadAsString());
+    // assertEquals("request" + "customResponse", getPayloadAsString(message));
```
pablokraan: Can you just remove the commented code instead of updating it?

tests/integration/src/test/resources/org/mule/properties/attachment-transformer-test-case.xml
```diff
@@ -95,7 +95,7 @@
         <expression-transformer expression="#[message.outboundAttachments['attach']]"/>
         <expression-transformer expression="#[groovy:message.getPayload().getInputStream()]"/>
         <object-to-string-transformer/>
-        <test:assert expression="#[groovy:message.getPayloadAsString().equals('some text')]"/>
```
pablokraan: What would be the way to get the payload as String now for objects that are not string?

core/src/main/java/org/mule/TransformationService.java
dfeist: Maybe, but prefer to conserve existing behaviour given this is part of a refactor. Can you create an JIRA issue for this if you think it's a bug?

tests/integration/src/test/java/org/mule/test/integration/transformer/response/ResponseTransformerScenariosTestCase.java
dfeist: This was the result of a find-replace :-)

core/src/main/java/org/mule/DefaultMessageCollection.java
dfeist: I need to remove these and move this logic to TransformationService. Missed this completely so thanks! Surprised no tests were affected..

core/src/main/java/org/mule/DefaultMuleMessage.java
dfeist: I'll move all these methods to public as part of MuleMessage interface.

core/src/main/java/org/mule/TransformationService.java
dfeist: It always was in DefaultMuleMessage it seems, I'll remove here.

core/src/main/java/org/mule/TransformationService.java
dfeist: Good idea, thanks.

tests/integration/src/test/resources/org/mule/properties/attachment-transformer-test-case.xml
dfeist: Using the transformation service.

modules/annotations/src/main/java/org/mule/config/expression/ExpressionFilterParser.java
dfeist: Not sure what javadoc is required "Creates a new instance of ExpressionFilterParse is implicit/obvious" especially given there is only a single constructor.

core/src/main/java/org/mule/DefaultMuleEvent.java
```diff
@@ -640,7 +640,7 @@ public String getMessageAsString(String encoding) throws MuleException
     {
         try
         {
-            return message.getPayloadForLogging(encoding);
+            return getMuleContext().getTransformationService().getPayloadAsString(message, encoding);
```
pablokraan: Why getPayloadAsString instead of getPayloadForLogging?

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,513 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+/**
+ * Provides the same operations previously exposed by {@link MuleMessage} but decoupled from MuleMessage.
+ *
+ * TODO Redefine this interface as part of Mule 4.0 transformation improvements.
```
pablokraan: Add a reference to the JIRA issue that will fix this TODO item

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,513 @@
...
+        applyTransformers(message, event, Arrays.asList(transformers), null);
+    }
+
+    /**
+     * Applies a list of transformers to the payload of the message. This *Will* change the payload of the
+     * message.
+     *
+     * @param event the event being processed
+     * @param transformers the transformers to apply to the message payload
+     * @param outputType the required output type for this transformation. by adding this parameter some additional
```
pablokraan: by adding this > By adding this

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,513 @@
...
+     * transformations will occur on the message payload to ensure that the final payload is of the specified type.
+     * If no transformers can be found in the registry that can transform from the return type of the transformation
+     * list to the outputType and exception will be thrown
+     * @throws TransformerException if a transformation error occurs or one or more of the transformers passed in a
+     * are incompatible with the message payload
+     */
+    public void applyTransformers(MuleMessage message, MuleEvent event, List<? extends Transformer> transformers, Class<?> outputType) throws MuleException
+    {
+        if (!transformers.isEmpty())
```
pablokraan: If there are no transformers, this method should fail

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,513 @@
...
+            message.setPayload(getPayload(message, DataTypeFactory.create(outputType)));
+        }
+    }
+
+
+    /**
+     * Attempts to obtain the payload of this message with the desired Class type. This will
+     * try and resolve a transformer that can do this transformation. If a transformer cannot be found
+     * an exception is thrown.  Any transformers added to the registry will be checked for compatibility.
+     *
```
pablokraan: Add \<p/> to separate the text instead of using empty lines

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,513 @@
...
+        {
+            MuleMessageCollection messageCollection = (MuleMessageCollection) message;
+            if (messageCollection.isInvalidatedPayload())
+            {
+                return getPayload(messageCollection, outputType);
+            }
+            else
+            {
+                List<T> results = new ArrayList<T>(messageCollection.getMessageList().size());
+                for (MuleMessage messageInList : messageCollection.getMessageList())
```
pablokraan: Ty to use lambdas :)

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,513 @@
...
+            }
+        }
+        else
+        {
+            return getPayload(message, DataType.BYTE_ARRAY_DATA_TYPE, message.getEncoding());
+        }
+    }
+
+    /**
+     * Get the message payload for logging without throwing exception
```
pablokraan: Get the message > Gets the message

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,513 @@
...
+        String encoding = transformed.getEncoding() == null ? message.getEncoding() : transformed.getEncoding();
+        Class<?> type = transformed.getType() == Object.class ? original.getType() : transformed.getType();
+
+        DataType mergedDataType = DataTypeFactory.create(type, mimeType);
+        mergedDataType.setEncoding(encoding);
+        return mergedDataType;
+    }
+
+    /**
+     * Will attempt to obtain the payload of this message with the desired Class type. This will
```
pablokraan: Will attempt to obtain > Attempts to obtain

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,513 @@
...
+     * @throws Exception Implementation may throw an endpoint specific exception
+     *
+     */
+    public String getPayloadAsString(MuleMessage message) throws Exception
+    {
+        return getPayloadAsString(message, message.getEncoding());
+    }
+
+    /**
+     * Converts the message implementation into a String representation
```
pablokraan: Converts the payload not the message. Fix all the javadocs, not just this one

core/src/main/java/org/mule/api/MuleContext.java
```diff
@@ -379,5 +380,21 @@
...
+     * {@link org.mule.api.transformer.Transformer}'s to a {@link MuleMessage} and for obtaining different
+     * representations of the message payload.
+     *
+     * @return transformation service
+     */
+    TransformationService getTransformationService();
+
+    /**
+     * Sets application wide instance of {@link TransformationService}
+     * @param transformationService transformation service instance
```
pablokraan: Add empty line between method and param descriptions

core/src/main/java/org/mule/DefaultMuleEvent.java
dfeist: Because this looks like an obvious bug/inconsistency, especially given this method throws an exception.

modules/annotations/src/main/java/org/mule/config/expression/ExpressionFilterParser.java
pablokraan: Is obvious that it creates an ExpressionFilterParser instance, but what about the context? is required or can be null?

tests/integration/src/test/resources/org/mule/properties/attachment-transformer-test-case.xml
pablokraan: I mean, how can you do the same using groovy. In this case, instead of returning the value (which is a String) it could be good to show how to use the transformation service

core/src/main/java/org/mule/TransformationService.java
dfeist: Ideally, but i'd prefer not to change transformation behavior as part of this refactor.  This will be looked at as part of MULE-9141 surely.

core/src/main/java/org/mule/TransformationService.java
pablokraan: MULE-9142

core/src/main/java/org/mule/api/MuleContext.java
```diff
@@ -379,5 +380,21 @@
      * @since 3.8.0
      */
     Collection<ExceptionContextProvider> getExceptionContextProviders();
+
+    /**
+     * Gets application wide instances of {@link TransformationService} used for applying
```
pablokraan: wide instances of > wide instance of

core/src/main/java/org/mule/api/MuleMessage.java
```diff
@@ -573,4 +458,25 @@
      * {@link org.mule.api.transport.PropertyScope#OUTBOUND}.
      */
     void clearAttachments();
+
+    /**
+     * Update the value of the message originalPayload.  This is typically updating automatically but there are some
```
pablokraan: If possible, always start the javadoc with a verb (update > updates).
Fix all javadoc formats

core/src/main/java/org/mule/TransformationService.java
dfeist: I think you mean streams, I could but the fact is that it's not possible when map() operations throw checked exceptions.

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -0,0 +1,514 @@
...
+     * compatibility.
+     *
+     * @param resultType the desired return type
+     * @param encoding   the encoding to use if required
+     * @return The converted payload of this message. Note that this method will not alter the
+     *         payload of this message <b>unless</b> the payload is an {@link InputStream} in which
+     *         case the stream will be read and the payload will become the fully read stream.
+     * @throws TransformerException if a transformer cannot be found or there is an error during
+     *                              transformation of the payload.
+     * @since 3.0
```
pablokraan: Sorry I didn't see this before. Remove @Since 3.0

PR #2172, opened 2015-11-18
MULE-9075: Remove RC4 cipher suites from tls-default.conf

afelisatti:


asequeira:
What happens when enabledCipherSuites is not defined (the default configuration)? Would RC4 be allowed?

afelisatti:
That depends on the default enabled cipher suites in the VM. This is already being addressed by Oracle though: http://bugs.java.com/view_bug.do?bug_id=8076221

PR #2173, opened 2015-11-18
MULE-9066: Adding validation when set-property and set-variable have …

FedeAmdan:
…an empty key


core/src/main/java/org/mule/transformer/simple/AbstractAddVariablePropertyTransformer.java
```diff
@@ -90,9 +90,9 @@ public Object clone() throws CloneNotSupportedException
 
     public void setIdentifier(String identifier)
     {
-        if (identifier == null)
+        if (identifier == null || StringUtils.isBlank(identifier))
```
marianogonzalez: this is redundant. Stringtils.isBlack() already accounts for the value being null

core/src/main/java/org/mule/transformer/simple/AbstractAddVariablePropertyTransformer.java
```diff
@@ -90,9 +90,9 @@ public Object clone() throws CloneNotSupportedException
...
         {
-            throw new IllegalArgumentException("Key must not be null");
+            throw new IllegalArgumentException("Key must not be null or empty");
```
marianogonzalez: rephrase as Key cannot be blank

tests/integration/src/test/java/org/mule/properties/InvalidSetVariableTestCase.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.mule.api.MuleContext;
+import org.mule.api.config.ConfigurationException;
+import org.mule.config.spring.SpringXmlConfigurationBuilder;
+import org.mule.context.DefaultMuleContextFactory;
+import org.mule.tck.junit4.AbstractMuleTestCase;
+
+import org.junit.Test;
+
+public class InvalidSetVariableTestCase extends AbstractMuleTestCase
```
marianogonzalez: this test is exactly the same as the one before. Have only one which uses @Parametrized to test each scenario

PR #2174, opened 2015-11-19
MULE-9085: MuleDocumentLoader must generate a level 3 DOM tree

elrodro83:


modules/spring-config/src/main/java/org/mule/config/spring/MuleDocumentLoader.java
```diff
@@ -43,30 +32,27 @@
...
-    /**
-     * JAXP attribute used to configure the schema language for validation.
-     */
-    private static final String SCHEMA_LANGUAGE_ATTRIBUTE = "http://java.sun.com/xml/jaxp/properties/schemaLanguage";
-
-    /**
-     * JAXP attribute value indicating the XSD schema language.
-     */
-    private static final String XSD_SCHEMA_LANGUAGE = "http://www.w3.org/2001/XMLSchema";
+    private static final UserDataHandler NOP_DATA_HANDLER = new UserDataHandler()
```
marianogonzalez: rename as NULL_DATA_HANDLER

modules/spring-config/src/main/java/org/mule/config/spring/MuleDocumentLoader.java
```diff
@@ -43,30 +32,27 @@
...
- * Alternative to Spring's default document loader that uses <b>SAX</b> instead of <b>DOM</b> to parse the bean
- * definitions.
- * <p/>
- * Additionally, the elements in the parsed elements are augmented with metadata annotations.
+ * Alternative to Spring's default document loader that uses <b>SAX</b> to add metadata to the <b>DOM</b> elements that
+ * are the result of the default parser.
  * 
  * @since 3.8.0
  */
 public class MuleDocumentLoader implements DocumentLoader
```
marianogonzalez: all classes should be final until proven otherwise

marianogonzalez: if possible, reduce this class visibility

modules/spring-config/src/main/java/org/mule/config/spring/parsers/generic/TextDefinitionParser.java
```diff
@@ -27,16 +27,36 @@
  */
 public class TextDefinitionParser extends ChildDefinitionParser
```
marianogonzalez: same as before... make final and try to reduce visibility

tests/integration/src/test/java/org/mule/test/config/ConfigurationAnnotationsTestCase.java
```diff
@@ -83,8 +83,7 @@ public void testFlowAnnotations()
         assertThat(getDocDescription(flow), is("Main flow"));
         assertThat(getSourceFile(flow), is("annotations.xml"));
         assertThat(getSourceFileLine(flow), is(10));
-        // initialState is added by the xsd with a default value
-        assertThat(getSourceElement(flow), is("<flow name=\"Bridge\" doc:name=\"Bridge flow\" initialState=\"started\">" + SystemUtils.LINE_SEPARATOR +
+        assertThat(getSourceElement(flow), is("<flow name=\"Bridge\" doc:name=\"Bridge flow\">" + SystemUtils.LINE_SEPARATOR +
```
marianogonzalez: why this change? seems like the only difference is that you removed the initialState attribute. WHat's the reason for this?

elrodro83: The attribute that was removed from this assertion is not present in the source XML but was added by the schema when the transformer was generating the DOM tree. Now that we don't use the schema to set the metadata (the schema is used by Spring default's parser) we can get the actual attributes from the tag in the source xml.

modules/spring-config/src/main/java/org/mule/config/spring/parsers/generic/TextDefinitionParser.java
elrodro83: This is just a rollback of a change that is no longer necessary with these parser changes.

PR #2186, opened 2015-11-23
MULE-8958: Allow insecure HTTPS connections

afelisatti:


afelisatti:
OAuth module also uses the TlsContext, but it passes it along to a requester so the requester test should cover it.

PR #2191, opened 2015-11-24
MULE-9103 - Decouple poll message source from transport infrastructure

pablolagreca:


core/src/main/java/org/mule/transport/polling/PollingMessageSource.java
```diff
@@ -0,0 +1,392 @@
...
+
+            @Override
+            public void stop() throws MuleException
+            {
+                PollingMessageSource.this.stop();
+            }
+        }, muleContext.getExceptionListener());
+    }
+
+    public TimeUnit getTimeUnit()
```
dfeist: Not used?

core/src/main/java/org/mule/transport/polling/PollingMessageSource.java
```diff
@@ -0,0 +1,392 @@
...
+     * The Polling name identifier. Used to create the scheduler name
+     */
+    public static final String POLLING_SCHEME = "polling";
+    /**
+     * Format string for all the Polling Schedulers name.
+     */
+    private static final String POLLING_SCHEDULER_NAME_FORMAT = POLLING_SCHEME + "://%s/%s";
+    private final SchedulerFactory<Runnable> schedulerFactory;
+    public static final long DEFAULT_POLL_FREQUENCY = 1000;
+    public static final TimeUnit DEFAULT_POLL_TIMEUNIT = TimeUnit.MILLISECONDS;
```
dfeist: Move both of the above to FixedFrequencySchedulerFactory?

PR #2201, opened 2015-11-24
MULE-9150: Extracting the configuration resources of the SpringXmlCon…

FedeAmdan:
…figurationBuilder class


marianogonzalez:
the commit message is wrong. It should not only reference the jira issue, it should also have the same title. Please make those match

PR #2202, opened 2015-11-24
MULE-9145: modifying CheckRequiredAttributesWhenNoChildren pre-processor to take the NS into account.

bayitajesi:
Changed all usages of this pre-processor and added test.

### DIFF
--- a/modules/spring-config/src/main/java/org/mule/config/spring/parsers/processors/CheckRequiredAttributesWhenNoChildren.java
+++ b/modules/spring-config/src/main/java/org/mule/config/spring/parsers/processors/CheckRequiredAttributesWhenNoChildren.java
@@ -17,6 +17,14 @@ import org.w3c.dom.Element;
 public class CheckRequiredAttributesWhenNoChildren extends CheckRequiredAttributes
 {
     private String elementName;
+    private String elementNamespaceUrl;
+
+    public CheckRequiredAttributesWhenNoChildren(String[][] attributeNames, String elementName, String elementNamespaceUrl)
+    {
+        super(attributeNames);
+        this.elementName = elementName;
+        this.elementNamespaceUrl = elementNamespaceUrl;
+    }
 
     public CheckRequiredAttributesWhenNoChildren(String[][] attributeNames, String elementName)
     {
@@ -27,7 +35,7 @@ public class CheckRequiredAttributesWhenNoChildren extends CheckRequiredAttribut
     public void preProcess(PropertyConfiguration config, Element element)
     {
         // If there are child elements we skip this check
-        if (element.getElementsByTagName(elementName).getLength() > 0)
+        if (element.getElementsByTagNameNS(elementNamespaceUrl, elementName).getLength() > 0)
         {
             return;
         }
--- a/modules/spring-config/src/test/java/org/mule/config/spring/parsers/processors/CheckRequiredAttributesTestCase.java
+++ b/modules/spring-config/src/test/java/org/mule/config/spring/parsers/processors/CheckRequiredAttributesTestCase.java
@@ -223,7 +223,8 @@ public class CheckRequiredAttributesTestCase extends AbstractPreProcessorTestCas
         assertOk(groups, "from id name type");
     }
 
-    protected PreProcessor createCheck(String[][] constraint)
+    @Override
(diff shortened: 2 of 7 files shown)


modules/spring-config/src/test/java/org/mule/config/spring/parsers/processors/CheckRequiredAttributesWhenNoChildrenTestCase.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * (c) 2003-2015 MuleSoft, Inc. This software is protected under international copyright
```
marianogonzalez: wrong header, this is the EE header but this PR is going to the CE project. Run mvn license:format on the root of your branch to fix

modules/spring-config/src/main/java/org/mule/config/spring/parsers/processors/CheckRequiredAttributesWhenNoChildren.java
```diff
@@ -17,17 +17,19 @@
 public class CheckRequiredAttributesWhenNoChildren extends CheckRequiredAttributes
 {
     private String elementName;
+    private String elementNamespaceUrl;
 
-    public CheckRequiredAttributesWhenNoChildren(String[][] attributeNames, String elementName)
+    public CheckRequiredAttributesWhenNoChildren(String[][] attributeNames, String elementName, String elementNamespaceUrl)
```
marianogonzalez: it is not ideal to force all uses of this constructor to change for a functionality you only need just now (presumably in just one place). Either add a constructor which preserves the old behaviour or create a new subclass of CheckRequiredAttributesWhenNoChildren which is namespace aware. I'm open to other solutions as long a they don't have an impact as big as this one

marianogonzalez:
the commit message should not only reference the jira issue, it should also have the same title. Please make those match

modules/spring-config/src/test/java/org/mule/config/spring/parsers/processors/CheckRequiredAttributesTestCase.java
```diff
@@ -223,7 +223,7 @@ public void testRealWorld() throws ParserConfigurationException
         assertOk(groups, "from id name type");
     }
 
-    protected PreProcessor createCheck(String[][] constraint)
+    protected PreProcessor createCheck(String[][] constraint, String elementName, String elementNamespaceUrl)
```
marianogonzalez: is this a method override? If yes, then add @Override... Otherwise why are you sending two attributes you don't use?

modules/spring-config/src/main/java/org/mule/config/spring/parsers/processors/CheckRequiredAttributesWhenNoChildren.java
```diff
@@ -27,7 +35,7 @@ public CheckRequiredAttributesWhenNoChildren(String[][] attributeNames, String e
     public void preProcess(PropertyConfiguration config, Element element)
     {
         // If there are child elements we skip this check
-        if (element.getElementsByTagName(elementName).getLength() > 0)
+        if (element.getElementsByTagNameNS(elementNamespaceUrl, elementName).getLength() > 0)
```
marianogonzalez: shouldn't you check elementNamespaceUrl to be not null?

bayitajesi: getElementsByTagNameNS allows null values and behaves as getElementsByTag

modules/spring-config/src/test/java/org/mule/config/spring/parsers/processors/CheckRequiredAttributesTestCase.java
bayitajesi: Yes, you are right, I forgot the @Override. Will change it.

modules/spring-config/src/main/java/org/mule/config/spring/parsers/processors/CheckRequiredAttributesWhenNoChildren.java
marianogonzalez: you win this time....

PR #2207, opened 2015-11-25
MULE-9107 - Remove servlet transport

pablolagreca:


modules/cxf/pom.xml
```diff
@@ -732,12 +732,6 @@
         </dependency>
         <dependency>
             <groupId>org.mule.transports</groupId>
-            <artifactId>mule-transport-servlet</artifactId>
```
marianogonzalez: looks like you didn't rebase this branch

PR #2212, opened 2015-11-25
MULE-9108 - Remove email transport

pablolagreca:


tools/schemadocs/src/main/resources/mule-30-index.xml
```diff
@@ -1,12 +1,5 @@
 <index version="3.0">
     <transport dist="ce">bpm</transport>
-    <transport dist="ce">email</transport>
```
marianogonzalez: i'll merge it but it looks like this ifle should go all together

PR #2214, opened 2015-11-26
MULE-9069: ExecutionInterceptor causes a 50% regression in the proxy

elrodro83:
scenario.

### DIFF
--- a/core/src/main/java/org/mule/context/notification/MessageProcessingFlowTraceManager.java
+++ b/core/src/main/java/org/mule/context/notification/MessageProcessingFlowTraceManager.java
@@ -14,18 +14,25 @@ import org.mule.api.context.notification.FlowStackElement;
 import org.mule.api.context.notification.FlowTraceManager;
 import org.mule.api.context.notification.ProcessorsTrace;
 import org.mule.api.execution.LocationExecutionContextProvider;
+import org.mule.api.lifecycle.Disposable;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/mule/config/DefaultMuleConfiguration.java
```diff
@@ -424,6 +427,26 @@ protected void validateXML() throws FatalException
         }
     }
 
+    /**
+     * Register a callback to be called when the logging configuration is changed.
+     * 
+     * @param listener the callback to execute
+     */
+    public static void registerLogConfigChangeListener(PropertyChangeListener listener)
```
dfeist: MuleConfiguration shouldn't expose behaviour, rather just expose configuration values.  See what others think maybe, but i'd prefer these methods were in a LogUtils than here.

core/src/main/java/org/mule/config/DefaultMuleConfiguration.java
```diff
@@ -424,6 +427,26 @@ protected void validateXML() throws FatalException
...
+    {
+        ((LoggerContext) LogManager.getContext(false)).addPropertyChangeListener(listener);
```
dfeist: Logging (in general at least) in Mule is not directly dependent on Log4J, even if it is used by default.  So this either needs to use the logging abstraction, or handle the case where log4j is not being used.

PR #2215, opened 2015-11-26
MULE-8873: Remove file transport

pablokraan:


tests/integration/src/test/java/org/mule/test/integration/ServerTools.java
```diff
@@ -1,187 +0,0 @@
...
-import org.apache.tools.ant.taskdefs.Java;
-import org.apache.tools.ant.types.Environment;
-import org.apache.tools.ant.types.Path;
-import org.apache.tools.ant.util.Watchdog;
-
-/**
- * Will start external test servers needed for the integration tests
- */
-
-public class ServerTools
```
marianogonzalez: looks like this can be useful for JMS testing and stuff. Why do you include it in this removal?

pablokraan: This class is not used at all, even on 3.x branch

PR #2222, opened 2015-11-30
MULE-9158 Change TransformationService to not mutate MuleMessage

dfeist:
This is part 1 of this JIRA issue.  Part 2 relates to single remaining mutation when the payload is consumable.

### DIFF
--- a/core-tests/src/test/java/org/mule/MuleMessageDataTypePropagationTestCase.java
+++ b/core-tests/src/test/java/org/mule/MuleMessageDataTypePropagationTestCase.java
@@ -190,9 +190,10 @@ public class MuleMessageDataTypePropagationTestCase extends AbstractMuleTestCase
 
         MuleEvent muleEvent = mock(MuleEvent.class);
 
-        transformationService.applyTransformers(muleMessage, muleEvent, Collections.singletonList(transformer));
+        MuleMessage result = transformationService.applyTransformers(muleMessage, muleEvent,
+                                                                                   Collections.singletonList(transformer));
 
-        assertDataType(muleMessage, Integer.class, APPLICATION_XML, DEFAULT_ENCODING);
+        assertDataType(result, Integer.class, APPLICATION_XML, DEFAULT_ENCODING);
     }
 
     @Test
@@ -209,9 +210,10 @@ public class MuleMessageDataTypePropagationTestCase extends AbstractMuleTestCase
 
         MuleEvent muleEvent = mock(MuleEvent.class);
 
-        transformationService.applyTransformers(muleMessage, muleEvent, Collections.singletonList(transformer));
+        MuleMessage result = transformationService.applyTransformers(muleMessage, muleEvent,
+                                                                                   Collections.singletonList(transformer));
 
-        assertDataType(muleMessage, Integer.class, APPLICATION_XML, CUSTOM_ENCODING);
+        assertDataType(result, Integer.class, APPLICATION_XML, CUSTOM_ENCODING);
     }
 
     @Test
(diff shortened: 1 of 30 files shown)


core-tests/src/test/java/org/mule/MuleMessageDataTypePropagationTestCase.java
```diff
@@ -190,7 +190,8 @@ public void updatesTypeOnTransformation() throws Exception
 
         MuleEvent muleEvent = mock(MuleEvent.class);
 
-        transformationService.applyTransformers(muleMessage, muleEvent, Collections.singletonList(transformer));
+        muleMessage = (DefaultMuleMessage) transformationService.applyTransformers(muleMessage, muleEvent,
```
pablokraan: Can you define the message as MuleMessage to avoid the cast?

core/src/main/java/org/mule/DefaultMuleMessage.java
```diff
@@ -219,7 +219,11 @@ private DefaultMuleMessage(Object message, MuleMessage previous, MuleContext mul
         }
 
         originalPayload = previous.getPayload();
-        setEncoding(previous.getEncoding());
+
+        if(this.dataType.getEncoding() == null)
```
pablokraan: Fix code style

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -56,9 +56,9 @@ public TransformationService(MuleContext muleContext)
      * @throws TransformerException if a transformation error occurs or one or more of the transformers passed in a
      * are incompatible with the message payload
      */
-    public void applyTransformers(MuleMessage message, MuleEvent event, List<? extends Transformer> transformers) throws MuleException
+    public MuleMessage applyTransformers(final MuleMessage message, final MuleEvent event, final List<? extends Transformer> transformers) throws MuleException
```
pablokraan: Javadocs must be updated

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -429,19 +427,29 @@ private void transformMessage(MuleMessage message, MuleEvent event, Transformer
                 // instance
                 synchronized (message)
                 {
-                    MuleMessage resultMessage = (MuleMessage) result;
-                    message.setPayload(resultMessage.getPayload(), resultMessage.getDataType());
-                    message.setOriginalPayload(resultMessage.getOriginalPayload());
-                    message.copyMessageProperties(resultMessage);
-                    message.copyAttachments(resultMessage);
+                    MuleMessage transformResult = (MuleMessage) result;
```
pablokraan: Looks like the original message was already mutated here

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -429,19 +427,29 @@ private void transformMessage(MuleMessage message, MuleEvent event, Transformer
...
+            resultMessage = new DefaultMuleMessage(result, message, muleContext, mergedDataType);
         }
+
+        if (message.getOriginalPayload() == null && muleContext.getConfiguration().isCacheMessageOriginalPayload())
+        {
+            resultMessage.setOriginalPayload(message.getPayload());
+        }
+
+        // Update the RequestContext with the result of the transformation.
+        RequestContext.internalRewriteEvent(resultMessage, false);
```
pablokraan: This means that the original message can't be used anymore (unless is set again on the RequestContext). Should not be that part properly documented?

core-tests/src/test/java/org/mule/MuleMessageDataTypePropagationTestCase.java
dfeist: No, because mimeType is set above and the setter is only available on MuleMessage.  I could define a second message instance for the transformation result though if you think that is best.

core/src/main/java/org/mule/TransformationService.java
dfeist: This was always here, I just moved it down to after construction of new message instance.

core/src/main/java/org/mule/TransformationService.java
dfeist: I think this can be simplified now as the DefaultMuleMessage constructor does this.

core/src/main/java/org/mule/TransformationService.java
dfeist: done

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -48,36 +48,35 @@ public TransformationService(MuleContext muleContext)
     }
 
     /**
-     * Applies a list of transformers to the payload of the message. This *Will* change the payload of the
-     * message.
+     * Applies a list of transformers returning the result of the transformation as a new message instance.
```
pablokraan: Regarding the "as a new message instance" part. Is a requirement that this method always return a new instance? What if the requested transformation is not required because the message payload is already an instance of the output type?

PR #2224, opened 2015-11-30
MULE-7663: tls-default.conf entries are ignored sometimes (fixing cli…

afelisatti:
…ent side and adding tests)


afelisatti:
Moved the TlsRestrictedProtocolsAndCiphers test cases to integration to reuse the logic for the HTTP transport tests. Updated TLS stores and removed unnecessary tls-default.conf file. Branch has executed successfully in CI.

PR #2225, opened 2015-12-01
MULE-9166: Move tests from mule core to a separated modules

pablokraan:
_ Fixing tests

### DIFF
--- a/modules/cxf/pom.xml
+++ b/modules/cxf/pom.xml
@@ -667,6 +667,13 @@
 
         <!-- Unit tests -->
         <dependency>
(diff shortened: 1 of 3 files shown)


modules/cxf/pom.xml
```diff
@@ -667,6 +667,13 @@
 
         <!-- Unit tests -->
         <dependency>
+            <groupId>org.mule</groupId>
```
dfeist: I think the client module needs this dependency too doesn't it? Anything that uses the test transport right?

PR #2233, opened 2015-12-09
MULE-9182: CXF proxy trying to send response for one-way operations

elrodro83:


modules/http/src/main/java/org/mule/module/http/internal/listener/HttpMessageProcessorTemplate.java
```diff
@@ -100,6 +101,11 @@ public void sendResponseToClient(MuleEvent muleEvent, ResponseCompletionCallback
         }
         catch (Exception e)
         {
+            HttpResponse errorResponse = responseBuilder
```
afelisatti: Why was this necessary?

PR #2234, opened 2015-12-09
MULE-8287: Allow cipher suite and protocol configuration per connector

afelisatti:
REVIEW ONLY. DO NOT MERGE: tests are still running.

### DIFF
--- a/modules/http/src/test/java/org/mule/module/http/functional/HttpRestrictedCiphersAndProtocolsTestCase.java
+++ b/modules/http/src/test/java/org/mule/module/http/functional/HttpRestrictedCiphersAndProtocolsTestCase.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.module.http.functional;
+
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.isA;
+import static org.junit.Assert.assertThat;
+import static org.mule.module.http.api.HttpConstants.Methods.POST;
+import org.mule.api.MuleEvent;
+import org.mule.api.MuleMessage;
+import org.mule.module.http.api.client.HttpRequestOptions;
+import org.mule.module.http.api.client.HttpRequestOptionsBuilder;
+import org.mule.tck.junit4.FunctionalTestCase;
+import org.mule.tck.junit4.rule.DynamicPort;
+import org.mule.tck.junit4.rule.SystemProperty;
+import org.mule.transport.ssl.DefaultTlsContextFactory;
+
+import java.io.IOException;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+/**
+ * Sets up some HTTPS servers and clients with different protocols and ciphers.
+ * Verifies only matching configurations are successful interacting with each other.
+ */
+public class HttpRestrictedCiphersAndProtocolsTestCase extends FunctionalTestCase
+{
+    @Rule
+    public DynamicPort port1 = new DynamicPort("port1");
+    @Rule
+    public DynamicPort port2 = new DynamicPort("port2");
+    @Rule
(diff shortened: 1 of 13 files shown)


transports/ssl/src/main/java/org/mule/transport/ssl/DefaultTlsContextFactory.java
```diff
@@ -28,18 +35,76 @@
...
+            return;
+        }
+
+        try
+        {
+            tlsConfiguration.initialise(null == getKeyStorePath(), null);
+        }
+        catch (CreateException e)
+        {
+            throw new InitialisationException(MessageFactory.createStaticMessage("Unable to initialise TLS configuration"), e, this);
```
elrodro83: If this exception is thrown, this object will appear as already initialized.

transports/ssl/src/main/java/org/mule/transport/ssl/DefaultTlsContextFactory.java
```diff
@@ -28,18 +35,76 @@
...
-    private boolean initialized = false;
+    private AtomicBoolean initialized = new AtomicBoolean(false);
     private boolean trustStoreInsecure = false;
+    private String[] enabledProtocols;
+    private String[] enabledCipherSuites;
+
+    @Override
+    public void initialise() throws InitialisationException
+    {
+        if (initialized.getAndSet(true))
```
marianogonzalez: shouldn't you use compareAndSet instead?

transports/ssl/src/main/java/org/mule/transport/ssl/DefaultTlsContextFactory.java
marianogonzalez: doesn't really matter... the muleContext will be discarded

transports/ssl/src/main/java/org/mule/transport/ssl/DefaultTlsContextFactory.java
```diff
@@ -28,18 +35,76 @@
...
+                String[] validCipherSuites = ArrayUtils.intersection(enabledCipherSuites, globalEnabledCipherSuites);
+                if (validCipherSuites.length < enabledCipherSuites.length)
+                {
+                    globalConfigNotHonored("cipher suites", globalEnabledCipherSuites);
+                }
+            }
+        }
+    }
+
+    private boolean isDefault(String[] array)
```
marianogonzalez: 100%sure the array will not be null?

afelisatti: Yes, but only because of how I use it. It'd be better to add that here since that would also mean defaults values are to be used. Thanks.

modules/http/src/test/java/org/mule/module/http/functional/HttpRestrictedCiphersAndProtocolsTestCase.java
```diff
@@ -0,0 +1,102 @@
...
+    @Rule
+    public DynamicPort port1 = new DynamicPort("port1");
+    @Rule
+    public DynamicPort port2 = new DynamicPort("port2");
+    @Rule
+    public DynamicPort port3 = new DynamicPort("port3");
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+    @Rule
+    public SystemProperty cipherSuites = new SystemProperty("cipherSuites", "TLS_DHE_DSS_WITH_AES_128_CBC_SHA");
```
pablolagreca: Do you have a test case testing with several cipher suites and protocols ?

afelisatti: Not combined, I'll add one. Thanks.

transports/ssl/src/main/java/org/mule/transport/ssl/DefaultTlsContextFactory.java
```diff
@@ -28,18 +35,76 @@
...
+                {
+                    globalConfigNotHonored("cipher suites", globalEnabledCipherSuites);
+                }
+            }
+        }
+    }
+
+    private boolean isUseDefaults(String[] array)
+    {
+        return (array == null) || ((array.length == 1) && DEFAULT.equalsIgnoreCase(array[0]));
```
marianogonzalez: i'll merge it but unnecessary parenthesis makes this harder to read than it should

PR #2235, opened 2015-12-10
MULE-9183: CXF: Status code for successful one-way operations should be

elrodro83:
202


dfeist:
What happens with one-way http requests? IMO behavior when using CXF and when not using CXF should be the same.

elrodro83:
That would be dependant on the application. In this case, this is a particular case defined in the soap specification (http://www.w3.org/TR/soap12-part2/#http-reqbindwaitstate).

However, if there are any other cases other that soap where a similar spec is defined, we should review that.

PR #2239, opened 2015-12-11
MULE-9153: Add support for interfaces as parameters of an @Operation

estebanwasinger:
Now when using a non instantiable class used as `@Operation` parameter or as `ConnectionProvider`/`Config` `@Parameter` child elements are not generated, these kind of classes can be only configured by reference.

### DIFF
--- a/modules/extensions-spring-support/src/main/java/org/mule/module/extension/internal/capability/xml/schema/SchemaBuilder.java
+++ b/modules/extensions-spring-support/src/main/java/org/mule/module/extension/internal/capability/xml/schema/SchemaBuilder.java
@@ -48,10 +48,12 @@ import static org.mule.module.extension.internal.util.MuleExtensionUtils.getDyna
 import static org.mule.module.extension.internal.util.NameUtils.getTopLevelTypeName;
 import static org.mule.module.extension.internal.util.NameUtils.hyphenize;
 import static org.mule.util.Preconditions.checkArgument;
+
 import org.mule.extension.annotation.api.Extensible;
 import org.mule.extension.api.introspection.ConfigurationModel;
 import org.mule.extension.api.introspection.ConnectionProviderModel;
 import org.mule.extension.api.introspection.DataQualifier;
+import org.mule.extension.api.introspection.DataQualifierVisitor;
 import org.mule.extension.api.introspection.DataType;
 import org.mule.extension.api.introspection.ExpressionSupport;
 import org.mule.extension.api.introspection.ExtensionModel;
@@ -86,6 +88,7 @@ import org.mule.module.extension.internal.capability.xml.schema.model.Union;
 import org.mule.module.extension.internal.introspection.AbstractDataQualifierVisitor;
 import org.mule.module.extension.internal.model.property.ExtendingOperationModelProperty;
 import org.mule.module.extension.internal.model.property.TypeRestrictionModelProperty;
+import org.mule.module.extension.internal.util.IntrospectionUtils;
 import org.mule.module.extension.internal.util.NameUtils;
 import org.mule.util.ArrayUtils;
 import org.mule.util.StringUtils;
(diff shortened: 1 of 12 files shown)


modules/extensions-support/src/main/java/org/mule/module/extension/internal/util/IntrospectionUtils.java
```diff
@@ -239,6 +240,19 @@ public static void checkInstantiable(Class<?> declaringClass, boolean requireDef
...
     }
 
+    public static boolean isInstantiable(Class<?> declaringClass)
+    {
+        try
+        {
+            checkInstantiable(declaringClass);
+            return true;
+        }
+        catch (IllegalArgumentException e)
```
marianogonzalez: invert this, throwing exceptions is expensive

modules/extensions-support/src/test/java/org/mule/module/extension/Weapon.java
```diff
@@ -0,0 +1,13 @@
...
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.module.extension;
+
+public interface Weapon
+{
+
+    public String kill();
```
marianogonzalez: public is redundant on an interface

marianogonzalez:
Couple of extra comments:

* Add a test case which validates that you can configure a default weapon on XML. Add it on the ConfigParserTestCAse
* Add a test case which validates that you can actually invoke that operation (OperationExecutionTestCase)
* Apply code formatting on all the classes you touched, specially in SchemaBuilder

modules/extensions-spring-support/src/main/java/org/mule/module/extension/internal/capability/xml/schema/SchemaBuilder.java
```diff
@@ -791,6 +734,43 @@ protected void defaultOperation()
         schema.getSimpleTypeOrComplexTypeOrGroup().add(complexType);
     }
 
+    private AbstractDataQualifierVisitor getDataQualifierVisitor(final ExtensionType extensionType, final ExplicitGroup all, final ParameterModel parameterModel)
```
marianogonzalez: The  getDataQualifierVisitor () name is not very descriptive. Rename to getParameterDeclarationVisitor (or something of that sort) and have the return type been simply DataQualifierVisitor

modules/extensions-support/src/main/java/org/mule/module/extension/internal/util/IntrospectionUtils.java
```diff
@@ -228,15 +230,20 @@ public static void checkInstantiable(Class<?> declaringClass)
 
     public static void checkInstantiable(Class<?> declaringClass, boolean requireDefaultConstructor)
     {
-        checkArgument(declaringClass != null, "declaringClass cannot be null");
+        checkArgument(isInstantiable(declaringClass, requireDefaultConstructor), String.format("Class %s cannot be instantiated.", declaringClass));
```
marianogonzalez: replace with an if. If you do it like this, you pay the price of the string.format even if the validation passes

modules/extensions-support/src/test/java/org/mule/module/extension/Ricin.java
```diff
@@ -42,4 +48,10 @@ public int hashCode()
     {
         return microgramsPerKilo.hashCode();
     }
+
+    @Override
+    public String kill()
+    {
+        return "You have been killed with Ricin";
```
marianogonzalez: you should use the RICIN_KILL_MESSAGE constant here. It also looks like that constant should be moved to this class

modules/extensions-spring-support/src/main/java/org/mule/module/extension/internal/capability/xml/schema/SchemaBuilder.java
```diff
@@ -99,6 +102,7 @@
 import java.util.Set;
 
 import javax.xml.bind.JAXBElement;
+import javax.xml.crypto.Data;
```
marianogonzalez: wtf?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/util/IntrospectionUtils.java
```diff
@@ -138,21 +140,41 @@ private static DataType toDataType(ResolvableType type)
...
+        else
+        {
+            resolvedDataType = DataType.of(rawClass, toRawTypes(generics));
+        }
 
-        return DataType.of(rawClass, toRawTypes(generics));
+        return resolvedDataType;
+    }
+
+    private static boolean hasRawClass(ResolvableType generic)
```
marianogonzalez: you're only using this once. Just inline it

modules/extensions-support/src/main/java/org/mule/module/extension/internal/util/IntrospectionUtils.java
```diff
@@ -228,15 +250,23 @@ public static void checkInstantiable(Class<?> declaringClass)
 
     public static void checkInstantiable(Class<?> declaringClass, boolean requireDefaultConstructor)
     {
-        checkArgument(declaringClass != null, "declaringClass cannot be null");
-
-        if (requireDefaultConstructor)
+        if (!isInstantiable(declaringClass, requireDefaultConstructor))
         {
-            checkArgument(hasDefaultConstructor(declaringClass), String.format("Class %s cannot be instantiated since it doesn't have a default constructor", declaringClass.getName()));
+            checkArgument(false, String.format("Class %s cannot be instantiated.", declaringClass));
```
marianogonzalez: checkARgument(false) doesn't make much sense... Simply throw the IllegalArgumentException yourself

PR #2241, opened 2015-12-11
MULE-8607 - fixing reconnection so SFTP is always connected after con…

pablolagreca:
…nect, always disconnected after disconnect and delegates reconnetion logic to each inbound endpoint

### DIFF
--- a/tests/functional/src/main/java/org/mule/tck/listener/ConnectionListener.java
+++ b/tests/functional/src/main/java/org/mule/tck/listener/ConnectionListener.java
@@ -11,7 +11,10 @@ import org.mule.api.MuleContext;
 import org.mule.api.context.notification.ConnectionNotificationListener;
 import org.mule.context.notification.ConnectionNotification;
 import org.mule.context.notification.NotificationException;
+import org.mule.util.Preconditions;
 import org.mule.util.concurrent.Latch;
+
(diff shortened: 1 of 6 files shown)


tests/functional/src/main/java/org/mule/tck/listener/ConnectionListener.java
```diff
@@ -22,8 +25,12 @@
 public class ConnectionListener
 {
 
+    private static final int UNDEFINED = 0;
+
     private CountDownLatch notificationReceivedLatch = new Latch();
     private int timeout = 10000;
+    private long previousNotificationTimestamp = UNDEFINED;
```
elrodro83: Could this be an Optional too?

tests/functional/src/main/java/org/mule/tck/listener/ConnectionListener.java
```diff
@@ -93,4 +110,13 @@ public ConnectionListener reset()
         return this;
     }
 
+    /**
+     * @return the minimum time tracked between all received notifications
+     * @throws IllegalStateException if there were less than two notifications received
+     */
+    public Long getMinimumTimeBetweenNotifications()
```
elrodro83: Is this method used anywhere else other than the test? If so, this precondition could be an assertion there.

PR #2246, opened 2015-12-15
MULE-9205: XmppMessageDispatcher doesn't reconnect

elrodro83:


transports/xmpp/src/main/java/org/mule/transport/xmpp/XmppMessageDispatcher.java
```diff
@@ -80,6 +80,12 @@ protected MuleMessage doSend(MuleEvent event) throws Exception
 
     protected void sendMessage(MuleEvent event) throws Exception
     {
+        // Handle session closing by the server
```
marianogonzalez: what about the message receiver?

transports/xmpp/src/main/java/org/mule/transport/xmpp/XmppMessageDispatcher.java
```diff
@@ -80,6 +80,12 @@ protected MuleMessage doSend(MuleEvent event) throws Exception
...
+        if (!xmppConnector.getXmppConnection().isConnected())
+        {
+            xmppConnector.getXmppConnection().connect();
```
marianogonzalez: what if this fails?

transports/xmpp/src/main/java/org/mule/transport/xmpp/XmppMessageDispatcher.java
elrodro83: The receiver uses a Smack listener. For that case, Smack reconnects automatically.

transports/xmpp/src/main/java/org/mule/transport/xmpp/XmppMessageDispatcher.java
elrodro83: An exception will be thrown. If the remote server is down or rejects the connection, Smack's reconnection mechanism will kick in.

PR #2251, opened 2015-12-17
MULE-9216: Introduce the new File and FTP connectors

marianogonzalez:


modules/file-extension-common/src/main/java/org/mule/module/extension/file/api/FileContentWrapper.java
```diff
@@ -0,0 +1,165 @@
...
+            }
+        }
+    }
+
+    private Object tryTransform(DataType sourceDataType, DataType targetDataType)
+    {
+        Transformer transformer;
+        try
+        {
+            transformer = event.getMuleContext().getRegistry().lookupTransformer(sourceDataType, targetDataType);
```
pablolagreca: Can this be updated to use TransformationService?

PR #2258, opened 2015-12-21
MULE-9166: Changed 'mule-core-tests' version placeholder to depend in…

estebanwasinger:
… the mule version instead of the project version of the extension

### DIFF
--- a/extensions/pom.xml
+++ b/extensions/pom.xml
@@ -53,7 +53,7 @@
         <dependency>
             <groupId>org.mule</groupId>
             <artifactId>mule-core-tests</artifactId>
-            <version>${project.version}</version>
+            <version>${mule.version}</version>
             <scope>test</scope>
             <type>test-jar</type>
         </dependency>


extensions/pom.xml
```diff
@@ -53,7 +53,7 @@
         <dependency>
             <groupId>org.mule</groupId>
             <artifactId>mule-core-tests</artifactId>
-            <version>${project.version}</version>
+            <version>${mule.version}</version>
```
pablokraan: Why do you need to do this? mule-core-tests is just another module in our project, so it must be referenced using project.version.

estebanwasinger: We need this because the extensions versions not always are going to be aligned with the Mule version. The same approach was taken for all the modules in this pom -->
[MULE-9055](https://www.mulesoft.org/jira/browse/MULE-9055) (https://github.com/mulesoft/mule/commit/1385d0a0fd4cce70799ed07dffc49d902d6179a4) attacks the same issue.

marianogonzalez: to solve that problem we're going to do a separate pom for the extensions that are not shipped with mule. For the ones that are part of the distributions, you should refer to by project.version. This PR is invalid

PR #2268, opened 2015-12-23
MULE-8753: Give the Extension model the concept of vendor

estebanwasinger:
This PR requires to have previously merged the following pull requests: https://github.com/mulesoft/mule-extensions-annotations/pull/25 and https://github.com/mulesoft/mule-extensions-api/pull/35

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/module/extension/internal/introspection/ImmutableExtensionModel.java
+++ b/modules/extensions-support/src/main/java/org/mule/module/extension/internal/introspection/ImmutableExtensionModel.java
@@ -9,6 +9,8 @@ package org.mule.module.extension.internal.introspection;
 import static org.mule.module.extension.internal.util.MuleExtensionUtils.toMap;
 import static org.mule.util.CollectionUtils.immutableList;
 import static org.mule.util.Preconditions.checkArgument;
+
+import org.mule.extension.api.exception.IllegalModelDefinitionException;
 import org.mule.extension.api.exception.NoSuchConfigurationException;
 import org.mule.extension.api.exception.NoSuchOperationException;
 import org.mule.extension.api.introspection.ConfigurationModel;
@@ -36,6 +38,7 @@ final class ImmutableExtensionModel extends AbstractImmutableModel implements Ex
     private final Map<String, ConfigurationModel> configurations;
     private final Map<String, OperationModel> operations;
     private final List<ConnectionProviderModel> connectionProviders;
+    private final String vendor;
 
     /**
      * Creates a new instance with the given state
@@ -43,6 +46,7 @@ final class ImmutableExtensionModel extends AbstractImmutableModel implements Ex
      * @param name                the extension's name. Cannot be blank
      * @param description         the extension's description
      * @param version             the extension's version
+     * @param vendor              the extension's vendor name
      * @param configurationModels a {@link List} with the extension's {@link ConfigurationModel configurationModels}
      * @param operationModels     a {@link List} with the extension's {@link OperationModel operationModels}
      * @param connectionProviders a {@link List} with the extension's {@link ConnectionProviderModel connection provider models}
@@ -52,6 +56,7 @@ final class ImmutableExtensionModel extends AbstractImmutableModel implements Ex
     protected ImmutableExtensionModel(String name,
                                       String description,
                                       String version,
+                                      String vendor,
                                       List<ConfigurationModel> configurationModels,
                                       List<OperationModel> operationModels,
                                       List<ConnectionProviderModel> connectionProviders,
@@ -65,8 +70,15 @@ final class ImmutableExtensionModel extends AbstractImmutableModel implements Ex
         this.operations = toMap(operationModels);
         this.connectionProviders = immutableList(connectionProviders);
 
-        checkArgument(!StringUtils.isBlank(version), "version cannot be blank");
+        checkArgument(!StringUtils.isBlank(version), "Version cannot be blank");
+
+        if (vendor == null)
+        {
+            throw new IllegalModelDefinitionException("Vendor cannot be null");
+        }
+
         this.version = version;
+        this.vendor = vendor;
     }
 
     /**
@@ -133,5 +145,14 @@ final class ImmutableExtensionModel extends AbstractImmutableModel implements Ex
     public List<ConnectionProviderModel> getConnectionProviders()
     {
         return connectionProviders;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getVendor()
+    {
+        return vendor;
     }
 }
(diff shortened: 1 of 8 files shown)


modules/extensions-support/src/main/java/org/mule/module/extension/internal/introspection/ImmutableExtensionModel.java
```diff
@@ -67,6 +71,7 @@ protected ImmutableExtensionModel(String name,
 
         checkArgument(!StringUtils.isBlank(version), "version cannot be blank");
         this.version = version;
+        this.vendor = vendor;
```
marianogonzalez: validate this is not null

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultExtensionManager.java
```diff
@@ -123,20 +124,23 @@ public void stop() throws MuleException
...
+
+        LOGGER.info("Registering extension {} (version {})", extensionName, extensionVersion);
 
-        if (extensionRegistry.containsExtension(extensionName))
+        if (extensionRegistry.containsExtension(extensionName, extensionVendor))
         {
             if (LOGGER.isDebugEnabled())
             {
                 LOGGER.debug("A extension of name '{}' (version {}) is already registered. Skipping...",
-                             extensionModel.getName(), extensionModel.getVersion());
```
marianogonzalez: should add the vendor to this message as well?

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ExtensionEntityKey.java
```diff
@@ -0,0 +1,44 @@
...
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.module.extension.internal.manager;
+
+import org.mule.extension.annotation.api.Extension;
+
+final class ExtensionEntityKey
```
marianogonzalez: add javadoc

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ExtensionEntityKey.java
```diff
@@ -0,0 +1,44 @@
...
+{
+
+    private String name;
```
marianogonzalez: key fields should be final

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ExtensionEntityKey.java
```diff
@@ -0,0 +1,44 @@
...
+            return false;
+        }
+        ExtensionEntityKey entity = (ExtensionEntityKey) o;
+        return name.equals(entity.name) && vendor.equals(entity.vendor);
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return name.hashCode() + vendor.hashCode();
```
marianogonzalez: not a good hashcode function. Use multiplication instead and add a prime number into the mix. Or use the HashCodeBuilder from apache commons

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ExtensionEntityKey.java
```diff
@@ -0,0 +1,44 @@
...
+            vendor = Extension.NO_VENDOR;
+        }
+        this.name = name;
+        this.vendor = vendor;
+    }
+
+    @Override
+    public boolean equals(Object o)
+    {
+        if (!(o instanceof ExtensionEntityKey))
```
marianogonzalez: this is more clear to read if you remove the ! and the extra parenthesis and move the rest of the method inside the if block.

If the conditions doesn't checks out, you return false outside

modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/ExtensionRegistry.java
```diff
@@ -151,7 +154,7 @@ void registerConfigurationProvider(ConfigurationProvider configurationProvider)
         catch (RegistrationException e)
         {
             throw new MuleRuntimeException(createStaticMessage(format(
-                    "Found exception while registering configuration provider '%'", configurationProvider.getName()))
+                    "Found exception while registering configuration provider '%s'", configurationProvider.getName()))
```
marianogonzalez: cool catch! :)

modules/extensions-support/src/test/java/org/mule/module/extension/internal/manager/DefaultExtensionManagerTestCase.java
```diff
@@ -61,6 +61,8 @@
 public class DefaultExtensionManagerTestCase extends AbstractMuleTestCase
```
marianogonzalez: add tests that:

* Verify you can't have a null vendor
* Extensions of same name but different vendor are not consider equal
* Exensions of same name and vendor are consider equal

modules/extensions-support/src/main/java/org/mule/module/extension/internal/introspection/ImmutableExtensionModel.java
estebanwasinger: Validating this property it's not entire clear, meanwhile using annotations "you can avoid the vendor configuration", when you use the fluent api, forcing the configuration of the vendor it's not intuitive to think that **No Vendor** means `.withVendor("")`, already exist the validation in `DefaultExtensionManager` which checks if it is null, and in this case, register the extension with a blank vendor.

marianogonzalez: I see what you mean but I don't see such validation on the DefaultExtensionManager. I think that if vendor cannot be null, then it should not be possible to instantiate this class on an invalid state, specially when the containsExtension() method depends on it. I think that the validation should be moved here

modules/extensions-support/src/main/java/org/mule/module/extension/internal/introspection/ImmutableExtensionModel.java
```diff
@@ -65,8 +70,15 @@ protected ImmutableExtensionModel(String name,
         this.operations = toMap(operationModels);
         this.connectionProviders = immutableList(connectionProviders);
 
-        checkArgument(!StringUtils.isBlank(version), "version cannot be blank");
+        checkArgument(!StringUtils.isBlank(version), "Version cannot be blank");
+
+        if (vendor == null)
+        {
+            throw new IllegalModelDefinitionException("Vendor extension cannot be null");
```
marianogonzalez: Rephrase as Vendor cannot be null

PR #2270, opened 2015-12-23
MULE-9236: Make MuleDocumentLoader extensible

elrodro83:


modules/spring-config/src/main/java/org/mule/config/spring/XmlMetadataFactory.java
```diff
@@ -0,0 +1,27 @@
...
+import org.mule.config.spring.parsers.XmlMetadataAnnotations;
+
+import org.xml.sax.Locator;
+
+/**
+ * Factory object to get a fresh instance of a {@link XmlMetadataAnnotations}.
+ * 
+ * This interface is an extension point to other Mule projects. Please be careful if you are going to make changes here.
+ */
+public interface XmlMetadataFactory
```
dfeist: Minor, but better IMO to use:
i) XmlMetadataAnnotationsFactory
ii) create() rather than createWith(), to be more consistent with other factories in Mule which don't attempt to be fluent.
WDYT?

PR #2279, opened 2015-12-29
MULE-9168: Users should not be forced to use @Parameter to expose POJ…

juandesi:
…O fields

### DIFF
--- a/modules/extensions-spring-support/src/main/java/org/mule/module/extension/internal/capability/xml/schema/SchemaBuilder.java
+++ b/modules/extensions-spring-support/src/main/java/org/mule/module/extension/internal/capability/xml/schema/SchemaBuilder.java
@@ -11,6 +11,7 @@ import static java.math.BigInteger.ZERO;
 import static org.apache.commons.lang.StringUtils.EMPTY;
 import static org.mule.extension.api.introspection.DataQualifier.LIST;
 import static org.mule.extension.api.introspection.DataQualifier.OPERATION;
+import static org.mule.extension.api.introspection.DataQualifier.POJO;
 import static org.mule.extension.api.introspection.ExpressionSupport.NOT_SUPPORTED;
 import static org.mule.extension.api.introspection.ExpressionSupport.SUPPORTED;
 import static org.mule.module.extension.internal.capability.xml.schema.PoolingSupport.REQUIRED;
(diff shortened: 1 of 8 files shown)


modules/extensions-support/src/test/java/org/mule/module/extension/internal/util/IntrospectionUtilsTestCase.java
```diff
@@ -88,6 +99,29 @@ public void getNullFieldDataType() throws Exception
         IntrospectionUtils.getFieldDataType(null);
     }
 
+    @Test
+    public void getNoAnnotatedExposedPojoFields()
+    {
+        Collection<Field> exposedFields = IntrospectionUtils.getExposedFields(LifetimeInfo.class);
```
marianogonzalez: you should always use hamcrest

modules/extensions-support/src/main/java/org/mule/module/extension/internal/util/IntrospectionUtils.java
```diff
@@ -305,6 +309,27 @@ public static boolean isVoid(Method method)
...
+        if (!allFields.isEmpty())
+        {
+            return allFields;
+        }
+        try
+        {
+            PropertyDescriptor[] propertyDescriptors = Introspector.getBeanInfo(extensionType, Object.class).getPropertyDescriptors();
+            return Arrays.stream(propertyDescriptors)
+                    .filter(propertyDescriptor -> propertyDescriptor.getReadMethod() != null)
+                    .map(propertyDescriptor -> getField(extensionType, propertyDescriptor.getName(), propertyDescriptor.getPropertyType()))
```
marianogonzalez: add a test case which verifies that field will never be null, even if the getter is not backed by a field of a consistent name

modules/extensions-spring-support/src/main/java/org/mule/module/extension/internal/capability/xml/schema/SchemaBuilder.java
```diff
@@ -775,14 +775,6 @@ protected void defaultOperation()
         };
     }
 
-    private boolean shouldGenerateChildElements(DataType dataType)
```
marianogonzalez: buuug

modules/extensions-support/src/test/java/org/mule/module/extension/internal/util/IntrospectionUtilsTestCase.java
```diff
@@ -88,6 +100,36 @@ public void getNullFieldDataType() throws Exception
...
+        assertThat(exposedFields, is(not(empty())));
+        assertThat(exposedFields.size(), is(2));
```
marianogonzalez: not a reason to reject the PR, but just FYI, there's a hasSize matcher in the Matchers class

PR #2286, opened 2015-12-30
MULE-9228: Split extensions parent POM, for internal and external ext…

estebanwasinger:
…ensions


extensions/pom.xml
```diff
@@ -11,95 +11,17 @@
     <modelVersion>4.0.0</modelVersion>
 
     <groupId>org.mule.extensions</groupId>
-    <artifactId>mule-extensions</artifactId>
+    <artifactId>bundled-mule-extensions</artifactId>
```
marianogonzalez: pretty sure this is wrong

PR #2298, opened 2016-01-06
MULE-9272: Improve logging when a TLS error occurs

afelisatti:


modules/http/src/main/java/org/mule/module/http/internal/request/DefaultHttpRequester.java
```diff
@@ -286,6 +292,14 @@ private MuleEvent resetMuleEventForNewThread(MuleEvent event)
                              }, getWorkManager(muleEvent));
     }
 
+    private void checkIfRemotelyClosed(Exception exception)
+    {
+        if (requestConfig.getTlsContext() != null && StringUtils.containsIgnoreCase(exception.getMessage(), "Remotely closed"))
```
dfeist: Use static string.

modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/MuleSslFilter.java
```diff
@@ -35,10 +39,18 @@ public MuleSslFilter(SSLEngineConfigurator serverSSLEngineConfigurator, SSLEngin
...
+        try
+        {
+            ctx.getAttributes().setAttribute(HTTPS.getScheme(), true);
+            NextAction nextAction = super.handleRead(ctx);
+            ctx.getAttributes().setAttribute(SSL_SESSION_ATTRIBUTE_KEY, getSslSession(ctx));
+            return nextAction;
+        }
+        catch (SSLHandshakeException e)
+        {
+            logger.error("SSL handshake error: " + e.getMessage());
```
dfeist: Doesn't this cause error to be logged twice, once here and once when it gets to exception handler?

afelisatti: AFAIK this is not handled at all. I could only see this error if SSL debug was enabled.

dfeist: You are right, this is listener, for some reason I assumed it was requester too.

PR #2301, opened 2016-01-07
MULE-9252: Allow users to use password editor for sensitive data fields

pablocabrera85:


modules/extensions-support/src/main/java/org/mule/module/extension/internal/introspection/enricher/PasswordModelEnricher.java
```diff
@@ -0,0 +1,53 @@
...
+ * @since 4.0
+ */
+public final class PasswordModelEnricher extends AbstractAnnotatedModelEnricher
+{
+
+    @Override
+    public void enrich(DescribingContext describingContext)
+    {
+        DeclarationDescriptor descriptor = describingContext.getDeclarationDescriptor();
+        descriptor.getRootDeclaration().getDeclaration().getConfigurations().forEach(configurationDeclaration -> configurationDeclaration.getParameters().forEach(this::addModelPropertyIfRequired));
```
marianogonzalez: this statement is just the same as the three below. Use a first order function to allow repetition. Repetition is the worst thing that can happen to an artist :)

modules/extensions-support/src/test/java/org/mule/module/extension/internal/PasswordModelPropertyTestCase.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private ExtensionFactory extensionFactory;
+    private ExtensionModel extensionModel;
+
+    @Before
+    public void setUp()
+    {
+        ClassLoader classLoader = getClass().getClassLoader();
+        ServiceRegistry serviceRegistry = mock(ServiceRegistry.class);
+        when(serviceRegistry.lookupProviders(ModelEnricher.class, classLoader)).thenReturn(asList(new PasswordModelEnricher()));
```
marianogonzalez: better use the any(ClassLoader) matcher rather than depending on the test classloader. THis way this test won't become problematic if class loading handling changes

modules/extensions-support/src/main/java/org/mule/module/extension/internal/introspection/enricher/PasswordModelEnricher.java
pablocabrera85: Will do

modules/extensions-support/src/test/java/org/mule/module/extension/internal/PasswordModelPropertyTestCase.java
pablocabrera85: Done

PR #2307, opened 2016-01-11
MULE-9278 Add Attributes to MuleMessage and make MuleMessage extend T…

dfeist:
…ypedValue.

### DIFF
--- a/core/src/main/java/org/mule/DefaultMuleMessage.java
+++ b/core/src/main/java/org/mule/DefaultMuleMessage.java
@@ -66,7 +66,7 @@ import org.apache.commons.logging.LogFactory;
  * <code>DefaultMuleMessage</code> is a wrapper that contains a payload and properties
  * associated with the payload.
  */
-public class DefaultMuleMessage implements MuleMessage, ThreadSafeAccess, DeserializationPostInitialisable
+public class DefaultMuleMessage extends TypedValue implements MuleMessage, ThreadSafeAccess, DeserializationPostInitialisable
 {
     private static final String NOT_SET = "<not set>";
 
@@ -80,7 +80,6 @@ public class DefaultMuleMessage implements MuleMessage, ThreadSafeAccess, Deseri
     private String id;
     private String rootId;
 
-    private transient Object payload;
     private transient Object originalPayload;
 
     /**
@@ -111,7 +110,7 @@ public class DefaultMuleMessage implements MuleMessage, ThreadSafeAccess, Deseri
     private transient AtomicReference<Thread> ownerThread = null;
     private transient AtomicBoolean mutable = null;
 
-    private DataType<?> dataType;
+    private Serializable attributes;
 
     private static DataType<?> getMessageDataType(MuleMessage previous, Object payload)
     {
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/mule/transformer/types/TypedValue.java
```diff
@@ -9,31 +9,110 @@
...
+        this.value = value;
+    }
+
+    /**
+     * Performs custom serialization of the value.
+     * <p>
+     * This implementation simply throws a {@link NotSerializableException} as custom serialization is not supported.
+     * Sub-classes can override this method to perform custom 'value' serialization.
+     */
+    protected Serializable serializeValue(ObjectOutputStream out) throws Exception
```
marianogonzalez: I don't think this is correct. This method only executes when the payload is not serializable. I would expect a method with this name to always execute when serialization happens. I think this method should go away and simply throw the exception when the payload is not serializable

core/src/main/java/org/mule/transformer/types/TypedValue.java
```diff
@@ -9,31 +9,110 @@
...
+        throw new NotSerializableException(value.getClass().getName());
+    }
+
+    /**
+     * Performs custom deserialization of the value
+     * <p>
+     * This implementation simply throws a {@link NotSerializableException} as custom serialization is not supported.
+     * Sub-classes can override this method to perform custom 'value' deserialization.
+     */
+    protected T deserializeValue(ObjectInputStream out) throws Exception
```
marianogonzalez: same as above

core/src/main/java/org/mule/DefaultMuleMessage.java
```diff
@@ -1322,6 +1353,24 @@ private void writeObject(ObjectOutputStream out) throws Exception
         return toWrite;
     }
 
+    @Override
+    protected Serializable serializeValue(ObjectOutputStream out) throws Exception
+    {
+        byte[] serializablePayload = muleContext.getTransformationService().getPayloadAsBytes(this);
```
marianogonzalez: where is this used? I couldn't find usages of this method in the PR. Other than that, the method name doesn't seem correct to me. This logic should only be performed when the payload is not serializable.

core/src/main/java/org/mule/DefaultMuleMessage.java
```diff
@@ -1322,6 +1353,24 @@ private void writeObject(ObjectOutputStream out) throws Exception
...
+        out.writeInt(serializablePayload.length);
+        new DataOutputStream(out).write(serializablePayload);
+        return serializablePayload;
+    }
+
+    @Override
+    protected Object deserializeValue(ObjectInputStream in) throws Exception
```
marianogonzalez: same as above

core/src/main/java/org/mule/transformer/types/TypedValue.java
dfeist: This is here to allow DefaultMuleMessage to override the behaviour.

core/src/main/java/org/mule/transformer/types/TypedValue.java
dfeist: same as above

core/src/main/java/org/mule/transformer/types/TypedValue.java
marianogonzalez: Ok, I see it now. I still think it's not very nice though. Look at the comments on the equivalent methods on DefaultMuleMessage

core/src/main/java/org/mule/DefaultMuleMessage.java
dfeist: It's used in TypedValue.writeObject(). It is only performed when payload is not serializable.

dfeist: BTW: This can all go away if transform the message to byte[] earlier before serialization, but that's a separate issue/PR.  So this would potentially disappear soon.

PR #2311, opened 2016-01-12
MULE-9305: Remove "ESB" occurrences from logs and splash

elrodro83:


core/src/test/resources/META-INF/MANIFEST.MF
```diff
@@ -12,8 +12,7 @@ Implementation-Vendor-Id: org.mule.patterns
 Implementation-Vendor: MuleSoft, Inc.
 Build-Date: 2012-Apr-13 16:46:57
 Build-Revision: 24242
-Description: Mule ESB and Integration Platform
-Dev-List-Email: mule-esb@mulesoft.com
```
afelisatti: Are you sure this isn't require somehow? Same with the mailing list in the pom.

elrodro83: I just removed the code that used this entry. Other than that this is a custom field.
As for the pom entry, it is entirely optional, and used only for the site generation through maven.

core/src/main/java/org/mule/DefaultMuleMessage.java
```diff
@@ -6,7 +6,7 @@
  */
 package org.mule;
 
-import static org.mule.util.SystemUtils.LINE_SEPARATOR;
+import static org.apache.commons.lang.SystemUtils.LINE_SEPARATOR;
```
afelisatti: Was this a mistake?

elrodro83: Seems to be, i'll change it back.

PR #2313, opened 2016-01-12
MULE-9312 - HttpMessageProcessTemplate is consuming request content t…

pablolagreca:
…wice

### DIFF
--- a/core/src/main/java/org/mule/transport/AbstractTransportMessageProcessTemplate.java
+++ b/core/src/main/java/org/mule/transport/AbstractTransportMessageProcessTemplate.java
@@ -6,11 +6,7 @@
  */
 package org.mule.transport;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mule/transport/AbstractTransportMessageProcessTemplate.java
```diff
@@ -6,11 +6,7 @@
  */
 package org.mule.transport;
 
-import org.mule.api.MessagingException;
-import org.mule.api.MuleContext;
-import org.mule.api.MuleEvent;
-import org.mule.api.MuleException;
-import org.mule.api.MuleMessage;
+import org.mule.api.*;
```
marianogonzalez: wildcards not allowed

PR #2321, opened 2016-01-13
MULE-9312 - HttpMessageProcessTemplate is consuming request content t…

pablolagreca:
…wice

### DIFF
--- a/core/src/main/java/org/mule/transport/AbstractTransportMessageProcessTemplate.java
+++ b/core/src/main/java/org/mule/transport/AbstractTransportMessageProcessTemplate.java
@@ -11,12 +11,11 @@ import org.mule.api.MuleContext;
 import org.mule.api.MuleEvent;
 import org.mule.api.MuleException;
 import org.mule.api.MuleMessage;
+import org.mule.api.ThreadSafeAccess;
 import org.mule.api.config.MuleProperties;
 import org.mule.api.construct.FlowConstruct;
-import org.mule.api.context.WorkManager;
 import org.mule.api.endpoint.InboundEndpoint;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mule/transport/AbstractTransportMessageProcessTemplate.java
```diff
@@ -43,8 +43,20 @@ public AbstractTransportMessageProcessTemplate(MessageReceiverType messageReceiv
...
-        return createEventFromMuleMessage(messageFromSource);
+        if (muleEvent == null)
+        {
+            MuleMessage messageFromSource = createMessageFromSource(getOriginalMessage());
+            muleEvent = createEventFromMuleMessage(messageFromSource);
+        }
+        else
+        {
+            MuleMessage message = muleEvent.getMessage();
+            if(message instanceof ThreadSafeAccess)
```
pablokraan: Fix code style

core/src/main/java/org/mule/transport/AbstractTransportMessageProcessTemplate.java
```diff
@@ -43,8 +43,20 @@ public AbstractTransportMessageProcessTemplate(MessageReceiverType messageReceiv
...
+            {
+                ((ThreadSafeAccess)message).resetAccessControl();
```
pablokraan: Why is needed to reset the access control?

pablolagreca: Since getMuleEvent() may be called by different threads. For instance, Throttling may get the event, do some processing, and then retry again later in a different thread.

PR #2338, opened 2016-01-14
MULE-9083: Add the concept of Message Sources in the Ext-Api

marianogonzalez:


dfeist:
IMO The change to mule-core and refactor resulting from this should be be a separate commit (with different JIRA ID) to the addition of ext-api message source (this JIRA issue).   Also the mule-api commit should not really be with MULE-9083 either, as there isn't anything their directly related to sources there.  Given @pablolagreca has been reviewing this I'll let him decide if this is important enough to change..

marianogonzalez:
DataType didn't get a refactor. It just got moved so that it can be used. The refactor will come on M2

modules/extensions-spring-support/src/main/java/org/mule/module/extension/internal/config/SourceBeanDefinitionParser.java
```diff
@@ -0,0 +1,66 @@
...
+    {
+        super(ExtensionMessageSourceFactoryBean.class);
+        this.extensionModel = extensionModel;
+        this.sourceModel = sourceModel;
+    }
+
+    @Override
+    protected void doParse(BeanDefinitionBuilder builder, Element element, ParserContext parserContext)
+    {
+        builder.setScope(SCOPE_PROTOTYPE);
```
pablolagreca: Probably doesn't matter because it's only retrieve once from the registry but it must be a singleton

marianogonzalez: fixed

PR #2339, opened 2016-01-15
MULE-9324: No exception is thrown when folder in path is removed

elrodro83:


transports/file/src/main/java/org/mule/transport/file/FileMessageReceiver.java
```diff
@@ -674,6 +674,11 @@ protected boolean attemptFileLock(final File sourceFile) throws MuleException
      */
     List<File> listFiles() throws MuleException
     {
+        if (!readDirectory.exists())
+        {
+            throw new DefaultMuleException(FileMessages.errorWhileListingFiles());
```
afelisatti: Why not do this below if listFiles returns null?

PR #2341, opened 2016-01-15
MULE-9306 When using MuleClient from JavaComponent or from a MessageP…

gsfernandes:
…rocessor the RequestContext is left inconsistent so Flow and Session variables were lost, also exception handling was not using the event being processed to handle errors when the component and processor is invoked.

### DIFF
--- a/core/src/main/java/org/mule/component/DefaultComponentLifecycleAdapter.java
+++ b/core/src/main/java/org/mule/component/DefaultComponentLifecycleAdapter.java
@@ -9,7 +9,6 @@ package org.mule.component;
 import static org.mule.config.i18n.MessageFactory.createStaticMessage;
 
 import org.mule.DefaultMuleEventContext;
-import org.mule.RequestContext;
 import org.mule.api.DefaultMuleException;
 import org.mule.api.MuleContext;
 import org.mule.api.MuleEvent;
@@ -339,7 +338,7 @@ public class DefaultComponentLifecycleAdapter implements LifecycleAdapter
         {
             if (componentObject == null)
             {
-                throw new ComponentException(createStaticMessage("componentObject is null"), RequestContext.getEvent(), component);
+                throw new ComponentException(createStaticMessage("componentObject is null"), event, component);
             }
             // Use the overriding entrypoint resolver if one is set
             if (component.getEntryPointResolverSet() != null)
@@ -353,7 +352,7 @@ public class DefaultComponentLifecycleAdapter implements LifecycleAdapter
         }
         catch (Exception e)
         {
-            throw new ComponentException(createStaticMessage("%s: %s", e.getClass().getName(), e.getMessage()), RequestContext.getEvent(), component, e);
+            throw new ComponentException(createStaticMessage("%s: %s", e.getClass().getName(), e.getMessage()), event, component, e);
         }
 
         return result;
--- a/core/src/main/java/org/mule/execution/TransactionalErrorHandlingExecutionTemplate.java
+++ b/core/src/main/java/org/mule/execution/TransactionalErrorHandlingExecutionTemplate.java
@@ -85,6 +85,8 @@ public class TransactionalErrorHandlingExecutionTemplate implements ExecutionTem
      *
      * @param muleContext MuleContext for this application
      * @param transactionConfig Transaction configuration
+     *
+     * @deprecated 
      */
     public static TransactionalErrorHandlingExecutionTemplate createMainExecutionTemplate(MuleContext muleContext, TransactionConfig transactionConfig)
     {
--- a/tests/integration/src/test/java/org/mule/test/integration/client/MuleClientDispatchExceptionHandlingTestCase.java
+++ b/tests/integration/src/test/java/org/mule/test/integration/client/MuleClientDispatchExceptionHandlingTestCase.java
@@ -1,0 +1,216 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.test.integration.client;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.core.Is.is;
+
+import org.mule.DefaultMuleMessage;
+import org.mule.RequestContext;
+import org.mule.api.DefaultMuleException;
+import org.mule.api.MuleEvent;
+import org.mule.api.MuleEventContext;
+import org.mule.api.MuleException;
+import org.mule.api.MuleMessage;
+import org.mule.api.client.LocalMuleClient;
+import org.mule.api.lifecycle.Callable;
+import org.mule.api.processor.MessageProcessor;
+import org.mule.tck.junit4.FunctionalTestCase;
+import org.mule.tck.junit4.rule.DynamicPort;
+import org.mule.util.concurrent.Latch;
+
+import java.util.concurrent.TimeUnit;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.junit.Rule;
+import org.junit.Test;
+
+/**
+ * Tests to validate that MuleClient can be used from JavaComponent/MessageProcessor in order to dispatch an event to
+ * a sub-flow and if the component/processor throws an exception afterwards the main-flow exception strategy handles
+ * it.
+ */
+public class MuleClientDispatchExceptionHandlingTestCase extends FunctionalTestCase
+{
+    private static final Log logger = LogFactory.getLog(MuleClientDispatchExceptionHandlingTestCase.class);
+
+    @Rule
+    public DynamicPort dynamicPort1 = new DynamicPort("port1");
+
+    // These attributes need to be accessed from JavaComponent and MessageProcessor static classes therefore
+    // they are declared as static
+    private static Latch innerFlowLatch;
+    private static Latch exceptionLatch;
+    private static MuleEvent eventFromMainFlow;
+    private static MuleMessage messageFromMainFlow;
+    private static boolean eventPropagated;
+    private static boolean isSameMessage;
+
+    @Override
+    protected String getConfigFile()
+    {
+        return "org/mule/test/integration/client/client-dispatch-catch-exception-flow.xml";
+    }
+
+    /**
+     * Validates that a JavaComponent after doing a dispatch to a sub-flow using MuleClient
+     * throws an exception and the catch-exception-strategy defined in main-flow is called.
+     * It also validates that original event passed to JavaComponent is later propagated
+     * to the JavaComponent defined in catch-exception-strategy block.
+     *
+     * @throws Exception
+     */
+    @Test
+    public void testCatchExceptionThrowFromJavaComponentToJavaComponent() throws Exception
+    {
+        doSendMessageToEndpoint("vm://catchExceptionJavaComponentToJavaComponent");
+    }
+
+    @Test
+    public void tesCatchExceptionThrowFromJavaComponentToMessageProcessor() throws Exception
+    {
+        doSendMessageToEndpoint("vm://catchExceptionJavaComponentToMessageProcessor");
+    }
+
+    @Test
+    public void testCatchExceptionThrowFromMessageProcessorToJavaComponent() throws Exception
+    {
+        doSendMessageToEndpoint("vm://catchExceptionMessageProcessorToJavaComponent");
+    }
+
+    @Test
+    public void tesCatchExceptionThrowFromMessageProcessorToMessageProcessor() throws Exception
+    {
+        doSendMessageToEndpoint("vm://catchExceptionMessageProcessorToMessageProcessor");
+    }
+
+    @Test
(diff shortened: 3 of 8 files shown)


tests/integration/src/test/java/org/mule/test/integration/client/MuleClientDispatchExceptionHandlingTestCase.java
```diff
@@ -0,0 +1,187 @@
...
+import org.apache.commons.logging.LogFactory;
+import org.junit.Rule;
+import org.junit.Test;
+
+/**
+ * Tests to validate that MuleClient can be used from JavaComponent/MessageProcessor in order to dispatch an event to
+ * a sub-flow and if the component/processor throws an exception afterwards the main-flow exception strategy handles
+ * it.
+ *
+ * @author guillermo.fernandes@mulesoft.com
```
dfeist: We don't add author to java files.

gsfernandes: Ok, I'll remove it.

On Fri, Jan 15, 2016 at 2:36 PM, dfeist <notifications@github.com> wrote:

> In
> tests/integration/src/test/java/org/mule/test/integration/client/MuleClientDispatchExceptionHandlingTestCase.java
> <https://github.com/mulesoft/mule/pull/2341#discussion_r49881148>:
>
> > +import org.mule.tck.junit4.rule.DynamicPort;
> > +
> > +import java.util.concurrent.CountDownLatch;
> > +import java.util.concurrent.TimeUnit;
> > +
> > +import org.apache.commons.logging.Log;
> > +import org.apache.commons.logging.LogFactory;
> > +import org.junit.Rule;
> > +import org.junit.Test;
> > +
> > +/**
> > + * Tests to validate that MuleClient can be used from JavaComponent/MessageProcessor in order to dispatch an event to
> > + * a sub-flow and if the component/processor throws an exception afterwards the main-flow exception strategy handles
> > + * it.
> > + *
> > + * @author guillermo.fernandes@mulesoft.com
>
> We don't add author to java files.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mulesoft/mule/pull/2341/files#r49881148>.
>



-- 
<http://www.mulesoft.com/>Guillermo Fernandes, Senior Software Engineer Av.
Corrientes 316 Entre Piso, Buenos Aires, C1043AAQ, Argentina
  <https://www.linkedin.com/company/www.mulesoft.com>
<https://twitter.com/mulesoft>   <https://www.facebook.com/MuleSoft>
<https://www.youtube.com/user/mulesoftvids>We're hiring!
<http://www.mulesoft.com/careers>

tests/integration/src/test/java/org/mule/test/integration/client/MuleClientDispatchExceptionHandlingTestCase.java
```diff
@@ -0,0 +1,187 @@
...
+     */
+    @Test
+    public void testCatchExceptionUsingVMConnector() throws Exception
+    {
+        doSendMessageToEndpoint("vm://vmconnector", "Business Error!");
+    }
+
+    private void doSendMessageToEndpoint(String endpoint, String expectedPayload) throws Exception
+    {
+        innerFlowLatch = new CountDownLatch(1);
```
dfeist: Can use org.mule.util.concurrent.Latch().  Also isn't it better for the latch to not be static, this way it will automatically be initialized for each test method.

tests/integration/src/test/resources/org/mule/test/integration/client/client-dispatch-javamessageprocessor-flow.xml
```diff
@@ -0,0 +1,51 @@
...
+        <spring:bean id="myProcessor"
+                     class="org.mule.test.integration.client.MuleClientDispatchFromJavaMessageProcessorTestCase$FlowVarsDispatchFlowUsingNewMuleClient"/>
+    </spring:beans>
+
+    <flow name="flowVarsFlowUsingVM">
+        <vm:inbound-endpoint exchange-pattern="request-response" path="vmflowvars"/>
+        <set-variable variableName="team" value="Sales"/>
+        <processor ref="myProcessor"/>
+        <set-variable variableName="ammount" value="100"/>
+        <set-payload value="#[flowVars.team + ': ' + flowVars.ammount]"/>
```
dfeist: Consider using FlowAsserts.verify("flowVarsFlowUsingVM") and <test:assert/> processors as it make the test much more readable/direct.

tests/integration/src/test/resources/org/mule/test/integration/client/client-dispatch-catch-exception-flow.xml
```diff
@@ -0,0 +1,71 @@
...
+    </flow>
+
+    <flow name="catchExceptionJavaComponentUsingHTTP">
+        <http:listener config-ref="HTTP_Listener_Configuration" path="/test" />
+        <set-payload value="empty"/>
+        <component class="org.mule.test.integration.client.MuleClientDispatchExceptionHandlingTestCase$DispatchInnerFlowThrowExceptionJavaComponent"/>
+        <set-payload value="Sales"/>
+        <catch-exception-strategy>
+            <set-payload value="Business Error!"/>
+            <custom-processor class="org.mule.test.integration.client.MuleClientDispatchExceptionHandlingTestCase$ExecutionCountDownProcessor"/>
```
dfeist: Consider using FlowAsserts.verify("flowVarsFlowUsingVM") and test:assert/ processors as it make the test much more readable/direct.

tests/integration/src/test/java/org/mule/test/integration/client/MuleClientDispatchFromJavaMessageProcessorTestCase.java
```diff
@@ -0,0 +1,88 @@
...
+
+    public static class FlowVarsDispatchFlowUsingNewMuleClient implements MessageProcessor
+    {
+
+        @Override
+        public MuleEvent process(MuleEvent event) throws MuleException
+        {
+            event.getMuleContext().getClient().dispatch("vm://vminnertest", new DefaultMuleMessage("payload", event.getMuleContext()));
+            return event;
+
```
dfeist: Ideally remove extra lines..

tests/integration/src/test/java/org/mule/test/integration/client/MuleClientDispatchExceptionHandlingTestCase.java
```diff
@@ -0,0 +1,185 @@
...
+
+/**
+ * Tests to validate that MuleClient can be used from JavaComponent/MessageProcessor in order to dispatch an event to
+ * a sub-flow and if the component/processor throws an exception afterwards the main-flow exception strategy handles
+ * it.
+ */
+public class MuleClientDispatchExceptionHandlingTestCase extends FunctionalTestCase
+{
+
+    private static final int TIMEOUT = 5000;
```
dfeist: Use org.mule.tck.junit4.AbstractMuleContextTestCase#RECEIVE_TIMEOUT instead of defined timeout here.

tests/integration/src/test/java/org/mule/test/integration/client/MuleClientDispatchExceptionHandlingTestCase.java
```diff
@@ -0,0 +1,185 @@
...
+        innerFlowLatch = new CountDownLatch(1);
+        exceptionLatch = new CountDownLatch(1);
+
+        LocalMuleClient client = muleContext.getClient();
+        final HttpRequestOptions httpRequestOptions = newOptions().method(POST.name()).responseTimeout(TIMEOUT).build();
+        logger.debug("Sending initial message to trigger execution of flow to be tested");
+        MuleMessage result = client.send(endpoint, getTestMuleMessage("Original Message"), httpRequestOptions);
+        logger.debug("Flow execution completed, result:" + result);
+
+        boolean innerFlowCalled = innerFlowLatch.await(3, TimeUnit.SECONDS);
```
dfeist: Use org.mule.tck.junit4.AbstractMuleContextTestCase#RECEIVE_TIMEOUT instead of magic number?

core/src/main/java/org/mule/component/DefaultComponentLifecycleAdapter.java
```diff
@@ -339,7 +338,7 @@ public Object invoke(MuleEvent event) throws MuleException
         {
             if (componentObject == null)
             {
-                throw new ComponentException(createStaticMessage("componentObject is null"), RequestContext.getEvent(), component);
+                throw new ComponentException(createStaticMessage("componentObject is null"), event, component);
```
pablokraan: I think that is not correct to use the event. That could fix this particular issue, but suppose that in the exception strategy there is another component that uses Requestcontext.getEvent... that event will be incorrect.

gsfernandes: I've modified the junit to validate this scenario and the event in requestcontext is the correct one. 
I would say that during the catch-exception-strategy handling the event is stored in the requestcontext therefore this is not an issue. I will push the changes to the PR.

gsfernandes: @dfeist, @pablokraan I did the change to propagate the exceptionHandler from flowConstruct on AbstractMessageReceiver when the TransactionalErrorHandlingExecutionTemplate is created and reverted the change in DefaultComponentLifecycleAdapter to use RequestContext.getEvent() when handling the exceptions and as @pablokraan mentioned the MuleEvent in RequestContext is incorrect so if the flow has another component in catch-exception-strategy that access to it may incur in a problem.

What do you think would be the correct way to handle this?

dfeist: Put the RequestContext set in a finally block?

gsfernandes: Didn't work. Because when the ComponentException is created with a reference to the "wrong event" (the one created by the dispatch process) this event is later loaded in RequestContext before calling the component defined in the catch-exception-strategy block using that reference set in ComponentException and later stored in RequestContext.

gsfernandes:
@dfeist @pablokraan please take a look to the latest changes. As @dfeist suggested setting the event in a finally block on DefaultComponentLifecycleAdapter was a partial solution do to I also needed to pass the event to the ComponentException due to later the event referenced by the exception is set to the RequestContext and therefore overwrites the correct one that we set in finally block. I added a comment to these lines explaining the reason. Please take a look to it.

dfeist:
Sorry I wasn't clear.  I meant putting "MuleEvent result = messageProcessorExecutionTemplate.execute(processor, event);" in BlockingProcessorExecutor in a try/finally block.  Because putting it DefaultComponentLifecycleAdapter will only cover Components and not MP's

gsfernandes:
Ok, but still that won't fix the problem of having the incorrect event in RequestContext because we do the set in a finally block but the exception handling mechanism later runs the catch-exception-strategy block that has another JavaComponent and this one receives the incorrect event.
This is because the AbstractComponent sets the event in RequestContext when it is invoked and this one is invoked with the event attached to the ComponentException instead of the one located in RequestContext.

In addition to this I found that in ExceptionToMessagingExceptionExecutionInterceptor when a MessageProcessor throws an exception this one is handled differently because it is wrapped (yet) as a MessagingException therefore it is instantiated at that point and the event attached to it is the one processed:

 ExceptionToMessagingExceptionExecutionInterceptor
 catch (Throwable ex)
        {
            throw putContext(new MessagingException(event, ex, messageProcessor), messageProcessor, event);
        }

gsfernandes:
If I do the same logic to handling an exception when invoking a JavaComponent as it is being done when invoking a MessageProcessor we will fix issues without having to change the RequestContext.

What I mean is to leave the original fix that we added at first in  DefaultComponentLifecycleAdapter.invoke to avoid using RequestContext when handling an exception there. This would be similar behaviour as it is done when a custom MessageProcessor throws an exception (as mentioned in previous comment).
I just committed the change and added a comment to that line.

core/src/main/java/org/mule/component/DefaultComponentLifecycleAdapter.java
gsfernandes: The MessageProcessorNotificationExecutionInterceptor updates the RC with the eventToProcess that in this case will be the one passed to the ComponentException here.

core/src/main/java/org/mule/execution/TransactionalErrorHandlingExecutionTemplate.java
```diff
@@ -80,18 +80,6 @@ public static TransactionalErrorHandlingExecutionTemplate createMainExecutionTem
     }
 
     /**
-     * Creates a TransactionalErrorHandlingExecutionTemplate to be used as first processing template in a flow using no particular exception listener.
-     * Exception listener configured in the flow within this ExecutionTemplate is executed will be used.
-     *
-     * @param muleContext MuleContext for this application
-     * @param transactionConfig Transaction configuration
-     */
-    public static TransactionalErrorHandlingExecutionTemplate createMainExecutionTemplate(MuleContext muleContext, TransactionConfig transactionConfig)
```
dfeist: Deprecate instead of remove (unless you check with PLG and he thinks it's ok to just remove)

gsfernandes: I'll deprecate this and mention to PLG so we could remove this in another PR.

PR #2347, opened 2016-01-18
MULE-9083: Add the concept of Message Sources in the Ext-Api

marianogonzalez:


core-tests/src/test/java/org/mule/message/processing/AsyncResponseFlowProcessingPhaseTestCase.java
```diff
@@ -138,7 +138,7 @@ public void order()
     @Test
     public void runPhaseWithMessagingExceptionThrown() throws Exception
     {
-        when(mockContext.supportsAsynchronousProcessing()).thenReturn(true);
```
dfeist: Can you explain this change?

marianogonzalez: the default value for supportsAsynchronousProcessing() was changed in  AsyncResponseFlowProcessingPhase. The same change was missing in this test case

PR #2348, opened 2016-01-18
MULE-9299: Remove VM transport usage from Scripting Module.

anosenzo:


pablolagreca:
I think we should add an item in https://www.mulesoft.org/jira/browse/MULE-9307 so we don't forget to add support for calling operations inside an scripting component.

PR #2350, opened 2016-01-19
MULE-9158 Change TransformationService to not mutate MuleMessage

dfeist:


core-tests/src/test/java/org/mule/el/context/MessageContextTestCase.java
```diff
@@ -139,22 +141,22 @@ public void assignPayload() throws Exception
     public void payloadAsType() throws Exception
     {
         MuleMessage mockMessage = mock(MuleMessage.class);
+        MuleMessage transformedMessage = mock(MuleMessage.class, withSettings().defaultAnswer(RETURNS_DEEP_STUBS));
```
pablokraan: This should do the same and is more readable: mock(MuleMessage.class, RETURNS_DEEP_STUBS);

core/src/main/java/org/mule/DefaultMuleEvent.java
```diff
@@ -605,7 +605,13 @@ public MuleMessage getMessage()
         {
             throw new TransformerException(CoreMessages.objectIsNull("outputType"));
         }
-        return getMuleContext().getTransformationService().getPayload(message, outputType);
+
+        MuleMessage transformedMessage = getMuleContext().getTransformationService().transform(message, outputType);
+        if (ClassUtils.isConsumable(message.getPayload().getClass()))
```
pablokraan: Add static import.
Makes sense to have MuleMessage.isConsumable()?

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -98,32 +100,14 @@ public MuleMessage applyTransformers(final MuleMessage message, final MuleEvent
             result = applyAllTransformers(message, event, transformers);
         }
 
-        if (null != outputType && !result.getPayload().getClass().isAssignableFrom(outputType))
+        if (outputType != null)
```
pablokraan: Should not check if outputType is compatible with result.getPayload.getDatatype instead of checking for non null?

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -84,7 +86,7 @@ public MuleMessage applyTransformers(final MuleMessage message, final MuleEvent
      * @param outputType the required output type for this transformation. By adding this parameter some additional
      * transformations will occur on the message payload to ensure that the final payload is of the specified type.
      * If no transformers can be found in the registry that can transform from the return type of the transformation
-     * list to the outputType and exception will be thrown
+     * list to the outputType and exception will be thrown.  This parameter is optional.
```
pablokraan: This method is a bit weird, because it receives a list of transformers and then an optional datatype. When a datatype is given, that means that another transformation will be required. Is not more readable to have two different methods, one receiving a lit of transformers an another one receiving a datatype?

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -210,22 +149,27 @@ public String getPayloadForLogging(MuleMessage message)
     /**
      * Obtains a {@link String} representation of the message payload for logging without throwing exception.
      * If encoding is required it will use the encoding set on the message.
-     * <p/>
-     * If the existing payload is consumable (i.e. can't be read twice) then the existing payload of the message will be
-     * replaced with a byte[] representation as part of this operations.
+     * <p>
+     * If the existing payload is consumable (i.e. can't be read twice) or an exception occurs during transformation
+     * then the
```
pablokraan: then the?

core/src/main/java/org/mule/el/context/MessageContext.java
```diff
@@ -100,14 +101,37 @@ public Object getPayload()
...
+     * <p/>
+     * <b>Note:</b> If this current payload is a stream this operation will read the stream making it unreadable for
+     * further processing.  To avoid this transform the message prior to use of this expression.
+     * TODO MULE-9285
+     *
+     * @param type the java type the payload is to be transformed to
+     * @return the transformed payload
+     * @throws TransformerException
+     */
     public <T> T payloadAs(Class<T> type) throws TransformerException
```
pablokraan: Do we still need this method now that the transformation service is available?

core/src/main/java/org/mule/el/context/MessageContext.java
```diff
@@ -100,14 +101,37 @@ public Object getPayload()
...
+    }
+
+    /**
+     * Obtains the payload of the current message transformed to the given #dataType.
+     * <p/>
+     * <b>Note:</b> If this current payload is a stream this operation will read the stream making it unreadable for
+     * further processing.  To avoid this transform the message prior to use of this expression.
+     * TODO MULE-9285
+     *
+     * @param dt the DatType to transform the current message payload to
```
pablokraan: Can you rename dt?

core/src/main/java/org/mule/routing/Foreach.java
```diff
@@ -89,13 +89,19 @@ public MuleEvent process(MuleEvent event) throws MuleException
         boolean transformed = false;
         if (xpathCollection)
         {
-            transformed = transformPayloadIfNeeded(message);
+            MuleMessage transformedMessage = transformPayloadIfNeeded(message);
+            if(transformedMessage != message)
```
pablokraan: Fix code style

core/src/main/java/org/mule/transformer/types/DataTypeFactory.java
```diff
@@ -41,11 +41,11 @@
     public static final DataType<String> RSS_STRING = new SimpleDataType<>(String.class, MimeTypes.RSS);
 
     //Common Java types
-    public static final DataType<String> STRING = new SimpleDataType<>(String.class);
-    public static final DataType<String> OBJECT = new SimpleDataType<>(Object.class);
-    public static final DataType<String> BYTE_ARRAY = new SimpleDataType<>(byte[].class);
-    public static final DataType<String> INPUT_STREAM = new SimpleDataType<>(InputStream.class);
-    public static final DataType<String> MULE_MESSAGE = new SimpleDataType<>(MuleMessage.class);
+    public static final DataType<String> STRING = new SimpleDataType<String>(String.class);
```
pablokraan: is needed to use new SimpleDataType<String> instead of ew SimpleDataType<>?

modules/cxf/src/test/java/org/mule/module/cxf/DirectXmlTestCase.java
```diff
@@ -47,6 +48,7 @@ public void testInputStream() throws Exception
         test(xml);
     }
 
+    @Ignore // TODO MULE-9285
```
pablokraan: Use @Ignore("MULE-9285")

core/src/main/java/org/mule/DefaultMuleEvent.java
dfeist: I think we did at one point but it got moved out, otherwise message has to have list of consumable types.

core/src/main/java/org/mule/TransformationService.java
dfeist: Seems this method is not used with the refactoring i've done, removing.

core/src/main/java/org/mule/TransformationService.java
dfeist: Seems this method is not used with the refactoring i've done, removing.

core/src/main/java/org/mule/el/context/MessageContext.java
dfeist: Yes, at least for now to maintain MEL compatibility. If we change this then it's part of a separate issue, we should decide if we want to expose transformation like this to the user in EL or via transformation service itself (for now transformation service is just being used as a way to refactor transformation out of the message)

core/src/main/java/org/mule/transformer/types/DataTypeFactory.java
dfeist: Can't remember if this was strictly needed, but DataType if created with generics should use correct type and not Object right?

core-tests/src/test/java/org/mule/el/context/MessageContextTestCase.java
```diff
@@ -139,22 +141,22 @@ public void assignPayload() throws Exception
...
-        assertSame(b, evaluate("message.payloadAs(org.mule.tck.testmodels.fruit.Banana)", mockMessage));
+        when(transformationService.transform(any(MuleMessage.class), any(DataType.class))).thenReturn(transformedMessage);
+        assertSame(transformedMessage.getPayload(), evaluate("message.payloadAs(org.mule.tck.testmodels.fruit.Banana)", mockMessage));
     }
 
     @Test
     public void payloadAsDataType() throws Exception
     {
         MuleMessage mockMessage = mock(MuleMessage.class);
+        MuleMessage transformedMessage = mock(MuleMessage.class, withSettings().defaultAnswer(RETURNS_DEEP_STUBS));
```
pablokraan: change this mock too

core/src/main/java/org/mule/DefaultMuleEvent.java
pablokraan: We moved the code from DefaultMuleMessage to ClassUtils, that is correct. I'm asking if, from API point of view, makes sense to have MuleMessage#isConsumable. We can discuss it later

core/src/main/java/org/mule/TransformationService.java
```diff
@@ -210,22 +116,27 @@ public String getPayloadForLogging(MuleMessage message)
...
+     * then the an exeption won't be thrown but rather a description of the payload type will be returned.
      *
-     * @return message payload as a String
+     * @return message payload as a String or message with the payload type if payload can't be converted to a String
      */
     public String getPayloadForLogging(MuleMessage message, String encoding)
     {
-        try
+        Class type = message.getPayload().getClass();
+        if (!ClassUtils.isConsumable(type))
```
pablokraan: Static import?

PR #2353, opened 2016-01-19
MULE-9266: Default values are not registered in the Extension XSD

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -237,6 +238,11 @@ public static String getDefaultValue(Optional optional)
         return Optional.NULL.equals(defaultValue) ? null : defaultValue;
     }
 
+    public static Object getDefaultValue(AccessibleObject object)
```
pablokraan: And the javadoc? where is the javadoc?

marianogonzalez: javadoc added... in Gerli

PR #2362, opened 2016-01-20
MULE-9314: Upgrade libraries and remove deprecated components in XML …

marianogonzalez:
…module


core/src/main/java/org/mule/management/stats/ComponentStatistics.java
```diff
@@ -156,7 +156,7 @@ public synchronized void addExecutionBranchTime(boolean first, long branch, long
         {
             maxExecutionTime = effectiveTotal;
         }
-        averageExecutionTime = Math.round(totalExecTime / executedEvent);
+        averageExecutionTime = executedEvent > 0 ? Math.round(totalExecTime / executedEvent) : 0;
```
dfeist: Should this be in this PR?

dfeist:
You need to remove expression evaluators from enum in mule.xsd also.

core/src/main/java/org/mule/management/stats/ComponentStatistics.java
marianogonzalez: I ran into tests failing because of this. Could extract it to a separate issue but I don't think it's worth it

PR #2365, opened 2016-01-21
MULE-9338: Fix flaxy SFTP tests

elrodro83:


tests/functional/src/main/java/org/mule/tck/util/sftp/SftpServer.java
```diff
@@ -89,9 +95,9 @@ public void stop()
     {
         try
         {
-            sshdServer.stop(true);
+            sshdServer.stop(false);
```
afelisatti: What does this change imply?

elrodro83: This does a graceful shutdown of the sftp server instead of an immediate one.
This change didn't have any noticeable time penalty in my testing.

PR #2369, opened 2016-01-21
MULE-9335 Remove Expression Evaluators

dfeist:


core-tests/src/test/java/org/mule/expression/ExpressionManagerTestCase.java
```diff
@@ -30,32 +24,11 @@
     public void testManager() throws Exception
     {
         MuleMessage message = new DefaultMuleMessage("test", muleContext);
-        Object o = muleContext.getExpressionManager().evaluate("function:uuid", message);
+        Object o = muleContext.getExpressionManager().evaluate("org.mule.util.UUID.getUUID()", message);
         assertNotNull(o);
-        o = muleContext.getExpressionManager().evaluate("function:now", message);
+        o = muleContext.getExpressionManager().evaluate("server.dateTime.toDate()", message);
```
marianogonzalez: unrelated to this PR, but we should also remove the custom DateTime classes in mule-core and use java8 dates instead

core/src/main/java/org/mule/expression/transformers/BeanBuilderTransformer.java
```diff
@@ -114,7 +114,7 @@ public Object transformMessage(MuleMessage message, String outputEncoding) throw
             if (!argument.isOptional() && value == null)
             {
                 throw new TransformerException(CoreMessages.expressionEvaluatorReturnedNull(
-                        argument.getExpressionConfig().getEvaluator(), argument.getExpressionConfig().getExpression()), this);
+                        null, argument.getExpressionConfig().getExpression()), this);
```
marianogonzalez: it is weird to have a null argument here

core/src/main/java/org/mule/expression/transformers/BeanBuilderTransformer.java
```diff
@@ -114,7 +114,7 @@ public Object transformMessage(MuleMessage message, String outputEncoding) throw
             if (!argument.isOptional() && value == null)
             {
                 throw new TransformerException(CoreMessages.expressionEvaluatorReturnedNull(
```
marianogonzalez: if expression evaluators are gone, why do we still need a expression evaluator related message?

core/src/main/java/org/mule/expression/transformers/ExpressionTransformer.java
```diff
@@ -66,7 +66,7 @@ public Object transformMessage(MuleMessage message, String outputEncoding) throw
             if (!argument.isOptional() && results[i] == null)
             {
                 throw new TransformerException(CoreMessages.expressionEvaluatorReturnedNull(
```
marianogonzalez: same as before

PR #2374, opened 2016-01-22
MULE-9149: Upgrade antlr to 3.5

elrodro83:


modules/pom.xml
```diff
@@ -31,7 +31,6 @@
         <module>http</module>
         <module>jaas</module>
         <module>jboss-transactions</module>
-        <module>jbpm</module>
```
afelisatti: Not sure you should remove this. Check how it was done for axis.

PR #2375, opened 2016-01-22
MULE-9334: DB Connection leak with an invalid table in Select statement

elrodro83:
- Fix tests


marianogonzalez:
I'll merge it because machaval said this was ok, but file a jira to review this because this will much likely affect studio's performance

PR #2379, opened 2016-01-22
MULE-9269: Validation for name clashing in operation's arguments

pablomusumeci:


modules/extensions-support/src/main/java/org/mule/module/extension/internal/introspection/validation/NameClashModelValidator.java
```diff
@@ -162,6 +197,15 @@ private void validateClashes(Set<String> set1, Set<String> set2, String type1, S
                                                                  extensionModel.getName(), type1, type2, Joiner.on(", ").join(intersection)));
             }
         }
+
+        private void validateClash(String name1, String name2, String type1, String type2)
```
marianogonzalez: use better argument names

PR #2382, opened 2016-01-25
MULE-9306 Usage of MuleClient for dispatching or sending an event to …

gsfernandes:
…a sub-flow from  JavaComponent or MessageProcessor leaves the RequestContext inconsistent with the event dispatched/sent by MuleClient instead of the one being processed. This generated two issues related with losing flow, session variables and strategy listener defined in main flow. Even the event later passed later to the exception strategy listener was incorrect.

Created a different PR due to 3.6.x has a different code.
@dfeist @elrodro83 please take a look to this.

### DIFF
--- a/core/src/test/java/org/mule/processor/chain/DefaultMessageProcessorChainTestCase.java
+++ b/core/src/test/java/org/mule/processor/chain/DefaultMessageProcessorChainTestCase.java
@@ -10,6 +10,7 @@ package org.mule.processor.chain;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.not;
 import static org.hamcrest.CoreMatchers.nullValue;
+import static org.hamcrest.core.IsEqual.equalTo;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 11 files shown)


core/src/test/java/org/mule/processor/chain/DefaultMessageProcessorChainTestCase.java
```diff
@@ -156,7 +156,6 @@ public void testMPChainWithNullReturnAtEnd() throws MuleException, Exception
         DefaultMessageProcessorChainBuilder builder = new DefaultMessageProcessorChainBuilder();
         builder.chain(new AppendingMP("1"), new AppendingMP("2"), new AppendingMP("3"), new ReturnNullMP());
         assertNull(builder.build().process(getTestEventUsingFlow("0")));
-        assertThat(RequestContext.getEvent(), is(nullValue()));
```
dfeist: Can you assert something else instead?

gsfernandes: I changed some of these tests to assert for the latest event processed to be in RC after the process is executed. I think in this case I was not able to do that but let me check it.

PR #2386, opened 2016-01-26
MULE-9353 - Operation invocation using reflection fails since it does…

pablolagreca:
… not resolves all the method parameters


modules/extensions-support/src/main/java/org/mule/module/extension/internal/runtime/executor/ReflectiveMethodOperationExecutor.java
```diff
@@ -73,7 +73,62 @@
     @Override
     public Object execute(OperationContext operationContext) throws Exception
     {
-        return invokeMethod(operationMethod, executorDelegate, getParameterValues(operationContext));
+        return invokeMethod(operationMethod, executorDelegate, resolvePrimitiveTypes(operationMethod, getParameterValues(operationContext)));
+    }
+
+    private Object[] resolvePrimitiveTypes(Method operationMethod, Object[] parameterValues)
```
marianogonzalez: I think this method is best suited if moved to MethodArgumentResolverDelegate

PR #2387, opened 2016-01-26
MULE-9295: Remove VM transport usage from DB

elrodro83:


modules/db/src/test/java/org/mule/module/db/integration/xa/AbstractXaTransactionalTestCase.java
```diff
@@ -65,37 +68,47 @@ public void commitsChanges() throws Exception
...
-        MuleMessage response = client.send("vm://testRollback", TEST_MESSAGE, null);
-
-        assertTrue(response.getPayload() instanceof NullPayload);
-
-        List<Map<String, String>> result = selectData("select * from PLANET where POSITION=4", getDefaultDataSource());
-        assertRecords(result, new Record(new Field("NAME", MARS.getName()), new Field("POSITION", 4)));
+        try
+        {
+            runTransactionalFlow("jdbcRollback", ACTION_ALWAYS_BEGIN, new XaTransactionFactory());
+            fail("Expected 'java.lang.IllegalStateException'");
```
pablokraan: Why is expecting IllegalStateException and then catches only ComponentException?

tests/functional/src/main/java/org/mule/functional/junit4/FunctionalTestCase.java
```diff
@@ -267,8 +275,92 @@ protected MuleEvent runFlow(String flowName) throws Exception
     }
 
     /**
-     * Runs the given flow with a default event expecting a failure.
-     * Will fail if there's no failure running the flow.
+     * Runs the given flow with a default event inside a transaction
+     * 
+     * @param flowName the name of the flow to be executed
+     * @param action See {@link TransactionConfig} constants
+     * @param factory
```
pablokraan: Add missing param description

tests/functional/src/main/java/org/mule/functional/junit4/FunctionalTestCase.java
```diff
@@ -267,8 +275,92 @@ protected MuleEvent runFlow(String flowName) throws Exception
...
+    {
+        try
+        {
+            this.runTransactionalFlow(flowName, payload, action, factory);
+        }
+        catch (MessagingException e)
+        {
+            return e;
+        }
+        catch (Exception e)
```
pablokraan: Flow is executed inside a ExecutionCallback#process which throws Exception. Why are you adding this restriction here? Is not better to expect an exception and then check that is of an expected type only when requested?

modules/db/src/test/java/org/mule/module/db/integration/xa/AbstractXaTransactionalTestCase.java
```diff
@@ -65,37 +68,47 @@ public void commitsChanges() throws Exception
...
-        assertTrue(response.getPayload() instanceof NullPayload);
-
-        List<Map<String, String>> result = selectData("select * from PLANET where POSITION=4", getDefaultDataSource());
-        assertRecords(result, new Record(new Field("NAME", "Mercury"), new Field("POSITION", 4)));
+        try
+        {
+            runTransactionalFlow("rollbackWithNonTransactionalMP", ACTION_ALWAYS_BEGIN, new XaTransactionFactory());
+            fail("Expected 'java.lang.IllegalStateException'");
+        }
+        catch (ComponentException e)
```
pablokraan: Why can't you use runTransactionalFlowExpectingException on these test methods?

modules/db/src/test/java/org/mule/module/db/integration/xa/AbstractXaTransactionalTestCase.java
elrodro83: The IllegalStateException is thrown by the test component, and will be wrapped by the Component exception. Will change the failure message to reflect that.

PR #2389, opened 2016-01-26
Initialisation value should be respected when optional parameter defaults to null

juandesi:


modules/extensions-support/src/main/java/org/mule/module/extension/internal/util/SingleValueSetter.java
```diff
@@ -45,6 +46,6 @@ public SingleValueSetter(ParameterModel parameterModel, Field field)
     @Override
     public void set(Object target, ResolverSetResult resolverSetResult)
     {
-        fieldSetter.set(target, resolverSetResult.get(parameterModel));
+        Optional.ofNullable(resolverSetResult.get(parameterModel)).ifPresent(v -> fieldSetter.set(target, v));
```
marianogonzalez: Optional makes sense when you are returning nulllable values. In this case, which is completely internal, this becomes overkill. Use an if instead

juandesi: but in this case I have a nullable value, and only if its present i want to process it.  
This way I dont assing a variable just for checking the state Optional does it for me

marianogonzalez: yes, but that's something you can do with a simple if.

Instead you're:

* Instantiating an optional (which ends up checking null)
* Instantiating a lamba (which gets instantiated even if the value is indeed null)

Again, this would be just fine if you were planning to externalize the value. But since this is a private piece of code, simply check null with an if

PR #2394, opened 2016-01-27
MULE-9355 IMAP does not connect to MS Exchange shared mailboxes

gsfernandes:


transports/email/src/main/java/org/mule/transport/email/MicrosoftExchangeImapConnector.java
```diff
@@ -0,0 +1,59 @@
...
+     *     <li>using Acitve Directory name (note that the name used is the one defined as 'User SamAccountName logon' in AD): CONTOSO\jdoe</li>
+     *     <li>accessing a shared mailbox (username same as previous case): CONTOSO\jdoe\sharedmailbox</li>
+     * </ul>
+     *
+     * @param protocol
+     * @param properties
+     */
+    private void addExchangeAuthenticationProperties(String protocol, Properties properties)
+    {
+        properties.put(String.format("mail.%s.auth.plain.disable",
```
pablokraan: I'm not an expert on MSExchange, but seems strange to me that we didn't have problems in the past with it. Are you sure that these properties are OK for any user using it?

transports/email/src/main/java/org/mule/transport/email/MicrosoftExchangeImapsConnector.java
```diff
@@ -0,0 +1,59 @@
...
+     * Disables the following authentication methods in order to use LOGIN as this is the one that
+     * works with MS Exchange through IMAPS.
+     * The support IMAP username format are:
+     * <ul>
+     *     <li>using email: jdoe@contoso.com</li>
+     *     <li>using Acitve Directory name (note that the name used is the one defined as 'User SamAccountName logon' in AD): CONTOSO\jdoe</li>
+     *     <li>accessing a shared mailbox (username same as previous case): CONTOSO\jdoe\sharedmailbox</li>
+     * </ul>
+     *
+     * @param protocol
```
pablokraan: Add a description for these parameters

transports/email/src/main/java/org/mule/transport/email/MicrosoftExchangeImapsConnector.java
```diff
@@ -0,0 +1,59 @@
...
+    protected void extendPropertiesForSession(Properties global, Properties local, URLName url)
+    {
+        super.extendPropertiesForSession(global, local, url);
+        addExchangeAuthenticationProperties(ImapsConnector.IMAPS, local);
+    }
+
+    /**
+     * Disables the following authentication methods in order to use LOGIN as this is the one that
+     * works with MS Exchange through IMAPS.
+     * The support IMAP username format are:
```
pablokraan: Add paragraph separator

transports/email/src/main/java/org/mule/transport/email/MicrosoftExchangeImapsConnector.java
```diff
@@ -0,0 +1,59 @@
...
+     *     <li>using Acitve Directory name (note that the name used is the one defined as 'User SamAccountName logon' in AD): CONTOSO\jdoe</li>
+     *     <li>accessing a shared mailbox (username same as previous case): CONTOSO\jdoe\sharedmailbox</li>
+     * </ul>
+     *
+     * @param protocol
+     * @param properties
+     */
+    private void addExchangeAuthenticationProperties(String protocol, Properties properties)
+    {
+        properties.put(String.format("mail.%s.auth.plain.disable",
```
pablokraan: Don't wrap lines if not necessary (<120?)

transports/email/src/test/java/org/mule/transport/email/connectors/MicrosoftExchangeImapsConnectorTestCase.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Override
+    public Connector createConnector() throws Exception
+    {
+        MicrosoftExchangeImapsConnector connector = new MicrosoftExchangeImapsConnector(muleContext);
+        connector.setName("ImapsConnector");
+        connector.setCheckFrequency(POLL_PERIOD_MS);
+        connector.setServiceOverrides(newEmailToStringServiceOverrides());
+
+        connector.setTrustStore("greenmail.jks");
```
pablokraan: The only differences with the IMAP version seem to be the use of a trust store and ServerSetup.PROTOCOL_IMAPS instead of ServerSetup.PROTOCOL_IMAP. Can this class be refactored to avoid the duplication?

transports/email/src/main/java/org/mule/transport/email/MicrosoftExchangeImapConnector.java
gsfernandes: My neither, but I found several issues when trying to connect to shared mailboxes on Exchange. I would say that most of the use cases for IMAP connector with Exchange where used to connect to a user mailbox either by username or user email.
I have tried to just add these properties to the endpoint instead of using the explicit connector for MS Exchange created in this PR and that would be another way to solve this issue without adding custom IMAP connector that we would need to test with each version of MS Exchange.
```
<imaps:inbound-endpoint user="CONTOSO\jdoe\sharedmailbox" password="MyPassword" host="192.168.56.101" port="993">
            <property key="mail.imaps.auth.plain.disable" value="true"/>
            <property key="mail.imaps.auth.ntlm.disable" value="true"/>
            <property key="mail.imaps.auth.gssapi.disable" value="true"/>
        </imaps:inbound-endpoint>
```

What would you think? I know that I have implemented this but now that I found that by adding manually properties it works too I'm in favor or not adding a custom connector and close this PR without merging it.

gsfernandes: Closing PR as we are not going to add a new connector for something that could be solved by passing these properties to the endpoint.

gsfernandes:
Closing PR as we are not going to add a new connector for something that could be solved by passing these properties to the endpoint.

PR #2396, opened 2016-01-27
MULE-9258: Added new annotation @Expression with support for literal parameters

pablomusumeci:


modules/extensions-support/src/main/java/org/mule/module/extension/internal/introspection/utils/ImplicitObjectUtils.java
```diff
@@ -48,7 +49,7 @@ public static ResolverSet buildImplicitResolverSet(ParametrizedModel parametrize
             if (defaultValue != null)
             {
                 ValueResolver<Object> valueResolver;
-                if (defaultValue instanceof String && expressionManager.isExpression((String) defaultValue))
+                if (defaultValue instanceof String && expressionManager.isExpression((String) defaultValue) && !parameterModel.getExpressionSupport().equals(ExpressionSupport.LITERAL))
```
marianogonzalez: enums are compared using == and !=

modules/extensions-support/src/main/java/org/mule/module/extension/internal/util/IntrospectionUtils.java
```diff
@@ -280,8 +281,8 @@ public static boolean isRequired(ParameterModel parameterModel, boolean forceOpt
 
     public static ExpressionSupport getExpressionSupport(AccessibleObject object)
     {
-        Parameter parameter = object.getAnnotation(Parameter.class);
-        return parameter != null ? parameter.expressionSupport() : SUPPORTED;
+        Expression expression = object.getAnnotation(Expression.class);
```
marianogonzalez: this is the second time I read these three lines. Create a static method in IntrospectionUtils which do this comparisson and centralize the default value

modules/extensions-support/src/test/java/org/mule/module/extension/HeisenbergOperations.java
```diff
@@ -188,4 +190,10 @@ public String getSaulPhone(@Connection HeisenbergConnection connection)
     {
         return connection.getSaulPhoneNumber();
     }
+
+    @Operation
+    public String dontEvaluateLiteralExpression(@Expression(expressionSupport = ExpressionSupport.LITERAL) String literalExpression)
```
marianogonzalez: rename to literalEcho

modules/extensions-support/src/test/java/org/mule/module/extension/HeisenbergExtension.java
```diff
@@ -123,14 +125,26 @@
...
     private String firstEndevour;
 
+    @Parameter
+    @Optional(defaultValue = "#[payload]")
+    @Expression(expressionSupport = LITERAL)
+    private String literalExpressionWithDefault;
+
+    @Parameter
+    @Optional
+    @Expression(expressionSupport = LITERAL)
```
marianogonzalez: once the expressionSupport field has been renamed to value, you don't need the = anymore

modules/extensions-spring-support/src/main/java/org/mule/module/extension/internal/config/XmlExtensionParserUtils.java
```diff
@@ -307,6 +312,11 @@ static ResolverSet getResolverSet(ElementDescriptor element, List<ParameterModel
                     resolver = new StaticValueResolver(null);
                 }
 
+                if (parameterModel.getExpressionSupport() == ExpressionSupport.LITERAL)
```
marianogonzalez: this branch evaluates many conditions, but only one of them should be true for a given parameter model. That means that once you hit a possitve, you shouldn't continue evaluating the rest.

That been said, the literal case should be evaluated before the parseParameter case above

modules/extensions-support/src/main/java/org/mule/module/extension/internal/introspection/ParsedParameter.java
```diff
@@ -19,7 +20,7 @@
  *
  * @since 3.7.0
  */
-final class ParsedParameter
+final class ParsedParameter implements AnnotatedElement
```
marianogonzalez: why do you need to implement this interface?

pablomusumeci: Because ParameterModel implemented the method getAnnotation() (which is contained in the Interface AnnotatedElement) but not the other 2 methods I implemented in this commit. By implementing AnnotatedElement I could ask the ExpressionSupport of the ParsedParameter. See line https://github.com/mulesoft/mule/pull/2396/files#diff-3e704c3bd995eae525b12afc05490bddR413

PR #2403, opened 2016-02-09
MULE-9367 : Removed apostrophes from filePattern

ullgren:
Fixed default log filename to not include apostrophes.


gsfernandes:
Closing PR as it was already implemented in #2432. Thank you!

PR #2408, opened 2016-02-12
MULE-9378: Expand FlowRunner to support different FlowConstruct types

elrodro83:


tests/functional/src/main/java/org/mule/functional/junit4/FlowConstructRunner.java
```diff
@@ -0,0 +1,194 @@
...
+     * @return this {@link FlowRunner}
+     */
+    public R spyObjects()
+    {
+        eventBuilder.spyObjects();
+    
+        return (R) this;
+    }
+
+    public FlowConstructRunner(MuleContext muleContext)
```
pablokraan: Add javadoc

tests/functional/src/main/java/org/mule/functional/junit4/FlowConstructRunner.java
```diff
@@ -0,0 +1,194 @@
...
+    {
+        this.muleContext = muleContext;
+    }
+
+    /**
+     * Builds a new event based on this runner's config. If one has already been built, it will return that one.
```
pablokraan: I don't know what are the usages for this method, but seems a bit counterintuitive that it only creates an event the first time that is invoked.

tests/functional/src/main/java/org/mule/functional/junit4/FlowRunner.java
```diff
@@ -76,121 +69,24 @@ public Object transform(Object input)
...
+     * Configures the flow to run inside a transaction.
      * 
-     * @param key the key of the flow variable to put
-     * @param value the value of the flow variable to put
+     * @param action See {@link TransactionConfig} constants
+     * @param factory See {@link MuleTransactionConfig#setFactory(TransactionFactory)}.
      * @return this {@link FlowRunner}
      */
-    public FlowRunner withFlowVariable(String key, Object value)
+    public FlowRunner transactionally(byte action, TransactionFactory factory)
```
pablokraan: Can the byte parameter be replaced with something more meaningful?

PR #2417, opened 2016-02-16
Update xmlsec version 1.5.3 to 1.5.8 which is the version used in CXF…

anosenzo:
… 2.7.15

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1100,11 +1100,6 @@
                 </exclusions>
             </dependency>
             <dependency>
-                <groupId>org.apache.santuario</groupId>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -1100,11 +1100,6 @@
                 </exclusions>
             </dependency>
             <dependency>
-                <groupId>org.apache.santuario</groupId>
-                <artifactId>xmlsec</artifactId>
-                <version>${xmlSecurityVersion}</version>
```
marianogonzalez: shouldn't the property  be removed as well?

marianogonzalez:
what's the motivation for this other than matching cxf version?

PR #2436, opened 2016-02-29
MULE-9330: Misconfigured file:list throws unfriendly error

marianogonzalez:


extensions/file/src/main/java/org/mule/extension/file/internal/command/LocalFileCommand.java
```diff
@@ -43,7 +43,13 @@
     @Override
     protected Path resolvePath(String filePath)
     {
-        return Paths.get(config.getBaseDir()).resolve(filePath).toAbsolutePath();
+        Path path = Paths.get(config.getBaseDir());
```
elrodro83: Can this be extracted to FileCommand?

PR #2437, opened 2016-02-29
MULE-9377: Map elements are not properly generated in the XSD

estebanwasinger:


modules/extensions-spring-support/src/main/java/org/mule/module/extension/internal/capability/xml/schema/builder/SchemaBuilder.java
```diff
@@ -612,14 +612,17 @@ private LocalComplexType generateMapComplexType(String name, final String descri
             public void onPojo()
             {
                 entryComplexType.getAttributeOrAttributeGroup().add(createAttribute(ATTRIBUTE_NAME_VALUE, valueType, false, SUPPORTED));
```
alegmarra: if the child element is not generated, then the attribute should be required:
_createAttribute(ATTRIBUTE_NAME_VALUE, valueType, **true**, SUPPORTED)_

modules/extensions-spring-support/src/test/resources/map.xsd
```diff
@@ -19,15 +19,6 @@
...
-                                                <xs:element minOccurs="0" maxOccurs="1" name="object">
-                                                    <xs:complexType>
-                                                        <xs:complexContent>
-                                                            <xs:extension xmlns="http://www.mulesoft.org/schema/mule/map" base="java.lang.Object"></xs:extension>
-                                                        </xs:complexContent>
-                                                    </xs:complexType>
-                                                </xs:element>
-                                            </xs:sequence>
                                             <xs:attribute type="mule:expressionString" use="required" name="key"></xs:attribute>
                                             <xs:attribute type="mule:expressionObject" use="optional" name="value"></xs:attribute>
```
alegmarra: If not child element, then the _use_ of attribute _value_ should be required

PR #2439, opened 2016-02-29
MULE-9405 NullPointerException when setting a context-property with a…

gsfernandes:
… null value on org.mule.module.xml.transformer.XsltTransformer

### DIFF
--- a/modules/xml/src/main/java/org/mule/module/xml/transformer/XsltTransformer.java
+++ b/modules/xml/src/main/java/org/mule/module/xml/transformer/XsltTransformer.java
@@ -6,6 +6,7 @@
  */
 package org.mule.module.xml.transformer;
 
+import org.mule.api.DefaultMuleException;
 import org.mule.api.MuleMessage;
 import org.mule.api.lifecycle.InitialisationException;
 import org.mule.api.transformer.Transformer;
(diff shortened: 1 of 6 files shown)


modules/xml/src/main/java/org/mule/module/xml/transformer/XsltTransformer.java
```diff
@@ -230,7 +230,15 @@ protected void doTransform(MuleMessage message, String outputEncoding, Source so
                 for (Entry<String, Object> parameter : contextProperties.entrySet())
                 {
                     String key = parameter.getKey();
-                    transformer.setParameter(key, evaluateTransformParameter(key, parameter.getValue(), message));
+                    Object value = evaluateTransformParameter(key, parameter.getValue(), message);
+                    if(value != null)
```
marianogonzalez: code format

modules/xml/src/main/java/org/mule/module/xml/transformer/XsltTransformer.java
```diff
@@ -230,7 +230,15 @@ protected void doTransform(MuleMessage message, String outputEncoding, Source so
...
+                    {
+                        transformer.setParameter(key, value);
+                    }
+                    else
+                    {
+                        logger.debug("Value for context-property with key: " + key + " is null");
```
marianogonzalez: I would rather throw a DefaultMuleException or MuleRuntimeException...

PR #2442, opened 2016-03-01
MULE-9405 Updated migration file

gsfernandes:


gsfernandes:
Done, changed commit message to be the same as the original. Thank you!

PR #2445, opened 2016-03-02
MULE-9389: Mule Context's getTransactionManager randomly returns null if

elrodro83:
called during context start

### DIFF
--- a/tests/functional/src/main/java/org/mule/tck/junit4/ApplicationContextBuilder.java
+++ b/tests/functional/src/main/java/org/mule/tck/junit4/ApplicationContextBuilder.java
@@ -53,6 +53,13 @@ public class ApplicationContextBuilder
     public MuleContext build() throws Exception
     {
         // Should we set up the manager for every method?
+        MuleContext context = doBuildContext();
+        context.start();
+        return context;
+    }
+
+    protected MuleContext doBuildContext() throws Exception
+    {
         MuleContext context;
         MuleContextFactory muleContextFactory = new DefaultMuleContextFactory();
         List<ConfigurationBuilder> builders = new ArrayList<ConfigurationBuilder>();
@@ -66,7 +73,6 @@ public class ApplicationContextBuilder
         builders.add(getAppBuilder(this.applicationResources));
         configureMuleContext(muleContextBuilder);
         context = muleContextFactory.createMuleContext(builders, muleContextBuilder);
-        context.start();
         return context;
     }
 
@@ -88,5 +94,4 @@ public class ApplicationContextBuilder
     {
         contextBuilder.setWorkListener(new TestingWorkListener());
     }
-
 }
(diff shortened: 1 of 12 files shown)


tests/functional/src/main/java/org/mule/tck/junit4/ApplicationContextBuilder.java
```diff
@@ -26,6 +26,7 @@
 {
 
     private MuleContext domainContext;
+    private boolean disableMuleContextStart = false;
```
marianogonzalez: why would i want this?

tests/functional/src/main/java/org/mule/tck/junit4/ApplicationContextBuilder.java
```diff
@@ -89,4 +93,9 @@ protected void configureMuleContext(MuleContextBuilder contextBuilder)
         contextBuilder.setWorkListener(new TestingWorkListener());
     }
 
+    public void disableMuleContextStart()
```
marianogonzalez: setters is a bad idea for this because it can be called at any time but is only useful when used before build. Overload the build method instead

tests/integration/src/test/resources/org/mule/test/integration/exceptions/exception-dlq-flow.xml
```diff
@@ -14,8 +14,6 @@
     <jms:activemq-connector name="jmsConn" maxRedelivery="1"
         brokerURL="vm://localhost?broker.persistent=false&amp;broker.useJmx=false"/>
 
-    <jboss-transaction-manager/>
```
marianogonzalez: how's this related?

tests/integration/src/test/resources/org/mule/test/integration/exceptions/exception-dlq-service.xml
```diff
@@ -13,8 +13,6 @@
 
     <jms:activemq-connector name="jmsConn" maxRedelivery="1" brokerURL="vm://localhost?broker.persistent=false&amp;broker.useJmx=false"/>
 
-    <jboss-transaction-manager/>
```
marianogonzalez: why?

tests/functional/src/main/java/org/mule/tck/junit4/ApplicationContextBuilder.java
elrodro83: Will remove this from DomainContextBuilder also.

tests/integration/src/test/resources/org/mule/test/integration/exceptions/exception-dlq-flow.xml
elrodro83: This tx manager was not being initialized for this test. Now that the init is eager, it was failing because of a JNDI lookup.

tests/integration/src/test/resources/org/mule/test/integration/exceptions/exception-dlq-service.xml
elrodro83: This tx manager was not being initialized for this test. Now that the init is eager, it was failing because of a JNDI lookup.

marianogonzalez: i don't think you can just remove it. How do we know arjuna still works with your  change?

PR #2451, opened 2016-03-03
Update Maven plugin versions

asequeira:


aiannucci:
What about maven-gpg-plugin version? (line 2050)

aiannucci:
Shouldn't we have a JIRA issue for this?

aiannucci:
The maven-compiler-plugin and maven-javadoc-plugin properties should be grouped with the other maven ones and can be updated as well

PR #2452, opened 2016-03-04
MULE-9389: Mule Context's getTransactionManager randomly returns null if

elrodro83:
called during context start - Better Solution


distributions/standalone/src/main/resources/MIGRATION.txt
```diff
@@ -30,7 +30,7 @@ MULE-6298: AbstractMessageSequenceSplitter#process will now return null instead
 MULE-9306: Losing flow and session variables when using MuleClient to dispatch/send an event. Exception strategy not caching exceptions after using MuleClient on a JavaComponent.
 MULE-9383: HTTP Connector will allow invalid Content-Type header values. In order to strictly validate them as before, use the mule.strictContentType=true system property.
 MULE-9405: MuleException instead of NPE is now thrown when setting a null value for a context-property in an XstlTransformation.
-MULE-9389: Transaction manager configuration now occurs during context start. It will not be available during initialize phase.
+MULE-9389: MuleTransactionManagerFactoryBean should be removed. Transaction manager now will be injected without it. Also, injection of TransactionManager from TransactionManagerFactory must not be done manually.
```
marianogonzalez: this is not clear. What do you mean the factory bean should be removed? from where? how come the user even knows about it?

modules/spring-config/src/main/java/org/mule/config/spring/factories/TransactionManagerFactoryBean.java
```diff
@@ -0,0 +1,90 @@
...
+ * <p/>
+ * Creates a reference to the TransactionManager configured on the MuleContext. This is useful when you need to inject
+ * the TransactionManager into other objects such as XA data Sources.
+ * <p/>
+ * This will first look for a single {@link TransactionManagerFactory} and use it to build the
+ * {@link TransactionManager}.
+ * <p/>
+ * If no {@link TransactionManagerFactory} is found, then it will look for {@link TransactionManager} instances.
+ */
+public class TransactionManagerFactoryBean implements FactoryBean<TransactionManager>, MuleContextAware
```
marianogonzalez: add @ since

PR #2453, opened 2016-03-04
MULE-9426: Refactor plugin classloading features

pablokraan:
_ Adding mule-module-artifact module
_ Moving classloaders and related classes to artifact module
_ Extracting CompositeClassLoader from CompositeApplicationClassLoader
_ Extracting Artifactdjescriptor from PluginDescriptor
_ Refactoring descriptors from application plugin and application
_ Renaming classes
_ Cleaning tests


pablolagreca:
Changes look good. My only question is if this module should have an API? Or is everything part of the API?. Ideally only those things that need to be exposed to launcher should be part of the API.

pablokraan:
yes, this will have an API, but that will be part of another story. I want to separate those concerns so I can move faster on this.

PR #2454, opened 2016-03-04
MULE-9387: Add placement and group metadata to parameters

juandesi:


modules/extensions-support/src/main/java/org/mule/module/extension/internal/introspection/describer/AnnotationsBasedDescriber.java
```diff
@@ -292,17 +290,34 @@ private void declareAnnotatedParameters(Class<?> annotatedType, Descriptor descr
             if (!parameters.isEmpty())
             {
                 ParameterGroup group = new ParameterGroup(field.getType(), field);
+                PlacementModelProperty groupPlacement = null;
                 groups.add(group);
 
+                if (field.isAnnotationPresent(Placement.class))
```
marianogonzalez: why do you test isPresent and then get it? Isn't it better to get it and then test for null?

marianogonzalez:
add support for operation parameters

PR #2465, opened 2016-03-08
MULE-9444 Upgrade Saxon to 9.7.0.3-HE

gsfernandes:


tests/integration/src/test/resources/org/mule/test/integration/xml/xslt-functional-test-flow.xml
```diff
@@ -17,7 +17,7 @@
     <!--// START SNIPPET: transformer-config -->
     <mulexml:xslt-transformer name="xslt">
         <mulexml:xslt-text>
-            <xsl:stylesheet version="2.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
+            <xsl:stylesheet version="2.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns="">
```
marianogonzalez: this looks weird... why?

tests/integration/src/test/resources/org/mule/test/integration/xml/xslt-functional-test-service.xml
```diff
@@ -17,7 +17,7 @@
     <!--// START SNIPPET: transformer-config -->
     <mulexml:xslt-transformer name="xslt">
         <mulexml:xslt-text>
-            <xsl:stylesheet version="2.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
+            <xsl:stylesheet version="2.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns="">
```
marianogonzalez: idem

pom.xml
```diff
@@ -228,7 +228,7 @@
         <rhinoVersion>1.7R4</rhinoVersion>
         <rhinoJsr223Version>1.1</rhinoJsr223Version>
         <saajVersion>1.3</saajVersion>
-        <saxonVersion>9.6.0-1</saxonVersion>
+        <saxonVersion>9.6.0-7</saxonVersion>
```
marianogonzalez: 9.7.0-1 is available, which also includes support for XPath 3.1. Why don't you go to the latest path version instead of upgrading to the latest minor?

tests/integration/src/test/resources/org/mule/test/integration/xml/xslt-functional-test-flow.xml
gsfernandes: Because xsl stylesheet is embedded as it is on this sample configuration the xmlns used for the output transformation would be the one set in the xmlns attribute or the default one that is inherited by the xml where the stylesheet is defined. Here it would be mule/core due to it is set as default xmlns because it is embedded in mule configuration. It is actually better to have this in a separate xml.

This is how currently works but when you look at the junit that is doing the assertion for this transformation it is not validating the xmlns attribute in output because of the conflicting libraries issue with XmlUnit, therefore with Saxon upgraded version here the test started to fail when comparing expected and result transformation.

tests/integration/src/test/resources/org/mule/test/integration/xml/xslt-functional-test-service.xml
gsfernandes: Same as previous comment. Here is the service configuration for the same junit integration test.

pom.xml
gsfernandes: This PR will fix a few issues reported and solved in Saxon 9.6, for Saxon 9.7 and XPath 3.1 I have created another enhancement MULE-9452.

tests/integration/src/test/resources/org/mule/test/integration/xml/xslt-functional-test-flow.xml
gsfernandes: I have confirmed that this "issue" is not the same behaviour as it is in Saxon 9.7.0-x. But at the same time this is they way it was working since Saxon 9.6 was upgraded. So, I would say keep this as it is in order to avoid breaking compatibility and during the upgrade for Saxon 9.7 in order to support XPath 3.1 write this breaking change in migration guideline.
What do you think?

PR #2466, opened 2016-03-08
MULE-9160: Upgrade to Log4j latest version

elrodro83:


modules/launcher/src/main/java/org/mule/module/launcher/log4j2/LoggerContextConfigurer.java
```diff
@@ -116,26 +117,21 @@ private void disableShutdownHook(LoggerContext context)
...
-                        (Reconfigurable) configuration,
-                        configFile,
-                        getListeners(configuration),
-                        DEFAULT_MONITOR_INTERVAL_SECS));
-            }
+        if (configFile != null && configuration instanceof Reconfigurable)
+        {
+            configuration.getWatchManager().setIntervalSeconds(DEFAULT_MONITOR_INTERVAL_SECS);
+            FileWatcher watcher = new ConfiguratonFileWatcher((Reconfigurable) configuration, getListeners(configuration));
+            configuration.getWatchManager().watchFile(configFile, watcher);
```
marianogonzalez: have you empirically tested that this still works? we have no automatic coverage for this

elrodro83: Yes, i tested the correct refresh of the config. That's how i found the issue with the script engines.

PR #2475, opened 2016-03-11
MULE-9444 Upgrade Saxon to 9.7.0.3-HE

gsfernandes:
Added migration entry


gsfernandes:
Closing this PR as it won't need to be added to migration file due to there is no impact in how xpath was used in previous version of mule.

gsfernandes:
Saxon is available to users in order to build custom components and there was a change in 9.7 that required to modify the way Mule uses Saxon through JAXP API so the migration guide had to be changed in order to notify users about this.

PR #2478, opened 2016-03-14
MULE-9466: Add query timeout to DB connector operations

pablokraan:


modules/db/src/test/java/org/mule/module/db/integration/model/MySqlTestDatabase.java
```diff
@@ -142,6 +142,21 @@ public void createStoredProcedureConcatenateStrings(DataSource dataSource) throw
     }
 
     @Override
+    public void createDelayFunction(DataSource dataSource) throws SQLException
+    {
+        executeDdl(dataSource, "DROP FUNCTION IF EXISTS DELAY;\n");
+
+        final String sql =
+                "" +
```
elrodro83: Can this empty line be removed so the formatting is consistent with OracleTestDatabase?

modules/db/src/test/java/org/mule/module/db/integration/AbstractQueryTimeoutTestCase.java
```diff
@@ -0,0 +1,69 @@
...
+        assertThat(response.getPayload(), is(not(instanceOf(NullPayload.class))));
+
+        try
+        {
+            runFlow(QUERY_TIMEOUT_FLOW, 5);
+            fail("Expected query to timeout");
+        }
+        catch (MessagingException e)
+        {
+            // Expected
```
elrodro83: Can do some assertion on the thrown exception? Maybe it should be delegated to the config to account for the different db engines.

PR #2481, opened 2016-03-16
MULE-9361 Remove invocation and session property scopes from MuleMessage

dfeist:


core-tests/src/test/java/org/mule/MessagePropertiesContextTestCase.java
```diff
@@ -121,36 +94,18 @@ public void testCopyConstructor() throws Exception
         assertSame(mpc.getProperty("ABC", PropertyScope.OUTBOUND), "abc");
     }
 
-    /*@Test
-    public void testInboundScopeIsImmutable() throws Exception
```
pablokraan: Is not this test valid?

core-tests/src/test/java/org/mule/MuleEventTestCase.java
```diff
@@ -200,40 +197,32 @@ public void testFlowVarNamesRemoveMutable() throws Exception
     public void testFlowVarsNotShared() throws Exception
```
pablokraan: Not sure if this test is still relevant

core-tests/src/test/java/org/mule/MuleMessageDataTypePropagationTestCase.java
```diff
@@ -344,43 +343,49 @@ public void setsDefaultInboundScopePropertyDataType() throws Exception
...
-    public void setsDefaultInvocationPropertyDataType() throws Exception
+    public void setsDefaultFlowVariableDataType() throws Exception
     {
-        DefaultMuleMessage muleMessage = new DefaultMuleMessage(TEST, muleContext);
-        muleMessage.setInvocationProperty(TEST_PROPERTY, TEST);
+        MuleEvent muleEvent = MuleTestUtils.getTestEvent(TEST, muleContext);
+        muleEvent.setFlowVariable(TEST_PROPERTY, TEST);
 
-        assertDefaultPropertyDataType(muleMessage, PropertyScope.INVOCATION);
+        DataType<?> actualDataType = muleEvent.getFlowVariableDataType(TEST_PROPERTY);
```
pablokraan: Extract this lines into a assertVariableDataType, similar to assertPropertyDataType

core-tests/src/test/java/org/mule/MuleMessageDataTypePropagationTestCase.java
```diff
@@ -344,43 +343,49 @@ public void setsDefaultInboundScopePropertyDataType() throws Exception
     }
 
     @Test
-    public void setsDefaultInvocationPropertyDataType() throws Exception
+    public void setsDefaultFlowVariableDataType() throws Exception
     {
-        DefaultMuleMessage muleMessage = new DefaultMuleMessage(TEST, muleContext);
-        muleMessage.setInvocationProperty(TEST_PROPERTY, TEST);
+        MuleEvent muleEvent = MuleTestUtils.getTestEvent(TEST, muleContext);
```
pablokraan: Add static imports

core-tests/src/test/java/org/mule/MuleSessionHandlerTestCase.java
```diff
@@ -87,25 +90,24 @@ public static void restoreEncoding()
...
-        handler.storeSessionInfoToMessage(session, message);
+        handler.storeSessionInfoToMessage(event.getSession(), event.getMessage());
         // store save session to outbound, move it to the inbound
         // for retrieve to deserialize
-        Object s = removeProperty(message);
-        message.setInboundProperty(MuleProperties.MULE_SESSION_PROPERTY, s);
-        session = handler.retrieveSessionInfoFromMessage(message);
+        Object s = removeProperty(event);
+        ((DefaultMuleMessage) event.getMessage()).setInboundProperty(MuleProperties.MULE_SESSION_PROPERTY, s);
+        MuleSession session = handler.retrieveSessionInfoFromMessage(event.getMessage());
```
pablokraan: I'm not sure about this: session lives in the event, but the session handler is reading/writing from/to the message. Is that right?

core-tests/src/test/java/org/mule/MuleSessionHandlerTestCase.java
```diff
@@ -87,25 +90,24 @@ public static void restoreEncoding()
...
 
-        handler.storeSessionInfoToMessage(session, message);
+        handler.storeSessionInfoToMessage(event.getSession(), event.getMessage());
         // store save session to outbound, move it to the inbound
         // for retrieve to deserialize
-        Object s = removeProperty(message);
-        message.setInboundProperty(MuleProperties.MULE_SESSION_PROPERTY, s);
-        session = handler.retrieveSessionInfoFromMessage(message);
+        Object s = removeProperty(event);
+        ((DefaultMuleMessage) event.getMessage()).setInboundProperty(MuleProperties.MULE_SESSION_PROPERTY, s);
```
pablokraan: Is this cast required?

core-tests/src/test/java/org/mule/el/mvel/datatype/AbstractVarAssignmentDataTypePropagatorTestCase.java
```diff
@@ -50,7 +50,7 @@ protected AbstractVarAssignmentDataTypePropagatorTestCase(EnricherDataTypePropag
         this.dataTypePropagator = dataTypePropagator;
     }
 
-    protected void doAssignmentDataTypePropagationTest(PropertyScope scope, String expression) throws Exception
+    protected void doFlowVarAssignmentDataTypePropagationTest(String expression) throws Exception
```
pablokraan: Does not make sense to me to have these methods on an parent class as they are used only in an specific subclass (flowVar vs session)

core-tests/src/test/java/org/mule/el/mvel/datatype/SessionVarExpressionDataTypeResolverTestCase.java
```diff
@@ -7,13 +7,87 @@
...
+        final MVELExpressionLanguageContext context = createMvelExpressionLanguageContext(testEvent, parserConfiguration);
+
+        CompiledExpression compiledExpression = (CompiledExpression) compileExpression(expression, new ParserContext(parserConfiguration));
+        // Expression must be executed, otherwise the variable accessor is not properly configured
+        MVEL.executeExpression(compiledExpression, context);
+
+        assertThat(expressionDataTypeResolver.resolve(testEvent, compiledExpression), like(String.class, JSON, CUSTOM_ENCODING));
+    }
+
+    protected MVELExpressionLanguageContext createMvelExpressionLanguageContext(MuleEvent testEvent, ParserConfiguration parserConfiguration)
```
pablokraan: This method and the test look like a duplicate of FlowVarExpressionDataTypeResolverTestCase

core-tests/src/test/java/org/mule/el/mvel/datatype/FlowVarExpressionDataTypeResolverTestCase.java
```diff
@@ -7,13 +7,88 @@
...
+        final MVELExpressionLanguageContext context = createMvelExpressionLanguageContext(testEvent, parserConfiguration);
+
+        CompiledExpression compiledExpression = (CompiledExpression) compileExpression(expression, new ParserContext(parserConfiguration));
+        // Expression must be executed, otherwise the variable accessor is not properly configured
+        MVEL.executeExpression(compiledExpression, context);
+
+        assertThat(expressionDataTypeResolver.resolve(testEvent, compiledExpression), like(String.class, JSON, CUSTOM_ENCODING));
+    }
+
+    protected MVELExpressionLanguageContext createMvelExpressionLanguageContext(MuleEvent testEvent, ParserConfiguration parserConfiguration)
```
pablokraan: Another duplicated block of code?

core-tests/src/test/java/org/mule/el/mvel/datatype/PropertyEnricherDataTypePropagatorTestCase.java
```diff
@@ -37,17 +34,24 @@
     @Test
     public void propagatesDataTypeForInlinedInvocationProperty() throws Exception
     {
-        doInlinePropertyDataTypePropagationTest(INVOCATION);
+        final DataType expectedDataType = DataTypeFactory.create(String.class, JSON);
```
pablokraan: In all of these session/flowVar test, looks to me that the only different part is how the test reads/writes data. Can these test be refactored to avoid the duplication?

core-tests/src/test/java/org/mule/routing/MapSplitterTestCase.java
```diff
@@ -62,8 +59,8 @@ public void testSplit() throws Exception
         assertTrue(splitPayloads.contains("three"));
 
         assertEquals(3, splitPayloads.size());
-        assertEquals("1", splitKeyProperties.get(splitPayloads.indexOf("one")));
-        assertEquals("2", splitKeyProperties.get(splitPayloads.indexOf("two")));
+//        assertEquals("1", splitKeyProperties.get(splitPayloads.indexOf("one")));
```
pablokraan: Fix or remove these lines

core-tests/src/test/java/org/mule/routing/UntilSuccessfulTestCase.java
```diff
@@ -133,6 +134,7 @@ public void testSuccessfulDeliveryAckExpression() throws Exception
         untilSuccessful.start();
 
         final MuleEvent testEvent = getTestEvent("test_data");
+        ((DefaultMuleMessage)testEvent.getMessage()).assertAccess(true);
```
pablokraan: Is ugly to have to use a cast in order to assert the result of this test. Also, fix the code format

core-tests/src/test/java/org/mule/routing/correlation/EventCorrelatorTestCase.java
```diff
@@ -185,7 +195,7 @@ private EventCorrelator createEventCorrelator() throws Exception
         when(mockMuleContext.getRegistry().get(MuleProperties.OBJECT_STORE_MANAGER)).thenReturn(mockObjectStoreManager);
         memoryObjectStore.store(TEST_GROUP_ID, mockEventGroup, "prefix.eventGroups");
         when(mockEventGroup.getGroupId()).thenReturn(TEST_GROUP_ID);
-        when(mockEventGroup.toMessageCollection()).thenReturn(null);
+        //when(mockEventGroup.toMessageCollection()).thenReturn(null);
```
pablokraan: Fix this

core-tests/src/test/java/org/mule/routing/filters/ExpressionFilterTestCase.java
```diff
@@ -100,20 +96,19 @@ public void testHeaderFilterWithNotNullEL() throws Exception
         assertTrue(filter.accept(message));
     }
 
-    @Ignore("MULE-9341")
```
pablokraan: Why the tests is fixed without fixing MULE-9341?

core-tests/src/test/java/org/mule/transformer/simple/AddSessionVariableTransformerTestCase.java
```diff
@@ -0,0 +1,202 @@
...
+import javax.activation.MimeTypeParseException;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mockito;
+import org.mockito.internal.verification.VerificationModeFactory;
+
+@SmallTest
+public class AddSessionVariableTransformerTestCase extends AbstractMuleTestCase
```
pablokraan: looks like there is a too much duplication between AddFlowVariableTransformerTestCase , AddPropertyTransformerTestCase and AddSessionVariableTransformerTestCase classes

core-tests/src/test/java/org/mule/transformer/simple/RemoveFlowVariableTransformerTest.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.HashSet;
+
+import org.junit.Before;
+import org.junit.Ignore;
+import org.junit.Test;
+import org.mockito.Mockito;
+import org.mockito.internal.verification.VerificationModeFactory;
+
+@SmallTest
+public class RemoveFlowVariableTransformerTest extends AbstractMuleTestCase
```
pablokraan: Rename to RemoveFlowVariableTransformerTestCase

core-tests/src/test/java/org/mule/transformer/simple/RemoveSessionVariableTransformerTest.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.HashSet;
+
+import org.junit.Before;
+import org.junit.Ignore;
+import org.junit.Test;
+import org.mockito.Mockito;
+import org.mockito.internal.verification.VerificationModeFactory;
+
+@SmallTest
+public class RemoveSessionVariableTransformerTest extends AbstractMuleTestCase
```
pablokraan: Remove duplication on this and related test cases

core/src/main/java/org/mule/PropertyScope.java
```diff
@@ -10,25 +10,14 @@
 import java.util.Comparator;
 
 /**
- * A PropertyScope is used to associate a message property with a lifetime.  A scope may be very 
- * brief such as {@link #INVOCATION} which only lasts until a service has been invoked or longer 
- * running such as {@link #SESSION}.
+ * A PropertyScope is used to associate a message property with a lifetime.
  */
 public final class PropertyScope implements Serializable
```
pablokraan: If session and invocation scopes are removed, does not make sense to just remove this class and directly use getInboundProperties/getOutboundProperties(name) instead of getProperty(name, scope)?

core/src/main/java/org/mule/api/routing/filter/Filter.java
```diff
@@ -20,5 +21,18 @@
      * @param message a non null message to filter.
      * @return <code>true</code> if the message matches the filter
      */
+    @Deprecated
     boolean accept(MuleMessage message);
+
+    /**
+     * Check a given event against this filter.
```
pablokraan: Check > Checks

core/src/main/java/org/mule/component/BindingInvocationHandler.java
```diff
@@ -75,7 +74,9 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
         MuleMessage message = createMuleMessage(args);
 
         // Some transports such as Axis, RMI and EJB can use the method information
-        message.setInvocationProperty(MuleProperties.MULE_METHOD_PROPERTY, method.getName());
+
+        // TODO MULE-9449 Remove Java Component Bindings
+        //message.setInvocationProperty(MuleProperties.MULE_METHOD_PROPERTY, method.getName());
```
pablokraan: Looks like the binding won't work anymore after this change, so why to keep this commented?

core/src/main/java/org/mule/el/context/FlowVariableMapContext.java
```diff
@@ -0,0 +1,78 @@
...
+ */
+package org.mule.el.context;
+
+import org.mule.api.MuleEvent;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+public class FlowVariableMapContext extends AbstractMapContext<String, Object>
```
pablokraan: Add javadoc

pablokraan: As I didn't know about AbstractMapContext, I took a look and seems to me that MessagePropertyMapContext should be updated as it can manage only outbound properties now.

core/src/main/java/org/mule/el/context/SessionVariableMapContext.java
```diff
@@ -0,0 +1,78 @@
...
+ */
+package org.mule.el.context;
+
+import org.mule.api.MuleSession;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+public class SessionVariableMapContext extends AbstractMapContext<String, Object>
```
pablokraan: Add javadocs

pablokraan: Can you refactor SessionVariableMapContext and FlowVariableMapContext to reuse as much code as possible?

core/src/main/java/org/mule/el/mvel/datatype/FlowVarEnricherDataTypePropagator.java
```diff
@@ -17,6 +18,16 @@
 
     public FlowVarEnricherDataTypePropagator()
     {
-        super("flowVars", PropertyScope.INVOCATION);
+        super("flowVars");
+    }
+
+    protected void addVariable(MuleEvent event, TypedValue typedValue, String propertyName)
```
pablokraan: Add missing @Override

core/src/main/java/org/mule/el/mvel/datatype/FlowVarEnricherDataTypePropagator.java
```diff
@@ -17,6 +18,16 @@
 
     public FlowVarEnricherDataTypePropagator()
     {
-        super("flowVars", PropertyScope.INVOCATION);
+        super("flowVars");
```
pablokraan: can you use org.mule.el.mvel.MessageVariableResolverFactory.FLOW_VARS?

core-tests/src/test/java/org/mule/MessagePropertiesContextTestCase.java
dfeist: From what i remember this immutability was never implemented, but seems the test method was commented out instead of removed.

core-tests/src/test/java/org/mule/MuleEventTestCase.java
dfeist: It is, this is about if two events share the same flowVars map or not I believe.

core-tests/src/test/java/org/mule/MuleSessionHandlerTestCase.java
dfeist: Yes, but i can refactor to use a different method that doesn't need cast.

core-tests/src/test/java/org/mule/MuleSessionHandlerTestCase.java
dfeist: Yes, as it serializes to/from outbound property of message.

core/src/main/java/org/mule/el/mvel/datatype/PropertyEnricherDataTypePropagator.java
```diff
@@ -20,29 +22,24 @@
...
-                scope = PropertyScope.SESSION;
+                event.setFlowVariable(assignmentVar, typedValue.getValue(), typedValue.getDataType());
+                return true;
             }
-
-            if (scope != null)
+            else if (event.getSession().getPropertyNamesAsSet().contains(assignmentVar))
             {
-                message.setProperty(assignmentVar, typedValue.getValue(), scope, typedValue.getDataType());
+                event.getSession().setProperty(assignmentVar, (Serializable) typedValue.getValue(), typedValue.getDataType());
```
pablokraan: Should not check that the value is serializable before casting?

core/src/main/java/org/mule/el/mvel/datatype/SessionVarEnricherDataTypePropagator.java
```diff
@@ -17,6 +20,17 @@
 
     public SessionVarEnricherDataTypePropagator()
     {
-        super("sessionVars", PropertyScope.SESSION);
+        super("sessionVars");
+    }
+
+    protected void addVariable(MuleEvent event, TypedValue typedValue, String propertyName)
+    {
+        event.getSession().setProperty(propertyName, (Serializable) typedValue.getValue(), typedValue.getDataType());
```
pablokraan: No check for typeValue.getValue instanceOf serializable?

core/src/main/java/org/mule/model/resolvers/MethodHeaderPropertyEntryPointResolver.java
```diff
@@ -56,8 +55,8 @@ public InvocationResult invoke(Object component, MuleEventContext context) throw
 
         //TODO MULE-4953 I don't think the VM transport if managing invocation properties correctly, or maybe it is and this
```
pablokraan: Not related to your comment, but MULE-4953 is fixed and we don't have VM transport anymore. Can you remove it?

core/src/main/java/org/mule/processor/TransactionalInterceptingMessageProcessor.java
```diff
@@ -83,6 +86,10 @@ public void initialise() throws InitialisationException
         {
             this.exceptionListener = muleContext.getDefaultExceptionStrategy();
         }
+        if (this.exceptionListener instanceof FlowConstructAware)
+        {
+            ((FlowConstructAware)(this.exceptionListener)).setFlowConstruct(flowConstruct);
```
pablokraan: (this.exceptionListener) > this.exceptionListener

core/src/main/java/org/mule/routing/AbstractSplitter.java
```diff
@@ -33,6 +32,6 @@
         return new CollectionMessageSequence(splitMessage(event));
     }
 
-    protected abstract List<MuleMessage> splitMessage(MuleEvent event) throws MuleException;
+    protected abstract List<MuleEvent> splitMessage(MuleEvent event) throws MuleException;
```
pablokraan: This looks now as a splitEvent method. Rename it

core/src/main/java/org/mule/routing/correlation/EventCorrelator.java
```diff
@@ -371,18 +370,9 @@ protected void handleGroupExpiry(EventGroup group) throws MessagingException
...
-            try
-            {
-                messageCollection = group.toMessageCollection();
-            }
-            catch (ObjectStoreException e)
-            {
-                throw new MessagingException(group.getMessageCollectionEvent(), e);
-            }
-            muleContext.fireNotification(new RoutingNotification(messageCollection, null,
+            MuleEvent messageCollection = group.getMessageCollectionEvent();
```
pablokraan: messageCollection variable and messageCollectionEvent method sound weird to me. Are they talking about an event or a message?

core/src/main/java/org/mule/routing/filters/ExpressionFilter.java
```diff
@@ -103,18 +101,7 @@ public boolean accept(MuleMessage message)
             Thread.currentThread().setContextClassLoader(expressionEvaluationClassLoader);
 
             // TODO MULE-9341 Remove Filters. Expression filter will be replaced by something that uses MuleEvent.
```
pablokraan: Do this TODO still apply?

core/src/main/java/org/mule/routing/filters/ExpressionFilter.java
```diff
@@ -124,6 +111,34 @@ public boolean accept(MuleMessage message)
...
+    /**
+     * Check a given event against this filter.
+     *
+     * @param event a non null event to filter.
+     * @return <code>true</code> if the event matches the filter
+     */
+    @Override
+    public boolean accept(MuleEvent event)
+    {
+        // MULE-4797 Because there is no way to specify the class-loader that script
```
pablokraan: Is code was copied/pasted from the accept(MuleMessage message) method. I would like you to remove that duplication and ask you why there is still code for that other accept method?

core/src/main/java/org/mule/routing/filters/RegExFilter.java
```diff
@@ -100,6 +101,26 @@ public boolean accept(MuleMessage message)
         }
     }
 
+    @Override
+    public boolean accept(MuleEvent event)
```
pablokraan: Similar to previous comment: method accept(MuleMessage) is still there. Why?

core/src/main/java/org/mule/transformer/simple/AbstractAddVariablePropertyTransformer.java
```diff
@@ -68,17 +67,21 @@ public Object transformMessage(MuleEvent event, String outputEncoding) throws Tr
...
                 {
-                    event.getMessage().setProperty(key, typedValue.getValue(), getScope(), typedValue.getDataType());
+                    addProperty(event, key, typedValue.getValue(), typedValue.getDataType());
                 }
             }
         }
         return event.getMessage();
     }
 
+    protected abstract void addProperty(MuleEvent event, String propertyName, Object value, DataType dataType);
```
pablokraan: Would be good to have javadoc on abstract methods to clearly indicate what they are supposed to do

core/src/main/java/org/mule/transformer/simple/AbstractRemoveVariablePropertyTransformer.java
```diff
@@ -38,25 +39,20 @@ public Object transformMessage(MuleEvent event, String outputEncoding) throws Tr
...
+            wildcardAttributeEvaluator.processValues(getPropertyNames(event), matchedValue -> {
+                removeProperty(event, matchedValue);
+                if (logger.isDebugEnabled())
                 {
-                    event.getMessage().removeProperty(matchedValue,getScope());
-                    if (logger.isDebugEnabled())
-                    {
-                        logger.debug(String.format("Removing property: '%s' from scope: '%s'", matchedValue, getScope().getScopeName()));
-                    }
+                    logger.debug(String.format("Removing property: '%s'", matchedValue));
```
pablokraan: Can you add the scope of the removed property?

core/src/main/java/org/mule/transformer/simple/AddPropertyTransformer.java
```diff
@@ -6,19 +6,27 @@
...
-    @Override
-    protected PropertyScope getScope()
+    public void setPropertyName(String propertyName)
     {
-        return PropertyScope.OUTBOUND;
+        this.setIdentifier(propertyName);
     }
 
-    public void setPropertyName(String propertyName)
+    protected void addProperty(MuleEvent event, String propertyName, Object value, DataType dataType)
```
pablokraan: Missing override annotations

core/src/main/java/org/mule/transformer/simple/AddFlowVariableTransformer.java
```diff
@@ -6,20 +6,27 @@
...
 public class AddFlowVariableTransformer extends AbstractAddVariablePropertyTransformer
 {
 
-    @Override
-    protected PropertyScope getScope()
-    {
-        return PropertyScope.INVOCATION;
-    }
-    
     public void setVariableName(String variableName)
```
pablokraan: Is strange to have setPropertyName and setVariableName here

core/src/main/java/org/mule/transformer/simple/AddSessionVariableTransformer.java
```diff
@@ -6,20 +6,29 @@
...
-import org.mule.PropertyScope;
+import org.mule.api.MuleEvent;
+import org.mule.api.metadata.DataType;
+
+import java.io.Serializable;
 
 public class AddSessionVariableTransformer extends AbstractAddVariablePropertyTransformer
 {
 
+    public void setVariableName(String variableName)
```
pablokraan: Is strange in general to have methods talking about "variable" and others talking about "property"

core/src/main/java/org/mule/transformer/simple/RemoveFlowVariableTransformer.java
```diff
@@ -6,18 +6,27 @@
...
 
-import org.mule.PropertyScope;
+import org.mule.api.MuleEvent;
+
+import java.util.Set;
 
 public class RemoveFlowVariableTransformer extends AbstractRemoveVariablePropertyTransformer
 {
+
+    public void setVariableName(String variableName)
```
pablokraan: Add missing override annotations in all the overridden methods

modules/scripting/src/main/java/org/mule/module/scripting/component/Scriptable.java
```diff
@@ -260,34 +260,35 @@ public void populateBindings(Bindings bindings, MuleEvent event)
...
         if (message == null)
         {
             message = new DefaultMuleMessage(NullPayload.getInstance(), muleContext);
         }
 
-        populateVariablesInOrder(bindings, message);
+        populateVariablesInOrder(bindings, event);
 
-        bindings.put(BINDING_MESSAGE, message);
+        bindings.put(BINDING_MESSAGE, event.getMessage());
```
pablokraan: I'm not sure about this: in the past users could set properties using the message. Now that is not possible as they live on the event. Does not this mean that now we need a way to access the event from the bindings?

modules/scripting/src/main/java/org/mule/module/scripting/component/Scriptable.java
```diff
@@ -260,34 +260,35 @@ public void populateBindings(Bindings bindings, MuleEvent event)
...
+        populateHeadersVariablesAndException(bindings, event);
     }
 
-    private void populateHeadersVariablesAndException(Bindings bindings, MuleMessage message)
+    private void populateHeadersVariablesAndException(Bindings bindings, MuleEvent event)
     {
-        bindings.put(BINDING_FLOW_VARS, new MesssagePropertyMap(message, INVOCATION));
-        bindings.put(BINDING_SESSION_VARS, new MesssagePropertyMap(message, SESSION));
+        bindings.put(BINDING_FLOW_VARS, new FlowVariableMapContext(event));
+        bindings.put(BINDING_SESSION_VARS, new SessionVariableMapContext(event.getSession()));
```
pablokraan: I think this answers my previous question

modules/scripting/src/test/java/org/mule/module/scripting/GreedyTestCase.java
```diff
@@ -24,14 +22,12 @@ protected String getConfigFile()
         return "greedy-config.xml";
     }
 
-    @Ignore("MULE-9341")
```
pablokraan: Same question as before regarding MULE-9341

modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -3283,10 +3283,7 @@
                     </xsd:annotation>
                     <xsd:simpleType>
                         <xsd:restriction base="xsd:NMTOKEN">
-                            <xsd:enumeration value="invocation"/>
                             <xsd:enumeration value="outbound"/>
```
pablokraan: This definition could be simplified

tests/integration/src/test/java/org/mule/test/construct/FlowConfigurationFunctionalTestCase.java
```diff
@@ -315,10 +315,6 @@ public void testSplitAggregateMapFlow() throws Exception
         assertTrue(apple.isBitten());
         assertTrue(banana.isBitten());
         assertTrue(orange.isBitten());
-
-        assertNotNull(results[0].getProperty("key", PropertyScope.INVOCATION));
```
pablokraan: Should not be the invocation property still set on the event?

tests/integration/src/test/java/org/mule/test/integration/exceptions/ExceptionHandlingTestCase.java
```diff
@@ -132,7 +131,9 @@ public void testOutboundEndpointInScope() throws Exception
         MuleMessage response = client.request("test://outScope2", 3000);
 
         assertNotNull(response);
-        assertTrue(response.getProperty("expectedHandler", PropertyScope.SESSION));
+
+        FlowAssert.verify("outboundEndpointInScope");
```
pablokraan: don't forget static imports

tests/integration/src/test/resources/foreach-test.xml
```diff
@@ -51,9 +51,7 @@
     <flow name="parent-message-config">
         <foreach rootMessageVariableName="parent">
             <test:component/>
-            <message-properties-transformer scope="invocation">
```
pablokraan: Makes sense to maintain message-properties-transformer or it can be also removed as part of this change?

core/src/main/java/org/mule/routing/MessageFilter.java
```diff
@@ -90,7 +90,7 @@ protected boolean accept(MuleEvent event)
 
         if (event != null && !VoidMuleEvent.getInstance().equals(event))
```
pablokraan: MessageFilter should be renamed to EventFilter

core/src/main/java/org/mule/DefaultMuleMessage.java
```diff
@@ -475,18 +468,6 @@ public void setInboundProperty(String key, Object value, DataType<?> dataType)
     }
 
     @Override
-    public void setInvocationProperty(String key, Object value)
```
pablokraan: DefaultMuleMessage still have setProperty methods. Should not they be private instead of public?

core/src/main/java/org/mule/DefaultMuleMessage.java
```diff
@@ -1607,16 +1565,6 @@ private MuleMessage copyToInbound(DefaultMuleMessage currentMessage, Object payl
         return newMessage;
     }
 
-    void setSessionProperties(Map<String, TypedValue> sessionProperties)
-    {
-        properties.sessionMap = sessionProperties;
-    }
-
-    void setInvocationProperties(Map<String, TypedValue> invocationProperties)
```
pablokraan: Not sure about this: before these changes, invocation properties were serialized with the message. If that is right: what happens now?

core-tests/src/test/java/org/mule/routing/filters/ExpressionFilterTestCase.java
dfeist: Because as part of this work, i added support for Event with expression filter as it was needed for something else.

core/src/main/java/org/mule/DefaultMuleMessage.java
dfeist: They are still used widely in tests to set inbound properties. Best to clean this up in a different issue/pr/

core/src/main/java/org/mule/DefaultMuleMessage.java
dfeist: These methods were used for the event to inject whole map into the message when setMessage() is called, this is no longer needed.

core/src/main/java/org/mule/PropertyScope.java
dfeist: setProperty(name, INBOUND) is still used widely in tests.  It makes sense to clean this up, but not in this PR.

core/src/main/java/org/mule/component/BindingInvocationHandler.java
dfeist: In order to remove all bindings code in a single PR separately rather than contaminating this commit with other changes.  Rather than delete part of the code and not the rest, decided it was best to comment with a TODO.

core/src/main/java/org/mule/el/context/FlowVariableMapContext.java
dfeist: Why does MessagePropertyMapContext need updating?

core/src/main/java/org/mule/routing/AbstractSplitter.java
dfeist: The purpose of this method is still to split the message, as before, it just happens to receive event as context.

core/src/main/java/org/mule/routing/MessageFilter.java
dfeist: Don't disagree but don't see how this is related to this PR, especially as currently Filter interface still supports Message and Event and it is accept(Message msg) that implementers must implement still.

core/src/main/java/org/mule/routing/filters/ExpressionFilter.java
dfeist: Yes, because this method still needs to be supports it's implementing is creating an event to wrap message.

core/src/main/java/org/mule/transformer/simple/AddFlowVariableTransformer.java
dfeist: This is used by spring i imagine, ideally should be using an alias rather than adding this methods..

modules/scripting/src/main/java/org/mule/module/scripting/component/Scriptable.java
dfeist: Yes I'd say so..

tests/integration/src/test/java/org/mule/test/construct/FlowConfigurationFunctionalTestCase.java
dfeist: No, this is not supported as we aggregate messages and not events.

tests/integration/src/test/resources/foreach-test.xml
dfeist: Makes sense to remove, but as part of seperate PR.

core-tests/src/test/java/org/mule/el/mvel/datatype/PropertyEnricherDataTypePropagatorTestCase.java
dfeist: I reworked all these tests now

core-tests/src/test/java/org/mule/MuleMessageDataTypePropagationTestCase.java
```diff
@@ -344,43 +343,45 @@ public void setsDefaultInboundScopePropertyDataType() throws Exception
...
-    public void setsDefaultInvocationPropertyDataType() throws Exception
+    public void setsDefaultFlowVariableDataType() throws Exception
     {
-        DefaultMuleMessage muleMessage = new DefaultMuleMessage(TEST, muleContext);
-        muleMessage.setInvocationProperty(TEST_PROPERTY, TEST);
+        MuleEvent muleEvent = getTestEvent(TEST, muleContext);
+        muleEvent.setFlowVariable(TEST_PROPERTY, TEST);
 
-        assertDefaultPropertyDataType(muleMessage, PropertyScope.INVOCATION);
+        assertVariableDataType(muleEvent, DataType.STRING_DATA_TYPE);
```
pablokraan: Always use static imports

core/src/main/java/org/mule/PropertyScope.java
pablokraan: Can you create a jira for that?

core/src/main/java/org/mule/component/BindingInvocationHandler.java
pablokraan: I'm OK with adding a  TODO, but in the class. Remove this commented line

core/src/main/java/org/mule/el/context/FlowVariableMapContext.java
pablokraan: forget about my previous comment, I misread the code

core/src/main/java/org/mule/el/context/SessionVariableMapContext.java
```diff
@@ -0,0 +1,81 @@
...
+package org.mule.el.context;
+
+import org.mule.api.MuleSession;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Flow variables wrapper {@link Map} for exposing Flow variables via an {@link org.mule.api.el.ExpressionLanguage}
```
pablokraan: Copy and paste is a bad practice

core/src/main/java/org/mule/PropertyScope.java
dfeist: https://www.mulesoft.org/jira/browse/MULE-9504

core/src/main/java/org/mule/routing/AbstractSplitter.java
pablokraan: That is not clear from the API, as now it takes an event and return a list of events, in the past it took a Message and returned a list of messages. I think the name must change

core/src/main/java/org/mule/api/routing/filter/Filter.java
```diff
@@ -15,10 +16,23 @@
...
 {
     /**
-     * Check a given message against this filter.
+     * Checks a given message against this filter.
      * 
      * @param message a non null message to filter.
      * @return <code>true</code> if the message matches the filter
      */
+    @Deprecated
     boolean accept(MuleMessage message);
```
pablokraan: This method must be removed, not deprecated

core/src/main/java/org/mule/transformer/simple/AbstractAddVariablePropertyTransformer.java
```diff
@@ -68,17 +67,35 @@ public Object transformMessage(MuleEvent event, String outputEncoding) throws Tr
...
                 }
             }
         }
         return event.getMessage();
     }
 
+    /**
+     * Adds the property with its value and dataType to a property or variables scope.
+     *
+     * @param event
```
pablokraan: add param description

modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -3275,20 +3272,12 @@
                         </xsd:documentation>
                     </xsd:annotation>
                 </xsd:attribute>
-                <xsd:attribute name="scope" use="optional" default="outbound">
+                <xsd:attribute name="scope" use="optional" fixed="outbound">
                     <xsd:annotation>
                         <xsd:documentation>
-                            Property scope to/from which properties are added/removed.  The scope determines the lifespan of the properties. Default scope is outbound.
+                            Property scope to/from which properties are added/removed.  The only scope supported is outbound.
```
pablokraan: What is the meaning of having an attribute that only supports one value?

tests/integration/src/test/resources/foreach-test.xml
pablokraan: Can you create another jira for that?

core/src/main/java/org/mule/routing/MessageFilter.java
pablokraan: Is related as you added another method to filter an event and deprecating the other one, which I think must be removed instead

core/src/main/java/org/mule/DefaultMuleMessage.java
pablokraan: Please, add another jira then

core-tests/src/test/java/org/mule/el/mvel/datatype/FlowVarExpressionDataTypeResolverTestCase.java
```diff
@@ -7,88 +7,20 @@
...
-    {
-        doVarDataTypeTest(variableName + "." + PROPERTY_NAME );
-    }
-
-    @Test
-    public void returnsFlowVarDataTypeUsingEscapedDotSyntax() throws Exception
+    public FlowVarExpressionDataTypeResolverTestCase()
     {
-        doVarDataTypeTest(variableName + ".'" + PROPERTY_NAME + "'");
+        super(new FlowVarExpressionDataTypeResolver(),  "flowVars");
```
pablokraan: Use the constant for "flowVars"

core-tests/src/test/java/org/mule/el/mvel/datatype/SessionVarExpressionDataTypeResolverTestCase.java
```diff
@@ -7,87 +7,22 @@
...
-    {
-        doVarDataTypeTest(variableName + "." + PROPERTY_NAME );
-    }
 
-    @Test
-    public void returnsFlowVarDataTypeUsingEscapedDotSyntax() throws Exception
+    public SessionVarExpressionDataTypeResolverTestCase()
     {
-        doVarDataTypeTest(variableName + ".'" + PROPERTY_NAME + "'");
+        super(new SessionVarExpressionDataTypeResolver(), "sessionVars");
```
pablokraan: Use the constant for "sessionVars"

core-tests/src/test/java/org/mule/transformer/simple/RemoveSessionVariableTransformerTestCase.java
```diff
@@ -0,0 +1,44 @@
...
+@SmallTest
+public class RemoveSessionVariableTransformerTestCase extends AbstractRemoveVariablePropertyTransformerTestCase
+{
+    public RemoveSessionVariableTransformerTestCase()
+    {
+        super(new RemoveSessionVariableTransformer());
+    }
+
+    @Override
+    protected void addMockedPropeerties(MuleEvent mockEvent, HashSet properties)
```
pablokraan: mockEvent > event

core/src/main/java/org/mule/api/routing/filter/Filter.java
dfeist: It is still used, but will be removed as part of MULE-9341.  Will add a TODO.

tests/integration/src/test/resources/foreach-test.xml
dfeist: https://www.mulesoft.org/jira/browse/MULE-9506

core/src/main/java/org/mule/DefaultMuleMessage.java
dfeist: https://www.mulesoft.org/jira/browse/MULE-9504

core-tests/src/test/java/org/mule/transformer/simple/RemoveSessionVariableTransformerTestCase.java
dfeist: Why? these methods clearly need to receive a mock because of their name and what they do, so don't see an issue with using 'mockEvent'.

tests/integration/src/test/java/org/mule/test/construct/FlowConfigurationFunctionalTestCase.java
dfeist: I reviewed this, this behavior actually makes sense, split Map messages have a key invocation property in Mule 3, and while this is not support in Mule 4 currently (MULE-9502) there is no reason this should be returned as a response once the responses are aggregated.

tests/integration/src/test/java/org/mule/test/integration/exceptions/ExceptionHandlingTestCase.java
dfeist: Can't static import FlowAssert.verify and Mockito.verrify.  In this cases where both are used it's best not to static import either and be explicit.

PR #2482, opened 2016-03-17
MULE-9457: Update of log4j2 library caused a regression in multiple

elrodro83:
Scenarios

### DIFF
--- a/modules/launcher/src/main/java/org/mule/module/launcher/log4j2/LoggerContextCache.java
+++ b/modules/launcher/src/main/java/org/mule/module/launcher/log4j2/LoggerContextCache.java
@@ -8,6 +8,7 @@ package org.mule.module.launcher.log4j2;
 
 import static java.util.concurrent.Executors.newScheduledThreadPool;
 import static org.mule.config.i18n.MessageFactory.createStaticMessage;
+
 import org.mule.api.MuleRuntimeException;
 import org.mule.api.config.MuleProperties;
 import org.mule.api.lifecycle.Disposable;
@@ -21,6 +22,7 @@ import com.google.common.collect.ImmutableList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
@@ -64,6 +66,9 @@ final class LoggerContextCache implements Disposable
     private static final long DEFAULT_DISPOSE_DELAY_IN_MILLIS = 15000;
 
     private final ArtifactAwareContextSelector artifactAwareContextSelector;
+    // Extra cache layer to avid some nasty implications for using guava cache at this point. See the comments in
+    // #doGetLoggerContext(final ClassLoader classLoader) for details.
+    private final Map<Integer, LoggerContext> builtContexts = new ConcurrentHashMap<>();
     private final Cache<Integer, LoggerContext> activeContexts;
     private final Cache<Integer, LoggerContext> disposedContexts;
     private final ScheduledExecutorService executorService;
@@ -84,6 +89,7 @@ final class LoggerContextCache implements Disposable
                     {
                         stop(notification.getValue());
                         activeContexts.invalidate(notification.getKey());
+                        builtContexts.remove(notification.getKey());
                     }
                 })
                 .build();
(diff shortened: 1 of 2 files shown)


modules/launcher/src/main/java/org/mule/module/launcher/log4j2/LoggerContextCache.java
```diff
@@ -64,6 +66,8 @@
     private static final long DEFAULT_DISPOSE_DELAY_IN_MILLIS = 15000;
 
     private final ArtifactAwareContextSelector artifactAwareContextSelector;
+    // Extra cache layer to avid some nasty implications for using guava cache at this point.
```
marianogonzalez: nasty implications is not very descriptive. Since this is a very subtle bug in which a new developer could easily fall into again, I would add more documentation around this.... Maybe consider extracting this to a class with a generous javadoc

modules/launcher/src/main/java/org/mule/module/launcher/log4j2/LoggerContextCache.java
```diff
@@ -83,7 +87,8 @@
                     public void onRemoval(RemovalNotification<Integer, LoggerContext> notification)
                     {
                         stop(notification.getValue());
-                        activeContexts.invalidate(notification.getKey());
+                                               activeContexts.invalidate(notification.getKey());
```
marianogonzalez: maybe a github bug, but formatting doesn't look good here

modules/launcher/src/main/java/org/mule/module/launcher/log4j2/LoggerContextCache.java
```diff
@@ -121,15 +126,45 @@ LoggerContext getLoggerContext(final ClassLoader classLoader)
         final LoggerContext ctx;
         try
         {
-            final int key = computeKey(classLoader);
-            ctx = activeContexts.get(key, new Callable<LoggerContext>()
+            final Integer key = computeKey(classLoader);
```
marianogonzalez: why move from a native to a wrapper?

modules/launcher/src/main/java/org/mule/module/launcher/log4j2/MuleLoggerContext.java
```diff
@@ -73,13 +80,24 @@
...
+            loggersField = null;
+        }
     }
 
     @Override
     public synchronized void reconfigure()
     {
-        super.reconfigure();
         applyContainerConfiguration();
+        super.reconfigure();
```
marianogonzalez: why the swap?

modules/launcher/src/main/java/org/mule/module/launcher/log4j2/MuleLoggerContext.java
```diff
@@ -73,13 +80,24 @@
             applicationClassloader = false;
             artifactName = null;
         }
+
+        Collection<Field> candidateFields = getAllFields(LoggerContext.class, withName("loggers"));
```
marianogonzalez: why?

modules/launcher/src/main/java/org/mule/module/launcher/log4j2/MuleLoggerContext.java
```diff
@@ -129,6 +147,45 @@ public String getName()
...
+     *            logger but will log a warning if mismatched.
+     * @return The Logger.
+     */
+    @Override
+    public Logger getLogger(final String name, final MessageFactory messageFactory)
+    {
+        Logger logger = lookupLoggersField().get(name);
+        if (logger != null)
+        {
+            AbstractLogger.checkMessageFactory(logger, messageFactory);
```
marianogonzalez: why not skipping this check at all?

modules/launcher/src/main/java/org/mule/module/launcher/log4j2/MuleLoggerContext.java
elrodro83: As of log4j 2.5, the configuration has to be set before the lifcecycle management of the log4j objects (that lifecycle is hadled by the reconfigure of the superclass). This seems to have been a benign issue before the upgrade.

modules/launcher/src/main/java/org/mule/module/launcher/log4j2/MuleLoggerContext.java
elrodro83: Will do.

modules/launcher/src/main/java/org/mule/module/launcher/log4j2/MuleLoggerContext.java
elrodro83: the loggers field is private in the superclass. This allows to override the #getLogger(final String name, final MessageFactory messageFactory) method.

modules/launcher/src/main/java/org/mule/module/launcher/log4j2/LoggerContextCache.java
elrodro83: Just so the wrapping is done at this point only and not each time it is used in call to a Map method.

PR #2483, opened 2016-03-17
MULE-9465: QName with invalid character is generated when using an In…

pablomusumeci:
…ner Enum class as parameter

### DIFF
--- a/modules/extensions-spring-support/src/test/resources/heisenberg-operation-config.xml
+++ b/modules/extensions-spring-support/src/test/resources/heisenberg-operation-config.xml
@@ -160,7 +160,9 @@
     </flow>
 
     <flow name="killWithWeapon">
-        <heisenberg:kill-with-weapon weapon="#[weapon]"/>
+        <heisenberg:kill-with-weapon weapon="#[weapon]" type="MELEE_WEAPON">
+            <heisenberg:weapon-attributes brand="Pizza on the rooftop"/>
+        </heisenberg:kill-with-weapon>
     </flow>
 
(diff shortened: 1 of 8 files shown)


modules/extensions-spring-support/src/test/resources/heisenberg-operation-config.xml
```diff
@@ -160,7 +160,7 @@
     </flow>
 
     <flow name="killWithWeapon">
-        <heisenberg:kill-with-weapon weapon="#[weapon]"/>
+        <heisenberg:kill-with-weapon weapon="#[weapon]" type="MELEE_WEAPON"/>
```
marianogonzalez: add another test case... this one was intentionally using a expression

marianogonzalez: add another test which also tests an inner pojo class

modules/extensions-support/src/main/java/org/mule/module/extension/internal/util/NameUtils.java
```diff
@@ -325,4 +326,14 @@ public String replace(String word)
             return java.util.regex.Pattern.compile(pattern, flags).matcher(word).replaceAll(replacement);
         }
     }
+
+    /**
+     * Removes everything that's not a word, a dot nor a hyphen
+     * @param originalName name that needs the removal of invalid characters
+     * @return name without invalid characters
+     */
+    public static String getCuratedName(String originalName)
```
marianogonzalez: getCuratedName -> curateName

PR #2484, opened 2016-03-17
MULE-9464: No TypeIdAnnotation for operation migrated through DMT

marianogonzalez:


modules/extensions-spring-support/src/main/java/org/mule/module/extension/internal/capability/xml/schema/builder/SchemaBuilder.java
```diff
@@ -509,7 +510,7 @@ private void visitEnum(StringType stringType)
                 Class<? extends Enum> enumType;
                 try
                 {
-                    enumType = getType(stringType);
+                    enumType = JavaTypeUtils.getType(stringType);
```
elrodro83: Static import

marianogonzalez: done

PR #2487, opened 2016-03-18
MULE-9498: Timezone attribute in Quartz connector and scheduler module

elrodro83:


modules/schedulers/src/main/resources/META-INF/mule-schedulers.xsd
```diff
@@ -26,6 +26,13 @@
                         </xsd:documentation>
                     </xsd:annotation>
                 </xsd:attribute>
+                <xsd:attribute name="timeZone" type="xsd:string">
```
afelisatti: Should this support properties?

elrodro83: It already does defined as it is. The cron expression is declared in the same way and it is tested to support expressions.

modules/schedulers/src/test/java/org/mule/modules/schedulers/cron/CronsTimeZoneSchedulerTest.java
```diff
@@ -0,0 +1,63 @@
...
+        Collection<Scheduler> schedulers = muleContext.getRegistry().lookupScheduler(
+                Schedulers.flowConstructPollingSchedulers("pollWithTimeZone"));
+        
+        assertThat(schedulers, hasSize(1));
+        for (Scheduler scheduler : schedulers)
+        {
+            assertThat(scheduler, instanceOf(CronScheduler.class));
+            // Just can assert that the tx in the config gets to the mule scheduler. To instrospect the quatz object to
+            // check its there would be very complex.
+            assertThat(((CronScheduler) scheduler).getTimeZone(), is(TimeZone.getTimeZone("America/Argentina/Buenos_Aires")));
```
afelisatti: Static import

transports/quartz/src/test/resources/quartz-namespace-config-flow.xml
```diff
@@ -79,6 +79,25 @@
     </flow>
     <!--// END SNIPPET: example-1 -->
 
+    <flow name="testService1Timezone">
+        <description>
+            This configuration will create an inbound event for testService1 at 12
+            noon GMT every day. The event payload
```
afelisatti: Shouldn't this be Buenos Aires time?

modules/schedulers/src/main/resources/META-INF/mule-schedulers.xsd
afelisatti: Is this optional?

elrodro83: Yes, attributes are optional by default in xsd

PR #2491, opened 2016-03-21
MULE-9429: Update artifact export policy to support packages and fold…

pablokraan:
…ers only

_ Extracting loader override parsing from FineGrainedControlClassLoader to ClassLoaderLookupPolicy and ClassLoaderLookupPolicyFactory
_ Updating class loaders to use ClassLoaderLookupPolicy
_ Replacing ArtifactDescriptor loaderOverrides and blockedPrefixNames with a ClassLoaderLookupPolicy instance
_ Improving ClassLoaderFilter to differentiate between class and resource packages
_ Replacing ArtifactDescriptor exportedPrefixNames with a ClassLoaderFilter instance
_ Updating FilteringClassLoader to filter using the new functionality
_ Clean up


modules/artifact/src/main/java/org/mule/module/artifact/classloader/ArtifactClassLoaderFilter.java
```diff
@@ -23,49 +25,72 @@
...
+        return exportedResourcePackages.contains(resourcePackage);
     }
 
-    private boolean isExportedPrefix(String name)
+    private String getResourceFolder(String resourceName)
     {
-        return hasListedPrefix(name, descriptor.getExportedPrefixNames());
+        if (resourceName == null)
+        {
+            return EMPTY_PACKAGE;
```
dfeist: Does this make sense, why would you get the resource folder for "null"?  Or is this simply preferred approach than failing with an exception?

modules/artifact/src/main/java/org/mule/module/artifact/classloader/ArtifactClassLoaderFilter.java
```diff
@@ -23,49 +25,72 @@
...
+        final String resourcePackage = getResourceFolder(name);
 
-    private boolean isExportedClass(String name)
-    {
-        return descriptor.getExportedPrefixNames().contains(name);
+        return exportedResourcePackages.contains(resourcePackage);
     }
 
-    private boolean isExportedPrefix(String name)
+    private String getResourceFolder(String resourceName)
```
dfeist: getResourcePackage?  Given seems the string returned is dot separated and not a path?

modules/artifact/src/main/java/org/mule/module/artifact/classloader/ArtifactClassLoaderFilter.java
```diff
@@ -23,49 +25,72 @@
...
         {
-            if (name.startsWith(exported))
-            {
-                return true;
-            }
+            return EMPTY_PACKAGE;
+        }
+        else
+        {
+            return (className.lastIndexOf(PACKAGE_SEPARATOR) < 0) ? EMPTY_PACKAGE : className.substring(0, className.lastIndexOf('.'));
```
dfeist: Use constant for package separator.

modules/artifact/src/main/java/org/mule/module/artifact/classloader/ArtifactClassLoaderFilterFactory.java
```diff
@@ -0,0 +1,56 @@
...
+    private Set<String> getPackages(String exportedPackages)
+    {
+        Set<String> exported = new HashSet<>();
+        if (StringUtils.isNotBlank(exportedPackages))
+        {
+            final String[] exports = exportedPackages.split(",");
+            for (String export : exports)
+            {
+                export = export.trim();
+                if (export.startsWith("/"))
```
dfeist: Isn't this already handled in ArtifactClassLoaderFilter, does it need handling in both places?  Also here it's being applied to classes and resources but in ArtifactClassLoaderFilter only resources I think..

modules/artifact/src/main/java/org/mule/module/artifact/classloader/ClassLoaderLookupPolicyFactory.java
```diff
@@ -0,0 +1,98 @@
...
+     * a system package.
+     */
+    public ClassLoaderLookupPolicy create(String config)
+    {
+        Set<String> overridden = new HashSet<>();
+        Set<String> blocked = new HashSet<>();
+
+        if (!StringUtils.isEmpty(config))
+        {
+            final String[] overrides = config.split(",");
```
dfeist: Use constant?

modules/artifact/src/main/java/org/mule/module/artifact/classloader/ClassLoaderLookupPolicyFactory.java
```diff
@@ -0,0 +1,98 @@
...
+
+                    // 'blocked' package definitions come with a '-' prefix
+                    boolean isBlocked = override.startsWith("-");
+                    if (isBlocked)
+                    {
+                        override = override.substring(1);
+                    }
+
+                    String dottedOverride;
+                    if (override.endsWith("."))
```
dfeist: Use constants everywhere, especially when same string literal is used multiple times in same class.

modules/artifact/src/main/java/org/mule/module/artifact/classloader/FineGrainedControlClassLoader.java
```diff
@@ -6,66 +6,22 @@
...
-import org.mule.util.StringUtils;
-
 import java.net.URL;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Set;
 
-/**
- * TODO document overrides, blocked, systemPackages and syntax for specifying those.
- */
```
dfeist: Javadoc?  What does this implementation do, and when should it be used?

modules/artifact/src/main/java/org/mule/module/artifact/classloader/ArtifactClassLoaderFilterFactory.java
pablokraan: Is similar, but not the same. This is parsing the configuration and normalizing the data to create the filter. Then, the filter processes parameter data to check against the configured packages

modules/artifact/src/test/java/org/mule/module/artifact/classloader/ClassLoaderLookupPolicyTestCase.java
```diff
@@ -0,0 +1,86 @@
...
+package org.mule.module.artifact.classloader;
+
+import org.mule.tck.junit4.AbstractMuleTestCase;
+import org.mule.tck.size.SmallTest;
+
+@SmallTest
+public class ClassLoaderLookupPolicyTestCase extends AbstractMuleTestCase
+{
+
+    /*
```
dfeist: Uncomment or remove. Or is there a very good reason to leave this commented?

modules/artifact/src/test/java/org/mule/module/artifact/classloader/FineGrainedControlClassLoaderTestCase.java
```diff
@@ -109,9 +37,8 @@ public void childResolvesOverriden() throws Exception
     {
         URLClassLoader parent = new URLClassLoader(new URL[] { hello() }, Thread.currentThread().getContextClassLoader());
 
-        Set<String> overrides = new HashSet<>();
-        overrides.add("mypackage");
-        FineGrainedControlClassLoader ext = new FineGrainedControlClassLoader(new URL[] { bye() }, parent, overrides);
+        final ClassLoaderLookupPolicy lookupPolicy = new ClassLoaderLookupPolicy(singleton("mypackage"), emptySet());
```
dfeist: Minor: ideally use constant for "mypackage"

modules/artifact/src/test/java/org/mule/module/artifact/classloader/ClassLoaderLookupPolicyTestCase.java
pablokraan: I don't know what happened with this code... maybe I didn't pushed my latest changes before I created the PR

PR #2496, opened 2016-03-22
MULE-9503 Add support for non-blocking dispatchers to transports

dfeist:


core/src/main/java/org/mule/processor/EndpointTransactionalInterceptingMessageProcessor.java
```diff
@@ -36,7 +38,7 @@ public MuleEvent process(final MuleEvent event) throws MuleException
         {
             return event;
         }
-        else
+        else if(TransactionCoordination.getInstance().getTransaction() != null)
```
pablokraan: I'm not sure about this, but seems to me that if EndpointTransactionalInterceptingMessageProcessor is used, then is because there is a transaction in place. If my assumption is correct, the right change would be to avoid using this class.

dfeist: This chain that includes processor is constructed once for the endpoint but if a composite-source is being used, transactions may or may not be in used for a specific request, that is why this is needed here.

core/src/main/java/org/mule/execution/MessageProcessorNotificationExecutionInterceptor.java
```diff
@@ -120,7 +120,7 @@ public void processExceptionReplyTo(MessagingException exception, Object replyTo
         }
         finally
         {
-            if (!NonBlockingVoidMuleEvent.getInstance().equals(result) && fireNotification)
+            if (!(result instanceof VoidMuleEvent) && fireNotification)
```
pablokraan: Why is this change required? Was not possible to receive a VoidMuleEvent before this PR?

dfeist: This now matches sub-clasess of VoidMuleEvent (such as NonBlockingVoidEvent) which it didn't before.  This wasn't required before as endpoints didn't support non-blocking.  The other updates like this are for the same reason.

core/src/main/java/org/mule/transport/AbstractMessageDispatcher.java
```diff
@@ -213,4 +255,88 @@ protected void applyResponseTransformers(MuleEvent event) throws MuleException
     protected abstract void doDispatch(MuleEvent event) throws Exception;
 
     protected abstract MuleMessage doSend(MuleEvent event) throws Exception;
+
+    protected void doSendNonBlocking(MuleEvent event, CompletionHandler<MuleMessage, Exception> completionHandler)
+    {
+        throw new IllegalStateException("This MessageDispatcher does not support non-blocking");
+    }
+
+    private class NonBlockingSendCompletionHandler implements CompletionHandler<MuleMessage, Exception>
```
pablokraan: Would be better to have this class a a top level element and have it properly tested

core/src/test/java/org/mule/endpoint/outbound/OutboundEndpointTestCase.java
```diff
@@ -80,12 +89,26 @@ public void testDefaultFlowSync() throws Exception
...
+    {
+        OutboundEndpoint endpoint = createOutboundEndpoint(null, null, null, null, MessageExchangePattern.REQUEST_RESPONSE, null);
+
+        SensingNullReplyToHandler nullReplyToHandler = new SensingNullReplyToHandler();
+        MuleEvent event = getNonBlockingTestEventUsingFlow(TEST_MESSAGE, nullReplyToHandler);
+
+        MuleEvent response = endpoint.process(event);
+        assertThat(response, CoreMatchers.<MuleEvent>equalTo(NonBlockingVoidMuleEvent.getInstance()));
+
+        assertThat(getNonBlockingResponse(nullReplyToHandler), CoreMatchers.equalTo(event));
```
pablokraan: Use static imports

core/src/test/java/org/mule/endpoint/outbound/OutboundNotificationMessageProcessorTestCase.java
```diff
@@ -66,4 +66,26 @@ public void testSend() throws Exception
...
     }
 
+    @Test
+    public void testSendNonBlocking() throws Exception
+    {
+        TestEndpointMessageNotificationListener listener = new TestEndpointMessageNotificationListener();
+        muleContext.registerListener(listener);
+
+        OutboundEndpoint endpoint = createTestOutboundEndpoint(null, null, null, null,
+                                                               MessageExchangePattern.REQUEST_RESPONSE, null);
```
pablokraan: always use static imports...

core/src/test/java/org/mule/endpoint/outbound/OutboundNotificationMessageProcessorTestCase.java
```diff
@@ -66,4 +66,26 @@ public void testSend() throws Exception
...
+        MessageProcessor mp = new OutboundNotificationMessageProcessor(endpoint);
+        SensingNullReplyToHandler nullReplyToHandler = new SensingNullReplyToHandler();
+        MuleEvent event = getNonBlockingTestEventUsingFlow(TEST_MESSAGE, nullReplyToHandler);
+        mp.process(event);
+
+        assertTrue(listener.latch.await(RECEIVE_TIMEOUT, TimeUnit.MILLISECONDS));
```
pablokraan: Use assertThat

core/src/test/java/org/mule/processor/TestNonBlockingProcessor.java
```diff
@@ -4,7 +4,7 @@
  * license, a copy of which has been included with this distribution in the
  * LICENSE.txt file.
  */
-package org.mule.tck.functional;
+package org.mule.processor;
```
pablokraan: Why is this change of package required?

dfeist: As I'm now using this in mule-core.

core/src/test/java/org/mule/tck/junit4/AbstractMuleContextTestCase.java
```diff
@@ -611,4 +617,25 @@ protected File getFileInsideWorkingDirectory(String fileName)
     {
         return new File(getWorkingDirectory(), fileName);
     }
+
+    protected MuleEvent getNonBlockingTestEventUsingFlow(Object payload, ReplyToHandler replyToHandler) throws Exception
```
pablokraan: When merging this to mule4.x, these methods should be on TestEventBuilder

core/src/test/java/org/mule/tck/testmodels/mule/TestMessageDispatcher.java
```diff
@@ -73,4 +95,10 @@ protected void doStop()
     {
         // no op
     }
+
+    @Override
+    protected boolean isSupportsNonBlocking()
```
pablokraan: Is not better to have a NonBlockingTestMessageDispatcher extending from TestMessageDispatcher? (asking as this class is used in some tests that would expect this method to return false)

core/src/main/java/org/mule/transport/AbstractMessageDispatcher.java
dfeist: I'd prefer to maintain this a private inner class.  I'll look at adding some more test tests to cover more of this though..

core/src/test/java/org/mule/tck/junit4/AbstractMuleContextTestCase.java
dfeist: This PR won't be merged to Mule 4, as Mule 4 has no transports.

core/src/test/java/org/mule/tck/testmodels/mule/TestMessageDispatcher.java
dfeist: I spent a while moving to the approach you suggest but it just complicates functional testing, and doesn't add much value and there are no existing tests that depend on this.  We currently also use a single dispatcher for testing success/failure caes, rather than two different dispatchers, this is similar.

PR #2497, opened 2016-03-22
MULE-9503 Add support for non-blocking dispatchers to transports

dfeist:


core/src/test/java/org/mule/tck/junit4/AbstractMuleContextTestCase.java
```diff
@@ -611,4 +617,33 @@ protected File getFileInsideWorkingDirectory(String fileName)
...
+            }
+            if (replyToHandler.exception != null)
+            {
+                throw replyToHandler.exception;
+            }
+            return replyToHandler.event;
+        }
+        else
+        {
+            return result;
```
gsfernandes: I encountered in an issue due to this change, in previous version it was always returning the replyToHandler.event as result of invoking a flow as nonBlocking, therefore if the flow config didn't have a nonblocking component (for instance a test:non-blocking) I got a null response.
Now with this change if you run a flow and the flow doesn't have a non-blocking component the response is always returned, ended up having for instance a "wrong test" due to you are supposed to get the response through a replyToHandler mechanism but instead you got the response obtained as result of the flow invocation.
In my case the test "was conceptually wrong" due to I was not getting the response via replyToHandler but I was actually not testing that, just the serialization of the event in cache.

Was this change intended to modify the behaviour for FunctionalTestCase nonBlocking support from now on?
If that's the case I think that it would be good to have the same behaviour in all the branches.

PR #2499, opened 2016-03-23
MULE-9500: Allow specifying custom log4j config locations

elrodro83:


modules/launcher/src/test/resources/log4j2-test-custom.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
marianogonzalez: this file doesn't seem referenced anywhere

modules/launcher/src/main/java/org/mule/module/launcher/log4j2/ArtifactAwareContextSelector.java
```diff
@@ -180,10 +183,30 @@ public void execute()
 
     private URI getArtifactLoggingConfig(ArtifactClassLoader muleCL)
```
marianogonzalez: extract the entire buildContext() method (which includes this sub rutine) into a factory which you can unit test using mock application descriptors and classloaders. In that way you can test all this logic without relying on classpath discovery of test log4j config files which we know ends up in flackyness

modules/launcher/src/test/resources/log4j2-test-custom.xml
elrodro83: Redid the test that uses this file.

modules/launcher/src/main/java/org/mule/module/launcher/log4j2/ArtifactAwareContextSelector.java
```diff
@@ -57,10 +47,11 @@
 class ArtifactAwareContextSelector implements ContextSelector, Disposable
 {
 
-    private static final StatusLogger logger = StatusLogger.getLogger();
+    static final StatusLogger logger = StatusLogger.getLogger();
 
     private LoggerContextCache cache = new LoggerContextCache(this, getClass().getClassLoader());
 
+    private static MuleLoggerContextFactory loggerContextFactory = new MuleLoggerContextFactory();
```
marianogonzalez: static state is evil

PR #2507, opened 2016-03-25
MULE-9503 Add support for non-blocking dispatchers to transports. Fix…

dfeist:
… couple of issues causing test failures.


core/src/main/java/org/mule/processor/EndpointTransactionalInterceptingMessageProcessor.java
```diff
@@ -38,7 +39,8 @@ public MuleEvent process(final MuleEvent event) throws MuleException
         {
             return event;
         }
-        else if(TransactionCoordination.getInstance().getTransaction() != null)
+        // This optimization is not valid with Services because of how outbound routers work
+        else if(TransactionCoordination.getInstance().getTransaction() != null || event.getFlowConstruct() instanceof Service)
```
pablokraan: Fix code style

PR #2508, opened 2016-03-29
MULE-9431: Invert classloading lookup process for deployed artifacts  

pablokraan:
_ Adding getClassLoaderLookupPolicy to ArtifactClassloader to propagate the right lookup policy
_ Adding ClassLoaderLookupMethod and refactoring ClassLoaderLookupPolicy to return a lookupMethod for a given class name
_ Extracting CompositeArtifactClassLoader from CompositeApplicationClassLoader in order to reuse it
_ Updating class loaders to return the used lookup policy
_ Removing lookup policy form artifact descriptors
_ Removing the ability to configure an application using different types of mule-deploy files, as only properties have been used so far.
_ Configuring a simple container lookup policy on deployment service


modules/artifact/src/main/java/org/mule/module/artifact/classloader/ClassLoaderLookupMethod.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Indicates how to resolve class loading lookup related to parent/child relationships
+ * between two different classLoaders.
+ */
+public enum ClassLoaderLookupMethod
+{
+
+    /**
+     * Class will be lookup on the parent classloader first. If the class is found,
```
dfeist: "lookup on" -> "looked up from"

modules/artifact/src/main/java/org/mule/module/artifact/classloader/ClassLoaderLookupMethod.java
```diff
@@ -0,0 +1,38 @@
...
+ * LICENSE.txt file.
+ */
+
+package org.mule.module.artifact.classloader;
+
+/**
+ * Indicates how to resolve class loading lookup related to parent/child relationships
+ * between two different classLoaders.
+ */
+public enum ClassLoaderLookupMethod
```
dfeist: Think "Strategy" might be clearer than "Method"?

modules/artifact/src/main/java/org/mule/module/artifact/classloader/CompositeClassLoader.java
```diff
@@ -32,45 +36,85 @@
     protected static final Log logger = LogFactory.getLog(CompositeClassLoader.class);
 
     protected final List<ClassLoader> classLoaders;
-
-    public CompositeClassLoader(ClassLoader parent, List<ClassLoader> classLoaders)
+    private final ClassLoaderLookupPolicy lookupPolicy;
+
+    /**
```
dfeist: Update class javadoc to explain how with CL uses lookup policy rather than simply stating that it delegates to a list of CL's.

modules/artifact/src/main/java/org/mule/module/artifact/classloader/ArtifactClassLoader.java
```diff
@@ -58,4 +58,9 @@
      */
     void addShutdownListener(ShutdownListener listener);
 
+    /**
+     * @return lookup policy used on the classLoader. Non null.
+     */
+    ClassLoaderLookupPolicy getClassLoaderLookupPolicy();
```
dfeist: Seems there are two or three hierarchies all with this method, anyway to clean this up? At least there are three definitions or implementations without override annotation.

modules/launcher/src/main/java/org/mule/module/launcher/application/MuleApplicationClassLoaderFactory.java
```diff
@@ -40,21 +44,68 @@
 public class MuleApplicationClassLoaderFactory implements ArtifactClassLoaderFactory<ApplicationDescriptor>
 {
 
+    public static final String CLASS_SUFFIX = ".class";
```
dfeist: SUFFIX -> EXTENSION?

modules/launcher/src/main/java/org/mule/module/launcher/application/MuleApplicationClassLoaderFactory.java
```diff
@@ -40,21 +44,68 @@
...
+        if (!plugins.isEmpty())
+        {
+            // Re-assigns parent classloader if there are shared plugin libraries
+            final Map<String, ClassLoaderLookupMethod> lookupMethods;
+            URL[] pluginLibs = descriptor.getSharedPluginLibs();
+            if (pluginLibs != null && pluginLibs.length != 0)
+            {
+                lookupMethods = getLookMethodsFrom(pluginLibs);
+                lookupPolicy = lookupPolicy.extend(lookupMethods);
+                parent = new MuleArtifactClassLoader("sharedLibs", pluginLibs, parent.getClassLoader(), lookupPolicy);
```
dfeist: Constant?

PR #2514, opened 2016-03-31
MULE-9533 JMS Outbound endpoint does not respect response timeout

dfeist:


transports/jms/src/main/java/org/mule/transport/jms/JmsMessageDispatcher.java
```diff
@@ -186,6 +186,8 @@ private MuleMessage dispatchMessage(MuleEvent event, boolean doSend) throws Exce
 
             if (useReplyToDestination && replyTo != null)
             {
+                final int timeout = endpoint.getResponseTimeout();
```
pablokraan: Should not check if the endpoint timeout is set and if not, then use the event timeout?

transports/jms/src/test/java/org/mule/transport/jms/integration/JmsTemporaryReplyToTestCase.java
```diff
@@ -22,43 +24,81 @@
...
+
+    private static String ECHO_FLOW_NAME = "EchoFlow";
+
     @Override
     protected String getConfigFile()
     {
         return "integration/jms-temporary-replyTo.xml";
     }
 
+    // Fixed Reply Queue
```
pablokraan: Can you avoid these kind of "test group separator" comments and improve test names instead?

transports/jms/src/main/java/org/mule/transport/jms/JmsMessageDispatcher.java
pablokraan: This change in the behaviour must be properly documented on the migration guide

transports/jms/src/test/resources/integration/jms-temporary-replyTo.xml
```diff
@@ -1,86 +1,68 @@
...
+      xmlns:jms="http://www.mulesoft.org/schema/mule/jms"
+      xmlns:test="http://www.mulesoft.org/schema/mule/test"
+      xsi:schemaLocation="
        http://www.mulesoft.org/schema/mule/test http://www.mulesoft.org/schema/mule/test/current/mule-test.xsd
-       http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-current.xsd
        http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd
-       http://www.mulesoft.org/schema/mule/vm http://www.mulesoft.org/schema/mule/vm/current/mule-vm.xsd">
+       http://www.mulesoft.org/schema/mule/jms http://www.mulesoft.org/schema/mule/jms/current/mule-jms.xsd">
 
-    <vm:connector name="vmConnector"/>
```
pablokraan: Should not the original test using VM queues also available?

transports/jms/src/main/java/org/mule/transport/jms/JmsMessageDispatcher.java
dfeist: There is no way to check if endpoint value is set (by user) or not, as it is an int with a default value.  This behaviour while different to Mule 3.1 and before is consistent with other connectors like HTTP.

The fact that timeout is now respected I don't think belongs in the migration guide, but the fact that you can no longer customize JMS request-response timeout via event.setTimeout() i'll add a note on.

transports/jms/src/test/java/org/mule/transport/jms/integration/JmsTemporaryReplyToTestCase.java
dfeist: I improved test names too eventually. Can remove this..

transports/jms/src/test/resources/integration/jms-temporary-replyTo.xml
dfeist: Why? It is only used to call flow as we used to do for all tests, it's not part of the test case. Same for getting response and testing response instead of using <test:assert/>

PR #2516, opened 2016-03-31
MULE-9534 JMS outbound endpoint does not use Mule correlationId for J…

dfeist:
…MS message correlationId

### DIFF
--- a/core/src/main/java/org/mule/client/DefaultLocalMuleClient.java
+++ b/core/src/main/java/org/mule/client/DefaultLocalMuleClient.java
@@ -36,6 +36,7 @@ import org.mule.config.i18n.CoreMessages;
 import org.mule.endpoint.SimpleEndpointCache;
 import org.mule.exception.DefaultMessagingExceptionStrategy;
 import org.mule.management.stats.FlowConstructStatistics;
+import org.mule.routing.MuleMessageInfoMapping;
 
 import java.util.Map;
 
@@ -234,6 +235,8 @@ public class DefaultLocalMuleClient implements LocalMuleClient
      */
     static public class MuleClientFlowConstruct implements FlowConstruct
     {
+        static MessageInfoMapping messageInfoMapping = new MuleMessageInfoMapping();
+
         MuleContext muleContext;
 
         public MuleClientFlowConstruct(MuleContext muleContext)
@@ -268,7 +271,7 @@ public class DefaultLocalMuleClient implements LocalMuleClient
 
         public MessageInfoMapping getMessageInfoMapping()
         {
-            return null;
+            return messageInfoMapping;
         }
 
         public MessageProcessorChain getMessageProcessorChain()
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/mule/client/DefaultLocalMuleClient.java
```diff
@@ -268,7 +269,7 @@ public MuleContext getMuleContext()
 
         public MessageInfoMapping getMessageInfoMapping()
         {
-            return null;
+            return MuleMessageInfoMapping.INSTANCE;
```
pablokraan: This singleton looks very ugly to me. Why can't create just a new instance?

transports/jms/src/main/java/org/mule/transport/jms/JmsMessageDispatcher.java
```diff
@@ -276,6 +280,35 @@ private MuleMessage dispatchMessage(MuleEvent event, boolean doSend) throws Exce
...
 
+    /**
+     * Resolve the value of correlationID that should be used for the JMS Message.  This is done here and not as part of
+     * transformation because of the need for visibility of MuleEvent and OutboundEndpoint.
+     *
+     * @param event the current MuleEvent
+     */
+    protected String resolveJmsCorrelationId(MuleEvent event) throws JMSException
+    {
+        return resolveMuleCorrelationId(event);
```
pablokraan: cannot just inline resolveMuleCorrelationId?

transports/jms/src/test/resources/integration/jms-correlation-id-propagation.xml
```diff
@@ -1,27 +1,47 @@
...
 
-    <flow name="passFromQueueInToQueueOut">
+    <flow name="withCustomCorrelationId">
+        <outbound-endpoint ref="in" />
+        <expression-message-info-mapping messageIdExpression="#[message.id]" correlationIdExpression="#['custom-cid']"/>
+    </flow>
+
+    <flow name="withNoCorrelationId">
+        <expression-component>_muleMessage.uniqueId='custom-cid'</expression-component>
+        <logger level="ERROR" message="#[message.id]"/>
```
pablokraan: Don't use loggers in the test configurations

transports/jms/src/main/java/org/mule/transport/jms/JmsMessageDispatcher.java
dfeist: This is because other implementations of JMS may use different values for MULE_CORRELATION_ID and JMSMCorrelationID.

core/src/main/java/org/mule/client/DefaultLocalMuleClient.java
dfeist: Added private constructor.

pablokraan: I was not talking about the missing private constructor bu why are you using a singleton instead of creating a new instance

transports/jms/src/main/java/org/mule/transport/jms/JmsMessageDispatcher.java
pablokraan: What is the problem with that? They will have to override this method anyway

PR #2523, opened 2016-04-01
MULE-9399: Mule throws a LifecycleException when disposing an Extension

elrodro83:
Source - partial fix, ensure correct fetching of disposable objects

### DIFF
--- a/core/src/main/java/org/mule/lifecycle/phases/MuleContextDisposePhase.java
+++ b/core/src/main/java/org/mule/lifecycle/phases/MuleContextDisposePhase.java
@@ -15,7 +15,6 @@ import org.mule.api.lifecycle.Disposable;
 import org.mule.api.lifecycle.Initialisable;
 import org.mule.api.lifecycle.LifecycleException;
 import org.mule.api.lifecycle.LifecyclePhase;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mule/lifecycle/phases/MuleContextDisposePhase.java
```diff
@@ -65,7 +64,7 @@ public MuleContextDisposePhase()
         orderedObjects.add(new NotificationLifecycleObject(Agent.class));
         orderedObjects.add(new NotificationLifecycleObject(ConfigurationProvider.class));
         orderedObjects.add(new NotificationLifecycleObject(Config.class));
-        orderedObjects.add(new NotificationLifecycleObject(Stoppable.class));
+        orderedObjects.add(new NotificationLifecycleObject(Disposable.class));
         orderedObjects.add(new NotificationLifecycleObject(Object.class));
```
marianogonzalez: shouldn't you remove this as well?

elrodro83: That's needed to call the @preDestroy annotated methods on other objects.

PR #2540, opened 2016-04-07
MULE-9431: Invert classloading lookup process for deployed artifacts

pablokraan:
_ Adding log4j and slf4j as system packages


modules/launcher/src/main/java/org/mule/module/launcher/MuleDeploymentService.java
```diff
@@ -55,7 +55,9 @@
     public static final Set<String> SYSTEM_PACKAGES = ImmutableSet.of(
             "java.", "org.mule", "com.mulesoft.mule.config", "com.mulesoft.mule.bti", "com.mulesoft.mule.cache",
             "com.mulesoft.mule.cluster", "com.mulesoft.mule.plugin", "com.mulesoft.mule.saml", "com.mulesoft.mule.throttling",
-            "com.mulesoft.mule.tracking", "com.mulesoft.mule.transaction", "com.mulesoft.module");
+            "com.mulesoft.mule.tracking", "com.mulesoft.mule.transaction", "com.mulesoft.module",
+            "org.apache.logging.log4j", "org.slf4j"
```
elrodro83: Should org.apache.commons.logging and org.apache.log4j be registered too?

PR #2544, opened 2016-04-07
MULE-9556 Support non-blocking processing in one-way flows

dfeist:


core/src/main/java/org/mule/DefaultMuleMessage.java
```diff
@@ -2111,7 +2111,6 @@ protected void copyToInbound(DefaultMuleMessage newMessage) throws Exception
         }
 
         newMessage.clearProperties(PropertyScope.INBOUND);
-        newMessage.clearProperties(PropertyScope.INVOCATION);
```
pablokraan: Why are you removing this line?

core/src/main/java/org/mule/construct/Flow.java
```diff
@@ -155,13 +137,40 @@ public MuleEvent process() throws Exception
     private MuleEvent createMuleEventForCurrentFlow(MuleEvent event, Object replyToDestination, ReplyToHandler
             replyToHandler)
     {
+        if (replyToHandler != null && !(replyToHandler instanceof DefaultReplyToHandler))
```
pablokraan: This instance of seems ugly to me. Why only the DefaultReplyToHandler is a special case?

core/src/main/java/org/mule/construct/Flow.java
```diff
@@ -155,13 +137,40 @@ public MuleEvent process() throws Exception
...
+        {
+            replyToHandler = createNonBlockingReplyToHandler(event, replyToHandler);
+        }
+        else
+        {
+            replyToHandler = null;
```
pablokraan: Why is original replyToHandler nullified here?

core/src/main/java/org/mule/DefaultMuleMessage.java
dfeist: I thought this was incorrect, but I see now it is used in VM and in this use it is correct to remove these.  I'll put this back and instead use a different method in TestMessageDispatcher.

pablokraan:
This PR is for a new feature... but we are on feature freeze mode....

core/src/main/java/org/mule/DefaultMuleMessage.java
dfeist: I thought this was wrong, but now see it's used by VM transport. I'll revert this and create response message differently in TestMessageDispatcher.

core/src/main/java/org/mule/construct/Flow.java
dfeist: This is also being done before this PR, just differently.

PR #2546, opened 2016-04-08
MULE-9558: HTTP failing response streaming calls exception strategy but

elrodro83:
returns generic error

### DIFF
--- a/core/src/main/java/org/mule/execution/AsyncResponseFlowProcessingPhase.java
+++ b/core/src/main/java/org/mule/execution/AsyncResponseFlowProcessingPhase.java
@@ -17,6 +17,7 @@ import org.mule.api.MessagingException;
 import org.mule.api.MuleEvent;
 import org.mule.api.MuleException;
 import org.mule.api.MuleMessage;
+import org.mule.api.MuleRuntimeException;
 import org.mule.api.exception.MessagingExceptionHandler;
 import org.mule.api.execution.ExecutionCallback;
 import org.mule.api.transport.ExceptionHandlingReplyToHandlerDecorator;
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/mule/execution/AsyncResponseFlowProcessingPhase.java
```diff
@@ -126,7 +127,7 @@ public void responseSentWithFailure(final Exception e, final MuleEvent event)
                     public void execute() throws Exception
                     {
                         ((DefaultMuleEvent) event).resetAccessControl();
-                        exceptionListener.handleException(e, event);
+                        RequestContext.setEvent(exceptionListener.handleException(e, event));
```
afelisatti: @dfeist should take a look at this to see if it makes sense

core/src/main/java/org/mule/execution/AsyncResponseFlowProcessingPhase.java
```diff
@@ -118,31 +120,33 @@ public void responseSentSuccessfully()
...
     {
         try
         {
-            callback.execute();
+            return callback.execute();
         }
         catch (Exception callbackException)
         {
             phaseResultNotifier.phaseFailure(callbackException);
+            throw new MuleRuntimeException(callbackException);
```
afelisatti: Why do you need to throw this after handling it?

modules/http/src/main/java/org/mule/module/http/internal/listener/HttpMessageProcessorTemplate.java
```diff
@@ -83,44 +83,58 @@ public void afterFailureProcessingFlow(Exception exception)
...
+                MuleEvent exceptionStrategyResult = responseCompletationCallback.responseSentWithFailure(e, muleEvent);
+                // Send the result from the event that was built from the Exception Strategy.
+                return this.responseBuilder.build(responseBuilder, exceptionStrategyResult);
             }
-            else
+            catch (Exception innerException)
             {
-                httpResponse = this.responseBuilder.build(responseBuilder, muleEvent);
+                // The failure occurred while executing the ES, or while building the response from the result of the ES
+                return buildErrorResponse();
```
afelisatti: Would it make sense to add some data from the exception in the error response?

PR #2547, opened 2016-04-08
MULE-9386: Stream not closed when using async scope

elrodro83:


core/src/main/java/org/mule/processor/AsyncDelegateMessageProcessor.java
```diff
@@ -50,6 +52,7 @@
 {
 
     protected Log logger = LogFactory.getLog(getClass());
+    private boolean consumablePayloadWarned = false;
```
marianogonzalez: this should be an AtomicBoolean

core/src/main/java/org/mule/processor/AsyncDelegateMessageProcessor.java
```diff
@@ -119,18 +122,28 @@ else if (!(flowConstruct instanceof StageNameSourceProvider))
...
     {
         if (event.isTransacted())
         {
             throw new MessagingException(CoreMessages.asyncDoesNotSupportTransactions(), event, this);
         }
 
+        final MuleMessage message = event.getMessage();
+        if (!consumablePayloadWarned && isConsumable(message.getPayload().getClass()))
+        {
+            consumablePayloadWarned = true;
```
marianogonzalez: I don't agree with this behaviour. I think we should either consume the payload or fail (like scatter-gather does). Allowing this can open the door for race conditions without much knowledge about concurrency will have a hard time understanding.

I'm also pretty sure that somewhere down the line, the `<async>` scope is either validating or consuming the payload in this case. @pablolagreca  do you remember where that is? @vromero didn't you spend a lot of time with a race condition problem on the `<async>` as well? any feedback on this?

elrodro83:
@marianogonzalez This will support the case in which the processors inside the async do not intend to consume the payload, which is a valid case, and something that some applications may already be doing.

Also, if we consume the message just before dispatching the async call, we will incur in an unwanted memory penalty.

Anyway, we should be clear on what the intended functionality and the use cases to support are, so anyfeedback form @pablolagreca and @vromero  is welcome.

PR #2551, opened 2016-04-09
MULE-9503 Add support for non-blocking dispatchers to transports. Add…

dfeist:
… missing support for response block within endpoint element. Rework to fix tests.


core/src/main/java/org/mule/endpoint/OutboundEndpointMessageProcessorChainBuilder.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package org.mule.endpoint;
+
+import org.mule.api.MuleException;
+import org.mule.api.construct.FlowConstruct;
+import org.mule.api.endpoint.ImmutableEndpoint;
+import org.mule.api.processor.MessageProcessor;
+import org.mule.processor.chain.DefaultMessageProcessorChainBuilder;
+
+public class OutboundEndpointMessageProcessorChainBuilder extends DefaultMessageProcessorChainBuilder
```
pablokraan: Add javadoc

PR #2553, opened 2016-04-11
MULE-9399: Mule throws a LifecycleException when disposing an Extensi…

marianogonzalez:
…on Source

### DIFF
--- a/core/src/main/java/org/mule/lifecycle/InitDisposeLifecycleInterceptor.java
+++ b/core/src/main/java/org/mule/lifecycle/InitDisposeLifecycleInterceptor.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.lifecycle;
+
+import org.mule.api.lifecycle.Disposable;
+import org.mule.api.lifecycle.Initialisable;
+import org.mule.api.lifecycle.LifecycleInterceptor;
+import org.mule.api.lifecycle.LifecyclePhase;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A {@link LifecycleInterceptor} which tracks the completion of the {@link Initialisable}
+ * phase so that if it fails to complete, the {@link Disposable} phase is only applied on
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/mule/lifecycle/InitDisposeLifecycleInterceptor.java
```diff
@@ -0,0 +1,123 @@
...
+        if (isInitialise(phase))
+        {
+            initialiseCompleted = true;
+            initialisedObjects = null;
+        }
+    }
+
+    private int getLifecycleTrackingKey(Object object)
+    {
+        return System.identityHashCode(object);
```
elrodro83: Can you use something less collisionable than hashCode?

marianogonzalez: this is the identity hash code... not collisionable

core/src/main/java/org/mule/lifecycle/RegistryLifecycleManager.java
```diff
@@ -32,9 +34,12 @@
...
 public class RegistryLifecycleManager extends AbstractLifecycleManager<Registry> implements RegistryLifecycleHelpers
 {
-    protected Map<String, LifecyclePhase> phases = new HashMap<String, LifecyclePhase>();
-    protected TreeMap<String, LifecycleCallback> callbacks = new TreeMap<String, LifecycleCallback>();
+
+
+    protected Map<String, LifecyclePhase> phases = new HashMap<>();
+    protected TreeMap<String, LifecycleCallback> callbacks = new TreeMap<>();
     protected MuleContext muleContext;
+    private final LifecycleInterceptor lifecycleInterceptor = new InitDisposeLifecycleInterceptor();
```
elrodro83: since this attribute is only used for init-dispose, its name should reflect that.

core/src/main/java/org/mule/lifecycle/RegistryLifecycleManager.java
```diff
@@ -32,9 +34,12 @@
 
 public class RegistryLifecycleManager extends AbstractLifecycleManager<Registry> implements RegistryLifecycleHelpers
 {
-    protected Map<String, LifecyclePhase> phases = new HashMap<String, LifecyclePhase>();
-    protected TreeMap<String, LifecycleCallback> callbacks = new TreeMap<String, LifecycleCallback>();
+
+
+    protected Map<String, LifecyclePhase> phases = new HashMap<>();
+    protected TreeMap<String, LifecycleCallback> callbacks = new TreeMap<>();
```
elrodro83: Change to SortedMap?

PR #2574, opened 2016-04-15
MULE-9586: Fix flakyness in HttpRequestProxyConfigTestCase

elrodro83:


modules/http/src/test/java/org/mule/module/http/functional/requester/HttpRequestProxyConfigTestCase.java
```diff
@@ -78,10 +78,14 @@ protected String getConfigFile()
     }
 
     @Before
-    public void startMockProxy() throws IOException
+    public void startMockProxy() throws IOException, InterruptedException
     {
         mockProxyAcceptor = new MockProxy();
         mockProxyAcceptor.start();
+
+        // Give time to the proxy thread to start up completely
```
afelisatti: It would be better to make the mock accept at least 2 connections and probe for the first one.

PR #2584, opened 2016-04-19
MULE-9592: MuleEndpointURI fails to extract right user and password. …

afelisatti:
…Refactoring to use raw data when available.

### DIFF
--- a/core/src/main/java/org/mule/endpoint/MuleEndpointURI.java
+++ b/core/src/main/java/org/mule/endpoint/MuleEndpointURI.java
@@ -22,6 +22,7 @@ import org.mule.util.StringUtils;
 
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.URLDecoder;
 import java.util.Properties;
 
 import org.apache.commons.logging.Log;
@@ -69,6 +70,9 @@ public class MuleEndpointURI implements EndpointURI
     private boolean dynamic;
     private transient MuleContext muleContext;
     private Properties serviceOverrides;
+
+    private String user;
+    private String password;
 
     MuleEndpointURI(String address,
                     String endpointName,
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mule/endpoint/MuleEndpointURI.java
```diff
@@ -494,37 +498,70 @@ public String getFilterAddress()
...
-            int i = userInfo.indexOf(':');
-            if (i > -1)
-            {
-                return userInfo.substring(i + 1);
-            }
+            userInfoData = userInfo;
+            decode = false;
+        }
+        String data = null;
+        if (StringUtils.isNotBlank(userInfoData))
```
elrodro83: Static import

core/src/main/java/org/mule/endpoint/MuleEndpointURI.java
```diff
@@ -494,37 +498,70 @@ public String getFilterAddress()
...
+        {
+            data = extractor.extract(userInfoData);
         }
-        return null;
+        return (data != null && decode) ? URLDecoder.decode(data) : data;
```
elrodro83: Static import

core/src/main/java/org/mule/endpoint/MuleEndpointURI.java
```diff
@@ -494,37 +498,70 @@ public String getFilterAddress()
 
     public String getUser()
```
elrodro83: Add javadoc to the implemented method in the interface to indicate that the returned value is url-decoded.
Also add @Override

core/src/main/java/org/mule/endpoint/MuleEndpointURI.java
```diff
@@ -494,37 +498,70 @@ public String getFilterAddress()
...
+                    }
+                }
+            });
         }
-        return null;
+        return user;
     }
 
-    public String getResponseTransformers()
+    public String getPassword()
```
elrodro83: Add javadoc to the implemented method in the interface to indicate that the returned value is url-decoded.
Also add @Override

PR #2585, opened 2016-04-19
MULE-9566: Content-Type set in request builder is not considered

canoasan:


modules/http/src/main/java/org/mule/module/http/internal/request/MuleEventToHttpRequest.java
```diff
@@ -74,7 +74,10 @@ public HttpRequestBuilder create(MuleEvent event, String resolvedMethod, String
 
         for (String outboundProperty : event.getMessage().getOutboundPropertyNames())
         {
-            builder.addHeader(outboundProperty, event.getMessage().getOutboundProperty(outboundProperty).toString());
+            if (!builder.getHeaders().containsKey(outboundProperty))
```
afelisatti: Wouldn't this make the 3.6.x branch different from 3.x? Outbound properties take precedence over the request builder AFAIK. Why is this needed?

canoasan: What about this specific scenario? The test is the same in both branches and in 3.6.x it fails if I let outbound properties replace the headers defined in the request builder.

PR #2587, opened 2016-04-19
MULE-9556 Support non-blocking processing in one-way flows. Fixed Fun…

gsfernandes:
…ctionalTestCases for nonBlocking


gsfernandes:
No need to change the processing strategy. Just set up the strategy in config for the flow.

PR #2591, opened 2016-04-20
MULE-8551 Add support for non-blocking components in catch exception …

dfeist:
…strategies

### DIFF
--- a/core/src/main/java/org/mule/exception/TemplateMessagingExceptionStrategy.java
+++ b/core/src/main/java/org/mule/exception/TemplateMessagingExceptionStrategy.java
@@ -22,6 +22,7 @@ import org.mule.api.transport.NonBlockingReplyToHandler;
 import org.mule.context.notification.ExceptionStrategyNotification;
 import org.mule.management.stats.FlowConstructStatistics;
 import org.mule.message.DefaultExceptionPayload;
+import org.mule.processor.AbstractRequestResponseMessageProcessor;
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/mule/exception/TemplateMessagingExceptionStrategy.java
```diff
@@ -36,66 +37,89 @@
...
+        return result;
+    }
 
-            muleContext.getNotificationManager().fireNotification(new ExceptionStrategyNotification(event, ExceptionStrategyNotification.PROCESS_START));
-            FlowConstruct flowConstruct = event.getFlowConstruct();
-            fireNotification(exception);
-            logException(exception, event);
-            processStatistics(event);
-            event.getMessage().setExceptionPayload(new DefaultExceptionPayload(exception));
+    private class ExceptionProcessingMessageProcessor extends AbstractRequestResponseMessageProcessor{
```
pablokraan: fix code style

core/src/test/java/org/mule/tck/SensingNullMessageProcessor.java
```diff
@@ -60,6 +61,8 @@ public void run()
                     {
                         eventToProcess = append(eventToProcess);
                     }
+                    // Reset access given we will be using the same event with a new thread
+                    ((ThreadSafeAccess) eventToProcess).resetAccessControl();
```
pablokraan: Don't you have to check that is an instance of ThreadSafeAccess before casting?

PR #2600, opened 2016-04-22
MULE-9608: Plugin extensions must be processed using the plugin class…

marianogonzalez:
…loader

_ Setting extension's classloader as the  current thread's context classloader

### DIFF
--- a/core-tests/src/test/java/org/mule/runtime/core/el/mvel/MVELExpressionExecutorTestCase.java
+++ b/core-tests/src/test/java/org/mule/runtime/core/el/mvel/MVELExpressionExecutorTestCase.java
@@ -13,7 +13,7 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.fail;
-import static org.mule.runtime.core.util.ClassUtils.withClassLoader;
+import static org.mule.runtime.core.util.ClassUtils.withContextClassLoader;
 import org.mule.mvel2.CompileException;
 import org.mule.mvel2.ParserConfiguration;
 import org.mule.runtime.core.api.lifecycle.InitialisationException;
@@ -84,7 +84,7 @@ public class MVELExpressionExecutorTestCase extends AbstractELTestCase
     @Test
     public void useContextClassLoader() throws ClassNotFoundException
     {
-        withClassLoader(new MyClassClassLoader(), () -> {
+        withContextClassLoader(new MyClassClassLoader(), () -> {
             try
             {
                 assertFalse((Boolean) mvel.execute("1 is org.MyClass", null));
--- a/core-tests/src/test/java/org/mule/runtime/core/el/mvel/MVELExpressionExecutorTestCase.java
+++ b/core-tests/src/test/java/org/mule/runtime/core/el/mvel/MVELExpressionExecutorTestCase.java
@@ -13,10 +13,11 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.fail;
-import org.mule.runtime.core.api.lifecycle.InitialisationException;
-import org.mule.runtime.core.el.context.AbstractELTestCase;
+import static org.mule.runtime.core.util.ClassUtils.withClassLoader;
 import org.mule.mvel2.CompileException;
 import org.mule.mvel2.ParserConfiguration;
+import org.mule.runtime.core.api.lifecycle.InitialisationException;
+import org.mule.runtime.core.el.context.AbstractELTestCase;
 import org.mule.tck.MuleTestUtils;
 import org.mule.tck.size.SmallTest;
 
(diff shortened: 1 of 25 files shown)


core-tests/src/test/java/org/mule/runtime/core/el/mvel/MVELExpressionExecutorTestCase.java
```diff
@@ -83,21 +84,18 @@ public void validExpression()
...
-        catch (Exception e)
-        {
-            e.printStackTrace();
-            fail(e.getMessage());
-        }
-        finally
-        {
-            Thread.currentThread().setContextClassLoader(originalClassLoader);
-        }
+        withClassLoader(new MyClassClassLoader(), () -> {
```
pablokraan: withClassLoader > withThreadContextClassLoader or similar

core-tests/src/test/java/org/mule/runtime/core/util/ClassUtilsTestCase.java
```diff
@@ -186,37 +190,37 @@ public void testLoadingResourceEnumeration() throws Exception
     @Test
     public void testGetSatisfiableMethods() throws Exception
     {
-        List methods = ClassUtils.getSatisfiableMethods(FruitBowl.class, new Class[]{Apple.class}, true,
+        List methods = ClassUtils.getSatisfiableMethods(FruitBowl.class, new Class[] {Apple.class}, true,
```
pablokraan: Don't change code style on code that is not related to the main commit, but if you want to do it... add static imports

marianogonzalez: static import added

core-tests/src/test/java/org/mule/runtime/core/el/mvel/MVELExpressionExecutorTestCase.java
marianogonzalez: renamed to withContextClassLoader

core/src/main/java/org/mule/runtime/core/util/ClassUtils.java
```diff
@@ -1208,4 +1171,55 @@ public static boolean isConsumable(Class<?> payloadClass)
...
+     * <p>
+     * This method guarantees that whatever the outcome, the thread's context classloader
+     * is set back to the value that it had before executing this method
+     *
+     * @param classLoader the context {@link ClassLoader} on which the {@code runnable} should be executed
+     * @param runnable    a closure
+     */
+    public static void withClassLoader(ClassLoader classLoader, Runnable runnable)
+    {
+        final Thread currentThread = Thread.currentThread();
```
pablokraan: Can't you just invoke the other method?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/enricher/ClassLoaderModelEnricher.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package org.mule.runtime.module.extension.internal.introspection.enricher;
+
+import org.mule.runtime.extension.api.introspection.ExtensionModel;
+import org.mule.runtime.extension.api.introspection.declaration.DescribingContext;
+import org.mule.runtime.extension.api.introspection.declaration.spi.ModelEnricher;
+import org.mule.runtime.extension.api.introspection.property.ClassLoaderModelProperty;
+
+/**
+ * Adds a {@link ClassLoaderModelProperty} at the {@link ExtensionModel} level, which
```
pablokraan: I don't know if there is currently a difference in the implementation, but the classloader must be the plugin's classloader, which could not necessarily be the one on the thread's context

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -298,6 +299,13 @@ public static Method getImplementingMethod(OperationDeclaration operationDeclara
         return operationDeclaration.getModelProperty(ImplementingMethodModelProperty.class).map(ImplementingMethodModelProperty::getMethod).orElse(null);
     }
 
+    public static ClassLoader getClassLoader(ExtensionModel extensionModel)
```
pablokraan: Add javadoc

core/src/main/java/org/mule/runtime/core/util/ClassUtils.java
marianogonzalez: Runnable and Supplier are not replaceable with each other. I would have to wrap the runnable into a supplier which returns null. IMO just as ugly as the current solution

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -298,6 +299,21 @@ public static Method getImplementingMethod(OperationDeclaration operationDeclara
...
+     * it returns the current {@link Thread}'s context classLoader.
+     *
+     * @param extensionModel a {@link ExtensionModel}
+     * @return a {@link ClassLoader}
+     */
+    public static ClassLoader getClassLoader(ExtensionModel extensionModel)
+    {
+        return extensionModel.getModelProperty(ClassLoaderModelProperty.class)
+                .map(ClassLoaderModelProperty::getClassLoader)
+                .orElse(Thread.currentThread().getContextClassLoader());
```
pablokraan: Not sure that is correct to default to the Thread's context classLoader here, because if you do this, there is no reason to define the ClassLoaderModelProperty. I would use ExtensionModel.class.classloader instead, so if something depends on extension's code, then it will fail and we will know that there is an error because the ClassLoaderModelProperty is not processed correctly.

PR #2613, opened 2016-04-28
MULE-9579 Using transactions when non blocking should log a warning n…

dfeist:
…ot an info


core/src/main/java/org/mule/util/OneTimeWarning.java
```diff
@@ -35,10 +35,17 @@ public void warn()
         delegate.warn();
     }
 
+    public void warn(Object... args)
```
marianogonzalez: add javadocs to this method and the other overload so that the difference between them is clear to users

PR #2615, opened 2016-04-28
MULE-9635: Include AHC/Grizzly performance fix in mule's custom ahc code

elrodro83:


modules/http/src/main/java/com/ning/http/client/providers/grizzly/HttpTransactionContext.java
```diff
@@ -131,20 +129,7 @@ static void cleanupTransaction(final HttpContext httpCtx,
...
-            if (httpTxContext.isRequestFullySent &&
-                httpTxContext.removeRequestSentCompletionHandler(completionHandler))
-            {
-                completionHandler.completed(httpTxContext);
-            }
-        }
+        httpTxContext.scheduleCleanup(httpCtx, completionHandler);
     }
 
     static void cleanupTransaction(final HttpContext httpCtx,
```
gsfernandes: This method was also removed in Grizzly fix. Please remove this too.

PR #2623, opened 2016-05-02
MULE-9640 JMS request-response with non-blocking support fails to use…

dfeist:
… connections properly on load when using DW.


transports/jms/src/main/java/org/mule/transport/jms/JmsMessageDispatcher.java
```diff
@@ -309,6 +311,7 @@ public void onFailure(Exception exception)
                 }
                 finally
                 {
+                    closeProducer(producer);
```
elrodro83: This code is repeated 3 times. Can it be extracted somewhere else?

PR #2634, opened 2016-05-04
MULE-9461: Implement watcher message source in new File connector

marianogonzalez:


extensions/file/src/main/java/org/mule/extension/file/internal/DirectoryListener.java
```diff
@@ -0,0 +1,487 @@
...
+                }
+            }
+        }
+    }
+
+    private void createWatcherService() throws IOException
+    {
+        try
+        {
+            watcher = FileSystems.getDefault().newWatchService();
```
elrodro83: Shouldn't this use the fileSystem attribute already in this object?

extensions/file/src/test/java/org/mule/extension/file/DirectoryListenerTestCase.java
```diff
@@ -0,0 +1,224 @@
...
+        }, "source did not stop"));
+    }
+
+    private void assertEvent(MuleMessage<?, ListenerFileAttributes> message, FileEventType type, Object expectedContent, String fileName) throws Exception
+    {
+        Object payload = message.getPayload();
+        if (payload instanceof InputStream)
+        {
+            payload = IOUtils.toString((InputStream) payload);
+            assertThat((String) payload, not((containsString(DR_MANHATTAN))));
```
elrodro83: extra parenthesis for containsString

extensions/file/src/main/java/org/mule/extension/file/internal/DirectoryListener.java
marianogonzalez: no. The fileSystem you refer to is the connector's abstration of it. This is FileSystemS... a Java API class

PR #2635, opened 2016-05-04
MULE-9660 OAuth2 authorization-code-grant-type is always expecting a …

gsfernandes:
…refresh_token when it is optional according to the spec


gsfernandes:
#2655

PR #2638, opened 2016-05-05
MULE-9565: Return UnionType when static metadata is fetch on a type that declares SubTypes

juandesi:


modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/metadata/MetadataExtensionFunctionalTestCase.java
```diff
@@ -171,4 +173,8 @@ protected void assertExpectedType(TypeMetadataDescriptor param, String name, Met
         assertThat(param.getType(), is(type));
     }
 
+    protected MetadataType getMetadataTypeOf(Class clazz)
+    {
+        return new JavaTypeLoader(this.getClass().getClassLoader()).load(clazz.getName()).get();
```
estebanwasinger: instead of doing the loading using the name of the class, just use the class `public MetadataType load(Type type)`

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/metadata/MetadataExtensionFunctionalTestCase.java
```diff
@@ -171,4 +173,8 @@ protected void assertExpectedType(TypeMetadataDescriptor param, String name, Met
         assertThat(param.getType(), is(type));
     }
 
+    protected MetadataType getMetadataTypeOf(Class clazz)
+    {
+        return new JavaTypeLoader(this.getClass().getClassLoader()).load(clazz);
```
marianogonzalez: use org.mule.runtime.module.extension.internal.util.ExtensionsTestUtils#TYPE_LOADER

PR #2644, opened 2016-05-05
MULE-9432: Isolate domains from core artifact

pablokraan:
_ Defining a filtering classloader to control what is exported from the container to the domains/plugins/apps

### DIFF
--- a/modules/artifact/src/main/java/org/mule/runtime/module/artifact/classloader/FilteringArtifactClassLoader.java
+++ b/modules/artifact/src/main/java/org/mule/runtime/module/artifact/classloader/FilteringArtifactClassLoader.java
@@ -7,11 +7,12 @@
 
 package org.mule.runtime.module.artifact.classloader;
 
+import static java.util.Collections.emptyList;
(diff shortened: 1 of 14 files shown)


modules/launcher/src/main/java/org/mule/runtime/module/launcher/NullClassLoaderFilter.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package org.mule.runtime.module.launcher;
+
+import org.mule.runtime.module.artifact.classloader.ClassLoaderFilter;
+
+/**
+ * Provides a {@link ClassLoaderFilter} that does not filter any class or resource.
+ */
+public class NullClassLoaderFilter implements ClassLoaderFilter
```
pablolagreca: ExportEverythingClassLoaderFilter?

gsfernandes: NoOpClassLoaderFilter?

PR #2649, opened 2016-05-06
MULE-9565: Return UnionType when static metadata is fetch on a type that declares SubTypes

juandesi:
Added logic for more specific cases like Subtype(Interface, ConcreteClass) and Subtype(ConcreteClass, ChildConcreteClass)

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataMediator.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataMediator.java
@@ -7,6 +7,7 @@
 package org.mule.runtime.module.extension.internal.metadata;
 
 import static java.util.stream.Collectors.toList;
+import static org.mule.metadata.java.utils.JavaTypeUtils.getType;
 import static org.mule.runtime.api.metadata.descriptor.builder.MetadataDescriptorBuilder.typeDescriptor;
 import static org.mule.runtime.api.metadata.resolving.FailureCode.NO_DYNAMIC_TYPE_AVAILABLE;
 import static org.mule.runtime.api.metadata.resolving.MetadataResult.failure;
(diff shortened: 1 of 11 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataMediator.java
```diff
@@ -205,14 +208,28 @@ public MetadataMediator(RuntimeExtensionModel extensionModel, RuntimeComponentMo
...
     private TypeMetadataDescriptor buildParameterTypeMetadataDescriptor(ParameterModel parameterModel)
     {
-        List<MetadataType> subTypes = subTypesMappingContainer.getSubTypes(parameterModel.getType());
+        MetadataType parameterType = parameterModel.getType();
+        TypeMetadataDescriptorBuilder typeDescriptorBuilder = typeDescriptor(parameterModel.getName());
+
+        List<MetadataType> subTypes = subTypesMappingContainer.getSubTypes(parameterType);
         if (!subTypes.isEmpty())
         {
+            boolean isIntantiable = isInstantiable(getType(parameterType));
```
marianogonzalez: isIntantiable -> isInstantiable

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataMediator.java
```diff
@@ -205,14 +208,28 @@ public MetadataMediator(RuntimeExtensionModel extensionModel, RuntimeComponentMo
      */
     private TypeMetadataDescriptor buildParameterTypeMetadataDescriptor(ParameterModel parameterModel)
     {
-        List<MetadataType> subTypes = subTypesMappingContainer.getSubTypes(parameterModel.getType());
+        MetadataType parameterType = parameterModel.getType();
+        TypeMetadataDescriptorBuilder typeDescriptorBuilder = typeDescriptor(parameterModel.getName());
+
+        List<MetadataType> subTypes = subTypesMappingContainer.getSubTypes(parameterType);
         if (!subTypes.isEmpty())
```
marianogonzalez: revert this... if it's empty then return early instead of returning at the end of the if block

PR #2653, opened 2016-05-09
MULE-9530: Improve Config retrieval when resolving Metadata for MetadataAware components

juandesi:


modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/metadata/MetadataImplicitStaticConfigurationTestCase.java
```diff
@@ -0,0 +1,38 @@
...
+        return METADATA_TEST_STATIC_IMPLICIT_CONFIGURATION;
+    }
+
+    @Test
+    public void resolveMetadataWithImplicitStaticConfig() throws Exception
+    {
+        componentId = new ProcessorId(RESOLVER_WITH_IMPLICIT_DYNAMIC_CONFIG, FIRST_PROCESSOR_INDEX);
+        MetadataKey key = newKey(AMERICA).withChild(newKey(USA).withChild(newKey(SAN_FRANCISCO))).build();
+        final MetadataResult<ComponentMetadataDescriptor> metadataKeysResult = metadataManager.getMetadata(componentId, key);
+        assertThat(metadataKeysResult.isSuccess(), is(false));
```
marianogonzalez: shouldn't be assertFailure?

juandesi: this one is for a Static but implicit referenced configuration. Shouldn't fail

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
```diff
@@ -149,16 +149,24 @@ public void setFlowConstruct(FlowConstruct flowConstruct)
...
+            {
+                throw new MetadataResolvingException("Configuration used for Metadata fetch cannot be dynamic", FailureCode.INVALID_CONFIGURATION);
+            }
+
+            configuration = configurationProvider.get(getInitialiserEvent(muleContext));
+        }
+        else
         {
-            throw new MetadataResolvingException("Configuration used for Metadata fetch cannot be dynamic", FailureCode.INVALID_CONFIGURATION);
+            configuration = extensionManager.getConfiguration(extensionModel, getInitialiserEvent(muleContext));
```
marianogonzalez: this doesn't look correct. This method will try to create an implicit config if none available. That config might be dynamic. It looks like you should do this if the provider is not present, and move the validation about it not being dynamic below it

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
```diff
@@ -170,21 +178,13 @@ private MetadataContext getMetadataContext() throws MetadataResolvingException
     {
         return getConfigurationProvider()
                 .map(provider -> provider.get(event))
-                .orElseGet(() -> {
-                    if (StringUtils.isBlank(configurationProviderName))
-                    {
-                        return extensionManager.getConfiguration(extensionModel, event);
-                    }
-                    return extensionManager.getConfiguration(configurationProviderName, event);
```
marianogonzalez: why is this case no longer valid?

tests/test-extensions/metadata-extension/src/main/java/org/mule/test/metadata/extension/MetadataExtension.java
```diff
@@ -34,5 +38,13 @@
         outputResolver = TestContentAndOutputResolverWithKeyResolver.class)
 public class MetadataExtension
 {
+    @Parameter
+    @Optional
+    @Expression(SUPPORTED)
```
marianogonzalez: redundant

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
alegmarra: `extensionManager.getConfiguration(configurationProviderName, event);` is redundant since that case should already be covered by `getConfigurationProvider()`

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/metadata/MetadataImplicitDynamicConfigurationTestCase.java
```diff
@@ -0,0 +1,43 @@
...
+{
+
+    @Override
+    protected String getConfigFile()
+    {
+        return METADATA_TEST_DYNAMIC_IMPLICIT_CONFIGURATION;
+    }
+
+    @Override
+    protected TestRule createTestTimeoutRule()
```
marianogonzalez: remove this

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/DefaultExtensionManager.java
```diff
@@ -220,25 +214,15 @@ else if (matches > 1)
...
     }
 
-    private boolean attemptToCreateImplicitConfiguration(ExtensionModel extensionModel, MuleEvent muleEvent)
+    private void createImplicitConfiguration(ExtensionModel extensionModel, MuleEvent muleEvent)
     {
         synchronized (extensionModel)
         {
             //check that another thread didn't beat us to create the instance
-            if (!extensionRegistry.getConfigurationProviders(extensionModel).isEmpty())
+            if (extensionRegistry.getConfigurationProviders(extensionModel).isEmpty())
```
marianogonzalez: this logic is not equivalent and doesn't look correct

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
```diff
@@ -149,16 +149,18 @@ public void setFlowConstruct(FlowConstruct flowConstruct)
 
     private MetadataContext getMetadataContext() throws MetadataResolvingException
     {
-        //TODO MULE-9530: Improve Config retrieval for Metadata resolution
-        if (!StringUtils.isBlank(configurationProviderName) &&
-            muleContext.getRegistry().get(configurationProviderName) instanceof DynamicConfigurationProvider)
+        MuleEvent fakeEvent = getInitialiserEvent(muleContext);
+        ConfigurationInstance<Object> configuration = getConfiguration(fakeEvent);
```
marianogonzalez: doesn't make much sense to fetch both provider and instance

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/DefaultExtensionManager.java
alegmarra: The previous case always returned true, since the created `configurationInstance` was never null, either one was created, or an exception was thrown. Now, instead of `attemp...` returning false and throwing an exception later, we just propagate the original exception (found [here](https://github.com/mulesoft/mule/pull/2653/commits/e0b8f1b993cb40fe1e8edad66c38fca15089bbf0#diff-edecab14dce16d65b4e6ac6d5eb80df8R41)). 

Also, instead of generating just the instance and forcing the StaticConfigProvider, we now delegate to the factory the Provider creation, which [can also create a Dynamic provider](https://github.com/mulesoft/mule/pull/2653/commits/e0b8f1b993cb40fe1e8edad66c38fca15089bbf0#diff-edecab14dce16d65b4e6ac6d5eb80df8R54) if the configuration is dynamic.
Having the dynamic provider created implicitly allowed us to identify dynamic configs if metadata was invoked using an implicit configuration that contained parameters with dynamic default values.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
alegmarra: The provider is only acquired to verify that it is not of Dynamic type. If we first tried to fetch the provider, and then obtain the configuration from it (like we did before), we could face the case where the provider is not present, forcing us to trigger the implicit configuration creation, and then again fetch the provider to verify if the one created for the implicit configuration is dynamic or not.

I think is cleaner just to get the configuration, let the manager do its magic if the config has to be created implicitly, and then verify if the registered provider is dynamic or not.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/DefaultExtensionManager.java
marianogonzalez: but if you already have configuration providers registered, by do you call registerConfigurationPRovder() again? you should use the one that exists.

What I'm questioning is that the if seems to be missing a `!`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
marianogonzalez: how can the provider not being present if you already obtain a configuration?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/DefaultExtensionManager.java
alegmarra: if `getConfigurationProviders` is empty, then you need to register a new connection provider. Otherwise, at least one has already been registered, so do nothing.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
alegmarra: yeah, probably the `.orElseThrow(()` will never be called and could be replaced by a plain `get`

marianogonzalez: obtain the  provider and if it suits you get a config out of it. Otherwise, you might have initialised a dynamc config which you later discard just because the provider wasn't of the correct ttype

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/DefaultExtensionManager.java
marianogonzalez: ok

PR #2668, opened 2016-05-16
MULE-9696: Lifecycle wrongfully applied on connections when disconnected

marianogonzalez:


core/src/main/java/org/mule/runtime/core/registry/MuleRegistryHelper.java
```diff
@@ -440,15 +441,36 @@ public Object applyLifecycle(Object object) throws MuleException
     @Override
     public Object applyLifecycle(Object object, String phase) throws MuleException
     {
+        return withLifecycleRegistry(object, registry -> registry.applyLifecycle(object, phase));
+    }
+
+    @Override
+    public void applyLifecycle(Object object, String startPhase, String toPhase) throws MuleException
```
elrodro83: Shouldn't this return Object for consistency with the method right before?

marianogonzalez: i would rather not. I think that what we should actually review is why the other one is returning a new object... It makes sense for injection but not for lifecycle

PR #2671, opened 2016-05-17
Update DefaultRouterResultsHandler.java

JeyaramD:
Copies the DataType of the Flow Variables. Which allows DataWeave to detects the flowVars type for processing


marianogonzalez:
Hello, thank you for your contribution.

Per the contributions guide, please:

* Review the [MuleSoft Contributor's Agreement.](http://www.mulesoft.org/legal/contributor-agreement.html) Before any contribution is accepted, we need you to run the following [script](https://api-notebook.anypoint.mulesoft.com/notebooks#bc1cf75a0284268407e4). 
* File an issue on Mule's Jira and have the commit message reference it. For example: MULE-XXXX: DefaultReourteResultHandlers doesn't copy flow variables data types
* Add a test case

Thank you

JeyaramD:
MULE-9712: Fix for FlowVars' DataTypes not being copied after scatter-gather

JeyaramD:
@marianogonzalez have updated the pull request with tests and JIRA reference.  Have also run the script to accept MuleSoft Contributor's Agreement.

core/src/main/java/org/mule/routing/CollectAllAndFlowVarsAggregationStrategy.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * If no routes generated exeption then it returns a new {@link MuleEvent} under the
+ * rules of {@link DefaultRouterResultsHandler} (you can change this behaviour by
+ * overriding {@link #aggregateWithoutFailedRoutes(AggregationContext)}. Otherwise, a
+ * {@link CompositeRoutingException} is thrown (override
+ * {@link #aggregateWithFailedRoutes(AggregationContext) to customize}
+ * 
+ * @since 3.5.0
+ */
+public class CollectAllAndFlowVarsAggregationStrategy extends CollectAllAggregationStrategy
```
marianogonzalez: this is not correct. The original fix that you submitted was on the `DefaultRouterResultsHandler` class. You're not introducing this class into the productive code base and only using it in the test case. This class will never be used in production since nobody else references it. 

Therefore, your original fix remains untested and none of the test cases you added below add any value

core/src/test/java/org/mule/processor/chain/DefaultMessageProcessorChainTestCase.java
```diff
@@ -904,6 +905,29 @@ public void testAll() throws MuleException, Exception
 
         assertEquals(1, threads);
     }
+    
+    @Test
+    public void testScatterGatherRouterWithCollectAllAndFlowVarsAggregationStrategy() throws MuleException, Exception
+    {
+        ScatterGatherRouter scatterGatherRouter = new ScatterGatherRouter();
```
marianogonzalez: this tests initialises a scatter-gather and then doesn't stop it. Because it's manually creating the scatter-gather, it's also highly coupled to its internal structure and the assertion becomes hard to read.

Replace by a functional test

core/src/test/java/org/mule/routing/DefaultRouterResultsHandlerTestCase.java
```diff
@@ -130,7 +133,8 @@ public void aggregateMultipleEvents() throws Exception
         // Because a new MuleMessageCollection is created, propagate properties from
         // original event
         // TODO DF
-        // assertEquals("value1", result.getMessage().getInvocationProperty("key1"));
+        assertEquals("value1", result.getMessage().getInvocationProperty("key1"));
+        assertEquals(simpleDateType1.getMimeType(), result.getFlowVariableDataType("key1").getMimeType());
```
marianogonzalez: you should not assert just over the mimeType. You should rely on the DataType's equals(Object) method

JeyaramD:
removed CollectAllAndFlowVarsAggregationStrategy and it's test cases
Updated test cases for DefaultRouterResultsHandler to use equals()

PR #2680, opened 2016-05-19
MULE-9725: Refactor FTP connector code to allow FTPS one to reuse ita…

marianogonzalez:
…uth issues

### DIFF
--- a/modules/tls/src/main/java/org/mule/runtime/module/tls/internal/MuleTlsContextFactoryBuilder.java
+++ b/modules/tls/src/main/java/org/mule/runtime/module/tls/internal/MuleTlsContextFactoryBuilder.java
@@ -7,14 +7,16 @@
 package org.mule.runtime.module.tls.internal;
 
 import static org.mule.runtime.core.config.i18n.MessageFactory.createStaticMessage;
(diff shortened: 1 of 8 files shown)


modules/tls/src/main/java/org/mule/runtime/module/tls/internal/MuleTlsContextFactoryBuilder.java
```diff
@@ -32,6 +35,11 @@
     @Override
     public void initialise() throws InitialisationException
     {
+        if (!initialised.compareAndSet(false, true))
```
afelisatti: Why was this necessary?

marianogonzalez: same reason why TlsContextFactory does the same thing. The initialise needs to be idempotent. Otherwise the bug referenced in the connection provider (but not exclusively that one) could mess this up

marianogonzalez: the connection provider i'm talking about is in the EE PR

PR #2688, opened 2016-05-23
MULE-9662:Child element not generated for SubType in Source

alegmarra:


modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/SubTypesMappingParserTestCase.java
```diff
@@ -170,4 +172,17 @@ public void subtypedAndConcreteParametersAsAttributes() throws Exception
...
     }
+
+    @Test
+    public void mappedParamSource() throws Exception
+    {
+        try
+        {
+            ((Flow) getFlowConstruct("sourceWithParameterMapping")).start();
+        }
+        catch (Exception e)
```
marianogonzalez: unnecesary catch

PR #2698, opened 2016-05-26
MULE-8196 Listener 503 responses do not include an informative http b…

mtqp:
…ody.

### DIFF
--- a/modules/http/src/main/java/org/mule/module/http/internal/listener/ServiceTemporarilyUnavailableListenerRequestHandler.java
+++ b/modules/http/src/main/java/org/mule/module/http/internal/listener/ServiceTemporarilyUnavailableListenerRequestHandler.java
@@ -6,60 +6,23 @@
  */
 package org.mule.module.http.internal.listener;
 
-import org.mule.module.http.internal.domain.request.HttpRequestContext;
-import org.mule.module.http.internal.domain.response.HttpResponseBuilder;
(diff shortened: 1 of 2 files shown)


modules/http/src/main/java/org/mule/module/http/internal/listener/ServiceTemporarilyUnavailableListenerRequestHandler.java
```diff
@@ -6,60 +6,23 @@
...
-    public static final int SERVICE_TEMPORARILY_UNAVAILABLE_STATUS_CODE = 503;
-
-    private Logger logger = LoggerFactory.getLogger(ServiceTemporarilyUnavailableListenerRequestHandler.class);
-
     private static ServiceTemporarilyUnavailableListenerRequestHandler instance = new ServiceTemporarilyUnavailableListenerRequestHandler();
 
     private ServiceTemporarilyUnavailableListenerRequestHandler()
     {
-
+        super(SERVICE_UNAVAILABLE.getStatusCode(), SERVICE_UNAVAILABLE.getReasonPhrase(), "Service not available for endpoint: %s");
```
elrodro83: Replace 'endpoint' with 'request uri'

mtqp: Thanks for the review! Your suggestion has already been fixed.

PR #2702, opened 2016-05-26
MULE-9711: Change namespaces of the transports

elrodro83:


tests/integration-tranports/src/test/java/org/mule/shutdown/ShutdownAppInDomainTestCase.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Set;
+
+import org.junit.Before;
+import org.junit.Test;
+
+/**
+ * Tests that threads in pools defined in a domain do not hold references to objects of the application in their thread
+ * locals.
+ */
+public class ShutdownAppInDomainTestCase extends DomainFunctionalTestCase
```
pablolagreca: Is this change supposed to be here?

elrodro83: Yes, it contains some tests that were commented out since they used JMS, I moved it to the transports-integration module.

PR #2705, opened 2016-05-27
MULE-9654 Add a way to deploy applications that use plugins not packa…

gsfernandes:
…ged in the app - Fixed all-modules dependencies

_There duplicated dependencies in all-modules
_Removed specific exclude for core app plugins

### DIFF
--- a/modules/all-modules/pom.xml
+++ b/modules/all-modules/pom.xml
@@ -47,16 +47,6 @@
         </dependency>
         <dependency>
             <groupId>org.mule.modules</groupId>
-            <artifactId>mule-module-file-extension-common</artifactId>
-            <version>${project.version}</version>
-        </dependency>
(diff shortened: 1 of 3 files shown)


modules/all-modules/pom.xml
```diff
@@ -47,16 +47,6 @@
         </dependency>
         <dependency>
             <groupId>org.mule.modules</groupId>
-            <artifactId>mule-module-file-extension-common</artifactId>
```
pablokraan: I think this jar must be maintained because file, ftp and ftps depend on it.

gsfernandes: You are right, it is maintained. The thing is this dependency was duplicated that's why I'm removing it here. Line 130 has the same dependency declared.

PR #2706, opened 2016-05-27
MULE-9759:Change test extensions packaging from jar to mule-extension

pablocabrera85:


pablocabrera85:
Right now we cannot build this due to cyclic dependencies

PR #2712, opened 2016-05-28
MULE-9727 Extension plugin not working if annotated class defines an ...

gsfernandes:
...operation that uses an enum class not included in the same package


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/resources/manifest/ExportedArtifactsCollector.java
```diff
@@ -198,6 +201,23 @@ public void visitObject(ObjectType objectType)
                 {
                     accept.set(true);
                 }
+
+                @Override
+                public void visitString(StringType stringType)
+                {
+                    accept.set(stringType.getAnnotation(TypeIdAnnotation.class).stream().anyMatch(
```
marianogonzalez: MetadataUtils class has a method which does this and returns the class associated with the type. That would make for a clerarer implementation of this method

PR #2719, opened 2016-05-30
MULE-9649: Check that different application plugins do not export the…

pablokraan:
… same java packages

_ Returning exportedClassPackages from ArtifactClassLoaderFilter
_ Renaming ClassLoaderFilterFactory related classes
_ Moving ArtifactDescriptor#classLoaderFilter down to ApplicationPluginDescriptor
_ Adding DuplicateExportedPackageException to indicate the validation error
_ Updating ApplicationPluginRepository#getContainerApplicationPluginDescriptors to throw ArtifactDescriptorCreateException instead of IOException
_ Updating DeploymentServiceTestCase broken after adding the new validation


modules/launcher/src/main/java/org/mule/runtime/module/launcher/ApplicationDescriptorFactory.java
```diff
@@ -97,6 +104,30 @@ public ApplicationDescriptor create(File artifactFolder) throws ArtifactDescript
...
+    {
+        final HashMap<String, String> exportedPackages = new HashMap<>();
+
+        for (ApplicationPluginDescriptor plugin : plugins)
+        {
+            for (String exportedPackage : plugin.getClassLoaderFilter().getExportedClassPackages())
+            {
+                if (exportedPackages.containsKey(exportedPackage))
+                {
+                    throw new DuplicateExportedPackageException(String.format("Package '%s' is exported on plugins '%s' and '%s'", exportedPackage, exportedPackages.get(exportedPackage), plugin.getName()));
```
elrodro83: Maybe a different check/wording can be done to account for the case that a same package is exported by more than 2 plugins, otherwise another deployment would be needed to detect that case.

modules/artifact/src/main/java/org/mule/runtime/module/artifact/classloader/ArtifactClassLoaderFilterFactory.java
```diff
@@ -7,56 +7,18 @@
...
 package org.mule.runtime.module.artifact.classloader;
 
-import org.mule.runtime.core.util.StringUtils;
-
-import java.util.HashSet;
-import java.util.Set;
-
 /**
- * Creates {@link ArtifactClassLoaderFilter} instances
+ * Creates {@link ClassLoaderFilter} instances
```
elrodro83: Should be ArtifactClassLoaderFilter

modules/launcher/src/main/java/org/mule/runtime/module/launcher/ApplicationDescriptorFactory.java
```diff
@@ -97,6 +104,30 @@ public ApplicationDescriptor create(File artifactFolder) throws ArtifactDescript
...
+    private List<ApplicationPluginDescriptor> getAllApplicationPlugins(Set<ApplicationPluginDescriptor> plugins)
+    {
+        final List<ApplicationPluginDescriptor> result = new LinkedList<>(applicationPluginRepository.getContainerApplicationPluginDescriptors());
+        result.addAll(plugins);
+        return result;
+    }
+
+    private void verifyPluginExportedPackages(List<ApplicationPluginDescriptor> plugins)
+    {
+        final HashMap<String, String> exportedPackages = new HashMap<>();
```
elrodro83: Change declaration to Map

modules/launcher/src/main/java/org/mule/runtime/module/launcher/ApplicationDescriptorFactory.java
```diff
@@ -108,24 +109,77 @@ public ApplicationDescriptor create(File artifactFolder) throws ArtifactDescript
...
+            }
+        }
+
+        if (error)
+        {
+            throw new DuplicateExportedPackageException(buildPackageDuplicationErrorMessage(exportedPackages));
+        }
+    }
+
+    private String buildPackageDuplicationErrorMessage(Map<String, List<String>> exportedPackages)
```
elrodro83: Move this to DuplicateExportedPackageException. Ideally the exception could have the exportedPackages map and build the message on demand, so that while debugging/inspecting it can be accessed and not just the string message.

PR #2725, opened 2016-05-31
MULE-9730: Support grizzly's PRESERVE_HEADER_CASE property

afelisatti:


modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/GrizzlyHttpRequestAdapter.java
```diff
@@ -49,7 +51,7 @@ public GrizzlyHttpRequestAdapter(FilterChainContext filterChainContext, HttpCont
         this.requestPacket = (HttpRequestPacket) httpContent.getHttpHeader();
         isTransferEncodingChunked = httpContent.getHttpHeader().isChunked();
         int contentLengthAsInt = 0;
-        String contentLengthAsString = requestPacket.getHeader(HttpHeaders.Names.CONTENT_LENGTH);
+        String contentLengthAsString = requestPacket.getHeader(CONTENT_LENGTH);
```
gsfernandes: Shouldn't code the new method? getHeaderValueIgnoreCase instead of getHeader?

gsfernandes: Ignore this, it is requestPacket called here.

PR #2726, opened 2016-05-31
MULE-8252 When the status code is set, http listener should auto comp…

mtqp:
MULE 8252 When the status code is set, http listener should auto complete de reason phrase

### DIFF
--- a/modules/http/src/main/java/org/mule/module/http/api/HttpConstants.java
+++ b/modules/http/src/main/java/org/mule/module/http/api/HttpConstants.java
@@ -6,10 +6,6 @@
  */
 package org.mule.module.http.api;
 
-import com.google.common.collect.Collections2;
-
-import java.util.Arrays;
-
 import static org.mule.module.http.api.HttpConstants.RequestProperties.HTTP_PREFIX;
 
 /**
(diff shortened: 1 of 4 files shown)


modules/http/src/main/java/org/mule/module/http/api/HttpConstants.java
```diff
@@ -6,6 +6,10 @@
  */
 package org.mule.module.http.api;
 
+import com.google.common.collect.Collections2;
```
afelisatti: Remove import, it's not used.

modules/http/src/main/java/org/mule/module/http/api/HttpConstants.java
```diff
@@ -6,6 +6,10 @@
...
+
+import java.util.Arrays;
```
afelisatti: Same as above.

modules/http/src/main/java/org/mule/module/http/api/HttpConstants.java
mtqp: Done

modules/http/src/main/java/org/mule/module/http/api/HttpConstants.java
mtqp: Done

PR #2727, opened 2016-05-31
MULE-9615: Extension configs can not define the same Source/Operation…

marianogonzalez:
… Class


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/schema/builder/ConnectionProviderSchemaDelegate.java
```diff
@@ -45,8 +44,13 @@ public ConnectionProviderSchemaDelegate(SchemaBuilder builder)
         this.builder = builder;
     }
 
-    public void registerConnectionProviderElement(Schema schema, ConnectionProviderModel providerModel)
+    public void registerConnectionProviderElement(ConnectionProviderModel providerModel)
```
elrodro83: Should these changes also be in ConfigurationSchemaDelegate?

PR #2728, opened 2016-05-31
MULE-9687 - Fix domain content validation failing test case

pablolagreca:


modules/spring-config/src/main/java/org/mule/runtime/config/spring/MuleArtifactContext.java
```diff
@@ -27,6 +27,7 @@
 
 import java.io.IOException;
 import java.util.ServiceLoader;
+import java.util.function.Function;
```
marianogonzalez: what's this import for?

PR #2741, opened 2016-06-01
MULE-9798 If a domain fails to be deployed a NPE is thrown when deplo… 

gsfernandes:
…ying and app that is registered for the same domain


modules/launcher/src/test/java/org/mule/runtime/module/launcher/application/DefaultApplicationFactoryTestCase.java
```diff
@@ -116,4 +120,17 @@ private Domain createDomain(String name)
 
         return domain;
     }
+
+    @Test
+    public void applicationDesployFailDueToDomainNotDeployed() throws Exception
+    {
+        expectedException.expect(DeploymentException.class);
```
pablolagreca: This line must be just before the method that causes it.

PR #2743, opened 2016-06-02
MULE-9793 ClassNotFoundException: org.apache.commons.io.input.ReaderI…

gsfernandes:
…nputStream when using file extension

### DIFF
--- a/extensions/all-extensions/pom.xml
+++ b/extensions/all-extensions/pom.xml
@@ -23,6 +23,21 @@
             <artifactId>mule-module-file</artifactId>
             <version>${project.version}</version>
             <type>zip</type>
+            <exclusions>
+                <!--TODO(gfernandes): remove this, see validation's comment-->
+                <exclusion>
+                    <groupId>commons-io</groupId>
+                    <artifactId>commons-io</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>commons-lang</groupId>
+                    <artifactId>commons-lang</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>com.google.guava</groupId>
+                    <artifactId>guava</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <dependency>
@@ -45,6 +60,22 @@
                 <exclusion>
                     <groupId>com.jcraft</groupId>
                     <artifactId>jsch</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>commons-io</groupId>
+                    <artifactId>commons-io</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>commons-lang</groupId>
+                    <artifactId>commons-lang</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>commons-collections</groupId>
+                    <artifactId>commons-collections</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>com.google.guava</groupId>
+                    <artifactId>guava</artifactId>
                 </exclusion>
             </exclusions>
         </dependency>
@@ -70,6 +101,10 @@
                 <exclusion>
                     <groupId>commons-validator</groupId>
                     <artifactId>commons-validator</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>commons-lang</groupId>
+                    <artifactId>commons-lang</artifactId>
                 </exclusion>
                 <exclusion>
                     <groupId>com.google.guava</groupId>
--- a/extensions/all-extensions/pom.xml
+++ b/extensions/all-extensions/pom.xml
@@ -91,11 +91,9 @@
             <artifactId>mule-module-validation</artifactId>
             <version>${project.version}</version>
             <type>zip</type>
-            <!--TODO(gfernandes): https://issues.apache.org/jira/browse/MASSEMBLY-799
-                wildcard exclusions were supported in maven-core https://issues.apache.org/jira/browse/MNG-3832
-                but still assembly is not handling this correctly. Remove commons-validator and add * for groupId and artifacId
-                so transitive dependencies will not end up in lib/opt (and we don't want to explicitly
-                exclude them from the assembly descriptor)
+            <!--TODO(gfernandes): https://www.mulesoft.org/jira/browse/MULE-9810
+                Remove explicit exclusions whenever this issue is fixed (this applies to all the zip type
+                dependencies
             -->
             <exclusions>
                 <exclusion>
--- a/extensions/ftp/pom.xml
+++ b/extensions/ftp/pom.xml
@@ -29,11 +29,33 @@
             <groupId>commons-net</groupId>
             <artifactId>commons-net</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>commons-lang</groupId>
+            <artifactId>commons-lang</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>commons-collections</groupId>
+            <artifactId>commons-collections</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
+
         <dependency>
             <groupId>com.jcraft</groupId>
             <artifactId>jsch</artifactId>
             <version>${jschVersion}</version>
         </dependency>
+
         <dependency>
             <groupId>org.mule.modules</groupId>
             <artifactId>mule-module-file-extension-common</artifactId>
--- a/extensions/pom.xml
+++ b/extensions/pom.xml
@@ -107,6 +107,40 @@
                 <version>${mule.extensions.maven.plugin.version}</version>
                 <extensions>true</extensions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>analyze-dependencies</id>
+                        <goals>
+                            <goal>analyze-only</goal>
+                        </goals>
+                        <phase>validate</phase>
+                        <configuration>
+                            <failOnWarning>true</failOnWarning>
+                            <verbose>true</verbose>
+                            <ignoredUsedUndeclaredDependencies>
+                                <!-- Mule dependencies are ignored -->
+                                <ignoredUsedUndeclaredDependency>org.mule*</ignoredUsedUndeclaredDependency>
+                                <!-- Testing dependencies are ignored -->
+                                <ignoredUsedUndeclaredDependency>org.hamcrest</ignoredUsedUndeclaredDependency>
+                                <ignoredUsedUndeclaredDependency>org.mockito</ignoredUsedUndeclaredDependency>
+                                <ignoredUsedUndeclaredDependency>junit:junit</ignoredUsedUndeclaredDependency>
+                                <ignoredUsedUndeclaredDependency>commons-logging:commons-logging</ignoredUsedUndeclaredDependency>
+
+                                <!-- Logging related artifacts are exposed by the BOOTSTRAP ClassLoader -->
+                                <ignoredUsedUndeclaredDependency>org.slf4j:slf4j-api</ignoredUsedUndeclaredDependency>
+                                <ignoredUsedUndeclaredDependency>javax.inject:javax.inject</ignoredUsedUndeclaredDependency>
+                            </ignoredUsedUndeclaredDependencies>
+                            <ignoredUnusedDeclaredDependencies>
+                                <!-- Don't need to worry about this -->
+                                <ignoredUnusedDeclaredDependency>:::</ignoredUnusedDeclaredDependency>
+                            </ignoredUnusedDeclaredDependencies>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
         </plugins>
     </build>
 
--- a/extensions/file/pom.xml
+++ b/extensions/file/pom.xml
@@ -24,6 +24,22 @@
             <version>${project.version}</version>
             <scope>provided</scope>
         </dependency>
+
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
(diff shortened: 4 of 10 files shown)


extensions/ftp/pom.xml
```diff
@@ -29,11 +29,38 @@
             <groupId>commons-net</groupId>
             <artifactId>commons-net</artifactId>
         </dependency>
+
+        <dependency>
```
marianogonzalez: this one should be on the runtime classloader. Not repeated on every plugin

extensions/all-extensions/pom.xml
```diff
@@ -46,6 +65,26 @@
...
+                <exclusion>
+                    <groupId>commons-lang</groupId>
+                    <artifactId>commons-lang</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>commons-collections</groupId>
+                    <artifactId>commons-collections</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>com.google.guava</groupId>
```
marianogonzalez: this connector doesn't use guava

extensions/all-extensions/pom.xml
```diff
@@ -46,6 +65,26 @@
...
+                <exclusion>
+                    <groupId>javax.inject</groupId>
+                    <artifactId>javax.inject</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>commons-io</groupId>
+                    <artifactId>commons-io</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>commons-lang</groupId>
```
marianogonzalez: where does this connector use this dependency?

extensions/ftp/pom.xml
```diff
@@ -29,11 +29,38 @@
...
+            <groupId>javax.inject</groupId>
+            <artifactId>javax.inject</artifactId>
+        </dependency>
+
+        <dependency>
```
marianogonzalez: where does FTP use these dependencies?

extensions/all-extensions/pom.xml
```diff
@@ -23,6 +23,25 @@
             <artifactId>mule-module-file</artifactId>
             <version>${project.version}</version>
             <type>zip</type>
+            <exclusions>
```
marianogonzalez: I don' think this is the right solution. As a developer, how would I know at design time that those dependencies won't be visible in runtime and I have to repackage them? These exclusions should be on the core poms

gsfernandes: The exclusion is just a matter of having fixed the maven-assembly-plugin to exclude every transitive dependency on these extensions.
It is still not simple to answer that question, but the goal is to only expose API from Mule and a few common third party libraries (actually their packages) in order to have a lightweight distribution.

As a golden rule we have to assume that these dependencies would not be at runtime because the runtime engine will not even have them (in a later stage whenever we have cleaned things in distribution).

What I have being doing in order to validate if the extension pom has the correct dependencies is to use maven-dependency-plugin:analyze and with the output check when I have a dependency not declared in my pom.

For instance: file extension without the change proposed here in this PR
{code}
[INFO] --- maven-dependency-plugin:2.10:analyze (default-cli) @ mule-module-file ---
[WARNING] Used undeclared dependencies found:
[WARNING]    javax.inject:javax.inject:jar:1:provided
[WARNING]    commons-io:commons-io:jar:2.4:provided
[WARNING]    org.hamcrest:hamcrest-core:jar:1.3:test
[WARNING]    org.mule.tests:mule-tests-unit:jar:4.0-SNAPSHOT:test
[WARNING]    org.mule:mule-api:jar:1.0.0-SNAPSHOT:provided
[WARNING]    org.hamcrest:hamcrest-library:jar:1.3:test
[WARNING]    commons-lang:commons-lang:jar:2.4:provided
[WARNING]    com.google.guava:guava:jar:18.0:provided
[WARNING]    junit:junit:jar:4.11:test
[WARNING]    org.mockito:mockito-core:jar:1.9.5:test
[WARNING]    org.mule.modules:mule-module-extensions-support:jar:4.0-SNAPSHOT:provided
[WARNING]    org.slf4j:slf4j-api:jar:1.7.7:provided
[WARNING] Unused declared dependencies found:
[WARNING]    com.googlecode.multithreadedtc:multithreadedtc:jar:1.01:test
[WARNING]    org.mule:mule-core-tests:test-jar:tests:4.0-SNAPSHOT:test
[WARNING]    org.mule.modules:mule-module-extensions-spring-support:jar:4.0-SNAPSHOT:provided
{code}

You have to know that these libraries are exposed from the container:
{code}
org.slf4j:slf4j-api:jar:1.7.7:provided
org.mule:mule-api:jar:1.0.0-SNAPSHOT:provided
org.mule.modules:mule-module-extensions-support:jar:4.0-SNAPSHOT:provided
{code}
But the rest of them (not the one for testing or even the extension support) has to be declared in order to get it packaged in this extension's lib folder.

I would improve the mule-extensions parent pom to fail the build if this analysis has any dependency in a plugin that is used and not declared or the other way around, excluding only the core extension support and testing stuff.
I think that it would help to avoid having issues like the one reported here?

extensions/all-extensions/pom.xml
gsfernandes:
```
public final class LocalFileConnectionProvider implements ConnectionProvider<FileConnector, FileSystem>
{
    @Inject
```

FileInputStream uses org.apache.commons.io.input.ReaderInputStream

LocalCreateDirectoryCommand uses org.apache.commons.lang.StringUtils

extensions/all-extensions/pom.xml
gsfernandes: SftpClient uses com.google.common.base.Supplier from guava.

extensions/ftp/pom.xml
gsfernandes: In order to do that we have to expose the packages from this library but that's something that we would not do (at least for the time being). Or we could have commons-net as a plugin instead and that should go as a shared plugins /lib/opt (as core, and other mule plugins).
Still it is a conversion that needs to happen in order to understand that isolation goes against reuse, at least as jars.

extensions/all-extensions/pom.xml
marianogonzalez: All of that is fine, but what I'm saying is that the problem goes way beyond extensions. It will be very hard for a developer to figure out that the commons-io classes that the IDE and test cases allows to use, will not be visible in runtime.

What I'm saying is that we should:

* Make sure that just because my maven module depend on mule-core, it doesn't get exposed to mule-core dependencies which are not going to be available in runtime
* Improve our unit test infrastructure so that its classloaders are closer to the runtime ones
* Develop a Mojo which automatically adds the exclusions which you're explicitly adding here... Otherwise this doesn't scale

extensions/ftp/pom.xml
marianogonzalez: check with Kraan, but I'm under the impression that the `java` and `javax` ones are going to be runtime only

extensions/all-extensions/pom.xml
marianogonzalez: that's a bug we need to fix. For sure I mean `java.util.function.Supplier` but screwed up with the import

extensions/all-extensions/pom.xml
gsfernandes: Yes, you are right. We have to do all of that. Don't merge this yet while
I'm trying to add dependency:analyze as part of extensions pom.xml. I'll
let you know when I'm done.

On Thu, Jun 2, 2016 at 3:04 PM, Mariano Gonzalez <notifications@github.com>
wrote:

> In extensions/all-extensions/pom.xml
> <https://github.com/mulesoft/mule/pull/2743#discussion_r65589012>:
>
> > @@ -23,6 +23,25 @@
> >              <artifactId>mule-module-file</artifactId>
> >              <version>${project.version}</version>
> >              <type>zip</type>
> > +            <exclusions>
>
> All of that is fine, but what I'm saying is that the problem goes way
> beyond extensions. It will be very hard for a developer to figure out that
> the commons-io classes that the IDE and test cases allows to use, will not
> be visible in runtime.
>
> What I'm saying is that we should:
>
>    - Make sure that just because my maven module depend on mule-core, it
>    doesn't get exposed to mule-core dependencies which are not going to be
>    available in runtime
>    - Improve our unit test infrastructure so that its classloaders are
>    closer to the runtime ones
>    - Develop a Mojo which automatically adds the exclusions which you're
>    explicitly adding here... Otherwise this doesn't scale
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/mulesoft/mule/pull/2743/files/c98fb624efa9b9914f2b46038515d8dd47bd2159#r65589012>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AG96KYRmKhy7a-y283MUzc9pqxC8hbalks5qHxtGgaJpZM4Isq0Y>
> .
>



-- 
<http://www.mulesoft.com/>
Guillermo Fernandes, Senior Software Engineer

Av. Corrientes 316 Entre Piso, Buenos Aires, C1043AAQ, Argentina


<https://www.linkedin.com/company/www.mulesoft.com>
<https://twitter.com/mulesoft>   <https://www.facebook.com/MuleSoft>
<https://www.youtube.com/user/mulesoftvids> We're hiring!
<http://www.mulesoft.com/careers>

extensions/ftp/pom.xml
gsfernandes: Ok, sure. Thanks!

On Thu, Jun 2, 2016 at 3:10 PM, Mariano Gonzalez <notifications@github.com>
wrote:

> In extensions/ftp/pom.xml
> <https://github.com/mulesoft/mule/pull/2743#discussion_r65590038>:
>
> > @@ -29,11 +29,38 @@
> >              <groupId>commons-net</groupId>
> >              <artifactId>commons-net</artifactId>
> >          </dependency>
> > +
> > +        <dependency>
>
> check with Kraan, but I'm under the impression that the java and javax
> ones are going to be runtime only
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/mulesoft/mule/pull/2743/files/c98fb624efa9b9914f2b46038515d8dd47bd2159#r65590038>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AG96KaETnm2Ugrj-QbnCnywu3rA6ULgeks5qHxyzgaJpZM4Isq0Y>
> .
>



-- 
<http://www.mulesoft.com/>
Guillermo Fernandes, Senior Software Engineer

Av. Corrientes 316 Entre Piso, Buenos Aires, C1043AAQ, Argentina


<https://www.linkedin.com/company/www.mulesoft.com>
<https://twitter.com/mulesoft>   <https://www.facebook.com/MuleSoft>
<https://www.youtube.com/user/mulesoftvids> We're hiring!
<http://www.mulesoft.com/careers>

extensions/all-extensions/pom.xml
gsfernandes: Ok, I'll remove the import and use Supplier from JDK.

On Thu, Jun 2, 2016 at 3:11 PM, Mariano Gonzalez <notifications@github.com>
wrote:

> In extensions/all-extensions/pom.xml
> <https://github.com/mulesoft/mule/pull/2743#discussion_r65590151>:
>
> > +                    <artifactId>javax.inject</artifactId>
> > +                </exclusion>
> > +                <exclusion>
> > +                    <groupId>commons-io</groupId>
> > +                    <artifactId>commons-io</artifactId>
> > +                </exclusion>
> > +                <exclusion>
> > +                    <groupId>commons-lang</groupId>
> > +                    <artifactId>commons-lang</artifactId>
> > +                </exclusion>
> > +                <exclusion>
> > +                    <groupId>commons-collections</groupId>
> > +                    <artifactId>commons-collections</artifactId>
> > +                </exclusion>
> > +                <exclusion>
> > +                    <groupId>com.google.guava</groupId>
>
> that's a bug we need to fix. For sure I mean java.util.function.Supplier
> but screwed up with the import
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/mulesoft/mule/pull/2743/files/c98fb624efa9b9914f2b46038515d8dd47bd2159#r65590151>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AG96KS03xd6IJVbHF_i_bJe-H_mvf0IZks5qHxzZgaJpZM4Isq0Y>
> .
>



-- 
<http://www.mulesoft.com/>
Guillermo Fernandes, Senior Software Engineer

Av. Corrientes 316 Entre Piso, Buenos Aires, C1043AAQ, Argentina


<https://www.linkedin.com/company/www.mulesoft.com>
<https://twitter.com/mulesoft>   <https://www.facebook.com/MuleSoft>
<https://www.youtube.com/user/mulesoftvids> We're hiring!
<http://www.mulesoft.com/careers>

gsfernandes: I couldn't remove it because it is also using InmutableSet and InmutableList from guava. 

Also while adding maven-dependency-plugin to check this on extensions parent pom I realised that commons-lang is used by validation extension so I needed to add it.

extensions/pom.xml
```diff
@@ -129,8 +129,9 @@
                                 <ignoredUsedUndeclaredDependency>junit:junit::</ignoredUsedUndeclaredDependency>
                                 <ignoredUsedUndeclaredDependency>commons-logging:commons-logging::</ignoredUsedUndeclaredDependency>
 
-                                <!-- Logging related artifacts are exposed by the runtime -->
+                                <!-- Logging related artifacts are exposed by the BOOTSTRAP ClassLoader -->
                                 <ignoredUsedUndeclaredDependency>org.slf4j:slf4j-api:jar:</ignoredUsedUndeclaredDependency>
+                                <ignoredUsedUndeclaredDependency>javax.inject:javax.inject::</ignoredUsedUndeclaredDependency>
```
marianogonzalez: extra semicolon?

gsfernandes: No, it is too bad that these properties cannot be have also the scope so it would be more easy to exclude all test dependencies with just one rule. But is optional so let me remove the unnecessary segment from those patterns.

```
List of dependencies that will be ignored if they are used but undeclared. The filter syntax is:
[groupId]:[artifactId]:[type]:[version]
where each pattern segment is optional and supports full and partial * wildcards. An empty pattern segment is treated as an implicit wildcard. *
For example, org.apache.* will match all artifacts whose group id starts with org.apache., and :::*-SNAPSHOT will match all snapshot artifacts.
```

PR #2754, opened 2016-06-03
MULE-9615: Extension configs can not define the same Source/Operation…

marianogonzalez:
… Class


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/describer/AnnotationsBasedDescriber.java
```diff
@@ -708,6 +749,7 @@ private void addTypeRestrictions(ParameterDeclarer parameter, ParsedParameter de
 
     private interface ModelPropertyContributor
     {
+
```
elrodro83: remove extra line

marianogonzalez: intellJ is the one that adds this line when the format runs

PR #2755, opened 2016-06-06
MULE-9810 Avoid exclusions for transitive dependencies in all-extensi…

gsfernandes:
…ons pom

_Changed way the distro is assemble in order to avoid https://issues.apache.org/jira/browse/MASSEMBLY-799


gsfernandes:
We'll research another alternative to solve this issue.

PR #2757, opened 2016-06-06
MULE-9824: Rename transports packages so they don't collide with the

elrodro83:
core packages


core/src/main/java/org/mule/runtime/core/DefaultMuleEvent.java
```diff
@@ -1135,4 +1135,29 @@ public ProcessorsTrace getProcessorsTrace()
...
+     * @param credentials
+     * @param encoding
+     * @param exchangePattern
+     * @param name
+     * @param uri
+     * @param timeout
+     * @param transacted
+     */
+    @Deprecated
+    public void setEndpointFileds(Credentials credentials, String encoding, MessageExchangePattern exchangePattern, String name, URI uri, int timeout, boolean transacted)
```
pablolagreca: setEndpointFileds -> setEndpointFields

PR #2762, opened 2016-06-07
MULE-9838:StringIndexOutOfBoundsException while documenting the schema

alegmarra:


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/schema/AnnotationProcessorUtils.java
```diff
@@ -338,6 +338,10 @@ private static void parseJavaDoc(ProcessingEnvironment processingEnv, Element el
         while (st.hasMoreTokens())
         {
             String nextToken = st.nextToken().trim();
+            if (nextToken.isEmpty())
```
marianogonzalez: test case?

PR #2766, opened 2016-06-07
MULE-7780: Adding property to configure whether to delete Mule after test is finished. In that case logs are moved if troubleshooting is needed.

asequeira:


tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleInstallation.java
```diff
@@ -71,13 +77,21 @@ protected void before() throws Throwable
     @Override
     protected void after()
     {
-        try
+        File logs = new File(muleHome, "logs");
+        File dest = new File(testname + ".logs");
+        deleteQuietly(dest);
+        String deleteOnExit = getProperty("mule.test.deleteOnExit");
```
aiannucci: We can save this property value in a constant.

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleInstallation.java
```diff
@@ -71,13 +77,21 @@ protected void before() throws Throwable
...
+        if (StringUtils.isEmpty(deleteOnExit) || parseBoolean(deleteOnExit))
```
aiannucci: static import

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleInstallation.java
```diff
@@ -99,7 +113,7 @@ private void unzip(File file, File destDir) throws IOException
             }
             else
             {
-                FileUtils.copyInputStreamToFile(zip.getInputStream(entry), destFile);
+                copyInputStreamToFile(zip.getInputStream(entry), destFile);
                 chmodRwx(destFile);
             }
         }
```
aiannucci: You should close() and/or finalize() the ZIP file after you finish using it.

PR #2774, opened 2016-06-09
MULE-9848 - Need to be able to provide a supplier when parsing a List

pablolagreca:


modules/spring-config/src/main/java/org/mule/runtime/config/spring/dsl/api/AttributeDefinition.java
```diff
@@ -264,6 +268,20 @@ public static Builder fromChildListConfiguration(Class<?> type)
         }
 
         /**
+         * Used when the attribute type is collection and the collection implementation must
+         * be of a given type.
+         *
+         * @param type collection type.
+         * @return the builder.
+         */
+        public Builder withCollectionType(Class<? extends Collection> type)
```
marianogonzalez: wouldn't it be better if you receive a Supplier instead so that you can fine tune how the collection is created?

modules/spring-config/src/main/java/org/mule/runtime/config/spring/dsl/spring/WrapperComponentConfig.java
```diff
@@ -0,0 +1,41 @@
...
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.config.spring.dsl.spring;
+
+import java.util.Collection;
+import java.util.Optional;
+
+/**
+ * Holder for wrapper element configuration configuration
```
marianogonzalez: add `@since`

PR #2781, opened 2016-06-10
MULE-9288: Make DataType immutable

elrodro83:


core/src/main/java/org/mule/runtime/core/DefaultMuleMessage.java
```diff
@@ -124,17 +127,19 @@
         }
         else
         {
-            DataType<?> dataType = DataTypeFactory.create(payload.getClass(), previous.getDataType().getMimeType());
-            dataType.setEncoding(previous.getDataType().getEncoding());
+            DataType<?> dataType = DataTypeFactory.create(payload.getClass(),
```
dfeist: Add TODO and reference MULE-9855

core/src/main/java/org/mule/runtime/core/DefaultMuleMessage.java
```diff
@@ -124,17 +127,19 @@
...
+                    previous.getDataType().getMimeType(),
```
dfeist: Wouldn't it be best to add copy (template) functionality to factory and unit test it there?

core/src/main/java/org/mule/runtime/core/DefaultMuleMessage.java
```diff
@@ -124,17 +127,19 @@
...
 
             return dataType;
         }
     }
 
     private static DataType<?> getCloningMessageDataType(MuleMessage previous)
     {
-        DataType<?> dataType = DataTypeFactory.create(previous.getDataType().getType(), previous.getDataType().getMimeType());
-        dataType.setEncoding(previous.getDataType().getEncoding());
+        DataType<?> dataType = DataTypeFactory.create(previous.getDataType().getType(),
```
dfeist: Ditto above

core/src/main/java/org/mule/runtime/core/transformer/types/DataTypeFactory.java
```diff
@@ -308,4 +306,52 @@ else if (classInMap != null)
         proxyClassCache.put(typeName, new ProxyIndicator(type, isProxy));
         return isProxy;
     }
+
+    public static <T> DataType<T> createFromDataType(DataType<T> typeFrom, String mimeType, String encoding)
```
dfeist: Are these methods temporary, until proper builder is defined? If so, add TODO's.

core/src/main/java/org/mule/runtime/core/transformer/types/DataTypeFactory.java
```diff
@@ -308,4 +306,52 @@ else if (classInMap != null)
...
+                return new CollectionDataType(collectionType, itemType, mimeType, encoding);
+            }
+        }
+        else
+        {
+            return new SimpleDataType<>(typeFrom.getType(), mimeType, encoding);
+        }
+    }
+
+    public static <T> DataType<T> createFromDataTypeWithMimeType(DataType<T> typeFrom, String mimeType)
```
dfeist: Method name is inconsistent with other methods. I understand it's because you can't overload, but it's not very nice.

core/src/main/java/org/mule/runtime/core/DefaultMuleMessage.java
elrodro83: Being added as part of https://www.mulesoft.org/jira/browse/MULE-9895

core/src/main/java/org/mule/runtime/core/transformer/types/DataTypeFactory.java
elrodro83: Will be refactored as part of https://www.mulesoft.org/jira/browse/MULE-9895

core/src/main/java/org/mule/runtime/core/transformer/types/CollectionDataType.java
```diff
@@ -36,33 +36,46 @@
     public CollectionDataType(Class<C> collectionType)
     {
         super(Object.class);
+        checkCollectionTypeForNull(collectionType);
```
dfeist: Use PreConditions

PR #2785, opened 2016-06-13
MULE-9897 - Support type converter on simple type collections

pablolagreca:


modules/spring-config/src/main/java/org/mule/runtime/config/spring/dsl/api/ComponentBuildingDefinition.java
```diff
@@ -251,6 +285,9 @@ public ComponentBuildingDefinition build()
             checkState(definition.typeDefinition != null, "You must specify the type");
             checkState(identifier != null, "You must specify the identifier");
             checkState(namespace != null, "You must specify the namespace");
+            Optional<Class> componentType = getType();
+            checkState(definition.typeConverter == null || (definition.typeConverter != null && componentType.isPresent()), TYPE_CONVERTER_AND_UNKNOWN_TYPE_MESSAGE);
```
marianogonzalez: optional

PR #2788, opened 2016-06-13
MULE-9338: Fix flaky SFTP tests

elrodro83:
* Downgrade apache mina sshd to a version compatible with java 6


tests/functional/src/main/java/org/mule/tck/util/sftp/SftpServer.java
```diff
@@ -97,7 +96,7 @@ public void stop()
         {
             sshdServer.stop(false);
         }
-        catch (IOException e)
+        catch (InterruptedException e)
         {
             e.printStackTrace();
```
marianogonzalez: this could give you jail time in some countries

PR #2790, opened 2016-06-14
MULE-9866: Provide better logging for classloading process

elrodro83:


core/src/main/java/org/mule/runtime/core/api/config/MuleProperties.java
```diff
@@ -189,5 +189,6 @@
     public static final String MULE_LOG_CONTEXT_DISPOSE_DELAY_MILLIS = SYSTEM_PROPERTY_PREFIX + "log.context.dispose.delay.millis";
     public static final String MULE_DEFAULT_PROCESSING_STRATEGY = SYSTEM_PROPERTY_PREFIX + "default.processing.strategy";
     public static final String MULE_FLOW_TRACE = SYSTEM_PROPERTY_PREFIX + "flowTrace";
+    public static final String MULE_LOG_FILTERED_CLASSLOADING = SYSTEM_PROPERTY_PREFIX + "filteredClassLoading.log";
```
pablokraan: I would use something like "mule.classloading.vebose"

modules/artifact/src/main/java/org/mule/runtime/module/artifact/classloader/CompositeClassLoader.java
```diff
@@ -75,6 +78,12 @@ public ClassLoaderLookupPolicy getClassLoaderLookupPolicy()
         final ClassLoaderLookupStrategy lookupStrategy = lookupPolicy.getLookupStrategy(name);
         Class<?> result;
 
+        /*
+         * Gather information about the exceptions in each of the contained classloaders to provide
+         * troubleshooting information in case of throwing a ClassNotFoundException.
```
pablokraan: Add link to ClassNotFoundException

modules/artifact/src/main/java/org/mule/runtime/module/artifact/classloader/CompositeClassLoader.java
```diff
@@ -87,12 +96,13 @@ else if (lookupStrategy == PARENT_FIRST)
             }
             catch (ClassNotFoundException e)
             {
-                result = doLoadClass(name);
+                exceptions.add(e);
```
pablokraan: Not sure that this info will be helpful. Each composed classloader will be following the path up to the container classloader. So, I guess the list of exceptions will contain multiple occurrences of a classnotfound on the container classloader

modules/artifact/src/main/java/org/mule/runtime/module/artifact/classloader/FilteringArtifactClassLoader.java
```diff
@@ -66,6 +75,16 @@ public URL getResource(String name)
...
         else
         {
+            if (logFilteredClassloading)
+            {
+                logger.info(format("Resource '%s' not found in classloader for '%s'.", name, getArtifactName()));
+                logger.info(format("Filter applied for resource '%s': %s", name, filter.toString()));
+            }
+            else if (logger.isTraceEnabled())
+            {
+                logger.trace(format("Resource '%s' not found in classloader for '%s'.", name, getArtifactName()));
```
pablokraan: Try to remove the duplication in the message formating

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleApplicationClassLoader.java
```diff
@@ -27,14 +27,6 @@ public MuleApplicationClassLoader(String appName, ClassLoader parentCl, NativeLi
     }
 
     @Override
-    public String toString()
```
pablokraan: Why did you remove the toString from classloader classes?

modules/artifact/src/main/java/org/mule/runtime/module/artifact/classloader/CompositeClassLoader.java
elrodro83: This is a regular comment, not a javadoc

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleApplicationClassLoader.java
elrodro83: It was moved to the superclass

modules/artifact/src/main/java/org/mule/runtime/module/artifact/classloader/exception/CompositeClassNotFoundException.java
```diff
@@ -0,0 +1,69 @@
...
+
+import org.mule.runtime.module.artifact.classloader.ClassLoaderLookupStrategy;
+
+import java.util.List;
+
+/**
+ * Extends {@link ClassNotFoundException}, composing the individual exceptions of each place where
+ * the class was looked for and wasn't found.
+ */
+public class CompositeClassNotFoundException extends ClassNotFoundException
```
pablokraan: We would need some test for this new feature. Not sure if that is doable, but having a feature that has no test and won't be activated by default  makes it prone to problems. Can you review if is possible to add tests ensuring that the right exceptions are gathered and the right messages created?

modules/artifact/src/test/java/org/mule/runtime/module/artifact/classloader/CompositeClassLoaderTestCase.java
```diff
@@ -96,9 +106,12 @@ public void usesParentFirstThenChildLookup() throws ClassNotFoundException
         assertThat(aClass, equalTo(CLASS_FROM_CLASSLOADER1));
     }
 
-    @Test(expected = ClassNotFoundException.class)
+    @Test
     public void  usesParentFirstAndChildLookupAndFails() throws ClassNotFoundException
     {
+        expected.expect(CompositeClassNotFoundException.class);
+        expected.expectMessage(startsWith("Cannot load class '" + CLASS_NAME + "': ["));
```
pablokraan: Should not assert that there are multiple exceptions in the composite exception?

modules/artifact/src/test/java/org/mule/runtime/module/artifact/classloader/CompositeClassLoaderTestCase.java
```diff
@@ -146,9 +159,11 @@ public void usesChildFirstThenParentLookup() throws ClassNotFoundException
         assertThat(aClass, equalTo(CLASS_FROM_CLASSLOADER1));
     }
 
-    @Test(expected = ClassNotFoundException.class)
+    @Test
     public void usesChildFirstThenParentLookupAndFails() throws ClassNotFoundException
     {
+        expected.expect(ClassNotFoundException.class);
```
pablokraan: Same here

modules/artifact/src/test/java/org/mule/runtime/module/artifact/classloader/FineGrainedControlClassLoaderTestCase.java
```diff
@@ -52,6 +61,9 @@ public void usesParentOnlyLookupAndFails() throws Exception
         final ClassLoaderLookupPolicy lookupPolicy = mock(ClassLoaderLookupPolicy.class);
         when(lookupPolicy.getLookupStrategy(TEST_CLASS_NAME)).thenReturn(PARENT_ONLY);
 
+        expected.expect(CompositeClassNotFoundException.class);
+        expected.expectMessage(startsWith("Cannot load class '" + TEST_CLASS_NAME + "': [ERROR" + lineSeparator() + "]"));
```
pablokraan: No test is checking that the composite exception contains the different exceptions (from the parent and the child if corresponds in each case)

PR #2793, opened 2016-06-14
MULE-9914: Unify loggers used in mule classes

pablokraan:
_ Updating core logger in core module


core/src/main/java/org/mule/runtime/core/transaction/MuleTransactionConfig.java
```diff
@@ -221,7 +220,7 @@ public ConstraintFilter getConstraint()
         }
         catch (CloneNotSupportedException e)
         {
-            logger.fatal("Failed to clone ConstraintFilter: " + e.getMessage(), e);
+            logger.error("Failed to clone ConstraintFilter: " + e.getMessage(), e);
```
gsfernandes: Don't know if we are using >= 1.6.0 but if we do, we could use the formatting support when logging exceptions too:
http://slf4j.org/faq.html#paramException
```
String s = "Hello world";
try {
  Integer i = Integer.valueOf(s);
} catch (NumberFormatException e) {
  logger.error("Failed to format {}", s, e);
}
```

core/src/main/java/org/mule/runtime/core/util/xa/XaTransactionRecoverer.java
```diff
@@ -91,15 +91,15 @@ public void rollbackDandlingTransaction(Xid xid) throws XAException
     {
         try
         {
-            logger.info("Rollbacking danling tx with id " + xid);
+            logger.info("Rollbacking dangling tx with id " + xid);
```
gsfernandes: Minor: but we could also use the formatting support here
```
logger.info("Rollbacking dangling tx with id {}", xid);
```

dfeist:
At some point we should make declarations consistent too: i) consistently use (or not) static instance ii) remove 'transient'

pablokraan:
@dfeist can you create a jira to track that?

dfeist:
https://www.mulesoft.org/jira/browse/MULE-9922

PR #2801, opened 2016-06-15
MULE-9914: Unify loggers used in mule classes

pablokraan:
_ Improving logging of exceptions

### DIFF
--- a/transports/http/src/main/java/org/mule/compatibility/transport/http/HttpMessageProcessTemplate.java
+++ b/transports/http/src/main/java/org/mule/compatibility/transport/http/HttpMessageProcessTemplate.java
@@ -224,8 +224,8 @@ public class HttpMessageProcessTemplate extends AbstractTransportMessageProcessT
             }
             catch (Exception e)
             {
-                final String errorMessage = "Exception sending http response after error: ";
-                logger.warn(errorMessage + e.getMessage());
+                final String errorMessage = "Exception sending http response after error";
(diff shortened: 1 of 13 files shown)


transports/http/src/main/java/org/mule/compatibility/transport/http/HttpMessageProcessTemplate.java
```diff
@@ -224,10 +224,11 @@ public void afterFailureProcessingFlow(MuleException exception) throws MuleExcep
             }
             catch (Exception e)
             {
-                logger.warn("Exception sending http response after error: " + e.getMessage());
+                final String errorMessage = "Exception sending http response after error: ";
+                logger.warn(errorMessage + e.getMessage());
                 if (logger.isDebugEnabled())
                 {
-                    logger.debug(e.toString());
+                    logger.debug(errorMessage, e);
```
elrodro83: The ": " ending of the message shouldn't be logged in this case, only on the previous one.

transports/jms/src/main/java/org/mule/compatibility/transport/jms/MultiConsumerJmsMessageReceiver.java
```diff
@@ -299,7 +299,7 @@ private void recoverSession()
             }
             catch (Exception jmsEx)
             {
-                logger.error(jmsEx.toString());
+                logger.error("CAnnot recover session", jmsEx);
```
elrodro83: Cannot

PR #2803, opened 2016-06-16
MULE-9930 Remove original payload from MuleMessage

dfeist:


core/src/main/java/org/mule/runtime/core/api/config/MuleConfiguration.java
```diff
@@ -41,8 +41,6 @@
 
     boolean isCacheMessageAsBytes();
```
pablokraan: Should not remove also this method?

dfeist: This is different functionality separate to keeping hold of the original message. This needs thinking about separately.

PR #2805, opened 2016-06-16
MULE-9147 Remove all backward compatibility flags and configuration o…

dfeist:
…ptions


core/src/main/java/org/mule/runtime/core/transformer/AbstractTransformer.java
```diff
@@ -259,28 +259,12 @@ public Object transform(Object src, String enc) throws TransformerException
...
-                return payload;
-            }
-            else
-            {
-                Message msg = CoreMessages.transformOnObjectUnsupportedTypeOfEndpoint(getName(),
-                        payload.getClass());
-                /// FIXME
-                throw new TransformerException(msg, this);
-            }
+            Message msg = CoreMessages.transformOnObjectUnsupportedTypeOfEndpoint(getName(),
```
pablolagreca: static import

PR #2806, opened 2016-06-16
MULE-9934: Extract HTTP response creation logic from response builder…

afelisatti:
… component

### DIFF
--- a/extensions/http/src/main/java/org/mule/extension/http/api/listener/builder/HttpListenerResponseBuilder.java
+++ b/extensions/http/src/main/java/org/mule/extension/http/api/listener/builder/HttpListenerResponseBuilder.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.extension.http.api.listener.builder;
+
+import org.mule.extension.http.api.HttpMessageBuilder;
+import org.mule.runtime.core.api.MuleEvent;
+import org.mule.runtime.extension.api.annotation.Alias;
+import org.mule.runtime.extension.api.annotation.Parameter;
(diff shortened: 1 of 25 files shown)


tests/http/src/test/resources/http-request-query-params-config.xml
```diff
@@ -47,8 +47,8 @@
         <httpn:request config-ref="config" path="testPath"  key="ANY">
             <httpn:request-builder>
                 <!-- Not currently supported. -->
-                <!--<httpn:requester-request-builder queryParams="#[{'testName1':null, 'testName2': [null]}]"/>-->
-                <httpn:requester-request-builder queryParams="#[{'testName1':null, 'testName2': null}]"/>
+                <!--<httpn:simple-request-builder queryParams="#[{'testName1':null, 'testName2': [null]}]"/>-->
```
pablokraan: Remove commented lines

extensions/http/src/main/java/org/mule/extension/http/api/listener/builder/HttpListenerResponseBuilder.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.function.Function;
+
+import javax.activation.DataHandler;
+
+/**
+ * Component that specifies how to create a proper HTTP response.
+ *
+ * @since 4.0
+ */
+@Alias("simple-response-builder")
```
pablokraan: Why "simple"? Is there a "complex-response-bulder"?

tests/http/src/test/resources/http-request-query-params-config.xml
afelisatti: I'm keeping the lines commented to have a clear idea of what we should test once we support query params collections.

extensions/http/src/main/java/org/mule/extension/http/api/listener/builder/HttpListenerResponseBuilder.java
afelisatti: No, it's just to standardise both listener and request side builders. These should not survive long in any case.

PR #2808, opened 2016-06-16
MULE-9933: RestrictedSSLSocketFactory needs to implement getDefault

bayitajesi:
method so it can be set as the socket factory used by the Ldap provider in the JVM


aiannucci:
Please create the PR to merge into mule-3.x branch

PR #2809, opened 2016-06-16
MULE-9933: RestrictedSSLSocketFactory needs to implement getDefault

bayitajesi:
method so it can be set as the socket factory used by the Ldap provider in the JVM

### DIFF
--- a/core/src/main/java/org/mule/api/security/tls/RestrictedSSLSocketFactory.java
+++ b/core/src/main/java/org/mule/api/security/tls/RestrictedSSLSocketFactory.java
@@ -6,12 +6,16 @@
  */
 package org.mule.api.security.tls;
 
+import org.mule.api.MuleRuntimeException;
+import org.mule.config.i18n.Message;
+import org.mule.config.i18n.MessageFactory;
 import org.mule.util.ArrayUtils;
 
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.Socket;
 
+import javax.net.SocketFactory;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLSocket;
 import javax.net.ssl.SSLSocketFactory;
@@ -27,6 +31,7 @@ public class RestrictedSSLSocketFactory extends SSLSocketFactory
     private final String[] enabledCipherSuites;
     private final String[] enabledProtocols;
     private final String[] defaultCipherSuites;
+    private static RestrictedSSLSocketFactory defaultSocketFactory = null;
 
     public RestrictedSSLSocketFactory(SSLContext sslContext, String[] cipherSuites, String[] protocols)
     {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mule/api/security/tls/RestrictedSSLSocketFactory.java
```diff
@@ -100,4 +101,22 @@ private SSLSocket restrictCipherSuites(SSLSocket socket)
         socket.setEnabledProtocols(enabledProtocols);
         return socket;
     }
+
+    public static synchronized SocketFactory getDefault()
+    {
+        SocketFactory factory = null;
+        try
+        {
+            TlsPropertiesMapper propertiesMapper = new TlsPropertiesMapper(TlsConfiguration.JSSE_NAMESPACE);
```
afelisatti: This is not required.

core/src/main/java/org/mule/api/security/tls/RestrictedSSLSocketFactory.java
```diff
@@ -100,4 +101,22 @@ private SSLSocket restrictCipherSuites(SSLSocket socket)
...
+            TlsConfiguration configuration = new TlsConfiguration(TlsConfiguration.DEFAULT_KEYSTORE);
```
afelisatti: You shouldn't need a keystore for client side connections.

core/src/main/java/org/mule/api/security/tls/RestrictedSSLSocketFactory.java
```diff
@@ -100,4 +101,22 @@ private SSLSocket restrictCipherSuites(SSLSocket socket)
...
+            propertiesMapper.readFromProperties(configuration, System.getProperties());
+            configuration.initialise(null == configuration.getKeyStore(), TlsConfiguration.JSSE_NAMESPACE);
```
afelisatti: The namespace is not required.

core/src/main/java/org/mule/api/security/tls/RestrictedSSLSocketFactory.java
```diff
@@ -100,4 +102,24 @@ private SSLSocket restrictCipherSuites(SSLSocket socket)
         socket.setEnabledProtocols(enabledProtocols);
         return socket;
     }
+
+    public static synchronized SocketFactory getDefault()
+    {
+        if(defaultSocketFactory != null)
```
afelisatti: I think it would make more sense to check for null and in that case create and assign the field. Then you'd only have one return.

core/src/main/java/org/mule/api/security/tls/RestrictedSSLSocketFactory.java
```diff
@@ -100,4 +102,24 @@ private SSLSocket restrictCipherSuites(SSLSocket socket)
...
+
+        try
+        {
+            TlsConfiguration configuration = new TlsConfiguration(null);
+            configuration.initialise(true, null);
+            defaultSocketFactory = new RestrictedSSLSocketFactory(configuration.getSslContext(), configuration.getEnabledCipherSuites(), configuration.getEnabledProtocols());
+        }
+        catch (Exception e)
+        {
+            throw new RuntimeException("Could not create the default RestrictedSSLSocketFactory", e);
```
afelisatti: Throw a MuleRuntimeException just so it's clear we are failing and not something else.

PR #2812, opened 2016-06-17
MULE-9885: Interface can be used as the type of a @ParameterGroup yet…

pablomusumeci:
… no error or warning is thrown

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/ParameterModelValidator.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/ParameterModelValidator.java
@@ -6,6 +6,7 @@
  */
 package org.mule.runtime.module.extension.internal.introspection.validation;
 
+import static java.lang.String.format;
 import static java.util.Collections.emptyMap;
 import static java.util.stream.Collectors.toList;
 import static org.mule.metadata.java.utils.JavaTypeUtils.getType;
@@ -20,16 +21,20 @@ import org.mule.metadata.api.model.MetadataType;
 import org.mule.metadata.api.model.ObjectFieldType;
 import org.mule.metadata.api.model.ObjectType;
 import org.mule.metadata.api.visitor.MetadataTypeVisitor;
+import org.mule.runtime.extension.api.ExtensionWalker;
 import org.mule.runtime.extension.api.exception.IllegalModelDefinitionException;
 import org.mule.runtime.extension.api.introspection.ExtensionModel;
 import org.mule.runtime.extension.api.introspection.config.ConfigurationModel;
 import org.mule.runtime.extension.api.introspection.connection.ConnectionProviderModel;
 import org.mule.runtime.extension.api.introspection.operation.OperationModel;
 import org.mule.runtime.extension.api.introspection.parameter.ParameterModel;
+import org.mule.runtime.extension.api.introspection.parameter.ParameterizedModel;
 import org.mule.runtime.extension.api.introspection.property.ImportedTypesModelProperty;
 import org.mule.runtime.extension.api.introspection.property.SubTypesModelProperty;
+import org.mule.runtime.extension.api.introspection.source.SourceModel;
 import org.mule.runtime.module.extension.internal.exception.IllegalParameterModelDefinitionException;
 import org.mule.runtime.module.extension.internal.introspection.SubTypesMappingContainer;
+import org.mule.runtime.module.extension.internal.model.property.ParameterGroupModelProperty;
 
 import com.google.common.collect.ImmutableMap;
 
@@ -58,6 +63,7 @@ public final class ParameterModelValidator implements ModelValidator
     private static final String CONFIGURATION = "configuration";
     private static final String OPERATION = "operation";
     private static final String CONNECTION_PROVIDER = "connection provider";
+    private static final String SOURCE = "source";
 
     private SubTypesMappingContainer subTypesMapping;
     private Map<MetadataType, MetadataType> importedTypes;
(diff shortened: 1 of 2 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/describer/AnnotationsBasedDescriber.java
```diff
@@ -443,6 +446,15 @@ private void declareParameterGroups(Class<?> annotatedType, ParameterizedDeclare
                                                                           annotatedType));
             }
 
+            if (!isInstantiable(field.getType(), false))
```
marianogonzalez: this is in the wrong place. This should be done in a ModelEnricher which uses an ExtensionWalker to verify all the parameters which have the proper model property. If you do it here, then you're not validating alternative describers which might have the same bug

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/ParameterModelValidator.java
```diff
@@ -117,36 +121,20 @@ public void visitObject(ObjectType objectType)
...
-            validateNameCollisionWithTypes(parameterModel, ownerName, ownerModelType, extensionName,
-                                           parameters.stream().map(p -> hyphenize(p.getName())).collect(toList()));
-        });
+        new ExtensionWalker()
+        {
+            @Override
+            public void onParameter(ParameterizedModel owner, ParameterModel model)
+            {
+                String ownerName = owner.getName();
+                String ownerModelType = owner.getClass().getSimpleName();
```
pablomusumeci: We used to have an specific tag (CONNECTION_PROVIDER, OPERATION, CONFIGURATION, etc) to use in the error message,  but now with the ExtensionWalker it is no longer possible to identify what kind of component is owner. Any idea?

marianogonzalez: a simple instanceof on the owner will do

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/ParameterModelValidator.java
```diff
@@ -191,4 +181,50 @@ private void validateNameCollisionWithTypes(ParameterModel parameterModel, Strin
...
+
+        if (parameterGroup != null)
+        {
+            parameterGroup.getGroups().stream()
+                    .map(p -> p.getField())
+                    .forEach(field -> {
+                        if (!isInstantiable(field.getType(), false))
+                        {
+                            throw new IllegalParameterModelDefinitionException(
+                                    format("Parameter group type should be instantiable. Affected field [%s] in [%s].",
```
marianogonzalez: rephase this to something more understandable. Say something like `an instantiable type with a default constructor was expected, however type %s was expected`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/ParameterModelValidator.java
```diff
@@ -191,4 +181,50 @@ private void validateNameCollisionWithTypes(ParameterModel parameterModel, Strin
...
+    private void validateParameterGroup(ParameterModel parameterModel)
+    {
+        ParameterGroupModelProperty parameterGroup = parameterModel
+                .getModelProperty(ParameterGroupModelProperty.class)
+                .orElse(null);
+
+        if (parameterGroup != null)
+        {
+            parameterGroup.getGroups().stream()
+                    .map(p -> p.getField())
```
marianogonzalez: you don't have to check the fields... you have to check the type

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/ParameterModelValidator.java
```diff
@@ -190,14 +190,14 @@ private void validateParameterGroup(ParameterModel parameterModel)
...
+                    .forEach(parameter -> {
+                        if (!isInstantiable(parameter.getType()))
                         {
                             throw new IllegalParameterModelDefinitionException(
-                                    format("Parameter group type should be instantiable. Affected field [%s] in [%s].",
-                                           field.getName(),
-                                           field.getAnnotatedType()));
+                                    format("ParameterGroup type should be non abstract with a default constructor, however type '%s' was received. Affected field [%s] in [%s].",
+                                           parameter.getType().getName(),
+                                           parameter.getField().getName(),
```
marianogonzalez: why do you insist with the field?? the field is an internal thing which has nothing to do with this. If the `@ParameterGroup` is on an operation, what does the field tell the user. Just use parameter.getType() and nothing else

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/ParameterModelValidator.java
```diff
@@ -191,4 +181,46 @@ private void validateNameCollisionWithTypes(ParameterModel parameterModel, Strin
 
     }
 
-}
+    private void validateParameterGroup(ParameterModel parameterModel, String ownerName, String ownerModelType, String extensionName)
+    {
+        ParameterGroupModelProperty parameterGroup = parameterModel
+                .getModelProperty(ParameterGroupModelProperty.class)
+                .orElse(null);
```
alegmarra: don't do this and check for null later, it's already an optional, go for `ifPresent`

PR #2817, opened 2016-06-21
Update raml parser dependencies

svacas:


distributions/standalone/assembly-whitelist.txt
```diff
@@ -374,7 +375,7 @@
 /mule-standalone-${productVersion}/lib/opt/parser-interface-impl-v2-3.8.1-+
 /mule-standalone-${productVersion}/lib/opt/quartz-2.2.1.jar
 /mule-standalone-${productVersion}/lib/opt/raml-parser-0.9-+
-/mule-standalone-${productVersion}/lib/opt/raml-parser-2-0.1.0-+
+/mule-standalone-${productVersion}/lib/opt/raml-parser-2-1.0.0-+
```
pablokraan: Why you use the -+ suffix instead of using the full file name?

svacas: because it's a SNAPSHOT

PR #2820, opened 2016-06-21
MULE-9944 Make mule 4 depend on 3.8.0 of mule common

machaval:


pom.xml
```diff
@@ -231,7 +231,7 @@
         <xmlSecurityVersion>1.5.8</xmlSecurityVersion>
         <xstreamVersion>1.4.7</xstreamVersion>
         <xaPoolVersion>1.5.0</xaPoolVersion>
-        <muleCommonVersion>3.8.0-SNAPSHOT</muleCommonVersion>
```
asequeira: What about the update in the assembly-whitelist? You know we have a Jenkins job to do this?

machaval: No I don't if so can you make the change. 3.8.0-SNAPSHOT was removed by the SNAPSHOT cleaner

PR #2825, opened 2016-06-21
MULE-9815: Thrown exceptions are unwrapped and trace is lost

juandesi:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/exception/ExceptionEnricherManager.java
```diff
@@ -38,36 +40,33 @@ public ExceptionEnricherManager(RuntimeExtensionModel extensionModel, ExceptionE
...
         }
         else
         {
-            root = ExceptionUtils.getRootCause(e);
-            if (root == null)
-            {
-                root = e;
-            }
+            // unwraps the exception thrown by the reflective operation.
+            Optional<? extends Throwable> failureCause = extractCauseOfType(e, UndeclaredThrowableException.class);
```
marianogonzalez: use orElse()

PR #2837, opened 2016-06-23
MULE-9679: Review rests for new Email Connector

juandesi:


extensions/email/src/main/java/org/mule/extension/email/api/sender/SenderOperations.java
```diff
@@ -60,7 +60,7 @@
      */
     public void send(@Connection SenderConnection connection,
                      @UseConfig SMTPConfiguration configuration,
-                     @Optional(defaultValue = PAYLOAD) EmailContent content,
```
marianogonzalez: the EMailCOntent#body needs to default to the payload, otherwise proxy scenaio becomes complicated

juandesi: I'll add the default again, but my concern is about transforming whats incoming in the payload to an "EmailContent instance", this is not going to work now.

extensions/email/src/main/java/org/mule/extension/email/api/EmailContent.java
```diff
@@ -59,6 +67,16 @@ public String getBody()
      */
     public String getContentType()
     {
-        return contentType;
+        // TODO: remove if when MULE-9960 is fixed since default values are not being injected properly.
+        return contentType == null ? TEXT : contentType;
```
marianogonzalez: this should be fixed now

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/XmlExtensionParserDelegate.java
```diff
@@ -665,12 +665,14 @@ private ValueResolver getPojoValueResolver(ObjectType objectType, ElementDescrip
...
             final String parameterName = objectField.getKey().getName().getLocalPart();
 
+            // TODO: MULE-9960
             ValueResolver resolver = getResolverFromAttribute(element, parameterName, fieldType, null);
             if (resolver == null)
             {
                 ElementDescriptor childElement = element.getChildByName(hyphenize(parameterName));
                 if (childElement != null)
                 {
+                    // TODO: MULE-9960
```
marianogonzalez: this class doesn't exists anymore

PR #2838, opened 2016-06-23
MULE-9776: Fix the http-ext to include it as a zip in the distribution

anosenzo:


modules/http/src/main/resources/META-INF/mule-module.properties
```diff
@@ -12,6 +11,111 @@ artifact.export.classPackages=org.mule.runtime.module.http.api,\
...
+                              org.glassfish.grizzly.utils,\
+                              org.glassfish.grizzly.http.util,\
+                              org.glassfish.grizzly.nio,\
+                              org.glassfish.grizzly.streams,\
+                              org.glassfish.grizzly.nio.transport,\
+                              org.glassfish.grizzly.nio.tmpselectors,\
+                              javax.net.ssl,\
+                              com.ning.http.client.providers.grizzly,\
+                              org.glassfish.grizzly.threadpool,\
+                              com.ning.http.client.simple,\
```
aiannucci: Should we order these alphabetically, just to make it look nicer?

PR #2845, opened 2016-06-24
Mule 9680: Review tests for new TCP/UDP/SSL connectors

pablomusumeci:


extensions/sockets/src/test/java/org/mule/extension/socket/TcpReconnectionTestCase.java
```diff
@@ -0,0 +1,140 @@
...
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@SmallTest
+@RunWith(MockitoJUnitRunner.class)
+public class TcpReconnectionTestCase extends AbstractMuleTestCase
+{
+
+    @Inject
+    MuleContext muleContext;
```
marianogonzalez: private

extensions/sockets/pom.xml
```diff
@@ -33,6 +33,11 @@
 
     <dependencies>
         <dependency>
+            <groupId>commons-io</groupId>
```
marianogonzalez: if you do this then you have to exclude it on the all-extensions pom

extensions/sockets/src/test/java/org/mule/extension/socket/TcpReconnectionTestCase.java
```diff
@@ -0,0 +1,140 @@
...
+
+    @Rule
+    public DynamicPort dynamicPort = new DynamicPort("port");
```
marianogonzalez: per conventions, declare the public rules first

extensions/sockets/src/test/java/org/mule/extension/socket/TcpReconnectionTestCase.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @Test
+    public void closeListener() throws Exception
+    {
+        expectedException.expect(SocketException.class);
+        expectedException.expectMessage("Socket is closed");
+        listenerConnection.connect();
+
+        new Runnable()
```
marianogonzalez: what's this runnable for? this is the same as calling listenerConnection.disconnect(); directly

pablomusumeci: listenerConnection.listen is blocking, so we need to disconnect it from a different thread

marianogonzalez: but you're not doing it in a different thread. You're running it in the current thread

marianogonzalez: besides if you were doing it in a different thread, you need a latch to control race conditions. This test is not doing what you say it does, and if it did, it'd be flaky

PR #2849, opened 2016-06-24
MULE-9911: No error when using file:copy with empty targetPath

marianogonzalez:


modules/file-extension-common/src/main/java/org/mule/runtime/module/extension/file/api/StandardFileSystemOperations.java
```diff
@@ -272,6 +281,7 @@ public void move(@Connection FileSystem fileSystem,
                      @Optional(defaultValue = "true") boolean createParentDirectories,
                      MuleEvent event)
     {
+        validateTargetPath(targetPath);
```
afelisatti: Why are there not test cases for the move case too?

PR #2860, opened 2016-06-27
MULE-9900: Support encoding while sending Strings in the Sockets Exte…

pablomusumeci:
…nsion

### DIFF
--- a/extensions/sockets/src/main/java/org/mule/module/socket/api/SocketOperations.java
+++ b/extensions/sockets/src/main/java/org/mule/module/socket/api/SocketOperations.java
@@ -7,6 +7,7 @@
 package org.mule.module.socket.api;
 
 import org.mule.module.socket.api.client.SocketClient;
+import org.mule.module.socket.api.config.RequesterConfig;
 import org.mule.module.socket.api.connection.RequesterConnection;
 import org.mule.module.socket.api.metadata.SocketMetadataResolver;
 import org.mule.runtime.api.connection.ConnectionException;
@@ -16,6 +17,7 @@ import org.mule.runtime.extension.api.annotation.metadata.MetadataScope;
 import org.mule.runtime.extension.api.annotation.param.Connection;
 import org.mule.runtime.extension.api.annotation.param.NoRef;
 import org.mule.runtime.extension.api.annotation.param.Optional;
+import org.mule.runtime.extension.api.annotation.param.UseConfig;
 
 import java.io.IOException;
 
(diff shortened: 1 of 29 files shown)


extensions/sockets/src/main/java/org/mule/module/socket/api/SocketOperations.java
```diff
@@ -29,27 +31,34 @@
...
+     * @param outputEncoding encoding that will be used to serialize the {@code data} if its type is {@link String}.
      * @param muleMessage if there is no response expected, the outcome of the operation will be
      *                    the same {@link MuleMessage} as the input.
      * @throws ConnectionException if the connection couldn't be established, if the remote host was unavailable.
      */
     @MetadataScope(outputResolver = SocketMetadataResolver.class, keysResolver = SocketMetadataResolver.class)
     public MuleMessage<?, ?> send(@Connection RequesterConnection connection,
+                                  @UseConfig RequesterConfig config,
+                                  @Optional String outputEncoding,
                                   @Optional(defaultValue = "#[payload]") Object data,
```
marianogonzalez: rename to content for consistency with other connectors

extensions/sockets/src/main/java/org/mule/module/socket/api/config/AbstractSocketConfig.java
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Default encoding used for serializing {@link String}. This encoding is used in
+     * {@link SocketOperations#send(RequesterConnection, RequesterConfig, String, Object, String, MuleMessage)} operation
+     * for serializing {@link String} types if no encoding parameter is specified.
+     *
+     * It is also used by the {@link SocketWorker} for encoding the replies send by the {@link SocketListener} source.
+     */
+    @Parameter
+    @Optional
+    String defaultEncoding;
```
marianogonzalez: private

extensions/sockets/src/test/java/org/mule/extension/socket/SendEncodedStringTestCase.java
```diff
@@ -0,0 +1,41 @@
...
+import org.mule.runtime.core.util.IOUtils;
+
+import java.io.InputStream;
+
+import org.junit.Test;
+
+public class SendEncodedStringTestCase extends SocketExtensionTestCase
+{
+    private static final String WEIRD_CHAR_MESSAGE = "This is a messag\u00ea with weird chars \u00f1.";
+    private static final String ENCODING = "iso-8859-1";
```
marianogonzalez: this test is flaky. You're assuming that the default encoding is not iso-8859 just because in your machine this isn't the case. What you should do is to make this dynamic and make sure that you pick a supported Charset which is not your machine's default. 

Look at the FileWriteTestCase for inspiration

PR #2861, opened 2016-06-27
MULE-9958: Make DataType getters return Objects instead of Strings

elrodro83:


core/src/main/java/org/mule/runtime/core/client/DefaultLocalMuleClient.java
```diff
@@ -111,7 +168,35 @@ private MuleException createUnsupportedUrlException(String url)
     public MuleMessage send(String url, Object payload, Map<String, Serializable> messageProperties, long timeout)
         throws MuleException
     {
-        return send(url, new DefaultMuleMessage(payload, messageProperties, muleContext), timeout);
+        return send(url, new DefaultMuleMessage(payload, emptyMap(), messageProperties, emptyMap(), muleContext/*
+                                                                                                                * ,
```
dfeist: ?

PR #2873, opened 2016-06-28
MULE-10000: Add the concept of minimum Mule Version and Extension Cat…

estebanwasinger:
…egory


modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/AnnotationsBasedDescriberTestCase.java
```diff
@@ -413,6 +418,38 @@ public void flyweight()
         assertThat(appleOperation, is(sameInstance(bananaOperation)));
     }
 
+    @Test
+    public void minMuleVersionIsDescribedCorrectly(){
```
marianogonzalez: format

PR #2876, opened 2016-06-29
MULE-9856 Create MuleMessage builder. Fix conflict.

dfeist:


modules/xml/src/main/java/org/mule/runtime/module/xml/transformer/datatype/CollectionDataTypeXStreamConverter.java
```diff
@@ -6,19 +6,19 @@
  */
 package org.mule.runtime.module.xml.transformer.datatype;
 
+import org.mule.runtime.api.metadata.CollectionDataType;
```
elrodro83: Use DefaultCollectionDataType (consistent with SimpleDataTypeXStreamConverter)

modules/xml/src/main/java/org/mule/runtime/module/xml/transformer/datatype/CollectionDataTypeXStreamConverter.java
```diff
@@ -31,7 +32,7 @@
     @Override
     public boolean canConvert(Class type)
     {
-        return CollectionDataType.class.isAssignableFrom(type);
+        return DefaultCollectionDataType.class.isAssignableFrom(type);
```
elrodro83: Make this type change in the rest of this class

PR #2877, opened 2016-06-29
MULE-9988: ConnectionProvider.connect() should not depend on the config

marianogonzalez:


extensions/ftp/src/main/java/org/mule/extension/ftp/internal/ftp/connection/ClassicFtpConnectionProvider.java
```diff
@@ -120,21 +119,20 @@ protected FTPClient createClient()
...
+    public ConnectionHandlingStrategy<ClassicFtpFileSystem> getHandlingStrategy(ConnectionHandlingStrategyFactory<ClassicFtpFileSystem> handlingStrategyFactory)
     {
-        return handlingStrategyFactory.supportsPooling(new PoolingListener<FtpConnector, ClassicFtpFileSystem>()
+        return handlingStrategyFactory.supportsPooling(new PoolingListener<ClassicFtpFileSystem>()
         {
             @Override
-            public void onBorrow(FtpConnector config, ClassicFtpFileSystem connection)
+            public void onBorrow(ClassicFtpFileSystem connection)
             {
-                connection.changeToBaseDir();
```
afelisatti: Why is this no longer needed? Bug fix?

marianogonzalez: this can no longer be done there because you need the baseDir to do that, and base dir is a config attribute. It's no done on the operations

PR #2881, opened 2016-06-29
Mule-9964

afelisatti:


extensions/http/src/main/java/org/mule/extension/http/api/listener/HttpListenerProvider.java
```diff
@@ -106,6 +118,10 @@
     @Inject
     private MuleContext muleContext;
 
+    private ThreadingProfile workerThreadingProfile;
```
marianogonzalez: shouldn't this be a `@Parameter`

extensions/http/src/main/java/org/mule/extension/http/api/listener/HttpListenerProvider.java
```diff
@@ -136,37 +152,68 @@ public void initialise() throws InitialisationException
...
         try
         {
+            workManager = createWorkManager(configName);
+            workManager.start();
             server.start();
         }
         catch (IOException e)
         {
-            throw new ConnectionException("Could not start server", e);
+            throw new DefaultMuleException("Could not start HTTP server", e);
```
marianogonzalez: if this is not a connection exception then there won't be reconnection. Is that ok? shouldn't the source support retrying to establish the listener, if say, network is down temporarily?

extensions/http/src/main/java/org/mule/extension/http/api/listener/HttpListenerProvider.java
```diff
@@ -136,37 +152,68 @@ public void initialise() throws InitialisationException
...
+    {
+        try
+        {
+            workManager.dispose();
+        }
+        finally
+        {
+            workManager = null;
         }
+        server.stop();
```
marianogonzalez: is this atomic? no need to wait for tasks to finish?

extensions/http/src/main/java/org/mule/extension/http/api/listener/HttpListenerProvider.java
afelisatti: It should in a way for compatibility, but since the whole threading thing is going to be reviewed I decided to avoid it for now.

extensions/http/src/main/java/org/mule/extension/http/api/listener/HttpListenerProvider.java
```diff
@@ -136,37 +152,68 @@ public void initialise() throws InitialisationException
...
+    public Server connect() throws ConnectionException
+    {
         return server;
     }
 
     @Override
     public void disconnect(Server server)
     {
-        server.stop();
+        //server could be shared with others, do nothing.
```
marianogonzalez: how would it be shared with others if the handling strategy is none()?

extensions/http/src/main/java/org/mule/extension/http/api/listener/HttpListenerProvider.java
```diff
@@ -199,4 +246,18 @@ private void verifyConnectionsParameters() throws InitialisationException
         }
     }
 
+    private WorkManager createWorkManager(String name)
+    {
+        final WorkManager workManager = workerThreadingProfile.createWorkManager(format("%s%s.%s", ThreadNameHelper.getPrefix(muleContext), name, "worker"), muleContext.getConfiguration().getShutdownTimeout());
```
marianogonzalez: static import

extensions/http/src/main/java/org/mule/extension/http/api/listener/HttpListenerProvider.java
```diff
@@ -199,4 +246,18 @@ private void verifyConnectionsParameters() throws InitialisationException
...
+        if (workManager instanceof MuleContextAware)
```
marianogonzalez: use initiliaseIfNeeded

extensions/http/src/main/java/org/mule/extension/http/api/listener/HttpListenerProvider.java
afelisatti: yes

extensions/http/src/main/java/org/mule/extension/http/api/listener/HttpListenerProvider.java
afelisatti: I meant listeners sharing the same config.

extensions/http/src/main/java/org/mule/extension/http/api/listener/HttpListenerProvider.java
afelisatti: I'm not initialising it, only setting the muleContext

extensions/http/src/main/java/org/mule/extension/http/api/listener/HttpListenerProvider.java
marianogonzalez: shouldn't you then add the same TODO you added below and explain that depending on what we do this might be a parameter?

extensions/http/src/main/java/org/mule/extension/http/api/listener/HttpListenerProvider.java
marianogonzalez: initialise does that too

extensions/http/src/main/java/org/mule/extension/http/api/listener/HttpListenerProvider.java
marianogonzalez: shouldn't the server be stopped before the workManager so that you avoid situations in which you still accept connections but there're no workers to handle them? Also I think that one should be stopped on the other's finally. As it is now, if `workManager.dispose()` were to fail the server would leak (I know that dispose() doesn't throw exceptions but a RTE can happen unexpectedly)

PR #2889, opened 2016-06-30
MULE-9981: Migrate every possible functional test to use the test runner

pablokraan:
_ Moving tests out of org.mule.runtime system package


core/src/main/java/org/mule/runtime/core/MuleServer.java
```diff
@@ -61,7 +61,7 @@
     /**
      * Don't use a class object so the core doesn't depend on mule-module-spring-config.
      */
-    protected static final String CLASSNAME_DEFAULT_CONFIG_BUILDER = "org.mule.runtime.core.config.builders.AutoConfigurationBuilder";
+    public static final String CLASSNAME_DEFAULT_CONFIG_BUILDER = "org.mule.runtime.core.config.builders.AutoConfigurationBuilder";
```
gsfernandes: Perhaps we should add a TODO in order to revert this and find a better way to write the test that now it is in another package.

core/src/main/java/org/mule/runtime/core/MuleServer.java
```diff
@@ -115,7 +115,7 @@
      * handle to the local MuleContext can be obtained from anywhere by calling
      * MuleServer.getMuleContext()
      */
-    protected static MuleContext muleContext = null;
+    protected MuleContext muleContext = null;
```
gsfernandes: Same here

tests/integration/src/test/resources/META-INF/mule-module.properties
```diff
@@ -1,11 +1,19 @@
...
+                              org.mule.test.config.spring.handlers,\
+                              org.mule.test.config.spring.parsers,\
+                              org.mule.test.config.spring.parsers.beans,\
+                              org.mule.test.config.spring.parsers.specific,\
+                              org.mule.test.core.context,\
+                              org.mule.test.core.context.notification,\
+                              org.mule.test.core.context.notification.processors,\
+                              org.mule.test.core.routing.outbound,\
+                              org.mule.test.core.transformers.simple,\
+                              org.mule.test.xml.functional,\
```
gsfernandes: This is also a fix related to missing export packages, right?

PR #2893, opened 2016-06-30
MULE-9960: POJO @Parameters default values are not being set

pablomusumeci:


extensions/email/src/main/java/org/mule/extension/email/internal/util/EmailConnectorUtils.java
```diff
@@ -48,6 +48,16 @@
     public static final String TEXT = "text/*";
 
     /**
+     * Default email content encoding
+     */
+    public static final String UTF_8 = "UTF-8";
```
marianogonzalez: are these really necessary? Why don't you just use the MediaType constants?

pablomusumeci:
I needed the UTF-8 string for the default value of the annotation (couldn't use String.valueOf(StandardCharsets.UTF_8)). 

Ended up using  org.apache.commons.lang3.CharEncoding.UTF_8

http://stackoverflow.com/questions/6698354/where-to-get-utf-8-string-literal-in-java

extensions/email/src/main/java/org/mule/extension/email/internal/util/EmailConnectorUtils.java
pablomusumeci: MediaType constants are not Strings, they are MediaType and cannot be used as a defaultValue in  the annotation. The alternative I can think of is having a constant in EmailConnectorUtils converting that MediaType.TEXT.toString() and using that as value in the annotation.

extensions/email/src/main/java/org/mule/extension/email/api/sender/SMTPConfiguration.java
```diff
@@ -32,10 +41,32 @@
     private String from;
 
     /**
+     * Default encoding to be used if there is encoding specified in {@link EmailContent#charset}
```
marianogonzalez: jdoc is wrong. This is used if NO enconding is specified. Also rephrase this to not reference one particular entity. Make it general since it more that surely will end up applying to some other stuff

extensions/email/src/test/java/org/mule/extension/email/sender/SMTPTestCase.java
```diff
@@ -174,10 +178,24 @@ public void forwardEmailWithContent() throws Exception
     @Test
     public void sendEncodedMessage() throws Exception
     {
-        flowRunner(SEND_ENCODED_MESSAGE).withPayload(WEIRD_CHAR_MESSAGE).run();
+        final String defaultEncoding = muleContext.getConfiguration().getDefaultEncoding();
+        Assert.assertThat(defaultEncoding, CoreMatchers.is(notNullValue()));
```
marianogonzalez: static imports

PR #2895, opened 2016-06-30
MULE-9836: Wrong XSD is generated for List<Object>

juandesi:


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/schema/builder/SchemaBuilder.java
```diff
@@ -853,8 +854,11 @@ public void visitArrayType(ArrayType arrayType)
             {
                 MetadataType genericType = arrayType.getType();
                 final boolean supportsChildElement = shouldGenerateDataTypeChildElements(genericType, expressionSupport);
-                forceOptional = supportsChildElement || shouldForceOptional(genericType);
 
+                Optional<TypeIdAnnotation> genericTypeId = getSingleAnnotation(genericType, TypeIdAnnotation.class);
```
marianogonzalez: Use JavaTypeUtils to get the actual class instead

PR #2903, opened 2016-07-01
MULE-10024: OperationContext should distinguish between null-value

elrodro83:
parameters and not passed parameters

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ParameterGroupArgumentResolver.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ParameterGroupArgumentResolver.java
@@ -80,10 +80,6 @@ public final class ParameterGroupArgumentResolver<T> implements ArgumentResolver
                 {
                     parameterField.set(group, operationContext.getParameter(getAlias(parameterField)));
                 }
-                else
-                {
-                    parameterField.set(group, null);
-                }
             }
 
             for (Map.Entry<Field, ParameterGroupArgumentResolver<?>> childGroup : childGroups.entrySet())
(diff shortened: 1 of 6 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ParameterGroupArgumentResolver.java
```diff
@@ -76,10 +76,9 @@ public T resolve(OperationContext operationContext)
             T group = type.newInstance();
             for (Field parameterField : parameterFields)
             {
-                Object value = operationContext.getParameter(getAlias(parameterField));
-                if (value != null)
+                if (operationContext.hasParameter(getAlias(parameterField)))
                 {
-                    parameterField.set(group, value);
+                    parameterField.set(group, operationContext.getParameter(getAlias(parameterField)));
```
marianogonzalez: this doesn't seem like the exact same behaviour. If the parameter was not applied, then nothing is set. That means that a field with a default value will no longer be assigned to `null`

elrodro83: added an else to keep that behavior. However, there seems to be a test case missing to fail on that case...

marianogonzalez: the code still looks weird. I took a second look an realised I was wrong. Your original commit was indeed equivalent to the original behaviour, so this second version should be rolled back. That been said, if your original commit is equivalent, then why to change it at all?

elrodro83: I had to make that original change because there were some tests failures. Some extensions code is trying to access the parameters that have not been set.

PR #2904, opened 2016-07-01
MULE-9856 Create MuleMessage builder.

elrodro83:
* fix tests


tests/http/src/test/java/org/mule/runtime/module/http/functional/listener/HttpListenerMethodRoutingTestCase.java
```diff
@@ -65,7 +66,7 @@ public void callWithMethod() throws Exception
     @Test
     public void callWithMethodEmptyBody() throws Exception
     {
-        sendRequestAndAssertMethod(null);
+        sendRequestAndAssertMethod("");
```
dfeist: NullPayload?

elrodro83: The called method receives String, and sending an empty one is okay for this test.

PR #2912, opened 2016-07-04
MULE-9856: Create MuleMessage builder - Core module

elrodro83:


core-tests/src/test/java/org/mule/runtime/core/DefaultMuleMessageTestCase.java
```diff
@@ -319,19 +299,21 @@ public void testNewMuleMessageFromMuleMessageWithAttachment() throws Exception
...
-    private MutableMuleMessage createMuleMessage()
+    private MuleMessage createMuleMessage()
     {
+        return MuleMessage.builder().payload(TEST_PAYLOAD).attributes(testAttributes).addOutboundProperty("MuleMessage", "MuleMessage").build();
         // TODO MULE-9856 Replace with the builder
-        MutableMuleMessage previousMessage = new DefaultMuleMessage(AbstractMuleContextTestCase.TEST_PAYLOAD, DataType.builder(DataType.STRING).charset(getDefaultEncoding(muleContext)).build(),
-                testAttributes);
-        previousMessage.setOutboundProperty("MuleMessage", "MuleMessage");
-        return previousMessage;
+        // MutableMuleMessage previousMessage = new DefaultMuleMessage(AbstractMuleContextTestCase.TEST_PAYLOAD,
```
dfeist: ?

core/src/main/java/org/mule/runtime/core/DefaultMuleMessageBuilder.java
```diff
@@ -44,13 +52,13 @@
...
-    private int correlationSequence;
-    private int correlationGroupSize;
+    private int correlationSequence = -1;
+    private int correlationGroupSize = -1;
     private Object replyTo;
     private ExceptionPayload exceptionPayload;
 
-    private Map<String, TypedValue<Serializable>> inboundProperties = new HashMap<>();
-    private Map<String, TypedValue<Serializable>> outboundProperties = new HashMap<>();
+    private Map<String, TypedValue<Serializable>> inboundProperties = new CaseInsensitiveMapWrapper<>(HashMap.class);
```
dfeist: We use CaseInsensitiveHashMap in DefaultMuleMessage.  We either need to continue to us this, or deprecate/remove it if the use of CaseInsensitiveMapWrapper is preferred.

core/src/main/java/org/mule/runtime/core/DefaultMuleMessage.java
```diff
@@ -473,7 +474,14 @@ public int getCorrelationSequence()
     public void setCorrelationSequence(int sequence)
     {
         assertAccess(WRITE);
-        setOutboundProperty(MULE_CORRELATION_SEQUENCE_PROPERTY, sequence);
+        if (sequence >= 0)
```
dfeist: Wouldn't it be better to use nullable Integer or better Optional here? Also this assumes 0 is not a valid sequence number doesn't it?

core/src/main/java/org/mule/runtime/core/DefaultMuleMessage.java
```diff
@@ -495,7 +503,14 @@ public int getCorrelationGroupSize()
     public void setCorrelationGroupSize(int size)
     {
         assertAccess(WRITE);
-        setOutboundProperty(MULE_CORRELATION_GROUP_SIZE_PROPERTY, size);
+        if (size >= 0)
```
dfeist: Ditto

core/src/main/java/org/mule/runtime/core/DefaultMuleMessageBuilder.java
```diff
@@ -215,24 +249,60 @@ public DefaultMuleMessageBuilder(org.mule.runtime.api.message.MuleMessage<PAYLOA
...
+
+    @Override
+    public Builder<PAYLOAD, ATTRIBUTES> removeOutboundAttachment(String key)
+    {
+        outboundAttachments.remove(key);
+        return this;
+    }
+
+    @Override
+    public Builder<PAYLOAD, ATTRIBUTES> clearOutbloundAttachments()
```
dfeist: Typo.

core/src/main/java/org/mule/runtime/core/api/MuleMessage.java
```diff
@@ -227,17 +226,49 @@
...
+         * @return
+         */
+        Builder<PAYLOAD, ATTRIBUTES> removeOutboundAttachment(String key);
+
+        /**
+         * Removes all outbound attachments on this message builder.
+         * 
+         * @return
+         */
+        Builder<PAYLOAD, ATTRIBUTES> clearOutbloundAttachments();
```
dfeist: Typo.

PR #2913, opened 2016-07-04
MULE-10025: Minimize the amount of classes that extensions need to ex…

marianogonzalez:
…pose - Part I


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/describer/DefaultConnectionProviderFactory.java
```diff
@@ -7,35 +7,40 @@
...
+
 /**
  * Creates instances of {@link ConnectionProvider} based on a {@link #providerClass}
  * * @param <Config>     the generic type for the configuration objects that the created {@link ConnectionProvider providers} accept
  *
  * @param <Connection> the generic type for the connections that the created  {@link ConnectionProvider providers} produce
  * @since 4.0
  */
-final class DefaultConnectionProviderFactory<Config, Connection> implements ConnectionProviderFactory<Connection>
+final class DefaultConnectionProviderFactory<Connection> implements ConnectionProviderFactory<Connection>
```
elrodro83: remove <Config> from the javadoc.

PR #2915, opened 2016-07-04
MULE-7780 - JUnit rules and matchers for Mule Standalone server        Adding a Rule for deploying Mule applications

asequeira:


tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/DistroUnzipper.java
```diff
@@ -0,0 +1,69 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.util.Enumeration;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+
+import org.apache.commons.io.FileUtils;
+
+
+public class DistroUnzipper
```
aiannucci: Is there a reason to keep this class non-static?

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleDeployment.java
```diff
@@ -0,0 +1,207 @@
...
+ *
+ *  &#064;Test
+ *  public void useApplication() throws IOException {
+ *      // App is deployed
+ *      // This code exercises the application
+ *  }
+ * }
+ * </pre>
+ */
+public class MuleDeployment implements TestRule
```
pablolagreca: Would it make sense to just have one rule for deploying in a container any set of apps + domains + whatever?. It may be out of scope for this PR.

asequeira: Adding domain support is part of the plan, but this is an initial version that we need to test classloading.

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/DistroUnzipper.java
```diff
@@ -0,0 +1,69 @@
...
+            {
+                ZipEntry entry = zipFileEntries.nextElement();
+                File destFile = new File(entry.getName());
+                if (entry.isDirectory())
+                {
+                    destFile.mkdir();
+                }
+                else
+                {
+                    FileUtils.copyInputStreamToFile(zip.getInputStream(entry), destFile);
```
elrodro83: static import

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleDeployment.java
```diff
@@ -0,0 +1,207 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * JUnit rule to deploy a Mule application for testing in a local Mule server. Usage:
+ * <p>
+ * <pre>
+ * public static class MuleDeployment {
+ *  &#064;Rule
+ *  public static MuleDeployment deployment = application(&quot;/path/to/application.zip&quot;).withProperty("-M-Dproperty", "value").timeout(120).deploy();
```
aiannucci: if you make it static, it should be a ClassRule, if not, it shouldn't be static

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleDeployment.java
```diff
@@ -0,0 +1,207 @@
...
+    private MuleDeployment()
+    {
+        String zippedDistribution = System.getProperty(DISTRIBUTION_PROPERTY);
+        if (StringUtils.isEmpty(zippedDistribution))
+        {
+            logger.error("You must configure the location for Mule distribution in the system property: " + DISTRIBUTION_PROPERTY);
+        }
+        File distribution = new File(zippedDistribution);
+        zipExtractor = new DistroUnzipper(distribution, WORKING_DIRECTORY);
+        if (!distribution.exists())
```
aiannucci: Shouldn't you validate this before passing it to the DistroUnzipper?

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleDeployment.java
```diff
@@ -0,0 +1,207 @@
...
+public class MuleDeployment implements TestRule
+{
+
+    private static final String DISTRIBUTION_PROPERTY = "com.mulesoft.muleesb.distributions:mule-ee-distribution-standalone:zip";
+    private static final long POLL_DELAY_MILLIS = 1000;
+    private static Logger logger = LoggerFactory.getLogger(MuleDeployment.class);
+    private static final File WORKING_DIRECTORY = new File(getProperty("user.dir"));
+    private static final String DELETE_ON_EXIT = getProperty("mule.test.deleteOnExit");
+    private static final int DEFAULT_DEPLOYMENT_TIMEOUT = 60000;
+    private int deploymentTimeout = getProperty("mule.test.deployment.timeout") == null ? DEFAULT_DEPLOYMENT_TIMEOUT : parseInt(getProperty("mule.test.deployment.timeout"));
```
anosenzo: You can use the getProperty() with the second attribute as the default value and avoid calling the getProperty() twice. In this case it will be parseInt( getProperty("mule.test.deployment.timeout", DEFAULT_DEPLOYMENT_TIMEOUT) )

aiannucci:
Create new JIRA targetting 4.0 for this

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleDeployment.java
elrodro83: extract this string to a constant

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleInstallation.java
```diff
@@ -74,19 +71,19 @@ public Statement apply(final Statement base, final Description description)
...
-        unzip(distribution, WORKING_DIRECTORY);
+        muleHome = new DistroUnzipper(distribution, WORKING_DIRECTORY).unzip().muleHome();
     }
 
     @Override
     protected void after()
     {
-        File logs = new File(muleHome, "logs");
-        File dest = new File(testname + ".logs");
+        File dest = new File(new File("logs"), testname);
```
elrodro83: This method is a partial duplicate of MuleDeployment#after()

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleDeployment.java
asequeira: What is your suggestion? I consider that a user should know how to use JUnit rules before using it. Besides the most usual use case would be to use a class rule because you will run several tests against the same application.

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleDeployment.java
asequeira: which one? mule.test.deployment.timeout? does it make any sense to define a constant when declaring a variable?

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleDeployment.java
aiannucci: So this Javadoc should say @ClassRule, otherwise it will fail if used as it is

PR #2924, opened 2016-07-05
MULE-10042: Operations should no longer be allowed to return a MuleMe…

marianogonzalez:
…ssage instance


extensions/ftp/src/main/java/org/mule/extension/ftp/internal/ftp/command/FtpReadCommand.java
```diff
@@ -80,7 +81,7 @@ public FtpReadCommand(ClassicFtpFileSystem fileSystem, FTPClient client)
         {
             InputStream payload = ClassicFtpInputStream.newInstance((FtpConnector) config, attributes, pathLock);
             MediaType mediaType = fileSystem.getFileMessageMediaType(message.getDataType().getMediaType(), attributes);
-            return MuleMessage.builder().payload(payload).mediaType(mediaType).attributes((FileAttributes) attributes).build();
+            return OperationResult.<InputStream, FileAttributes>builder().output(payload).mediaType(mediaType).attributes((FileAttributes) attributes).build();
```
pablolagreca: If we use an static method, how will the user access the default encoding?

marianogonzalez: not follwoing

PR #2926, opened 2016-07-06
MULE-9855 Make MuleMessage immutable. Step 1

dfeist:
PENDING BUILD.


core-tests/src/test/java/org/mule/runtime/core/AbstractThreadSafeAccessTestCase.java
```diff
@@ -1,86 +0,0 @@
...
- */
-package org.mule.runtime.core;
-
-import static org.junit.Assert.assertEquals;
-
-import org.mule.runtime.core.api.MuleMessage;
-import org.mule.runtime.core.api.ThreadSafeAccess;
-import org.mule.tck.junit4.AbstractMuleContextTestCase;
-
-public abstract class AbstractThreadSafeAccessTestCase extends AbstractMuleContextTestCase
```
elrodro83: Would it make sense to keep testing this for events?

core-tests/src/test/java/org/mule/runtime/core/mule/enricher/MessageEnricherTestCase.java
```diff
@@ -344,7 +343,7 @@ public void enricherConservesSameEventInstanceNonBlockingTargetNonBlocking() thr
 
         nullReplyToHandler.latch.await(RECEIVE_TIMEOUT, TimeUnit.MILLISECONDS);
 
-        assertThat(sensingNullMessageProcessor.event.getMessage(), not(sameInstance(in.getMessage())));
+        assertThat(sensingNullMessageProcessor.event.getMessage(), sameInstance(in.getMessage()));
```
elrodro83: is this assertion still meaningful?

core/src/main/java/org/mule/runtime/core/el/context/OutboundAttachmentMapContext.java
```diff
@@ -49,7 +50,7 @@ public void doRemove(String key)
     @Override
     public void clear()
     {
-        event.setMessage(MuleMessage.builder(event.getMessage()).clearOutboundAttachments().build());
+        event.setMessage(MuleMessage.builder(event.getMessage()).outboundAttachments(emptyMap()).build());
```
elrodro83: did we remove the clear... method from the builder?

transports/vm/src/main/java/org/mule/compatibility/transport/vm/VMMessageReceiver.java
```diff
@@ -115,10 +113,6 @@ public MuleEvent process() throws Exception
                     if (event != null && !VoidMuleEvent.getInstance().equals(event) && getEndpoint().getExchangePattern().hasResponse())
                     {
                         MuleMessage returnedMessage = event.getMessage();
```
elrodro83: remove this line too

elrodro83:
Is there any way to also check there are no synchronization on a message object?

core-tests/src/test/java/org/mule/runtime/core/AbstractThreadSafeAccessTestCase.java
dfeist: Event doesn't use thread safe access.

core-tests/src/test/java/org/mule/runtime/core/mule/enricher/MessageEnricherTestCase.java
dfeist: Less meaningful, but no harm asserting that message isn't copied.  Can remove if you prefer..

core/src/main/java/org/mule/runtime/core/el/context/OutboundAttachmentMapContext.java
dfeist: Yes.  Do you think we should put it back?

core-tests/src/test/java/org/mule/runtime/core/AbstractThreadSafeAccessTestCase.java
elrodro83: It was delegating to its message

PR #2936, opened 2016-07-07
MULE-9855 Make MuleMessage immutable. Step 2

dfeist:


core/src/main/java/org/mule/runtime/core/MessagePropertiesContext.java
```diff
@@ -1,250 +0,0 @@
...
-/**
- * This object maintains case-sensitive inbound and outbound scoped messages properties.
- * <ol>
- * <li> {@link PropertyScope#INBOUND} Contains properties that were on the message when
- * it was received by Mule. This scope is read-only.</li>
- * <li>{@link PropertyScope#OUTBOUND} Any properties set in this scope will be attached
- * to any outbound messages resulting from this message. This is the default scope.</li>
- * </ol>
- */
-public class MessagePropertiesContext implements MessageProperties, Serializable
```
elrodro83: can you briefly explain why this is no longer needed?

dfeist: It's moved inside MuleMessageImplementation. We need to clean it up, but for now it's moved private/inner.

PR #2937, opened 2016-07-07
MULE-10043: Content-Type header is not used if it is present as a Hea…

afelisatti:
…der because it don't compare its key as case insensitive

### DIFF
--- a/modules/http/src/main/java/org/mule/runtime/module/http/internal/CaseInsensitiveParameterMap.java
+++ b/modules/http/src/main/java/org/mule/runtime/module/http/internal/CaseInsensitiveParameterMap.java
@@ -13,6 +13,11 @@ import org.mule.runtime.core.util.CaseInsensitiveMapWrapper;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
 
+/**
+ * {@link ParameterMap} where the key's case is not taken into account when looking for it, adding or aggregating it.
+ *
+ * @since 4.0
+ */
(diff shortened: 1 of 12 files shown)


modules/http/src/main/java/org/mule/runtime/module/http/internal/CaseInsensitiveParameterMap.java
```diff
@@ -0,0 +1,29 @@
...
+
+package org.mule.runtime.module.http.internal;
+
+import static java.util.stream.Collectors.toCollection;
+import org.mule.runtime.core.util.CaseInsensitiveMapWrapper;
+
+import java.util.LinkedHashMap;
+import java.util.LinkedList;
+
+public class CaseInsensitiveParameterMap extends ParameterMap
```
elrodro83: javadoc

modules/http/src/main/java/org/mule/runtime/module/http/internal/CaseInsensitiveParameterMap.java
```diff
@@ -0,0 +1,29 @@
...
+{
+
+    public CaseInsensitiveParameterMap(ParameterMap paramsMap)
+    {
+        this.paramsMap = new CaseInsensitiveMapWrapper<>(LinkedHashMap.class);
+        for (String key : paramsMap.keySet())
+        {
+            LinkedList<String> values = paramsMap.getAll(key).stream().collect(toCollection(LinkedList::new));
```
elrodro83: declare as list instead of linked list

PR #2938, opened 2016-07-07
MULE-9856 Introduce Attributes instead of Serializable in MuleMessage…

dfeist:
… and builder

### DIFF
--- a/core-tests/src/test/java/org/mule/runtime/core/DefaultMuleMessageTestCase.java
+++ b/core-tests/src/test/java/org/mule/runtime/core/DefaultMuleMessageTestCase.java
@@ -15,6 +15,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
+import org.mule.runtime.api.message.Attributes;
 import org.mule.runtime.api.message.NullPayload;
 import org.mule.runtime.api.metadata.MediaType;
 import org.mule.runtime.core.api.MuleMessage;
@@ -39,7 +40,7 @@ public class DefaultMuleMessageTestCase extends AbstractMuleContextTestCase
 {
 
     public static final String FOO_PROPERTY = "foo";
-    private Serializable testAttributes = new Apple();
+    private Attributes testAttributes = new Attributes() {};
 
     @Test
     public void testMessagePropertiesAccessors()
--- a/core-tests/src/test/java/org/mule/runtime/core/DefaultMuleMessageTestCase.java
+++ b/core-tests/src/test/java/org/mule/runtime/core/DefaultMuleMessageTestCase.java
@@ -13,12 +13,12 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
+import static org.mule.runtime.core.message.NullAttributes.NULL_ATTRIBUTES;
 
 import org.mule.runtime.api.message.Attributes;
 import org.mule.runtime.api.message.NullPayload;
 import org.mule.runtime.api.metadata.MediaType;
 import org.mule.runtime.core.api.MuleMessage;
-import org.mule.runtime.core.message.NullAttributes;
 import org.mule.runtime.core.util.IOUtils;
 import org.mule.tck.junit4.AbstractMuleContextTestCase;
 import org.mule.tck.testmodels.fruit.Apple;
@@ -43,7 +43,7 @@ public class DefaultMuleMessageTestCase extends AbstractMuleContextTestCase
 {
 
     public static final String FOO_PROPERTY = "foo";
-    private Attributes testAttributes = new NullAttributes();
+    private Attributes testAttributes = NULL_ATTRIBUTES;
 
     @Test
     public void testMessagePropertiesAccessors()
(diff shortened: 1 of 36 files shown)


core-tests/src/test/java/org/mule/runtime/core/DefaultMuleMessageTestCase.java
```diff
@@ -39,7 +40,7 @@
 {
 
     public static final String FOO_PROPERTY = "foo";
-    private Serializable testAttributes = new Apple();
+    private Attributes testAttributes = new Attributes() {};
```
marianogonzalez: this will not be very supportable/debugable. When you see the toString(), or when you serialize, you will see an ugly name. It will also be hard to put a condition breakpoint... I would either make this an instantiable class or create a NullAttributes kind of thing

core/src/main/java/org/mule/runtime/core/DefaultAttributes.java
```diff
@@ -0,0 +1,17 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core;
+
+import org.mule.runtime.api.message.Attributes;
+
+/**
+ * Default implementation of {@link Attributes} to be used when no other connector specific attributes instance is set.
+ */
+public class DefaultAttributes implements Attributes
```
marianogonzalez: I don't see the value of this class. Besides if this exists, then why do you use `new Attributes{}` above?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/AbstractReturnDelegate.java
```diff
@@ -61,9 +61,9 @@ protected MuleMessage toMessage(Object value, OperationContextAdapter operationC
         }
     }
 
-    private Serializable resolveAttributes(OperationResult operationResult, OperationContextAdapter operationContext)
+    private Attributes resolveAttributes(OperationResult operationResult, OperationContextAdapter operationContext)
     {
-        return (Serializable) operationResult.getAttributes().orElseGet(() -> operationContext.getEvent().getMessage().getAttributes());
+        return (Attributes) operationResult.getAttributes().orElseGet(() -> operationContext.getEvent().getMessage().getAttributes());
```
elrodro83: is this cast needed?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultOperationResultBuilder.java
```diff
@@ -8,29 +8,29 @@
...
 
-import java.io.Serializable;
 import java.util.Optional;
 
 /**
  * Default implementation of {@link OperationResult.Builder}
  *
  * @param <Output>     the generic type of the output value
- * @param <Attributes> the generic type of the message attributes
+ * @param <A> the generic type of the message attributes
```
elrodro83: identation

core/src/main/java/org/mule/runtime/core/DefaultAttributes.java
elrodro83: change to NullAttributes.

core-tests/src/test/java/org/mule/runtime/core/DefaultMuleMessageTestCase.java
dfeist: This is just a test case, where i want testAttributes to be a unique object instance to then assert with.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/AbstractReturnDelegate.java
dfeist: yes

core/src/main/java/org/mule/runtime/core/message/NullAttributes.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private static final long serialVersionUID = 1201393762712713465L;
+
+    private Object readResolve() throws ObjectStreamException
+    {
+        return NULL_ATTRIBUTES;
+    }
+
+    @Override
+    public int hashCode()
```
marianogonzalez: remove

core/src/main/java/org/mule/runtime/core/message/BaseAttributes.java
```diff
@@ -0,0 +1,26 @@
...
+
+import static org.apache.commons.lang3.builder.ToStringStyle.SHORT_PREFIX_STYLE;
+
+import org.mule.runtime.api.message.Attributes;
+
+import org.apache.commons.lang3.builder.ReflectionToStringBuilder;
+
+/**
+ * Basic attributes implementation that defines a generic {@link #toString()} method.
+ */
```
marianogonzalez: `@since 4.0`

core/src/main/java/org/mule/runtime/core/message/NullAttributes.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package org.mule.runtime.core.message;
+
+import org.mule.runtime.api.message.Attributes;
+
+import java.io.ObjectStreamException;
+
+/**
+ * Default implementation of {@link Attributes} to be used when no other connector specific attributes instance is set.
+ */
```
marianogonzalez: since

PR #2945, opened 2016-07-08
MULE-10058 - Create a simple builder for creating a MuleContext

pablolagreca:


core-tests/src/test/java/org/mule/runtime/core/config/bootstrap/SimpleRegistryBootstrapTestCase.java
```diff
@@ -63,7 +64,7 @@ public void registerOnlyDomainPropertiesType() throws Exception
     private SimpleRegistryBootstrap createTestRegistryBootstrap(ArtifactType artifactType) throws InitialisationException
     {
         final Properties properties = new Properties();
-        properties.put("1", String.format("java.lang.String,%s=%s", ArtifactType.APPLY_TO_ARTIFACT_TYPE_PARAMETER_KEY, ArtifactType.APP.getAsString()));
+        properties.put("1", String.format("java.lang.String,%s=%s", ArtifactType.APPLY_TO_ARTIFACT_TYPE_PARAMETER_KEY, APP.getAsString()));
```
marianogonzalez: static imports, class wide

core/src/main/java/org/mule/runtime/core/registry/MuleRegistryHelper.java
```diff
@@ -181,6 +182,24 @@ public Transformer lookupTransformer(DataType source, DataType result) throws Tr
     protected Transformer resolveTransformer(DataType source, DataType result) throws TransformerException
     {
         Lock readLock = transformerResolversLock.readLock();
+        //if (transformerResolvers.isEmpty())
```
marianogonzalez: remove. Ugly bug you found though. I wonder how many WTF moments we had because of this

modules/launcher/src/main/java/org/mule/runtime/module/launcher/application/DefaultMuleApplication.java
```diff
@@ -172,24 +174,25 @@ public void init()
 
         try
         {
-            ConfigurationBuilder cfgBuilder = domainRepository.getDomain(descriptor.getDomain()).createApplicationConfigurationBuilder(this);
-            if (!cfgBuilder.isConfigured())
+            ArtifactMuleContextBuilder artifactBuilder = new ArtifactMuleContextBuilder()
+                    .setArtifactProperties(descriptor.getAppProperties())
+                    .setArtifactType(APP)
```
marianogonzalez: would it make sense to have a specific builder class for each artifact type? I understand that right now it might be quite overkill, but my concern is that in the future we want to add something that only applies for one of the possible types and then the API becomes ugly

core/src/main/java/org/mule/runtime/core/registry/MuleRegistryHelper.java
dfeist: Remove or add TODO referencing JIRA issue

modules/launcher/src/test/java/org/mule/runtime/module/launcher/DeploymentServiceTestCase.java
```diff
@@ -174,10 +177,10 @@
     @Rule
     public DynamicPort httpPort = new DynamicPort("httpPort");
 
-    @Override
-    protected void doSetUp() throws Exception
+    @Before
+    public void setUp() throws Exception
     {
-        super.doSetUp();
+        //super.doSetUp();
```
marianogonzalez: remove

modules/launcher/src/main/java/org/mule/runtime/module/launcher/artifact/ArtifactMuleContextBuilder.java
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    /**
+     * Sets the default encoding for the {@code MuleContext} if the use did not define one
+     * explicitly within the configuration.
+     *
+     * @param defaultEncoding default encoding to use within the {@code MuleContext}
+     * @return the builder
+     */
+    public ArtifactMuleContextBuilder setDefaultEncoding(String defaultEncoding)
```
dfeist: Does it not make sense to use Configuration object which currently holds encoding.  Or what is the relationship between this builder and MuleConfiguration?

modules/launcher/src/test/java/org/mule/runtime/module/launcher/DeploymentServiceTestCase.java
```diff
@@ -203,16 +206,15 @@ protected void doSetUp() throws Exception
         deploymentService.addDomainDeploymentListener(domainDeploymentListener);
     }
 
-    @Override
-    protected void doTearDown() throws Exception
+    @After
+    public void tearDown() throws Exception
     {
         // comment out the deletion to analyze results after test is done
```
marianogonzalez: this comment doesn't seem to be current

modules/launcher/src/test/java/org/mule/runtime/module/launcher/artifact/ArtifactMuleContextBuilderTestCase.java
```diff
@@ -0,0 +1,84 @@
...
+        muleContext.start();
+        assertThat(muleContext.isStarted(), is(true));
+    }
+
+    @Test
+    public void buildWithoutClassloader() throws Exception
+    {
+        expectedException.expectMessage(EXECUTION_CLASSLOADER_WAS_NOT_SET);
+        new ArtifactMuleContextBuilder()
+                .build();
```
marianogonzalez: inline

modules/launcher/src/test/java/org/mule/runtime/module/launcher/artifact/ArtifactMuleContextBuilderTestCase.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    @Test
+    public void setNullArtifactProperties() throws Exception
+    {
+        expectedException.expectMessage(MULE_CONTEXT_ARTIFACT_PROPERTIES_CANNOT_BE_NULL);
+        new ArtifactMuleContextBuilder()
+                .setArtifactProperties(null);
```
marianogonzalez: inline

modules/spring-config/src/main/java/org/mule/runtime/config/spring/Notification.java
```diff
@@ -0,0 +1,234 @@
...
+ * Notification helper object for creating notification manager configuration.
+ *
+ * For now we create specific types {@code Notification.EnabledNotification} and {@code Notification.DisabledNotification}
+ * for elements that have children of the same type but that must be injected in different collections of the
+ * parent object. Ideally this should not be needed or we don't have to support such use case anymore but for
+ * previous elements coming from previous mule version we have to keep the XML syntax as it is.
+ *
+ * @since 4.0
+ */
+public class Notification
```
marianogonzalez: this class name is not descriptive. The Jdoc is also confusing since it says that it's a helper for creating a manager configuration, yet it doesn't look like a factory or builder, it rather looks like an actual notification

modules/spring-config/src/main/java/org/mule/runtime/config/spring/Notification.java
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    private String event;
+    private Class eventClass;
+    private String interfase;
+    private Class interfaceClass;
+    private boolean eventExplicitlyConfigured;
+    private boolean interfaceExplicitlyConfigured;
+
+    public void setInterfase(String interfase)
```
marianogonzalez: interfase -> interface . Both, method and argument name

modules/spring-config/src/main/java/org/mule/runtime/config/spring/SpringMuleContextServiceConfigurator.java
```diff
@@ -0,0 +1,264 @@
...
+    private void createEndpointFactory(BeanDefinitionRegistry beanDefinitionRegistry)
+    {
+        try
+        {
+            Class endpointFactoryClass = ClassUtils.loadClass("org.mule.compatibility.core.endpoint.DefaultEndpointFactory", Thread.currentThread().getContextClassLoader());
+            beanDefinitionRegistry.registerBeanDefinition("_muleEndpointFactory", getBeanDefinition(endpointFactoryClass));
+        }
+        catch (ClassNotFoundException e)
+        {
+            //Nothing to do.
```
marianogonzalez: i would log as WARN, but willing to settle on DEBUG... I'm cheap like that

modules/spring-config/src/main/java/org/mule/runtime/config/spring/SpringMuleContextServiceConfigurator.java
```diff
@@ -0,0 +1,264 @@
...
+    private void createBootstrapBeanDefinitions(BeanDefinitionRegistry beanDefinitionRegistry, MuleContext muleContext, ArtifactType artifactType, OptionalObjectsController optionalObjectsController)
+    {
+        try
+        {
+            SpringRegistryBootstrap springRegistryBootstrap = new SpringRegistryBootstrap(artifactType, muleContext, optionalObjectsController, beanDefinitionRegistry);
+            springRegistryBootstrap.initialise();
+        }
+        catch (InitialisationException e)
+        {
+            throw new RuntimeException(e);
```
marianogonzalez: maybe it makes sense for this method to throw InitialisationException? Not sure myself

modules/spring-config/src/main/java/org/mule/runtime/config/spring/SpringMuleContextServiceConfigurator.java
```diff
@@ -0,0 +1,264 @@
...
+        }
+    }
+
+    private AbstractBeanDefinition getBeanDefinition(Class<?> beanType)
+    {
+        return getBeanDefinitionBuilder(beanType)
+                .getBeanDefinition();
```
marianogonzalez: inline

modules/spring-config/src/main/java/org/mule/runtime/config/spring/SpringXmlConfigurationBuilder.java
```diff
@@ -52,72 +42,40 @@
...
-        this(configResource, emptyMap());
+        this(configResource, emptyMap(), APP);
     }
 
     public SpringXmlConfigurationBuilder(String[] configFiles) throws ConfigurationException
     {
         super(configFiles, emptyMap());
-    }
-
-    protected List<ConfigResource> getConfigResources() throws IOException
```
marianogonzalez: Gateway requires overriding this method to add custom stuff of their own. Will they have a similar extension point for them to create custom contexts?

tests/integration/src/test/resources/org/mule/test/integration/notifications/message-processor-notification-test-flow.xml
```diff
@@ -14,15 +14,16 @@
...
     </notifications>
 
     <http:request-config name="requestConfig" host="localhost" />
 
     <spring:bean name="notificationLogger" class="org.mule.test.core.context.notification.processors.MessageProcessorNotificationLogger"/>
 
-    <notifications>
-        <notification-listener ref="notificationLogger"/>
-    </notifications>
+    <!--<notifications>-->
```
marianogonzalez: remove or uncomment

modules/launcher/src/main/java/org/mule/runtime/module/launcher/application/DefaultMuleApplication.java
pablolagreca: May make sense what you say in the future. That's not the case right now. The only difference is the set of services that are available in each case plus a couple of restrictions that may be removed in the future. This is part of a big refactoring of the whole spring context creation and mule context as is only internal API. If there's a good reason to refactor the builder we can do it in the future.

modules/launcher/src/test/java/org/mule/runtime/module/launcher/artifact/ArtifactMuleContextBuilderTestCase.java
pablolagreca: I will change this but there's no reason to do so. We don't have a policy for this.

modules/spring-config/src/main/java/org/mule/runtime/config/spring/SpringMuleContextServiceConfigurator.java
pablolagreca: Warn doesn't make sense, CNFE will be the most common case.

modules/spring-config/src/main/java/org/mule/runtime/config/spring/SpringMuleContextServiceConfigurator.java
pablolagreca: One we can get rid completely of SimpleRegistryBootstrap we will be availabe to remove the InitialisationException thrown by it because the initialise method won't be used anymore in the lifecycle of the MuleContext creation. So it doesn't make sense to throw InitialisationException and force the callers to handle it since we are going to remove it in the future.

modules/spring-config/src/main/java/org/mule/runtime/config/spring/SpringXmlConfigurationBuilder.java
pablolagreca: The whole mechanism they use won't be available anymore. Once they start integration we will have to figure out a  proper interface for the to plug-in behaviour.

modules/launcher/src/main/java/org/mule/runtime/module/launcher/artifact/ArtifactMuleContextBuilder.java
pablolagreca: For now the idea is that the builder does not know about specific object within the MuleContext like MuleConfiguration. For now the idea is just to create a builder with all the information we are getting from the artifact descriptors to actually create a MuleContext so we can do so decoupled from those artifacts configuration. It may be that we need to move more MuleConfiguration properties to the builder or at some point allow to just set the MuleConfiguration. I would wait until we keep using the builder, but I would like to have it decoupled from MuleConfiguration.

PR #2967, opened 2016-07-13
MULE-10078: Properly handle disposal of XaTransactedJmsMessageReceiver

elrodro83:


core/src/main/java/org/mule/transport/AbstractConnector.java
```diff
@@ -1783,7 +1784,7 @@ public String getConnectionDescription()
     }
 
     @Override
-    public final boolean isConnected()
+    public boolean isConnected()
```
pablolagreca: do you really need to remove the final modifier?

elrodro83: yes, for mocking the connector

PR #2972, opened 2016-07-14
MULE-9738 Improve functional test case to work with extensions - Migr…

gsfernandes:
…ate file extension to use isolated runner

### DIFF
--- a/extensions/file/src/main/resources/META-INF/mule-module.properties
+++ b/extensions/file/src/main/resources/META-INF/mule-module.properties
@@ -1,4 +1,0 @@
-module.name=file
-
-artifact.export.classPackages=org.mule.extension.file.api
-artifact.export.resourcePackages=/META-INF
--- a/extensions/file/src/test/java/org/mule/extension/file/FileConnectorTestCase.java
+++ b/extensions/file/src/test/java/org/mule/extension/file/FileConnectorTestCase.java
@@ -9,9 +9,7 @@ package org.mule.extension.file;
 import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.assertThat;
 import static org.junit.rules.ExpectedException.none;
-
-import org.mule.extension.file.internal.FileConnector;
-import org.mule.functional.junit4.ExtensionFunctionalTestCase;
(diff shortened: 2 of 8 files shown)


extensions/file/pom.xml
```diff
@@ -36,11 +36,6 @@
         </dependency>
 
         <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-        </dependency>
-
-        <dependency>
```
gsfernandes: It is already exposed by mule-core.

extensions/file/src/main/resources/META-INF/mule-module.properties
```diff
@@ -1,4 +0,0 @@
-module.name=file
-
-artifact.export.classPackages=org.mule.extension.file.api
-artifact.export.resourcePackages=/META-INF
```
gsfernandes: No longer needed, any test that needs to use this extension should be migrated to use isolation runner. So far, not even integration tests are using this API for their tests.

extensions/file/src/test/java/org/mule/extension/file/FileConnectorTestCase.java
```diff
@@ -62,12 +60,6 @@ protected void doTearDownAfterMuleContextDispose() throws Exception
         temporaryFolder.delete();
     }
 
-    @Override
-    protected Class<?>[] getAnnotatedExtensionClasses()
-    {
-        return new Class<?>[] {FileConnector.class};
-    }
-
```
gsfernandes: Now extensions are auto-discovered by the runner and their metadata is generated while classloader is built.

gsfernandes:
Ready to be reviewed!

extensions/file/src/test/resources/log4j2-test.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+  ~ The software in this package is published under the terms of the CPAL v1.0
```
pablokraan: We don't add license header to XML files

PR #2977, opened 2016-07-14
Removing tools module

asequeira:


aiannucci:
Reference JIRA issue in the commit

PR #2980, opened 2016-07-15
MULE-10100: Until successful returns event copy causing uses in other…

afelisatti:
… scopes to fail

### DIFF
--- a/tests/integration/src/test/resources/for-each-until-config.xml
+++ b/tests/integration/src/test/resources/for-each-until-config.xml
@@ -26,14 +26,14 @@
 
     <flow name="flowVarSyncNoForEach">
         <set-variable variableName="count" value="#[3]"/>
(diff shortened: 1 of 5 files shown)


tests/integration/src/test/resources/for-each-until-config.xml
```diff
@@ -0,0 +1,50 @@
...
+                    <set-variable variableName="until" value="#[payload]"/>
+                </processor-chain>
+            </until-successful>
+            <set-variable variableName="count" value="#[flowVars.count + payload]"/>
+        </foreach>
+    </flow>
+
+    <flow name="flowVarSyncNoForEach">
+        <set-variable variableName="count" value="#[3]"/>
+            <set-variable variableName="current" value="#[payload]"/>
```
dfeist: formatting

PR #2984, opened 2016-07-18
MULE-10012: file:matcher minSize accepts negative numbers

pablomusumeci:


modules/file-extension-common/src/test/java/org/mule/runtime/module/extension/file/FilePredicateBuilderContractTestCase.java
```diff
@@ -269,6 +275,20 @@ public void rejectSymbolicLink()
         assertReject();
     }
 
+    @Test
+    public void failOnInvalidMinSize()
+    {
+        expectedException.expect(IllegalArgumentException.class);
```
asequeira: Why do you use a rule instead of using `@Test (expected=IllegalArgumentException.class)` if you are not validating the error message?

pablomusumeci: I used expectedException rule because that's the way I saw it was used in all File connector test cases (even though the test didn't check the exception message).

modules/file-extension-common/src/main/java/org/mule/runtime/module/extension/file/api/FilePredicateBuilder.java
```diff
@@ -192,12 +195,14 @@ public T setSymbolicLink(Boolean symbolicLink)
 
     public T setMinSize(Long minSize)
     {
+        checkArgument(minSize >= 0, format(SIZE_MUST_BE_POSITIVE_MESSAGE, "minSize", minSize));
```
marianogonzalez: this is the wrong place to do this. You should do this on the build() method, since the setter is not invoked when the object is create through the SDK.

modules/file-extension-common/src/test/java/org/mule/runtime/module/extension/file/FilePredicateBuilderContractTestCase.java
```diff
@@ -269,6 +275,22 @@ public void rejectSymbolicLink()
...
+        builder.setMinSize(-1L);
+        builder.build();
+    }
+
+    @Test
+    public void failOnInvalidMaxSize()
+    {
+        expectedException.expect(IllegalArgumentException.class);
+        builder.setMinSize(-1L);
```
marianogonzalez: you're setting minSize instead of maxSize

PR #2985, opened 2016-07-18
MULE-10107: High contention when many threads try to create exceptions

elrodro83:
Ref: http://docs.oracle.com/javase/7/docs/technotes/guides/lang/cl-mt.html


modules/launcher/src/test/java/org/mule/module/launcher/SynchronizedClassLoaderTestCase.java
```diff
@@ -8,16 +8,19 @@
...
 
 import java.net.URL;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
+import org.junit.Ignore;
 import org.junit.Test;
 
 @MediumTest
+@Ignore("Synchronization is not a concern since java 7 (MULE-10107)")
```
pablokraan: Why not remove the test instead of ignoring it?

PR #2986, opened 2016-07-18
MULE-9858: Make Message CorrelationId/SequenceID etc. message attributes

elrodro83:
rather than properties

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/MuleMessageCorrelation.java
+++ b/core/src/main/java/org/mule/runtime/core/MuleMessageCorrelation.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core;
+
+import static java.util.Optional.ofNullable;
+import static org.mule.runtime.core.routing.CorrelationMode.ALWAYS;
+import static org.mule.runtime.core.routing.CorrelationMode.IF_NOT_SET;
+import static org.mule.runtime.core.routing.CorrelationMode.NEVER;
+
+import org.mule.runtime.api.message.MuleMessage;
+import org.mule.runtime.core.routing.CorrelationMode;
+
+import java.io.Serializable;
+import java.util.Optional;
+
+/**
+ * Immutable container for correlation properties relative to a {@link MuleMessage}.
+ * 
+ * @since 4.0
+ */
+public class MuleMessageCorrelation implements Serializable
+{
+    private static final long serialVersionUID = -5687080761804624442L;
+
+    public static final String NOT_SET = "<not set>";
+
+    private final String id;
+    private final Integer groupSize;
+    private final Integer sequence;
+
+    /**
+     * Builds a new {@link MuleMessageCorrelation} with the given parameters.
+     * 
+     * @param id see {@link #getId()}.
+     * @param groupSize see {@link #getGroupSize()}.
+     * @param sequence see {@link #getSequence()}.
+     */
+    public MuleMessageCorrelation(String id, Integer groupSize, Integer sequence)
+    {
+        this.id = id;
+        this.groupSize = groupSize;
+        this.sequence = sequence;
+    }
+
+    /**
+     * The correlation Id can be used by components in the system to manage message relations.
+     * <p>
+     * The id is associated with the message using the underlying transport protocol. As such not all messages will
+     * support the notion of a id i.e. tcp or file. In this situation the correlation Id is set as a property of the
+     * message where it's up to developer to keep the association with the message. For example if the message is
+     * serialised to xml the id will be available in the message.
+     *
+     * @return the id for this message or null value if one hasn't been set
+     */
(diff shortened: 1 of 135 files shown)


core/src/main/java/org/mule/runtime/core/MuleMessageCorrelation.java
```diff
@@ -0,0 +1,109 @@
...
+ * @since 4.0
+ */
+public class MuleMessageCorrelation implements Serializable
+{
+    private static final long serialVersionUID = -5687080761804624442L;
+
+    public static final String NOT_SET = "<not set>";
+
+    private final String id;
+    private final Integer groupSize;
```
dfeist: Why not use Optional internally to?

core/src/main/java/org/mule/runtime/core/MuleMessageCorrelation.java
```diff
@@ -0,0 +1,109 @@
...
+ * 
+ * @since 4.0
+ */
+public class MuleMessageCorrelation implements Serializable
+{
+    private static final long serialVersionUID = -5687080761804624442L;
+
+    public static final String NOT_SET = "<not set>";
+
+    private final String id;
```
dfeist: I'm not convinced it makes sense to mixed correlationId and group size/sequence in the same container object, because for a lot of use cases a correlationID is required but group size/sequence is only required to limited number of use cases when groups of messages need to be represented.

core/src/main/java/org/mule/runtime/core/MuleMessageCorrelation.java
```diff
@@ -0,0 +1,109 @@
...
+    public Optional<Integer> getGroupSize()
+    {
+        return ofNullable(groupSize);
+    }
+
+    /**
+     * @param enableCorrelation the correlation mode of the calling component.
+     * @return whether correlation has to be handled for the message that has this {@link MuleMessageCorrelation}.
+     */
+    public boolean doCorrelation(CorrelationMode enableCorrelation)
```
dfeist: I'd put this in CorrelationMode instead.

core/src/main/java/org/mule/runtime/core/MuleMessageCorrelation.java
```diff
@@ -0,0 +1,109 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core;
```
dfeist: message package

core/src/main/java/org/mule/runtime/core/MuleMessageCorrelation.java
```diff
@@ -0,0 +1,109 @@
...
+
+import java.io.Serializable;
+import java.util.Optional;
+
+/**
+ * Immutable container for correlation properties relative to a {@link MuleMessage}.
+ * 
+ * @since 4.0
+ */
+public class MuleMessageCorrelation implements Serializable
```
dfeist: Simpler name?

core/src/main/java/org/mule/runtime/core/MuleMessageCorrelation.java
```diff
@@ -0,0 +1,109 @@
...
+    private final Integer sequence;
+
+    /**
+     * Builds a new {@link MuleMessageCorrelation} with the given parameters.
+     * 
+     * @param id see {@link #getId()}.
+     * @param groupSize see {@link #getGroupSize()}.
+     * @param sequence see {@link #getSequence()}.
+     */
+    public MuleMessageCorrelation(String id, Integer groupSize, Integer sequence)
```
dfeist: Isn't it better to use two constructors, one that just takes id, and another that takes string and 2 int's (but not use Integer)?

core/src/main/java/org/mule/runtime/core/MuleMessageCorrelation.java
elrodro83: Optional is not serializable

dfeist:
Remove org.mule.runtime.core.message.DefaultMuleMessageBuilder.MuleMessageImplementation#findProperty

core/src/main/java/org/mule/runtime/core/routing/CorrelationMode.java
```diff
@@ -6,7 +6,22 @@
...
 public enum CorrelationMode
 {
-    IF_NOT_SET, ALWAYS, NEVER
+    IF_NOT_SET, ALWAYS, NEVER;
+
+    /**
+     * @param message the message to check for its correlation attributes.
+     * @return whether correlation has to be handled for the message that has this {@link Correlation}.
+     */
+    public boolean doCorrelation(MuleMessage message)
```
dfeist: Why not pass Correlation object instead of whole message?

core/src/main/java/org/mule/runtime/core/message/Correlation.java
```diff
@@ -22,7 +18,7 @@
  * 
  * @since 4.0
  */
-public class MuleMessageCorrelation implements Serializable
+public class Correlation implements Serializable
```
dfeist: On second thought can we use CorrelationIdentifier?  I don't think Correlation really describes what it is..

PR #2987, opened 2016-07-18
MULE-9956: Error when triggering directory-listener

marianogonzalez:


extensions/file/src/main/java/org/mule/extension/file/internal/DirectoryListener.java
```diff
@@ -239,16 +238,17 @@ private synchronized void initialiseClusterListener()
...
-                catch (Exception e)
-                {
-                    throw new MuleRuntimeException(e);
-                }
-            }, muleContext);
+            clusterListener = new PrimaryNodeLifecycleNotificationListener(() ->
+                                                                           {
+                                                                               try
+                                                                               {
+                                                                                   DirectoryListener.this.start();
```
elrodro83: can't you just do 'start();' ?

PR #2990, opened 2016-07-18
MULE-10088: Create a class loader builder to easily create a classloa…

gsfernandes:
…der for plugins - Fixed distro plugins folder


gsfernandes:
@pablolagreca this was something missed in your original PR please include this change too.

PR #2994, opened 2016-07-18
MULE-9973: directory-listener fails when file is locked by file:read

marianogonzalez:


extensions/file/src/main/java/org/mule/extension/file/internal/DirectoryListener.java
```diff
@@ -504,6 +497,11 @@ private void registerPath(Path path) throws IOException
         keyPaths.put(key, path);
     }
 
+    private Path resolveRootPath()
+    {
+        return new DirectoryListenerCommand((LocalFileSystem) fileSystem).resolveRootPath(config, directory);
```
elrodro83: can this command be an attribute, to avoid recreating it in each call?

PR #2995, opened 2016-07-18
MULE-10016: Fail matcher symbolic link filter doesn't work

marianogonzalez:


extensions/file/src/main/java/org/mule/extension/file/api/LocalFileAttributes.java
```diff
@@ -94,7 +95,7 @@ public boolean isDirectory()
     @Override
     public boolean isSymbolicLink()
     {
-        return getAttributes().isSymbolicLink();
+        return Files.isSymbolicLink(Paths.get(getPath()));
```
elrodro83: static import

marianogonzalez: can't. The static method has the same name as the method invoking it

PR #3003, opened 2016-07-19
MULE-10040: Fail to overwrite directory when renaming

pablomusumeci:


extensions/file/src/test/java/org/mule/extension/file/FileRenameTestCase.java
```diff
@@ -86,7 +114,8 @@ public void targetAlreadyExistsWithoutOverwrite() throws Exception
     public void targetAlreadyExistsWithOverwrite() throws Exception
     {
         File origin = createHelloWorldFile();
-        temporaryFolder.newFile(RENAME_TO);
+        File targetFile = new File(origin.getParent(), RENAME_TO);
+        targetFile.createNewFile();
```
marianogonzalez: why this change? what's the difference?

pablomusumeci: `temporaryFolder.newFile(RENAME_TO);` was creating a file in the root folder. So we had:

current/files/hello.json
current/rename

When we were checking if the file already existed, we were doing that in the same directory of hello.json (current/files/rename). The file wasn't there, so it could be renamed without any problem. We didn't notice any error on the test because the original file was being moved to the desired location (current/files/rename), but no file was being overwritten.  So, the the things the test was testing were both truth (the file renamed was no longer there, and at the end it was where it was supposed to be).

PR #3004, opened 2016-07-19
MULE-10018: Cannot use datetime constant in file matchers

juandesi:
* Added support for LocalDateTime parsing.


tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/model/LifetimeInfo.java
```diff
@@ -38,4 +43,14 @@ public void setDateOfDeath(Calendar dateOfDeath)
     {
         this.dateOfDeath = dateOfDeath;
     }
+
+    public LocalDateTime getDateOfConception()
```
marianogonzalez: you're missing a test in which you verifiy that this was parsed correctly

PR #3009, opened 2016-07-20
MULE-9901: Enrich File,FTP,SFTP and FTPS extension with display prope…

estebanwasinger:
…rties


modules/file-extension-common/src/main/java/org/mule/runtime/module/extension/file/api/FileDisplayConstants.java
```diff
@@ -0,0 +1,13 @@
+package org.mule.runtime.module.extension.file.api;
```
aiannucci: Missing license header

PR #3013, opened 2016-07-20
MULE-9855 Make message immutable. Cleanup message implementation.

dfeist:


elrodro83:
I think there were some groovy tests that used some of the removed methods.

PR #3027, opened 2016-07-21
MULE-9901: Enrich Socket extension with display properties

estebanwasinger:


extensions/sockets/src/main/java/org/mule/extension/socket/api/connection/tcp/protocol/CustomProtocol.java
```diff
@@ -31,8 +34,15 @@
...
     private TcpProtocol delegate;
 
+    /**
+     * Reference to full qualifier class name that should extend {@link TcpProtocol} that will be used as a custom
+     * protocol
+     */
     @Parameter
     @Alias("class")
+    @Summary("Reference to full qualifier class name that must implement " +
+             "'org.mule.extension.socket.api.connection.tcp.protocol.TcpProtocol' that will be used as a custom protocol")
```
marianogonzalez: remove the classname reference. Just say the name of the class in which the thingy is implemented

PR #3028, opened 2016-07-21
MULE-9856: Create MuleMessage builder - Api improvements

elrodro83:


core/src/main/java/org/mule/runtime/core/api/MuleMessage.java
```diff
@@ -113,7 +112,7 @@ static MuleMessage of(Object payload)
         Builder payload(Object payload);
 
         @Override
-        CollectionBuilder collectionPayload(Collection payload, Class<?> itemType);
+        CollectionBuilder collectionPayload(Iterable payload, Class<?> itemType);
```
dfeist: Collection or iterable?

core/src/main/java/org/mule/runtime/core/api/MuleMessage.java
```diff
@@ -113,7 +114,13 @@ static MuleMessage of(Object payload)
         Builder payload(Object payload);
 
         @Override
+        CollectionBuilder collectionPayload(Iterator payload, Class<?> itemType);
+
+        @Override
         CollectionBuilder collectionPayload(Collection payload, Class<?> itemType);
+
+        @Override
+        CollectionBuilder collectionPayload(Object[] payload, Class<?> itemType);
```
dfeist: Do we need this, given the type can be obtained via payload.getClass().getComponentType()?  Can't we just add logic to create a CollectionDataType if `payload(String[] stringArray)` is used?

core/src/main/java/org/mule/runtime/core/api/MuleMessage.java
```diff
@@ -113,7 +114,13 @@ static MuleMessage of(Object payload)
         Builder payload(Object payload);
 
         @Override
+        CollectionBuilder collectionPayload(Iterator payload, Class<?> itemType);
```
dfeist: After thinking about this naming I prefer streamPayload.  Why? Because we will eventually use this not just for sequences of messages (which are also streams, we are used to thinking about them as a collection) but also for streams of ByteBuffers as we move from blocking InputStream streaming to non-blocking streaming.  And given this collectionPayload(Iterator<ByteBuffer>, ByteBuffer.class) or collectionPayload(Publisher<ByteBuffer>, ByteBuffer.class) don't work for me, when these are quite clearly streams and not finite.

core/src/main/java/org/mule/runtime/core/message/DefaultMuleMessageBuilder.java
```diff
@@ -150,16 +150,35 @@ public DefaultMuleMessageBuilder(org.mule.runtime.api.message.MuleMessage messag
...
         this.payload = payload;
         this.dataType = DataType.builder().collectionType(payload.getClass()).itemType(clazz).build();
         return this;
     }
 
     @Override
+    public MuleMessage.CollectionBuilder collectionPayload(Iterator payload, Class<?> clazz)
+    {
+        requireNonNull(payload);
+        return innerCollectionPayload(() -> payload, clazz);
```
dfeist: By converting to iterable, we loose the Iterator DataType.type.

PR #3041, opened 2016-07-22
MULE-10019: SDK date attributes should support ISO 8601

pablomusumeci:


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ExtensionDefinitionParser.java
```diff
@@ -743,71 +743,90 @@ private boolean isExpressionFunction(MetadataType metadataType)
         }
     }
 
-    private ValueResolver parseDateTime(Object value, DateTimeType dataType, Object defaultValue)
+
+    private static ValueResolver doParseDate(Object value, Class<?> type)
```
marianogonzalez: why static?

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ExtensionDefinitionParser.java
```diff
@@ -743,71 +743,90 @@ private boolean isExpressionFunction(MetadataType metadataType)
...
+        if (value instanceof String)
         {
-            Date date = doParseDate(value, DATE_FORMAT, defaultValue);
-            return new StaticValueResolver(LocalDateTime.ofInstant(date.toInstant(), ZoneId.systemDefault()));
+            Object constructedValue = null;
+            DateTime dateTime = getParsedDateTime((String) value);
+
+            if (type.equals(LocalDate.class))
+            {
+                constructedValue = LocalDate.of(dateTime.getYear(), dateTime.getMonthOfYear(), dateTime.getDayOfMonth());
```
marianogonzalez: LocalDate also includes hours. You're missing information here. Can't you construct the value from the dateTime's milliseconds?

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ExtensionDefinitionParser.java
pablomusumeci: my bad

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ExtensionDefinitionParser.java
```diff
@@ -743,71 +743,90 @@ private boolean isExpressionFunction(MetadataType metadataType)
...
-            calendar.setTime(date);
-            return new StaticValueResolver(calendar);
+            return new StaticValueResolver(value);
         }
+
+        throw new IllegalArgumentException(format("Could not transform value of type '%s' to a valid date type", value != null ? value.getClass().getName() : "null"));
     }
 
-    private ValueResolver parseDate(Object value, DateType dateType, Object defaultValue)
+    private static DateTime getParsedDateTime(String value)
```
marianogonzalez: why static?

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ExtensionDefinitionParser.java
```diff
@@ -743,71 +743,90 @@ private boolean isExpressionFunction(MetadataType metadataType)
...
 
-            value = defaultValue;
+        Class<?> type = getType(dateType);
+        if (isExpression(value, parser))
+        {
+            return new TypeSafeExpressionValueResolver((String) value, type);
         }
 
-        if (value instanceof String)
+        if (value == null)
```
marianogonzalez: this is wrong. Nothing of this makes sense if the value is not an String. You should keep the original logic with the only exception of the SimpleDateFormat thing

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ExtensionDefinitionParser.java
pablomusumeci: LocalDate jdoc says:

{@code LocalDate} is an immutable date-time object that represents a date, often viewed as year-month-day. This class does not store or represent a time or time-zone.

marianogonzalez: if that's true then you shouldn't be using this class at all. You're using this same parse method for both Date and DateTime metadata types. It looks like you're consciously dropping the time part of it. Actually, given your response I took a second look at the test cases and it looks like although you do have a time value in your tests, you're never asserting that. Did you mean that or was it an oversight?

In any case, you need a parse method which accounts for the full ISO spec. That means, including the time and timezone

pablomusumeci: I'm dropping the time part when the type I need to construct doesn't make use of that information (like Date for instance). The thing here is, that because LocalDate adapts the time to the local time zone, we cannot make assertions against the hour/date/minute. 

Looking through the code I saw that he have the same exact problem in [some other test](https://github.com/mulesoft/mule/blob/mule-4.x/modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/config/ParameterizedConfigParserTestCase.java#L237).
 
I think that, in order to make a complete test (including time zones) we should give support to types like [java.time.ZonedDateTime](https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html).

marianogonzalez: No. That is wrong. Date class can handle time. Calendar class can handle time and timeZone as well. The only real limitant that you have here is that of the framework you want to use to parse the String. If Joda doesn't do what you want then drop Joda, don't change the feature to match the code you have. I'm sure the internet is full of examples of how to do that.

What we need is to be able to parse ANY iso String into ANY of the aforementioned types.

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ExtensionDefinitionParser.java
```diff
@@ -743,71 +746,85 @@ private boolean isExpressionFunction(MetadataType metadataType)
...
+        try
         {
-            return new TypeSafeExpressionValueResolver((String) value, getType(dateType));
+            return ISODateTimeFormat.dateTimeParser().withOffsetParsed().parseDateTime(value);
         }
-        else
+        catch (DateTimeParseException e)
         {
-            return new StaticValueResolver(doParseDate(value, DATE_FORMAT, defaultValue));
+            throw new IllegalArgumentException(format("Could not parse value '%s' according to ISO 8601", value));
```
marianogonzalez: do not hide the original exception

PR #3045, opened 2016-07-25
MULE-10107: High contention when many threads try to create exception…

gsfernandes:
…s - Fixing race condition for multiple threads loading same class


pablokraan:
This PR has to be done against 3.x branch

gsfernandes:
Done, #3046 created for 3.x branch.

PR #3050, opened 2016-07-25
MULE-10077: Metadata Type resolution fails when an operation returns …

alegmarra:
…int[][]


modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/AnnotationsBasedDescriberTestCase.java
```diff
@@ -590,6 +592,10 @@ private void assertTestModuleOperations(ExtensionDeclaration extensionDeclaratio
                                 .build(),
                         true, SUPPORTED, null);
 
+        operation = getOperation(extensionDeclaration, GET_GRAMS_IN_STORAGE);
+        assertThat(operation, is(notNullValue()));
+        assertThat(operation.getParameters().isEmpty(), is(true));
```
marianogonzalez: add an assertion to verify the operation type

PR #3055, opened 2016-07-26
MULE-9699: Migrate static resource handler to HTTP extension

afelisatti:


extensions/http/src/main/java/org/mule/extension/http/internal/StaticResourceLoader.java
```diff
@@ -0,0 +1,118 @@
...
+
+public class StaticResourceLoader
+{
+    private static final String ANY_PATH = "/*";
+    private static final String ROOT_PATH = "/";
+    private static final String DEFAULT_MIME_TYPE = "application/octet-stream";
+    private MimetypesFileTypeMap mimeTypes = new MimetypesFileTypeMap();
+
+    /**
+     * The resource base for where documents are served up from.
```
estebanwasinger: It would nice to add a description or example of how to reference the "resource base", generally this is confusing step

extensions/http/src/main/java/org/mule/extension/http/internal/StaticResourceLoader.java
```diff
@@ -0,0 +1,118 @@
...
+     */
+    @Parameter
+    private String resourceBase;
```
estebanwasinger: The parameter name doesn't indicate precisely what is required, at least I would change it to "resourceBasePath" to be clear that what is required is a folder path

extensions/http/src/main/java/org/mule/extension/http/internal/StaticResourceLoader.java
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * The default file to serve when a directory is specified. The default value is 'index.html'.
+     */
+    @Parameter
+    @Optional(defaultValue = "index.html")
+    private String defaultFile;
+
+    public MuleMessage load(MuleEvent event) throws ResourceNotFoundException
+    {
+        checkArgument(event.getMessage().getAttributes() instanceof HttpRequestAttributes, "Message attributes must be HttpRequestAttributes.");
```
pablolagreca: Create an issue to track this as something TODO before releasing Mule 4. I would prefer to avoid this kind of requirement from components. As we discussed yesterday, it's better if the tooling helps the user to autocomplete the required input parameters of the load-static-resource operation instead of automatically discovering those from an HttpRequestAttributes object.

PR #3060, opened 2016-07-28
MULE-10134: Migrate extensions to use isolated test runner - Migrate …

gsfernandes:
…HTTP extension to use isolated test runner


tests/http/src/test/java/org/mule/test/module/http/functional/AbstractHttpTestCase.java
```diff
@@ -0,0 +1,18 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.test.module.http.functional;
+
+import org.mule.extension.http.internal.request.validator.HttpRequesterConfig;
+import org.mule.extension.http.internal.request.validator.HttpRequesterProvider;
+import org.mule.functional.junit4.MuleArtifactFunctionalTestCase;
+import org.mule.functional.junit4.runners.ArtifactClassLoaderRunnerConfig;
+
+@ArtifactClassLoaderRunnerConfig(exportClasses = {HttpRequesterProvider.class, HttpRequesterConfig.class})
```
pablokraan: Why is required to export this classes?

tests/http/src/test/java/org/mule/test/module/http/functional/AbstractClientTlsRestrictedTlsProtocolsAndCiphersTestCase.java
```diff
@@ -35,7 +38,13 @@
 import org.junit.Rule;
 import org.junit.Test;
 
-public abstract class AbstractClientTlsRestrictedTlsProtocolsAndCiphersTestCase extends AbstractHttpTestCase
+/**
+ * Needs to access resources for validating different TLS scenarios and it cannot be done with
```
pablokraan: Create a abstract class to avoid duplicating this javadoc and the getAnnotatedExtensionClasses in the two classes that require this changes.

tests/http/src/test/java/org/mule/test/module/http/functional/HttpListenerStaticResourcesTestCase.java
```diff
@@ -113,6 +114,7 @@ public void contentTypeShouldBeDeterminedFromResource() throws Exception
         assertResponseContentType("image/gif");
     }
 
+    @Ignore("It is not supported to access application resources from extension classloader")
```
pablokraan: Add the reference to a JIRA so the exclusion can be removed when is fixed

tests/http/src/test/resources/log4j2-test.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
pablokraan: Is this file required?

tests/http/src/test/java/org/mule/test/module/http/functional/AbstractHttpTestCase.java
gsfernandes: There are two tests that use those classes, both validate if the connector is well configured so it asserts those objects.

tests/http/src/test/resources/log4j2-test.xml
gsfernandes: No, just added for debugging the class loaders built for the test.

PR #3062, opened 2016-07-28
MULE-10038: Improve System tests framework.

asequeira:


tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleDeployment.java
```diff
@@ -6,19 +6,25 @@
...
+import static java.lang.Boolean.parseBoolean;
 import static java.lang.Integer.parseInt;
 import static java.lang.System.getProperty;
+import static java.util.concurrent.TimeUnit.*;
 import static org.apache.commons.io.FilenameUtils.removeExtension;
 import org.mule.tck.probe.JUnitProbe;
 import org.mule.tck.probe.PollingProber;
 import org.mule.test.infrastructure.process.MuleProcessController;
 
+import com.jayway.awaitility.Awaitility;
```
aiannucci: is this import needed?

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleDeployment.java
```diff
@@ -192,16 +199,46 @@ protected void before() throws Throwable
...
+            logger.info("Deployment successful");
+        }
+        else
+        {
+            logger.warn("Mule Server was already running");
+            libraries.forEach((library) -> mule.addLibrary(new File(library)));
+            logger.info("Redeploying domains");
+            domains.forEach((domain) -> redeployDomain(domain));
+            logger.info("Redeploying applications");
+            applications.forEach((application) -> redeploy(application));
```
aiannucci: What happens if Mule is running but I'm trying to deploy a different application (i changed its name), this is going to call **redeploy** and try to undeploy the unexisting application. Should we consider this scenario?

asequeira: That is out of scope, renaming an application is not as usual as one would like to provide support to it.

PR #3065, opened 2016-07-28
MULE-9489: Provide an InterceptiongMessageProcessor equivalent in SDK

marianogonzalez:


modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/InterceptingOperationMessageProcessorTestCase.java
```diff
@@ -0,0 +1,121 @@
...
+
+        messageProcessor.setListener(next);
+
+        return messageProcessor;
+    }
+
+    @Test
+    public void nextExecuted() throws Exception
+    {
+        org.mule.runtime.core.api.MuleEvent resultEvent = mock(org.mule.runtime.core.api.MuleEvent.class);
```
pablolagreca: Do you really need this to be a qualified name?

marianogonzalez: i'm using both the new and the old type of events because MessageProcessor API hasn't been updated yet. I try not to import the old API so that it's easier to visualize when the MUle API is more mature and these components ready for update

pablolagreca: make sense

PR #3066, opened 2016-07-29
MULE-10171: String attachments do not maintain content type when sent…

pablokraan:
… on HTTP transport

### DIFF
--- a/transports/http/src/main/java/org/mule/transport/http/transformers/ObjectToHttpClientMethodRequest.java
+++ b/transports/http/src/main/java/org/mule/transport/http/transformers/ObjectToHttpClientMethodRequest.java
@@ -31,7 +31,9 @@ import org.mule.util.IOUtils;
 import org.mule.util.ObjectUtils;
 import org.mule.util.StringUtils;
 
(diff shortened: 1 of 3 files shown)


transports/http/src/main/java/org/mule/transport/http/transformers/ObjectToHttpClientMethodRequest.java
```diff
@@ -520,4 +528,63 @@ else if (dh.getDataSource() instanceof PartDataSource)
...
+        public CustomStringPart(String name, String value, String charset, String contentType)
+        {
+            super(name, contentType, charset == null ? DEFAULT_CHARSET : charset, DEFAULT_TRANSFER_ENCODING);
+            if (value == null)
+            {
+                throw new IllegalArgumentException("Value may not be null");
+            }
+            else if (value.indexOf(0) != -1)
+            {
+                throw new IllegalArgumentException("NULs may not be present in string parts");
```
elrodro83: NULLs

elrodro83:
Can you check if the http connector has a similar test case?

PR #3078, opened 2016-08-02
MULE-10180: Classloader leak when Oracle JDBC Driver is not used but …

gsfernandes:
…included in application lib folder

_URLClassLoader in Java 7 added a close method that closes any files that were opened by it, removed workaround. This fixed the locks for files in windows, files were locked by OracleDriver
_Improvements in DefaultReleaser to avoid classloader leaks

### DIFF
--- a/modules/launcher/src/main/java/org/mule/module/launcher/GoodCitizenClassLoader.java
+++ b/modules/launcher/src/main/java/org/mule/module/launcher/GoodCitizenClassLoader.java
@@ -15,8 +15,6 @@ import java.net.URL;
 import java.net.URLClassLoader;
 import java.net.URLStreamHandler;
 import java.net.URLStreamHandlerFactory;
-import java.util.Collection;
-import java.util.jar.JarFile;
 
 import sun.net.www.protocol.jar.Handler;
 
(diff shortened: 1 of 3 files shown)


gsfernandes:
@pablokraan it would be great if you could review this please, you have more experience with this topic. Thanks!

distributions/standalone/src/main/resources/conf/wrapper.conf
```diff
@@ -14,27 +14,30 @@ wrapper.java.additional.3=-Djava.net.preferIPv4Stack=TRUE
 # Needed to avoid a memory leak on mvel (see MULE-7650)
 wrapper.java.additional.4=-Dmvel2.disable.jit=TRUE
 
+# Needed to avoid memory leak on JDBC DriverManager (see MULE-10180)
+wrapper.java.additional.5=-Dmule.jdbcDriverManagerInitializer=TRUE
```
pablokraan: mule.jdbcDriverManagerInitializer > mule.db.initializeJdbcDrivers

modules/launcher/src/main/java/org/mule/module/launcher/util/JdbcDriverManagerInitializer.java
```diff
@@ -0,0 +1,64 @@
...
+        {
+            // Use the system classloader as the victim for all this
+            // ClassLoader pinning we're about to do.
+            Thread.currentThread().setContextClassLoader(getSystemClassLoader());
+
+            /*
+             * First call to this loads all drivers in the current class
+             * loader
+             */
+            if (JDBC_DRIVER_MANAGER_INITIALIZER_PROPERTY.isEnabled())
```
pablokraan: This check should be the first thing this method does

modules/launcher/src/main/java/org/mule/module/launcher/MuleContainer.java
```diff
@@ -135,6 +136,10 @@ protected void init(String[] args) throws IllegalArgumentException
         {
             setStartupPropertiesFile(propertiesFile);
         }
+
+        JdbcDriverManagerInitializer jdbcDriverManagerInitializer = new JdbcDriverManagerInitializer();
```
pablokraan: I don't get this: the jdbcDriverManagerInitializer is created on the container and uses the system classloader to initialize the drivers. How is this helping to fix the memory leak if the DB driver is inside the application classloader, which does not even exists at this point?

modules/launcher/src/main/java/org/mule/module/launcher/util/JdbcDriverManagerInitializer.java
```diff
@@ -0,0 +1,64 @@
...
+            {
+                getDrivers();
```
pablokraan: I remember some code was added to the application classloader or something like that in order to release resources on application undeployment caused by drivers that were registered but never unregistered. Is not this code going to cause that same issue we fixed?

modules/launcher/src/main/java/org/mule/module/launcher/MuleContainer.java
gsfernandes: DriverManager uses Service Provider for loading JDBC 4.0 drivers as part of DriverManager class initialization.

If an application didn't use the driver (didn't get a connection from DriverManager neither did a Class.forName("driverName")) when application classloader is disposed the DriverManager is called for the first time therefore it was loading all the drivers found by SPI before they were just about to be deregistered.

This "side-effect" ended up locking the OracleDriver jar due to OracleDriver leaves inputStreams open to get error mappings from properties file. But the GoodCitizenClassLoader was doing a workaround for closing any opened jar and resources, instead now calling URLClassLoader.close() (method added in Java 7) it does a better job and resources are released and the Oracle jar file is not locked and those input streams closed.

I found this after researched how DriverManager could be handled. I just validated and  confirmed that it is not needed to do the DriverManager initialization in order to fix this issue. 
I don't see that avoiding loading drivers when not used and are about to be deregistered could have any other issue now that the classloaders are disposed correctly.

PR #3079, opened 2016-08-02
MULE-10140: Support XA transactions in SDK extensions at the operatio…

marianogonzalez:
…n level

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/ConnectionProviderModelValidator.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/ConnectionProviderModelValidator.java
@@ -6,16 +6,36 @@
  */
 package org.mule.runtime.module.extension.internal.introspection.validation;
 
-import static org.mule.runtime.module.extension.internal.util.MuleExtensionUtils.getOperationsConnectionType;
+import static java.lang.String.format;
+import static org.apache.commons.collections.CollectionUtils.isEmpty;
+import static org.mule.metadata.java.api.utils.JavaTypeUtils.getType;
+import org.mule.runtime.api.connection.CachedConnectionProvider;
+import org.mule.runtime.api.connection.ConnectionProvider;
+import org.mule.runtime.extension.api.ExtensionWalker;
+import org.mule.runtime.extension.api.connectivity.TransactionalConnection;
 import org.mule.runtime.extension.api.exception.IllegalModelDefinitionException;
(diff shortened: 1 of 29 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/ConnectionProviderModelValidator.java
```diff
@@ -36,32 +56,121 @@
...
+                                                               Class<?> connectionType = ((RuntimeConnectionProviderModel) providerModel).getConnectionType();
+                                                               configModel.getOperationModels().forEach(operationModel -> validateConnectionTypes(extensionModel, providerModel, operationModel, connectionType));
+                                                           }
+                                                       });
+    }
+
+    private <T> Optional<Class<T>> getConnectionType(EnrichableModel model)
+    {
+        Optional<ConnectivityModelProperty> connectivityProperty = model.getModelProperty(ConnectivityModelProperty.class);
+        if (!connectivityProperty.isPresent()) {
```
elrodro83: use && instead of nested ifs

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/transaction/XAExtensionTransactionalResource.java
```diff
@@ -0,0 +1,168 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public void close() throws Exception
+    {
+        try
+        {
+            getConnection().close();
+        }
+        catch (Exception e)
```
elrodro83: shouldn't this be rethrown, according to the signature?

marianogonzalez: this and other things like the synchronized blocks and stuff are patterns already present in existing XA enabled classes such as ConnectorWrapper or BitronixConnectorWrapper. Because all of them have the same pattern, I take it these are somehow correct and necessary. 

They can however be reviewed when we revisit transactions as a whole at the runtime level

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/ConnectionProviderModelValidator.java
marianogonzalez: fixed

PR #3082, opened 2016-08-03
MULE-10191: Query named parameters are not validated properly

pablokraan:


modules/db/src/main/java/org/mule/module/db/internal/config/domain/query/ParameterizedQueryTemplateFactoryBean.java
```diff
@@ -59,13 +60,36 @@ public QueryTemplate getObject() throws Exception
...
+
+        for (QueryParam queryParam : unresolvedQueryParams)
+        {
+            if (errorMsgBuilder.length() > 0)
+            {
+                errorMsgBuilder.append(", ");
+            }
+            errorMsgBuilder.append("'").append(queryParam.getName()).append("'");
+        }
+        errorMsgBuilder.insert(0, "There is at least a query parameter that do not match the name of any parameter defined in the query text. Unresolved parameters: ");
```
elrodro83: do -> does

PR #3085, opened 2016-08-03
MULE-9986: Introduce MultipartPayload

elrodro83:


core/src/main/java/org/mule/runtime/core/message/AttachmentAttributes.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.Map;
+
+/**
+ * Representation of metadata associated to an attachment.
+ *
+ * @since 4.0
+ */
+public class AttachmentAttributes extends BaseAttributes
```
dfeist: Shouldn't this be PartAttributes?  Can this be an inner class in MultiPartPayload maybe?

core/src/main/java/org/mule/runtime/core/message/AttachmentAttributes.java
```diff
@@ -0,0 +1,98 @@
...
+    public String getFileName()
+    {
+        return fileName;
+    }
+
+    /**
+     * @return The size in bytes of the attachment, as provided by the source. In case it is unknown or cannot be
+     *         determined, -1 will be returned.
+     */
+    public long getSize()
```
dfeist: Is this `fileSize`? Would an optional be better than -1?

core/src/main/java/org/mule/runtime/core/message/MultiPartPayload.java
```diff
@@ -0,0 +1,123 @@
...
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.NoSuchElementException;
+
+/**
+ * Represents a payload of a {@link MuleMessage} composed of many different parts. Each parts is in itself a
+ * {@link MuleMessage}, and has {@code attributes} specific to that parts (such as http part headers).
+ * <p>
+ * This is useful for representing attachments as part of the payload of a message.
```
dfeist: I'd remove the "attachment" terminology everywhere IMO.

core/src/main/java/org/mule/runtime/core/message/MultiPartPayload.java
```diff
@@ -0,0 +1,123 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableList.Builder;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.NoSuchElementException;
+
+/**
+ * Represents a payload of a {@link MuleMessage} composed of many different parts. Each parts is in itself a
+ * {@link MuleMessage}, and has {@code attributes} specific to that parts (such as http part headers).
```
dfeist: Are the attributes are always the same type, or extend the same type? If so link to it instead of  {@code attributes}

dfeist: http part headers -> ?

core/src/main/java/org/mule/runtime/core/message/MultiPartPayload.java
```diff
@@ -0,0 +1,123 @@
...
+ * @since 4.0
+ */
+public class MultiPartPayload implements Serializable
+{
+    private static final long serialVersionUID = -1435622001805748221L;
+
+    /**
+     * The name of a part that does <b>not</b> represent an attachment.
+     */
+    public static final String BODY_PART_NAME = "_body";
```
dfeist: It would be best to not expose this, but rather expose a `Optional getBody()` to obtain body.

core/src/main/java/org/mule/runtime/core/message/MultiPartPayload.java
```diff
@@ -0,0 +1,123 @@
...
+    {
+        this(asList(parts));
+    }
+
+    /**
+     * Builds a new {@link MultiPartPayload} with the given {@link MuleMessage}s as parts.
+     * 
+     * @param parts
+     */
+    public MultiPartPayload(List<MuleMessage> parts)
```
dfeist: What about the iterator case?  Also we use collection for message collections but require list for collection of parts, does this need to be list?

core/src/main/java/org/mule/runtime/core/message/AttachmentAttributes.java
elrodro83: In the case of http, -1 is the value returned when the size is unknown. The idea of the attributes object would be to provide the value as received, rather than interpret it somehow.

core/src/main/java/org/mule/runtime/core/message/AttachmentAttributes.java
elrodro83: Maybe, but the contained state is specific for attachments.

core/src/main/java/org/mule/runtime/core/message/MultiPartPayload.java
elrodro83: not necessarily. So far i saw 2 cases: AttachmentAttributes for attachments and NullAttributes for a body.

core/src/main/java/org/mule/runtime/core/message/PartAttributes.java
```diff
@@ -0,0 +1,100 @@
...
+     * @param size The size in bytes of the attachment, or -1 if unknown or undetermined.
+     * @param headers The headers relative to the attachment part.
+     */
+    public PartAttributes(String name,
+            String fileName,
+            long size,
+            Map<String, LinkedList<String>> headers)
+    {
+        requireNonNull(name);
+        this.name = name;
```
dfeist: we can do `this.name = requireNonNull(name);`

core/src/main/java/org/mule/runtime/core/message/PartAttributes.java
```diff
@@ -0,0 +1,100 @@
...
+        this.fileName = fileName;
+        this.size = size;
+        this.headers = ImmutableMap.copyOf(headers);
```
dfeist: Do we want to use static imports for copy() and emptyMap() in this class? We do in most places...

core/src/main/java/org/mule/runtime/core/transformer/AbstractTransformer.java
```diff
@@ -274,6 +276,13 @@ public Object transform(Object src, Charset enc) throws TransformerException
         }
 
         DataType sourceType = DataType.fromObject(payload);
+        if (MultiPartPayload.class.isAssignableFrom(sourceType.getType()))
+        {
+            Message msg = MessageFactory.createStaticMessage("\"%s\" cannot be transformed to %s.", MultiPartPayload.class.getSimpleName(), getReturnDataType().getType().getName());
+            /// FIXME
```
dfeist: // TODO + JIRA?

PR #3088, opened 2016-08-04
MULE-10196: When AbstractConnector fails to connect receivers, it leaves

elrodro83:
the connector connection active

### DIFF
--- a/core/src/main/java/org/mule/transport/AbstractConnector.java
+++ b/core/src/main/java/org/mule/transport/AbstractConnector.java
@@ -1663,39 +1663,49 @@ public abstract class AbstractConnector extends AbstractAnnotatedObject implemen
     {
         doConnect();
 
-        if (receivers != null)
+        try
         {
-            for (MessageReceiver receiver : receivers.values())
+            if (receivers != null)
             {
-                final List<MuleException> errors = new ArrayList<>();
-                try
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mule/transport/AbstractConnector.java
```diff
@@ -1663,40 +1663,50 @@ protected void connectConnectorAndReceivers() throws Exception
...
-                    // every exception has been logged above already
-                    // api needs refactoring to support the multi-cause exception
-                    // here
-                    throw errors.get(0);
+                    if (!errors.isEmpty())
+                    {
+                        // throw the first one in order not to break the reconnection
+                        // strategy logic,
+                        // every exception has been logged above already
+                        // api needs refactoring to support the multi-cause exception
```
pablokraan: api -> API

core/src/main/java/org/mule/transport/AbstractConnector.java
```diff
@@ -1663,40 +1663,50 @@ protected void connectConnectorAndReceivers() throws Exception
...
+                        // here
+                        throw errors.get(0);
+                    }
                 }
             }
         }
+        catch (Exception e)
+        {
+            // If an exception occurs while connection the receivers, disconnect the connector to
```
pablokraan: connection -> connecting

PR #3094, opened 2016-08-04
MULE-10112: file:list should fail when referencing non existing matcher

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/RegistryLookupValueResolver.java
```diff
@@ -47,12 +50,19 @@ public RegistryLookupValueResolver(String key)
...
+     * @throws ConfigurationException if no object is registered under {@link #key}
      */
     @Override
     public T resolve(MuleEvent event) throws MuleException
     {
-        return event.getMuleContext().getRegistry().get(key);
+        T value = event.getMuleContext().getRegistry().get(key);
+        if (value == null)
+        {
+            throw new ConfigurationException(MessageFactory.createStaticMessage(format("Element '%s' is not defined in the Mule Registry", key)));
```
elrodro83: static import

PR #3095, opened 2016-08-04
MULE-10193: HttpListener - MuleMessage cast exception when sending du…

pablokraan:
…plicate Content-Type header

_ DefaultMuleMessage expects to receive a String value for the content-type and enconding properties. When HTTP connector is used and there are multiple content-type headers in the request, the property value is a list of Strings. As we don't want to loose any value from the original request, in these cases, the content-type will be  obtained from the list, taking the first element.

### DIFF
--- a/core/src/main/java/org/mule/DefaultMuleMessage.java
+++ b/core/src/main/java/org/mule/DefaultMuleMessage.java
@@ -6,6 +6,7 @@
  */
 package org.mule;
 
+import static java.lang.String.format;
 import static org.mule.api.config.MuleProperties.CONTENT_TYPE_PROPERTY;
 import static org.mule.api.config.MuleProperties.MULE_ENCODING_PROPERTY;
 import static org.mule.api.config.MuleProperties.SYSTEM_PROPERTY_PREFIX;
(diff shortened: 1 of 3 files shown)


core/src/test/java/org/mule/DefaultMuleMessageTestCase.java
```diff
@@ -460,4 +467,75 @@ public void usesNullPayloadAsNull() throws Exception
...
         assertThat(message.getProperty(FOO_PROPERTY, PropertyScope.OUTBOUND), is(nullValue()));
     }
+
+    @Test
+    public void updatesEncodingPropertyWithStringCollection() throws Exception
+    {
+        MuleMessage message = createMuleMessage();
+        List<String> encodings = new ArrayList<>();
+        encodings.add(UTF_16.name());
+        encodings.add(UTF_16.name());
```
elrodro83: put a different value here to check that the first is used.

core/src/test/java/org/mule/DefaultMuleMessageTestCase.java
```diff
@@ -460,4 +467,75 @@ public void usesNullPayloadAsNull() throws Exception
...
+        message.setProperty(MULE_ENCODING_PROPERTY, new Object());
+    }
+
+    @Test
+    public void updatesContentTypePropertyWithStringCollection() throws Exception
+    {
+        MuleMessage message = createMuleMessage();
+        List<String> contentTypes = new ArrayList<>();
+        contentTypes.add("application/json; charset=UTF-8");
+        contentTypes.add("application/json; charset=UTF-8");
```
elrodro83: put a different value here to check that the first is used.

core/src/main/java/org/mule/DefaultMuleMessage.java
```diff
@@ -560,6 +564,55 @@ else if (MuleProperties.CONTENT_TYPE_PROPERTY.equalsIgnoreCase(key))
...
+
+        if (value instanceof String)
+        {
+           propertyValue = (String) value;
+        }
+        else if (value instanceof Collection)
+        {
+            if (((Collection) value).isEmpty())
+            {
+                throw new IllegalArgumentException(String.format("Unsupported value for '%s' property. Expected 'java.lang.String' but was an empty collection", propertyName));
```
elrodro83: static imports

PR #3096, opened 2016-08-04
MULE-10075: Support text parameters

alegmarra:


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/schema/builder/SchemaBuilder.java
```diff
@@ -811,4 +822,13 @@ TopLevelElement createTopLevelElement(String name, BigInteger minOccurs, String
         return element;
     }
 
+    private void generateTextElement(DslElementSyntax paramDsl, String description, Group all)
+    {
+        TopLevelElement textElement = createTopLevelElement(paramDsl.getElementName(), ZERO, "1");
```
marianogonzalez: shouldn't the ZERO be a 1 if the attribute is required?

alegmarra: child elements are never required, since passing it as an attribute is always an option

marianogonzalez: not sure this should be the case. Text parameter should always be a child element an never an attribute. It would be very confusing otherwise.

PR #3102, opened 2016-08-05
MULE-9826: server doesn't close connection when returning Maps in str…

alepulver:
…eaming (chunked) mode

### DIFF
--- a/modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseBuilder.java
+++ b/modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseBuilder.java
@@ -174,8 +174,13 @@ public class HttpResponseBuilder extends HttpMessageBuilder implements Initialis
                 }
                 else
                 {
-                    ByteArrayHttpEntity byteArrayHttpEntity = (ByteArrayHttpEntity) httpEntity;
-                    setupContentLengthEncoding(httpResponseHeaderBuilder, byteArrayHttpEntity.getContent().length);
+                    if (httpEntity instanceof EmptyHttpEntity)
+                    {
+                        setupContentLengthEncoding(httpResponseHeaderBuilder, 0);
+                    } else {
+                        ByteArrayHttpEntity byteArrayHttpEntity = (ByteArrayHttpEntity) httpEntity;
+                        setupContentLengthEncoding(httpResponseHeaderBuilder, byteArrayHttpEntity.getContent().length);
+                    }
                 }
             }
             else if (payload instanceof InputStream)
--- a/modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseBuilder.java
+++ b/modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseBuilder.java
@@ -177,7 +177,9 @@ public class HttpResponseBuilder extends HttpMessageBuilder implements Initialis
                     if (httpEntity instanceof EmptyHttpEntity)
                     {
                         setupContentLengthEncoding(httpResponseHeaderBuilder, 0);
-                    } else {
+                    }
+                    else
+                    {
                         ByteArrayHttpEntity byteArrayHttpEntity = (ByteArrayHttpEntity) httpEntity;
                         setupContentLengthEncoding(httpResponseHeaderBuilder, byteArrayHttpEntity.getContent().length);
                     }
--- a/modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseBuilder.java
+++ b/modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseBuilder.java
@@ -167,6 +167,16 @@ public class HttpResponseBuilder extends HttpMessageBuilder implements Initialis
                     warnMapPayloadButNoUrlEncodedContentType(httpResponseHeaderBuilder.getContentType());
                 }
                 httpEntity = createUrlEncodedEntity(event, (Map) payload);
+
+                if (responseStreaming == ALWAYS && supportsTransferEncoding(event))
+                {
+                    setupChunkedEncoding(httpResponseHeaderBuilder);
+                }
+                else
+                {
+                    ByteArrayHttpEntity byteArrayHttpEntity = (ByteArrayHttpEntity) httpEntity;
+                    setupContentLengthEncoding(httpResponseHeaderBuilder, byteArrayHttpEntity.getContent().length);
+                }
             }
             else if (payload instanceof InputStream)
             {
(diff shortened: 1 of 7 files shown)


modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerUrlEncodedTestCase.java
```diff
@@ -128,6 +133,27 @@ public void urlEncodedEmptyParamsUrlEncodedContentTypeGenerateANullPayload() thr
...
 
+    @Test
+    public void serverClosesConnectionAfterSendingData() throws Exception
+    {
+        // Apache Fluent doesn't fail while other clients such as curl, postman and this one do
+        AsyncHttpClientConfig asyncHttpClientConfig = getAsyncHttpClientConfig();
+        AsyncHttpClient asyncHttpClient = new AsyncHttpClient(new GrizzlyAsyncHttpProvider(asyncHttpClientConfig), asyncHttpClientConfig);
+        ListenableFuture<com.ning.http.client.Response> responseFuture = asyncHttpClient.
+                preparePost(getListenerUrl()).setBody("a=1&b=2").
+                addHeader("Content-Type", "application/x-www-form-urlencoded").execute();
```
afelisatti: There are constants already defined in HttpHeaders for both Strings.

modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerUrlEncodedTestCase.java
```diff
@@ -128,6 +133,27 @@ public void urlEncodedEmptyParamsUrlEncodedContentTypeGenerateANullPayload() thr
...
+        com.ning.http.client.Response response = responseFuture.get();
+
+        assertThat(response.getStatusCode(), is(200));
+    }
+
+    private AsyncHttpClientConfig getAsyncHttpClientConfig()
```
afelisatti: I don't think this is necessary. You can just inline it and get a simple config.

modules/http/src/test/resources/http-listener-response-streaming-config.xml
```diff
@@ -245,4 +246,21 @@
         <set-property propertyName="Content-Length" value="#[stringPayload.length()]"/>
     </flow>
 
+    <flow name="mapFlow">
+        <http:listener config-ref="listenerConfig" path="/map"/>
+        <set-property propertyName="Content-Type" value="application/x-www-form-urlencoded"/>
```
afelisatti: To make the 4.x cherry-pick process easier, I'd use a response builder with the Content-Type header instead of a set-property element which will no longer work.

modules/http/src/test/resources/http-listener-url-encoded-config.xml
```diff
@@ -3,6 +3,7 @@
       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
       xmlns:http="http://www.mulesoft.org/schema/mule/http"
       xmlns:vm="http://www.mulesoft.org/schema/mule/vm"
+      xmlns:doc="http://www.mulesoft.org/schema/mule/documentation"
```
afelisatti: I don't think this is needed.

modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseBuilder.java
```diff
@@ -167,6 +167,16 @@ else if (!configuredContentType.startsWith(HttpHeaders.Values.APPLICATION_X_WWW_
...
                 }
                 httpEntity = createUrlEncodedEntity(event, (Map) payload);
+
+                if (responseStreaming == ALWAYS && supportsTransferEncoding(event))
+                {
+                    setupChunkedEncoding(httpResponseHeaderBuilder);
+                }
+                else
+                {
+                    ByteArrayHttpEntity byteArrayHttpEntity = (ByteArrayHttpEntity) httpEntity;
```
afelisatti: Could you avoid this cast using an intermediate variable?

modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerUrlEncodedTestCase.java
```diff
@@ -137,23 +137,16 @@ public void urlEncodedEmptyParamsUrlEncodedContentTypeGenerateANullPayload() thr
     public void serverClosesConnectionAfterSendingData() throws Exception
     {
         // Apache Fluent doesn't fail while other clients such as curl, postman and this one do
-        AsyncHttpClientConfig asyncHttpClientConfig = getAsyncHttpClientConfig();
+        AsyncHttpClientConfig asyncHttpClientConfig = new AsyncHttpClientConfig.Builder().build();
         AsyncHttpClient asyncHttpClient = new AsyncHttpClient(new GrizzlyAsyncHttpProvider(asyncHttpClientConfig), asyncHttpClientConfig);
         ListenableFuture<com.ning.http.client.Response> responseFuture = asyncHttpClient.
                 preparePost(getListenerUrl()).setBody("a=1&b=2").
-                addHeader("Content-Type", "application/x-www-form-urlencoded").execute();
+                addHeader(HttpHeaders.Names.CONTENT_TYPE, HttpHeaders.Values.APPLICATION_X_WWW_FORM_URLENCODED).execute();
```
afelisatti: Static imports.

modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseBuilder.java
afelisatti: How are you handling here the case where the map is empty and createUrlEncodedEntity returns an EmptyHttpEntity?

alepulver: It goes through another if before, because it's a NullPayload.

afelisatti: Yes, but that is in the case where a form request is sent to the listener, in which case it propagates a NullPayload through the flow which then makes it back to be sent as a response. I think you need to consider the scenario where the payload changes to an empty map mid flow so that reaches the response builder.

modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseBuilder.java
```diff
@@ -167,6 +167,21 @@ else if (!configuredContentType.startsWith(HttpHeaders.Values.APPLICATION_X_WWW_
...
+                if (responseStreaming == ALWAYS && supportsTransferEncoding(event))
+                {
+                    setupChunkedEncoding(httpResponseHeaderBuilder);
+                }
+                else
+                {
+                    if (httpEntity instanceof EmptyHttpEntity)
+                    {
+                        setupContentLengthEncoding(httpResponseHeaderBuilder, 0);
+                    } else {
```
afelisatti: code format

PR #3105, opened 2016-08-08
MULE-10069: Artifact must export resource files instead of resource's…

pablokraan:
… folder paths

### DIFF
--- a/modules/artifact/src/main/java/org/mule/runtime/module/artifact/classloader/ArtifactClassLoaderFilter.java
+++ b/modules/artifact/src/main/java/org/mule/runtime/module/artifact/classloader/ArtifactClassLoaderFilter.java
@@ -33,7 +33,7 @@ public class ArtifactClassLoaderFilter implements ClassLoaderFilter
     public static final ArtifactClassLoaderFilter NULL_CLASSLOADER_FILTER = new ArtifactClassLoaderFilter(Collections.EMPTY_SET, Collections.EMPTY_SET);
 
     public static final String EXPORTED_CLASS_PACKAGES_PROPERTY = "artifact.export.classPackages";
-    public static final String EXPORTED_RESOURCE_PACKAGES_PROPERTY = "artifact.export.resourcePackages";
+    public static final String EXPORTED_RESOURCE_PROPERTY = "artifact.export.resources";
 
     private static final char PACKAGE_SEPARATOR = '.';
     private static final String EMPTY_PACKAGE = "";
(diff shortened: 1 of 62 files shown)


modules/artifact/src/main/java/org/mule/runtime/module/artifact/classloader/ArtifactClassLoaderFilter.java
```diff
@@ -40,21 +40,38 @@
...
-        this.exportedResourcePackages = unmodifiableSet(exportedResourcePackages);
+    private Set<String> sanitizeClassPackageS(Set<String> exportedClassPackages)
+    {
+        return exportedClassPackages.stream().map(this::sanitizePackageName).collect(Collectors.toSet());
+    }
+
+    private String sanitizePackageName(String exportedClassPackage)
+    {
+        exportedClassPackage = exportedClassPackage.trim();
+        exportedClassPackage = exportedClassPackage.endsWith(".") ? exportedClassPackage.substring(0, exportedClassPackage.length()-1) : exportedClassPackage;
```
gsfernandes: Whitespace, length() - 1

modules/artifact/src/main/java/org/mule/runtime/module/artifact/classloader/ArtifactClassLoaderFilter.java
```diff
@@ -70,9 +87,9 @@ public boolean exportsClass(String className)
     public boolean exportsResource(String name)
     {
         checkArgument(name != null, "Resource name cannot be null");
-        final String resourcePackage = getResourceFolder(name);
+        final String sanitizeResourceName = sanitizeResourceName(name);
 
-        return exportedResourcePackages.contains(resourcePackage);
+        return exportedResources.contains(sanitizeResourceName);
```
gsfernandes: Wouldn't you need to change all the mule-module.properties to start exporting files instead of folders? For instance, I'm looking at core and it is exporting:
```
artifact.export.resourcePackages=/,\
                                /META-INF,\
                                /META-INF/services,\
                                /META-INF/services/org/mule/runtime/core/config,\
                                /META-INF/services/org/mule/runtime/core/i18n,\
                                /META-INF/services/org/mule/runtime/core/models,\
                                /META-INF/services/org/mule/runtime/transport
```
Perhaps I' m missing something here but it should be explicitly exported the file if not it won't return true here. Right?

pablokraan: container supports exporting folders and files, changing what is exported in every module is not part of this PR, but it will be part of future changes once the API is clearly defined

PR #3107, opened 2016-08-09
MULE-9072 Cleanup MuleSession. Move session handlers to compatibility…

dfeist:
… module.


core/src/main/java/org/mule/runtime/core/VoidMuleEvent.java
```diff
@@ -273,4 +274,16 @@ public ProcessorsTrace getProcessorsTrace()
...
+    @Override
+    public SecurityContext getSecurityContext()
+    {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void setSecurityContext(SecurityContext context)
+    {
+
```
elrodro83: shouldn't this throw UnsupporteOperationException too?

transports/core-transports-tests/src/test/java/org/mule/compatibility/core/session/MuleSessionHandlerTestCase.java
```diff
@@ -31,7 +32,7 @@
 import org.mule.runtime.core.security.DefaultSecurityContextFactory;
 import org.mule.runtime.core.security.MuleCredentials;
 import org.mule.runtime.core.session.DefaultMuleSession;
-import org.mule.runtime.core.session.SerializeAndEncodeSessionHandler;
+;
```
elrodro83: remove this

PR #3117, opened 2016-08-10
MULE-10230: Allow SQL variable assignment operator

MaxOcklind:
https://www.mulesoft.org/jira/browse/MULE-10230


pablokraan:
Hi @MaxOcklind, thanks for sending this PR. 
Can you add tests to ensure your change works as expected on SimpleQueryTemplateParserTestCase?

marianogonzalez:
shouldn't this PR point to the 3.x branch?

MaxOcklind:
Hi @pablokraan, I have now added a test case.

Do you agree with @marianogonzalez that this PR should be pointed at the 3.x branch? And in that case I'm guessing that I need to checkout that branch, make the changes again and create a new PR? (I'm new to git and github so expect a stupid question or two)

pablokraan:
@MaxOcklind yes, the PR against mule-3.x makes more sense. The changes are probably to be the same but the class you changed is going to be in a different package. Don't forget to check the code format before pushing the changes.

pablokraan:
I'm closing this PR as there is another one created on mule-3.x branch

PR #3119, opened 2016-08-11
MULE-10218: Exclusive optionals should support nested groups

pablomusumeci:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ParameterGroupArgumentResolver.java
```diff
@@ -72,7 +73,7 @@ public T resolve(OperationContext operationContext) {
       }
 
       for (ParameterGroupArgumentResolver<?> childResolver : childResolvers) {
-        childResolver.resolve(operationContext);
+        ((Field) childResolver.getContainer()).set(parameterGroup, childResolver.resolve(operationContext));
```
pablomusumeci: Not glad about this cast but the thing here is that when you are dealing with a "child resolver" (aka nested resolver), you know that the container is going to be a Field (no chance of being a Parameter).

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/ExclusiveParameterModelValidator.java
```diff
@@ -106,6 +109,9 @@ protected void defaultVisit(MetadataType metadataType) {
 
           @Override
           public void visitSimpleType(SimpleType simpleType) {}
+
+          @Override
+          public void visitObject(ObjectType objectType) {}
```
marianogonzalez: why do you define this if it doesn nothing. Same ot the `visitSimpleType` above

pablomusumeci: The idea was throwing an IllegalModelDefinitionException for all the different MetadataTypes except for these 2 situations.

PR #3122, opened 2016-08-11
MULE-10115: Review how correlation is implemented/achieved

elrodro83:
* Introduce MessageExecutionContext
* Add Correlation to Event
* Remove correlation from message
* Remove enableCorrelation flag
* Introduce MuleCompatibilityMessage


core/src/main/java/org/mule/runtime/core/DefaultMessageExecutionContext.java
```diff
@@ -0,0 +1,63 @@
...
+import java.io.Serializable;
+import java.util.Date;
+import java.util.Optional;
+
+/**
+ * Default immutable implementation of {@link MessageExecutionContext}.
+ *
+ * @since 4.0
+ */
+public class DefaultMessageExecutionContext implements MessageExecutionContext, Serializable {
```
dfeist: Can this be final?

core/src/main/java/org/mule/runtime/core/api/MessageExecutionContext.java
```diff
@@ -0,0 +1,36 @@
...
+  /**
+   * @return the unique id that identifies all {@link MuleEvent}s of the same context.
+   */
+  String getId();
+
+  /**
+   * @return the correlation id that was set by the {@link MessageSource} for the first {@link MuleEvent} of this context, if
+   *         available.
+   */
+  Optional<String> getSourceCorrelationId();
```
dfeist: I'd use `getCorrelationId()` and in implementation class document that if context was created without an explicit correlationId then `id` is used.

core/src/main/java/org/mule/runtime/core/api/MessageExecutionContext.java
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * @return a timestamp indicating when the message was received by the {@link MessageSource}.
+   */
+  Date getReceivedDate();
+}
```
dfeist: We need to determine what else goes here.  If this isn't going to have muleContext/flowConstruct then we need  server/app and flow identifiers ideally, unless the flow identifier we use composing server-app-flow id's.

If we have flow app, then maybe there isn't need for source name/uri, but we should check what notification requirements are, because it wouldn't be ideally to have to lookup uri from flow name in runtime.

core/src/main/java/org/mule/runtime/core/api/MessageExecutionContext.java
```diff
@@ -0,0 +1,36 @@
...
+  /**
+   * @return the correlation id that was set by the {@link MessageSource} for the first {@link MuleEvent} of this context, if
+   *         available.
+   */
+  Optional<String> getSourceCorrelationId();
+
+  /**
+   * @return a timestamp indicating when the message was received by the {@link MessageSource}.
+   */
+  Date getReceivedDate();
```
dfeist: Use `java.time.OffsetTime`.  I would call this method `getRecievedTime`.

core/src/main/java/org/mule/runtime/core/api/MuleEvent.java
```diff
@@ -40,6 +42,40 @@
...
+   * in the message.
+   *
+   * @return the id for this message or null value if one hasn't been set
+   */
+  String getCorrelationId();
+
+  /**
+   * @return {@code true} if the source of this event has set a correlation id (see {@link #getCorrelationId()}).
+   */
+  boolean hasSourceCorrelation();
```
dfeist: I don't think this is needed, from event perspective it should be transparent if source set a custom id or not.

core/src/main/java/org/mule/runtime/core/DefaultMessageExecutionContext.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  /**
+   * Builds a new execution context with the given parameters.
+   * 
+   * @param id the unique id that identifies all {@link MuleEvent}s of the same context.
+   * @param sourceCorrelationId the correlation id that was set by the {@link MessageSource} for the first {@link MuleEvent} of
+   *        this context, if available.
+   */
+  public DefaultMessageExecutionContext(String id, String sourceCorrelationId) {
```
dfeist: Id should be UUID not exposed in constructor i thought.  Also defining a separate/user correlation id should be optional and have it's own constructor IMO.

core/src/main/java/org/mule/runtime/core/DefaultMuleEvent.java
```diff
@@ -990,6 +958,37 @@ public void setSecurityContext(SecurityContext context) {
...
+  }
+
+  @Override
+  public Correlation getCorrelation() {
+    return correlation;
+  }
+
+  @Override
+  public String getCorrelationId() {
+    return getCorrelation().getId().orElse(getExecutionContext().getSourceCorrelationId().orElse(getMessage().getUniqueId()));
```
dfeist: Event shouldn't care about context.id vs context.correlationId.  This logic should take into account parent event shouldn't it?

core/src/main/java/org/mule/runtime/core/message/Correlation.java
```diff
@@ -37,21 +37,16 @@
    * @param groupSize see {@link #getGroupSize()}.
    * @param sequence see {@link #getSequence()}.
    */
-  Correlation(String id, Integer groupSize, Integer sequence) {
+  public Correlation(String id, Integer groupSize, Integer sequence) {
```
dfeist: Can you move class instead of change access?

core/src/main/java/org/mule/runtime/core/message/Correlation.java
```diff
@@ -37,21 +37,16 @@
...
-   * The id is associated with the message using the underlying transport protocol. As such not all messages will support the
-   * notion of a id i.e. tcp or file. In this situation the correlation Id is set as a property of the message where it's up to
-   * developer to keep the association with the message. For example if the message is serialised to xml the id will be available
-   * in the message.
+   * If present, this id will take precedence over any correlationId already set for the event.
    *
-   * @return the id for this message or null value if one hasn't been set
+   * @return the correlation id for this event or null value if one hasn't been set
    */
   public Optional<String> getId() {
```
dfeist: I'm not sure why this would be null.

core/src/main/java/org/mule/runtime/core/routing/filters/ExpressionFilter.java
```diff
@@ -80,7 +84,8 @@ public boolean accept(MuleMessage message) {
     }
 
     // TODO MULE-9341 Remove Filters. Expression filter will be replaced by something that uses MuleEvent.
-    return accept(new DefaultMuleEvent(message, MessageExchangePattern.ONE_WAY, new Flow("", muleContext)));
+    return accept(new DefaultMuleEvent(new DefaultMessageExecutionContext(muleContext.getUniqueIdString(), null), message,
```
dfeist: Why are you using app id for execution context id?

core/src/main/java/org/mule/runtime/core/message/Correlation.java
elrodro83: This is being called from more places now

core/src/main/java/org/mule/runtime/core/DefaultMuleEvent.java
elrodro83: Yes, implementing the parent-child is something i have to do yet

core/src/main/java/org/mule/runtime/core/message/Correlation.java
elrodro83: This will only be set by the compatibility code, otherwise should always be null

core/src/main/java/org/mule/runtime/core/routing/filters/ExpressionFilter.java
elrodro83: it is an uuid

core/src/main/java/org/mule/runtime/core/DefaultMessageExecutionContext.java
```diff
@@ -6,58 +6,110 @@
...
   private static final long serialVersionUID = -3664490832964509653L;
 
+  /**
+   * Builds a new execution context with the given parameters.
+   * 
+   * @param context the {@link MuleContext} of the application that is building this context.
+   * @param flow the flow that processes events of this context. 
+   * @param correlationId See {@link MessageExecutionContext#getCorrelationId()}.
+   */
+  public static MessageExecutionContext buildContext(MuleContext context, FlowConstruct flow) {
```
dfeist: FlowConstruct has MuleContext already doesn't it?  I think `create` works better here as this isn't a builder.

core/src/main/java/org/mule/runtime/core/DefaultMuleEvent.java
```diff
@@ -981,12 +980,7 @@ public Correlation getCorrelation() {
 
   @Override
   public String getCorrelationId() {
-    return getCorrelation().getId().orElse(getExecutionContext().getSourceCorrelationId().orElse(getMessage().getUniqueId()));
-  }
-
-  @Override
-  public boolean hasSourceCorrelation() {
-    return getExecutionContext().getSourceCorrelationId().isPresent();
+    return getCorrelation().getId().orElse(getExecutionContext().getCorrelationId().orElse(getMessage().getUniqueId()));
```
dfeist: If this logic needs to be here then fine, but ideally we'd encapsulate some of this inside Correlation object wouldn't we if possible?

core/src/main/java/org/mule/runtime/core/el/context/MessageContext.java
```diff
@@ -52,6 +54,10 @@ public String getRootId() {
     return event.getMessage().getMessageRootId();
   }
 
+  public Correlation getCorrelation() {
```
dfeist: Do we need to add this give MEL will be deprecated?

core/src/main/java/org/mule/runtime/core/DefaultMessageExecutionContext.java
```diff
@@ -0,0 +1,108 @@
...
+ */
+public final class DefaultMessageExecutionContext implements MessageExecutionContext, Serializable {
+
+  private static final long serialVersionUID = -3664490832964509653L;
+
+  /**
+   * Builds a new execution context with the given parameters.
+   * @param flow the flow that processes events of this context. 
+   */
+  public static MessageExecutionContext createContext(FlowConstruct flow) {
```
dfeist: Just "Create"? Because this can't be used with a static import as it is because it's not clear if it creates a MuleContext or ExceutionContext etc.

core/src/main/java/org/mule/runtime/core/DefaultMessageExecutionContext.java
```diff
@@ -0,0 +1,108 @@
...
+   */
+  public static MessageExecutionContext createContext(FlowConstruct flow, String correlationId) {
+    DefaultMessageExecutionContext executionContext =
+        new DefaultMessageExecutionContext(flow.getMuleContext().getUniqueIdString(), correlationId);
+    executionContext.serverId = generateId(flow.getMuleContext());
+    executionContext.flowName = flow.getName();
+    return executionContext;
+  }
+
+  private static String generateId(MuleContext context) {
```
dfeist: Why is server id being created here?

core/src/main/java/org/mule/runtime/core/DefaultMuleEvent.java
```diff
@@ -990,6 +905,35 @@ public void setSecurityContext(SecurityContext context) {
...
+  }
+
+  @Override
+  public Correlation getCorrelation() {
+    return correlation;
+  }
+
+  @Override
+  public String getCorrelationId() {
+    return getCorrelation().getId().orElse(getExecutionContext().getCorrelationId().orElse(getMessage().getUniqueId())
```
dfeist: For performance this should be fixed value defined when event is build.  But this can be done along with work on EventBuilder.

PR #3123, opened 2016-08-11
MULE-10235: Move services from launcher to a new services module

gsfernandes:
_mule-module-service module created
_code moved from mule-module-launcher, only left the code related to the launcher process
_added new module to distribution


distributions/standalone/assembly-whitelist.txt
```diff
@@ -139,6 +139,7 @@
 /mule-standalone-${productVersion}/lib/mule/mule-module-jbossts-${productVersion}.jar
 /mule-standalone-${productVersion}/lib/mule/mule-module-json-${productVersion}.jar
 /mule-standalone-${productVersion}/lib/mule/mule-module-launcher-${productVersion}.jar
+/mule-standalone-${productVersion}/lib/mule/mule-module-services-${productVersion}.jar
```
pablokraan: I think it should be mule-module-service

modules/services/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+
+        <dependency>
+            <groupId>org.mule.modules</groupId>
+            <artifactId>mule-module-artifact</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.mule.tests</groupId>
+            <artifactId>mule-tests-functional</artifactId>
```
pablokraan: Is this dependency needed?

modules/launcher/src/main/java/org/mule/runtime/module/launcher/service/FileSystemServiceProviderDiscoverer.java
```diff
@@ -56,6 +58,11 @@ public FileSystemServiceProviderDiscoverer(ArtifactClassLoader containerClassLoa
 
     final List<ServiceDescriptor> serviceDescriptors = new LinkedList<>();
 
+    File servicesFolder = getServicesFolder();
```
pablokraan: Not sure that this class must be on launcher module. Looks like the only reason is because of this dependency on MuleFoldersUtil.

gsfernandes:
jenkins test this

PR #3131, opened 2016-08-16
MULE-10254: @ParameterGroup doesn't work on abstract pojos

marianogonzalez:


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/schema/builder/ObjectTypeSchemaDelegate.java
```diff
@@ -179,7 +180,10 @@ private ComplexType registerPojoAsTypeWithBase(ObjectType metadataType, ObjectTy
     } else {
       DslElementSyntax baseDsl = builder.getDslResolver().resolve(baseType);
       base = new QName(baseDsl.getNamespaceUri(), getBaseTypeName(baseType), baseDsl.getNamespace());
-      fields = metadataType.getFields().stream().filter(f -> !baseType.getFields().contains(f)).collect(toList());
+      fields = metadataType.getFields().stream()
+          .filter(field -> !baseType.getFields().stream()
+              .anyMatch(other -> other.getKey().getName().getLocalPart().equals(field.getKey().getName().getLocalPart())))
```
alegmarra: `other.getKey().getName().getLocalPart().equals(field.getKey().getName().getLocalPart()))` => `MetadataTypeUtils` ?

marianogonzalez: I can hardly see this been reused anywhere else. No point in putting it in an utils if it only has one usage

PR #3135, opened 2016-08-16
MULE-10242: Allow an initial pipeline ID, when obtaining a dynamic pi…

balbifm:
…peline from a flow, to be proposed by the client


aiannucci:
@svacas does this look OK to you?

svacas:
LGTM

PR #3138, opened 2016-08-17
MULE-10248 - Migrate components to use new parsing mechanism

pablolagreca:


modules/spring-config/src/test/resources/mule-config.xml
```diff
@@ -5,9 +5,7 @@
 
     <flow name="service">
         <component>
-            <spring-object bean="orange">
-                <property key="brand" value="Tutti Frutti"/>
-            </spring-object>
+            <spring-object bean="orange"/>
```
afelisatti: Why was it necessary to delete this part of the test?

PR #3140, opened 2016-08-17
MULE-10268: Proxy Authentication Header is not included when using pr…

gsfernandes:
…oxy for HTTPS target server - Adding test

### DIFF
--- a/modules/http/pom.xml
+++ b/modules/http/pom.xml
@@ -126,6 +126,11 @@
         </dependency>
         <dependency>
             <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-proxy</artifactId>
+            <scope>test</scope>
+        </dependency>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -412,6 +412,11 @@
             </dependency>
             <dependency>
                 <groupId>org.eclipse.jetty</groupId>
+                <artifactId>jetty-proxy</artifactId>
```
afelisatti: I don't think this should go here.

modules/http/pom.xml
```diff
@@ -126,6 +126,11 @@
         </dependency>
         <dependency>
             <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-proxy</artifactId>
```
afelisatti: Add the version here.

PR #3143, opened 2016-08-18
MULE-10214: @Import should not require a class element

alegmarra:


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/schema/ClasspathBasedDslContext.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+
+  private void findExtensionsInClasspath() {
+
+    // default classpath for testing environment, were the extension classloader is not useful
+    Set<Class<?>> annotated = getExtensionTypes(forJavaClassPath());
+
+    if (annotated.isEmpty()) {
+      // we are building an standalone extension, thus the classloader of the extension is required
```
marianogonzalez: i would prefer to do this search first and do the testing one in a second phase. But let me know if this is not possible.

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/DefaultDslContext.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.util.Optional;
+
+/**
+ * Default implementation of {@link DslResolvingContext} that uses
+ * the {@link ExtensionManager} to provide the required {@link ExtensionModel}s
+ *
+ * @since 4.0
+ */
+class DefaultDslContext implements DslResolvingContext {
```
marianogonzalez: just to clarify, this is the one that will be used when an application is being deployed on a real mule right?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/ExtensionEntityKey.java
```diff
@@ -6,7 +6,7 @@
  */
 package org.mule.runtime.module.extension.internal.manager;
 
-import org.mule.runtime.extension.api.annotation.Extension;
+import static com.sun.imageio.plugins.jpeg.JPEG.vendor;
```
marianogonzalez: wtf?

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/DefaultDslContext.java
alegmarra: Yes, it's the one used by the `ExtensionBuildingDefinitionProvider`.
Other context are: 
- ClasspathBased, used during schema generation at compile time
- SchemaTest, used for testing only

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/schema/ClasspathBasedDslContext.java
alegmarra: Having the test classpath first and falling back to this case adds a penalty of >100ms during the compilation of the Http extension, while avoiding a 300 lines warning trace for each test that requires this classpath.

PR #3149, opened 2016-08-18
MULE-10284: @Text has no effect when applied to a parameter group in …

marianogonzalez:
…abstract class


elrodro83:
jenkins validate

PR #3155, opened 2016-08-19
MULE-10297: Remove event.getMuleContext() dependency from flow

elrodro83:
statistics

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/DefaultMessageExecutionContext.java
+++ b/core/src/main/java/org/mule/runtime/core/DefaultMessageExecutionContext.java
@@ -12,6 +12,7 @@ import org.mule.runtime.core.api.MessageExecutionContext;
 import org.mule.runtime.core.api.MuleEvent;
 import org.mule.runtime.core.api.construct.FlowConstruct;
 import org.mule.runtime.core.api.source.MessageSource;
+import org.mule.runtime.core.management.stats.ProcessingTime;
 
 import java.io.Serializable;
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/mule/runtime/core/DefaultMessageExecutionContext.java
```diff
@@ -53,6 +56,8 @@ public static MessageExecutionContext create(FlowConstruct flow, String correlat
   private String serverId;
   private String flowName;
 
+  private ProcessingTime processingTime;
```
dfeist: final?

core/src/main/java/org/mule/runtime/core/interceptor/AbstractEnvelopeInterceptor.java
```diff
@@ -49,7 +47,7 @@ public abstract MuleEvent last(MuleEvent event, ProcessingTime time, long startT
   @Override
   protected MuleEvent processBlocking(MuleEvent event) throws MuleException {
     long startTime = System.currentTimeMillis();
-    ProcessingTime time = event.getProcessingTime();
+    ProcessingTime time = ((DefaultMessageExecutionContext) event.getExecutionContext()).getProcessingTime();
```
dfeist: We really need to find a cleaner way to avoid this cast (we've been casting it in compatibility module, but shouldn't in core).  If it makes sense to do this later and not now then need to add a TODO with JIRA.

PR #3156, opened 2016-08-19
MULE-10214: @Import should not require a class element - Fix

alegmarra:


distributions/standalone/assembly-whitelist.txt
```diff
@@ -310,6 +310,7 @@
 /mule-standalone-${productVersion}/lib/opt/plexus-utils-3.0.22.jar
 /mule-standalone-${productVersion}/lib/opt/quartz-2.2.1.jar
 /mule-standalone-${productVersion}/lib/opt/reflections-0.9.10.jar
+/mule-standalone-${productVersion}/lib/opt/javassist-3.7.ga.jar
```
marianogonzalez: alpha sort

PR #3150, opened 2016-08-19
MULE-10306: Add option to disable internal entity expansion in XML (l…

alepulver:
…eads to DoS)

PR #3161, opened 2016-08-22
MULE-10306: Add option to disable internal entity expansion in XML (l…

alepulver:
…eads to DoS) (#3150)

### DIFF
--- a/distributions/standalone/src/main/resources/MIGRATION.txt
+++ b/distributions/standalone/src/main/resources/MIGRATION.txt
@@ -8,6 +8,7 @@ Migration changes from 3.8.x to 3.9.x
 MULE-9931: Transaction log files size are now restricted by size, using a default size of 500 mb. The configured size is just an approximated value which may be exceeded based on the size of the record store by the transaction. This is configurable using the attribute queueTransactionFilesSize of the <configuration> element.
     The size restriction applies to the set of transaction log files for local transactions and XA transactions independently meaning that if both types of transactions are used for queue the set of tx files will use up to 1 gb.
 MULE-10100: Processing after a synchronous until successful resulting in a VoidMuleEvent will now continue with the original event.
+MULE-10306: XML entity expansion is now disabled by default because it allows DoS attacks. To restore previous behaviour use the expandInternalEntities="true" attribute in the xslt transformer.
 
 Migration changes from 3.7.x to 3.8.x
 MULE-9368: System property mule.sftp.knownHostsFile is now removed. Instead, the file with the known hosts must now be provided through the mule xml config file, in the knownHostsFile attribute of the connector or the endpoints.
--- a/modules/xml/src/main/java/org/mule/module/xml/transformer/AbstractXmlTransformer.java
+++ b/modules/xml/src/main/java/org/mule/module/xml/transformer/AbstractXmlTransformer.java
@@ -6,6 +6,8 @@
  */
 package org.mule.module.xml.transformer;
 
+import static javax.xml.stream.XMLInputFactory.IS_REPLACING_ENTITY_REFERENCES;
+import static javax.xml.stream.XMLInputFactory.IS_SUPPORTING_EXTERNAL_ENTITIES;
 import org.mule.api.MuleRuntimeException;
 import org.mule.api.lifecycle.Initialisable;
 import org.mule.api.lifecycle.InitialisationException;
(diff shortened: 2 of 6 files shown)


modules/xml/src/main/java/org/mule/module/xml/transformer/AbstractXmlTransformer.java
```diff
@@ -42,11 +44,13 @@
  */
 public abstract class AbstractXmlTransformer extends AbstractMessageTransformer implements Initialisable
 {
+
     private String outputEncoding;
     private XMLInputFactory xmlInputFactory;
     private XMLOutputFactory xmlOutputFactory;
     private boolean useStaxSource = false;
     private boolean acceptExternalEntities = false;
+    private boolean expandInternalEntities = true;
```
afelisatti: Shouldn't this be false by default to prevent the security issue?

modules/xml/src/main/java/org/mule/module/xml/transformer/AbstractXmlTransformer.java
```diff
@@ -373,4 +376,14 @@ public boolean getAcceptExternalEntities()
     {
         return this.acceptExternalEntities;
     }
+
+    public boolean getExpandInternalEntities()
```
afelisatti: Is this getter necessary?

modules/xml/src/main/resources/META-INF/mule-xml.xsd
```diff
@@ -232,6 +232,15 @@ The Jaxen filter allows you to route messages based on XPath expressions. The Ja
                         </xsd:documentation>
                     </xsd:annotation>
                 </xsd:attribute>
+                <xsd:attribute name="expandInternalEntities" type="mule:substitutableBoolean" default="true">
+                    <xsd:annotation>
+                        <xsd:documentation>
+                            Whether to expand internal entity references in the XML.
+                            Default value is true. Keeping this value will make your application
```
afelisatti: Again, I think we should by default NOT be vulnerable and inform here of the problem with changing the value to true. Since this is a change in behaviour from 3.8 you also need to add an entry to the migration guide (MIGRATION.txt).

modules/xml/src/test/java/org/mule/transformers/xml/xslt/XsltTransformerBLTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * This test case validates that the XSLT transformer is not vulnerable to
+ * Billion Laughs attack when internal entity expansion is disabled
+ * <p>
+ * <b>EIP Reference:</b> <a
+ * href="https://en.wikipedia.org/wiki/Billion_laughs"<a/>
+ * </p>
+ */
+public class XsltTransformerBLTest extends FunctionalTestCase
```
afelisatti: Couldn't we have just one test now that there's no system property involved?

modules/xml/src/main/java/org/mule/module/xml/transformer/AbstractXmlTransformer.java
alepulver: No, it was removed.

distributions/standalone/src/main/resources/MIGRATION.txt
```diff
@@ -8,6 +8,7 @@ Migration changes from 3.8.x to 3.9.x
 MULE-9931: Transaction log files size are now restricted by size, using a default size of 500 mb. The configured size is just an approximated value which may be exceeded based on the size of the record store by the transaction. This is configurable using the attribute queueTransactionFilesSize of the <configuration> element.
     The size restriction applies to the set of transaction log files for local transactions and XA transactions independently meaning that if both types of transactions are used for queue the set of tx files will use up to 1 gb.
 MULE-10100: Processing after a synchronous until successful resulting in a VoidMuleEvent will now continue with the original event.
+MULE-10306: Disable XML internal entity expansion by default because if allows DoS attacks. To restore previous behaviour use the expandInternalEntities="true" attribute in the xslt transformer.
```
afelisatti: it allows?

afelisatti: I'd recommend following the "is now" standard to explain the changes, something like: XML internal entity expansion is now disabled by default because...

modules/xml/src/main/java/org/mule/module/xml/transformer/AbstractXmlTransformer.java
afelisatti: It's still showing.

modules/xml/src/main/resources/META-INF/mule-xml.xsd
```diff
@@ -227,11 +227,21 @@ The Jaxen filter allows you to route messages based on XPath expressions. The Ja
...
                             Default value is false. Setting this value to true will make your application
                             vulnerable to XXE attacks. Use with extreme care.
                         </xsd:documentation>
                     </xsd:annotation>
                 </xsd:attribute>
+                <xsd:attribute name="expandInternalEntities" type="mule:substitutableBoolean" default="false">
+                    <xsd:annotation>
+                        <xsd:documentation>
+                            Whether to expand internal entity references in the XML body.
+                            Default value is false. Setting this value yo true will make your application
```
afelisatti: to

PR #3163, opened 2016-08-22
MULE-10219: Support DDL operations in DB extension

marianogonzalez:


extensions/db/src/main/java/org/mule/extension/db/internal/operation/BaseDbOperations.java
```diff
@@ -0,0 +1,134 @@
...
+                               QueryType... validTypes) {
+    final Query resolvedQuery = storedProcedureResolver.resolve(call, connector, connection);
+    validateQueryType(resolvedQuery.getQueryTemplate(), asList(validTypes));
+
+    return resolvedQuery;
+  }
+
+  protected void validateQueryType(QueryTemplate queryTemplate, List<QueryType> validTypes) {
+    if (validTypes == null || !validTypes.contains(queryTemplate.getType())) {
+      throw new IllegalArgumentException(format("Query type must be one of [%s] but query '%s' is of type",
```
elrodro83: missing %s at the end

extensions/db/src/test/java/org/mule/extension/db/integration/ddl/ExecuteDdlTestCase.java
```diff
@@ -0,0 +1,77 @@
...
+    try {
+      DataSource dataSource = getDefaultDataSource();
+      connection = dataSource.getConnection();
+
+      QueryRunner qr = new QueryRunner(dataSource);
+      qr.update(connection, "DROP TABLE TestDdl");
+    } catch (SQLException e) {
+      // Ignore: table does not exist
+    } finally {
+      if (connection != null) {
```
elrodro83: why not use try-with-resources?

marianogonzalez: because this is kraanean code coming from 3.5.x. Changed it though

extensions/db/src/main/java/org/mule/extension/db/internal/operation/BaseDbOperations.java
marianogonzalez: nice catch. fixed

PR #3164, opened 2016-08-22
MULE-10298: Deploy applications in parallel

pablokraan:
_ Removing lazy initialisation from ApplicationStatusMapper
_ Synchronising access to LoggerContextCache#getLoggerContext
_ Synchronising ObservableList
_ Extracting isUpdatedZombieArtifact from ArchiveDeployer#deployExplodedApp so it can be executed without having to run a new thread
_ Removing deployment lock from DefaultArchiveDeployer so locking is managed form DeploymentService
_ Moving MuleContainer#initArgs before creation of deployment service so the StartupContext can be used on the MuleDeploymentService constructor
_ Updating MuleDeploymentService to create a sequential or parallel deploymentDirectoryWatcher depending on the configuration
_ Adding ParallelDeploymentDirectoryWatcher to manage parallel deployment
_ Making DeploymentServiceTestCase parameterised to run sequential/parallel deployments
_ Adding deploysMultipleAppsZipOnStartup
_ Fixing test flakiness and bugs (MULE-10094 and a couple of non reported ones)

### DIFF
--- a/modules/launcher/src/main/java/org/mule/module/launcher/ArchiveDeployer.java
+++ b/modules/launcher/src/main/java/org/mule/module/launcher/ArchiveDeployer.java
@@ -26,7 +26,7 @@ public interface ArchiveDeployer<T extends Artifact>
     T deployExplodedArtifact(String artifactDir) throws DeploymentException;
 
     /**
-     * Indicates if a previously failed artifact (zombie) was updated.
+     * Indicates if a previously failed artifact (zombie) configuration was updated on the file system.
      *
      * @param artifactName name of the artifact to check. Non empty.
      * @return true if the zombie artifact was updated, false it the artifact is not a zombie or it was not updated.
--- a/modules/launcher/src/main/java/org/mule/module/launcher/ArchiveDeployer.java
+++ b/modules/launcher/src/main/java/org/mule/module/launcher/ArchiveDeployer.java
@@ -6,7 +6,6 @@
  */
 package org.mule.module.launcher;
 
-import org.mule.module.launcher.application.Application;
 import org.mule.module.launcher.artifact.Artifact;
 import org.mule.module.launcher.artifact.ArtifactFactory;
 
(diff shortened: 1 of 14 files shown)


modules/launcher/src/main/java/org/mule/module/launcher/ArchiveDeployer.java
```diff
@@ -26,6 +25,14 @@
 
     T deployExplodedArtifact(String artifactDir) throws DeploymentException;
 
+    /**
+     * Indicates if a previously failed artifact (zombie) was updated.
```
pablolagreca: what do you mean by "updated" can it be better explained here?

modules/launcher/src/main/java/org/mule/module/launcher/DefaultArchiveDeployer.java
```diff
@@ -550,6 +468,7 @@ public void redeploy(T artifact) throws DeploymentException
             try
             {
                 logDeploymentFailure(t, artifact.getArtifactName());
+                addZombieApp(artifact);
```
pablolagreca: Is this removed from the zombie map after redeploy is done?

modules/launcher/src/main/java/org/mule/module/launcher/ParallelDeploymentDirectoryWatcher.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * Provides parallel deployment of Mule applications.
```
pablolagreca: since?

modules/launcher/src/test/java/org/mule/module/launcher/DeploymentServiceTestCase.java
```diff
@@ -1846,6 +1909,12 @@ public void undeploysDomainAndDomainsAppsRemovingAnchorFile() throws Exception
         doDomainUndeployAndVerifyAppsAreUndeployed(createUndeployDummyDomainAction());
     }
 
+    @Override
+    public int getTestTimeoutSecs()
```
pablolagreca: remove

modules/launcher/src/main/java/org/mule/module/launcher/DefaultArchiveDeployer.java
pablokraan: Here the artifact is added as zombie when the redeployment has failed. It will be removed from the zombie map on the next successful redeployment.

modules/launcher/src/main/java/org/mule/module/launcher/ParallelDeploymentDirectoryWatcher.java
pablokraan: What is version must be used on @since? This change has to be backported to some maintenance branches

PR #3171, opened 2016-08-23
MULE-10038: Improve System tests framework.     Reordering and removing imports

asequeira:


tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleDeployment.java
```diff
@@ -12,24 +12,24 @@
...
-import java.util.stream.Stream;
 
 import org.apache.commons.io.FilenameUtils;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
+
+import org.mule.tck.probe.JUnitProbe;
+import org.mule.tck.probe.PollingProber;
+import org.mule.test.infrastructure.process.MuleProcessController;
+
```
aiannucci: Mule imports should go above all others and just below static imports

PR #3175, opened 2016-08-24
MULE-10331: Make data source parameters exclusive in new DB connector

pablomusumeci:


marianogonzalez:
add a test to actually verify this in DB

PR #3176, opened 2016-08-24
MULE-10299: Remove MuleEvent dependency on FlowConstruct.

elrodro83:


core/src/main/java/org/mule/runtime/core/api/processor/LoggerMessageProcessor.java
```diff
@@ -35,8 +38,10 @@
   protected String level = "DEBUG";
 
   protected MuleContext muleContext;
+  protected FlowConstruct flowConstruct;
```
dfeist: private?

PR #3179, opened 2016-08-25
MULE-10348: Processors in DefaultMessageProcessorChain are not comple…

mbuchwald:
MULE-10348: Processors in DefaultMessageProcessorChain are not completely initialized

### DIFF
--- a/core/src/main/java/org/mule/processor/chain/InterceptingChainLifecycleWrapperPathSkip.java
+++ b/core/src/main/java/org/mule/processor/chain/InterceptingChainLifecycleWrapperPathSkip.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.processor.chain;
+
+import org.mule.api.processor.MessageProcessor;
+import org.mule.api.processor.MessageProcessorChain;
+
+import java.util.List;
+
+public class InterceptingChainLifecycleWrapperPathSkip extends InterceptingChainLifecycleWrapper
+{
+
+    public InterceptingChainLifecycleWrapperPathSkip(MessageProcessorChain chain,
+                                                     List<MessageProcessor> processors,
+                                                     String name)
+    {
+        super(chain, processors, name);
+    }
+
+}
--- a/core/src/main/java/org/mule/processor/chain/InterceptingChainLifecycleWrapperPathSkip.java
+++ b/core/src/main/java/org/mule/processor/chain/InterceptingChainLifecycleWrapperPathSkip.java
@@ -11,6 +11,12 @@ import org.mule.api.processor.MessageProcessorChain;
 
 import java.util.List;
 
+/**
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mule/processor/chain/InterceptingChainLifecycleWrapperPathSkip.java
```diff
@@ -0,0 +1,23 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.processor.chain;
+
+import org.mule.api.processor.MessageProcessor;
+import org.mule.api.processor.MessageProcessorChain;
+
+import java.util.List;
+
+public class InterceptingChainLifecycleWrapperPathSkip extends InterceptingChainLifecycleWrapper{
```
aiannucci: Fix format

core/src/main/java/org/mule/processor/chain/PathSkipMessageProcessorChainBuilder.java
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.processor.chain;
+
+import org.mule.api.processor.MessageProcessorChain;
+
+public class PathSkipMessageProcessorChainBuilder extends DefaultMessageProcessorChainBuilder {
```
aiannucci: Fix format

core/src/main/java/org/mule/processor/chain/InterceptingChainLifecycleWrapperPathSkip.java
mbuchwald: fixed

core/src/main/java/org/mule/processor/chain/PathSkipMessageProcessorChainBuilder.java
mbuchwald: fixed

core/src/main/java/org/mule/processor/chain/InterceptingChainLifecycleWrapperPathSkip.java
```diff
@@ -0,0 +1,24 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.processor.chain;
+
+import org.mule.api.processor.MessageProcessor;
+import org.mule.api.processor.MessageProcessorChain;
+
+import java.util.List;
+
+public class InterceptingChainLifecycleWrapperPathSkip extends InterceptingChainLifecycleWrapper
```
dfeist: Why do some uses of InterceptingChainLifecycleWrapper need to be skipped but others not?

mbuchwald: We are now returning the chain when getting the MPs from Batch. Without skipping the chains, this added a level in the paths. Firstly I tried skipping the InterceptingChainLifecycleWrapper, but for Choice and ScatterGather this removed the last levels (the processors where wrapped with this chain).. So, I needed a type of Interceptor that could be skipped.

PR #3183, opened 2016-08-25

elrodro83:
* remove messageSourceURI, messageSourceName and
ignoredPropertyOverrides

### DIFF
--- a/core-tests/src/test/java/org/mule/runtime/core/DefaultMessageContextTestCase.java
+++ b/core-tests/src/test/java/org/mule/runtime/core/DefaultMessageContextTestCase.java
@@ -51,7 +51,7 @@ public class DefaultMessageContextTestCase extends AbstractMuleTestCase {
 
   @Test
   public void noCorrelationIdInContext() {
-    final MessageContext executionContext = create(flow);
+    final MessageContext executionContext = create(flow, "test");
 
     final MuleMessage message = MuleMessage.builder().payload(TEST_PAYLOAD).build();
     final DefaultMuleEvent event = new DefaultMuleEvent(executionContext, message, flow);
@@ -61,7 +61,7 @@ public class DefaultMessageContextTestCase extends AbstractMuleTestCase {
 
   @Test
   public void parentNoCorrelationIdInContext() {
-    final MessageContext executionContext = create(flow);
+    final MessageContext executionContext = create(flow, "test");
 
     final MuleMessage messageParent = MuleMessage.builder().payload(TEST_PAYLOAD).build();
     final MuleMessage message = MuleMessage.builder().payload(TEST_PAYLOAD).build();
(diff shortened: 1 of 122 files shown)


core-tests/src/test/java/org/mule/runtime/core/DefaultMessageContextTestCase.java
```diff
@@ -51,7 +51,7 @@ public void before() {
 
   @Test
   public void noCorrelationIdInContext() {
-    final MessageContext executionContext = create(flow);
+    final MessageContext executionContext = create(flow, "test");
```
dfeist: - Refactor to use a test utility method rather than passing string "test" each time ideally.  Even if there isn't a util method this should be a constant defined once and reused.
- Static import is not ideal with `create`

core/src/main/java/org/mule/runtime/core/client/DefaultLocalMuleClient.java
```diff
@@ -198,11 +198,11 @@ public void dispatch(String url, MuleMessage message, OperationOptions operation
   }
 
   protected MuleEvent createRequestResponseMuleEvent(MuleMessage message) throws MuleException {
-    return new DefaultMuleEvent(create(flowConstruct), message, REQUEST_RESPONSE, flowConstruct);
+    return new DefaultMuleEvent(create(flowConstruct, "muleClient"), message, REQUEST_RESPONSE, flowConstruct);
```
dfeist: private method to create with, or at minimum constant.

core/src/main/java/org/mule/runtime/core/routing/filters/ExpressionFilter.java
```diff
@@ -85,7 +85,7 @@ public boolean accept(MuleMessage message) {
 
     // TODO MULE-9341 Remove Filters. Expression filter will be replaced by something that uses MuleEvent.
     Flow flowConstruct = new Flow("", muleContext);
-    return accept(new DefaultMuleEvent(create(flowConstruct), message, ONE_WAY, flowConstruct));
+    return accept(new DefaultMuleEvent(create(flowConstruct, "ExpressionFilter"), message, ONE_WAY, flowConstruct));
```
dfeist: Shouldn't these cases have a TODO?

core/src/main/java/org/mule/runtime/core/routing/filters/RegExFilter.java
```diff
@@ -77,7 +77,7 @@ public RegExFilter(String pattern, String value, int flags) {
   public boolean accept(MuleMessage message) {
     // TODO MULE-9341 Remove Filters that are not needed
     Flow flowConstruct = new Flow("", muleContext);
-    return accept(new DefaultMuleEvent(create(flowConstruct), message, ONE_WAY, flowConstruct));
+    return accept(new DefaultMuleEvent(create(flowConstruct, "RegExFilter"), message, ONE_WAY, flowConstruct));
```
dfeist: Shouldn't these cases have a TODO?

core/src/main/java/org/mule/runtime/core/routing/filters/ExpressionFilter.java
elrodro83: there is already a TODO to remove these filters

core/src/main/java/org/mule/runtime/core/routing/filters/RegExFilter.java
elrodro83: there is already a TODO to remove these filters

elrodro83:
jenkins validate

PR #3190, opened 2016-08-26
MULE-10338: Add bulk operation support on new DB connector

marianogonzalez:


extensions/db/src/main/java/org/mule/extension/db/api/param/ParameterizedStatementDefinition.java
```diff
@@ -0,0 +1,119 @@
...
+      final String paramName = inputParameter.getParamName();
+      if (resolvedParameterValues.containsKey(paramName)) {
+        inputParameter.setValue(resolvedParameterValues.get(paramName));
+        resolvedParameterValues.remove(paramName);
+      }
+    });
+
+    resolvedParameterValues.entrySet().stream()
+        .map(entry -> {
+          InputParameter inputParameter = new InputParameter();
```
elrodro83: Not related to this change, but couldn't InputParameter be immutable?

marianogonzalez: fields cannot be final, otherwise SDK wouldn't be able to inject values. I could however replace the setter by a new factory method and make it pseudo-immutable

marianogonzalez: forget about the above. No, I cannot do that

extensions/db/src/main/java/org/mule/extension/db/internal/operation/BulkOperations.java
```diff
@@ -0,0 +1,178 @@
...
+import org.mule.extension.db.internal.util.DefaultFileReader;
+import org.mule.runtime.extension.api.annotation.ParameterGroup;
+import org.mule.runtime.extension.api.annotation.param.Connection;
+import org.mule.runtime.extension.api.annotation.param.UseConfig;
+
+import java.sql.SQLException;
+import java.util.List;
+import java.util.Map;
+
+public class BulkOperations extends BaseDbOperations {
```
elrodro83: add a javadoc at the class level

extensions/db/src/main/java/org/mule/extension/db/internal/resolver/query/BulkQueryFactory.java
```diff
@@ -10,44 +10,35 @@
...
-      return null;
-    }
+  public BulkQuery resolve() {
 
-    BulkQuery bulkQuery = createBulkQuery(muleEvent);
+    BulkQuery bulkQuery = createBulkQuery();
 
     if (bulkQuery.getQueryTemplates().size() == 0) {
-      throw new QueryResolutionException("There are no queries on the resolved dynamic bulk query: " + this.bulkQueryText);
+      throw new QueryResolutionException("There are no queries on the resolved query script");
```
elrodro83: Can this message provide some information about where the script came from?

PR #3192, opened 2016-08-29
MULE-10262: Include error type attribute into on-error handlers

afelisatti:


core/src/main/java/org/mule/runtime/core/routing/AsynchronousUntilSuccessfulProcessingStrategy.java
```diff
@@ -68,8 +70,10 @@
   @Override
   public void initialise() throws InitialisationException {
     if (getUntilSuccessfulConfiguration().getObjectStore() == null) {
-      throw new InitialisationException(MessageFactory
-          .createStaticMessage("A ListableObjectStore must be configured on UntilSuccessful."), this);
+      throw new InitialisationException(
+                                        MessageFactory
+                                            .createStaticMessage("A ListableObjectStore must be configured on UntilSuccessful."),
```
pablolagreca: static imports

PR #3194, opened 2016-08-29
MULE-10352: Fix requester builder lifecycle since mule client depends…

afelisatti:
… on it.


modules/http/src/main/java/org/mule/module/http/internal/request/client/HttpRequesterBuilder.java
```diff
@@ -136,7 +136,7 @@ else if (httpRequester.getConfig() == null)
 
             if (requestConfig == null)
             {
-                requestConfig = new DefaultHttpRequesterConfig();
+                requestConfig = (DefaultHttpRequesterConfig) new HttpRequesterConfigBuilder(muleContext).build();
                 muleContext.getRegistry().registerObject(DEFAULT_HTTP_REQUEST_CONFIG_NAME, requestConfig);
```
elrodro83: this block should be synchronized

PR #3196, opened 2016-08-29
MULE-10376 - Implement support for error mapping and error type disco…

pablolagreca:
…very


core/src/main/java/org/mule/runtime/core/exception/ErrorTypeLocator.java
```diff
@@ -0,0 +1,155 @@
...
+   */
+  public ErrorType lookupErrorType(Exception exception) {
+    return defaultExceptionMapper.resolveErrorType(exception).get();
+  }
+
+  /**
+   * Finds the {@code ErrorType} related to a component defined by the {@link ComponentIdentifier} based on the exception thrown
+   * by the component and the mappings configured in the {@code ErrorTypeLocator}.
+   * 
+   * If no mapping is available then the {@link #getAnyErrorType()} rules applies.
```
afelisatti: The above says UNKNOWN but this one ANY in case of no mapping, which one is correct?

core/src/main/java/org/mule/runtime/core/exception/ErrorTypeRepository.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import org.mule.runtime.api.message.ErrorType;
+import org.mule.runtime.core.message.ErrorTypeBuilder;
+
+/**
+ * Repository for the different {@link ErrorType}s in a mule artifact.
+ *
+ * Only once instance of {@link ErrorType} must exists describing the same
```
afelisatti: one

core/src/main/java/org/mule/runtime/core/exception/ErrorTypeRepository.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  public void addErrorType(String namespace, String stringRepresentation, ErrorType parentErrorType) {
+    checkArgument(namespace != null, "namespace cannot be null");
+    checkArgument(stringRepresentation != null, "string representation cannot be null");
+    ErrorTypeBuilder errorTypeBuilder =
+        ErrorTypeBuilder.builder().namespace(namespace).stringRepresentation(stringRepresentation)
+            .parentErrorType(parentErrorType);
+    if (this.errorTypes.put(new ErrorTypeKey(namespace, stringRepresentation), errorTypeBuilder.build()) != null) {
+      throw new IllegalStateException(format("Already exists an error type with namespace %s and string representation",
```
afelisatti: The wording of this error seems weird to me.

core/src/main/java/org/mule/runtime/core/exception/ErrorTypeLocator.java
pablolagreca: The link to the method was wrong. It should be #lookupErrorType(..)

core/src/main/java/org/mule/runtime/core/exception/ErrorTypeRepository.java
pablolagreca: the placeholder (%s) for string representation was missing. But I'm not sure that fixes what you think it's weird. Can you explain better?

core/src/main/java/org/mule/runtime/core/exception/ErrorTypeRepository.java
```diff
@@ -0,0 +1,108 @@
...
+    public boolean equals(Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (o == null || getClass() != o.getClass()) {
+        return false;
+      }
+
+      ErrorTypeKey that =
+          (ErrorTypeKey) o;
```
afelisatti: Did the formatter fix this?

core/src/main/java/org/mule/runtime/core/exception/TemplateOnErrorHandler.java
```diff
@@ -110,11 +110,11 @@ protected MuleEvent processCatch(MuleEvent event, MessagingException exception)
         TransactionCoordination.getInstance().rollbackCurrentTransaction();
       } catch (Exception ex) {
         //Do nothing
+        logger.warn(ex.getMessage());
       }
 
-      event.setError(builder(exception).build());
       event.setMessage(MuleMessage.builder(event.getMessage()).exceptionPayload(new DefaultExceptionPayload(exception)).build());
-      return event;
+      throw exception;
```
afelisatti: Why were this changes necessary? Seems like they might mess with the catch/rollback approaches.

core/src/main/java/org/mule/runtime/core/exception/ErrorTypeRepository.java
pablolagreca: Yes, this has the formatter applied.

modules/spring-config/src/main/java/org/mule/runtime/config/spring/dsl/model/CoreComponentBuildingDefinitionProvider.java
```diff
@@ -559,13 +561,15 @@ public void init(MuleContext muleContext) {
...
     return (value) -> {
-      if (ANY.getStringRepresentation().equals(value)) {
-        return ANY;
+      ErrorType anyErrorType = muleContext.getErrorTypeLocator().getAnyErrorType();
+      if (anyErrorType.getStringRepresentation().equals(value)) {
+        return anyErrorType;
       } else {
         String[] values = value.split(":");
-        //For now, let's make a namespace mandatory and everything inherit from ANY
+        // For now, let's make a namespace mandatory and everything inherit from ANY
```
afelisatti: Shouldn't we now try to get the error type from the locator?

core/src/main/java/org/mule/runtime/core/exception/TemplateOnErrorHandler.java
pablolagreca: This is a very corner case that almost never happen. What was happening today was that the error handler was actually not failing when there's was an MP failing inside it. Since we are introducing the concept or an scope for error handling, if the users uses an MP that may fail, then they can use the scope to catch such exception. Worst case if there's no scope for error handling then the exception will be propagated to the parent caller.

PR #3197, opened 2016-08-30
Fix flaky test, see MULE-10371

alepulver:


aiannucci:
Commit message should be: <JIRA_ISSUE>: description

PR #3205, opened 2016-08-31
MULE-10417: When sending multiple HTTP headers with the same key using a Requester the format is not the expected by HTTP spec

bayitajesi:


modules/http/src/main/java/org/mule/module/http/internal/domain/request/HttpRequestBuilder.java
```diff
@@ -39,9 +45,22 @@ public HttpRequestBuilder setHeaders(ParameterMap headers)
         return this;
     }
 
-    public HttpRequestBuilder addHeader(String headerName, String headerValue)
+    public HttpRequestBuilder addHeader(String headerName, Object headerValue)
```
afelisatti: I think this should maintain it's signature and the outbound properties should be analysed instead.

PR #3208, opened 2016-09-01
MULE-9867: Add a way to share libraries/module between application pl…

pablokraan:
…ugins

_ Adding RegionClassLoader to manage all the classloaders related to an application
_ Removing sharedLibs classLoaderTestRunneroader
_ Adding getExportedClassPackages and getExportedResources to ClassLoaderFilter
_ Adding ArtifactClassLoaderFilter interface to provide access to the filter configuration
_ Replacing uses of ArtifactClassLoaderFilter with ClassLoaderFilter
_ Adding findLocalClass to ArtifactClassLoaders
_ Adding classloader filter for application classloader
_ Updating app classlaoder factory to properly expose classes and shared libraries packages and resources
_ Adding resources from apps's classes folder to the apps's classloader filter
_ Adding shared libraries to app descriptor
_ Adding shared libraries to classloader and app's lookup policy
_ Adding plugin exported packages into app's lookup policy as parent_first
_ Adding basic plugin dependencies
_ Adding plugin dependency resolver to detect if there are missing plugin dependencies
_ Converting PackageDiscoverer into JarExplorer in order to get both packages an resources.
_ Updating isolated test running to use the same structure of classloaders as in the container
_ Updating isolated classLoader factory to define a filter for app's packages and resources
_ Adding plugin's exported packages to app's lookup policy on isolated test runner
_ Adding test classes as exported by the app classloader on the isolated test runner
_ Migrating validation tests to use the isolated test runner
_ Adding ability to consume application resources from plugins
_ Adding deployment test with plugin depending on other plugin
_ Removing duplication of application's flow execution code
_ Removing Compositea and CompositeApplication classloaders
_ Removing unused class PassThroughClassLoaderFilter
_ Removing GoodCitizenClassLoader


modules/artifact/src/main/java/org/mule/runtime/module/artifact/util/FileJarExplorer.java
```diff
@@ -0,0 +1,81 @@
...
+    try {
+      final File libraryFile = new File(URLDecoder.decode(library.getFile(), "UTF-8"));
+      if (!libraryFile.exists()) {
+        throw new IllegalArgumentException("Library file does not exists: " + library);
+      }
+      if (libraryFile.isDirectory()) {
+        final Collection<File> files = listFiles(libraryFile, TrueFileFilter.TRUE, TrueFileFilter.INSTANCE);
+        for (File classFile : files) {
+          final String relativePath = classFile.getAbsolutePath().substring(libraryFile.getAbsolutePath().length() + 1);
+          if (relativePath.endsWith(".class")) {
```
gsfernandes: Use CLASS_EXTENSION as it is already a constant defined.

PR #3216, opened 2016-09-02
MULE-10265: Create scope for error handling + transactions

afelisatti:


core/src/main/java/org/mule/runtime/core/transaction/TransactionType.java
```diff
@@ -0,0 +1,14 @@
...
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.transaction;
+
+/**
+ * Represents the different kind of transaction available.
+ */
+public enum TransactionType {
+  STANDARD, MULTI, XA
```
pablolagreca: STANDARD -> LOCAL

core/src/main/java/org/mule/runtime/core/processor/BlockMessageProcessor.java
```diff
@@ -74,17 +81,37 @@ protected TransactionFactory getTransactionFactory() {
...
 
-  protected MuleTransactionConfig createTransactionConfig(String action) {
+  protected MuleTransactionConfig createTransactionConfig(String action, TransactionType type) {
     MuleTransactionConfig transactionConfig = new MuleTransactionConfig();
     transactionConfig.setActionAsString(action);
+    transactionConfig.setFactory(lookUpTransactionFactory(type));
     return transactionConfig;
   }
 
+  private TransactionFactory lookUpTransactionFactory(TransactionType type) {
```
pablolagreca: why do we need this in both, the block MP and its factory bean?

afelisatti: Just to make sure there are no incompatibilities if the factory bean is eventually removed

PR #3217, opened 2016-09-02

elrodro83:
* Second iteration of replacing MuleEvent.setMessage calls with the builder
* Change event mutators from transformers to messageProcessors


core/src/main/java/org/mule/runtime/core/event/mutator/AbstractAddVariablePropertyProcessor.java
```diff
@@ -99,4 +104,11 @@ public void setValue(String value) {
     this.valueEvaluator = new AttributeEvaluator(value);
   }
 
+  public void setReturnDataType(DataType type) {
```
dfeist: Are these needed, i don't think they make sense here.  I assume this is because XML element extends abstract-transformer, instead can we change that?

core/src/main/java/org/mule/runtime/core/event/mutator/AbstractEventMutatorProcessor.java
```diff
@@ -0,0 +1,33 @@
...
+import org.mule.runtime.core.AbstractAnnotatedObject;
+import org.mule.runtime.core.api.MuleContext;
+import org.mule.runtime.core.api.MuleEvent;
+import org.mule.runtime.core.api.MuleMessage;
+import org.mule.runtime.core.api.context.MuleContextAware;
+import org.mule.runtime.core.api.lifecycle.Initialisable;
+import org.mule.runtime.core.api.processor.MessageProcessor;
+
+/**
+ * Base clkass for {@link MessageProcessor}s that change the event or message. Implementations will return a new instance of the
```
dfeist: Typo

core/src/main/java/org/mule/runtime/core/event/mutator/AbstractEventMutatorProcessor.java
```diff
@@ -0,0 +1,33 @@
...
+ * event with the updated data, since the {@link MuleEvent} and {@link MuleMessage} objects are immutable.
+ *
+ * @since 4.0
+ */
+public abstract class AbstractEventMutatorProcessor extends AbstractAnnotatedObject
```
dfeist: All message processor are potentially event mutators by definition, so I don't see the need for this abstract class at all TBH.

core/src/main/java/org/mule/runtime/core/event/mutator/RemovePropertyProcessor.java
```diff
@@ -4,19 +4,20 @@
  * license, a copy of which has been included with this distribution in the
  * LICENSE.txt file.
  */
-package org.mule.runtime.core.transformer.simple;
+package org.mule.runtime.core.event.mutator;
```
dfeist: Don't like this package name at all. Not sure it was 100% required to move them from transformer.simple, but obviously they are no longer true transformers so the change does make sense.  how about just in `event` or `core.processor.simple`?

core/src/main/java/org/mule/runtime/core/exception/TemplateOnErrorHandler.java
```diff
@@ -109,7 +109,8 @@ protected MuleEvent processCatch(MuleEvent event, MessagingException exception)
         logger.warn(ex.getMessage());
       }
 
-      event.setMessage(MuleMessage.builder(event.getMessage()).exceptionPayload(new DefaultExceptionPayload(exception)).build());
+      // event.setMessage(MuleMessage.builder(event.getMessage()).exceptionPayload(new
```
dfeist: ?

transports/core/src/main/java/org/mule/compatibility/core/transformer/simple/MutatorTransformerWrapper.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.nio.charset.Charset;
+
+/**
+ * Provides a way to declare {@link AbstractEventMutatorProcessor} instances as transformers inside an {@link InboundEndpoint} or
+ * {@link OutboundEndpoint}.
+ *
+ * @since 4.0
+ */
+public class MutatorTransformerWrapper extends AbstractMessageTransformer implements MuleContextAware {
```
dfeist: MessageProcessorTransformerAdaptor?

transports/module-support/src/main/java/org/mule/compatibility/config/spring/handlers/MuleTransportsNamespaceHandler.java
```diff
@@ -68,15 +69,17 @@ public void init() {
     registerBeanDefinitionParser("default-exception-strategy",
                                  new ExceptionStrategyDefinitionParser(DefaultMessagingExceptionStrategy.class));
 
+    registerBeanDefinitionParser("mutator-transformer", new MessageProcessorDefinitionParser(MutatorTransformerWrapper.class));
```
dfeist: This shouldn't be needed, we should be able to wrap with adaptor transparently.

core/src/main/java/org/mule/runtime/core/event/mutator/AbstractAddVariablePropertyProcessor.java
elrodro83: Actually, this is defined for this specific element and its subchildren in `abstractAddPropertyProcessorType` in the xsd, plus there are tests testing this functionality outside of transformers.

PR #3221, opened 2016-09-04

elrodro83:
* Change in filters
* Core almost completed


core/src/main/java/org/mule/runtime/core/routing/AsynchronousUntilSuccessfulProcessingStrategy.java
```diff
@@ -154,7 +154,7 @@ private void incrementProcessAttemptCountAndRescheduleOrRemoveFromStore(final Se
                                                                           Exception lastException) {
     try {
       final MuleEvent event = getUntilSuccessfulConfiguration().getObjectStore().remove(eventStoreKey);
-      final MuleEvent mutableEvent = event;
+      MuleEvent mutableEvent = event;
```
dfeist: rename var?

core/src/main/java/org/mule/runtime/core/routing/EventGroup.java
```diff
@@ -217,6 +217,9 @@ public void addEvent(MuleEvent event) throws ObjectStoreException {
       // Using both event ID and CorrelationSequence since in certain instances
       // when an event is split up, the same event IDs are used.
       Serializable key = getEventKey(event);
+      // TODO MULE-9342 migrating this breaks
```
elrodro83: TODO MULE-9342

PR #3222, opened 2016-09-05
MULE-10367: support for excluding a type from top level elements

alegmarra:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/SubtypesModelValidator.java
```diff
@@ -44,71 +47,89 @@ public void validate(ExtensionModel model) throws IllegalModelDefinitionExceptio
...
 
+  private void validateSubtypesSupportGlobalDeclaration(ExtensionModel model,
+                                                        Map<MetadataType, List<MetadataType>> typesMapping) {
+    List<MetadataType> nonGlobalTypes = typesMapping.values().stream().flatMap(Collection::stream)
+        .filter(s -> s.getAnnotation(NotGlobalTypeAnnotation.class).isPresent()).collect(toList());
+
+    if (!nonGlobalTypes.isEmpty()) {
+      throw new IllegalModelDefinitionException(format("All the declared SubtypesMapping in extension %s must support to be declared"
+          + " as global types, but [%s] do not support it",
+                                                       model.getName(), Arrays.toString(nonGlobalTypes.toArray())));
```
estebanwasinger: Just doing a `toString()` for a `MetadataType` will generate an unreadable StackTrace

PR #3228, opened 2016-09-05
MULE-10466: Update javax transport version to 1.2

mbuchwald:


distributions/standalone/assembly-whitelist.txt
```diff
@@ -296,6 +295,7 @@
 /mule-standalone-${productVersion}/lib/opt/javassist-3.7.ga.jar
 /mule-standalone-${productVersion}/lib/opt/javax.annotation-api-1.2.jar
 /mule-standalone-${productVersion}/lib/opt/javax.inject-1.jar
+/mule-standalone-${productVersion}/lib/opt/javax.transaction-api-1.2-b03.jar
```
pablokraan: This version is a beta, use 1.2

mbuchwald: Fixed

aiannucci:
There should be an equivalent EE PR for this change.

mbuchwald:
Done

pom.xml
```diff
@@ -575,10 +575,6 @@
                     </exclusion>
                     <exclusion>
                         <groupId>org.apache.geronimo.specs</groupId>
-                        <artifactId>geronimo-jta_1.0.1B_spec</artifactId>
```
aiannucci: If you remove this exclusion it would cause the library to be potentially pulled again by the activemq dependency

mbuchwald: Fixed

PR #3242, opened 2016-09-07
MULE-10295: target attribute accepts expressions

pablomusumeci:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
```diff
@@ -147,7 +151,23 @@ private ReturnDelegate createReturnDelegate() {
...
-    return StringUtils.isBlank(target) ? new ValueReturnDelegate(muleContext) : new TargetReturnDelegate(target, muleContext);
+    return !isTargetPresent() ? new ValueReturnDelegate(muleContext) : new TargetReturnDelegate(target, muleContext);
+  }
+
+  private boolean isTargetPresent() {
+    if (isBlank(target)) {
+      return false;
+    }
+
+    if (target.startsWith(FLOW_VARS)) {
```
marianogonzalez: this is the wrong place to do this.. Application should not deploy. This should be validated on the OperationMessageProcessor constructor.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
```diff
@@ -147,7 +151,23 @@ private ReturnDelegate createReturnDelegate() {
...
+      throw new IllegalOperationException(format(INVALID_TARGET_MESSAGE, flowConstruct.getName(), operationModel.getName(),
+                                                 format("a flowVar with the '%s' prefix", FLOW_VARS)));
```
marianogonzalez: this message is redundant. Read it to yourself and you'll notice that as a user you wouldn't like to receive that message

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessorTestCase.java
```diff
@@ -41,6 +47,11 @@
 @RunWith(MockitoJUnitRunner.class)
 public class OperationMessageProcessorTestCase extends AbstractOperationMessageProcessorTestCase {
 
+  public static final String INVALID_TARGET =
```
marianogonzalez: this should be the same constant than the validation uses. Make the original constant package level and place this test in the same package

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
marianogonzalez: or you could also validate on the initialise phase

marianogonzalez: better to do it in the initialise

pablomusumeci: This validation runs when the return delegate is created, which happens in the `doInitialise` method.

PR #3244, opened 2016-09-07
MULE-10453: Memory leak caused by per-app logging

pablokraan:
_ Using appplication's RegionClassLoader to log everything coming from the app.
_ Cleaining how to loggers are obtained avoiding the creation of extra ArtifactDescriptors
_ Adding artifact descriptor to each artifact classloader
_ Making artifact descriptor's name fixed on the constructor
_ Implementing dispose on Region classlaoder to clean up every region's member
_ Extra cleanup

### DIFF
--- a/modules/artifact/src/main/java/org/mule/runtime/module/artifact/classloader/RegionClassLoader.java
+++ b/modules/artifact/src/main/java/org/mule/runtime/module/artifact/classloader/RegionClassLoader.java
@@ -7,8 +7,10 @@
 
 package org.mule.runtime.module.artifact.classloader;
 
+import static java.util.Collections.emptyList;
 import static org.mule.runtime.core.util.Preconditions.checkArgument;
 import org.mule.runtime.core.util.ClassUtils;
+import org.mule.runtime.module.artifact.descriptor.ArtifactDescriptor;
 
 import java.io.IOException;
 import java.net.URL;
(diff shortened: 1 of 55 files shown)


modules/artifact/src/main/java/org/mule/runtime/module/artifact/classloader/RegionClassLoader.java
```diff
@@ -133,4 +148,34 @@ public final URL findResource(final String name) {
...
     return new CompoundEnumeration<>(enumerations.toArray(new Enumeration[0]));
   }
+
+  @Override
+  public void dispose() {
+    for (ArtifactClassLoader classLoader : unfilteredClassLoaders) {
+      try {
+        classLoader.dispose();
+      } catch (Exception e) {
+        // Ignore and continue
```
pablolagreca: Even thought we can't do anything if there's an error and it should continue, it should log the exception in debug level.

modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/application/DefaultMuleApplication.java
```diff
@@ -257,6 +255,10 @@ public void dispose() {
     }
   }
 
+  private static boolean isRegionClassLoaderMember(ClassLoader loggerClassLoader) {
```
pablolagreca: param name should be classLoader instead of loggerClassLoader

modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/ArtifactAwareContextSelector.java
```diff
@@ -91,10 +92,16 @@ static ClassLoader resolveLoggerContextClassLoader(ClassLoader classLoader) {
     ClassLoader loggerClassLoader = classLoader == null ? Thread.currentThread().getContextClassLoader() : classLoader;
     if (!(loggerClassLoader instanceof ArtifactClassLoader)) {
       loggerClassLoader = loggerClassLoader.getSystemClassLoader();
+    } else if (isRegionClassLoaderMember(loggerClassLoader)) {
+      loggerClassLoader = loggerClassLoader.getParent();
     }
     return loggerClassLoader;
   }
 
+  private static boolean isRegionClassLoaderMember(ClassLoader loggerClassLoader) {
```
pablolagreca: same as before

tests/functional/src/main/resources/maven-module-mapping.properties
```diff
@@ -80,3 +80,5 @@ mule-api=/mule-api/
 mule-extensions-api=/mule-extensions-api/mule-extensions-api/
 mule-extensions-api-persistence=/mule-extensions-api/mule-extensions-api-persistence/
 mule-extensions-api-xml-dsl=/mule-extensions-api/mule-extensions-api-xml-dsl/
+mule-extensions-api-dsql=/mule-extensions-api/mule-extensions-api-dsql/
```
pablolagreca: Does this belong to this PR?

pablokraan: Not really, this is extra mapping is required when you have mule, mule-api. extensions-api in the same project. I can move it to other PR if you want... but this file will go way with Guille's PR about the test runner

PR #3248, opened 2016-09-07

elrodro83:
* Migrate reamining modules


extensions/validation/src/test/java/org/mule/extension/validation/ValidationElTestCase.java
```diff
@@ -38,12 +38,13 @@ protected void doSetUp() throws Exception {
   @Test
   public void email() throws Exception {
     final String expression = "#[validator.validateEmail(email)]";
-    MuleEvent event = getTestEvent("");
-    event.setFlowVariable("email", VALID_EMAIL);
+    MuleEvent event = MuleEvent.builder(getTestEvent(""))
+        .addFlowVariable("email", VALID_EMAIL)
```
dfeist: We need to update test infrastructure rather than do this.

PR #3251, opened 2016-09-08
MULE-10477: Add layout hints on DB connector

estebanwasinger:


extensions/db/src/main/java/org/mule/extension/db/api/param/ParameterizedStatementDefinition.java
```diff
@@ -40,8 +41,8 @@
    */
   @Parameter
   @Optional
-  @DisplayName("Input Parameters")
```
marianogonzalez: why this?

PR #3252, opened 2016-09-08
MULE-10357: Reduce test thread leaks

elrodro83:
* Cleanup partial init of muleContext when it fails.
* Remove leaked threads in BlockingProcessorExecutorTestCase
* Reduce memory usage of AbstractTransactionQueueManagerTestCase


core-tests/src/test/java/org/mule/runtime/core/util/queue/AbstractTransactionQueueManagerTestCase.java
```diff
@@ -647,18 +645,15 @@ public void testOffer() throws Exception {
...
-          } catch (Exception e) {
-            // unlikely to happen. But if it does lets show it in the test logs.
-            logger.warn("Error using queue session", e);
-          }
+      final Thread takeExecutionThread = new Thread(() -> {
+        try {
+          takeExecutionLatch.release();
+          QueueSession s1 = mgr.getQueueSession();
+          Queue q1 = s1.getQueue("queue1");
+          assertEquals("Queue content", "String1", q1.take());
```
afelisatti: Might want to take advantage and move to hamcrest.

PR #3256, opened 2016-09-08
MULE-10456: Exported types are not declared in the extension model

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/IntrospectionUtils.java
```diff
@@ -384,7 +385,11 @@ private static boolean isVoid(Class<?> type) {
   }
 
   public static Collection<Method> getOperationMethods(Class<?> declaringClass) {
-    return getAllMethods(declaringClass, withModifier(PUBLIC), not(withAnnotation(Ignore.class)));
+    return getAllSuperTypes(declaringClass).stream()
+        .flatMap(type -> Stream.of(type.getDeclaredMethods()))
+        .filter(method -> isPublic(method.getModifiers()))
+        .filter(method -> !method.isAnnotationPresent(Ignore.class))
+        .collect(toCollection(LinkedHashSet::new));
```
estebanwasinger: Which is the difference between doing it by hand or just calling `getAllMethods`?

marianogonzalez: getAllMethods returns a HashSet of unpredictable order, which makes the schema test cases flecky

PR #3262, opened 2016-09-09
MULE-9283: Review and update naming and deprecation

elrodro83:
Just the deprecation for now


core/src/main/java/org/mule/runtime/core/api/MuleEvent.java
```diff
@@ -124,21 +132,25 @@
...
    */
+  @Deprecated
   MuleSession getSession();
 
   /**
    * Retrieves the service for the current event
    * 
    * @return the service for the event
-   * @deprecated TODO MULE-10302 remove this.
+   * @deprecated TODO MULE-9731 Migrate 3.7 {@link ReplyToHandler}-centric non-blocking support to use new non-blocking API
```
dfeist: Is it only non-blocking that needs flowConstruct?

PR #3263, opened 2016-09-09
MULE-10266: Define common error types and create mapping for internal exceptions within core

afelisatti:


core-tests/src/test/java/org/mule/runtime/core/routing/filters/ExpressionFilterTestCase.java
```diff
@@ -137,15 +136,14 @@ public void testRegexFilterWithAngleBrackets() {
 
   @Test
   public void testExceptionTypeFilter() {
-    ExpressionFilter filter = new ExpressionFilter("exception is java.lang.Exception");
+    ExpressionFilter filter = new ExpressionFilter("exception.getCause() is java.lang.Exception");
```
pablolagreca: Can you add a comment about this in the compatibility section of the error handling document?

core/src/main/java/org/mule/runtime/core/api/routing/RoutePathNotFoundException.java
```diff
@@ -21,19 +20,19 @@
    */
   private static final long serialVersionUID = -8481434966594513066L;
 
-  public RoutePathNotFoundException(MuleEvent event, MessageProcessor target) {
-    super(event, target);
+  public RoutePathNotFoundException(MessageProcessor target) {
```
pablolagreca: We need a separate issue to move this exceptions to the API and refactor to extract also the need for the MessageProcessor as property of the exceptions. Can you create one? No need to add a TODO

modules/cxf/src/main/java/org/mule/runtime/module/cxf/CxfOutboundMessageProcessor.java
```diff
@@ -148,10 +147,12 @@ private MuleException wrapException(MuleEvent event, Throwable ex, boolean alway
...
       }
-      return new DispatchException(MessageFactory.createStaticMessage(fault.getMessage()), event, this, fault);
+      DispatchException dispatchException = new DispatchException(createStaticMessage(fault.getMessage()), this, fault);
+      return new MessagingException(event, dispatchException);
     }
-    return new DispatchException(MessageFactory.createStaticMessage(ExceptionHelper.getRootException(ex).getMessage()), event,
-                                 this, ex);
+    DispatchException dispatchException =
+        new DispatchException(createStaticMessage(ExceptionHelper.getRootException(ex).getMessage()), this, ex);
+    return new MessagingException(event, dispatchException);
```
pablolagreca: Why do we need a MessagingException here?

transports/jms/src/test/resources/jms-redelivery-flow.xml
```diff
@@ -22,7 +22,7 @@
         <default-exception-strategy>
             <commit-transaction exception-pattern="org.mule.compatibility.transport.jms.redelivery.MessageRedeliveredException"/>
             <choice>
-                <when expression="#[payload.exception instanceof org.mule.compatibility.transport.jms.redelivery.MessageRedeliveredException]">
+                <when expression="#[payload.getException() instanceof org.mule.compatibility.transport.jms.redelivery.MessageRedeliveredException]">
```
pablolagreca: why did you have to change from .exception to .getException() ?

afelisatti: I think I had mistakenly changed it and to restore it back I used getException(), I'll check.

modules/cxf/src/main/java/org/mule/runtime/module/cxf/CxfOutboundMessageProcessor.java
afelisatti: Probably because the event is needed, I'll check.

core-tests/src/test/java/org/mule/runtime/core/routing/filters/ExpressionFilterTestCase.java
afelisatti: This is caused by wrapping the exception in a MessagingException to support the "causedBy" and I think it should go away once we can get rid of all those uses. Should I add the "causedBy" stuff to the migration and create a ticket to get rid of it?

core/src/main/java/org/mule/runtime/core/api/routing/RoutePathNotFoundException.java
afelisatti: Done: MULE-10519. I kept the message processor in this case not because of the "failingProcessor" stuff but because it is used to check the route.

core-tests/src/test/java/org/mule/runtime/core/routing/filters/ExpressionFilterTestCase.java
afelisatti: I've added a comment in MULE-10519 (move everything away from ME) explaining that these change should be restored once that is gone and the wrapping no longer needed.

PR #3272, opened 2016-09-12
MULE-10527: Error processing stored procedure with output params

pablokraan:
_ Defining output parameters with type name for structured types only


marianogonzalez:
no test?

pablokraan:
This is basically a refactoring, moving a method that affects every types to just affect structured types. The issue, as described in the jira, is caused by the driver that does not respect the API.
Besides that, having to write a test for this would require to use the specific driver that has the problem

PR #3275, opened 2016-09-12
MULE-10303: Extensions Parser fails to load pojos declared as top ele…

estebanwasinger:
…ment if the pojo is only used as value of Map

### DIFF
--- a/tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/model/SaleInfo.java
+++ b/tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/model/SaleInfo.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.test.heisenberg.extension.model;
+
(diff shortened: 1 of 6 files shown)


tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/model/SaleInfo.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package org.mule.test.heisenberg.extension.model;
+
+import org.mule.runtime.extension.api.annotation.Parameter;
+import org.mule.test.heisenberg.extension.HeisenbergOperations;
+
+import java.util.Map;
+
+/**
+ * DON'T USE THIS CLASS BEYOND {@link HeisenbergOperations#processSale(Map)}
```
marianogonzalez: why is this restriction? I should be able to reuse this same class somewhere else. Otherwise the solution is not applicable

estebanwasinger: If this class is used somewhere else, the test will loose their sense. The bug that is being fixed happen when you used the class only as the value of the map, so using this class as a plain parameter will make the test not cover the bug to fix

marianogonzalez: the comment is misleading then. Just to make sure, I can have a Map<Apple, Kiwi> and and operation public Apple getFruit() and everything works perfectly, right?

PR #3277, opened 2016-09-12
MULE-10203: Move to a separate module ArtifactClassLoaderRunner

pablokraan:


modules/extensions-support/pom.xml
```diff
@@ -51,11 +51,11 @@
...
-        <dependency>
-            <groupId>org.mule.modules</groupId>
-            <artifactId>mule-module-spring-config</artifactId>
-            <version>${project.version}</version>
-        </dependency>
+        <!--<dependency>-->
+            <!--<groupId>org.mule.modules</groupId>-->
+            <!--<artifactId>mule-module-spring-config</artifactId>-->
+            <!--<version>${project.version}</version>-->
+        <!--</dependency>-->
```
gsfernandes: Why did you have to be comment this dependency?

pablokraan: I commented it when I was trying some things, it turns out that is not required. I'll delete it

tests/runner/src/main/resources/META-INF/mule-module.properties
```diff
@@ -0,0 +1,23 @@
...
+                              org.mule.test.runner,\
+                              org.junit.runners.model,\
+                              org.junit.runner.notification,\
+                              org.junit.runner,\
+                              junit.framework,\
+                              org.junit.runner.manipulation
+
+
+artifact.export.resources=/META-INF,\
+                          /META-INF/services/org/mule/runtime/core/config
```
gsfernandes: Do we need to expose this? I think that this module will not be loaded by a container or if it is we should not expose anything.

pablokraan: I moved the packages that were included in the module properties from the functional test module. I assume that they were required

PR #3282, opened 2016-09-13
MULE-10498: Review exported extension plugins when packaging.

juandesi:


extensions/pom.xml
```diff
@@ -111,6 +111,23 @@
                 </dependencies>
             </plugin>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <version>2.6</version>
```
aiannucci: use ${jar.plugin.version}, which is already defined in the root POM

gsfernandes:
This was already included as part of #3283

PR #3284, opened 2016-09-13
MULE-9283 Review and update naming and deprecation. Move message interfaces to message package.  Deprecate ExceptionPayload.

dfeist:


core-tests/src/test/java/org/mule/runtime/core/DefaultMessageContextTestCase.java
```diff
@@ -13,11 +13,9 @@
 import org.mule.runtime.core.api.EventContext;
 import org.mule.runtime.core.api.MuleContext;
 import org.mule.runtime.core.api.Event;
-import org.mule.runtime.core.api.InternalMessage;
+import org.mule.runtime.core.api.message.InternalMessage;
```
pablolagreca: This may be done in a separate issue, but InternalMessage shouldn't be in an internal package or at least out of core.api?

dfeist: The thing is, if we like it or not this is API for now because transport, studio, munit will all need to use this and mule-core will need to expose it.  If in a later milestone we update exports to only expose api this will be a problem if it's not here i think. Maybe internal wasn't the best name.

PR #3286, opened 2016-09-13
MULE-10303: Fix tests

estebanwasinger:


modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/AnnotationsBasedDescriberTestCase.java
```diff
@@ -413,7 +413,7 @@ private void assertExtensionProperties(ExtensionDeclaration extensionDeclaration
   }
 
   private void assertTestModuleOperations(ExtensionDeclaration extensionDeclaration) throws Exception {
-    assertThat(extensionDeclaration.getOperations(), hasSize(22));
+    assertThat(extensionDeclaration.getOperations(), hasSize(23));
```
marianogonzalez: each of those 22 other operations have assertions. This one should too

PR #3288, opened 2016-09-13
MULE-10474: Review email operations that return or receive a multipart mule message.

juandesi:


extensions/email/src/main/java/org/mule/extension/email/api/EmailContentProcessor.java
```diff
@@ -58,84 +63,91 @@ private EmailContentProcessor(javax.mail.Message message) {
...
+   * Processes an attachment part, adding it to the attachments list.
+   *
+   * @param part the attachment part to be processed
+   */
+  private void processAttachment(Part part) throws MessagingException, IOException {
+    Map<String, LinkedList<String>> headers = new HashMap<>();
+    final Enumeration allHeaders = part.getAllHeaders();
+    while (allHeaders.hasMoreElements()) {
+      Header h = (Header) allHeaders.nextElement();
+      headers.put(h.getName(), new LinkedList<>(singletonList(h.getValue())));
```
marianogonzalez: A non mutable copy of an immutable single element list... overkill... Just use Arrays.asList() instead since that list will never be exposed to the user anyway

extensions/email/src/main/java/org/mule/extension/email/internal/commands/SendCommand.java
```diff
@@ -28,45 +30,37 @@
...
-          .fromAddresses(fromAddress)
-          .to(toAddresses)
-          .cc(ccAddresses)
-          .bcc(bccAddresses)
-          .withSubject(subject)
-          .withAttachments(attachments)
-          .withContent(content.getBody(), content.getContentType(),
-                       content.getCharset() == null ? defaultCharset : content.getCharset())
-          .withHeaders(headers).build();
+          .fromAddresses(configuration.getFrom())
```
marianogonzalez: shouldn't the from be taken from the emailBuilder?

extensions/email/src/test/java/org/mule/extension/email/EmailContentProcessorTestCase.java
```diff
@@ -18,8 +17,8 @@
 import static org.mule.extension.email.util.EmailTestUtils.assertAttachmentContent;
 import static org.mule.extension.email.util.EmailTestUtils.getMultipartTestMessage;
 import static org.mule.extension.email.util.EmailTestUtils.getSinglePartTestMessage;
-
 import org.mule.runtime.api.message.Message;
+import org.mule.extension.email.api.EmailContentProcessor;
```
marianogonzalez: why is this class API?

extensions/email/src/main/java/org/mule/extension/email/internal/commands/SendCommand.java
juandesi: no, the from address is specified in the configs

PR #3290, opened 2016-09-13
MULE-10507: add system property to specify Oauth2 external redirect/c…

alepulver:
…allback path


modules/oauth/src/main/java/org/mule/module/oauth2/internal/authorizationcode/DefaultAuthorizationCodeGrantType.java
```diff
@@ -84,6 +88,16 @@ public String getRedirectionUrl()
         return redirectionUrl;
     }
 
+    public String getExternalRedirectionUrl()
+    {
+        if (externalRedirectionUrl == null) {
```
pablolagreca: You should do this inside the initialise method.

PR #3291, opened 2016-09-13
MULE-10203: Move to a separate module ArtifactClassLoaderRunner

pablokraan:


tests/runner/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <!-- Override scope="test" from parent -->
+            <scope>compile</scope>
+        </dependency>
+    </dependencies>
+</project>
```
gsfernandes: Today I was checking a dependency that is used by the runner and not declared. @pablolagreca removed this transitive dependency and found this issue, as part of this PR it would be nice to add the missing declared dependency:
```
org.mule.tests:mule-tests-functional:jar:4.0-SNAPSHOT->org.mule.modules:mule-module-extensions-support:jar:4.0-SNAPSHOT:compile->org.mule.modules:mule-module-tooling-support:jar:4.0-SNAPSHOT:compile->org.mule.modules:mule-module-repository:jar:4.0-SNAPSHOT:compile->org.apache.maven:maven-aether-provider:jar:3.3.9:compile->org.apache.maven:maven-model:jar:3.3.9:compile
```
org.apache.maven:maven-model:jar:3.3.9 could be added as compile here.

gsfernandes: This pom has to be added as module on mule/tests/pom.xml, that's why the check failed.

PR #3292, opened 2016-09-13
MULE-10507: use XSD parameter instead of system property for the inte…

alepulver:
…rface

### DIFF
--- a/modules/oauth/src/main/java/org/mule/module/oauth2/internal/authorizationcode/DefaultAuthorizationCodeGrantType.java
+++ b/modules/oauth/src/main/java/org/mule/module/oauth2/internal/authorizationcode/DefaultAuthorizationCodeGrantType.java
@@ -8,7 +8,6 @@ package org.mule.module.oauth2.internal.authorizationcode;
 
 import static org.mule.api.config.MuleProperties.SYSTEM_PROPERTY_PREFIX;
 import static org.mule.config.i18n.MessageFactory.createStaticMessage;
-
 import org.mule.api.MuleContext;
 import org.mule.api.MuleEvent;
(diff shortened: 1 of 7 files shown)


modules/oauth/src/main/java/org/mule/module/oauth2/internal/authorizationcode/DefaultAuthorizationCodeGrantType.java
```diff
@@ -169,8 +170,7 @@ public void initialise() throws InitialisationException
             resourceOwnerIdEvaluator.initialize(muleContext.getExpressionManager());
             if (externalRedirectionUrl == null)
             {
-                String url = System.getProperty(EXTERNAL_REDIRECT_URL_PROPERTY);
```
pablolagreca: I would still support the system property buy fail if both, the system property and the externalRedirectionUrl attribute are defined.

modules/oauth/src/main/resources/META-INF/mule-oauth2.xsd
```diff
@@ -92,6 +92,15 @@
                         </xsd:documentation>
                     </xsd:annotation>
                 </xsd:attribute>
+                <xsd:attribute name="externalRedirectionUrl" type="xsd:string" use="optional">
+                    <xsd:annotation>
+                        <xsd:documentation>
+                            This is the redirection url to send to the user, and must match the one configured in the
+                            oauth provider. It won't create an endpoint by itself, and should be used when the
```
afelisatti: What do you mean by "must match the one configured in the oauth provider"? Seems like you're referencing the AES module.

PR #3299, opened 2016-09-15
MULE-10556: Fix performance tests and run them in main build as junit…

LucianoGandini:
… tests so future errors are catched faster.


tests/performance/src/test/java/org/mule/test/performance/util/AbstractPerformanceTestCase.java
```diff
@@ -0,0 +1,28 @@
...
+
+package org.mule.test.performance.util;
+
+
+import org.mule.tck.junit4.AbstractMuleContextTestCase;
+
+import org.databene.contiperf.junit.ContiPerfRule;
+import org.junit.Rule;
+
+public class AbstractPerformanceTestCase extends AbstractMuleContextTestCase {
```
elrodro83: This class only has the rule. I think it's not necessary.

aiannucci: Although it defines an increased timeout. If you don't define this, many performance tests will fail

elrodro83: Also, configure the test timeout via system property passed through surefire (only when actually running the performance tests)

LucianoGandini: Created profile to pass the timeout and ContiPerf property only when runing perf tests.

LucianoGandini: Also the class is now the usual "AbstractMuleContextTestCase"

PR #3302, opened 2016-09-16
MULE-10578 Exception thrown form a Work#run() is ignored by the

elrodro83:
WorkManager


dfeist:
Make sense to add a test case?

PR #3317, opened 2016-09-19
MULE-10435: Add a way to shared libs on the isolated tests

gsfernandes:


tests/runner/pom.xml
```diff
@@ -47,43 +47,60 @@
...
         </dependency>
-
         <dependency>
             <groupId>org.mule.tests</groupId>
             <artifactId>mule-tests-unit</artifactId>
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
```
pablokraan: Email extension is also using this, it should be defined in mule as managed dependency

tests/runner/src/test/java/org/mule/test/runner/api/AetherClassPathClassifierTestCase.java
```diff
@@ -0,0 +1,189 @@
...
+  private Dependency muleArtifactDep;
+  private Dependency muleTestFunctionalDep;
+  private Dependency guavaDep;
+  private Dependency derbyDriverDep;
+
+  private List<Dependency> directDependencies;
+
+  @Before
+  public void before() throws Exception {
+    this.rootArtifact = new DefaultArtifact("org.mule.tests:mule-tests-runner:4.0-SNAPSHOT");
```
pablokraan: Is the 4.0-SNAPSHOT related to the Mule project version?

tests/runner/src/test/java/org/mule/test/runner/api/AetherClassPathClassifierTestCase.java
```diff
@@ -0,0 +1,189 @@
...
+
+  private ArtifactDescriptorResult noManagedDependencies() throws ArtifactDescriptorException {
+    ArtifactDescriptorResult artifactDescriptorResult = mock(ArtifactDescriptorResult.class);
+    List<Dependency> managedDependencies = Collections.<Dependency>emptyList();
+    when(artifactDescriptorResult.getManagedDependencies()).thenReturn(managedDependencies);
+    when(dependencyResolver.readArtifactDescriptor(any(Artifact.class))).thenReturn(artifactDescriptorResult);
+    return artifactDescriptorResult;
+  }
+
+  class ListOfElements extends ArgumentMatcher<List> {
```
pablokraan: cannot use IsCollectionWithSize matcher?

tests/runner/src/test/java/org/mule/test/runner/api/AetherClassPathClassifierTestCase.java
gsfernandes: No, it is just mock data.

pablokraan: the I would suggest you to use fantasy artifact IDs, like org.foo:foo:1.0

PR #3319, opened 2016-09-19
MULE-10593: App with missing config fails to deploy but has status "c…

alepulver:
…reated"

### DIFF
--- a/modules/launcher/src/main/java/org/mule/module/launcher/application/DefaultMuleApplication.java
+++ b/modules/launcher/src/main/java/org/mule/module/launcher/application/DefaultMuleApplication.java
@@ -6,6 +6,7 @@
  */
 package org.mule.module.launcher.application;
 
+import static org.mule.config.i18n.MessageFactory.createStaticMessage;
 import static org.mule.util.SplashScreen.miniSplash;
 import org.mule.MuleServer;
 import org.mule.api.MuleContext;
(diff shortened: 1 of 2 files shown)


modules/launcher/src/test/java/org/mule/module/launcher/DeploymentServiceTestCase.java
```diff
@@ -2215,6 +2215,25 @@ public void redeploysExplodedDomainAfterDeploymentError() throws Exception
...
+
+    @Test
+    public void deployFailsWhenMissingFile() throws Exception
+    {
+        addExplodedAppFromBuilder(emptyAppFileBuilder);
+
+        deploymentService.start();
+
+        assertApplicationDeploymentSuccess(applicationDeploymentListener, emptyAppFileBuilder.getId());
+        assertAppsDir(NONE, new String[] {emptyAppFileBuilder.getId()}, true);
```
pablokraan: These lines are redundant, just check that the app was deployed

alepulver: I suppose this checks that the test fails for the correct reason, and not a problem in the original application (if we could just test direct deployment this wouldn't be necessary). It was also in the other test.

modules/launcher/src/main/java/org/mule/module/launcher/application/DefaultMuleApplication.java
```diff
@@ -92,16 +92,22 @@ public void install()
...
-            if (!configResource.exists())
+            for (String configResourceAbsolutePatch : this.descriptor.getAbsoluteResourcePaths())
             {
-                String message = String.format("Config for app '%s' not found: %s", getArtifactName(), configResource);
-                throw new InstallException(MessageFactory.createStaticMessage(message));
+                File configResource = new File(configResourceAbsolutePatch);
+                if (!configResource.exists())
+                {
+                    String message = String.format("Config for app '%s' not found: %s", getArtifactName(), configResource);
+                    throw new InstallException(MessageFactory.createStaticMessage(message));
```
alepulver: Done.

PR #3322, opened 2016-09-19
MULE-10594: Generic types are not being exported in the extension manifest.

juandesi:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/resources/manifest/ExportedArtifactsCollector.java
```diff
@@ -162,18 +161,28 @@ public void visitArrayType(ArrayType arrayType) {
...
         Optional<EnumAnnotation> enumAnnotation = stringType.getAnnotation(EnumAnnotation.class);
         if (enumAnnotation.isPresent()) {
-          clazz.set(getType(stringType));
+          exportedClasses.add(getType(stringType));
         }
       }
-    });
 
-    return Optional.ofNullable(clazz.get());
+      private Class<?> getClassForName(String name) {
```
marianogonzalez: use `org.mule.runtime.core.util.ClassUtils#loadClass(java.lang.String, java.lang.ClassLoader)` instead. The class loader used should be the TCCL

marianogonzalez:
code format mismatch

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/resources/manifest/ExportedArtifactsCollector.java
```diff
@@ -162,18 +165,28 @@ public void visitArrayType(ArrayType arrayType) {
...
         }
       }
-    });
 
-    return Optional.ofNullable(clazz.get());
+      private Class loadClass(String name){
+        try {
+          return ClassUtils.loadClass(name, extensionClassloader);
+        } catch (ClassNotFoundException e) {
+          throw new IllegalStateException(e);
```
marianogonzalez: IllegalArgumentException

PR #3326, opened 2016-09-20
MULE-10581: Batch process cannot resume after a crash, if there is a corrupted file in the ObjectStore

mbuchwald:


mbuchwald:
(Wait until cloudbees finishes running)

core/src/main/java/org/mule/util/store/PersistentObjectStorePartition.java
```diff
@@ -265,9 +265,20 @@ private synchronized void loadStoredKeysAndFileNames() throws ObjectStoreExcepti
...
+                {
+                    File file = files[i];
+                    StoreValue<T> storeValue = deserialize(file);
+                    realKeyToUUIDIndex.put(storeValue.getKey(), file.getName());
+                }
+                catch (ObjectStoreException e)
+                {
+                    if (logger.isWarnEnabled())
+                    {
+                        logger.warn("Could not deserialize an ObjectStore file");
```
marianogonzalez: This should also make it clear that the data will be lost... Careful with the language you use not to freak the user out

marianogonzalez: why not moving the file to a special directory instead? if you delete it we'll never have any tools to try to diagnose or have a sense of why the file is corrupted. Also, at the customer's eyes, moving the file is less violent than deleting it

core/src/test/java/org/mule/util/store/PersistentObjectStorePartitionTestCase.java
```diff
@@ -66,6 +68,24 @@ public void indicatesUnexistentKeyOnRetrieveError() throws ObjectStoreException
...
+        try
+        {
+            File.createTempFile("temp", ".obj", objectStoreFolder.getRoot());
+            partition.store(KEY, VALUE);
+            // Expect the new stored object, and the partition-descriptor file
+            assertEquals(2, objectStoreFolder.getRoot().listFiles().length);
+        }
+        catch (Exception e)
+        {
+            fail("Supposed to have skipped corrupted or unreadable files");
```
marianogonzalez: you're missing an assertion that the file actually got moved

core/src/main/java/org/mule/util/store/PersistentObjectStorePartition.java
```diff
@@ -266,8 +267,23 @@ private synchronized void loadStoredKeysAndFileNames() throws ObjectStoreExcepti
...
+                {
+                    if (logger.isWarnEnabled())
+                    {
+                        logger.warn(String.format(
+                                "Could not deserialize the ObjectStore file: %s. The file will be skipped and moved " +
+                                "to the Garbage folder", file.getName()));
+                    }
+                    String garbageFileName = file.getParentFile().getParentFile().getParent() + File.separator +
+                                             GARBAGE_FOLDER + File.separator + file.getParentFile().getName();
+                    FileUtils.moveFileToDirectory(file, new File(garbageFileName), true);
```
marianogonzalez: static imports

core/src/main/java/org/mule/util/store/PersistentObjectStorePartition.java
```diff
@@ -266,8 +267,23 @@ private synchronized void loadStoredKeysAndFileNames() throws ObjectStoreExcepti
...
+                }
+                catch (ObjectStoreException e)
+                {
+                    if (logger.isWarnEnabled())
+                    {
+                        logger.warn(String.format(
+                                "Could not deserialize the ObjectStore file: %s. The file will be skipped and moved " +
+                                "to the Garbage folder", file.getName()));
+                    }
+                    String garbageFileName = file.getParentFile().getParentFile().getParent() + File.separator +
```
marianogonzalez: isn't it best if you just calculate the delta between the app.home and the actual file. This logic is very fragile when it comes to changes in the deployment structure

PR #3330, opened 2016-09-20
MULE-10298: Deploy applications in parallel

pablokraan:
_ Adding synchronization to MuleDeploymentService API.
   On previous versions, synchronization was done on the deployers, but that was removed with the parallel deployment work (as it is not possible to have concurrent deployments with a unique lock).
   This lack of synchronization should not be a problem in the current version as the only client of this API is the agent that synchronizes the deployment before using it.
   However, is responsibility of the service to synchronize every call and let clients to synchronize only when they must execute many API calls

### DIFF
--- a/modules/launcher/src/main/java/org/mule/module/launcher/MuleDeploymentService.java
+++ b/modules/launcher/src/main/java/org/mule/module/launcher/MuleDeploymentService.java
@@ -7,6 +7,7 @@
 package org.mule.module.launcher;
 
 import static java.lang.System.getProperties;
+import static java.util.concurrent.TimeUnit.SECONDS;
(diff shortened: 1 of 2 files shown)


mohdajami:
Hi Pablo, we started getting this error on Mule EE in the last 3 days, could this be related?

Deployment     system     SYSTEM
Worker(xx.xx.xx.xx): Your application has failed with exception java.io.IOException: Another deployment operation is in progress

pablokraan:
Hi, 
I don't think this change could be related to your problem as is not included on any EE release as far as I know. 
The error you see is because there is a request to deploy an app while another deployment request is still in progress, that restriction is there since the beginning.
If you are having problems, you should raise a support ticket

PR #3331, opened 2016-09-20
MULE-10495: Improve test infrastructure for creating Events/Messages

elrodro83:


core-tests/src/test/java/org/mule/runtime/core/DefaultMuleEventTestCase.java
```diff
@@ -51,7 +52,7 @@
 
   @Before
   public void before() throws Exception {
-    flow = getTestFlow();
+    flow = getTestFlow(muleContext);
```
dfeist: Is it preferred to use static method in MuleTestUtils over method in AbstractMuleContextTestCase that does not require MuleContext? The later seems simple..

core-tests/src/test/java/org/mule/runtime/core/MuleEventTestCase.java
```diff
@@ -39,7 +43,10 @@
 
   @Test
   public void testEventSerialization() throws Exception {
-    Event event = getTestEvent("payload");
+    FlowConstruct flowConstruct = MuleTestUtils.getTestFlow(muleContext);
+    Event event = Event.builder(DefaultEventContext.create(flowConstruct, TEST_CONNECTOR))
```
dfeist: Can we push the creation of DefaultEventContext up to AbstractMuleContextTestCase and make it reusable?

dfeist: Also, when all that is being created is an event from a payload, why does the test need to have all this boiler plate code?

core-tests/src/test/java/org/mule/runtime/core/MuleEventTestCase.java
```diff
@@ -43,10 +38,7 @@
 
   @Test
   public void testEventSerialization() throws Exception {
-    FlowConstruct flowConstruct = MuleTestUtils.getTestFlow(muleContext);
-    Event event = Event.builder(DefaultEventContext.create(flowConstruct, TEST_CONNECTOR))
-        .message(InternalMessage.of("payload"))
-        .build();
+    Event event = eventBuilder().message(InternalMessage.of("payload")).build();
```
dfeist: Any reason this can't just be `Event event = event("payload");` or  `Event event = getTestEvent("payload")`?  I understand it makes use to use EventBuilder in a consistent way to how it's used in Mule for the non-trivial cases, but still don't see why we cant' simply the simplest cases further.  There is even a TEST_PAYLOAD or similar in AbstractMuleTestCase which can be used to create an event without even passing a payload.  Also ideally this event should be reused so that it can be asserted against and reused.

core-tests/src/test/java/org/mule/runtime/core/el/mvel/MVELExpressionLanguageTestCase.java
```diff
@@ -264,11 +255,7 @@ public void appTakesPrecedenceOverEverything() throws Exception {
   @Test
   public void messageTakesPrecedenceOverEverything() throws Exception {
     mvel.setAliases(Collections.singletonMap("message", "'other1'"));
-    FlowConstruct flowConstruct = MuleTestUtils.getTestFlow(muleContext);
-    Event event = Event.builder(DefaultEventContext.create(flowConstruct, TEST_CONNECTOR))
-        .message(InternalMessage.of(""))
-        .addVariable("message", "other2")
-        .build();
+    Event event = eventBuilder().message(InternalMessage.of("")).addVariable("message", "other2").build();
```
dfeist: If the message is often not important in tests, then can we make it optional and use a default/null message instead?  This may mean a slightly different builder for tests though, so maybe not best idea..  Alternatively a NULL_MESSAGE or EMPTY_MESSAGE constant could help, and not require extending builder in test infrasturucture.

core-tests/src/test/java/org/mule/runtime/core/mule/enricher/MessageEnricherTestCase.java
```diff
@@ -179,11 +159,7 @@ public void propagatesVariables() throws Exception {
     enricher.setEnrichmentMessageProcessor(event -> {
       return Event.builder(event).message(InternalMessage.builder(event.getMessage()).payload("enriched").build()).build();
     });
-    FlowConstruct flowConstruct = MuleTestUtils.getTestFlow(muleContext);
-    Event in = Event.builder(DefaultEventContext.create(flowConstruct, TEST_CONNECTOR))
-        .message(InternalMessage.of(""))
-        .addVariable("flowFoo", "bar")
-        .build();
+    Event in = eventBuilder().message(InternalMessage.of("")).addVariable("flowFoo", "bar").build();
```
dfeist: Just thinking out loud, but another approach to exposing builder and requiring `.build` could be to take a builder function.
`event(payload)`
`event(builder -> builder.addVariable()`

PR #3332, opened 2016-09-21
MULE-10403: FTP Extension: Write operation doesn't work using SFTP wh…

pablomusumeci:
…en the destination is a new file.


extensions/ftp/src/main/java/org/mule/extension/ftp/internal/sftp/connection/SftpClient.java
```diff
@@ -113,7 +114,7 @@ public SftpFileAttributes getAttributes(Path path) {
     try {
       return new SftpFileAttributes(path, sftp.stat(path.toString()));
     } catch (SftpException e) {
-      if (e.getMessage().contains(FileNotFoundException.class.getName())) {
+      if (e.id == FILE_NOT_FOUND_ERROR_CODE) {
```
marianogonzalez: Unit test for the SftpClient class. Using a mock client, create a test which validates that null is returned when the error code is 2 and another one which throws exception otherwise

PR #3336, opened 2016-09-21
MULE-10515: Fail to update blob with file:read result

marianogonzalez:


extensions/db/src/main/java/org/mule/extension/db/api/param/JdbcType.java
```diff
@@ -56,30 +57,30 @@
...
-                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               .getInstance());
+                                                                                                                                                                                                                                                              new BlobDbType(Types.BLOB,
```
pablokraan: ?

extensions/db/src/main/java/org/mule/extension/db/internal/domain/type/BlobDbType.java
```diff
@@ -0,0 +1,47 @@
...
+   * Sets the parameter accounting for the case in which the {@code value} is
+   * an {@link InputStream}, in which case it is consumed into a {@code byte[]} and
+   * set.
+   * {@inheritDoc}
+   */
+  @Override
+  public void setParameterValue(PreparedStatement statement, int index, Object value) throws SQLException {
+    if (value instanceof InputStream) {
+      try {
+        value = toByteArray((InputStream) value);
```
pablokraan: Can't you create a blob instance? (connection.createBlob)

extensions/db/src/main/java/org/mule/extension/db/internal/domain/type/MetadataDbTypeManager.java
```diff
@@ -83,10 +83,12 @@ protected void initialise(DbConnection connection) {
       while (resultSetIterator.hasNext()) {
         Map<String, Object> typeRecord = resultSetIterator.next();
 
-        Number data_type = (Number) typeRecord.get(METADATA_TYPE_ID_COLUMN);
-        String type_name = (String) typeRecord.get(METADATA_TYPE_NAME_COLUMN);
+        Number dataType = (Number) typeRecord.get(METADATA_TYPE_ID_COLUMN);
+        String typeName = (String) typeRecord.get(METADATA_TYPE_NAME_COLUMN);
 
-        ResolvedDbType resolvedDbType = new ResolvedDbType(data_type.intValue(), type_name);
+        DbType resolvedDbType = dataType.equals(Types.BLOB)
```
pablokraan: I don't think this is the right place for this logic. Maybe it could go into JdbcTypes

extensions/db/src/main/java/org/mule/extension/db/api/param/JdbcType.java
marianogonzalez: don't look at me... talk to will weathon and his formatter

extensions/db/src/main/java/org/mule/extension/db/internal/domain/type/MetadataDbTypeManager.java
marianogonzalez: I disagree. This looks like the place to do it.

extensions/db/src/main/java/org/mule/extension/db/internal/domain/type/BlobDbType.java
marianogonzalez: done

extensions/db/src/main/java/org/mule/extension/db/internal/domain/type/MetadataDbTypeManager.java
pablokraan: There is something wrong here, because this code is not taking in account other types (like struct and array), same problem exists on 3.x :(

PR #3338, opened 2016-09-21
MULE-10299: Remove MuleEvent dependency on FlowConstruct.

elrodro83:
* Fix flowConstruct propagation among templates/processors

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/exception/TemplateOnErrorHandler.java
+++ b/core/src/main/java/org/mule/runtime/core/exception/TemplateOnErrorHandler.java
@@ -15,7 +15,6 @@ import org.mule.runtime.core.api.MuleContext;
 import org.mule.runtime.core.api.MuleException;
 import org.mule.runtime.core.api.connector.NonBlockingReplyToHandler;
 import org.mule.runtime.core.api.construct.FlowConstruct;
-import org.mule.runtime.core.api.construct.FlowConstructAware;
 import org.mule.runtime.core.api.exception.MessagingExceptionHandler;
 import org.mule.runtime.core.api.exception.MessagingExceptionHandlerAcceptor;
 import org.mule.runtime.core.api.lifecycle.InitialisationException;
@@ -181,9 +180,6 @@ public abstract class TemplateOnErrorHandler extends AbstractExceptionListener
         new DefaultMessageProcessorChainBuilder(this.flowConstruct);
     try {
       configuredMessageProcessors = defaultMessageProcessorChainBuilder.chain(getMessageProcessors()).build();
-      if (configuredMessageProcessors instanceof FlowConstructAware) {
-        ((FlowConstructAware) configuredMessageProcessors).setFlowConstruct(flowConstruct);
-      }
     } catch (MuleException e) {
       throw new InitialisationException(e, this);
     }
--- a/core/src/main/java/org/mule/runtime/core/exception/TemplateOnErrorHandler.java
+++ b/core/src/main/java/org/mule/runtime/core/exception/TemplateOnErrorHandler.java
@@ -8,18 +8,20 @@ package org.mule.runtime.core.exception;
 
 import static org.mule.runtime.core.context.notification.ExceptionStrategyNotification.PROCESS_END;
 import static org.mule.runtime.core.context.notification.ExceptionStrategyNotification.PROCESS_START;
+
 import org.mule.runtime.core.VoidMuleEvent;
-import org.mule.runtime.core.api.MuleContext;
 import org.mule.runtime.core.api.Event;
+import org.mule.runtime.core.api.MuleContext;
 import org.mule.runtime.core.api.MuleException;
-import org.mule.runtime.core.api.message.InternalMessage;
 import org.mule.runtime.core.api.connector.NonBlockingReplyToHandler;
 import org.mule.runtime.core.api.construct.FlowConstruct;
+import org.mule.runtime.core.api.construct.FlowConstructAware;
 import org.mule.runtime.core.api.exception.MessagingExceptionHandler;
 import org.mule.runtime.core.api.exception.MessagingExceptionHandlerAcceptor;
 import org.mule.runtime.core.api.lifecycle.InitialisationException;
-import org.mule.runtime.core.api.processor.Processor;
+import org.mule.runtime.core.api.message.InternalMessage;
 import org.mule.runtime.core.api.processor.MessageProcessorChain;
+import org.mule.runtime.core.api.processor.Processor;
 import org.mule.runtime.core.context.notification.ExceptionStrategyNotification;
 import org.mule.runtime.core.management.stats.FlowConstructStatistics;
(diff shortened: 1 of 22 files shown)


core/src/main/java/org/mule/runtime/core/exception/TemplateOnErrorHandler.java
```diff
@@ -179,6 +181,9 @@ protected void doInitialise(MuleContext muleContext) throws InitialisationExcept
         new DefaultMessageProcessorChainBuilder(this.flowConstruct);
     try {
       configuredMessageProcessors = defaultMessageProcessorChainBuilder.chain(getMessageProcessors()).build();
+      if (configuredMessageProcessors instanceof FlowConstructAware) {
```
dfeist: It's be good to add a utility method to LifecycleUtils for this, so we are not repeating it everywhere.

core/src/main/java/org/mule/runtime/core/routing/UntilSuccessful.java
```diff
@@ -267,6 +267,7 @@ public Processor getDlqMP() {
   public Processor getRoute() {
     final DefaultMessageProcessorChain chain = DefaultMessageProcessorChain.from(muleContext, routes.get(0));
     chain.setTemplateMuleContext(muleContext);
```
dfeist: This shouldn't be required.

core/src/main/java/org/mule/runtime/core/routing/UntilSuccessful.java
```diff
@@ -267,6 +267,7 @@ public Processor getDlqMP() {
...
+    chain.setFlowConstruct(flowConstruct);
```
dfeist: Is there no way to initialize this route when object is constructed and then implemented `getOwnedMessageProcessors()` for lifecycle?

core/src/main/java/org/mule/runtime/core/exception/TemplateOnErrorHandler.java
```diff
@@ -179,6 +181,9 @@ protected void doInitialise(MuleContext muleContext) throws InitialisationExcept
...
+        ((FlowConstructAware) configuredMessageProcessors).setFlowConstruct(flowConstruct);
```
dfeist: Shouldn't this be done via `getOwnedMessageProcessors`?

core/src/main/java/org/mule/runtime/core/routing/MessageFilter.java
```diff
@@ -124,6 +124,7 @@ public void setMuleContext(MuleContext context) {
 
   @Override
   public void setFlowConstruct(FlowConstruct flowConstruct) {
+    super.setFlowConstruct(flowConstruct);
     if (!onUnacceptedFlowConstruct && unacceptedMessageProcessor instanceof FlowConstructAware) {
       ((FlowConstructAware) unacceptedMessageProcessor).setFlowConstruct(flowConstruct);
```
dfeist: Can `getOwnedMessageProcessor` be used?

tests/integration/src/test/java/org/mule/test/core/context/notification/processors/MessageProcessorNotificationPathTestCase.java
```diff
@@ -63,7 +65,7 @@ public void scopes() throws Exception {
     testFlowPaths("enricher", "/0", "/0/0", "/1", "/1/0", "/1/0/0", "/1/0/1");
     testFlowPaths("until-successful", "/0", "/0/0", "/1");
     testFlowPaths("until-successful-with-processor-chain", "/0", "/0/0", "/0/0/0", "/0/0/1", "/1");
-    testFlowPaths("until-successful-with-enricher", "/0", "/0/0", "/0/0/0", "/1");
+    testFlowPaths("until-successful-with-enricher", "/0", "/0/0", "/0/0/0", "/0/0/0/0", "/0/0/1", "/1");
```
dfeist: Why does assertion changes, or was it wrong before due to a bug?

tests/integration/src/test/resources/org/mule/test/integration/notifications/message-processor-notification-test-flow.xml
```diff
@@ -152,10 +152,10 @@
 
     <flow name="all3">
         <all>
-            <logger message="check"/>
-            <logger message="check"/>
+            <logger message="check1" level="WARN"/>
```
dfeist: Can we reduce level now?

tests/integration/src/test/java/org/mule/test/core/context/notification/processors/MessageProcessorNotificationPathTestCase.java
elrodro83: I had to modify a config file that was causing a test to be flaky. This change accounts for that change

core/src/main/java/org/mule/runtime/core/routing/UntilSuccessful.java
elrodro83: Sadly no, this initalization is lazy and happens after initialize or setFlowConstruct have been called

core/src/main/java/org/mule/runtime/core/routing/MessageFilter.java
elrodro83: there is no such method on this context

core/src/main/java/org/mule/runtime/core/routing/UntilSuccessful.java
dfeist: There is no reason for this to be lazy though is there?

core/src/main/java/org/mule/runtime/core/routing/MessageFilter.java
dfeist: I thought we could make filter a MessageProessorOwner, but we can't because intercepting MP's aren't supported.

core/src/main/java/org/mule/runtime/core/exception/TemplateOnErrorHandler.java
dfeist: Looking at this closer, using getOwnedMessageProcessors() is not suitable here.

core/src/main/java/org/mule/runtime/core/processor/chain/DefaultMessageProcessorChainBuilder.java
```diff
@@ -56,9 +56,12 @@ public DefaultMessageProcessorChainBuilder(FlowConstruct flowConstruct) {
   public MessageProcessorChain build() throws MuleException {
     LinkedList<Processor> tempList = new LinkedList<>();
 
+    final List<Processor> builtProcessors = new ArrayList<>(processors.size());
+
     // Start from last but one message processor and work backwards
     for (int i = processors.size() - 1; i >= 0; i--) {
       Processor processor = initializeMessageProcessor(processors.get(i));
+      builtProcessors.add(processor);
```
dfeist: Can you add using index, just so we conserve order just in case..

PR #3345, opened 2016-09-22
MULE-10530: Rename and make optional the "key" attribute of HTTP request operation

afelisatti:


extensions/http/src/test/resources/basic-http-config.xml
```diff
@@ -14,7 +14,7 @@
     </httpn:listener-config> 
     
     <flow name="client">
-        <httpn:request path="/" config-ref="reqConfig" key="ANY">
+        <httpn:request path="/" config-ref="reqConfig" outputType="ANY">
```
marianogonzalez: shouldn't you have a test in which you set this to a non default value and assert over the output metadata?

PR #3351, opened 2016-09-22
MULE-10493: Refactor DB Connector parameter's structure

estebanwasinger:


extensions/db/src/main/java/org/mule/extension/db/api/config/DerbyDataSourceConfig.java
```diff
@@ -0,0 +1,139 @@
...
+  private String database;
+
+  /**
+   * Specifies the type of SubsubProtocol to be used by Derby. The avaiable options are:
+   * 'directory', 'memory', 'classpath' and 'jar.
+   */
+  @Parameter
+  @Optional(defaultValue = "directory")
+  @Placement(group = CONNECTION, order = 2)
+  private String subsubProtocol;
```
marianogonzalez: subsub?

extensions/db/src/main/java/org/mule/extension/db/api/config/DerbyDataSourceConfig.java
```diff
@@ -0,0 +1,139 @@
...
+
+  /**
+   * Indicates if the DataBase should be created if it this not exist.
```
marianogonzalez: DataBase -> Database

extensions/db/src/main/java/org/mule/extension/db/api/config/MySqlDbUtils.java
```diff
@@ -0,0 +1,83 @@
...
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Map;
+
+/**
+ * Utility class for MySQL Connection
+ *
+ * @since 4.0
+ */
+final class MySqlDbUtils {
```
marianogonzalez: the logic that was moved here applies to other vendor specific connections such as oracle. Why move them to a mysql Specfic utility class?

extensions/db/src/main/java/org/mule/extension/db/internal/domain/connection/datasource/DataSourceConnectionProvider.java
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * Reference to a JDBC {@link DataSource} object. This object is typically created using Spring.
+   * When using XA transactions, an {@link XADataSource} object must be provided.
+   */
+  @Parameter
+  @Alias("dataSource")
+  @Placement(group = CONNECTION)
+  //This, non optimal, name is due that this field clashes with the one defined in DbConnectionProvider
+  private DataSource aDataSource;
```
marianogonzalez: dataSourceRef?

extensions/db/src/main/java/org/mule/extension/db/internal/domain/connection/datasource/DataSourceConnectionProvider.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * {@link ConnectionProvider} implementation which creates DB connections from a referenced {@link
+ * DataSource}
+ *
+ * @since 4.0
+ */
+@DisplayName("Data Source Connection")
+@Alias("data-source")
+public class DataSourceConnectionProvider extends DbConnectionProvider {
```
marianogonzalez: The name should make it clear that this is about referencing an existing dataSource... Otherwise it seems to imply that only this one uses a DataSource

extensions/db/src/main/java/org/mule/extension/db/internal/domain/connection/mysql/MySqlConnectionParameters.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+@ExclusiveOptionals(isOneRequired = true)
+public class MySqlConnectionParameters implements DbConnectionParameters {
+
+  private static final String MYSQL_DRIVER_CLASS = "com.mysql.jdbc.Driver";
+
+  @Optional
+  @Parameter
+  @Alias("mySqlParameters")
+  private MySqlDataSourceConfig mySqlParameters;
```
marianogonzalez: why this? it doesn't make sense. Why can't you just have generic connection parameters as we used to. One thing is to have vendor specific providers , but common things should be handled in the same way

extensions/db/src/test/resources/integration/config/datasource-invalid-config.xml
```diff
@@ -7,7 +7,10 @@
 
 
     <db:config name="invalidDbConfig">
-        <db:connection dataSource="#[flowVars.jdbcDataSource]" user="This parameter violates the optional exclusiveness condition"/>
+        <db:my-sql-connection >
+            <db:database-url url="url"/>
```
marianogonzalez: what's the point of being able to specify a url in a vendor specific config?

extensions/db/src/main/java/org/mule/extension/db/api/config/DerbyDataSourceConfig.java
```diff
@@ -0,0 +1,139 @@
...
+  /**
+   * Specifies a list of custom key-value connectionProperties for the config.
+   */
+  @Parameter
+  @Optional
+  @Placement(tab = ADVANCED)
+  private Map<String, String> connectionProperties = new HashMap<>();
+
+  @ParameterGroup
+  private AdvancedDbParameters parametersConfig;
```
marianogonzalez: this should be on a base class. You shouldn't be able to make a provider which doesn't have this

extensions/db/src/main/java/org/mule/extension/db/api/config/MySqlDataSourceConfig.java
```diff
@@ -0,0 +1,113 @@
...
+import static org.mule.extension.db.api.config.MySqlDbUtils.getEffectiveUrl;
+import static org.mule.runtime.extension.api.annotation.param.display.Placement.ADVANCED;
+import static org.mule.runtime.extension.api.annotation.param.display.Placement.CONNECTION;
+
+/**
+ * {@link DataSourceConfig} implementation for MySQL databases.
+ *
+ * @since 4.0
+ */
+public final class MySqlDataSourceConfig implements DataSourceConfig {
```
marianogonzalez: The Config suffix is confusing (applies to several classes)

extensions/db/src/main/java/org/mule/extension/db/api/config/DerbyDataSourceConfig.java
estebanwasinger: just derby...

extensions/db/src/main/java/org/mule/extension/db/api/config/MySqlDbUtils.java
estebanwasinger: In a short investigation, Oracle JDBC URLs are not 100% compatibles with MySql ones. 
We should review this when giving support to OracleDB or another DB and maybe think in a better abstraction of `ParametersToURL` translators.

extensions/db/src/main/java/org/mule/extension/db/internal/domain/connection/derby/DerbyConnectionParameters.java
```diff
@@ -6,49 +6,115 @@
...
+  @Placement(group = CONNECTION, order = 3)
+  private boolean create;
 
+  /**
+   * Specifies a list of custom key-value connectionProperties for the config.
+   */
+  @Parameter
+  @Optional
+  @Placement(tab = ADVANCED)
+  private Map<String, String> connectionProperties = new HashMap<>();
```
estebanwasinger: This can't be generic for all the ConnectionProvider due that the DataSource constructor that we use doesn't let you configure additional properties, so the unique way to do it is adding these properties to the JDBC URL which differs between each DB.

PR #3367, opened 2016-09-27
MULE-10574: Add model validator for parameters that have the same name as then entries of a map

juandesi:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/NameClashModelValidator.java
```diff
@@ -213,6 +258,68 @@ private void validateClash(String existingNamingModel, String newNamingModel, St
                                                          typeOfNewNamingModel));
       }
     }
+
+    private void validateSingularizingClash(ParameterizedModel model) {
+      List<ParameterModel> parameters = model.getParameterModels();
+      parameters.forEach(
+                         p -> p.getType().accept(new MetadataTypeVisitor() {
```
alegmarra: create the visitor once, maybe even using a method to be more clear on what it is for. This way we avoid creating it N times, and also simplify this scope

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/NameClashModelValidator.java
```diff
@@ -213,6 +258,68 @@ private void validateClash(String existingNamingModel, String newNamingModel, St
...
+                           public void visitArrayType(ArrayType arrayType) {
+                             validateSingular();
+                           }
+
+                           @Override
+                           public void visitDictionary(DictionaryType dictionaryType) {
+                             validateSingular();
+                           }
+
+                           private void validateSingular() {
```
alegmarra: I think is better to put this outside the visitor class. also rename to something more expressive on what is doing?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/NameClashModelValidator.java
```diff
@@ -213,6 +258,68 @@ private void validateClash(String existingNamingModel, String newNamingModel, St
...
+                             String singularName = singularize(p.getName());
+                             boolean allowInline = p.getModelProperty(XmlHintsModelProperty.class)
+                                 .map(XmlHintsModelProperty::allowsInlineDefinition)
+                                 .orElse(true);
+
+                             if (singularName.equals(p.getName()) || !allowInline) {
+                               return;
+                             }
+
+                             Optional<ParameterModel> clashingParam =
```
alegmarra: all this, including the exception throwing => new method
validateNameClashWithParameters?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/NameClashModelValidator.java
```diff
@@ -213,6 +258,68 @@ private void validateClash(String existingNamingModel, String newNamingModel, St
...
+                               throw new IllegalModelDefinitionException(format("Extension '%s' defines an %s of name '%s' which contains a '%s' "
+                                   + "parameter '%s' that when singularized clashes with another "
+                                   + "parameter '%s' in the same %s", extensionName,
+                                                                                describedReference, model.getName(),
+                                                                                type.getSimpleName(),
+                                                                                p.getName(), clashingParam.get().getName(),
+                                                                                describedReference));
+                             }
+
+                             Optional<String> nameClash =
```
alegmarra: all this, including the exception throwing => new method
validateNameClashWithTopLevelElements?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/NameClashModelValidator.java
```diff
@@ -168,16 +184,45 @@ private void validateTopLevelParameter(ParameterModel parameter, ParameterizedMo
...
+            .map(k -> {
+              DescribedParameter reference = singularizedObjects.get(k);
+              Collection<Named> namedElements = names.get(k);
+              if (!namedElements.isEmpty()) {
+                Named next = namedElements.iterator().next();
+                return format(
+                              "%s name '%s' contains parameter '%s' that clash when singularized with parameter '%s' in '%s'",
+                              reference.parent.getDescription(), reference.parent.getName(), reference.getName(), k,
+                              next.getName());
+              } else {
```
alegmarra: filter this when creating the singularClashes map, and also see if is not better just to have a Map<String, Set> instead of <String, Named>, specially since the value is never used

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/NameClashModelValidator.java
```diff
@@ -66,13 +75,15 @@ public void validate(ExtensionModel model) throws IllegalModelDefinitionExceptio
...
     private final ExtensionModel extensionModel;
     private final Set<DescribedReference<Named>> namedObjects = new HashSet<>();
+    private final Map<String, DescribedParameter> singularizedObjects = new HashMap<>();
     private final Multimap<String, TopLevelParameter> topLevelParameters = LinkedListMultimap.create();
 
     public ValidationDelegate(ExtensionModel extensionModel) {
       this.extensionModel = extensionModel;
     }
 
     private void validate(ExtensionModel extensionModel) throws IllegalModelDefinitionException {
```
estebanwasinger: The actual logic doesn't check clashing between `@Configuration( name= "SomeConfig") ` and `@Configuration( name= "some-config")`

alegmarra: Good catch, in the DslSyntaxResolver we do `.withElementName(hyphenize(component.getName()))` for parameters. We can move this to `NameUtils.getElementName` and reuse it in both places.

marianogonzalez: was this resolved?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/NameClashModelValidator.java
```diff
@@ -168,27 +187,29 @@ private void validateTopLevelParameter(ParameterModel parameter, ParameterizedMo
...
-      return repeatedNames;
+      return namedObject.stream()
+          .filter(parameter -> !names.add(parameter.getName()))
+          .map(Named::getName).collect(toSet());
     }
 
     private void validateNameClashes(Collection<? extends Named>... collections) {
       Multimap<String, Named> names = LinkedListMultimap.create();
-      stream(collections).flatMap(Collection::stream).forEach(named -> names.put(named.getName(), named));
+      stream(collections).flatMap(Collection::stream).forEach(named -> names.put(NameUtils.hyphenize(named.getName()), named));
```
marianogonzalez: static import

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/NameClashModelValidator.java
```diff
@@ -199,12 +220,34 @@ private void validateNameClashes(Collection<? extends Named>... collections) {
             i++;
           }
 
-          errorMessage.append(format(" which name is '%s'. Names should be unique", entry.getKey()));
+          errorMessage.append(format(" which alias name is '%s'. Names should be unique", entry.getKey()));
```
marianogonzalez: why did you added the alias concept in the message? It doesn't always apply and when it does it's kind of obvious...

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/NameClashModelValidator.java
```diff
@@ -199,12 +220,34 @@ private void validateNameClashes(Collection<? extends Named>... collections) {
...
           throw new IllegalModelDefinitionException(errorMessage.toString());
         }
       });
     }
 
+    private void validateSingularizeNameClashesWithTopLevels() {
+      Map<String, Collection<TopLevelParameter>> singularClashes = topLevelParameters.keySet().stream()
```
marianogonzalez: this would be much more efficient and clear if you iterate over the entry set instead of the key set

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/NameClashModelValidator.java
```diff
@@ -199,12 +220,34 @@ private void validateNameClashes(Collection<? extends Named>... collections) {
...
+        singularClashes.entrySet().forEach(e -> {
+          DescribedParameter reference = singularizedObjects.get(e.getKey());
+          e.getValue().forEach(tp -> errorMessages.add(format(
+                                                              "%s '%s' contains parameter '%s' that clash when singularized with parameter '%s' from '%s'",
+                                                              reference.parent.getDescription(), reference.parent.getName(),
+                                                              reference.getName(),
+                                                              tp.getName(), tp.ownerType)));
+        });
+
+        throw new IllegalModelDefinitionException(format("Extension '%s' contains %d parameters that clash when singularized. %s",
```
marianogonzalez: "parameters that clash when singularized" is not a descriptive message. It should just say that it contains parameter names which clash when transformed into XML DSL language

marianogonzalez: or something like that :)

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/NameClashModelValidator.java
```diff
@@ -213,6 +256,49 @@ private void validateClash(String existingNamingModel, String newNamingModel, St
...
+                             boolean allowInline = p.getModelProperty(XmlHintsModelProperty.class)
+                                 .map(XmlHintsModelProperty::allowsInlineDefinition)
+                                 .orElse(true);
+
+                             if (!singularName.equals(p.getName()) && allowInline) {
+                               singularizedObjects.put(singularName, new DescribedParameter(p, model));
+                               Optional<ParameterModel> clashParam =
+                                   parameters.stream().filter(p -> p.getName().equals(singularName)).findAny();
+                               String describedReference = new DescribedReference<>(model).getDescription();
+                               Class<Object> type = getType(p.getType());
```
marianogonzalez: there isn't necessarily a class behind this. This validator should also work for smart connectors

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/NameClashModelValidator.java
```diff
@@ -213,6 +256,49 @@ private void validateClash(String existingNamingModel, String newNamingModel, St
...
+
+                               if (clashParam.isPresent()) {
+                                 throw new IllegalModelDefinitionException(
+                                                                           format("Extension '%s' defines an %s of name '%s' which contains a '%s' parameter '%s' that when singularized "
```
marianogonzalez: unclear error message

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/introspection/validation/NameClashModelValidatorTestCase.java
```diff
@@ -91,118 +107,193 @@ public void validModel() {
     validate();
   }
 
-  @Test(expected = IllegalModelDefinitionException.class)
+  @Test
   public void operationClashingWithConfig() {
+    exception.expect(IllegalModelDefinitionException.class);
```
marianogonzalez: instead of putting this line at the beginning of every method, initialise the rule to this value and only set it to `none()` when you expect the validation to pass

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/NameClashModelValidator.java
juandesi: yes

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/NameClashModelValidator.java
marianogonzalez: got it

PR #3368, opened 2016-09-27
MULE-10643: ResourceBundle class loader leak when undeploying applica…

gsfernandes:
…tion

- Clean up ResourceBundle cache for removing references to class loader being disposed.

### DIFF
--- a/modules/launcher/src/main/java/org/mule/module/launcher/artifact/DefaultResourceReleaser.java
+++ b/modules/launcher/src/main/java/org/mule/module/launcher/artifact/DefaultResourceReleaser.java
@@ -12,9 +12,11 @@ import static java.lang.management.ManagementFactory.getPlatformMBeanServer;
 import static java.sql.DriverManager.deregisterDriver;
 import static java.sql.DriverManager.getDrivers;
 
+import java.lang.reflect.Field;
 import java.sql.Driver;
 import java.util.Enumeration;
 import java.util.Hashtable;
+import java.util.ResourceBundle;
 
 import javax.management.MBeanServer;
 import javax.management.ObjectName;
@@ -32,6 +34,39 @@ public class DefaultResourceReleaser implements ResourceReleaser
     public void release()
     {
         deregisterJdbcDrivers();
+
+        cleanUpResourceBundle();
+    }
+
+    private void cleanUpResourceBundle()
+    {
+        try
+        {
+            ResourceBundle.clearCache(this.getClass().getClassLoader());
+        }
+        catch (Exception e)
+        {
+            logger.warn("Couldn't clean up ResourceBundle. This can cause a memory leak.", e);
+        }
+
+        try
+        {
+            // Even the cache is cleaned up there is a reference to a CacheKey that is hold by ResouceBundle.NONEXISTENT_BUNDLE
+            // so we have to cleanup this reference too instead of just leaving it and waiting for a next
+            // ResourceBundle.getBundle() call to a bundle that is missing that would change this reference
+            Field nonExistentBundleField = ResourceBundle.class.getDeclaredField("NONEXISTENT_BUNDLE");
+            nonExistentBundleField.setAccessible(true);
+
+            ResourceBundle resourceBundle = (ResourceBundle) nonExistentBundleField.get(null);
+
+            Field cacheKeyField = ResourceBundle.class.getDeclaredField("cacheKey");
+            cacheKeyField.setAccessible(true);
+            cacheKeyField.set(resourceBundle, null);
+        }
+        catch (Exception e)
+        {
+            logger.warn("Couldn't clean up ResourceBundle references.  This can cause a memory leak.", e);
+        }
     }
 
     private void deregisterJdbcDrivers()
--- a/modules/launcher/src/test/java/org/mule/module/launcher/ResourceReleaserTestCase.java
+++ b/modules/launcher/src/test/java/org/mule/module/launcher/ResourceReleaserTestCase.java
@@ -10,20 +10,27 @@ import static java.sql.DriverManager.deregisterDriver;
 import static java.sql.DriverManager.getDrivers;
 import static java.sql.DriverManager.registerDriver;
 import static java.util.Collections.list;
+import static java.util.Locale.getDefault;
 import static org.hamcrest.CoreMatchers.hasItem;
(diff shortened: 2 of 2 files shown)


modules/launcher/src/main/java/org/mule/module/launcher/artifact/DefaultResourceReleaser.java
```diff
@@ -32,6 +36,31 @@
...
         deregisterJdbcDrivers();
+
+        cleanUpResourceBundle();
+    }
+
+    private void cleanUpResourceBundle()
+    {
+        try
+        {
+            final Method clearCacheMethod = ResourceBundle.class.getMethod("clearCache", ClassLoader.class);
```
pablokraan: Commit message says "Uses reflection just in case if a Java version < 6" we know that that won't be the case on 3.x, 3.8.x and 3.7.x

modules/launcher/src/test/java/org/mule/module/launcher/ResourceReleaserTestCase.java
```diff
@@ -91,6 +99,37 @@ public void deregisterJdbcDriversSameClassLoaders() throws Exception
...
+
+        Map beforeCacheList = newHashMap();
+
+        Field cacheListField = ResourceBundle.class.getDeclaredField("cacheList");
+        cacheListField.setAccessible(true);
+        beforeCacheList.putAll((Map) cacheListField.get(null));
+
+        try
+        {
+            ResourceBundle.getBundle("aBundle", getDefault(), classLoader);
```
pablokraan: should fail if the resource is found

modules/launcher/src/test/java/org/mule/module/launcher/ResourceReleaserTestCase.java
```diff
@@ -91,6 +99,37 @@ public void deregisterJdbcDriversSameClassLoaders() throws Exception
...
+        }
+        catch (MissingResourceException e)
+        {
+            // OK
```
pablokraan: OK -> Expected

modules/launcher/src/test/java/org/mule/module/launcher/ResourceReleaserTestCase.java
```diff
@@ -91,6 +99,37 @@ public void deregisterJdbcDriversSameClassLoaders() throws Exception
...
+        }
+        Map actualCacheList = (Map) cacheListField.get(null);
+        int deltaForNonExistentBundle = actualCacheList.size() - beforeCacheList.size();
+
+        classLoader.dispose();
+
+        actualCacheList = (Map) cacheListField.get(null);
+
+        assertThat(actualCacheList.size(), equalTo(beforeCacheList.size() + deltaForNonExistentBundle));
+        assertThat(actualCacheList.entrySet(), (Matcher) hasItems(beforeCacheList.entrySet().toArray()));
```
pablokraan: Is the (Matcher) cast required?

modules/launcher/src/main/java/org/mule/module/launcher/artifact/DefaultResourceReleaser.java
gsfernandes: You were right, not need to use reflection here.

modules/launcher/src/test/java/org/mule/module/launcher/ResourceReleaserTestCase.java
gsfernandes: Yes, added fail() after this line.

modules/launcher/src/test/java/org/mule/module/launcher/ResourceReleaserTestCase.java
gsfernandes: This question made me think about how I was testing this and how I implemented the clean up for the reference from the NONEXISTENT_BUNDLE, so I changed the implementation to instead of asking for a non existent bundle just use reflection to nullate the reference and now the test is readable and the state of the ResourceBundle after dispose doesn't have a "dirty" non-existent bundle. Please review this changes and let me know.

modules/launcher/src/main/java/org/mule/module/launcher/artifact/DefaultResourceReleaser.java
```diff
@@ -32,6 +34,37 @@
...
+        cleanUpResourceBundle();
+    }
+
+    private void cleanUpResourceBundle()
+    {
+        try
+        {
+            ResourceBundle.clearCache(this.getClass().getClassLoader());
+        }
+        catch(Exception e)
```
pablokraan: Fix format

modules/launcher/src/main/java/org/mule/module/launcher/artifact/DefaultResourceReleaser.java
```diff
@@ -32,6 +34,37 @@
...
+        {
+            logger.warn("Couldn't clean up ResourceBundle", e);
+        }
+
+        try
+        {
+            // NONEXISTENT_BUNDLE leaves a cacheKey referencing to this classloader
+            Field nonExistentBundleField = ResourceBundle.class.getDeclaredField("NONEXISTENT_BUNDLE");
```
pablokraan: I don't get why this is required if the cache was cleaned up previously. Can you add a comment explaining the reasons?

modules/launcher/src/main/java/org/mule/module/launcher/artifact/DefaultResourceReleaser.java
```diff
@@ -32,6 +34,37 @@
...
+
+            ResourceBundle resourceBundle = (ResourceBundle) nonExistentBundleField.get(null);
+
+            Field cacheKeyField = ResourceBundle.class.getDeclaredField("cacheKey");
+            cacheKeyField.setAccessible(true);
+            cacheKeyField.set(resourceBundle, null);
+        }
+        catch (Exception e)
+        {
+            logger.warn("Couldn't clean up reference from NONEXISTENT_BUNDLE on ResourceBundle", e);
```
pablokraan: NONEXISTENT_BUNDLE is not something the user must know about. Message should say something like "Unable to clean up ResourceBundle references, this can cause a memory leak" or something like that

PR #3371, opened 2016-09-28
MULE-10533: When using a list of maps, the map parameter items xml tag is not recognized by the parser

juandesi:


tests/test-extensions/vegan-extension/src/main/java/org/mule/test/vegan/extension/VeganFidelityOperation.java
```diff
@@ -19,4 +22,8 @@ public String tryEat(@XmlHints(allowReferences = false) Object food) {
     throw new IllegalArgumentException("I SHALL NEVER EAT " + food.toString());
   }
 
+  public String tryToEatThisListOfMaps(List<Map<String, String>> food) {
```
alegmarra: does it affect the schema if we have a pojo that is only used by this map generic? I don't think it would, but if it does is a bug in the schema builder and we should update it too in order to avoid inconsistent schema generation.

marianogonzalez: moving forward we will no longer generate XML support for complex parameters in operations but rather rely on DW for that. I wouldn't stress too much about this

marianogonzalez:
this is just the test case. Where' s the actual fix?

alegmarra:
I think the fix is in the dsl

PR #3381, opened 2016-09-29
MULE-10647: Reconnection policy not present in XSD when ConnectionPro…

marianogonzalez:
…vider is not pooling


core/src/main/java/org/mule/runtime/core/internal/connection/ReconnectableConnectionProviderWrapper.java
```diff
@@ -6,45 +6,48 @@
...
+public class ReconnectableConnectionProviderWrapper<C> extends ConnectionProviderWrapper<C> {
 
   private final boolean disableValidation;
   private final RetryPolicyTemplate retryPolicyTemplate;
 
-  public CachedConnectionProviderWrapper(ConnectionProvider<Connection> provider, boolean disableValidation,
-                                         RetryPolicyTemplate retryPolicyTemplate) {
-    super(provider);
+  /**
+   * Creates a new instance
```
gsfernandes: There should be a line between this and params

PR #3389, opened 2016-09-30
MULE-10678: Email: Support deleteAfterRetrieve in the imap list operation

juandesi:


extensions/email/src/main/java/org/mule/extension/email/internal/commands/ListCommand.java
```diff
@@ -70,15 +75,29 @@
...
     }
   }
 
+  private <T extends BaseEmailAttributes> void deleteRetrieved(MailboxConnection connection, String folderName,
+                                                               List<OperationResult<Object, T>> retrievedEmails) {
+    retrievedEmails.forEach(e -> {
+      int number = e.getAttributes()
+          .orElseThrow(() -> new EmailException("Could not find attributes in retrieved email"))
+          .getNumber();
+      setFlagCommand.setByNumber(connection, folderName, DELETED, number);
```
marianogonzalez: wouldn't it be better to delegate into the delete command?

extensions/email/src/main/java/org/mule/extension/email/internal/commands/SetFlagCommand.java
```diff
@@ -52,4 +51,27 @@ public void set(MailboxConnection connection, String folderName, Flag flag, long
...
+   * Sets the specified {@code flag} to the email of mailbox number {@code number}
+   * <p>
+   * This method only works for {@link UIDFolder}s, that are handled by the IMAP protocol
+   *
+   * @param connection the associated {@link MailboxConnection}.
+   * @param folderName the name of the folder where the emails are stored.
+   * @param flag       the {@link Flag} that wants to be set in the email message.
+   * @param number     the number of the email in the corresponding {@link Folder} of name {@code folderName}
+   */
+  public void setByNumber(MailboxConnection connection, String folderName, Flag flag, int number) {
```
marianogonzalez: is this by number or by id? This isn't relative to the message position on the mailbox isn't it?

extensions/email/src/main/java/org/mule/extension/email/internal/mailbox/imap/IMAPOperations.java
```diff
@@ -93,7 +96,7 @@ public void markAsRead(@Connection MailboxConnection connection,
   public void markAsDeleted(@Connection MailboxConnection connection,
                             @Optional(defaultValue = INBOX_FOLDER) String mailboxFolder,
                             @Summary("Email ID Number of the email to mark as deleted") @DisplayName("Email ID") long emailId) {
-    setFlagByUID.set(connection, mailboxFolder, DELETED, emailId);
+    setFlagCommand.setByUID(connection, mailboxFolder, DELETED, emailId);
```
marianogonzalez: why is this a set by UID while the other is set by number?

extensions/email/src/test/resources/retriever/imap-flows.xml
```diff
@@ -73,9 +73,20 @@
     </flow>
 
     <flow name="retrieveAndDelete">
+        <email:list-imap config-ref="config" deleteAfterRetrieve="true"/>
```
marianogonzalez: is `deleteAfterRetrieve` the exact same parameter name that the `list-pop` operation has?

extensions/email/src/main/java/org/mule/extension/email/internal/commands/ListCommand.java
juandesi: no, as pop3 cannot delete a single email without losing consistency in the mailbox folder, we need to mark all the emails and then perform a deletion by closing the folder.

extensions/email/src/main/java/org/mule/extension/email/internal/commands/SetFlagCommand.java
juandesi: there are 2 methods, one relative to the position in the mailbox, and the other relative to the unique identifier.

extensions/email/src/main/java/org/mule/extension/email/internal/mailbox/imap/IMAPOperations.java
juandesi: because imap supports uids, pop3 doesn't

extensions/email/src/test/resources/retriever/imap-flows.xml
juandesi: yes

PR #3393, opened 2016-09-30
MULE-10663: Improve field ordering in new Http connector

afelisatti:


extensions/http/src/main/java/org/mule/extension/http/internal/request/validator/HttpRequesterProvider.java
```diff
@@ -73,7 +73,7 @@
    */
   @Parameter
   @Optional
-  @Placement(group = URL_CONFIGURATION, order = 2)
+  @Placement(group = CONNECTION, order = 2)
```
estebanwasinger: Maybe is better if the order of this fields follows the same order defined in `@Placement`

afelisatti: Makes sense.

PR #3397, opened 2016-10-03
MULE-10692: Improve exception messages when classes are not found in …

gsfernandes:
…RegionClassLoader


modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/plugin/DefaultArtifactPlugin.java
```diff
@@ -51,4 +53,9 @@ public String getArtifactName() {
   public ArtifactClassLoader getArtifactClassLoader() {
     return classLoader;
   }
+
+  @Override
+  public String toString() {
+    return format("%s[%s]@%s", getClass().getName(), getArtifactName(), Integer.toHexString(System.identityHashCode(this)));
```
pablokraan: Add static imports

PR #3398, opened 2016-10-03
MULE-10306: disable entity expansion by default in XML transformers

alepulver:


distributions/standalone/src/main/resources/MIGRATION.txt
```diff
@@ -7,6 +7,7 @@ Please check https://developer.mulesoft.com/docs/display/current/Mule+ESB for mo
 Migration changes from 3.8.0 to 3.8.x
 MULE-10100: Processing after a synchronous until successful resulting in a VoidMuleEvent will now continue with the original event.
 MULE-10686: XML entity expansion in Jersey is now disabled by default because it allows DoS attacks. To restore previous behaviour use the mule.xml.expandInternalEntities="true" property.
+MULE-10306: XML entity expansion is now disabled by default because it allows DoS attacks. To restore previous behaviour use the mule.xml.expandInternalEntities="true" property.
```
afelisatti: are the double quotes necessary? how would I enabled it from the console or add a wrapper.conf entry?

afelisatti: same for the line above regarding Jersey

modules/xml/src/test/java/org/mule/transformers/xml/xslt/XsltTransformerXXETestCase.java
```diff
@@ -28,6 +31,10 @@
 public class XsltTransformerXXETestCase extends FunctionalTestCase
 {
 
+    // enable entity expansion as it's required     to test external entity expansion
```
afelisatti: weird spacing

PR #3408, opened 2016-10-04
MULE-10644: Ensure MULE-9534 fix only applies to one-way endpoints and add tests for JMS Message ID request response correlation pattern

mbuchwald:


core/src/main/java/org/mule/runtime/core/api/Event.java
```diff
@@ -245,6 +245,13 @@
   SecurityContext getSecurityContext();
 
   /**
+   * @return the correlation id to use for this event.
```
dfeist: Missing jira #?

transports/jms/src/main/java/org/mule/compatibility/transport/jms/JmsMessageDispatcher.java
```diff
@@ -437,7 +456,13 @@ protected MessageConsumer createReplyToConsumer(Message jmsMessage, Event event,
     String durableName;
     // If we're not using
     if (!(replyTo instanceof TemporaryQueue || replyTo instanceof TemporaryTopic)) {
-      selector = "JMSCorrelationID='" + jmsMessage.getJMSCorrelationID() + "'";
+      // Since we are supporting the JMS Message ID Pattern for request-response, the correlationID will be null
+      // if it is not manually setted up, and the selector must be set to the messageID.
```
dfeist: "Set up"

PR #3410, opened 2016-10-04
MULE-10641: Allow declaration of MetadataResolver at parameter level

alegmarra:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/enricher/DynamicMetadataModelEnricher.java
```diff
@@ -96,20 +93,13 @@ private void enrichParameter(ParameterDeclaration declaration) {
...
-      if (metadataScope.isCustomScope()) {
-        declareOutput(sourceDeclaration.getOutput(), sourceDeclaration.getOutputAttributes(), metadataScope);
-        declareComponentMetadataKeyId(sourceDeclaration);
-      }
-    }
+  private void enrichSourceMetadata(SourceDeclaration declaration) {
+    declaration.getModelProperty(ImplementingTypeModelProperty.class)
+        .ifPresent(prop -> {
+          final Class<?> sourceType = prop.getType();
+          MetadataScopeAdapter metadataScope = new MetadataScopeAdapter(extensionType, sourceType);
```
juandesi: IMO Sources should not be able to use this annotation, instead, as they can only declare OutputMetadata, it should be annotated with OutputResolver just like all the operations, MetadataScope should be reserved for global scenarios only

extensions/db/src/main/java/org/mule/extension/db/api/param/ParameterizedStatementDefinition.java
```diff
@@ -38,8 +40,8 @@
    */
   @Parameter
   @Optional
-  //@Content TODO: MULE-10641 re-enable datasense for these parameter
   @Placement(order = 2)
+  @TypeResolver(BaseDbMetadataResolver.class)
```
marianogonzalez: The `Base` prefix denotes an abstract class meant to be extended. This is something you're actually using so rename

extensions/db/src/test/java/org/mule/extension/db/integration/delete/DeleteMetadataTestCase.java
```diff
@@ -58,15 +58,13 @@ public void bulkDeleteOutputMetadata() throws Exception {
   }
 
   @Test
-  @Ignore("TODO MULE-10641")
```
marianogonzalez: test case touched -> Allure

extensions/db/src/test/java/org/mule/extension/db/integration/delete/DeleteMetadataTestCase.java
```diff
@@ -58,15 +58,13 @@ public void bulkDeleteOutputMetadata() throws Exception {
...
   public void deleteNoParametersInputMetadata() throws Exception {
     ParameterMetadataDescriptor parameters =
         getInputMetadata("deleteMetadata", "DELETE FROM PLANET WHERE name = 'Mars'");
     assertThat(parameters.getType(), is(typeBuilder.nullType().build()));
   }
 
   @Test
-  @Ignore("TODO MULE-10641")
```
marianogonzalez: allure

extensions/db/src/test/java/org/mule/extension/db/integration/delete/DeleteMetadataTestCase.java
```diff
@@ -77,7 +75,6 @@ public void deleteParameterizedInputMetadata() throws Exception {
...
   public void deleteWithExpressionInputMetadata() throws Exception {
```
marianogonzalez: allure

extensions/db/src/test/java/org/mule/extension/db/integration/insert/InsertMetadataTestCase.java
```diff
@@ -52,23 +51,21 @@ public void insertOutputMetadata() throws Exception {
   }
 
   @Test
-  public void bulkUpdateOutputMetadata() throws Exception {
+  public void bulkInsertOutputMetadata() throws Exception {
```
marianogonzalez: congratulations ale.. you're allure master by now

extensions/db/src/test/java/org/mule/extension/db/integration/select/SelectMetadataInputTestCase.java
```diff
@@ -27,11 +27,9 @@
 import java.util.List;
 import java.util.Optional;
 
-import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runners.Parameterized;
 
-@Ignore("TODO MULE-10641")
 public class SelectMetadataInputTestCase extends AbstractDbIntegrationTestCase {
```
marianogonzalez: this test will look great once allured

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/metadata/MetadataExtensionFunctionalTestCase.java
```diff
@@ -181,21 +185,37 @@ protected void assertExpectedType(MetadataResult<ParameterMetadataDescriptor> de
...
+    assertThat(descriptor.get().isDynamic(), is(isDynamic));
   }
 
   protected Set<MetadataKey> getKeysFromContainer(MetadataKeysContainer metadataKeysContainer) {
     return metadataKeysContainer.getKeys(metadataKeysContainer.getCategories().iterator().next()).get();
   }
+
+  protected void assertSuccess(MetadataResult<?> metadata) {
+    if (!metadata.isSuccess()) {
+      if (metadata.getFailure().isPresent()) {
```
marianogonzalez: if it aint' a success, you should actually assert for the failure to be present rather than an innocent if

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/enricher/DynamicMetadataModelEnricher.java
marianogonzalez: there's some sense to this comment but it isn't completely accurate. Let's revisit this when we do the sources rework

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/BaseMetadataDelegate.java
```diff
@@ -0,0 +1,99 @@
...
+
+  /**
+   * Merge multiple failed {@link MetadataResult} into one {@link MetadataFailure}.
+   *
+   * @param results the results to be merged as one
+   * @return a new single {@link MetadataFailure}
+   */
+  protected <T> MetadataResult<T> mergeFailures(T descriptor, MetadataResult<?>... results) {
+    List<MetadataResult<?>> failedResults = Stream.of(results).filter(result -> !result.isSuccess()).collect(toList());
+    String messages = failedResults.stream().map(f -> f.getFailure().get().getMessage()).collect(joining(" and "));
```
marianogonzalez: didn't I comment on a method exactly like this just a while ago?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataInputDelegate.java
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Metadata service delegate implementations that handles the resolution
+ * of a {@link RuntimeComponentModel} {@link InputMetadataDescriptor}
+ *
+ * @since 4.0
+ */
+class MetadataInputDelegate extends BaseMetadataDelegate {
+
+
```
marianogonzalez: emptyness my old friend...

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataInputDelegate.java
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * For each of the Component's {@link ParameterModel} creates the corresponding {@link TypeMetadataDescriptor} using only its
+   * static {@link MetadataType} and ignoring if any parameter has a dynamic type.
+   *
+   * @return A {@link List} containing a {@link MetadataResult} of {@link TypeMetadataDescriptor} for each input parameter using
+   * only its static {@link MetadataType} and ignoring if any parameter has a dynamic type.
+   */
+  MetadataResult<InputMetadataDescriptor> getInputMetadataDescriptors(MetadataContext context, Object key) {
+    InputMetadataDescriptorBuilder input = MetadataDescriptorBuilder.inputDescriptor();
```
marianogonzalez: inputDescriptor() is not a good name for this... it should be newBuilder() or something like that

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataKeysDelegate.java
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Metadata service delegate implementations that handles the resolution
+ * of a {@link RuntimeComponentModel} {@link MetadataKeysContainer}
+ *
+ * @since 4.0
+ */
+class MetadataKeysDelegate extends BaseMetadataDelegate {
+
+
```
marianogonzalez: empty

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataKeysDelegate.java
```diff
@@ -0,0 +1,124 @@
...
+    try {
+      final Set<MetadataKey> metadataKeys = resolverFactory.getKeyResolver().getKeys(context);
+      final Map<Integer, String> partOrder = getPartOrderMapping(keyParts);
+      final Set<MetadataKey> enrichedMetadataKeys = metadataKeys.stream()
+          .map(metadataKey -> cloneAndEnrichMetadataKey(metadataKey, partOrder))
+          .map(MetadataKeyBuilder::build).collect(toSet());
+      keysContainer.add(componentResolverName, enrichedMetadataKeys);
+
+      return success(keysContainer.build());
+
```
marianogonzalez: empty

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataOutputDelegate.java
```diff
@@ -0,0 +1,127 @@
...
+                               "OutputAttributes");
+  }
+
+  private MetadataResult<TypeMetadataDescriptor> toTypeMetadataDescriptorResult(MetadataType type,
+                                                                                MetadataResult<MetadataType> result) {
+    MetadataType resultingType = result.get() == null ? type : result.get();
+
+    TypeMetadataDescriptor descriptor = typeDescriptor()
+        .withType(resultingType)
+        .dynamic(!type.equals(resultingType))
```
marianogonzalez: this logic is very fragile... all it takes is a badly written TypeAnnotation for this to give false positives. This information should actually be provided by the resolver.

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/introspection/validation/MetadataComponentModelValidatorTestCase.java
```diff
@@ -310,4 +319,23 @@ private ParameterModel getMockKeyPartParam(Object defaultValue, int order) {
     when(param.getModelProperty(MetadataKeyPartModelProperty.class)).thenReturn(of(new MetadataKeyPartModelProperty(order)));
     return param;
   }
+
+  public static class MockResolver implements TypeKeysResolver, OutputTypeResolver {
+
+    @Override
+    public MetadataType getOutputType(MetadataContext context, Object key)
+        throws MetadataResolvingException, ConnectionException {
+      return null;
```
marianogonzalez: maybe something for a follow up ticket, but I think that any resolver should be forbidden of returning null.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/BaseMetadataDelegate.java
alegmarra: yep, it turns out it was hidden in the metadataMediator, and now I need this in two places (because the mediator was splitted).  
I'll add a TODO so we review it in MULE-10707, once that ticket is completed, we probably won't need this anymore

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataInputDelegate.java
alegmarra: MetadataDescriptorBuilder provides the builders for all the possible metadata descriptors, which then have package-protected constructors. 
`newInputDescriptor`?

alegmarra: If I was not clear, you can check it in the current mule, it's been around for a while

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataOutputDelegate.java
alegmarra: I get what you meant about the `equals` returning false on the slightest change, but that is actually what I intended.
If the `type` is not exactly the same type returned by `component.getOutputAttributes().getType()` then it means that some kind of dynamic resolver was involved, and if that happens, then we are in a dynamic case.
(it will even be the same type instance, since once dynamic resolution is skyped, we use that instance as a default)

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/introspection/validation/MetadataComponentModelValidatorTestCase.java
alegmarra: Yeah, we could fail at runtime.
Currently we just default to the static type. I thought about throwing an exception, but left it that way just in case we needed a workaround of `dynamic resolver but if key is invalid the default to the static java type that is unknown to me because I'm resolving metadata for multiple operations`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataInputDelegate.java
marianogonzalez: ok, i get it now

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/introspection/validation/MetadataComponentModelValidatorTestCase.java
marianogonzalez: you can do it on a follow up, but have it blow up if null is returned

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataInputDelegate.java
alegmarra: feature for formatter?

marianogonzalez: could be

PR #3413, opened 2016-10-05
MULE-10709: Use custom factory builder for XML parsers to avoid vulne…

alepulver:
…rabilities

DON'T MERGE UNTIL https://muleesb.ci.cloudbees.com/job/Mule-3.x-CE-XMLSecurity/


modules/xml/src/test/java/org/mule/transformers/jaxb/JaxbSecurityTestCase.java
```diff
@@ -0,0 +1,123 @@
...
+        try
+        {
+            final MuleEvent testEvent = createEvent();
+            final MuleEvent response = ((Flow) getFlowConstruct("testFlow")).process(testEvent);
+
+            String payload = response.getMessage().getPayloadAsString();
+
+            fail("Should've thrown exception");
+        } catch (TransformerMessagingException e) {
+            assertThat(e.getTransformer(), instanceOf(JAXBUnmarshallerTransformer.class));
```
gsfernandes: Is it possible to check the cause exception?

modules/xml/src/test/java/org/mule/transformers/jaxb/JaxbSecurityTestCase.java
```diff
@@ -0,0 +1,123 @@
...
+    {
+        try
+        {
+            final MuleEvent testEvent = createEvent();
+            final MuleEvent response = ((Flow) getFlowConstruct("testFlow")).process(testEvent);
+
+            String payload = response.getMessage().getPayloadAsString();
+
+            fail("Should've thrown exception");
+        } catch (TransformerMessagingException e) {
```
gsfernandes: We use ExpectedException instead of try/catch, see examples from other junits and change this is possible.

alepulver: I'm using a try/catch to make assertions about the exception that was thrown.

modules/xml/src/test/java/org/mule/transformers/jaxb/JaxbSecurityTestCase.java
alepulver: I couldn't find it.

PR #3426, opened 2016-10-07
MULE-10731: ArtifactClassLoaderRunner - Does not work when using Allu…

gsfernandes:
…re - Added extraBootPackages and artifacts


gsfernandes:
Jenkins validate

PR #3430, opened 2016-10-11
MULE-10730: Event.Builder.variables does not clear previous values

elrodro83:
* Improve Message.toString() representation
* Move getCurrentEvent() and setCurrentEvent() to a less internal
location
* Reduce visibility of internal stuff

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/Event.java
+++ b/core/src/main/java/org/mule/runtime/core/api/Event.java
@@ -38,7 +38,10 @@ import java.util.NoSuchElementException;
  */
 public interface Event extends MuleEvent {
 
-  static final ThreadLocal<Event> currentEvent = new ThreadLocal<>();
+  static class CurrentEventHolder {
+
+    private static final ThreadLocal<Event> currentEvent = new ThreadLocal<>();
+  }
 
   /**
    * @return the context applicable to all events created from the same root {@link Event} from a {@link MessageSource}.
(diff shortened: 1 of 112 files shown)


core/src/main/java/org/mule/runtime/core/api/Event.java
```diff
@@ -318,6 +322,13 @@ static Builder builder(Event event) {
     Builder removeVariable(String key);
 
     /**
+     * Removes all variables.
+     *
+     * @return the builder instance
+     */
+    Builder clearVariables();
```
elrodro83: You're right, there's actually a bug in the variables method

core/src/main/java/org/mule/runtime/core/message/DefaultEventBuilder.java
```diff
@@ -280,35 +287,37 @@ protected boolean isFlowConstructNonBlockingProcessingStrategy() {
...
+    private final Error error;
 
     // Use this constructor from the builder
-    public EventImplementation(EventContext context, InternalMessage message,
-                               Map<String, DefaultTypedValue<Object>> variables,
-                               MessageExchangePattern exchangePattern, FlowConstruct flowConstruct, MuleSession session,
-                               boolean transacted, boolean synchronous, boolean nonBlocking, Object replyToDestination,
-                               ReplyToHandler replyToHandler, FlowCallStack flowCallStack, GroupCorrelation groupCorrelation,
-                               Error error, String legacyCorrelationId, boolean notificationsEnabled) {
+    private EventImplementation(EventContext context, InternalMessage message,
```
elrodro83: It is used in the kryo module

elrodro83:
jenkins validate

PR #3434, opened 2016-10-12
MULE-10750 - Basic support for locating a component in the configurat…

pablolagreca:
…ion by path and name

### DIFF
--- a/modules/spring-config/src/main/java/org/mule/runtime/config/spring/dsl/model/ComponentModel.java
+++ b/modules/spring-config/src/main/java/org/mule/runtime/config/spring/dsl/model/ComponentModel.java
@@ -178,10 +178,6 @@ public class ComponentModel {
     return this.schemaValueParameter.contains(parameterName);
   }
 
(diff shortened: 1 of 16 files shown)


modules/spring-config/src/main/java/org/mule/runtime/config/spring/dsl/model/ComponentModel.java
```diff
@@ -175,6 +178,14 @@ public boolean isParameterValueProvidedBySchema(String parameterName) {
     return this.schemaValueParameter.contains(parameterName);
   }
 
+  public void addAnnotation(String annotationKey, Object value) {
```
gsfernandes: Perhaps I'm missing something here but, what would be the meaning of an annotation on a ComponentModel?

pablolagreca: adding javadoc....

PR #3443, opened 2016-10-13
MULE-10424: Define bindings to use in Mule 4 from the event

afelisatti:


core/src/main/java/org/mule/runtime/core/el/v2/MuleExpressionLanguage.java
```diff
@@ -74,4 +87,17 @@ public ValidationResult validate(String expression) {
     return expressionExecutor.validate(expression);
   }
 
+  private void addEventBindings(MuleEvent event, BindingContext.Builder contextBuilder) {
+    Message message = event.getMessage();
+    Attributes attributes = message.getAttributes();
```
dfeist: Attributes should be TypedValue in Message. Ensure there is a TODO and JIRA issue.

afelisatti: Done: https://www.mulesoft.org/jira/browse/MULE-10774

core-tests/src/test/java/org/mule/runtime/core/el/MuleExpressionLanguageTestCase.java
```diff
@@ -27,4 +52,37 @@ public void nullExpressionExecutor() throws Exception {
     expressionLanguage.evaluate("hey", BindingContext.builder().build());
   }
 
+  @Test
+  public void eventBindings() throws Exception {
```
dfeist: Can this be split out to use a test method per-binding? This method is long..

dfeist: Also can we use/reuse any of the existing EL tests?

PR #3451, opened 2016-10-14
MULE-10681: Replace NullType with VoidType in places which makes sense

marianogonzalez:


extensions/db/src/main/java/org/mule/extension/db/internal/domain/metadata/DbInputMetadataResolver.java
```diff
@@ -59,7 +46,7 @@ public MetadataType getInputMetadata(MetadataContext context, String query)
     List<InputQueryParam> inputParams = queryTemplate.getInputParams();
     // No metadata when no input parameters
     if (inputParams.size() == 0) {
-      return typeBuilder.nullType().build();
+      return typeBuilder.voidType().build();
```
alegmarra: for me this is NullType, because we should communicate that they can ignore the parameter (ie, leave it null). For me a parameter with "no type" (void) is not a valid metadata description.

extensions/db/src/test/java/org/mule/extension/db/integration/delete/DeleteMetadataTestCase.java
```diff
@@ -46,7 +47,7 @@ public void bulkDeleteOutputMetadata() throws Exception {
   public void deleteNoParametersInputMetadata() throws Exception {
     ParameterMetadataDescriptor parameters =
         getInputMetadata("deleteMetadata", "DELETE FROM PLANET WHERE name = 'Mars'");
-    assertThat(parameters.getType(), is(typeBuilder.nullType().build()));
+    assertThat(parameters.getType(), is(instanceOf(VoidType.class)));
```
alegmarra: why asserting with a `instaceOf(class)` instead of using the MetadataType equals as we did before?

extensions/db/src/test/java/org/mule/extension/db/integration/insert/InsertMetadataTestCase.java
```diff
@@ -47,7 +48,7 @@ public void bulkInsertOutputMetadata() throws Exception {
   public void insertNoParametersInputMetadata() throws Exception {
     ParameterMetadataDescriptor parameters =
         getInputMetadata("insertMetadata", "INSERT INTO PLANET(POSITION, NAME) VALUES (777, 'Mercury')");
-    assertThat(parameters.getType(), is(typeBuilder.nullType().build()));
+    assertThat(parameters.getType(), is(instanceOf(VoidType.class)));
```
alegmarra: same question

extensions/db/src/test/java/org/mule/extension/db/integration/update/UpdateMetadataTestCase.java
```diff
@@ -49,7 +50,7 @@ public void bulkUpdateOutputMetadata() throws Exception {
   public void updateNoParametersInputMetadata() throws Exception {
     ParameterMetadataDescriptor parameters =
         getInputMetadata("updateMetadata", "update Planet set position = 1 where name = 'Mars'");
-    assertThat(parameters.getType(), is(typeBuilder.nullType().build()));
+    assertThat(parameters.getType(), is(instanceOf(VoidType.class)));
```
alegmarra: same

extensions/db/src/test/java/org/mule/extension/db/integration/delete/DeleteMetadataTestCase.java
marianogonzalez: because you don't need to instantiate anything just to check it's void

PR #3452, opened 2016-10-14
MULE-10516: Declare config-ref attribute as a reference in the Compon…

marianogonzalez:
…entBuildingDefinition


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -112,14 +112,19 @@ public void handle(Message message,
...
-    Event event = Event.builder(create(flowConstruct, getConfigurationProviderName()))
+    Event event = Event.builder(create(flowConstruct, getConfigName()))
         .message((InternalMessage) message).exchangePattern(REQUEST_RESPONSE).flow(flowConstruct)
         .build();
     messageProcessingManager
         .processMessage(new ExtensionFlowProcessingTemplate(event, messageProcessor, new NullCompletionHandler()),
                         createProcessingContext());
   }
 
+  private String getConfigName() {
```
pablolagreca: Use optional if it can be null.

PR #3456, opened 2016-10-17
MULE-10791: Re enable dynamic metadata resolution for operation output attributes

juandesi:


extensions/pom.xml
```diff
@@ -30,6 +30,7 @@
         <module>sockets</module>
         <module>validation</module>
         <module>email</module>
+        <module>wsc</module>
```
marianogonzalez: this should not be part of this PR

PR #3458, opened 2016-10-18
MULE-10613: The object type of List of Pojos defined inside a Pojo is…

pablomusumeci:
… not in the list of declared types

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/resources/manifest/ExportedArtifactsCollector.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/resources/manifest/ExportedArtifactsCollector.java
@@ -6,35 +6,37 @@
  */
 package org.mule.runtime.module.extension.internal.resources.manifest;
 
-import static java.util.stream.Collectors.toSet;
-import static org.mule.metadata.java.api.utils.JavaTypeUtils.getType;
-import static org.mule.runtime.module.extension.internal.ExtensionProperties.EXTENSION_MANIFEST_FILE_NAME;
-import static org.mule.runtime.module.extension.internal.util.MuleExtensionUtils.getClassLoader;
+import com.google.common.collect.ImmutableSet;
 import org.mule.metadata.api.annotation.EnumAnnotation;
 import org.mule.metadata.api.model.ArrayType;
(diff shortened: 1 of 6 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/resources/manifest/ExportedArtifactsCollector.java
```diff
@@ -151,13 +153,18 @@ private void collectExportedClass(MetadataType type) {
...
-        dictionaryType.getKeyType().accept(this);
-        dictionaryType.getValueType().accept(this);
+        addType(dictionaryType.getKeyType(), exportedClasses);
+        addType(dictionaryType.getValueType(), exportedClasses);
       }
 
       @Override
       public void visitArrayType(ArrayType arrayType) {
-        arrayType.getType().accept(this);
+        addType(arrayType.getType(), exportedClasses);
```
marianogonzalez: what is the root cause of the issue? why does this fix it? This solution is not valid since this wouldn't catch the case of List<Map<String, Apple>> . The reason why the original code used the visitor with the array's type was to allow for those combinations.

Why does this fix the issue? Isn't the real problem in `visitObjectField`?

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/resources/manifest/ExportedArtifactsCollectorTestCase.java
```diff
@@ -74,8 +74,8 @@ public void setup() {
   @Test
   public void collect() {
     Set<String> exportedPackages = collector.getExportedPackages();
-    assertThat(exportedPackages, hasSize(3));
-    assertThat(exportedPackages, hasItems(SHAPE_PACKAGE, APPLE_PACKAGE, VEGAN_PACKAGE));
+    assertThat(exportedPackages, hasSize(4));
+    assertThat(exportedPackages, hasItems(SHAPE_PACKAGE, APPLE_PACKAGE, VEGAN_PACKAGE, PEEL_PACKAGE));
```
marianogonzalez: add another test case considering composite cases, such as:

* List<List<Pojo>>
* List<Map<String, Pojo>>
*  Pojo which has a field of another pojo

PR #3461, opened 2016-10-18
MULE-10792: Model validator does not enforce returning an OperationRe…

pablomusumeci:
…sult when the OutputResolver specifies an attributes resolver


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/MetadataComponentModelValidator.java
```diff
@@ -118,6 +121,17 @@ public void visitObject(ObjectType objectType) {
 
   }
 
+  private void validateMetadataOutputAttributes(ExtensionModel extensionModel, ComponentModel component) {
+    MetadataResolverFactory resolverFactory = getMetadataResolverFactory(component);
+    if (isVoid(component.getOutputAttributes().getType())
+        && !(resolverFactory.getOutputAttributesResolver() instanceof NullMetadataResolver)) {
+      throw new IllegalModelDefinitionException(format("Extension '%s' has a/an %s named '%s' which specifies a non null %s but the output attributes were not defined",
```
marianogonzalez: the wording of the message is very confusing. Why not just say "the coso has an attributes metadata resolver but it doesn't set any attributes"?

PR #3462, opened 2016-10-18
MULE-10543: Implement the SchedulerService

elrodro83:
DO NOT MERGE - REVIEW ONLY!


services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultScheduler.java
```diff
@@ -0,0 +1,294 @@
...
+  }
+
+  @Override
+  public ScheduledFuture<?> schedule(Runnable command, long delay, TimeUnit unit) {
+    checkShutdown();
+    if (command == null) {
+      throw new NullPointerException();
+    }
+
+    final DefaultSchedulerTaskDecorator decorated = decorateRunnable(command);
```
dfeist: Isn't it a significantly simpler approach to keep references to `Future`s and then simply call `Future.cancel()` on each Future when the when the Scheduler is shutdown?  Or are there any other reasons to wrap runnable, other than performing per-scheduler stop?

services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultScheduler.java
```diff
@@ -0,0 +1,294 @@
...
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Proxy for a {@link ScheduledExecutorService} that adds tracking of the source of the dispatched tasks.
+ *
+ * @since 4.0
+ */
+public class DefaultScheduler implements Scheduler {
```
dfeist: Does it not make any sense to extend `AbstractExecutorService` and do any wrapping of task, or storing of futures in `newTaskFor` rather than implement all methods here?

services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultScheduler.java
```diff
@@ -0,0 +1,294 @@
...
+
+  private void tryTerminate() {
+    if (isTerminated()) {
+      terminationLatch.countDown();
+    }
+  }
+
+  @Override
+  public String toString() {
+    return super.toString() + "{" + System.lineSeparator()
```
dfeist: Static import

services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultScheduler.java
```diff
@@ -0,0 +1,294 @@
...
+    try {
+      synchronized (currentTasks) {
+        tasks = new ArrayList<>(currentTasks.size());
+        for (BaseSchedulerTaskDecorator task : currentTasks) {
+          task.stop();
+          tasks.add(task.getDecoratedRunnable());
+        }
+        currentTasks.clear();
+
+        return tasks;
```
dfeist: It seems this implementation will return both pending tasks and those that were executing when shutdown is called or those that are interupted, when it should only return pending tasks.  Task.stop() should probably remove tasks from currentTasks to achieve this but there is potentially a deadlock doing that, that needs to be avoided.

elrodro83: Will fox and add test for that scenario

core/src/main/java/org/mule/runtime/core/api/scheduler/Scheduler.java
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Provides access to a specific executor in the Mule runtime. Different {@link Scheduler} instances may be backed by the same
+ * {@link ExecutorService}, allowing for a fine control of the source of the tasks that the underlying {@link ExecutorService}
+ * will run.
+ * <p>
+ * See {@link ScheduledExecutorService} and {@link ExecutorService} for documentation on the provided methods.
+ * 
+ * @since 1.0
+ */
+public interface Scheduler extends ScheduledExecutorService {
```
elrodro83: There is an implementation issue with the scheduledExecutorService that it ignores the variable thread pool size and instead uses a fixed pool (ref: MULE-9146). So, in order to avoid it we use the scheduledExecutor just for handling the timer, and then dispatch the task to the actual pool.

services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultSchedulerService.java
```diff
@@ -0,0 +1,129 @@
...
+    logger.info("Stopping " + this.toString() + "...");
+
+    cpuLightExecutor.shutdown();
+    ioExecutor.shutdown();
+    computationExecutor.shutdown();
+    scheduledExecutor.shutdown();
+
+    try {
+      final long startMillis = currentTimeMillis();
+      waitForExecutorTermination(startMillis, scheduledExecutor, "scheduledExecutor");
```
dfeist: Add comment to say why it's important this executor is stopped first?

services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultSchedulerService.java
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  @Override
+  public Scheduler computationScheduler() {
+    return new DefaultScheduler(computationExecutor, scheduledExecutor);
+  }
+
+  @Override
+  public void start() throws MuleException {
+    int cores = getRuntime().availableProcessors();
```
dfeist: Move to a private static field?

services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultScheduler.java
```diff
@@ -0,0 +1,282 @@
...
+  public void execute(Runnable command) {
+    checkShutdown();
+    requireNonNull(command);
+
+    executor.execute(command);
+  }
+
+  protected void checkShutdown() {
+    if (isShutdown()) {
+      throw new RejectedExecutionException(this.toString() + " already shutdown");
```
dfeist: Use a RejectedExceptionHandler, unless we are sure about hard coding AbortPolicy behaviour in our Scheduler.

core/src/main/java/org/mule/runtime/core/api/scheduler/SchedulerService.java
```diff
@@ -0,0 +1,60 @@
...
+ * tasks on those.
+ * <p>
+ * This interface provides access to different schedulers, each with its own configured tuning options (TODO MULE-10585),
+ * optimized for some specific kind of work. Artifacts that need to dispatch tasks need to carefully determine which scheduler
+ * best fits the nature of the task in order to keep the resources usage and the impact on other artifacts at a minimum.
+ * <p>
+ * The {@link Scheduler}s returned by methods in the implementations must provide access the the thread pools in the Mule Runtime
+ * that have the expected tuning configuration. Each {@link Scheduler} will have its own lifecycle, managed by its user and NOT
+ * this service.
+ *
```
elrodro83: will add this, but in the implementation of the service, not the interface.

services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultScheduler.java
```diff
@@ -0,0 +1,282 @@
...
+
+  private final ExecutorService executor;
+  private final ScheduledExecutorService scheduledExecutor;
+
+  /**
+   * Wait condition to support awaitTermination
+   */
+  private final CountDownLatch terminationLatch = new CountDownLatch(1);
+
+  private Set<RunnableFutureDecorator<?>> currentTasks = synchronizedSet(new HashSet<>());
```
elrodro83: I was able to replace it with a ConcurrentHashMap, that should improve performance.

services/scheduler/src/main/java/org/mule/service/scheduler/internal/RunnableFutureDecorator.java
```diff
@@ -0,0 +1,104 @@
...
+  RunnableFutureDecorator(RunnableFuture<V> task, DefaultScheduler scheduler) {
+    this.task = task;
+    this.scheduler = scheduler;
+  }
+
+  @Override
+  public void run() {
+    try {
+      if (!startTask()) {
+        // In case the scheduler that owns this task was shutdown, we behave consistently by just not executing the task.
```
elrodro83: As it turns out, this same check is already being done in ThreadPoolExecutor#run(), so i will remove it from here.

services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultSchedulerService.java
```diff
@@ -0,0 +1,129 @@
...
+
+    logger.info("Starting " + this.toString() + "...");
+
+    // TODO MULE-10585
+    final String prefix = SchedulerService.class.getSimpleName();
+    cpuLightExecutor = new ThreadPoolExecutor(2 * cores, 2 * cores, 0, SECONDS, new ArrayBlockingQueue<>(2 * cores),
+                                              new NamedThreadFactory(prefix + "_cpuLight"));
+    ioExecutor = new ThreadPoolExecutor(cores, cores * cores, 0, SECONDS, new ArrayBlockingQueue<>(cores * cores),
```
elrodro83: just some random value to put for now until we do proper testing.

services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultSchedulerService.java
```diff
@@ -0,0 +1,129 @@
...
+
+  @Override
+  public void start() throws MuleException {
+    int cores = getRuntime().availableProcessors();
+
+    logger.info("Starting " + this.toString() + "...");
+
+    // TODO MULE-10585
+    final String prefix = SchedulerService.class.getSimpleName();
+    cpuLightExecutor = new ThreadPoolExecutor(2 * cores, 2 * cores, 0, SECONDS, new ArrayBlockingQueue<>(2 * cores),
```
elrodro83: just some random value to put for now until we do proper testing.

services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultSchedulerService.java
elrodro83: why? it is only used inside this method

core/src/main/java/org/mule/runtime/core/api/scheduler/SchedulerService.java
```diff
@@ -0,0 +1,60 @@
...
+ * @since 4.0
+ */
+public interface SchedulerService extends Service {
```
elrodro83: i have to do some implementation for the unit tests in core, that will not need lifecycle.

core/src/main/java/org/mule/runtime/core/api/scheduler/Scheduler.java
```diff
@@ -0,0 +1,33 @@
...
+
+  /**
+   * Tries to do a graceful shutdown.
+   * <p>
+   * If this hasn't terminated after a configured time, a forceful shutdown takes place.
+   * 
+   * @param gracefulShutdownTimeoutSecs the maximum time to wait for the running tasks to gracefully complete.
+   * @param unit the time unit of the {@code timeout} argument
+   */
+  void stop(long gracefulShutdownTimeoutSecs, TimeUnit unit);
```
elrodro83: done

PR #3463, opened 2016-10-19
MULE-10674: parameterValue for Database bulk-insert should be on top …

pablomusumeci:
…of the menue

### DIFF
--- a/extensions/db/src/main/java/org/mule/extension/db/internal/operation/BulkOperations.java
+++ b/extensions/db/src/main/java/org/mule/extension/db/internal/operation/BulkOperations.java
@@ -18,7 +18,6 @@ import org.mule.extension.db.api.param.QuerySettings;
 import org.mule.extension.db.internal.DbConnector;
 import org.mule.extension.db.internal.domain.connection.DbConnection;
 import org.mule.extension.db.internal.domain.executor.BulkUpdateExecutor;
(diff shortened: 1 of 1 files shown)


extensions/db/src/main/java/org/mule/extension/db/internal/operation/BulkOperations.java
```diff
@@ -62,8 +61,8 @@
    *         according to the order in which commands were added to the batch.
    * @throws SQLException if an error is produced
    */
-  public int[] bulkInsert(@ParameterGroup BulkQueryDefinition query,
-                          @XmlHints(allowReferences = false) List<Map<String, Object>> parameterValues,
+  public int[] bulkInsert(@XmlHints(allowReferences = false) List<Map<String, Object>> parameterValues,
```
marianogonzalez: this is fine. I'm not against this (although we're very soon to not supporting `XmlHints` on the parameters, but this is still correct).

However, how does this fix the issue? I was expecting the solution to include a `@Placement` saying that this should be on top. Have you tried the extension in mozart or studio to see how it looks like after this change?

pablomusumeci: Makes sense, rather have an explicit @Placement annotation to leave things clear.

PR #3471, opened 2016-10-21
MULE-10717: Flows can start processing messages before referenced flo…

pablokraan:
…ws are completely started

_ Delaying message sources  start until the rest of the mule context was started.
_ Message sources as before only when the flow construct is started on an already started mule context
 _ Fixing test that require a started mule context
 _ Removing flakiness from jbmp test case

### DIFF
--- a/core/src/main/java/org/mule/DefaultMuleContext.java
+++ b/core/src/main/java/org/mule/DefaultMuleContext.java
@@ -9,7 +9,6 @@ package org.mule;
 import static org.mule.api.config.MuleProperties.OBJECT_EXPRESSION_LANGUAGE;
 import static org.mule.api.config.MuleProperties.OBJECT_POLLING_CONTROLLER;
 import static org.mule.api.config.MuleProperties.OBJECT_TRANSACTION_MANAGER;
-
 import org.mule.api.Injector;
 import org.mule.api.MuleContext;
 import org.mule.api.MuleException;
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/mule/DefaultMuleContext.java
```diff
@@ -323,6 +329,73 @@ public synchronized void start() throws MuleException
...
+    {
+        startPipelineMessageSources();
+        startServicesMessageSources();
+    }
+
+    private void startServicesMessageSources() throws LifecycleException
+    {
+        for (Service service : this.getRegistry().lookupObjects(Service.class))
+        {
+            if (service.isStarted() || service.isPaused())
```
pablolagreca: What's the meaining of paused?. Wouldn't the source have to be paused and not started if the services is paused?

core/src/main/java/org/mule/DefaultMuleContext.java
```diff
@@ -323,6 +329,73 @@ public synchronized void start() throws MuleException
...
+            {
+                MessageSource messageSource = service.getMessageSource();
+
+                startMessageSource(messageSource);
+
+                if (service.isPaused() && messageSource instanceof Pausable)
```
pablolagreca: There can be an small portion of time where the source would be started, may try to process messages but the service would be stopped. Maybe there is not way to solve this. WHat do you think?

core/src/main/java/org/mule/DefaultMuleContext.java
```diff
@@ -323,6 +329,73 @@ public synchronized void start() throws MuleException
...
+                MessageSource messageSource = pipeline.getMessageSource();
+
+                startMessageSource(messageSource);
+            }
+        }
+    }
+
+    private void startMessageSource(MessageSource messageSource) throws LifecycleException
+    {
+        if (messageSource instanceof Startable)
```
pablolagreca: LifecycleUitls.startIfStatable?

core/src/main/java/org/mule/DefaultMuleContext.java
pablokraan: This is just to maintain the original behaviour. Some objects are created when the source is started only, there are tests that are broken otherwise.

PR #3472, opened 2016-10-24
MULE-10683: Make DataSense tooling service consistent with the test c…

estebanwasinger:
…onnectivity one


extensions/db/pom.xml
```diff
@@ -104,7 +104,7 @@
         <dependency>
             <groupId>org.apache.derby</groupId>
             <artifactId>derby</artifactId>
-            <version>${derbyVersion}</version>
+            <version>10.11.1.1</version>
```
estebanwasinger: Revert

modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/application/ApplicationStartedSplashScreen.java
```diff
@@ -40,6 +40,6 @@ private void listPlugins(ApplicationDescriptor descriptor) {
   }
 
   protected void listLibraries(ApplicationDescriptor descriptor) {
-    listItems(getLibraries(getAppLibFolder(descriptor.getName())), "Application libraries:");
+    listItems(getLibraries(getAppLibFolder(descriptor.getArtifactLocation())), "Application libraries:");
```
estebanwasinger: Revert

pom.xml
```diff
@@ -1541,7 +1547,7 @@
                         <execution>
                             <phase>compile</phase>
                             <goals>
-                                <goal>validate</goal>
+                                <goal>format</goal>
```
estebanwasinger: revert

pom.xml
```diff
@@ -1664,7 +1670,7 @@
                     <artifactId>maven-surefire-plugin</artifactId>
                     <version>${maven.surefire.plugin.version}</version>
                     <configuration>
-                        <argLine>-Xms64m -Xmx512m</argLine>
+                        <argLine>-Xms64m -Xmx512m -verbose:gc</argLine>
```
estebanwasinger: revert

core/src/main/java/org/mule/runtime/core/api/config/MuleProperties.java
```diff
@@ -160,7 +160,7 @@
   public static final String OBJECT_EXTENSION_MANAGER = "_muleExtensionManager";
   public static final String OBJECT_TIME_SUPPLIER = "_muleTimeSupplier";
   public static final String OBJECT_CONNECTION_MANAGER = "_muleConnectionManager";
-  public static final String OBJECT_METADATA_MANAGER = "_muleMetadataManager";
+  public static final String OBJECT_METADATA_SERVICE = "_muleMetadataManager";
```
marianogonzalez: shouldn't the registry key change as well for consistency?

modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/api/DeployableArtifact.java
```diff
@@ -65,4 +66,10 @@
    */
   ConnectivityTestingService getConnectivityTestingService();
 
+  /**
+   * @return the {@link MetadataService} which can resolve the metadata of the components inside
+   * the current {@link DeployableArtifact}
+   * @see MetadataService
+   */
+  MetadataService getMetadataService();
```
marianogonzalez: isn't this mixing concerns? It's weird to me that the MetadataService gets exposed at such a high level. @pablokraan can you confirm?

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/metadata/MetadataExtensionFunctionalTestCase.java
```diff
@@ -106,32 +114,59 @@
...
     event = eventBuilder().message(InternalMessage.of("")).build();
-    metadataManager = muleContext.getRegistry().lookupObject(MuleMetadataManager.class);
+    metadataService = muleContext.getRegistry().lookupObject(MuleMetadataService.class);
     personType = getMetadata(PERSON_METADATA_KEY.getId());
   }
 
-  protected ComponentMetadataDescriptor getComponentDynamicMetadata() {
-    return getComponentDynamicMetadata(PERSON_METADATA_KEY);
+  enum ResolutionType {
+    EXPLICIT_RESOLUTION, DSL_RESOLUTION
```
marianogonzalez: what exactly are these two modes?

modules/extensions-spring-support/src/test/resources/metadata-tests.xml
```diff
@@ -178,12 +182,12 @@
...
     <flow name="pagedOperationMetadata">
         <metadata:paged-operation-metadata config-ref="config"
-                                           animal="#[new org.mule.test.metadata.extension.model.animals.Bear()"/>
+                                           animal="#[new org.mule.test.metadata.extension.model.animals.Bear()]"/>
     </flow>
 
     <flow name="queryOperation">
         <metadata:do-query>
-            <metadata:query><![CDATA[dsql:SELECT id FROM Circle WHERE diameter < 18]]></metadata:query>
+            <metadata:query><![CDATA[dsql:SELECT id FROM Circle WHERE (diameter < 18)]]></metadata:query>
```
marianogonzalez: why are the parenthesis suddenly important?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataMediator.java
```diff
@@ -148,10 +122,22 @@ public MetadataMediator(ComponentModel componentModel) {
     return mergeResults(componentDescriptor, output, input);
   }
 
+  private MetadataResult<Object> getMetadataKeyObjectValue(ParameterValueResolver metadataKeyResolver) {
+    try {
+      Object keyValue = getContainerName().isPresent() ? metadataKeyResolver.getParameterValue(getContainerName().get()) : null;
+      return success(keyValue);
```
marianogonzalez: `success(null)` is a possible outcome and it doesn't sound right

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataMediator.java
```diff
@@ -148,10 +122,22 @@ public MetadataMediator(ComponentModel componentModel) {
...
+  }
+
   private List<ParameterModel> getMetadataKeyParts(ComponentModel componentModel) {
     return componentModel.getParameterModels().stream()
         .filter(p -> p.getModelProperty(MetadataKeyPartModelProperty.class).isPresent())
         .collect(toList());
   }
 
+  private Optional<String> getContainerName() {
+    return Optional.ofNullable(keyContainerName);
```
marianogonzalez: static import

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationParameterValueResolver.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * {@link ParameterValueResolver} implementation for {@link OperationMessageProcessor}
+ *
+ * @since 4.0
+ */
+public class OperationParameterValueResolver implements ParameterValueResolver {
```
marianogonzalez: this class doesn't make sense:

* First it's not for OperationMessageProcessor, since it's only value seems to be the `resolveSingleLevelMetadataKey` method which is Metadata specific
* Secondly, it takes an OperationContext, which means that this is more a wrapper than anything else
* It contains parameter group logic, which is also a smell

In summary, this looks like a big hack

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/LazyOperationContext.java
```diff
@@ -0,0 +1,112 @@
...
+
+    return (T) value;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Optional<ConfigurationInstance> getConfiguration() {
+    return Optional.empty();
```
marianogonzalez: static import

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceWrapper.java
```diff
@@ -165,6 +165,12 @@ private void releaseConnection() {
     return Optional.of(new FieldSetter<>(fields.iterator().next()));
   }
 
+  Object getFieldValue(String fieldName) throws NoSuchFieldException, IllegalAccessException {
+    final Field declaredField = delegate.getClass().getDeclaredField(fieldName);
```
marianogonzalez: looks like something that belongs in `IntrospectionUtils`

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessorTestCase.java
```diff
@@ -256,4 +287,80 @@ public void executesWithDefaultConfig() throws Exception {
     assertThat(operationContext, is(instanceOf(OperationContextAdapter.class)));
     assertThat(operationContext.getConfiguration().get().getValue(), is(sameInstance(defaultConfigInstance)));
   }
+
+
```
marianogonzalez: empty line

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSourceTestCase.java
```diff
@@ -131,6 +147,8 @@
   @Mock(answer = RETURNS_DEEP_STUBS)
   private Event event;
   private ExtensionMessageSource messageSource;
+  @Mock
```
marianogonzalez: add empty line

modules/spring-config/src/main/java/org/mule/runtime/config/spring/LazyMetadataService.java
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * {@link MetadataService} implementation that initialises the required components before doing test
+ * connectivity.
+ *
+ * This guarantees that if the application has been created lazily, the requested components exists
+ * before the execution of the actual {@link MetadataService}.
+ *
+ * @since 4.0
+ */
+public class LazyMetadataService implements MetadataService {
```
marianogonzalez: Looks like this is more like a decorator than an implementation

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -342,6 +343,17 @@ protected void validateOperationConfiguration(ConfigurationProvider configuratio
     }
   }
 
+  @Override
+  protected ParameterValueResolver getParameterValueResolver() {
+    return (fieldName) -> {
+      try {
+        return source.getFieldValue(fieldName);
```
marianogonzalez: getFieldValue() is not a responsibility that this class should have

modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/api/DeployableArtifact.java
pablokraan: This is similar to ArtifactContext#getConnectivityTestingService, so I guess it must be there too

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataMediator.java
alegmarra: why wrapping this in a metadata result? I'd just handle the exception outside and generate the failure there if needed.
regarding the `null`, I get that in the keyless case we have to pass on something to the resolver, but I'm not fond on injecting null to a piece of code that might be reused in cases where that key actually has a value. 
maybe we can think of a way of handling this cases from the resolver and all the way up to here (for example, if a resolver is used in a keyless way, we force them to parameterize the resolver with an Optional<> key, and we wrap whatever comes here in an optional.ofNullable to handle both cases)

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
```diff
@@ -323,4 +326,16 @@ private void validateConfigurationProviderIsNotExpression() throws Initialisatio
   protected ConfigurationProvider getConfigurationProvider() {
     return configurationProvider;
   }
+
+  private ParameterValueResolver getMetadataKeyValueResolver(MetadataKey key) {
```
alegmarra: this case should be handled by the mediator on its own. It has all the required information to resolve the value of the key. We can add a method that receives (context, key) in the mediator, and that method generates this resolver to delegate in the method expecting a PArameterValueResolver

core/src/main/java/org/mule/runtime/core/api/config/MuleProperties.java
estebanwasinger: yes, fixed

modules/extensions-spring-support/src/test/resources/metadata-tests.xml
estebanwasinger: reverted

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataMediator.java
estebanwasinger: Ale explained perfectly

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
estebanwasinger: The `ExtensionMessageSource` doesn't have the logic of how to resolve the field value, is delegating to the `SourceWrapper`. 
I don't see another way to expose the value of a field without exposing the entire concrete source `Class`.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationParameterValueResolver.java
estebanwasinger: 1. Yes, It's a bad and old method name, should be removed.
2. Yes, uses the OperationContext to reuse logic as much as possible.
3. In the SDK doesn't exist the logic of resolve a `ParameterGroup` from the name of the container of it, so this logic is necessary.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ParameterValueResolver.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.module.extension.internal.runtime;
+
+/**
+ * Contract for extension components that knows how to single parameter values
```
marianogonzalez: how to single -> how to resolve

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationParameterValueResolver.java
marianogonzalez: you didn't remove 1

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -347,19 +344,13 @@ protected void validateOperationConfiguration(ConfigurationProvider configuratio
...
-      if (keyIdModelProperty.isPresent()) {
-        try {
-          keyValue = source.getFieldValue(keyIdModelProperty.get().getParameterName());
-        } catch (NoSuchFieldException | IllegalAccessException e) {
-          throw new MetadataResolvingException(e.getMessage(), FailureCode.UNKNOWN, e);
-        }
+  protected ParameterValueResolver getParameterValueResolver() {
+    return (fieldName) -> {
+      try {
+        return source.getFieldValue(fieldName);
```
marianogonzalez: you didn't remove this... Sources should not have this method

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
marianogonzalez: i'd prefer SourceWrapper to have a getDelegate() which returns te actual source, and then use instrospectionUtils inside the parameter resolver

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/IntrospectionUtils.java
```diff
@@ -260,6 +260,17 @@ public static MetadataType getFieldMetadataType(Field field, ClassTypeLoader typ
     return CollectionUtils.isEmpty(candidates) ? Optional.empty() : Optional.of(candidates.iterator().next());
   }
 
+  public static Object getFieldValue(Object object, String fieldName) throws IllegalAccessException, NoSuchFieldException {
```
marianogonzalez: jdoc

PR #3473, opened 2016-10-24
MULE-10810: Deployment should look for application resources relative…

gsfernandes:
… to the artifactFolder

### DIFF
--- a/modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/application/ApplicationDescriptorFactory.java
+++ b/modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/application/ApplicationDescriptorFactory.java
@@ -11,12 +11,10 @@ import static java.lang.String.format;
 import static java.util.Collections.emptySet;
 import static org.apache.commons.io.FileUtils.listFiles;
 import static org.mule.runtime.container.api.MuleFoldersUtil.PLUGINS_FOLDER;
-import static org.mule.runtime.container.api.MuleFoldersUtil.getAppClassesFolder;
-import static org.mule.runtime.container.api.MuleFoldersUtil.getAppLibFolder;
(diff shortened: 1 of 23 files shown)


modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/application/ApplicationDescriptorFactory.java
```diff
@@ -90,9 +88,10 @@ public ApplicationDescriptor create(File artifactFolder) throws ArtifactDescript
 
       final Set<ArtifactPluginDescriptor> plugins = parsePluginDescriptors(artifactFolder, desc);
       verifyPluginExportedPackages(getAllApplicationPlugins(plugins));
+      desc.setArtifactLocation(artifactFolder);
```
pablokraan: There is a rootFolder on the artifact descriptor, that folder points to the folder that contains the exploded artifact file. Can that folder be used instead of the location?

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/application/TemporaryApplicationDescriptorFactoryTestCase.java
```diff
@@ -0,0 +1,145 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Set;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+public class TemporaryApplicationDescriptorFactoryTestCase extends AbstractMuleTestCase {
```
pablokraan: This test looks like a duplicate of DefaultApplicationFactoryTestCase

gsfernandes:
Ready for review. I found other places were it was considering apps to be always inside MULE_HOME/apps folder, I ended up moving this rely on to the descriptor that should be created by the factory with the resources needed for later. In particular, this case was about the classes folder the application.

PR #3475, opened 2016-10-24
MULE-10148: Refactor DeploymentServiceTestCase to avoid using pre-com…

pablokraan:
…piled classes and jars


tests/unit/src/main/java/org/mule/tck/util/CompilerUtils.java
```diff
@@ -0,0 +1,408 @@
...
+        options.add("-d");
+        options.add(target.getAbsolutePath());
+      }
+
+      if (jarFiles.length > 0) {
+        // Adds same classpath as the one used on the runner
+        String classPath = System.getProperty("java.class.path");
+        // Adds extra jars files required to compile the source classes
+        for (File jarFile : jarFiles) {
+          classPath = classPath + ":" + jarFile.getAbsolutePath();
```
gsfernandes: Use System.pathSeparator in order to make it work on Windows too, instead of :

PR #3484, opened 2016-10-26
moving classes from mule-api, some refactor after doc 'Plugin Extensi…

Ambrish12:
…onModel loading - Detailed process' with several TODOs in code, WIP


pablokraan:
@Ambrish12 can you explain us what is the purpose of creating a pull request like this one?

PR #3486, opened 2016-10-26
MULE-10790: Add Allure framework support in Mule-4.x.

anosenzo:


modules/tls/src/test/java/org/mule/runtime/module/tls/DefaultTlsContextFactoryTestCase.java
```diff
@@ -34,6 +37,10 @@
 
   @BeforeClass
   public static void createTlsPropertiesFile() throws Exception {
+
+    // This test will be ignored when the Allure framework is used because of this issue MULE-10805
+    assumeThat(System.getProperty("allure.profile.is.activated", "false"), Is.is(equalTo("false")));
```
aiannucci: If you use an assumption you will never be aware that this test is not running. The test will show as Passed while it didn't actually run.
We should use an @ Ignore and place a TODO tag.

extensions/validation/src/test/java/org/mule/extension/validation/NumberValidationTestCase.java
```diff
@@ -57,6 +60,10 @@ protected String getConfigFile() {
 
   @Test
   public void validateNumber() throws Exception {
+
+    // This test will be ignored when the Allure framework is used because of this issue MULE-10805
+    assumeThat(System.getProperty("allure.profile.is.activated", "false"), is(equalTo("false")));
```
aiannucci: If you use an assumption you will never be aware that this test is not running. The test will show as Passed while it didn't actually run.
We should use an @ Ignore and place a TODO tag.

pom.xml
```diff
@@ -2228,5 +2229,68 @@
...
+                                    <version>${allure.aspectj.version}</version>
+                                </dependency>
+                            </dependencies>
+                        </plugin>
+
+                        <!--Needed only to show reports locally. Run jetty:run and open localhost:8080 to show the report-->
+                        <plugin>
+                            <groupId>org.eclipse.jetty</groupId>
+                            <artifactId>jetty-maven-plugin</artifactId>
+                            <version>9.2.10.v20150310</version>
```
aiannucci: use a property for the version

pom.xml
```diff
@@ -2228,5 +2229,68 @@
...
+                </pluginManagement>
+            </build>
+
+            <reporting>
+                <excludeDefaults>true</excludeDefaults>
+                <plugins>
+                    <plugin>
+                        <groupId>ru.yandex.qatools.allure</groupId>
+                        <artifactId>allure-maven-plugin</artifactId>
+                        <version>2.5</version>
```
aiannucci: use a property for the version

pom.xml
```diff
@@ -2228,5 +2229,68 @@
                 </plugins>
             </build>
         </profile>
+        <profile>
+            <id>allure</id>
+
+            <properties>
+                <allure.aspectj.version>${aspectjVersion}</allure.aspectj.version>
```
aiannucci: Remove this line

pom.xml
```diff
@@ -2228,5 +2229,68 @@
...
+            <build>
+                <pluginManagement>
+                    <plugins>
+                        <plugin>
+                            <groupId>org.apache.maven.plugins</groupId>
+                            <artifactId>maven-surefire-plugin</artifactId>
+                            <configuration>
+                                <testFailureIgnore>true</testFailureIgnore>
+                                <argLine>
+                                    -javaagent:${settings.localRepository}/org/aspectj/aspectjweaver/${allure.aspectj.version}/aspectjweaver-${allure.aspectj.version}.jar
```
aiannucci: use ${aspectjVersion}

pom.xml
```diff
@@ -2228,5 +2229,68 @@
...
+                                </properties>
+                                <systemPropertyVariables>
+                                    <allure.profile.is.activated>true</allure.profile.is.activated>
+                                </systemPropertyVariables>
+                            </configuration>
+                            <dependencies>
+                                <dependency>
+                                    <groupId>org.aspectj</groupId>
+                                    <artifactId>aspectjweaver</artifactId>
+                                    <version>${allure.aspectj.version}</version>
```
aiannucci: use ${aspectjVersion}

PR #3487, opened 2016-10-26
MULE-10146: SchemaBuilder doesn't support child elements for List of …

pablomusumeci:
…Date types


marianogonzalez:
there's something weird with how this was formatted. static imports were moved to the bottom. Why that superfluous change? something is wrong on your IDE

PR #3488, opened 2016-10-26
MULE-10784: Add soap faults support in the new Web Service Consumer.

juandesi:


extensions/wsc/src/main/java/org/mule/extension/ws/internal/WebServiceConsumer.java
```diff
@@ -17,6 +19,7 @@
  *
  * @since 4.0
  */
+@Export(classes = WscException.class)
```
marianogonzalez: this shouldn't be necessary. If any of your operations have this on their throws signature it should be exported automatically

juandesi: ok, didn't know that was already fixed

PR #3489, opened 2016-10-26
MULE-9439: Improve serialization of isolated objects

pablokraan:
_ Adding SerializationProtocol to differentiate between internal/external serialization (both returned serializers protocols are the same until the next pull request)
_ Moving ObjectSerializer API to SerializationProtocol
_ Updating ObjectSerializer to provide internal/external serialization protocols
_ Using internal/external serializer depending on the context

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/serialization/ObjectSerializer.java
+++ b/core/src/main/java/org/mule/runtime/core/api/serialization/ObjectSerializer.java
@@ -6,109 +6,30 @@
  */
 package org.mule.runtime.core.api.serialization;
 
-import org.mule.runtime.core.util.store.DeserializationPostInitialisable;
-
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.io.Serializable;
-
 /**
- * Defines a component capable to serialize/deserialize objects into/from an array of {@link byte}s. Unlike usual serializing
- * components, this one doesn't enforce the serialized object to implement {@link Serializable}. However, some implementations
- * might require that condition and throw {@link IllegalArgumentException} if not met.
- * <p/>
- * Implementations are also responsible for the correct initialization of classes implementing the
- * {@link DeserializationPostInitialisable} interface.
- * <p/>
- * <p/>
- * Unexpected behavior can result of deserializing objects that were generated with a different implementation of
- * {@link ObjectSerializer}.
- * <p/>
- * Implementations are required to be thread-safe
- *
- * @since 3.7.0
+ * Provides access to the different {@link SerializationProtocol} used in the container.
  */
 public interface ObjectSerializer {
 
   /**
-   * Serializes the given object into a an array of {@link byte}s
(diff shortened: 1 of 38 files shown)


core/src/main/java/org/mule/runtime/core/api/serialization/ObjectSerializer.java
```diff
@@ -6,109 +6,30 @@
...
- * <p/>
- * Unexpected behavior can result of deserializing objects that were generated with a different implementation of
- * {@link ObjectSerializer}.
- * <p/>
- * Implementations are required to be thread-safe
- *
- * @since 3.7.0
+ * Provides access to the different {@link SerializationProtocol} used in the container.
  */
 public interface ObjectSerializer {
```
pablolagreca: Shouldn't this be moved to mule-api? Perhaps it can be done in the next PR.

core/src/main/java/org/mule/runtime/core/api/serialization/SerializationProtocol.java
```diff
@@ -0,0 +1,113 @@
...
+ * Implementations are also responsible for the correct initialization of classes implementing the
+ * {@link DeserializationPostInitialisable} interface.
+ * <p/>
+ * <p/>
+ * Unexpected behavior can result of deserializing objects that were generated with a different implementation of
+ * {@link SerializationProtocol}.
+ * <p/>
+ * Implementations are required to be thread-safe
+ */
+public interface SerializationProtocol {
```
pablolagreca: Shouldn't this be moved to mule-api? Perhaps it can be done in the next PR.

core/src/main/java/org/mule/runtime/core/api/serialization/SerializationProtocol.java
```diff
@@ -0,0 +1,113 @@
...
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.Serializable;
+
+/**
+ * Defines a component capable to serialize/deserialize objects into/from an array of {@link byte}s. Unlike usual serializing
+ * components, this one doesn't enforce the serialized object to implement {@link Serializable}. However, some implementations
+ * might require that condition and throw {@link IllegalArgumentException} if not met.
+ * <p/>
+ * Implementations are also responsible for the correct initialization of classes implementing the
```
pablolagreca: Is there a way to get rid of this restriction?. Maybe handle this internally without exposing this concept in this API?

pablokraan: We could try to get rid of DeserializationPostInitialisable, but as long as that class exists, serializers must use it

core/src/main/java/org/mule/runtime/core/api/serialization/SerializationProtocol.java
pablokraan: created https://www.mulesoft.org/jira/browse/MULE-10844 to track this

PR #3493, opened 2016-10-27
MULE-10038: Improve System tests framework.   Adding configurable location suffix to be able to start more than one server in the same test.

asequeira:


tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleDeployment.java
```diff
@@ -80,6 +80,7 @@
   private static final String DEBUG_PORT = "5005";
   private static Logger logger = LoggerFactory.getLogger(MuleDeployment.class);
   private static PollingProber prober;
+  protected String locationSuffix = "";
```
aiannucci: Can be private

PR #3505, opened 2016-10-31
MULE-10544: Change ProcessingStragy to use SchedulerService instead of

elrodro83:
flow ThreadingProfile
* Make ProcessingStrategy prototypes


core/src/main/java/org/mule/runtime/core/api/construct/Pipeline.java
```diff
@@ -28,7 +29,9 @@
 
   public List<Processor> getMessageProcessors();
 
-  public void setProcessingStrategy(ProcessingStrategy processingStrategy);
+  public void setProcessingStrategyFactory(ProcessingStrategyFactory processingStrategyFactory);
+
+  public ProcessingStrategyFactory getProcessingStrategyFactory();
```
dfeist: Is this public getter really needed?  Surely the processing strategy shouldn't be used outside of Pipeline, and even if it is then the ProcessingStrategy should be exposed and not the factory.

core/src/main/java/org/mule/runtime/core/api/processor/factory/AsynchronousProcessingStrategyFactory.java
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.api.processor.factory;
```
dfeist: Don't think this is the best package name as these are not processor factories.  I'd put these in the existing `.strategy.` package or maybe `.strategy.factory`, but I don't think that's necessary really.

core/src/main/java/org/mule/runtime/core/api/processor/factory/DefaultFlowProcessingStrategyFactory.java
```diff
@@ -0,0 +1,20 @@
...
+
+import org.mule.runtime.core.api.processor.ProcessingStrategy;
+import org.mule.runtime.core.construct.flow.DefaultFlowProcessingStrategy;
+
+
+public class DefaultFlowProcessingStrategyFactory implements ProcessingStrategyFactory {
+
+  @Override
+  public ProcessingStrategy create() {
+    return new DefaultFlowProcessingStrategy();
```
dfeist: Can we make the strategy implementations inner classes.  They should never be used or instantiated outside the context of their factory.

core/src/main/java/org/mule/runtime/core/api/processor/factory/ProcessingStrategyFactory.java
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.api.processor.factory;
```
dfeist: -> `org.mule.runtime.core.api.processor.strategy` (move ProcessingStrategy too in this case) or `org.mule.runtime.core.api.processor`.

core/src/main/java/org/mule/runtime/core/api/processor/factory/DefaultFlowProcessingStrategyFactory.java
elrodro83: Some test cases are referencing those classes, will try to refactor those cases

core/src/main/java/org/mule/runtime/core/api/strategy/factory/NonBlockingProcessingStrategyFactory.java
```diff
@@ -0,0 +1,187 @@
...
+      processingStrategy.setThreadWaitTimeout(threadWaitTimeout);
+    }
+    if (poolExhaustedAction != null) {
+      processingStrategy.setPoolExhaustedAction(poolExhaustedAction);
+    }
+
+    return processingStrategy;
+  }
+
+  // TODO MULE-10544 use internal executor, do not allow it to be passed form outside
```
dfeist: Need to take into account the ability to test still though.  Maybe using a TestScheduler if not via this means?

core/src/main/resources/META-INF/mule-module.properties
```diff
@@ -41,6 +42,7 @@ artifact.export.classPackages=org.mule.runtime.core,\
                               org.mule.runtime.core.api.service,\
                               org.mule.runtime.core.api.source,\
                               org.mule.runtime.core.api.store,\
+                              org.mule.runtime.core.api.strategy.factory,\
```
dfeist: Remove.

PR #3507, opened 2016-11-01
MULE-10870 - Duplicate content-type created in response builder

canoasan:
Added a check in the case a given header does not allow multiple values and more than one is set.
In that case, a warn message is shown and the older values are removed from the map instead of throwing an exception. I opted for this solution to keep backwards compatibility and throw the exception in a new commit in Mule 4.

### DIFF
--- a/modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseHeaderBuilder.java
+++ b/modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseHeaderBuilder.java
@@ -6,6 +6,7 @@
  */
 package org.mule.module.http.internal.listener;
 
+import static java.util.Arrays.asList;
 import static org.mule.module.http.api.HttpHeaders.Names.CONTENT_LENGTH;
 import static org.mule.module.http.api.HttpHeaders.Names.CONTENT_TYPE;
 import static org.mule.module.http.api.HttpHeaders.Names.TRANSFER_ENCODING;
@@ -19,7 +20,6 @@ import com.google.common.collect.Multimap;
 import com.google.common.collect.Multimaps;
 import com.google.common.collect.Sets;
 
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
@@ -29,10 +29,10 @@ import org.slf4j.LoggerFactory;
 
 public class HttpResponseHeaderBuilder
 {
-    private Logger logger = LoggerFactory.getLogger(getClass());
+    private static final Logger LOGGER = LoggerFactory.getLogger(HttpResponseHeaderBuilder.class);
 
-    private List<String> calculatedHeadersNames = Arrays.asList(TRANSFER_ENCODING, CONTENT_LENGTH);
-    private List<String> uniqueHeadersNames = Arrays.asList(TRANSFER_ENCODING, CONTENT_LENGTH, CONTENT_TYPE);
+    private List<String> calculatedHeadersNames = asList(TRANSFER_ENCODING, CONTENT_LENGTH);
+    private List<String> uniqueHeadersNames = asList(TRANSFER_ENCODING.toLowerCase(), CONTENT_LENGTH.toLowerCase(), CONTENT_TYPE.toLowerCase());
 
     Multimap<String, String> headers =
             Multimaps.newMultimap(new CaseInsensitiveMapWrapper<Collection<String>>(HashMap.class), new Supplier<Collection<String>>(){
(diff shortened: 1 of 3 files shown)


modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseHeaderBuilder.java
```diff
@@ -23,10 +24,15 @@
 import java.util.HashMap;
 import java.util.List;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 public class HttpResponseHeaderBuilder
 {
+    private Logger logger = LoggerFactory.getLogger(getClass());
```
marianogonzalez: logger should be static and final

modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerResponseBuilderTestCase.java
```diff
@@ -272,4 +293,20 @@ public boolean isDateValid(String dateToValidate){
         }
         return true;
     }
+
+    private static class HeaderPredicate implements Predicate {
```
marianogonzalez: wrong format

modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerResponseBuilderTestCase.java
```diff
@@ -75,6 +79,10 @@
     public SystemProperty responseBuilderAndErrorResponseBuilderNotTheSamePath = new SystemProperty("responseBuilderAndErrorResponseBuilderNotTheSamePath","responseBuilderAndErrorResponseBuilderNotTheSamePath");
     @Rule
     public SystemProperty httpHeadersResponseBuilderPath = new SystemProperty("httpHeadersResponseBuilderPath","httpHeadersResponseBuilderPath");
+    @Rule
```
marianogonzalez: add an empty line between rules

marianogonzalez:
commits in the CE project should always point to a MULE issue

modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseHeaderBuilder.java
```diff
@@ -73,11 +79,36 @@ private void failIfHeaderDoesNotSupportMultipleValues(String headerName)
...
+    /**
+     *
+     * Check if the given header allows multiple values. In case it doesn't and the header already has a value,
+     * a warn message is shown and the older values are removed from the map.
+     *
+     * @param headerName the header name to check if it has already a value set.
+     */
+    private void logIfHeaderDoesNotSupportMultipleValues(String headerName)
+    {
+        if (uniqueHeadersNames.contains(headerName))
```
afelisatti: Should we consider the case here? Just thinking about the several issues we've had around that.

modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerResponseBuilderTestCase.java
```diff
@@ -210,6 +218,19 @@ public void httpHeadersResponseBuilder() throws Exception
...
 
+    @Test
+    public void twoHeadersResponseBuilder() throws Exception
+    {
+        final String url = getUrl(twoHeadersResponseBuilderPath);
+        final Response response = Request.Get(url).connectTimeout(1000).execute();
+        final HttpResponse httpResponse = response.returnResponse();
+        final List<Header> headers = Arrays.asList(httpResponse.getAllHeaders());
+
+        assertTrue(exists(headers, new HeaderPredicate("Content-Type", "application/json")));
```
afelisatti: Use hamcrest.

modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerResponseBuilderTestCase.java
```diff
@@ -75,6 +79,10 @@
     public SystemProperty responseBuilderAndErrorResponseBuilderNotTheSamePath = new SystemProperty("responseBuilderAndErrorResponseBuilderNotTheSamePath","responseBuilderAndErrorResponseBuilderNotTheSamePath");
     @Rule
     public SystemProperty httpHeadersResponseBuilderPath = new SystemProperty("httpHeadersResponseBuilderPath","httpHeadersResponseBuilderPath");
+    @Rule
+    public SystemProperty httpContentTypeResponseBuilderPath = new SystemProperty("httpContentTypeResponseBuilderPath","twoContentTypeResponseBuilderFlow");
```
afelisatti: Is this being used? I don't see it in the xml nor in the tests.

modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseHeaderBuilder.java
```diff
@@ -23,10 +24,15 @@
...
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 public class HttpResponseHeaderBuilder
 {
+    private Logger logger = LoggerFactory.getLogger(getClass());
 
     private List<String> calculatedHeadersNames = Arrays.asList(TRANSFER_ENCODING, CONTENT_LENGTH);
+    private List<String> uniqueHeadersNames = Arrays.asList(TRANSFER_ENCODING, CONTENT_LENGTH, CONTENT_TYPE);
```
elrodro83: static imports

afelisatti:
Everything looks good to me. @marianogonzalez, @elrodro83?

afelisatti:
Whoever ends up merging this, remember to edit the commit message so that the SE thing is not visible.

PR #3509, opened 2016-11-01
MULE-10698: Parameter not supporting expressions accepts them

pablomusumeci:


marianogonzalez:
how does this fix the issue? Where are you validating that if ExpressionSupport.NONE then no literal value has an expression format?

pablomusumeci:
What was happening here was that we were retrieving the ExpressionSupport from the value  MetadataType of the field, instead of retrieving the value from the ObjectFieldType itself. The parameter mentioned in this issue is one that shouldn't support an expression.

marianogonzalez:
but then isn't the solution incomplete? What if the parameter is a simple type and yet I still go and give it an expression value?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/IntrospectionUtils.java
```diff
@@ -536,24 +541,87 @@ public static String getSourceName(Class<? extends Source> sourceType) {
...
+
+  /**
+   * Given a {@link MetadataType} it adds all the {@link Class} that are related from that type. This includes generics of an
+   * {@link ArrayType}, key and value of an {@link DictionaryType} and classes from the fields of {@link ObjectType}.
+   *
+   * @param type {@link MetadataType} to inspect
+   * @param relativeClasses {@link Set} were the classes should be added
+   * @param extensionClassLoader extension class loader
+   */
+  public static void collectRelativeClasses(MetadataType type, Set<Class<?>> relativeClasses, ClassLoader extensionClassLoader) {
```
marianogonzalez: should not receive the set... should return a new set instead. If you move this method into an utils, then it should have utils semantics, not speculate with how it should be used

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/ExtensionRegistry.java
```diff
@@ -80,17 +80,19 @@
...
-      if (enumClasses.add(enumClass)) {
-        try {
-          registry.registerTransformer(new StringToEnum(enumClass));
-        } catch (MuleException e) {
-          throw new MuleRuntimeException(createStaticMessage("Could not register transformer for enum " + enumClass.getName()),
-                                         e);
-        }
-      }
-    });
+    getParameterClasses(extensionModel, Thread.currentThread().getContextClassLoader()).stream()
```
alegmarra: MetadataTypes everywhere, check if is instantiable before doing a getType

PR #3511, opened 2016-11-01
MULE-10661: ArtifactClassLoaderRunner - URLClassLoaders are created f…

gsfernandes:
…or classification process, these CLs have to be closed

### DIFF
--- a/tests/runner/src/main/java/org/mule/test/runner/classloader/TestContainerClassLoaderFactory.java
+++ b/tests/runner/src/main/java/org/mule/test/runner/classloader/TestContainerClassLoaderFactory.java
@@ -20,6 +20,8 @@ import org.mule.runtime.module.artifact.descriptor.ArtifactDescriptor;
 
 import com.google.common.collect.ImmutableSet;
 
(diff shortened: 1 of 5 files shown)


tests/runner/src/main/java/org/mule/test/runner/classloader/TestContainerClassLoaderFactory.java
```diff
@@ -113,7 +121,11 @@ protected ArtifactClassLoader createArtifactClassLoader(final ClassLoader parent
    * @return uses only the set of {@link URL}s defined for the container to create the {@link ClassLoaderLookupPolicy}
    */
   public ClassLoaderLookupPolicy getContainerClassLoaderLookupPolicy() {
-    return withContextClassLoader(classLoader, () -> super.getContainerClassLoaderLookupPolicy(discoverModules()));
+    try (URLClassLoader classLoader = new URLClassLoader(urls, null)) {
```
elrodro83: Doesn't this change have an impact on performance? before this was being created once for this object and now it is on each method call.

PR #3515, opened 2016-11-02
MULE-10882: Unregister resources from Bitronix (ResourceRegistrar and ManagementRegistrar)

mbuchwald:


modules/db/src/test/java/org/mule/module/db/internal/resolver/database/StaticDbConfigResolverTestCase.java
```diff
@@ -72,4 +78,15 @@ public void returnsMetaData() throws Exception
...
         assertThat(metaData, is(expectedMetaData));
     }
+
+    @Test
+    public void disposesDisposableDbConfig() throws Exception
+    {
+        DataSource dataSource = mock(DataSource.class, withSettings().extraInterfaces(Disposable.class));
+        when(dbConfig.getDataSource()).thenReturn(dataSource);
+
+        ((StaticDbConfigResolver) dbConfigResolver).dispose();
```
pablokraan: dbConfigResolver should be defined as StaticDbConfigResolver instead of using a cast

modules/db/src/test/java/org/mule/module/db/internal/resolver/database/StaticDbConfigResolverTestCase.java
```diff
@@ -72,4 +78,15 @@ public void returnsMetaData() throws Exception
 
         assertThat(metaData, is(expectedMetaData));
     }
+
+    @Test
+    public void disposesDisposableDbConfig() throws Exception
+    {
+        DataSource dataSource = mock(DataSource.class, withSettings().extraInterfaces(Disposable.class));
```
pablokraan: You can create a DisposableDataSource interface to mock and to avoid the cast at the end

PR #3517, opened 2016-11-02
MULE-10544: Change ProcessingStragy to use SchedulerService instead of flow ThreadingProfile

elrodro83:


core/src/main/java/org/mule/runtime/core/processor/strategy/NonBlockingProcessingStrategyFactory.java
```diff
@@ -168,20 +60,22 @@ public void configureProcessors(List<Processor> processors,
...
       } else {
-        return super.onProcessor(processor, publisherFunction);
+        return publisher -> from(publisher).transform(publisherFunction);
       }
     }
 
     @Override
     public void start() throws MuleException {
-      executorService = newFixedThreadPool(maxThreads);
+      this.scheduler = schedulerService.cpuLightScheduler();
```
dfeist: This should IO pool, as blocking operations occur using these thread. (Today pool of 128 is used by default in 3.x)

PR #3519, opened 2016-11-02
MULE-10349: Support expression for response parameters in sources - fix

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/LazyExecutionContext.java
```diff
@@ -29,18 +29,18 @@
  *
  * @since 4.0
  */
-public class LazyOperationContext implements ExecutionContext<OperationModel> {
+public class LazyExecutionContext<M extends ComponentModel> implements ExecutionContext<M> {
```
afelisatti: javadoc for M

PR #3520, opened 2016-11-03
MULE-10349: Support expression for response parameters in sources - y…

marianogonzalez:
…et another fix


extensions/http/src/main/java/org/mule/extension/http/internal/listener/HttpResponseFactory.java
```diff
@@ -159,18 +164,18 @@ public HttpResponse create(HttpResponseBuilder responseBuilder,
...
-      } else {
-        try {
-          bytes = (byte[]) objectToByteArray.transform(payload);
-        } catch (Exception e) {
-          throw new RuntimeException(e);
-        }
-      }
-
-      ByteArrayHttpEntity byteArrayHttpEntity = new ByteArrayHttpEntity(bytes);
+      //byte[] bytes;
```
elrodro83: remove

PR #3521, opened 2016-11-03
MULE-10847: Implement internal object serializers

pablokraan:
_ Injecting the ObjectSerializer on the application on creation (through builder and factory)
_ Renaming ArtifactClassLoaderRepository to ClassLoaderRepository: can’t depend on ArtifactClassLoader because it does not work when the isolated test runner is used (because classloader classes are loaded on a different classloader)
_ Registering ClassLoaderRepository on the muleContext, as serializers can be defined from the mule config and they will need a reference to it
_ Adding custom Java object serializer
_ Adding custom object input and output streams

### DIFF
--- a/modules/artifact/src/main/java/org/mule/runtime/module/artifact/serializer/CustomJavaSerializationProtocol.java
+++ b/modules/artifact/src/main/java/org/mule/runtime/module/artifact/serializer/CustomJavaSerializationProtocol.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+
+package org.mule.runtime.module.artifact.serializer;
+
+import static org.mule.runtime.api.util.Preconditions.checkArgument;
+import org.mule.runtime.core.api.serialization.SerializationException;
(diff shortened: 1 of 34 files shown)


modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/DefaultClassLoaderManager.java
```diff
@@ -39,10 +39,21 @@ public ArtifactClassLoader unregister(String classLoaderId) {
...
+    ArtifactClassLoader artifactClassLoader = artifactClassLoaders.get(classLoaderId);
+    if (artifactClassLoader != null) {
+      result = artifactClassLoader.getClassLoader();
+    }
+
+    return result;
+  }
+
+  @Override
+  public String getId(ClassLoader classLoader) {
```
pablolagreca: Should this be Optional?

modules/artifact/src/main/java/org/mule/runtime/module/artifact/serializer/CustomJavaSerializationProtocol.java
```diff
@@ -0,0 +1,81 @@
...
+
+      return (T) obj;
+    } catch (Exception ex) {
+      throw new SerializationException("Cannot deserialize object", ex);
+    }
+  }
+
+  private void validateForSerialization(Object object) {
+    if (object != null && !(object instanceof Serializable)) {
+      throw new SerializationException(String.format("Was expecting a Serializable type. %s was found instead",
```
pablolagreca: static imports

PR #3522, opened 2016-11-03
MULE-10646: Refactor GrizzlyHttpClient to use BodyDeferringAsyncHandler

afelisatti:


afelisatti:
This branch is currently running on CI to validate everything works as expected. Job id is "http-streaming".

modules/http/pom.xml
```diff
@@ -148,6 +148,12 @@
             <version>4.3.3</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpasyncclient</artifactId>
+            <version>4.0.2</version>
```
dfeist: Do we not use this anywhere else, if we do then ideally version should be in base pom.xml

modules/http/src/test/java/org/mule/module/http/functional/HttpStreamingNonBlockingTestCase.java
```diff
@@ -0,0 +1,67 @@
...
+            stop.set(true);
+            //consume stream
+            event.getMessageAsString();
+            latch.release();
+        }
+
+        @Override
+        public void processExceptionReplyTo(MessagingException exception, Object replyTo)
+        {
+            fail();
```
dfeist: Fail with a message.

modules/http/src/test/java/org/mule/module/http/functional/requester/HttpRequestConnectionsPersistenceTestCase.java
```diff
@@ -25,9 +26,26 @@
 
 public class HttpRequestConnectionsPersistenceTestCase extends AbstractHttpRequestTestCase
 {
+
     private static final int GRIZZLY_IDLE_CHECK_TIMEOUT_MILLIS = 6000;
     private static final int POLL_DELAY_MILLIS = 200;
+    public static final int SMALL_TIMEOUT_MILLIS = 500;
```
dfeist: Can these be pushed up?

modules/http/src/main/java/org/mule/module/http/internal/request/grizzly/GrizzlyHttpClient.java
```diff
@@ -515,4 +480,109 @@ public void stop()
     {
         asyncHttpClient.close();
     }
+
+    private class BodyDeferringWorkManagerSource implements AsyncHandler<Response>, WorkManagerSource
```
dfeist: Unsure why "WorkManagerSource" is in the class name, this is really secondary to the purpose of this class.  Instead isn't this a BodyDefferingAsyncHAndler?  (or was name bad before too)?

modules/http/pom.xml
afelisatti: We are not using it anywhere else.

modules/http/src/main/java/org/mule/module/http/internal/request/grizzly/GrizzlyHttpClient.java
afelisatti: Just because I was replacing "WorkManagerSourceAsyncCompletionHandler" and it was both a handler and a WorkManagerSource but I guess a better name would be "WorkManagerSourceBodyDeferringHandler" or something like that.

modules/http/src/test/java/org/mule/module/http/functional/requester/HttpRequestConnectionsPersistenceTestCase.java
afelisatti: Because of them being public? Sure.

PR #3533, opened 2016-11-07
MULE-10896: Extract classLoader model definition from artifact descri…

gsfernandes:
…ptors - Fixed TemporaryApplicationFactory


gsfernandes:
Jenkins validate

PR #3537, opened 2016-11-08
MULE-10546: Update notifications threading with usage of the ThreadingService

elrodro83:
* Fix issue in SchedulerService under load
* Refactor tests startup/disposal of the context and services
* Use the SchedulerServices for async notifications


elrodro83:
DO NOT MERGE YET!!! BUILD PENDING!!!

core/src/main/java/org/mule/runtime/core/context/notification/ServerNotificationManager.java
```diff
@@ -147,24 +158,36 @@ public boolean isListenerRegistered(ServerNotificationListener listener) {
...
-        notifyListeners(notification);
+        notifyListeners(notification, (listener, nfn) -> listener.onNotification(nfn));
       } else {
-        try {
-          eventQueue.put(notification);
-        } catch (InterruptedException e) {
-          if (!disposed.get()) {
-            logger.error("Failed to queue notification: " + notification, e);
+        notifyListeners(notification, (listener, nfn) -> {
+          if (CPU_LITE.equals(listener.getProcessingType())) {
```
dfeist: All notifications should be non-blocking.  So there should be a JIRA for this and a TODO here to only use cpuLite pool.  Also, even in the current state where some listeners may be blocking I don't see any need for computation pool here at all.

core/src/main/java/org/mule/runtime/core/api/context/notification/ServerNotificationListener.java
```diff
@@ -6,11 +6,17 @@
...
+import org.mule.runtime.core.api.processor.ReactiveProcessor;
+
 /**
- * <code>ServerNotificationListener</code> is an observer interface that ojects can implement and register themselves with the
+ * <code>ServerNotificationListener</code> is an observer interface that objects can implement and register themselves with the
  * Mule Server to receive notifications when the server, model and components stop, start, initialise, etc.
  */
 public interface ServerNotificationListener<T extends ServerNotification> {
 
+  default ReactiveProcessor.ProcessingType getProcessingType() {
```
dfeist: Use IO as default until all notifications are non-blocking.  Also update javadoc of notificationListeners to say they should be non-blocking.

core/src/main/java/org/mule/runtime/core/api/context/notification/SynchronousServerEvent.java
```diff
@@ -11,6 +11,6 @@
  * thread, thus blocking the current thread of execution until all listeners have been processed
  */
 
-public interface BlockingServerEvent {
+public interface SynchronousServerEvent {
```
dfeist: Update javadoc.

PR #3538, opened 2016-11-08
MULE-10905: Add support for implicit Keys Resolvers for Enum and Bool…

estebanwasinger:
…ean MetadataKeyIds

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/metadata/DefaultMetadataResolverFactory.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/metadata/DefaultMetadataResolverFactory.java
@@ -52,6 +52,11 @@ public final class DefaultMetadataResolverFactory implements MetadataResolverFac
     keysResolver = keyResolver.get();
     outputTypeResolver = outputResolver.get();
     attributesTypeResolver = attributesResolver.get();
+
+    checkArgument(keysResolver != null, "MetadataKeyResolver type cannot be null");
+    inputResolvers.values().forEach(resolver -> checkArgument(resolver != null, "Input Type Resolver cannot be null"));
+    checkArgument(outputTypeResolver != null, "OutputTypeResolver type cannot be null");
+    checkArgument(attributesTypeResolver != null, "AttributesTypeResolver type cannot be null");
   }
 
   /**
--- a/core/src/main/java/org/mule/runtime/core/internal/metadata/DefaultMetadataResolverFactory.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/metadata/DefaultMetadataResolverFactory.java
@@ -21,6 +21,7 @@ import org.mule.runtime.extension.api.metadata.NullMetadataResolver;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.function.Supplier;
 
 
 /**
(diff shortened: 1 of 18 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/ResolverSupplier.java
```diff
@@ -0,0 +1,66 @@
...
+
+import java.util.function.Supplier;
+
+/**
+ * {@link Supplier} implementation that returns instances of type {@link T} from a given {@link Class}
+ *
+ * @since 4.0
+ * @param <T> The type of the instances to supply
+ */
+public final class ResolverSupplier<T> implements Supplier<T> {
```
marianogonzalez: if this is a resolver supplier, then T should extend Resolver. Otherwise, this is just a supplier.

core/src/main/java/org/mule/runtime/core/internal/metadata/NullMetadataResolverSupplier.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.util.function.Supplier;
+
+/**
+ * {@link Supplier} implementation which returns a {@link NullMetadataResolver} instance.
+ * This supplier always returns the same instance.
+ *
+ * @since 4.0
+ */
+public class NullMetadataResolverSupplier implements Supplier<NullMetadataResolver> {
```
marianogonzalez: is this class really needed? Can't you just declare an inline supplier like `() -> myNullResolver`... I would get the point for this class if you tell me that you reuse it a lot of times. But then again, I would argue that a `StaticSupplier<T>` in `org.mule.runtime.core.util.func` would be more useful

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/BooleanKeyResolver.java
```diff
@@ -0,0 +1,63 @@
...
+
+    {
+      add(newKey("TRUE").build());
+      add(newKey("FALSE").build());
+    }
+  };
+  private final String categoryName;
+
+  BooleanKeyResolver(String categoryName) {
+    checkArgument(isNotEmpty(categoryName), "Category Name can't be null");
```
pablomusumeci: Is this validation necessary? Take a look at https://github.com/mulesoft/mule/blob/mule-4.x/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/MetadataComponentModelValidator.java#L172

core/src/main/java/org/mule/runtime/core/internal/metadata/NullMetadataResolverSupplier.java
alegmarra: +1

core/src/main/java/org/mule/runtime/core/internal/metadata/DefaultMetadataResolverFactory.java
```diff
@@ -37,20 +38,20 @@
...
-  public DefaultMetadataResolverFactory(Class<? extends TypeKeysResolver> keyResolver,
-                                        Map<String, Class<? extends InputTypeResolver>> typeResolvers,
-                                        Class<? extends OutputTypeResolver> outputResolver,
-                                        Class<? extends AttributesTypeResolver> attributesResolver) {
+  public DefaultMetadataResolverFactory(Supplier<? extends TypeKeysResolver> keyResolver,
+                                        Map<String, Supplier<? extends InputTypeResolver>> typeResolvers,
+                                        Supplier<? extends OutputTypeResolver> outputResolver,
+                                        Supplier<? extends AttributesTypeResolver> attributesResolver) {
 
     checkArgument(keyResolver != null, "MetadataKeyResolver type cannot be null");
```
alegmarra: this doesn't make much sense now... or at least should be validated also for the result of the supplier

core/src/main/java/org/mule/runtime/core/internal/metadata/NullMetadataResolverSupplier.java
estebanwasinger: I understand, the problem that came up from using `Suppliers` is that you can't know if the resolver is a proper resolver or a `NullTypeResolver`.
**Why I need that?**
The logic of knowing if the type is dynamic or not is checking the type of the resolver, due that now is wrapped in a supplier that is not possible.
Which are the possible fixes?
* Instantiate the resolver and maintain the old logic (I don't like to instantiate a lot of classes just to ask that)
* Ask for the supplier type, having the `NullMetadataResolverSupplier` that will be returned every time that a `NullTypeResolver` fix the issue.

marianogonzalez: Ok, I understand that and I'm willing to let it go. But we should be looking forward of not depending on the resolver type for this.

PR #3544, opened 2016-11-08
MULE-10654: Model Validator should not complain if the singularized n…

pablomusumeci:
…ame clashes with another parameter name, when the generic of the list is of the same as the parameter type

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/NameClashModelValidator.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/NameClashModelValidator.java
@@ -12,11 +12,10 @@ import static java.util.function.Function.identity;
 import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toMap;
 import static java.util.stream.Collectors.toSet;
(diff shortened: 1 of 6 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/NameClashModelValidator.java
```diff
@@ -75,13 +74,21 @@ public void validate(ExtensionModel model) throws IllegalModelDefinitionExceptio
...
     private final ExtensionModel extensionModel;
     private final Set<DescribedReference<NamedObject>> namedObjects = new HashSet<>();
     private final Map<String, DescribedParameter> singularizedObjects = new HashMap<>();
     private final Multimap<String, TopLevelParameter> topLevelParameters = LinkedListMultimap.create();
+    private final DslSyntaxResolver dslSyntaxResolver;
 
     public ValidationDelegate(ExtensionModel extensionModel) {
       this.extensionModel = extensionModel;
+      this.dslSyntaxResolver = new DslSyntaxResolver(extensionModel,
+                                                     new SingleExtensionImportTypesStrategy());
```
pablomusumeci: The second parameter is how you configure the DslSyntaxResolver to work without any imported type without failing. Check if that's the way to go.

PR #3546, opened 2016-11-09
MULE-10913: DB Pool retrieving invalid connections

pablokraan:
_ Validating connections on checkout. This will use JDBC 4 connection checking, which should be usually available as most of the connector functionality depends on it. If this is not the case, a default query will be used to validate the connection.


marianogonzalez:
the commit comment says that is the driver doesn't support this a default query will be used, but I don't see that fallback code. Did you forget to push something?

pablokraan:
@marianogonzalez that logic lives inside c3p0

PR #3560, opened 2016-11-10
MULE-10753: SDK: Inconsistent behavior when injecting a tlsContext in…

pablomusumeci:
… a pojo

### DIFF
--- a/modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/parameter/ObjectTypeParameterParser.java
+++ b/modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/parameter/ObjectTypeParameterParser.java
@@ -14,7 +14,8 @@ import static org.mule.runtime.dsl.api.component.ComponentBuildingDefinition.Bui
 import static org.mule.runtime.dsl.api.component.TypeDefinition.fromType;
 import static org.mule.runtime.extension.api.declaration.type.TypeUtils.acceptsReferences;
 import static org.mule.runtime.extension.api.declaration.type.TypeUtils.getExpressionSupport;
-import org.mule.metadata.api.annotation.TypeIdAnnotation;
+import static org.mule.runtime.module.extension.internal.introspection.describer.model.InfrastructureTypeMapping.getNameMap;
+import static org.mule.runtime.module.extension.internal.util.ExtensionMetadataTypeUtils.getId;
 import org.mule.metadata.api.model.ArrayType;
 import org.mule.metadata.api.model.DictionaryType;
 import org.mule.metadata.api.model.MetadataType;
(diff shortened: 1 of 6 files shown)


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/parameter/ObjectTypeParameterParser.java
```diff
@@ -85,13 +88,22 @@ protected void doParse(Builder definitionBuilder) throws ConfigurationException
     type.getFields().forEach(this::parseField);
   }
 
+  private String getFieldName(ObjectFieldType objectFieldType, String fieldName) {
+
+    return objectFieldType.getValue().getAnnotation(TypeIdAnnotation.class)
```
alegmarra: `InfrastructureTypeMapping.getNameMap().getOrDefault(ExtensionMetadataTypeUtils.getId(field.getValue()), fieldName)`
+ static imports

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/parameter/ObjectTypeParameterParser.java
```diff
@@ -85,13 +88,22 @@ protected void doParse(Builder definitionBuilder) throws ConfigurationException
...
   private void parseField(ObjectFieldType objectField) {
     final MetadataType fieldType = objectField.getValue();
-    final String fieldName = objectField.getKey().getName().getLocalPart();
+    final String key = objectField.getKey().getName().getLocalPart();
+    final String fieldName = getFieldName(objectField, key);
     final boolean acceptsReferences = acceptsReferences(objectField);
     final Object defaultValue = getDefaultValue(fieldType).orElse(null);
     final ExpressionSupport expressionSupport = getExpressionSupport(objectField);
-    Optional<DslElementSyntax> fieldDsl = typeDsl.getChild(fieldName);
+    Optional<DslElementSyntax> fieldDsl = typeDsl.getChild(key);
```
pablomusumeci: I'm not really sure about doing that. An infrastructure parameter is kinda special, but it is still a parameter. You are going to have a different key name but the parsing is all the same. I know that we can shortcut it and go straight to the `parseObject(...)` thing, but if the parameter parsing changes in the future, we are still going to want those infrastructure parameters to be parsed in the same way as the others.

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/parameter/ObjectTypeParameterParser.java
pablomusumeci: You will end up needing the .orElse() to deal with the case in which the objectField doesn't have the TypeIdAnnotation.

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/parameter/ObjectTypeParameterParser.java
```diff
@@ -85,6 +86,11 @@ protected void doParse(Builder definitionBuilder) throws ConfigurationException
     type.getFields().forEach(this::parseField);
   }
 
+
+  private Optional<String> getInfrastructureParameterName(MetadataType fieldType) {
+    return Optional.ofNullable(getNameMap().get(getId(fieldType)));
```
marianogonzalez: obtain the namaeMap once, and cache it into a field with a descriptive name

PR #3567, opened 2016-11-11
MULE-10938: Add DW plugin to our tests dependencies

afelisatti:


pom.xml
```diff
@@ -229,6 +229,7 @@
         <testFtpServerVersion>1.0.6</testFtpServerVersion>
         <tomcatVersion>6.0.44</tomcatVersion>
         <unirestVersion>1.4.9</unirestVersion>
+        <weaveVersion>${project.version}</weaveVersion>
```
pablolagreca: I'm not sure this is fine. The weave version should not be tied to the version of mule. It must be versioned independently from the runtime so we don't need to release DW everytime we want to release a new mule version.
Is this just temporary?

afelisatti: For now the version matches ours but yes, perhaps we should use 4.0-SNAPSHOT since we are most likely going to diverge soon.

afelisatti:
jenkins validate

PR #3573, opened 2016-11-14
MULE-10885: Support operation UX

marianogonzalez:


core/src/main/java/org/mule/runtime/core/util/AttributeEvaluator.java
```diff
@@ -42,6 +42,14 @@ public AttributeEvaluator initialize(final ExtendedExpressionManager expressionM
     return this;
   }
 
+  private String sanitize(String attributeValue) {
```
pablolagreca: What is this?. Seems this must be solved elsewhere. Perhaps in the parsers?

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/schema/builder/SchemaBuilder.java
```diff
@@ -494,8 +496,17 @@ TopLevelElement createTypeRef(DslElementSyntax typeDsl, ObjectType type, boolean
     return createRefElement(refQName, isRequired);
   }
 
-  MetadataTypeVisitor getParameterDeclarationVisitor(final ExtensionType extensionType, final ExplicitGroup all,
-                                                     final ParameterModel parameterModel) {
+  void declareAsParameter(MetadataType type, ExtensionType extensionType, ExplicitGroup all, ParameterModel parameterModel) {
+    if (ExtensionModelUtils.isContent(parameterModel) || TypeUtils.isContent(type)) {
```
pablolagreca: why do you have to check both?. Shouldn't the parameter model be enough?

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/parameter/TopLevelParameterObjectFactory.java
```diff
@@ -61,24 +68,56 @@ public Object getObject() throws Exception {
...
+      if (valueResolver != null) {
+        builder.addPropertyResolver(objectField.getName(), valueResolver);
       }
-    }
+    });
+  }
+
+  private Field getField(Class<?> objectClass, String key) {
+    return getFieldByNameOrAlias(objectClass, key)
+        .orElseThrow(() -> new IllegalModelDefinitionException(String.format("Class '%s' does not contain field %s",
```
pablolagreca: static import

core/src/main/java/org/mule/runtime/core/util/AttributeEvaluator.java
marianogonzalez: because the expression can be in a CDATA, it means it can start or end with `\n`, `\t`, spaces and things of the sort....  This removes that and makes it so that the validaiton regex recognizes this as an expression.

What was discussed with Ana is that this should move to the ExpressionLanguage, but at the time of sending this PR that was still not finished up.

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/schema/builder/SchemaBuilder.java
marianogonzalez: no. when parsing top level elements, a synthetic parameterModel is used temporarily in order to reuse code and keep it simpler. It's just an implementation detail

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/parameter/TopLevelParameterObjectFactory.java
marianogonzalez: done

PR #3575, opened 2016-11-14
MULE-10440: Define plugin versions and improve plugin dependencies

pablokraan:
_ Splitting BundleDescriptor in two: BundleDescriptor contains bundle identifier, BundleDependency defines how a dependency is used
_ Updating plugin factory to manage bundleDescriptors instead of plain names
_ Updating plugin dependencies resolver to manage different types of versions

### DIFF
--- a/modules/artifact/src/main/java/org/mule/runtime/module/artifact/descriptor/BundleDependency.java
+++ b/modules/artifact/src/main/java/org/mule/runtime/module/artifact/descriptor/BundleDependency.java
@@ -1,0 +1,163 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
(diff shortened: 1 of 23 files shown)


modules/artifact/src/main/java/org/mule/runtime/module/artifact/descriptor/BundleDependency.java
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * Describes a dependency on a bundle.
+ *
+ * @since 4.0
+ */
+public class BundleDependency {
+
+  private BundleDescriptor descriptor;
+  private String type;
+  private Optional<String> classifier = empty();
```
pablolagreca: shouldn't both the type and classifier be part of the BundleDescriptor?. If BundleDescriptor is the one that uniquely identifies the dependency then it should also have type and classifier. The I guess the BundleDependency declares a particular dependency to a bundle descriptor and that's why it has the scope attribute

PR #3577, opened 2016-11-14
MULE-10868: Migrate scheduler module

elrodro83:


core/src/main/java/org/mule/runtime/core/source/polling/schedule/ScheduledPoll.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+
+/**
+ * Definition of a Scheduler for poll.
+ *
+ * @since 3.5.0, moved from {@link org.mule.runtime.core.api.schedule.Scheduler}.
+ */
+public class ScheduledPoll implements Lifecycle, NameableObject {
```
dfeist: Again with this name, any reason it shouldn't be called something broader than "poll"?  To poll is just one reason you might need to schedule a recurrent task.  Also, this should be with scheduler infrastructure rather than in package with polling source shouldn't it?

PR #3582, opened 2016-11-15
MULE-10974: JDK Resources are not being found when running a test with MuleArtifactClassloader

juandesi:


gsfernandes:
Jenkins validate

PR #3583, opened 2016-11-15
MULE-10937: Separate deployment artifact implementations from deploym…

pablokraan:
…ent module

_ Adding mule-module-deployment-impl module
_ Adding  MuleContextListenerFactory (and default implementation DeploymentMuleContextListenerFactory) to avoid dependency to DeploymentListener on the deployment-impl module
_ Adding missing unit tests
_ Moving classes from  deployment to deployment-impl module
_ Fixing distros


modules/deployment-impl/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>mule-modules</artifactId>
+        <groupId>org.mule.modules</groupId>
+        <version>4.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>mule-module-deployment-impl</artifactId>
+    <packaging>jar</packaging>
+    <name>Mule Deployment Model Implementation Module</name>
```
pablolagreca: Is this the implementation of deployment model or deployment?. If it were the implementation of deployment model I would expect the artifact id to be org.mule.modules:mule-module-deployment-model-impl instead of org.mule.modules:mule-module-deployment-impl

PR #3586, opened 2016-11-15
MULE-10975: AbstractAggregator eventGroups object store ignores the

elrodro83:
object store configuration in the registry

### DIFF
--- a/core/src/main/java/org/mule/routing/AbstractAggregator.java
+++ b/core/src/main/java/org/mule/routing/AbstractAggregator.java
@@ -6,13 +6,15 @@
  */
 package org.mule.routing;
 
+import static org.mule.api.config.MuleProperties.OBJECT_STORE_DEFAULT_IN_MEMORY_NAME;
+import static org.mule.api.config.MuleProperties.OBJECT_STORE_DEFAULT_PERSISTENT_NAME;
+
 import org.mule.VoidMuleEvent;
 import org.mule.api.MessagingException;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mule/routing/AbstractAggregator.java
```diff
@@ -148,27 +148,20 @@ protected Factory internalEventsGroupsObjectStoreFactory()
             @Override
             public Object create()
             {
-                try
+                ObjectStore objectStore;
+                if (persistentStores)
+                {
+                    objectStore = muleContext.getRegistry().lookupObject(OBJECT_STORE_DEFAULT_PERSISTENT_NAME);
```
pablolagreca: This must be done using the ObjectStoreManager, just ask for a persistent object store.

pablolagreca:
Test case?

elrodro83:
The test case is in the cluster module, the pr is in ee

core/src/main/java/org/mule/routing/AbstractAggregator.java
elrodro83: The thing is that in order to support the configurable objectStore, the partitions have to be handled a bit differently. The obtained object will be wrapped in a ProvidedPartitionableObjectStoreWrapper to handle that logic (the creation of the corresponding partitions). The type returned when using the ObjectStoreManager does not support this.

PR #3588, opened 2016-11-15
Mule 10979

facundovs:
Pull Request of mule-10979.

System properties feature has been removed.


core/src/main/java/org/mule/config/DefaultMuleConfiguration.java
```diff
@@ -276,119 +272,6 @@ else if (isStandalone())
...
-        {
-            fullStackTraces = false;
-        }
-        p = System.getProperty(MuleProperties.SYSTEM_PROPERTY_PREFIX + "stacktrace.filter");
-        if (p != null)
-        {
-            stackTraceFilter = p.split(",");
-        }
-
-        p = System.getProperty(MuleProperties.SYSTEM_PROPERTY_PREFIX + "verbose.exceptions");
```
pablokraan: Seems to me that the clean up should not delete applySystemProperties method, as this property is not configurable from the XML, so it does not going to be overwritten when the XML is parsed. Check every property configured in this method and delete only the ones that are configurable on the XML

distributions/standalone/src/main/resources/conf/wrapper.conf
```diff
@@ -43,9 +43,6 @@ wrapper.java.additional.15=-Dorg.quartz.scheduler.skipUpdateCheck=true
 # (see http://wrapper.tanukisoftware.org/doc/english/prop-java-additional-n.html)
 # *** IMPORTANT ***
 
-# Stack traces are truncated in the logs by default
-#wrapper.java.additional.<n>=-Dmule.verbose.exceptions=true
```
pablokraan: Similar to previous comment, delete only the properties that are configurable from the XML

distributions/standalone/src/main/resources/conf/wrapper.conf
```diff
@@ -59,12 +56,10 @@ wrapper.java.additional.15=-Dorg.quartz.scheduler.skipUpdateCheck=true
 #wrapper.java.additional.<n>=-Dhttp.proxyPassword=YOUR_PASSWORD
 
 # Identification of your Mule server
-#wrapper.java.additional.<n>=-Dmule.serverId=YOUR_MULE_SERVER
 #wrapper.java.additional.<n>=-Dmule.clusterId=YOUR_MULE_CLUSTER
-#wrapper.java.additional.<n>=-Dmule.domainId=YOUR_MULE_DOMAIN
 
-# Mule's working directory for SEDA queue persistence, transactions, etc.
-#wrapper.java.additional.<n>=-Dmule.workingDirectory=./.mule
+
```
pablokraan: Remove extra empty lines

tests/integration/src/test/java/org/mule/test/config/MuleConfigurationTestCase.java
```diff
@@ -79,57 +79,6 @@ public void testConfigureProgramatically() throws Exception
         verifyConfiguration();
     }
 
-    /** Test for MULE-3092 */
-    @Test
-    public void testConfigureWithSystemProperties() throws Exception
```
pablokraan: Review if the test has to be delete or just cleaned up

tests/integration/src/test/java/org/mule/test/integration/config/DisableTimeoutsConfigTestCase.java
```diff
@@ -1,56 +0,0 @@
...
-    @Rule
-    public DynamicPort dynamicPort2 = new DynamicPort("port2");
-
-    @Rule
-    public SystemProperty disableTimeouts = new SystemProperty(MuleProperties.SYSTEM_PROPERTY_PREFIX + "timeout.disable", "true");
-
-    @Override
-    protected String getConfigFile()
-    {
-        return "org/mule/test/integration/config/disable-timeouts-config.xml";
```
pablokraan: If this test is removed, you must also remove the configuration file org/mule/test/integration/config/disable-timeouts-config.xml if it is not being used on another test

PR #3596, opened 2016-11-16
MULE-10857: Definition parser doesn't support Connection Providers wi…

pablomusumeci:
…th alias


marianogonzalez:
this PR contains no test case

pablomusumeci:
The schema generated was not the right one. I added an alias to one provider and corrected the XSD we test against to, the SchemaGeneratorTestCase is the one that covers this case.

marianogonzalez:
that's incomplete. You don't only have to test that the schema is correctly generated. You need to test that the parser can eat it. USe the feature in a functional test case

PR #3600, opened 2016-11-17
MULE-10548: Cleanup WorkManager and WorkListener

elrodro83:


core-tests/src/test/java/org/mule/runtime/core/routing/AsynchronousUntilSuccessfulProcessingStrategyTestCase.java
```diff
@@ -120,6 +120,14 @@ public void setUp() throws Exception {
                                                                                                           .build());
   }
 
+  @After
+  public void after() {
```
dfeist: Shouldn't abstract test case be doing this?

core-tests/src/test/java/org/mule/runtime/core/work/MuleEventWorkTestCase.java
```diff
@@ -41,7 +41,7 @@ protected void doSetUp() throws Exception {
 
   @Test
   public void testScheduleMuleEventWork() throws Exception {
-    muleContext.getWorkManager().scheduleWork(new TestMuleEventWork(originalEvent));
+    muleContext.getSchedulerService().computationScheduler().submit(new TestMuleEventWork(originalEvent));
```
dfeist: Why computation? Or is this not important, any pool is the same?

core/src/main/java/org/mule/runtime/core/connector/ConnectException.java
```diff
@@ -67,7 +68,7 @@ private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundE
     }
   }
 
-  public void handleReconnection() {
+  public void handleReconnection(Executor retryExecutor) {
```
dfeist: Why Executor and not ExecutorService?

core/src/main/java/org/mule/runtime/core/routing/AsynchronousUntilSuccessfulProcessingStrategy.java
```diff
@@ -72,19 +69,13 @@ public void initialise() throws InitialisationException {
 
   @Override
   public void start() {
-    final String threadPrefix =
-        String.format("%s%s.%s", ThreadNameHelper.getPrefix(getUntilSuccessfulConfiguration().getMuleContext()),
-                      getUntilSuccessfulConfiguration().getFlowConstruct().getName(), "until-successful");
-    pool = getUntilSuccessfulConfiguration().getThreadingProfile().createPool(threadPrefix);
-    scheduledRetriesPool = getUntilSuccessfulConfiguration().createScheduledRetriesPool(threadPrefix);
+    pool = muleContext.getSchedulerService().ioScheduler();
```
dfeist: Would it be a good idea to allow a name to be passed when a pool is received, so any logging or error that occur can be associated back to where pool is used.  Or is there another mechanism to do this?

tests/integration/src/test/resources/until-successful-seconds-test.xml
```diff
@@ -76,20 +76,4 @@
             </processor-chain>
         </until-successful>
     </flow>
-
-    <flow name="asynchronous-using-threading-profile">
```
dfeist: Why is this removed?

tests/performance/src/test/resources/scatter-gather-perf-test.xml
```diff
@@ -22,15 +22,6 @@
         </scatter-gather>
     </flow>
 
-    <flow name="sequentialProcessing">
```
dfeist: Isn't there another wy to do sequential processing now, to migrate this test to use?

core-tests/src/test/java/org/mule/runtime/core/routing/AsynchronousUntilSuccessfulProcessingStrategyTestCase.java
elrodro83: That's only the case for AbstractMuleContextTestCase. This text extends from AbstractMuleTestCase and creates the service within itself.

core-tests/src/test/java/org/mule/runtime/core/work/MuleEventWorkTestCase.java
elrodro83: Any pool is the same here, since just a test task is scheduled. This task just performs some assertions.

core/src/main/java/org/mule/runtime/core/routing/AsynchronousUntilSuccessfulProcessingStrategy.java
elrodro83: No, that's a good idea actually. Created MULE-11018 for that

tests/performance/src/test/resources/scatter-gather-perf-test.xml
elrodro83: no. Is that still a requirement for scatter gather?

dfeist: I'd check with Mariano G..

PR #3609, opened 2016-11-17
MULE-10997: Show SQL when it fails

rscadrde:


marianogonzalez:
Hello @rscadrde ,

Thank you for this contribution. We will incorporate it into the product. Some comments through:

1) Please sign the contributor's agreement following the instructions in the contribution guide
2) notice that the change you submitted is for Mule 4, which will be released only next year. If you need this for Mule 3.x, you need a separate PR pointing to the old DB connector.

Thanks

extensions/db/src/main/java/org/mule/extension/db/internal/domain/executor/AbstractSingleQueryExecutor.java
```diff
@@ -32,7 +32,12 @@ public Object execute(DbConnection connection, Query query) throws SQLException
 
     prepareQuery(statement, query);
 
-    return doExecuteQuery(connection, statement, query);
+    try {
+    	return doExecuteQuery(connection, statement, query);
+    } catch (SQLException e) {
+    	LOGGER.error("Error executing "+query.getQueryTemplate().getSqlText(), e);
```
pablokraan: This must be configurable, not everybody will want to log the query here

rscadrde:
@marianogonzalez 
is it sufficient that I read 
http://www.mulesoft.org/legal/contributor-agreement.html
or do I need to sign on some kind of mule page?

marianogonzalez:
@rscadrde you need to read the contributors agreement at http://www.mulesoft.org/legal/contributor-agreement.html and sign it by running the script at https://api-notebook.anypoint.mulesoft.com/notebooks#bc1cf75a0284268407e4

rscadrde:
Ok. Agreement notebook done.

@pablokraan where should I add the configuration?

extensions/db/src/main/java/org/mule/extension/db/internal/domain/executor/AbstractSingleQueryExecutor.java
pablokraan: I was thinking about this, and not convinced that this must be parameterized. So skip that until someone complains :)
However, there a couple of things that should be addressed:
1) This change is covering only the single query case, but it should cover the bulk update too.
2) Makes sense to throw a new exception instead of logging it here? (just to avoid logging the same exception twice)
3) The sqlText won't be helpful in all the cases, as it contains placeholders for parameterized queries. You can use the QueryLogger to properly log query parameters too

rscadrde:
1) 
I added the SQL output to bulk query, too.


2) 
Rethrow is difficult as correct SQLException subclass should be thrown with correct codes inside.
So I just left out the Exception in log.

3) 
The QueryLogger is heavily depending on AbstractExecutor and DefaultQueryLoggerFactory.
e.g. org.mule.extension.db.internal.domain.executor.AbstractSingleQueryExecutor.prepareQuery(Statement, Query)
=> I just added the parameter values as log-line.

StoicNZ:
From the past 4 years of manually adding logging for this, thank you **rscadrde**!

PR #3610, opened 2016-11-17
MULE-10937: Separate deployment artifact implementations from deploym…

pablokraan:
…ent module

_ Fixing tests


modules/deployment-model-impl/src/test/resources/org/foo/EchoTest.java
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+
+package org.foo;
+
+public class EchoTest {
```
canoasan: What's the purpose of this class?

PR #3613, opened 2016-11-18
MULE-11002: add cache for XML factories to avoid configuring them (wi…

alepulver:
…th setFeature) which is expensive


core/src/main/java/org/mule/util/xmlsecurity/XMLFactoryConfig.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.util.xmlsecurity;
+
+/**
+ * Cache key for XMLSecureFactoriesCache. Also creates the corresponding factory.
```
elrodro83: link in javadoc

core/src/main/java/org/mule/util/xmlsecurity/XMLFactoryConfig.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+public abstract class XMLFactoryConfig
+{
+    protected Boolean externalEntities;
```
elrodro83: are this required to be protected? cannot be private?

core/src/main/java/org/mule/util/xmlsecurity/XMLFactoryConfig.java
```diff
@@ -0,0 +1,54 @@
...
+        final int primeNumber = 31;
+
+        int result = this.externalEntities.hashCode();
+        result = primeNumber * result + this.expandEntities.hashCode();
+        result = primeNumber * result + this.factoryName.hashCode();
+
+        return result;
+    }
+
+    public abstract Object createFactory();
```
elrodro83: is there no common superclass to what is being built?

core/src/main/java/org/mule/util/xmlsecurity/XMLSecureFactories.java
```diff
@@ -0,0 +1,67 @@
...
+    {
+        XMLSecureFactories factory = new XMLSecureFactories();
+
+        factory.externalEntities = externalEntities;
+        factory.expandEntities = expandEntities;
+
+        return factory;
+    }
+
+    public static XMLSecureFactories createDefault()
```
elrodro83: why not getDefault and return always the same instance?

core/src/main/java/org/mule/util/xmlsecurity/XMLSecureFactoriesImpl.java
```diff
@@ -18,40 +17,13 @@
 import org.apache.commons.logging.LogFactory;
 
 /**
- * Provide XML parser factories configured to avoid XXE and BL attacks according to global configuration (safe by default)
+ * Create different XML factories configured through the same interface for disabling vulnerabilities.
  */
-public class XMLSecureFactories
+public class XMLSecureFactoriesImpl
```
elrodro83: rename to DefualtXMLSecureFactories

core/src/main/java/org/mule/util/xmlsecurity/XMLFactoryConfig.java
alepulver: If they are private they would require getters to be accessed from the conflict classes that implement createFactory.

core/src/main/java/org/mule/util/xmlsecurity/XMLFactoryConfig.java
alepulver: No, they are different interfaces to XML parsers made at different times. All "standard" but not compatible :)

PR #3618, opened 2016-11-18
MULE-11001: ArtifactClassLoaderRunner - Lookup policies for Plugin in…

gsfernandes:
… Region should honor dependencies


transports/file/src/main/resources/plugin.properties
```diff
@@ -1,5 +1,6 @@
 #Plugin Properties
 #TODO MULE-10735 generate this file from maven dependencies so we won't have issues with the hard-coded version, packages or resources there
+plugin.dependencies=mule-module-http-ext-4.0-SNAPSHOT
```
pablokraan: Why there are many transports that depend on HTTP extension?

gsfernandes: Because mule-transport-module-support depends on HTTP extension (for some constants needed that would be later added to the HTTP service API), therefore as all the transports depends on this support module they require to have this plugin dependency declared in order to access those classes.

tests/runner/src/test/java/org/mule/test/runner/classloader/PluginLookPolicyFactoryTestCase.java
```diff
@@ -0,0 +1,100 @@
...
+import java.net.URL;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.ArgumentCaptor;
+
+public class PluginLookPolicyFactoryTestCase {
```
pablokraan: Extends from AbstractMuleTestCase

PR #3591, opened 2016-11-16
MULE-10866: support to use ExtensionModel for executing the templatin…

fernandezlautaro:
…g mechanism.

*do not accept this PR until https://github.com/mulesoft/mule-extensions-api/pull/201 is merged*
This PR enables the templating mechanism to work through the `ExtensionModel`.
The custom Describer was coded in extensions-xml-support, where from an XML it reads it completely while generating a ExtensionDeclaration.

@marianogonzalez: I've found some minor issues while using the `DefaultExtensionFactory` (that's why I added a constructor without enrichers neither validators), but the entire loading mechanism works similarly to what the SDK does. The ugly part comes to the deployment, where does a fallback mechanism.

@pablokraan + @marianogonzalez  (see above): the current implementation of `org.mule.runtime.module.deployment.internal.application.ApplicationExtensionsManagerConfigurationBuilder` tries to look for an SDK extension from a given ArtifactPlugin, and iff it's missing then fallbacks to the JSON describer in the method `#discoverExtensionThroughJsonDescriber`. This is a WIP as there is other JIRA (MULE-10876) meant to unify both concepts (the `ExtensionManifest` should die and the JSON descriptor must take its place).

@pablolagreca: changed bunch of stuff on spring-config module, but it's far simpler than before. The dynamism it's entirely based on `ExtensionModel` objects, which means that we could potentially drop the XSD generations (see `org.mule.runtime.config.spring.ModuleDelegatingEntityResolver#getSchema`) for any extension The only thing I don't like at all is the discovery mechanism:
```
Optional<ExtensionModel> extensionModel = extensionManager.get().getExtensions().stream()
          .filter(em -> systemId.startsWith(em.getXmlDslModel().getNamespaceUri())) // <-- not the best code here
          .findAny();
```
we look by the `namespaceUri` with a `startsWith` (we could talk directly)

I did *not* run the tests, do not merge until I confirm this doesn't blows everything up

PR #3629, opened 2016-11-21
MULE-10866 Make templating mechanism work with ExtensionModel for its generation

fernandezlautaro:
Same PR as [before](https://github.com/mulesoft/mule/pull/3591), fixing the broken build due to missing dependencies (and a some tests I didn't run)


modules/spring-config/src/main/java/org/mule/runtime/config/spring/ModuleDelegatingEntityResolver.java
```diff
@@ -35,14 +37,19 @@
...
   public ModuleDelegatingEntityResolver(Optional<ExtensionManager> extensionManager) {
     this.entityResolver = new DelegatingEntityResolver(Thread.currentThread().getContextClassLoader());
     this.extensionManager = extensionManager;
     ServiceRegistry spiServiceRegistry = new SpiServiceRegistry();
-    schemaResourceFactory =
-        spiServiceRegistry.lookupProvider(SchemaResourceFactory.class, getClass().getClassLoader());
+    try {
+      // TODO(fernandezlautaro): MULE-11024 until the implementation is moved up to extensions-api, we need to work with Optional to avoid breaking the mule testing framework (cannot add the dependency, as it will imply a circular dependency)
+      schemaResourceFactory =
+          of(spiServiceRegistry.lookupProvider(SchemaResourceFactory.class, getClass().getClassLoader()));
```
pablokraan: LookupProvider will throw IllegalStateException when there is zero or multiple providers. Setting empty is not the proper way to manage this situation. I think that you should use getProviders instead and manage the different results

modules/spring-config/src/main/java/org/mule/runtime/config/spring/ModuleDelegatingEntityResolver.java
```diff
@@ -56,7 +63,8 @@ public InputSource resolveEntity(String publicId, String systemId) throws SAXExc
 
   private InputSource generateModuleXsd(String publicId, String systemId) {
     InputSource inputSource = null;
-    if (extensionManager.isPresent()) {
+    if (extensionManager.isPresent()
```
pablokraan: Is not an error if the extensionManager is absent and the schemaResourceFactory is present? (and viceversa)

fernandezlautaro: ExtensionManager could be absent, but the schemaResourceFactory must not.
I'll add the same TODO to the `if` condition so it gets removed once MULE-11024 is done

modules/spring-config/src/main/java/org/mule/runtime/config/spring/ModuleDelegatingEntityResolver.java
fernandezlautaro: done

PR #3636, opened 2016-11-21
MULE-11031: Cannot reference flowVars directly using DW

afelisatti:


core/src/main/java/org/mule/runtime/core/el/v2/MuleExpressionLanguage.java
```diff
@@ -122,7 +122,11 @@ private void addEventBindings(Event event, BindingContext.Builder contextBuilder
       Error error = event.getError().isPresent() ? event.getError().get() : null;
       contextBuilder.addBinding(ERROR, new DefaultTypedValue(error, fromType(Error.class)));
       Map<String, TypedValue> flowVars = new HashMap<>();
-      event.getVariableNames().forEach(name -> flowVars.put(name, event.getVariable(name)));
+      event.getVariableNames().forEach(name -> {
```
pablolagreca: This must be done before any other binding, so variables names do not override other binding like "error", "payload"

PR #3637, opened 2016-11-21
	MULE-10992 Change default processing strategy to remove single-threaded limitation for request-response. Some tests need more threads.

dfeist:


modules/cxf/src/test/java/org/mule/runtime/module/cxf/ProxySoapVersionTestCase.java
```diff
@@ -18,6 +18,7 @@
 import org.junit.Rule;
 import org.junit.Test;
 
+// TODO MULE-11035 - Migrate extension tests that depend on multiple threads to use MuleArtifactFunctionalTestCase has been successfully created.
```
elrodro83: what does "has been successfully created" mean?

PR #3642, opened 2016-11-22
MULE-11039: Handle threading errors with the error handling mechanism

elrodro83:
Do not merge yet!! Discussion pending

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/exception/ErrorTypeRepositoryFactory.java
+++ b/core/src/main/java/org/mule/runtime/core/exception/ErrorTypeRepositoryFactory.java
@@ -8,11 +8,13 @@ package org.mule.runtime.core.exception;
 
 import static org.mule.runtime.core.exception.Errors.ComponentIdentifiers.CONNECTIVITY;
 import static org.mule.runtime.core.exception.Errors.ComponentIdentifiers.EXPRESSION;
+import static org.mule.runtime.core.exception.Errors.ComponentIdentifiers.OVERLOAD;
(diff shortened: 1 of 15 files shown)


core/src/main/java/org/mule/runtime/core/exception/ErrorTypeRepositoryFactory.java
```diff
@@ -38,6 +40,7 @@ public static ErrorTypeRepository createDefaultErrorTypeRepository() {
     errorTypeRepository.addErrorType(RETRY_EXHAUSTED, connectivityErrorType);
     errorTypeRepository.addErrorType(ROUTING, errorTypeRepository.getAnyErrorType());
     errorTypeRepository.addErrorType(SECURITY, errorTypeRepository.getAnyErrorType());
+    errorTypeRepository.addErrorType(OVERLOAD, errorTypeRepository.getAnyErrorType());
```
afelisatti: Use CRITICAL if you don't want it to be handled by users.

services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultSchedulerService.java
```diff
@@ -111,7 +111,10 @@ public Scheduler cpuIntensiveScheduler() {
   @Override
   public Scheduler customScheduler(String name, int corePoolSize) {
     final ExecutorService executor =
-        newFixedThreadPool(corePoolSize, new SchedulerThreadFactory(customGroup, "%s." + name + ".%02d"));
+        new ThreadPoolExecutor(corePoolSize, corePoolSize,
```
dfeist: Why? javadoc to explain?

PR #3647, opened 2016-11-23
MULE-10926 - Support policy composition

pablolagreca:


core/src/main/java/org/mule/runtime/core/policy/ParameterizedPolicy.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.util.Map;
+
+/**
+ * This class represents a policy injection.
+ * <p>
+ * It contains the set of parameters configured to apply the policy and the chain of
+ * {@link org.mule.runtime.core.api.processor.Processor}s to be applied.
+ */
+public class ParameterizedPolicy {
```
pablokraan: Should not implement Policy?

core/src/main/java/org/mule/runtime/core/policy/CompositeOperationPolicy.java
```diff
@@ -0,0 +1,96 @@
...
+import org.mule.runtime.core.api.Event;
+import org.mule.runtime.core.api.policy.OperationPolicyParametersTransformer;
+import org.mule.runtime.core.api.policy.SourcePolicyParametersTransformer;
+import org.mule.runtime.core.api.processor.Processor;
+
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * {@link OperationPolicy} created from a list of {@link ParameterizedPolicy}.
```
pablokraan: Why operation and source policies depend on a concrete ParameterizedPolicy class?

core/src/main/java/org/mule/runtime/core/policy/DefaultSourcePolicy.java
```diff
@@ -0,0 +1,104 @@
...
+      throws Exception {
+    Processor sourceNextOperation =
+        buildFlowExecutionWithPolicyFunction(nextOperation, sourceEvent, messageSourceResponseParametersProcessor);
+    policyStateHandler.updateNextOperation(sourceEvent.getContext().getId(), sourceNextOperation);
+    Event result = policyChain
+        .process(policyEventConverter.createEvent(sourceEvent.getMessage(), Event.builder(sourceEvent.getContext()).build()));
+    return Event.builder(result.getContext()).message(result.getMessage()).build();
+  }
+
+  private Processor buildFlowExecutionWithPolicyFunction(Processor nextOperation, Event sourceEvent,
```
pablokraan: I understand that this code was moved... but I don't get that it does...

core/src/main/java/org/mule/runtime/core/policy/CompositeOperationPolicy.java
```diff
@@ -0,0 +1,96 @@
...
+
+    public NextOperationCall(Event operationEvent, Processor operationProcessor,
+                             OperationParametersProcessor operationParametersProcessor) {
+      this.operationEvent = operationEvent;
+      this.operationProcessor = operationProcessor;
+      this.operationParametersProcessor = operationParametersProcessor;
+    }
+
+    @Override
+    public Event process(Event event) throws MuleException {
```
pablokraan: All this logic must be tested on a unit test

core/src/main/java/org/mule/runtime/core/policy/CompositeOperationPolicy.java
pablolagreca: It's an utility class that we have just to group a policy and it's parameters. It does not have any behaviour that we can benefit from having an interface, at least for now.

core/src/main/java/org/mule/runtime/core/policy/CompositeOperationPolicy.java
```diff
@@ -0,0 +1,96 @@
...
+      if (index >= parameterizedPolicies.size()) {
```
pablolagreca: Not, is just the end of the loop.

core/src/main/java/org/mule/runtime/core/policy/PolicyManager.java
```diff
@@ -29,7 +29,7 @@
    *
    * @param executionIdentifier the unique identifier associated to the exection of a policy instance.
    * @param sourceIdentifier the identifier of the message source
-   * @return a {@link OperationPolicy} associated to that source.
+   * @return a {@link DefaultOperationPolicy} associated to that source.
```
pablolagreca: yes, that's a mistake in the javadoc.

core/src/main/java/org/mule/runtime/core/policy/DefaultOperationPolicy.java
```diff
@@ -0,0 +1,93 @@
...
+      throws Exception {
+    Optional<Event> latestPolicyState = policyStateHandler.getLatestState(operationEvent.getContext().getId());
+    Event variablesProviderEvent = latestPolicyState.orElseGet(() -> Event.builder(operationEvent.getContext()).build());
+    Map<String, Object> originalParametersMap = operationParametersProcessor.getOperationParameters();
+    Message message = operationEvent.getMessage();
+    message = operationPolicyParametersTransformer
+        .map(parametersTransformer -> parametersTransformer.fromParametersToMessage(originalParametersMap)).orElse(message);
+    Event policyEvent = policyEventConverter.createEvent(message, variablesProviderEvent);
+    Processor operationCall = buildOperationExecutionWithPolicyFunction(nextProcessor);
+    policyStateHandler.updateNextOperation(operationEvent.getContext().getId(), operationCall);
```
pablolagreca: The operationCall is pass to the stateHandler so it can be stored and later retrieve by PolicyNextActionMessageProcessor.

pablolagreca: Yes it is. Do you mean because many policies will be store state using the same ID?. If that's the case then you are right. I will fix it.

core/src/main/java/org/mule/runtime/core/policy/DefaultPolicy.java
```diff
@@ -0,0 +1,122 @@
...
+import org.mule.runtime.core.util.UUID;
+
+import java.util.Optional;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+//TODO MULE-10963 - Remove FlowConstruct implementation once MPs don't depend on FlowConstructAware anymore.
+public class DefaultPolicy implements Policy, FlowConstruct, MuleContextAware, Lifecycle {
```
pablolagreca: It's used in EE. EE will define the namespaces of the different policies types and will always use this class as object created from the XML.

core/src/main/java/org/mule/runtime/core/policy/DefaultSourcePolicy.java
```diff
@@ -0,0 +1,104 @@
...
+ * used to execute.
+ */
+public class DefaultSourcePolicy implements SourcePolicy {
+
+  private final PolicyChain policyChain;
+  private final Optional<SourcePolicyParametersTransformer> sourcePolicyParametersTransformer;
+  private final PolicyStateHandler policyStateHandler;
+  private final PolicyEventConverter policyEventConverter = new PolicyEventConverter();
+
+  public DefaultSourcePolicy(PolicyChain policyChain,
```
pablolagreca: done

core/src/main/java/org/mule/runtime/core/policy/ParameterizedPolicy.java
pablolagreca: I think there's a naming issue here. I will rename this class to Policy since the sum of the parameters + the policy chain represents a policy and rename the Policy to PolicyInstance. Eventually there will be a factory for PolicyInstance which will be called PolicyFactory.

core/src/main/java/org/mule/runtime/core/policy/DefaultSourcePolicy.java
pablolagreca: I added javadoc

core/src/main/java/org/mule/runtime/core/policy/CompositeOperationPolicy.java
```diff
@@ -0,0 +1,96 @@
...
+        operationResponseEvent = operationProcessor.process(event);
+        return operationResponseEvent;
+      }
+      ParameterizedPolicy parameterizedPolicy = parameterizedPolicies.get(index);
```
pablolagreca: I added the explanation as javadoc.
/**
     * When this policy is processed, it will use the {@link NextOperationCall} which will keep track of the different policies to be applied and
     * the current index of the policy under execution.
     * <p>
     * The first time, the first policy in the {@code parameterizedPolicies} will be executed, it will receive as it next operation the same
     * instance of {@link NextOperationCall}, and since {@link NextOperationCall} keeps track of the policy executed, it will execute the following
     * policy in the chain until the finaly policy it's executed in which case then next operation of it, it will be the operation execution.
     */

core/src/main/java/org/mule/runtime/core/policy/CompositeOperationPolicy.java
pablokraan: End of the loop is when index == size, otherwise is an error because it entered in the loop more than the expected.

core/src/main/java/org/mule/runtime/core/policy/CompositeSourcePolicy.java
```diff
@@ -0,0 +1,92 @@
...
+
+  /**
+   * Inner class that implements the actually chaining of policies.
+   */
+  public class NextOperationCall implements Processor {
+
+    private int index = 0;
+    private Processor nonPolicyNextOperation;
+    private Event sourceEvent;
+    private MessageSourceResponseParametersProcessor messageSourceResponseParametersProcessor;
```
pablokraan: Right, but don you think that there is enough duplicate code to try to generalize it?

core/src/main/java/org/mule/runtime/core/policy/DefaultOperationPolicy.java
pablokraan: Yes, that is what I mean. So, is this inconsistent with the previous comment saying that policyStateHandler does not have to be thread safe?

core/src/main/java/org/mule/runtime/core/policy/DefaultPolicyInstance.java
```diff
@@ -27,9 +27,9 @@
 
 
 //TODO MULE-10963 - Remove FlowConstruct implementation once MPs don't depend on FlowConstructAware anymore.
-public class DefaultPolicy implements Policy, FlowConstruct, MuleContextAware, Lifecycle {
+public class DefaultPolicyInstance implements PolicyInstance, FlowConstruct, MuleContextAware, Lifecycle {
```
pablokraan: Why is not this class part of EE if is not used on CE?

core/src/main/java/org/mule/runtime/core/policy/CompositeSourcePolicy.java
pablolagreca: I'm doing that.

core/src/main/java/org/mule/runtime/core/policy/DefaultSourcePolicyFactory.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Default implementation for {@link SourcePolicyFactory}.
+ *
+ * @since 4.0
+ */
+public class DefaultSourcePolicyFactory implements SourcePolicyFactory {
+
+  private final PolicyStateHandler policyStateHandler;
+
+  public DefaultSourcePolicyFactory(PolicyStateHandler policyStateHandler) {
```
pablokraan: Add javadoc

core/src/main/java/org/mule/runtime/core/policy/AbstractCompositePolicy.java
```diff
@@ -0,0 +1,126 @@
...
+      this.nextProcessor = nextProcessor;
+      this.originalEvent = originalEvent;
+      this.parametersProcessor = parametersProcessor;
+    }
+
+    @Override
+    public Event process(Event event) throws MuleException {
+      if (index == parameterizedPolicies.size()) {
+        return processNextOperation(nextProcessor, event);
+      }
```
pablokraan: if index> parameterizedPolicies.size() throw exception

PR #3655, opened 2016-11-24
MULE-10942	Review and define processing type for all core processors.

dfeist:


core/src/main/java/org/mule/runtime/core/api/transformer/Transformer.java
```diff
@@ -95,4 +97,14 @@
    * @since 3.0.0
    */
   DataType getReturnDataType();
+
+  @Override
+  default ProcessingType getProccesingType() {
+    if (getSourceDataTypes().stream().filter(dataType -> !dataType.isStreamType()).count() > 0) {
+      return IO_RW;
+    } else {
+      return CPU_LITE;
```
elrodro83: shouldn't transformers be CPU_INTENSIVE by default?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
```diff
@@ -274,4 +275,10 @@ protected ParameterValueResolver getParameterValueResolver() {
     final Event event = getInitialiserEvent(muleContext);
     return new OperationParameterValueResolver(new LazyExecutionContext<>(resolverSet, operationModel, extensionModel, event));
   }
+
+  @Override
+  public ProcessingType getProccesingType() {
+    return BLOCKING;
```
elrodro83: this is just until there is non-blocking support in extensions, right? can you add a TODO?

core/src/main/java/org/mule/runtime/core/api/transformer/Transformer.java
dfeist: Only complex transformers should be CPU intensive, most of the built-in transformers are very simple.  I can make some changes to improve this though..

PR #3657, opened 2016-11-24
MULE-10711: Include body in request/response builder with payload as default

afelisatti:


tests/http/src/test/resources/http-request-source-target-config.xml
```diff
@@ -11,11 +11,23 @@
     </httpn:request-config>
 
     <flow name="payloadSourceFlow">
-        <httpn:request config-ref="config" path="test" method="POST" source="#[payload]"/>
+        <httpn:request config-ref="config" path="test" method="POST">
+            <httpn:request-builder>
+                <httpn:body>
+                    #[dw:payload]
```
marianogonzalez: redundant. body already defaults to payload automatically

afelisatti: Yes, but the test is to verify this scenario also works.

PR #3672, opened 2016-11-25
MULE-10585: Make the container ThreadPools configurable

elrodro83:


services/scheduler/src/main/java/org/mule/service/scheduler/internal/config/ThreadPoolsConfig.java
```diff
@@ -0,0 +1,215 @@
...
+      compile("(" + NUMBER_OR_C_REGEXP + ")?(\\s*[-+\\/*]\\s*(" + NUMBER_OR_C_REGEXP + ")?)*");
+
+  /**
+   * Loads the configuration from the {@code &#123;mule.home&#125;/conf/scheduler-pools.conf} file.
+   * 
+   * @return The loaded configuration, or the default if the file is unavailable.
+   * @throws MuleException for any trouble that happens while parsing the file.
+   */
+  public static ThreadPoolsConfig loadThreadPoolsConfig() throws MuleException {
+    File muleHome = getProperty("mule.home") != null ? new File(getProperty("mule.home")) : null;
```
dfeist: Is there an existing constant you can use for this, if not then introduce one here in this class.

distributions/standalone/src/main/resources/conf/scheduler-pools.conf
```diff
@@ -0,0 +1,27 @@
+# Configures the thread pools of the Mule Runtime
+# Certain attributes support expressions. Those are arithmetic expressions (+, -, / or * operators supported only) that are
+# evaluated when the runtime starts.
+# Those expressions may be applied to numbers (integers or decimals) or any of the following variables:
+# * C: the number of processors available to the Java virtual machine.
+
+# The maximum time to wait until all tasks in all the runtime thread pools have completed execution when stopping the
+# scheduler service.
+SchedulerService.gracefulShutdownTimeoutSeconds=15
```
dfeist: Is this case sensitive? Typically property files use lower case for keys. Like, for example, what quartz does `org.quartz.threadPool.threadCount`

distributions/standalone/src/main/resources/conf/scheduler-pools.conf
```diff
@@ -0,0 +1,27 @@
...
+
+# The number of threads to keep in the cpu_lite pool, even if they are idle.
+# Supports Expressions
+SchedulerService.cpuLight.threadPoolSize=2*C
```
dfeist: any reason not to use `2*cores`.  It more readable and more obviously apparent what it is.

distributions/standalone/src/main/resources/conf/scheduler-pools.conf
dfeist: Isn't this best exposed as `ms`, else smallest timeout is 1s and resolution is limited.  (Also can unit go in comment, rather than making property name longer?)

distributions/standalone/src/main/resources/conf/scheduler-pools.conf
```diff
@@ -0,0 +1,27 @@
...
+
+# The number of threads to keep in the I/O pool.
+# Supports Expressions
+SchedulerService.io.coreThreadPoolSize=C
+# The maximum number of threads to allow in the I/O pool.
+# Supports Expressions
+SchedulerService.io.maxThreadPoolSize=C*C
+# When the number of threads in the I/O pool is greater than SchedulerService.io.coreThreadPoolSize, this is the maximum
+# time (in seconds) that excess idle threads will wait for new tasks before terminating.
+SchedulerService.io.threadKeepAliveSeconds=30
```
dfeist: Isn't this best exposed as `ms`, else smallest timeout is 1s and resolution is limited.  (Also can unit go in comment, rather than making property name longer?)

services/scheduler/src/main/java/org/mule/service/scheduler/internal/config/ThreadPoolsConfig.java
```diff
@@ -0,0 +1,217 @@
...
+      properties.load(configIs);
+    } catch (IOException e) {
+      throw new DefaultMuleException(e);
+    }
+
+    ScriptEngineManager manager = new ScriptEngineManager();
+    ScriptEngine engine = manager.getEngineByName("js");
+    engine.put("C", config.cores);
+
+    config.setGracefulShutdownTimeoutSeconds(resolveNumber(properties, PROP_PREFIX + "gracefulShutdownTimeoutSeconds"));
```
dfeist: Compose these strings with constants for "cpuLight" "IO" "coreThreadPoolSize" etc.

services/scheduler/src/main/java/org/mule/service/scheduler/internal/config/ThreadPoolsConfig.java
```diff
@@ -0,0 +1,217 @@
...
+        throw new DefaultMuleException(propName + ": Value has to be greater than 0");
+      }
+
+      return result;
+    } catch (ScriptException e) {
+      throw new DefaultMuleException(propName + ": " + e.getMessage(), e);
+    }
+  }
+
+  private int cores = getRuntime().availableProcessors();
```
dfeist: Any reason this can't be static, although I guess it doesn't make much difference.

distributions/standalone/src/main/resources/conf/scheduler-pools.conf
```diff
@@ -0,0 +1,27 @@
+# Configures the thread pools of the Mule Runtime
```
dfeist: IMO we should use a  mule specific prefix for these properties and also support allowing them to be configured via system properties.

e.g. `org.mule.runtime.scheduler.io.core_threads`

services/scheduler/src/main/java/org/mule/service/scheduler/internal/config/ThreadPoolsConfig.java
```diff
@@ -0,0 +1,217 @@
...
+    final Properties properties = new Properties();
+    try (final FileInputStream configIs = new FileInputStream(defaultConfigFile)) {
+      properties.load(configIs);
+    } catch (IOException e) {
+      throw new DefaultMuleException(e);
+    }
+
+    ScriptEngineManager manager = new ScriptEngineManager();
+    ScriptEngine engine = manager.getEngineByName("js");
+    engine.put("C", config.cores);
```
dfeist: `c` -> `cores`

distributions/standalone/src/main/resources/conf/scheduler-pools.conf
elrodro83: For this case, even a time of 1 sec would be too low and affect performance badly (threads being created and destroyed all time). I'd leave it as seconds.

distributions/standalone/src/main/resources/conf/scheduler-pools.conf
```diff
@@ -2,26 +2,26 @@
...
 
 # The number of threads to keep in the cpu_lite pool, even if they are idle.
 # Supports Expressions
-SchedulerService.cpuLight.threadPoolSize=2*C
+org.mule.runtime.scheduler.cpuLight.threadPoolSize=2*cores
 
 # The number of threads to keep in the I/O pool.
 # Supports Expressions
-SchedulerService.io.coreThreadPoolSize=C
+org.mule.runtime.scheduler.io.threadPoolSize.core=cores
```
dfeist: Can we simplify this to simple `scheduler.io.coreThreads` or`scheduler.io.minThreads`?

distributions/standalone/src/main/resources/conf/scheduler-pools.conf
```diff
@@ -2,26 +2,26 @@
...
-# scheduler service.
-SchedulerService.gracefulShutdownTimeoutSeconds=15
+# The maximum time (in milliseconds) to wait until all tasks in all the runtime thread pools have completed execution when
+# stopping the scheduler service.
+org.mule.runtime.scheduler.gracefulShutdownTimeout=15000
 
 # The number of threads to keep in the cpu_lite pool, even if they are idle.
 # Supports Expressions
-SchedulerService.cpuLight.threadPoolSize=2*C
+org.mule.runtime.scheduler.cpuLight.threadPoolSize=2*cores
```
dfeist: Is there any way to make this naming consistent with other thread size?  The fact that this doesn't have core/max and others do is confusing. It might be clearer if:
- we supported core/max, explained why we recommend they are the same in description. (i prefer this one)
- use just max size, and explain why we dont expose core size, or the other way around.

distributions/standalone/src/main/resources/conf/scheduler-pools.conf
dfeist: I'm not sure it's fair to assume that? .5s is a long time for a CPU but can't configure 500ms. ThreadPoolExecutor supports a resolution of nanoseconds, so why restrict this to seconds.  Of course, it's easier if we don't have to support units here in this properties file, but exposing as ms is a good solution for me, and also makes it consistent with other time value in this properties file making them all the same unit.

distributions/standalone/src/main/resources/conf/scheduler-pools.conf
elrodro83: I'd prefer having a consistent prefix, so core and max look like attributes of the pool

dfeist: Then we should use:
`threadPool.coreSize`
`threadPool.maxSize`
`threadPool.keepAlive`  (not this should be attribute of theedPool too.)

PR #3674, opened 2016-11-26
MULE-11084: Improve schema for validations

alegmarra:
* Improve validation of required/optional elements
* Order parameters: 
- First any custom/prefixed element like "reconnection"
- Then the parameters of the "Generic" group
- Finally the parameters from other groups, sorted by how they were declared in the group 
(no extra sorting, just taking the order provided by the model)

### DIFF
--- a/modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/schema/builder/ConfigurationSchemaDelegate.java
+++ b/modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/schema/builder/ConfigurationSchemaDelegate.java
@@ -8,13 +8,14 @@ package org.mule.runtime.module.extension.internal.capability.xml.schema.builder
 
 import static java.math.BigInteger.ONE;
 import static java.math.BigInteger.ZERO;
+import static java.util.Optional.empty;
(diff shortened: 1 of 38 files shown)


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/schema/builder/ConfigurationSchemaDelegate.java
```diff
@@ -38,25 +43,48 @@
...
-    choice.setMaxOccurs(UNBOUNDED);
+    Optional<TopLevelElement> connectionElement = addConnectionProviderElement(configurationModel);
+    Optional<TopLevelElement> policyElement = addDynamicConfigPolicyElement(configurationModel);
 
-    addConnectionProviderElement(choice, configurationModel);
-    addDynamicConfigPolicyElement(choice, configurationModel);
-    builder.registerParameters(config, choice, configurationModel.getAllParameterModels());
-    config.setAnnotation(builder.createDocAnnotation(configurationModel.getDescription()));
+    Map<String, Map<String, TopLevelElement>> groups = new LinkedHashMap<>();
+    configurationModel.getParameterGroupModels()
```
alegmarra: @marianogonzalez please check this makes sense for you, since is an example of the rules for sorting the parameters in the schema

PR #3692, opened 2016-11-30
MULE-11118: return 503 when thread pool is exhausted

alepulver:


modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/ExecutorPerServerAddressIOStrategy.java
```diff
@@ -29,6 +30,7 @@
  */
 public class ExecutorPerServerAddressIOStrategy extends AbstractIOStrategy
 {
+    protected final static String UNAVAILABLE_ATTRIBUTE = "__reject_request__";
```
dfeist: You could use `org.glassfish.grizzly.attributes.Attribute<Boolean>` here, but up to you.

dfeist: Can you use an attribute slightly more explicit?  Whats unavailable?

modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerWorkerThreadingProfileTestCase.java
```diff
@@ -100,11 +94,12 @@ private void assertMaxThreadsActive(String flowName, String url, int maxThreadsA
         sendRequestUntilNoMoreWorkers(flowName, url, maxThreadsActive);
         // Due to differences in buffer sizes, the exception that is caused client side may be one of two different
         // exceptions.
-        expectedException.expect(anyOf(instanceOf(NoHttpResponseException.class), instanceOf(SocketException
-                                                                                                     .class)));
+        //expectedException.expect(anyOf(instanceOf(NoHttpResponseException.class), instanceOf(SocketException.class)));
```
dfeist: Why is this commented out?

PR #3694, opened 2016-11-30
MULE-10585: Make the container ThreadPools configurable

elrodro83:


distributions/standalone/src/main/resources/conf/scheduler-pools.conf
```diff
@@ -11,17 +13,26 @@ org.mule.runtime.scheduler.gracefulShutdownTimeout=15000
 # The number of threads to keep in the cpu_lite pool, even if they are idle.
 # Supports Expressions
 org.mule.runtime.scheduler.cpuLight.threadPool.size=2*cores
+# The size of the queue to use for holding cpu_lite tasks before they are executed.
+# Supports Expressions
+org.mule.runtime.scheduler.cpuLight.workQueue.size=mem / (2*3*32)
```
dfeist: Why not simply `cpuLight.queueSize`, are there, or will there be, other properties of the queue?

distributions/standalone/src/main/resources/conf/scheduler-pools.conf
```diff
@@ -11,17 +13,26 @@ org.mule.runtime.scheduler.gracefulShutdownTimeout=15000
...
 
 # The number of threads to keep in the I/O pool.
 # Supports Expressions
 org.mule.runtime.scheduler.io.threadPool.coreSize=cores
 # The maximum number of threads to allow in the I/O pool.
 # Supports Expressions
-org.mule.runtime.scheduler.io.threadPool.maxSize=cores*cores
+org.mule.runtime.scheduler.io.threadPool.maxSize=mem / (2*30*64)
```
dfeist: Where do these values come from?  Why should size be based on memory available? IMO we should just be using fixed values, or potentially a multiple `threadPool.maxSize` of threads for these, at least until testing.

distributions/standalone/src/main/resources/conf/scheduler-pools.conf
```diff
@@ -11,17 +13,26 @@ org.mule.runtime.scheduler.gracefulShutdownTimeout=15000
...
+# The size of the queue to use for holding I/O tasks before they are executed.
+# Supports Expressions
+org.mule.runtime.scheduler.io.workQueue.size=mem / (2*3*32)
```
dfeist: Where do these values come from? Why should size be based on memory available? IMO we should just be using fixed values, or potentially a multiple threadPool.maxSize for these, at least until testing.

distributions/standalone/src/main/resources/conf/scheduler-pools.conf
elrodro83: could be, but at least this is consistent with the pool size property

PR #3696, opened 2016-12-01
MULE-11114: Add DataWeave to Runtime distribution

anosenzo:


distributions/standalone/pom.xml
```diff
@@ -38,6 +38,33 @@
                 </executions>
             </plugin>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>copy-plugins</id>
+                        <phase>process-resources</phase>
```
aiannucci: Why not `package` phase?

distributions/standalone/pom.xml
```diff
@@ -38,6 +38,33 @@
...
+                        <goals>
+                            <goal>copy</goal>
+                        </goals>
+                        <configuration>
+                            <skip>${skipDistributions}</skip>
+                            <useBaseVersion>true</useBaseVersion>
```
aiannucci: If you set `outputDirectory` as `${project.build.directory}/plugins` then you avoid modifying the `assembly.xml`

distributions/standalone/pom.xml
```diff
@@ -38,6 +38,33 @@
...
+                            <artifactItems>
+                                <artifactItem>
+                                    <groupId>com.mulesoft.weave</groupId>
+                                    <artifactId>mule-plugin-weave</artifactId>
+                                    <version>${weaveVersion}</version>
+                                    <classifier>dist</classifier>
+                                    <type>zip</type>
+                                </artifactItem>
```
aiannucci: Instead of specifying the artifact here, can't you use the `copy-dependencies` goal? It will make more sense if we want to see what are the distribution dependencies with `dependency:tree`

PR #3698, opened 2016-12-01
MULE-10870 - Duplicate content-type created in response builder

canoasan:


tests/http/src/test/java/org/mule/test/module/http/functional/listener/HttpListenerSuccessResponseBuilderTestCase.java
```diff
@@ -108,6 +113,16 @@ public void headersResponseBuilder() throws Exception {
   }
 
   @Test
+  public void twoHeadersResponseBuilder() throws Exception {
+    final String url = getUrl(twoHeadersResponseBuilderPath);
+    final Response response = Request.Get(url).connectTimeout(1000).execute();
+    final StatusLine statusLine = response.returnResponse().getStatusLine();
+
+    assertThat(statusLine.getStatusCode(), is(500));
+    assertThat(statusLine.getReasonPhrase(), is("Header Content-Type does not support multiple values"));
```
afelisatti: Can you open a JIRA about this? Reason phrase should be the 500 one and the error should be in the body IMO.

PR #3699, opened 2016-12-01
In progress

alepulver:


afelisatti:
Fix the commit message so that it references a MULE issue. Might simply be "Add test for..." in this case but we still need to track it.

PR #3700, opened 2016-12-01
MULE-10970: Make all message processor AnnotatedObject

gsfernandes:


transports/core/src/main/java/org/mule/compatibility/core/transport/AbstractMessageDispatcher.java
```diff
@@ -21,17 +22,23 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import javax.xml.namespace.QName;
 
 /**
  * Abstract implementation of an outbound channel adaptors. Outbound channel adaptors send messages over over a specific
  * transport. Different implementations may support different Message Exchange Patterns.
  */
 public abstract class AbstractMessageDispatcher extends AbstractTransportMessageHandler
```
afelisatti: Why didn't it make sense to make things higher up AnnotatedObjects?

afelisatti:
Are you sure none of this require ObjectFactories to be parsed?

gsfernandes:
https://www.mulesoft.org/jira/browse/MULE-11140 in order to migrate ObjectFactories

PR #3706, opened 2016-12-01
MULE-8777: add support for multiple values for the same http query pa…

alepulver:
…rameter


modules/http/src/main/java/org/mule/module/http/internal/HttpMapParam.java
```diff
@@ -37,30 +37,54 @@ public void setExpression(String expression)
...
             else
             {
                 parameterMap.put(paramName, toStringIfPossible(paramValue));
             }
-
         }
     }
 
-    private String toStringIfPossible(Object paramValue)
+    private static String toStringIfPossible(Object paramValue)
```
afelisatti: Why did you need to make this static?

PR #3708, opened 2016-12-02
MULE-11133 Create initial flow performance test case

dfeist:


tests/performance/src/test/java/org/mule/test/core/FlowPerformanceTestCase.java
```diff
@@ -0,0 +1,197 @@
...
+
+  protected void addBuilders(List<ConfigurationBuilder> builders) {
+    builders.add(new BasicRuntimeServicesConfigurationBuilder());
+  }
+
+  @Before
+  public void setup() throws MuleException {
+    muleContext.start();
+
+    defaultSource = new TriggerableMessageSource();
```
elrodro83: wouldn't it be better to use a parameterized test case?

PR #3715, opened 2016-12-03
MULE-11138: Make easier to work with UDT on DB connector

pablokraan:
_ Adding support to create java.sql.Struct and java.sql.Array values from Object[] and List values
_ Adding EL functions to create array and struct values from MEL expressions

### DIFF
--- a/modules/db/src/main/java/org/mule/module/db/internal/domain/connection/OracleDbConnection.java
+++ b/modules/db/src/main/java/org/mule/module/db/internal/domain/connection/OracleDbConnection.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+
+package org.mule.module.db.internal.domain.connection;
+
+import org.mule.module.db.internal.domain.transaction.TransactionalAction;
+import org.mule.module.db.internal.resolver.param.ParamTypeResolverFactory;
+
+import java.lang.reflect.Method;
+import java.sql.Array;
+import java.sql.Connection;
+import java.sql.SQLException;
+
+/**
+ * Custom {@link DbConnection} for Oracle databases
+ */
+public class OracleDbConnection extends DefaultDbConnection
+{
+
+    /**
+     * {@inheritDoc}
+     */
+    public OracleDbConnection(Connection delegate, TransactionalAction transactionalAction, DefaultDbConnectionReleaser connectionReleaseListener, ParamTypeResolverFactory paramTypeResolverFactory)
+    {
+        super(delegate, transactionalAction, connectionReleaseListener, paramTypeResolverFactory);
+    }
+
+    @Override
+    public Array createArrayOf(String typeName, Object[] elements) throws SQLException
+    {
+        Method createArrayMethod;
+        try
+        {
+            createArrayMethod = delegate.getClass().getMethod("createARRAY", String.class, Object.class);
+            createArrayMethod.setAccessible(true);
+        }
+        catch (NoSuchMethodException e)
+        {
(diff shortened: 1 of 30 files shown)


modules/db/src/main/java/org/mule/module/db/internal/domain/connection/OracleDbConnection.java
```diff
@@ -0,0 +1,56 @@
...
+        super(delegate, transactionalAction, connectionReleaseListener, paramTypeResolverFactory);
+    }
+
+    @Override
+    public Array createArrayOf(String typeName, Object[] elements) throws SQLException
+    {
+        Method createArrayMethod;
+        try
+        {
+            createArrayMethod = delegate.getClass().getMethod("createARRAY", String.class, Object.class);
```
marianogonzalez: don't do this lookup every time. cache this method on the constructor and wrap it in some kind of strategy object

modules/db/src/main/java/org/mule/module/db/internal/domain/type/ArrayResolvedDbType.java
```diff
@@ -32,9 +37,35 @@ public ArrayResolvedDbType(int id, String name)
...
     }
 
+    /**
+     * Creates error message for the case when a given class is not supported
+     *
+     * @param value value that was attempted to be converted
+     * @return the error message for the provided value's class
+     */
+    protected static String createUnsupportedTypeErrorMessage(Object value) {
+        return format("Cannot create a %s from a value of type %s", Struct.class.getName(), value.getClass());
```
marianogonzalez: Shouldn't you use the class simpleName instead?

modules/db/src/main/java/org/mule/module/db/internal/el/AbstractDbFunction.java
```diff
@@ -0,0 +1,134 @@
...
+import org.mule.module.db.internal.domain.transaction.TransactionalAction;
+import org.mule.module.db.internal.resolver.database.DbConfigResolver;
+
+import java.sql.SQLException;
+import java.util.List;
+
+/**
+ * Provides a base class for DB functions that creates custom DB types.
+ */
+public abstract class AbstractDbFunction  implements ExpressionLanguageFunction
```
marianogonzalez: good luck with this in Mule 4. I wish you all the best

modules/db/src/test/java/org/mule/module/db/integration/function/DbCreateArrayFunctionTestCase.java
```diff
@@ -0,0 +1,74 @@
...
+        testDatabase.createStoredProcedureUpdateZipCodes(dataSource);
+        testDatabase.createStoredProcedureUpdateContactDetails(dataSource);
+    }
+
+    @Test
+    public void createsDefaultTypeArray() throws Exception
+    {
+        LocalMuleClient client = muleContext.getClient();
+
+        MuleMessage response = client.send("vm://createsDefaultTypeArray", NORTHWEST.getZips(), null);
```
marianogonzalez: do you really need to use vm for this tests? CAn't you just runFlow()?

modules/db/src/main/java/org/mule/module/db/internal/domain/type/ArrayResolvedDbType.java
pablokraan: No, full qualified name is required to univocally reference a class

modules/db/src/main/java/org/mule/module/db/internal/el/AbstractDbFunction.java
pablokraan: I don't need good luck, I'm Muleman

modules/db/src/test/java/org/mule/module/db/integration/function/DbCreateArrayFunctionTestCase.java
pablokraan: No really required, I did it in this way as all the tests in the module use the same approach. I can change it

PR #3726, opened 2016-12-07
MULE-10890: In Email when you don't add a mandatory field like body i…

pablomusumeci:
…t throws an exception with "null" message


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/validation/NullSafeModelValidator.java
```diff
@@ -0,0 +1,151 @@
...
+                                                                            fieldName,
+                                                                            declaringClass.getName(),
+                                                                            NullSafe.class.getSimpleName(),
+                                                                            fieldType.getName()));
+                }
+              }
+
+              @Override
+              public void visitObject(ObjectType objectType) {
+                // TODO: Reenable these checks when MULE-10679 brings introspectionUtils to a visible scope
```
marianogonzalez: uncomment this code. IntrospectionUtils is in scope now

PR #3734, opened 2016-12-12
MULE-11109: getPayloadForLogging changes stream.

facundovs:
ObjectToString and DefaultMuleMessage classes were modified.
PayloadForLogginBreaksStreamTestCase and bug-login-breajks-stream were added.

### DIFF
--- a/core/src/main/java/org/mule/DefaultMuleMessage.java
+++ b/core/src/main/java/org/mule/DefaultMuleMessage.java
@@ -80,6 +80,7 @@ import org.apache.commons.logging.LogFactory;
 public class DefaultMuleMessage implements MuleMessage, ThreadSafeAccess, DeserializationPostInitialisable
 {
     protected static final String NOT_SET = "<not set>";
+    private static final int UTF_INVALID_CHARACTER= 65533;
 
     private static final long serialVersionUID = 1541720810851984845L;
     private static final Log logger = LogFactory.getLog(DefaultMuleMessage.class);
@@ -760,7 +761,10 @@ public class DefaultMuleMessage implements MuleMessage, ThreadSafeAccess, Deseri
         String result = getPayload(DataType.STRING_DATA_TYPE, encoding);
         if (muleContext.getConfiguration().isCacheMessageAsBytes())
         {
-            cache = result.getBytes(encoding);
+            if (isValidUT8Character(result))
+            {
+                cache = result.getBytes(encoding);
+            }
         }
         return result;
     }
@@ -1424,7 +1428,7 @@ public class DefaultMuleMessage implements MuleMessage, ThreadSafeAccess, Deseri
         assertAccess(WRITE);
         properties.clearProperties(scope);
     }
-    
+
     /**
      * {@inheritDoc}
      */
@@ -1470,6 +1474,13 @@ public class DefaultMuleMessage implements MuleMessage, ThreadSafeAccess, Deseri
         if (payload == null)
         {
             this.payload = NullPayload.getInstance();
+        }
+        else if (payload instanceof String)
+        {
+            if (isValidUT8Character((String) payload))
+            {
+                this.payload = payload;
+            }
         }
         else
         {
@@ -1825,7 +1836,7 @@ public class DefaultMuleMessage implements MuleMessage, ThreadSafeAccess, Deseri
 
     /**
      * @deprecated since 3.8.0. Use {@link ClassUtils#isConsumable(Class)} instead.
-     * 
+     *
      * Determines if the payload of this message is consumable i.e. it can't be read
      * more than once.
      */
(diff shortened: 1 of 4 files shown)


tests/integration/src/test/java/org/mule/PayloadForLoggingBreaksStreamTestCase.java
```diff
@@ -0,0 +1,67 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule;
+
+import static org.junit.Assert.*;
```
aiannucci: Don't use wildcards in imports

tests/integration/src/test/java/org/mule/PayloadForLoggingBreaksStreamTestCase.java
```diff
@@ -0,0 +1,67 @@
...
+import org.junit.Test;
+
+public class PayloadForLoggingBreaksStreamTestCase extends FunctionalTestCase
+{
+    private static final int invalidUTFValue = -128;
+
+    @Test
+    public void testCallGetPayloadForLogging() throws Exception
+    {
+        assertEquals("Set file encoding to UTF-8 for this test", "UTF-8", System.getProperty("file.encoding"));
```
aiannucci: Get "UTF-8" literal from `StandardCharsets` class

core/src/main/java/org/mule/DefaultMuleMessage.java
```diff
@@ -760,7 +761,10 @@ public String getPayloadAsString(String encoding) throws Exception
         String result = getPayload(DataType.STRING_DATA_TYPE, encoding);
         if (muleContext.getConfiguration().isCacheMessageAsBytes())
         {
-            cache = result.getBytes(encoding);
+            if (isValidUT8Character(result))
```
marianogonzalez: this makes no sense. WHy to check for the first character to be utf-8 valid when you don't even know if `encoding` is utf-8 in the first place. This also seems to be fixing a symptom and not the root problem

core/src/main/java/org/mule/transformer/simple/ObjectToString.java
```diff
@@ -85,6 +85,7 @@ protected String createStringFromInputStream(InputStream input, String outputEnc
             try
             {
                 input.close();
+                input.reset();
```
marianogonzalez: you can't do this.

1) It makes no sense to reset a closed stream. With a simple ByteArrayInputStream it might work, but with an InputStream coming from ftp, http, sockets, etc. there're no guarantees. 
2) You're violating the reset() contract. This method only works on SOME input streams and provided the mark() method was previously called. You can't assume either here
3) Why did you do this in the first place? Looks like there's some issue that's been hidden rather than fixed

tests/integration/src/test/java/org/mule/PayloadForLoggingBreaksStreamTestCase.java
```diff
@@ -0,0 +1,71 @@
...
+
+public class PayloadForLoggingBreaksStreamTestCase extends FunctionalTestCase
+{
+    private static final int invalidUTFValue = -128;
+    private static final String UTF_8_CONSTANT = UTF_8.name();
+
+    @Test
+    public void testCallGetPayloadForLogging() throws Exception
+    {
+        assertEquals("Set file encoding to UTF-8 for this test",UTF_8_CONSTANT, System.getProperty("file.encoding"));
```
marianogonzalez: test should not assume that the system encoding is utf-8. This is flaky

tests/integration/src/test/resources/bug-login-breaks-stream.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<mule xmlns="http://www.mulesoft.org/schema/mule/core"
```
marianogonzalez: don't create an empty config just because. Use one of the existing configs.... On a second thought, you shouldn't need a config at all to reproduce this issue

core/src/main/java/org/mule/DefaultMuleMessage.java
facundovs: I suppoused the encoding condition is implicit (if the encoding isn't utf-8, the � character wouldn't exist). However, this condition could be added in the if statement explicitly.

tests/integration/src/test/resources/bug-login-breaks-stream.xml
facundovs: Ok. I tried to do a unit test for this issue. However, DefaultMuleMessage needs a MuleContext to be tested. I could reuse an existing config.

tests/integration/src/test/java/org/mule/PayloadForLoggingBreaksStreamTestCase.java
facundovs: Ok. I could set the encoding through a system property.

core/src/main/java/org/mule/transformer/simple/ObjectToString.java
marianogonzalez: i don't fully understand what you're saying, but it sounds like the solution doesn't work without this, so I guess that revert it is.

PR #3735, opened 2016-12-12
MULE-11159: Fix: FileToString Transform is not able to process incoming message payload retrieved from file connector when streaming attribute is set to false.

fsgonz:


transports/file/src/main/java/org/mule/transport/file/transformers/FileToString.java
```diff
@@ -19,6 +19,7 @@
 
     public FileToString()
     {
+        registerSourceType(DataTypeFactory.BYTE_ARRAY);
```
fsgonz: Originally, the conflict arised because TypeBasedTransformerResolver was unable to resolve the transformer to be used between FileToByteArray transformer and ObjectToByteArray. 

In the case of FileToString and ObjectToString, no conflict would arise as FileToString (as well as FileToByteArray) has a default priority weighting, while ObjectToString transformer (as well as ObjectToByteArray) has a higher priority weighting than default. I downloaded the app from the concerning issue (MULE-7630, SE-596) and no test failures occurred.  

I've changed the test so that FileToString and ObjectToString are involved and the transformer is resolved correctly. In case the weighting priority is modified in these latter classes, the issue arises. So the problems similar to the one described in SE-596 should not be replicated becaused of the change implemented in this commit.

transports/file/src/test/java/org/mule/transport/file/FileNoStreamingTestCase.java
```diff
@@ -0,0 +1,57 @@
...
+    protected String getConfigFile()
+    {
+        return "file-no-streaming-config.xml";
+    }
+
+    @Test
+    public void fileToStringNoStreamingFileConnector() throws Exception
+    {
+        File tmpDir = FileUtils.openDirectory(getFileInsideWorkingDirectory("in").getAbsolutePath());
+        createDataFile(tmpDir, TEST_MESSAGE, StandardCharsets.UTF_8.name());
```
afelisatti: static import

PR #3736, opened 2016-12-12
MULE-11174: Missing exports in spring-config module

gsfernandes:


marianogonzalez:
i'll merge it because this is already exporting yaml... but this will have to go away once Kraan comes with the guadana

gsfernandes:
Agree, but right now it was wrong the packages exported due to it required to be exported those two too.

PR #3737, opened 2016-12-12
MULE-11161: change PGP encryption cipher from CAST5 to AES-256

alepulver:


marianogonzalez:
this should include a note in the migration guide. Also, is there a use case in which the user might one to go back to the prior one? Shouldn't we have a mechanism to specify the cipher?

aiannucci:
I agree that this should be user configurable. Also, I'm almost sure that this change breaks the security module implementation. Changes will be needed there as well.

alepulver:
Yes, I'll add the note. We could add a system property, but it wasn't configurable before.

marianogonzalez:
@alepulver I understand that it wasn't configurable before, but you cannot claim backwards compatibility without a mechanism to go back to the old behavior.

As @aiannucci said:

* you need to check how this impacts AES. If a new AES version is needed to work with this change, then that has to be clearly documented on the migration guide and you need to let the documentation team now as well.
* A system property is a good way to do it if this fix needs to go on a patch release. However, if this is only for 3.9 onwards, then make it explicit on the DSL.

marianogonzalez:
where does the need for this change comes from anyway? Is this source of problems or did you just saw an easy improvement and went for it?

alepulver:
@marianogonzalez the issue was assigned to me with high priority. I'll make it configurable for future releases and make a patch for the customer that requested it.

modules/pgp/src/main/java/org/mule/module/pgp/KeyBasedEncryptionStrategy.java
```diff
@@ -43,6 +45,11 @@ public void initialise() throws InitialisationException
             java.security.Security.addProvider(new BouncyCastleProvider());
         }
         provider = SecurityUtils.getDefaultSecurityProvider();
+
+        if (encryptionAlgorithm == null)
+        {
+            encryptionAlgorithm = SymmetricKeyAlgorithmTags.CAST5;
```
marianogonzalez: static import

modules/pgp/src/main/java/org/mule/module/pgp/KeyBasedEncryptionStrategy.java
```diff
@@ -137,4 +144,9 @@ public void setCheckKeyExpirity(boolean checkKeyExpirity)
     {
         this.checkKeyExpirity = checkKeyExpirity;
     }
+
+    public void setEncryptionAlgorithm(Integer encryptionAlgorithm)
+    {
+        this.encryptionAlgorithm = encryptionAlgorithm;
```
marianogonzalez: where is this setter invoked? I don't see any code using it

alepulver: From the seccurity module.

PR #3738, opened 2016-12-12
MULE-10904: Provide a way for the extension developer to inject mule …

pablomusumeci:
…default encoding

### DIFF
--- a/extensions/email/src/main/java/org/mule/extension/email/api/EmailBody.java
+++ b/extensions/email/src/main/java/org/mule/extension/email/api/EmailBody.java
@@ -44,19 +44,19 @@ public class EmailBody {
   private String contentType;
 
   /**
-   * The character encoding of the body. If not specified, it defaults to the default charset in the mule configuration
+   * The character encoding of the body. If not specified, it defaults to the default encoding in the mule configuration
    */
   @Parameter
   @Optional
   @Placement(order = 3)
-  private String charset;
+  private String encoding;
 
   public EmailBody() {}
 
-  public EmailBody(String content, MediaType contentType, String charset) {
+  public EmailBody(String content, MediaType contentType, String encoding) {
     this.content = content;
     this.contentType = contentType.toString();
-    this.charset = charset;
+    this.encoding = encoding;
   }
 
   /**
@@ -74,9 +74,9 @@ public class EmailBody {
   }
 
   /**
-   * @return the charset of the body.
+   * @return the encoding of the body.
    */
-  public String getCharset() {
-    return charset;
+  public String getEncoding() {
+    return encoding;
   }
 }
--- a/extensions/jms/src/main/java/org/mule/extensions/jms/api/config/JmsConfig.java
+++ b/extensions/jms/src/main/java/org/mule/extensions/jms/api/config/JmsConfig.java
@@ -13,19 +13,16 @@ import org.mule.extensions.jms.api.operation.JmsConsume;
 import org.mule.extensions.jms.api.operation.JmsPublish;
 import org.mule.extensions.jms.api.operation.JmsPublishConsume;
 import org.mule.extensions.jms.api.source.JmsSubscribe;
-import org.mule.runtime.api.lifecycle.Initialisable;
-import org.mule.runtime.api.lifecycle.InitialisationException;
-import org.mule.runtime.core.api.MuleContext;
 import org.mule.runtime.extension.api.annotation.Configuration;
 import org.mule.runtime.extension.api.annotation.Expression;
 import org.mule.runtime.extension.api.annotation.Operations;
+import org.mule.runtime.extension.api.annotation.param.DefaultEncoding;
 import org.mule.runtime.extension.api.annotation.Sources;
 import org.mule.runtime.extension.api.annotation.param.NullSafe;
 import org.mule.runtime.extension.api.annotation.param.Optional;
 import org.mule.runtime.extension.api.annotation.param.Parameter;
 import org.mule.runtime.extension.api.annotation.param.display.Placement;
 
-import javax.inject.Inject;
 import javax.jms.Message;
 import javax.jms.Queue;
 import javax.jms.Topic;
@@ -38,24 +35,13 @@ import javax.jms.Topic;
 @Configuration(name = "config")
 @Operations({JmsConsume.class, JmsPublish.class, JmsPublishConsume.class, JmsAck.class})
 @Sources({JmsSubscribe.class})
-public class JmsConfig implements Initialisable {
-
-  @Inject
-  private MuleContext muleContext;
-
-  //TODO MULE-10904: remove this logic
-  @Override
-  public void initialise() throws InitialisationException {
-    if (encoding == null) {
-      encoding = muleContext.getConfiguration().getDefaultEncoding();
-    }
-  }
+public class JmsConfig {
 
   /**
    * the encoding of the {@link Message} {@code body}
    */
   @Parameter
-  @Optional
+  @DefaultEncoding
   @Expression(NOT_SUPPORTED)
   private String encoding;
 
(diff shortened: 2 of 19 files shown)


extensions/jms/src/main/java/org/mule/extensions/jms/api/config/JmsConfig.java
```diff
@@ -32,24 +31,17 @@
  */
 @Configuration(name = "config")
 @Operations({JmsConsume.class, JmsPublish.class, JmsPublishConsume.class, JmsAck.class})
-public class JmsConfig implements Initialisable {
+public class JmsConfig {
 
   @Inject
   private MuleContext muleContext;
```
pablomusumeci: Forget to delete this

extensions/email/src/main/java/org/mule/extension/email/api/EmailBody.java
```diff
@@ -44,19 +44,19 @@
...
 
   /**
-   * The character encoding of the body. If not specified, it defaults to the default charset in the mule configuration
+   * The character encoding of the body. If not specified, it defaults to the default encoding in the mule configuration
    */
   @Parameter
   @Optional
   @Placement(order = 3)
-  private String charset;
+  private String encoding;
```
marianogonzalez: make sure that no system or integration test is affected by this

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/parameter/TopLevelParameterObjectFactory.java
```diff
@@ -107,9 +110,11 @@ private void resolveParameters(ObjectType objectType, DefaultObjectBuilder build
       if (parameters.containsKey(key)) {
         valueResolver = toValueResolver(parameters.get(key));
       } else if (!isParameterGroup) {
-        valueResolver = getDefaultValue(field)
-            .map(value -> new TypeSafeExpressionValueResolver<>(value, objectField.getType(), muleContext))
-            .orElse(null);
+        if (getDefaultValue(field).isPresent()) {
```
marianogonzalez: bceause of the prior comments, this code doesn't make sense anymore. and for the record, you could still do this with an Optional.map(), just put the if inside of it

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/parameter/TopLevelParameterObjectFactory.java
```diff
@@ -132,4 +137,9 @@ private Field getField(Class<?> objectClass, String key) {
                                                                       objectClass.getName(),
                                                                       key)));
   }
+
+  @Override
+  public void setMuleContext(MuleContext context) {
```
marianogonzalez: don't. If you need the mule context get it in the constructor.

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/parameter/TopLevelParameterObjectFactory.java
```diff
@@ -43,8 +45,9 @@
  *
  * @since 4.0
  */
-public class TopLevelParameterObjectFactory extends AbstractExtensionObjectFactory<Object> {
+public class TopLevelParameterObjectFactory extends AbstractExtensionObjectFactory<Object> implements MuleContextAware {
```
marianogonzalez: don't implement MuleContextAware

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/describer/AnnotationsBasedDescriber.java
```diff
@@ -568,6 +571,22 @@ private void declareConnectionProvider(HasConnectionProviderDeclarer declarer, C
     return declarerList;
   }
 
+  private void parseDefaultEncoding(ExtensionParameter extensionParameter, ParameterDeclarer parameter) {
+    // TODO: MULE-9220 - Add a syntax validator which checks that the annotated parameter is a String
+    if (extensionParameter.getAnnotation(DefaultEncoding.class).isPresent()) {
+      if (!extensionParameter.isAnnotatedWith(org.mule.runtime.extension.api.annotation.param.Optional.class)) {
```
marianogonzalez: forget about the optional

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/model/property/DefaulEncodingModelProperty.java
```diff
@@ -0,0 +1,36 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.module.extension.internal.model.property;
+
+import org.mule.runtime.api.meta.model.ModelProperty;
+import org.mule.runtime.api.meta.model.parameter.ParameterModel;
+
+/**
+ * A {@link ModelProperty} intended to be used on {@link ParameterModel parameters}
+ * to signal that if the parameter is resolved to {@code null}, then the runtime should
```
marianogonzalez: careful here. One thing is `resolves to null` and another thing is `is not provided`. What you did is not about resolving to null , but about not being provided. Two very different things. If you want to make this about not resolving to null, then you need to wrap the resolver instead of using a static one like you did. I think you should allow values resolving to null. Users might want to default to different stuff, like email does.

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/AbstractExtensionObjectFactory.java
```diff
@@ -275,6 +276,30 @@ private ConfigurationException buildExclusiveParametersException(ParameterizedMo
...
   }
 
+  /**
+   * Gets a {@link ValueResolver} for the {@code parameterModel} if it has an associated a default value or encoding.
+   *
+   * @param parameterModel to be resolved
+   * @return a {@link ValueResolver} for cases such having a non null default value or depending on the runtime's default
+   *         encoding, {@code null} otherwise.
+   */
+  private ValueResolver<?> getValueResolverForAbsentParameter(ParameterModel parameterModel) {
```
marianogonzalez: this method seems to overlap with the piece of code you added on resolveParameters(). Also the name of this method is misleading, since it's only for default encoding, not for any absent one

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/describer/model/ExtensionParameter.java
```diff
@@ -53,7 +54,7 @@ default boolean shouldBeAdvertised() {
    * @return A {@code boolean} indicating whether the parameter is a required or not
    */
   default boolean isRequired() {
-    return !(isAnnotatedWith(Optional.class));
+    return !(isAnnotatedWith(Optional.class) || isAnnotatedWith(DefaultEncoding.class));
```
marianogonzalez: don't do this. instead make the change in annotationBasedDescriber

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/describer/AnnotationsBasedDescriber.java
```diff
@@ -568,6 +571,13 @@ private void declareConnectionProvider(HasConnectionProviderDeclarer declarer, C
     return declarerList;
   }
 
+  private void parseDefaultEncoding(ExtensionParameter extensionParameter, ParameterDeclarer parameter) {
+    // TODO: MULE-9220 - Add a syntax validator which checks that the annotated parameter is a String
+    if (extensionParameter.getAnnotation(DefaultEncoding.class).isPresent()) {
+      parameter.withModelProperty(new DefaultEncodingModelProperty());
```
marianogonzalez: also set the parameter as not required here

PR #3739, opened 2016-12-12
MULE-10569: Support non-blocking operations in Ext-API

marianogonzalez:


core/src/main/resources/META-INF/mule-module.properties
```diff
@@ -215,7 +215,10 @@ artifact.export.classPackages=org.mule.runtime.core,\
                               javax.net.ssl,\
                               org.apache.commons.io.filefilter,\
                               javax.net,\
-                              org.reactivestreams, \
+                              org.reactivestreams,\
```
dfeist: Shouldn't expose reactor really.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/DefaultExecutionMediator.java
```diff
@@ -146,7 +150,11 @@ private OperationExecutionResult executeWithInterceptors(OperationExecutor execu
       after(context, result, interceptors);
     }
 
-    return new OperationExecutionResult(result, exception, Optional.ofNullable(retryRequestHolder.get()));
+    Mono<Object> reactiveResult = result instanceof Publisher
```
dfeist: Why is this needed?  WhenIs `result` not a Mono?

dfeist:
Are there any tests that assert that the scheduler task after a non-blocking operation is not the same as the one before?

marianogonzalez:
there's no way to test that AFAIK, and even if there were , that would be flaky since I might get the same thread anyway.

This relies on the mule runtime to actually work and in the fact that the NB methods are void, so you HAVE to work with the callback

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
```diff
@@ -120,41 +125,54 @@ public OperationMessageProcessor(ExtensionModel extensionModel,
...
+        .build();
+    classLoader = getExtensionClassLoader();
   }
 
   @Override
   public Event process(Event event) throws MuleException {
-    return withContextClassLoader(getExtensionClassLoader(), () -> {
-      Optional<ConfigurationInstance> configuration = getConfiguration(event);
+    try {
+      return just(event).transform(this).subscribe().block();
```
dfeist: You don't need subscribe() here I don't think.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
```diff
@@ -120,41 +125,54 @@ public OperationMessageProcessor(ExtensionModel extensionModel,
...
-        MuleEvent muleEvent = doProcess(event, operationContext);
-        return (Event) muleEvent;
-      };
-
-      OperationPolicy policy =
-          policyManager.createOperationPolicy(operationIdentifier, event, operationParameters, operationExecutionFunction);
-      return policy.process(event);
+      ExecutionContextAdapter operationContext = createExecutionContext(configuration, operationParameters, event);
+
+      if (operationModel.isBlocking()) {
```
dfeist: If would be better if you did this if outside of concatMap, and simply use map() operator for this case.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
```diff
@@ -120,41 +125,54 @@ public OperationMessageProcessor(ExtensionModel extensionModel,
...
+        OperationExecutionFunction operationExecutionFunction =
+            (parameters, operationEvent) -> doProcess(operationEvent, operationContext).block();
+        OperationPolicy policy =
+            policyManager.createOperationPolicy(operationIdentifier, event,
+                                                operationParameters,
+                                                operationExecutionFunction);
+        return justOrEmpty(policy.process(event));
```
dfeist: Is null an expected result here?  Doing this will drop the event, and no callback will be invoked handing the source.  You can avoid this by using `nullSafeMap` which completes the eventContext.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
```diff
@@ -120,41 +125,54 @@ public OperationMessageProcessor(ExtensionModel extensionModel,
...
 
-  protected org.mule.runtime.api.message.MuleEvent doProcess(Event event, ExecutionContextAdapter operationContext)
+  protected Mono<Event> doProcess(Event event, ExecutionContextAdapter operationContext)
       throws MuleException {
-    Object result = executeOperation(operationContext, event);
-    MuleEvent resultEvent = returnDelegate.asReturnValue(result, operationContext);
-    return resultEvent;
+    return executeOperation(operationContext, event)
+        .map(value -> returnDelegate.asReturnValue(value, operationContext))
+        .otherwiseIfEmpty(justOrEmpty(returnDelegate.asReturnValue(null, operationContext)));
```
dfeist: I don't think you need to create a Mono here, as `asReturnValue` doesn't do anything async does it?  Would `Mono.otherwiseReturn` be suitable instead?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
```diff
@@ -120,41 +125,54 @@ public OperationMessageProcessor(ExtensionModel extensionModel,
...
   }
 
-  private Object executeOperation(ExecutionContextAdapter operationContext, Event event) throws MuleException {
+  private Mono<Object> executeOperation(ExecutionContextAdapter operationContext, Event event) throws MuleException {
     try {
       return executionMediator.execute(operationExecutor, operationContext);
     } catch (MessagingException e) {
```
dfeist: If you are working with Mono you shouldn't be throwing/catching exceptions. instead using onErrrorXX methods.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/DefaultExecutionMediator.java
```diff
@@ -103,17 +109,15 @@ public Object execute(OperationExecutor executor, ExecutionContextAdapter contex
     }
```
dfeist: Be aware that when returning a Mono, finally doesn't work in the same way.  You might need to use `Mono.doAfterTerminate()` instead.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/DefaultExecutionMediator.java
```diff
@@ -103,17 +109,15 @@ public Object execute(OperationExecutor executor, ExecutionContextAdapter contex
...
   }
 
-  private Object executeWithRetryPolicy(OperationExecutor executor, ExecutionContextAdapter context,
-                                        List<Interceptor> interceptors)
+  private Mono<Object> executeWithRetryPolicy(OperationExecutor executor, ExecutionContextAdapter context,
+                                              List<Interceptor> interceptors)
       throws Throwable {
```
dfeist: Why does this throw Throwable?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/DefaultExecutionMediator.java
```diff
@@ -88,7 +93,8 @@ public DefaultExecutionMediator(ExtensionModel extensionModel, OperationModel op
    * @return the operation's result
    * @throws Exception if the operation or a {@link Interceptor#before(ExecutionContext)} invokation fails
    */
-  public Object execute(OperationExecutor executor, ExecutionContextAdapter context) throws Throwable {
+  @Override
+  public Mono<Object> execute(OperationExecutor executor, ExecutionContextAdapter context) throws Throwable {
```
dfeist: Why does this throw Throwable?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/DefaultExecutionMediator.java
```diff
@@ -146,7 +150,11 @@ private OperationExecutionResult executeWithInterceptors(OperationExecutor execu
       after(context, result, interceptors);
     }
 
-    return new OperationExecutionResult(result, exception, Optional.ofNullable(retryRequestHolder.get()));
+    Mono<Object> reactiveResult = result instanceof Publisher
+        ? from((Publisher) result)
+        : justOrEmpty(result);
```
dfeist: In what case would empty need to be returned?  Empty will filter event in flow.. unless it's handled in `OperationMessageProcessor` specifically.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExecutionMediator.java
```diff
@@ -29,5 +31,5 @@
    * @return the operation's result
    * @throws Exception if any exception is encountered
```
dfeist: No exceptions with Mono

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/InterceptingExecutionMediator.java
```diff
@@ -39,8 +42,8 @@ public InterceptingExecutionMediator(ExecutionMediator intercepted) {
   }
 
   @Override
-  public Object execute(OperationExecutor executor, ExecutionContextAdapter context) throws Throwable {
-    Object resultValue = intercepted.execute(executor, context);
+  public Mono<Object> execute(OperationExecutor executor, ExecutionContextAdapter context) throws Throwable {
+    Object resultValue = intercepted.execute(executor, context).block();
```
dfeist: Why block here.  Seems you are blocking and then using `just` to create Mono from value later?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/DefaultExecutionMediator.java
```diff
@@ -146,7 +150,11 @@ private OperationExecutionResult executeWithInterceptors(OperationExecutor execu
       after(context, result, interceptors);
```
dfeist: Is it safe to assume all before/after interceptors are ok to be run on an event loop?

dfeist:
There is some magic Rodro put in place to assert on task instead of thread. See https://github.com/mulesoft/mule/blob/mule-4.x/tests/integration/src/test/resources/non-blocking-fully-supported-test-config.xml#L134

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/ReactiveOperationExecutionWrapper.java
```diff
@@ -0,0 +1,94 @@
...
+
+    ExecutionContextAdapter<OperationModel> context = (ExecutionContextAdapter<OperationModel>) executionContext;
+    return Mono.create(sink -> {
+      ReactorCallback callback = new ReactorCallback(sink, context.getEvent());
+      context.setVariable(REACTIVE_CALLBACK_CONTEXT_PARAM, callback);
+
+      try {
+        delegate.execute(executionContext);
+      } catch (Exception e) {
+        sink.error(e);
```
dfeist: Doesn't this need wrapping in a MessagingException as we discussed?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/ReactorCallback.java
```diff
@@ -0,0 +1,44 @@
...
+ * project Reactor.
+ *
+ * @since 4.0
+ */
+final class ReactorCallback implements NonBlockingCallback<Object, Attributes> {
+
+  private final MonoSink<Object> sink;
+  private final Event event;
+
+  ReactorCallback(MonoSink<Object> sink, Event event) {
```
dfeist: Isn't generic type Result<Object, Attributes>?

core/src/main/resources/META-INF/mule-module.properties
marianogonzalez: i will add reactor as a dependency of extension-support to avoid exposing it. It makes no sense to build a facade for it IMO

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/ReactiveOperationExecutionWrapper.java
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  @Override
+  public Object execute(ExecutionContext<OperationModel> executionContext) throws Exception {
+    if (executionContext.getComponentModel().isBlocking()) {
+      return delegate.execute(executionContext);
+    }
+
+    ExecutionContextAdapter<OperationModel> context = (ExecutionContextAdapter<OperationModel>) executionContext;
+    return Mono.create(sink -> {
```
dfeist: type?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/ReactiveOperationExecutionWrapper.java
dfeist: In fact, why is this catch even needed?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/DefaultExecutionMediator.java
marianogonzalez: that will be reviewed in a separate PR now that mule-api has excepetions

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/DefaultExecutionMediator.java
marianogonzalez: that will be reviewed in a separate PR now that mule-api has excepetions

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/DefaultExecutionMediator.java
marianogonzalez: null doesn't mean that the even was dropped. Null means that the operation was void or that it simply returned null.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/DefaultExecutionMediator.java
marianogonzalez: mmm... good catch

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/ReactiveOperationExecutionWrapper.java
marianogonzalez: I did... for some reason that change is missing... Fuck github

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/ReactorCallback.java
marianogonzalez: no

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/ReactiveOperationExecutionWrapper.java
marianogonzalez: oh no.. now I remember.. That happens on reactorCallback

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/DefaultExecutionMediator.java
dfeist: If it's not safe to assume that, there is a lot more work to do here.. :-(

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/DefaultExecutionMediator.java
dfeist: `concatMap` will drop empty mono's, so you'll need to handle this in OperationMessageProcessor if you don't want that and return the original event instead.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
dfeist: Don't you need to return original event if there was no result from the operation rather than `returnDelegate.asReturnValue(null, operationContext))`?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
marianogonzalez: fixed

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
marianogonzalez: null doesn't mean to drop. Null means void operation or simply returned null

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
marianogonzalez: fixed, but this code is still ugly. Mono is missing something similar to Optiona.orElseGet(). In either case, the fallback return delegate gets executed even if Mono was not empty. I need to send a PR to Reactor for that

core/src/main/java/org/mule/runtime/core/api/retry/RetryPolicyTemplate.java
```diff
@@ -29,4 +34,11 @@
   void setNotifier(RetryNotifier retryNotifier);
 
   RetryContext execute(RetryCallback callback, Executor workManager) throws Exception;
+
+  default <T> Publisher<T> applyPolicy(Publisher<T> publisher,
```
marianogonzalez: this is following the same pattern as when blocking retry is performed. The user sees the template but the actual retry logic resides in the policy, and a new instance of that policy is created for each retryable task. This allows the user to have a single way of configuring reconnection, and then internally we decide when to do it in a blocking or non blocking fashion.

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/ReflectiveMethodOperationExecutorTestCase.java
```diff
@@ -185,6 +186,10 @@ private void assertResult(Object value, Object expected) throws Exception {
     assertThat(value, is(expected));
   }
 
+  private Object execute() {
+    return from(executor.execute(operationContext)).block();
```
marianogonzalez: no. this method doesn't fail. At most it returns an error mono

core/src/main/java/org/mule/runtime/core/api/retry/RetryPolicy.java
```diff
@@ -21,7 +21,10 @@
 
   PolicyStatus applyPolicy(Throwable cause);
 
-  void applyOn(Mono<?> publisher, Predicate<Throwable> predicate);
+  default <T> Publisher<T> applyPolicy(Publisher<T> publisher,
+                                       Predicate<Throwable> shouldRetry,
+                                       Consumer<Publisher<T>> onExhausted) {
```
marianogonzalez: yes, this was a copy paste problem. Fixed

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultExecutionMediator.java
```diff
@@ -89,13 +91,14 @@ public DefaultExecutionMediator(ExtensionModel extensionModel, OperationModel op
    * @throws Exception if the operation or a {@link Interceptor#before(ExecutionContext)} invokation fails
    */
   @Override
-  public Mono<Object> execute(OperationExecutor executor, ExecutionContextAdapter context) throws MuleException {
+  public Publisher<Object> execute(OperationExecutor executor, ExecutionContextAdapter context) throws MuleException {
```
marianogonzalez: There's no way I know or care about the actual type and I don't like raw types. The Object doesn't hurt at all

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultExecutionMediator.java
```diff
@@ -89,13 +91,14 @@ public DefaultExecutionMediator(ExtensionModel extensionModel, OperationModel op
...
     final List<Interceptor> interceptors = collectInterceptors(context.getConfiguration(), executor);
     final Optional<MutableConfigurationStats> stats = getMutableConfigurationStats(context);
     stats.ifPresent(s -> s.addInflightOperation());
 
     try {
-      return (Mono<Object>) getExecutionTemplate(context).execute(() -> executeWithInterceptors(executor, context, interceptors, stats));
+      return (Mono<Object>) getExecutionTemplate(context)
```
marianogonzalez: Yes. this PR is long enough I don't want to add changes on the execution template just for avoiding a cast.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultExecutionMediator.java
```diff
@@ -154,9 +168,9 @@ private InterceptorsExecutionResult before(ExecutionContext executionContext, Li
   private void onSuccess(ExecutionContext executionContext, Object result, List<Interceptor> interceptors) {
     intercept(interceptors,
               interceptor -> interceptor.onSuccess(executionContext, result), interceptor -> format(
```
marianogonzalez: AFAIK is better than concatenating. Also consider that the format will only be executed upon errors which will be rare at this point. I don't think it makes an actual differnece

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
```diff
@@ -185,20 +182,18 @@ public Event process(Event event) throws MuleException {
   protected Mono<Event> doProcess(Event event, ExecutionContextAdapter operationContext) throws MuleException {
     return executeOperation(operationContext)
         .map(value -> returnDelegate.asReturnValue(value, operationContext))
-        .defaultIfEmpty(returnDelegate.asReturnValue(null, operationContext));
+        .defaultIfEmpty(returnDelegate.asReturnValue(null, operationContext))
```
marianogonzalez: A void operation returns null. That's just how reflection works in Java. Additionally, an operation returning a String might end up returning null. In either case, I will get an empty Mono. I need this hack because empty monos skip the map operator.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/ExtensionProperties.java
```diff
@@ -32,6 +33,12 @@
   public static final String SOURCE_CALLBACK_CONTEXT_PARAM = PROPERTY_PREFIX + "SOURCE_CALLBACK_CONTEXT_PARAM";
 
   /**
+   * The key of an execution context variable which holds the {@link NonBlockingCallback} that a non blocking
+   * component will use to notify completion or failure
+   */
+  public static final String REACTIVE_CALLBACK_CONTEXT_PARAM = PROPERTY_PREFIX + "REACTIVE_CALLBACK_CONTEXT_PARAM";
```
dfeist: Use same name as type: NON_BLOCKING_CALLACK.., or OPERATION_COMPLETION_HANDLER..., rather than use a new term.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/ReactorCallback.java
```diff
@@ -0,0 +1,44 @@
...
+    this.sink = sink;
+    this.event = event;
+  }
+
+  @Override
+  public void success(Result<Object, Attributes> result) {
```
dfeist: How about the success with no response case, does this also use this signature?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultExecutionMediator.java
```diff
@@ -84,71 +86,69 @@ public DefaultExecutionMediator(ExtensionModel extensionModel, OperationModel op
...
-    }
 
-    return new OperationExecutionResult(result, exception, Optional.ofNullable(retryRequestHolder.get()));
-  }
+      result.doOnSuccess(value -> {
+        onSuccess(context, value, interceptors);
+        stats.ifPresent(s -> s.discountInflightOperation());
+        sink.success(value);
+      })
+          .mapError(e -> {
```
dfeist: Formatting?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultExecutionMediator.java
```diff
@@ -84,71 +86,69 @@ public DefaultExecutionMediator(ExtensionModel extensionModel, OperationModel op
...
+            e = exceptionEnricherManager.processException(e);
+            e = moduleExceptionHandler.processException(e);
+            e = onError(context, e, interceptors);
+
+            return e;
+          }).subscribe(value -> {
```
dfeist: Some comment(s) to explains what you are doing here?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultExecutionMediator.java
```diff
@@ -84,71 +86,69 @@ public DefaultExecutionMediator(ExtensionModel extensionModel, OperationModel op
...
+    }).doOnTerminate((value, e) -> {
+      try {
+        after(context, value, executedInterceptors);
+      } finally {
+        executedInterceptors.clear();
+      }
+    });
 
+    return from(getRetryPolicyTemplate(context.getConfiguration()).applyPolicy(publisher,
+                                                                               e -> ExceptionUtils.extractConnectionException(e)
```
dfeist: static import?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ExecutionMediator.java
```diff
@@ -25,9 +31,9 @@
...
    * @param executor an {@link OperationExecutor}
-   * @param context an {@link ExecutionContextAdapter}
-   * @return the operation's result
-   * @throws Exception if any exception is encountered
+   * @param context  an {@link ExecutionContextAdapter}
+   * @return a {@link Mono} with the operation's result
+   * @throws MuleException if the {@link Mono} could not be created
    */
-  Object execute(OperationExecutor executor, ExecutionContextAdapter context) throws Throwable;
+  Publisher<Object> execute(OperationExecutor executor, ExecutionContextAdapter context) throws MuleException;
```
dfeist: Is this the same method I said shouldn't throw an exception before, or a different one?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/InterceptingExecutionMediator.java
```diff
@@ -39,8 +46,8 @@ public InterceptingExecutionMediator(ExecutionMediator intercepted) {
   }
 
   @Override
-  public Object execute(OperationExecutor executor, ExecutionContextAdapter context) throws Throwable {
-    Object resultValue = intercepted.execute(executor, context);
+  public Publisher<Object> execute(OperationExecutor executor, ExecutionContextAdapter context) throws MuleException {
+    Object resultValue = from(intercepted.execute(executor, context)).block();
```
dfeist: Why block? Should this have a TODO?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
dfeist: Seems this condition is in two places now though, unless github is lying..

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
dfeist: If this statement returns Mono.emty() then the current event gets dropped by concatMap.  I fail to see how that it's correct.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
```diff
@@ -120,55 +122,72 @@ public OperationMessageProcessor(ExtensionModel extensionModel,
...
-        throw wrapInMessagingException(event, e);
-      }
-      throw e;
-    } catch (Throwable e) {
-      throw wrapInMessagingException(event, e);
-    }
+  protected Mono<Event> doProcess(Event event, ExecutionContextAdapter operationContext) throws MuleException {
+    return executeOperation(operationContext)
+        .map(value -> returnDelegate.asReturnValue(value, operationContext))
+        .defaultIfEmpty(returnDelegate.asReturnValue(null, operationContext))
```
dfeist: Also here, I'm not sure why if the operation is void you create a null message,

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
```diff
@@ -120,55 +122,72 @@ public OperationMessageProcessor(ExtensionModel extensionModel,
...
+        .mapError(Exceptions::unwrap);
   }
 
-  private MessagingException wrapInMessagingException(Event event, Throwable e) {
-    return new MessagingException(createStaticMessage(e.getMessage()), event, e, this);
+  private Mono<Object> executeOperation(ExecutionContextAdapter operationContext) throws MuleException {
```
dfeist: This throws exeption too

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/PagedOperationMessageProcessor.java
```diff
@@ -42,10 +45,10 @@ public PagedOperationMessageProcessor(ExtensionModel extensionModel,
   }
 
   @Override
-  protected org.mule.runtime.api.message.MuleEvent doProcess(Event event, ExecutionContextAdapter operationContext)
+  protected Mono<Event> doProcess(Event event, ExecutionContextAdapter operationContext)
```
dfeist: This throws exeption too

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/connector/PetStoreRetryPolicyProviderConnectionTestCase.java
```diff
@@ -53,24 +51,22 @@ protected String getConfigFile() {
...
+    exception.expectCause(is(instanceOf(ConnectionException.class)));
     runFlow("fail-operation-with-connection-exception");
   }
 
   @Test
   public void retryPolicyExhaustedDueToInvalidConnectionAtValidateTime() throws Exception {
     exception.expect(MessagingException.class);
     exception.expectCause(is(instanceOf(ConnectionException.class)));
-    exception.expectMessage(is(CONNECTION_FAIL_DOT));
+    //exception.expectMessage(is(CONNECTION_FAIL_DOT));
```
marianogonzalez: fixed

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/ExtensionProperties.java
marianogonzalez: fixed

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/ReactorCallback.java
marianogonzalez: that,s the case of a void operation and you can always pass an empty result.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultExecutionMediator.java
marianogonzalez: this is how rodro's monster formats... Complain to him

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultExecutionMediator.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultExecutionMediator.java
marianogonzalez: done

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ExecutionMediator.java
marianogonzalez: fixed

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ExecutionMediator.java
```diff
@@ -25,9 +31,9 @@
    * Coordinates the execution of the {@code executor} using the given {@code context}
    *
    * @param executor an {@link OperationExecutor}
-   * @param context an {@link ExecutionContextAdapter}
-   * @return the operation's result
-   * @throws Exception if any exception is encountered
+   * @param context  an {@link ExecutionContextAdapter}
+   * @return a {@link Mono} with the operation's result
+   * @throws MuleException if the {@link Mono} could not be created
```
marianogonzalez: fixed

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/InterceptingExecutionMediator.java
marianogonzalez: no. no plans to support NB here

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
marianogonzalez: we discussed this last night

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
marianogonzalez: fixed

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/PagedOperationMessageProcessor.java
marianogonzalez: fixed

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/ReactorCallback.java
```diff
@@ -0,0 +1,44 @@
...
+
+import reactor.core.publisher.MonoSink;
+
+/**
+ * Implementation of {@link CompletionCallback} which works using
+ * project Reactor.
+ *
+ * @since 4.0
+ */
+final class ReactorCallback implements CompletionCallback<Object, Attributes> {
```
dfeist: Minor: ReactorCallback ReactorCompletionCallback

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/InterceptingExecutionMediator.java
```diff
@@ -39,8 +45,8 @@ public InterceptingExecutionMediator(ExecutionMediator intercepted) {
   }
 
   @Override
-  public Object execute(OperationExecutor executor, ExecutionContextAdapter context) throws Throwable {
-    Object resultValue = intercepted.execute(executor, context);
+  public Publisher<Object> execute(OperationExecutor executor, ExecutionContextAdapter context) {
+    Object resultValue = from(intercepted.execute(executor, context)).block();
```
dfeist: What happens with exceptions that might be thrown here by block?  They should be caught and return error Mono shouldn't they? (same in other use like this)

PR #3740, opened 2016-12-13
MULE-11185: Fix: sftpclient unable to validate duplicate files for relative paths involving ~ symbol.

fsgonz:


transports/sftp/src/test/java/org/mule/transport/sftp/SftpEndpointWithRelativePathTestCase.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+
+    @Test
+    public void writeFileWithRelativePath() throws Exception
+    {
+        MuleClient client = muleContext.getClient();
+        MuleMessage result = client.send("vm://in", "some data", null);
+
+        assertNotNull(result);
```
marianogonzalez: use hamcrest

transports/sftp/src/test/java/org/mule/transport/sftp/SftpEndpointWithRelativePathTestCase.java
```diff
@@ -0,0 +1,39 @@
...
+        assertNull(result.getExceptionPayload());
+        assertFalse(result.getPayload() instanceof NullPayload);
```
marianogonzalez: this is only testing that it didn't fail. It's not testing that it actually wrote the file in the correct path

transports/sftp/src/test/java/org/mule/transport/sftp/SftpEndpointWithRelativePathTestCase.java
```diff
@@ -30,10 +30,9 @@ protected String getConfigFile()
     public void writeFileWithRelativePath() throws Exception
     {
         MuleClient client = muleContext.getClient();
-        MuleMessage result = client.send("vm://in", "some data", null);
+        client.send("vm://in", FILE_CONTENT, null);
+        MuleMessage message = client.request("file://testdir/file.txt", RECEIVE_TIMEOUT);
```
marianogonzalez: make sure that the file is deleted after the test ends.

PR #3743, opened 2016-12-13
MULE-11147 Retry HTTP requests where connection has become stale sinc…

dfeist:
…e obtaining it from connection pool

### DIFF
--- a/modules/http/src/main/java/org/mule/module/http/internal/request/DefaultHttpRequester.java
+++ b/modules/http/src/main/java/org/mule/module/http/internal/request/DefaultHttpRequester.java
@@ -14,6 +14,11 @@ import org.mule.DefaultMuleEvent;
 import org.mule.OptimizedRequestContext;
 import org.mule.RequestContext;
 import org.mule.api.CompletionHandler;
+import static org.apache.commons.lang.StringUtils.containsIgnoreCase;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import org.mule.api.MessagingException;
 import org.mule.api.MuleContext;
 import org.mule.api.MuleEvent;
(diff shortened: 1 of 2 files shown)


modules/http/src/main/java/org/mule/module/http/internal/request/DefaultHttpRequester.java
```diff
@@ -328,16 +372,23 @@ private MuleEvent innerProcess(MuleEvent muleEvent, boolean checkRetry) throws M
...
         {
             checkIfRemotelyClosed(e);
-            throw new MessagingException(CoreMessages.createStaticMessage(getErrorMessage(httpRequest)), muleEvent, e, this);
+            if(shouldRetryRemotelyClosed(e, retryCount))
+            {
+                return innerProcess(muleEvent, retryCount - 1);
+            }
+            else
+            {
+                throw new MessagingException(CoreMessages.createStaticMessage("Error sending HTTP request"), muleEvent, e);
```
afelisatti: Why change the error message? The above shows the failed URL.

modules/http/src/main/java/org/mule/module/http/internal/request/DefaultHttpRequester.java
```diff
@@ -214,83 +222,120 @@ private void initializeAttributeEvaluators(AttributeEvaluator ... attributeEvalu
...
+                checkIfRemotelyClosed(exception);
+                if(shouldRetryRemotelyClosed(exception, retryCount))
+                {
+                    try
+                    {
+                        innerProcessNonBlocking(muleEvent, originalCompletionHandler, retryCount - 1);
+                    }
+                    catch (MuleException e)
+                    {
+                        originalCompletionHandler.onFailure(new MessagingException(CoreMessages.createStaticMessage("Error sending HTTP request"), muleEvent, e));
```
afelisatti: Same comment as before, getErrorMessage will show the request that failed which was an ER to improve troubleshooting.

PR #3752, opened 2016-12-15
MULE-11203: Improving Error Message in AsynchronousRetryTemplate.

facundovs:
Referencing to SE-1963 issue, in DB Connections, non-blocking retry policies aren't supported.
However, the message is unclear and has to be modified.


marianogonzalez:
NO! It's the RetryConnectionFactory the one which has to validate that it's not supplied with an Async template. This PR is invalid

PR #3754, opened 2016-12-15
MULE-11203:  Add Error Message about not supported Asynchronous Retry Policies in DB Connection.

facundovs:
Referencing to SE-1963 issue, in DB Connections, non-blocking retry policies aren't supported.
Because of that, a verification was added in the RetryConnectionFactory class.
Now if the Retry Policy is not blocking, an exception is throwed.


marianogonzalez:
code is fine but issue title and commit message are wrong

marianogonzalez:
missing test case

PR #3757, opened 2016-12-15
MULE-10781: Ensure test parity between the old an new Web Service Consumer

juandesi:


extensions/db/src/main/java/org/mule/extension/db/internal/result/resultset/NullResultSetHandler.java
```diff
@@ -17,7 +17,7 @@
  * <p/>
  * <b>Warning!</b>
  * <p/>
- * The consumer of the returned {@link ResultSet} will be responsible of closing it.
```
juandesi: buen rename

tests/wsc/src/test/java/org/mule/extension/ws/runtime/wss/WssVerifySignatureTestCase.java
```diff
@@ -0,0 +1,21 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.extension.ws.runtime.wss;
+
+import ru.yandex.qatools.allure.annotations.Features;
+import ru.yandex.qatools.allure.annotations.Stories;
+
+@Features("Web Service Consumer")
+@Stories("WSS")
+public class WssVerifySignatureTestCase extends AbstractWebServiceSecurityTestCase {
```
marianogonzalez: sounds like the last 5 tests should be collapsed into one parameterized one

PR #3759, opened 2016-12-16
MULE-10568 Support non-blocking API in Ext-API sources

dfeist:
REVIEW ONLY.

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/processor/Sink.java
+++ b/core/src/main/java/org/mule/runtime/core/api/processor/Sink.java
@@ -1,0 +1,22 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.api.processor;
+
+import org.mule.runtime.core.api.Event;
+
+/**
+ * Sink
+ *
+ * @since 4.0
+ */
+public interface Sink {
+
+  void accept(Event event);
+
+  void complete();
+
+}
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhaseTemplate.java
```diff
@@ -39,6 +41,8 @@
    */
   Event routeEvent(Event event) throws MuleException;
 
+  Publisher<Event> dispatchEvent(Event event);
```
marianogonzalez: should the introduction of this method call for deprecation of  Event routeEvent(Event) ?

core/src/main/java/org/mule/runtime/core/api/processor/Sink.java
```diff
@@ -0,0 +1,22 @@
...
+package org.mule.runtime.core.api.processor;
+
+import org.mule.runtime.core.api.Event;
+
+/**
+ * Sink
+ *
+ * @since 4.0
+ */
+public interface Sink {
```
marianogonzalez: This should have a different name

core/src/main/java/org/mule/runtime/core/api/processor/strategy/ProcessingStrategy.java
```diff
@@ -60,6 +75,42 @@
...
+    Cancellation cancellation = from(publisher)
+        .onErrorResumeWith(MessagingException.class, flowConstruct.getExceptionListener())
+        .doOnNext(response -> response.getContext().success(response))
+        // We no longer need error here, so supress error to avoid retry.
+        .doOnError(MessagingException.class, me -> me.getEvent().getContext().error(me))
+        .onErrorResumeWith(EventDroppedException.class, ede -> {
+          ede.getEvent().getContext().success();
+          return empty();
+        })
+        .retry()
```
marianogonzalez: this hsould use the new RetryPolicyTemplate method I added

PR #3763, opened 2016-12-16
MULE-11194: Database UX cleanup

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/introspection/enricher/ContentParameterModelEnricher.java
```diff
@@ -85,8 +85,10 @@ private void configureDsl(ParameterDeclaration p) {
   }
 
   private void defaultToPayload(ParameterDeclaration p) {
-    p.setRequired(false);
-    p.setDefaultValue(PAYLOAD);
+    if (p.getDefaultValue() == null) {
+      p.setRequired(false);
```
alegmarra: is this because the `required` is going to be handled by the @Optional enricher?

marianogonzalez: this is because if the user didn't give a default, then we force it as optional:defaultPayload... This was already like this before the change

PR #3765, opened 2016-12-16
MULE-11199: WSC does not support operation level style definition, failing with NPE

juandesi:


extensions/wsc/src/main/java/org/mule/extension/ws/internal/introspection/WsdlIntrospecter.java
```diff
@@ -164,19 +167,32 @@ public boolean isDocumentStyle() {
     return isWsdlStyle(DOCUMENT_STYLE);
   }
 
+  /**
```
marianogonzalez: run formatter

PR #3767, opened 2016-12-17
MULE-10954: Email Attachments are added with empty content when setti…

pablomusumeci:
…ng the content with an expresion

### DIFF
--- a/extensions/email/src/main/java/org/mule/extension/email/internal/MessageBuilder.java
+++ b/extensions/email/src/main/java/org/mule/extension/email/internal/MessageBuilder.java
@@ -12,10 +12,11 @@ import static javax.mail.Message.RecipientType.CC;
 import static javax.mail.Message.RecipientType.TO;
 import static javax.mail.Part.ATTACHMENT;
 import static javax.mail.Part.INLINE;
+import static org.mule.extension.email.internal.util.EmailConnectorConstants.CONTENT_TRANSFER_ENCODING;
 import static org.mule.extension.email.internal.util.EmailConnectorConstants.CONTENT_TYPE_HEADER;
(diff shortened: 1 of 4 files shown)


extensions/email/src/main/java/org/mule/extension/email/internal/MessageBuilder.java
```diff
@@ -180,15 +182,18 @@ public MessageBuilder withHeaders(Map<String, String> headers) throws MessagingE
...
-        DataHandler dataHandler = toDataHandler(a.getId(), a.getContent(), a.getContentType());
-        attachmentsMap.put(a.getId(), dataHandler);
-      } catch (Exception e) {
-        throw new EmailException(ERROR + " could not add attachments", e);
-      }
-    });
+    Map<String, DataHandler> attachmentsMap = new LinkedHashMap<>();
+    attachments
+        .stream().map(a -> a.getContent() instanceof InputStream
+            ? new EmailAttachment(a.getId(), toByteArray((InputStream) a.getContent()), a.getContentType()) : a)
```
marianogonzalez: why? This is likely to run the server out of memory.

extensions/email/src/main/java/org/mule/extension/email/internal/MessageBuilder.java
```diff
@@ -180,15 +182,18 @@ public MessageBuilder withHeaders(Map<String, String> headers) throws MessagingE
...
+        .forEach(a -> {
+          try {
+            DataHandler dataHandler = toDataHandler(a.getId(), a.getContent(), a.getContentType());
```
marianogonzalez: why can't you just build the data handler based on the input stream?

PR #3774, opened 2016-12-20
MULE-10929:  Support policy packaging and deployment

pablokraan:
_ Adding PolicyParametrization to parameterize policy templates
_ Adding PolicyPointcut to provide a way to decide whether a policy must be applied on given a request
_ Adding RegionOwnerArtifact to provide access to the region classloader where policies are included
_ Adding TestPolicyManager to simulate API gateway
_ Adding ApplicationPolicyManager to manage policies per application.
_Adding a applicationPolicyManager to applications
_ Adding PolicyTemplateDescriptor and PolicyTemplate (and related factories and builders) to represent policy template artifacts
_ Refactoring AbstractArtifactClassLoaderBuilder to receive a PluginDependenciesResolver instead of the values needed to create it
_ Refactoring AbstractArtifactClassLoaderBuilder to let subclases to define how to createArtifactClassLoader
_ Adding ApplicationPolicyProvider to manage and provide application policies
_ Adding PolicyInstance (and related factories/builders) to represent instances of polices applied to applications
_ Adding PolicyFileBuilder to create policy file for testing purposes
_ Adding deployment test with policies

### DIFF
--- a/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultPolicyInstance.java
+++ b/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultPolicyInstance.java
@@ -1,0 +1,157 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+
+package org.mule.runtime.module.deployment.impl.internal.policy;
+
+import static java.util.Arrays.asList;
+import static java.util.Collections.emptyList;
+import static org.mule.runtime.core.config.bootstrap.ArtifactType.APP;
+import static org.mule.runtime.module.deployment.impl.internal.artifact.ArtifactContextBuilder.newBuilder;
+import org.mule.runtime.api.exception.MuleException;
+import org.mule.runtime.core.api.config.MuleProperties;
+import org.mule.runtime.core.api.registry.RegistrationException;
+import org.mule.runtime.core.policy.Policy;
+import org.mule.runtime.core.policy.PolicyParametrization;
+import org.mule.runtime.core.policy.PolicyPointcut;
+import org.mule.runtime.core.policy.PolicyPointcutParameters;
+import org.mule.runtime.deployment.model.api.application.Application;
+import org.mule.runtime.deployment.model.api.artifact.ArtifactContext;
+import org.mule.runtime.deployment.model.api.policy.PolicyTemplate;
+import org.mule.runtime.module.artifact.classloader.ClassLoaderRepository;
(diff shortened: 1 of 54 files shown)


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultPolicyInstance.java
```diff
@@ -0,0 +1,157 @@
...
+ */
+public class DefaultPolicyInstance implements PolicyInstance {
+
+  private ArtifactContext policyContext;
+  private final Application application;
+  private final PolicyTemplate template;
+  private final PolicyParametrization parametrization;
+  private final ServiceRepository serviceRepository;
+  private final ClassLoaderRepository classLoaderRepository;
+  private org.mule.runtime.core.policy.PolicyInstance policyInstance;
```
marianogonzalez: this is weird... is this class a wrapper?

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultPolicyInstance.java
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  private void initPolicyInstance() {
+
+    if (policyInstance == null) {
+      synchronized (this) {
+        initPolicyContext();
+
+        try {
+          policyInstance = policyContext.getMuleContext().getRegistry().lookupObject(
```
marianogonzalez: definitively weird to have this in this class. Sounds like a third party should be injected with this default policy instance and then wrap it in this other class

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultPolicyInstance.java
```diff
@@ -0,0 +1,157 @@
...
+                                                                                     org.mule.runtime.core.policy.DefaultPolicyInstance.class);
+        } catch (RegistrationException e) {
+          throw new IllegalStateException(String.format("More than one %s found on context", PolicyInstance.class), e);
+        }
+
+        try {
+          policyInstance.initialise();
+          policyInstance.start();
```
marianogonzalez: how do you know you should start it if this class is not lifecycle aware? If this is coming from the registry, then mule should be in charge of applying lifecycle

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultPolicyInstance.java
pablokraan: Yeah... I know is weird. 
This is class not a wrapper of org.mule.runtime.core.policy.PolicyInstance. There are duplicated concepts and at this point I don't know which one has to be renamed. I think that the PolicyInstance from core should not be there or should be renamed, but I'm not sure at this point.

PR #3776, opened 2016-12-21
MULE-11176: Default Scheduler is running tasks with the wrong

elrodro83:
classloader

### DIFF
--- a/services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultScheduler.java
+++ b/services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultScheduler.java
@@ -319,7 +319,7 @@ class DefaultScheduler extends AbstractExecutorService implements Scheduler {
     return newDecoratedTaskFor(super.newTaskFor(runnable, value), runnable.getClass());
   }
 
-  protected <T> RunnableFuture<T> newDecoratedTaskFor(final RunnableFuture<T> newTaskFor, final Class<?> taskClass) {
+  private <T> RunnableFuture<T> newDecoratedTaskFor(final RunnableFuture<T> newTaskFor, final Class<?> taskClass) {
     return new RunnableFutureDecorator<>(newTaskFor, currentThread().getContextClassLoader(), this, taskClass.getName(),
                                          idGenerator.getAndIncrement());
   }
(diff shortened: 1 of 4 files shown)


elrodro83:
* Also fixes a ConcurrentModificationException when stopping with custom schedulers

services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultScheduler.java
```diff
@@ -311,13 +311,16 @@ public void stop(long gracefulShutdownTimeout, TimeUnit unit) {
...
+    return newDecoratedTaskFor(super.newTaskFor(callable), callable.getClass());
   }
 
   @Override
   protected <T> RunnableFuture<T> newTaskFor(Runnable runnable, T value) {
-    return new RunnableFutureDecorator<>(super.newTaskFor(runnable, value), this, runnable.getClass().getName(),
+    return newDecoratedTaskFor(super.newTaskFor(runnable, value), runnable.getClass());
+  }
+
+  protected <T> RunnableFuture<T> newDecoratedTaskFor(final RunnableFuture<T> newTaskFor, final Class<?> taskClass) {
```
pablokraan: Do this have to be protected, if yes add javadoc

services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultSchedulerService.java
```diff
@@ -305,7 +309,7 @@ public void stop() throws MuleException {
       waitForExecutorTermination(startMillis, cpuLightExecutor, CPU_LIGHT_THREADS_NAME);
       waitForExecutorTermination(startMillis, ioExecutor, IO_THREADS_NAME);
       waitForExecutorTermination(startMillis, computationExecutor, COMPUTATION_THREADS_NAME);
-      for (ExecutorService customSchedulerExecutor : customSchedulersExecutors) {
+      for (ExecutorService customSchedulerExecutor : new ArrayList<>(customSchedulersExecutors)) {
```
pablokraan: Why do you need to create a list here?

services/scheduler/src/main/java/org/mule/service/scheduler/internal/RunnableFutureDecorator.java
```diff
@@ -36,23 +37,33 @@
...
     long startTime = beforeRun();
+
+    final Thread currentThread = Thread.currentThread();
+    final ClassLoader currentClassLoader = currentThread.getContextClassLoader();
+    currentThread.setContextClassLoader(classLoader);
+
     try {
       task.run();
     } finally {
+      currentThread.setContextClassLoader(currentClassLoader);
```
pablokraan: I think this should be the last thing to be done on the task, because other things can be require the same context classloader. For example, on line 68 there is a logger, if I'm not wrong the context use dto log depends on the context classloader

PR #3789, opened 2016-12-22
MULE-11265: ArtifactClassLoaderRunner - Cannot handle classpath URLs …

gsfernandes:
…escaped on Windows


tests/runner/src/test/java/org/mule/test/runner/api/DefaultWorkspaceReaderTestCase.java
```diff
@@ -0,0 +1,94 @@
...
+    if (!bar.exists()) {
+      assumeThat(bar.mkdir(), is(true));
+    }
+  }
+
+  @Test
+  public void handleEncodedURLs() throws Exception {
+    List<URL> urls = newArrayList();
+    File targetClasses = new File(new File(bar, TARGET), CLASSES);
+    assumeThat(targetClasses.mkdirs(), is(true));
```
pablokraan: This must be an assert not an assume

pablokraan: extract both lines (new file and assertion) into a utility method

PR #3793, opened 2016-12-23
MULE-11104 fixing build after maven-extensions-maven-plugin's revert

fernandezlautaro:


extensions/all-extensions/pom.xml
```diff
@@ -24,7 +24,6 @@
             <artifactId>mule-module-email</artifactId>
             <version>${project.version}</version>
             <type>zip</type>
-            <classifier>mule-plugin</classifier>
```
marianogonzalez: don't delete these. Comment and add a todo

PR #3798, opened 2016-12-23
MULE-10996: Content-Length header case conflicts with streaming.

anosenzo:
* Fix in the request builder to append the values of the headers with same name but different case.
* Added test cases to validate that the headers with same name but different case are appended fine in HttpRequestHeadersTestCase and HttpListenerResponseBuilderTestCase
* Added log4j2-test.xml file to add debug logging when needed in the HTTP tests.

### DIFF
--- a/modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerResponseBuilderTestCase.java
+++ b/modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerResponseBuilderTestCase.java
@@ -74,6 +74,9 @@ public class HttpListenerResponseBuilderTestCase extends FunctionalTestCase
     public SystemProperty headerDuplicatesResponseBuilderPath = new SystemProperty("headerDuplicatesResponseBuilderPath","headerDuplicatesResponseBuilderPath");
 
     @Rule
+    public SystemProperty headerDuplicatesAndDifferentCaseResponseBuilderPath = new SystemProperty("headerDuplicatesAndDifferentCaseResponseBuilderPath","headerDuplicatesAndDifferentCaseResponseBuilderPath");
+
+    @Rule
     public SystemProperty errorEmptyResponseBuilderPath = new SystemProperty("emptyResponseBuilderPath","emptyResponseBuilderPath");
 
     @Rule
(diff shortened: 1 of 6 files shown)


modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerResponseBuilderTestCase.java
```diff
@@ -86,7 +89,10 @@
     public SystemProperty errorHeadersResponseBuilderPath = new SystemProperty("headersResponseBuilderPath","headersResponseBuilderPath");
 
     @Rule
-    public SystemProperty errorHeaderDuplicatesResponseBuilderPath = new SystemProperty("headerDuplicatesResponseBuilderPath","headerDuplicatesResponseBuilderPath");
+    public SystemProperty errorHeaderDuplicatesResponseBuilderPath = new SystemProperty("errorHeaderDuplicatesResponseBuilderPath","errorHeaderDuplicatesResponseBuilderPath");
```
anosenzo: This line wasn't part of the fix but had wrong the path name, so it was testing the same as the "headerDuplicatesResponseBuilderPath" test.

modules/http/src/test/java/org/mule/module/http/functional/requester/HttpRequestHeadersTestCase.java
```diff
@@ -99,11 +99,33 @@ public void overridesHeaders() throws Exception
         flow.process(event);
 
         final Collection<String> values = headers.get("testName1");
-        assertThat(values, Matchers.containsInAnyOrder(Arrays.asList("testValue1", "testValueNew").toArray(new String[2])));
+        assertThat(values, Matchers.containsInAnyOrder(Arrays.asList("testValue1", "testValueNew", "testValue2").toArray()));
```
afelisatti: static import

modules/http/src/test/java/org/mule/module/http/functional/requester/HttpRequestHeadersTestCase.java
```diff
@@ -99,11 +99,33 @@ public void overridesHeaders() throws Exception
...
+        params.put("testName1", "testValueNewHeadersExpression1");
+        params.put("TestName1", "testValueNewHeadersExpression2");
+        params.put("testName2", "testValue2NewHeadersExpression");
+
+        event.getMessage().setInvocationProperty("headers", params);
+
+        flow.process(event);
+
+        final Collection<String> values = headers.get("testName1");
+        assertThat(values, Matchers.containsInAnyOrder(Arrays.asList("testValueNewHeadersExpression1", "testValueNewHeadersExpression2","testValueConfigHeader1", "testValueConfigHeader2", "testValueOutboundPropertyInConfigXml").toArray()));
```
afelisatti: static imports

PR #3800, opened 2016-12-23
MULE-7211: Fix Flaky Tests in Mule - SFTP Reconnection

afelisatti:


transports/sftp/src/test/java/org/mule/transport/sftp/SftpReconnectionTestCase.java
```diff
@@ -6,25 +6,24 @@
...
+import org.mule.context.notification.ConnectionNotification;
+import org.mule.tck.listener.ConnectionListener;
+import org.mule.tck.listener.FlowExecutionListener;
 
 public class SftpReconnectionTestCase extends AbstractSftpTestCase
 {
 
-    private static final long RECONNECTION_FREQUENCY = 1000l;
+    private static final long DELTA = 20l;
+    private static final long RECONNECTION_FREQUENCY = 1000l - DELTA;
```
elrodro83: there is a hamcrest matcher for this, instead of modifying the value to verify (IsCloseTo)

PR #3804, opened 2016-12-26
MULE-11266: Fix: Text file object store does not update persistent file stored record according to expiration policy.

fsgonz:


core/src/main/java/org/mule/util/store/InMemoryObjectStore.java
```diff
@@ -139,44 +139,75 @@ public void clear() throws ObjectStoreException
...
         {
-            final long now = System.nanoTime();
-            int expiredEntries = 0;
-            Map.Entry<?, ?> oldestEntry;
+            Long oldestKey = (Long) oldestEntry.getKey();
+            long oldestKeyValue = oldestKey.longValue();
 
-            purge:
-            while ((oldestEntry = store.firstEntry()) != null)
+            if (TimeUnit.NANOSECONDS.toMillis(now - oldestKeyValue) >= entryTTL)
```
pablolagreca: static import

core/src/main/java/org/mule/util/store/InMemoryObjectStore.java
```diff
@@ -139,44 +139,75 @@ public void clear() throws ObjectStoreException
...
+        // should expire further if entry TTLs are enabled
+        return entryTTL > 0 && currentSize != 0;
+    }
+
+    @Override
+    public void expire()
+    {
+        int expiredEntries = doTrimAndExpire();
+
+        if (logger.isDebugEnabled() && expiredEntries != 0)
```
pablolagreca: why the expiredEntries != 0 condition?. Should log it anyway so if we put this class in debug mode we can actually know if it's not expiring entries.

PR #3812, opened 2016-12-26
MULE-11254: Policies must be sorted before begin applied

pablokraan:
_ Adding policy order to be able to sort applied policies
_ Replacing PolicyInstanceProvider with ApplicationPolicyInstance as there is no need to have a nested PolicyProvider
_ Refactoring tests

### DIFF
--- a/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/MuleApplicationPolicyProvider.java
+++ b/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/MuleApplicationPolicyProvider.java
@@ -7,17 +7,19 @@
 
 package org.mule.runtime.module.deployment.impl.internal.application;
 
+import static java.lang.Integer.compare;
 import static java.lang.String.format;
 import static java.util.Optional.of;
 import static org.mule.runtime.api.util.Preconditions.checkArgument;
 import org.mule.runtime.api.lifecycle.Disposable;
+import org.mule.runtime.core.policy.Policy;
 import org.mule.runtime.core.policy.PolicyParametrization;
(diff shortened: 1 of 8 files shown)


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/MuleApplicationPolicyProvider.java
```diff
@@ -108,15 +111,15 @@ public synchronized boolean removePolicy(String parametrizedPolicyId) {
...
+  public List<Policy> findSourceParameterizedPolicies(PolicyPointcutParameters policyPointcutParameters) {
+    List<Policy> policies = new ArrayList<>();
 
     if (!registeredPolicyInstanceProviders.isEmpty()) {
       for (RegisteredPolicyInstanceProvider registeredPolicyInstanceProvider : registeredPolicyInstanceProviders) {
-        if (registeredPolicyInstanceProvider.policyInstanceProvider.getPointcut().matches(policyPointcutParameters)) {
-          policies.addAll(registeredPolicyInstanceProvider.policyInstanceProvider
-              .findSourceParameterizedPolicies(policyPointcutParameters));
+        if (registeredPolicyInstanceProvider.applicationPolicyInstance.getPointcut().matches(policyPointcutParameters)) {
+          if (registeredPolicyInstanceProvider.applicationPolicyInstance.getSourcePolicy().isPresent()) {
```
marianogonzalez: the instance should be exposed through a getter instead of a field directly. This is hard to mock this way

pablokraan: This private to the class, no one will mock it

PR #3826, opened 2016-12-29
MULE-11293: Fix: Session property disappearing after dispatchEvent()

facundovs:
CallableEntryPointResolver class was modified.
SessionVariableAfterJavaComponentTestCase and his respective config was added.

### DIFF
--- a/core/src/main/java/org/mule/client/DefaultLocalMuleClient.java
+++ b/core/src/main/java/org/mule/client/DefaultLocalMuleClient.java
@@ -18,6 +18,8 @@ import org.mule.api.MuleEvent;
 import org.mule.api.MuleException;
 import org.mule.api.MuleMessage;
 import org.mule.api.MuleRuntimeException;
+import org.mule.api.MuleSession;
+import org.mule.RequestContext;
 import org.mule.api.client.LocalMuleClient;
 import org.mule.api.client.OperationOptions;
 import org.mule.api.connector.ConnectorOperationLocator;
@@ -42,6 +44,7 @@ import java.util.Map;
 
 public class DefaultLocalMuleClient implements LocalMuleClient
 {
+
     protected final MuleContext muleContext;
     private final EndpointCache endpointCache;
     private ConnectorOperationLocator connectorOperatorLocator;
@@ -91,13 +94,13 @@ public class DefaultLocalMuleClient implements LocalMuleClient
     }
 
     public void dispatch(String url, Object payload, Map<String, Object> messageProperties)
-        throws MuleException
+            throws MuleException
     {
         dispatch(url, new DefaultMuleMessage(payload, messageProperties, muleContext));
     }
 
     public MuleMessage send(String url, Object payload, Map<String, Object> messageProperties)
-        throws MuleException
+            throws MuleException
     {
         return send(url, new DefaultMuleMessage(payload, messageProperties, muleContext));
     }
@@ -112,7 +115,14 @@ public class DefaultLocalMuleClient implements LocalMuleClient
         else
         {
             OutboundEndpoint endpoint = endpointCache.getOutboundEndpoint(url, MessageExchangePattern.REQUEST_RESPONSE, null);
-            return returnMessage(endpoint.process(createMuleEvent(message, endpoint)));
+            if (RequestContext.getEvent() != null)
+            {
+                return returnMessage(endpoint.process(createMuleEvent(message, endpoint, RequestContext.getEvent().getSession())));
+            }
+            else
+            {
+                return returnMessage(endpoint.process(createMuleEvent(message, endpoint)));
+            }
         }
     }
 
@@ -131,7 +141,7 @@ public class DefaultLocalMuleClient implements LocalMuleClient
     }
 
     public MuleMessage send(String url, Object payload, Map<String, Object> messageProperties, long timeout)
-        throws MuleException
+            throws MuleException
     {
         return send(url, new DefaultMuleMessage(payload, messageProperties, muleContext), timeout);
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mule/model/resolvers/CallableEntryPointResolver.java
```diff
@@ -43,6 +48,7 @@ public InvocationResult invoke(Object component, MuleEventContext context) throw
     {
         if (component instanceof Callable)
         {
+            copySessionProperties(context.getSession(), context.getMessage());
```
pablokraan: If properties need to be copied, that should happen inside the MuleEventContext instance creation, not here

tests/integration/src/test/java/org/mule/transformers/simple/SessionVariableAfterJavaComponentTestCase.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Rule
+    public SystemProperty sessionVariableSystemProperty = new SystemProperty("sessionVariable", SESSION_VARIABLE_VALUE);
+
+    @Override
+    protected String getConfigFile()
+    {
+        return "session-variable-after-java-component.xml";
+    }
+
```
pablokraan: Remove extra line

tests/integration/src/test/java/org/mule/transformers/simple/SessionVariableAfterJavaComponentTestCase.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Test
+    public void testSessionVariableExistsAfterJavaComponent() throws Exception
+    {
+        runFlow("main");
+        latch.await(RECEIVE_TIMEOUT, MILLISECONDS);
+        assertThat(DispatchJavaComponent.sessionVariableValue, is(SESSION_VARIABLE_VALUE));
+        latch.await(RECEIVE_TIMEOUT, MILLISECONDS);
```
pablokraan: Why is this second await needed? Latch was previously released, otherwise the first call would fail

tests/integration/src/test/java/org/mule/transformers/simple/SessionVariableAfterJavaComponentTestCase.java
```diff
@@ -0,0 +1,78 @@
...
+    {
+
+        private static String sessionVariableValue = null;
+
+        @Override
+        public Object onCall(MuleEventContext eventContext) throws Exception
+        {
+            sessionVariableValue = eventContext.getMessage().getProperty("testVariable", SESSION);
+            eventContext.getMuleContext().getClient().dispatch("vm://in", eventContext.getMessage());
+            latch.release();
```
pablokraan: Why are you calling latch.release twice?

tests/integration/src/test/java/org/mule/transformers/simple/SessionVariableAfterJavaComponentTestCase.java
```diff
@@ -0,0 +1,72 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+
+public class SessionVariableAfterJavaComponentTestCase extends FunctionalTestCase
+{
+
+    private static final String SESSION_VARIABLE_VALUE = "TEST";
+
+    @Rule
+    public SystemProperty sessionVariableSystemProperty = new SystemProperty("sessionVariable", SESSION_VARIABLE_VALUE);
```
pablokraan: Seems to me that the property value can be fixed, why do you need a system property?

tests/integration/src/test/java/org/mule/transformers/simple/SessionVariableAfterJavaComponentTestCase.java
```diff
@@ -0,0 +1,72 @@
...
+    protected String getConfigFile()
+    {
+        return "session-variable-after-java-component.xml";
+    }
+
+    @Test
+    public void testSessionVariableExistsAfterJavaComponent() throws Exception
+    {
+        runFlow("main");
+        assertThat(DispatchJavaComponent.sessionVariableValue, is(SESSION_VARIABLE_VALUE));
```
pablokraan: Would be better to rename sessionVariableValue to inboundSessionVariableValue and outboundSessionVariableValue and use static imports or move them to the test class

tests/integration/src/test/resources/session-variable-after-java-component.xml
```diff
@@ -0,0 +1,16 @@
...
+<mule xmlns="http://www.mulesoft.org/schema/mule/core"
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xmlns:vm="http://www.mulesoft.org/schema/mule/vm"
+      xsi:schemaLocation="http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd
+        http://www.mulesoft.org/schema/mule/vm http://www.mulesoft.org/schema/mule/vm/current/mule-vm.xsd">
+    <flow name="dispatch-flow">
+        <vm:inbound-endpoint path="in"/>
+        <echo-component/>
+    </flow>
+    <flow name="main">
```
pablokraan: Minor, but having the entry flow as the first flow in the config makes the file more readable

tests/integration/src/test/java/org/mule/transformers/simple/SessionVariableAfterJavaComponentTestCase.java
```diff
@@ -0,0 +1,65 @@
...
+        assertThat(outboundSessionVariableValue, is(SESSION_VARIABLE_VALUE));
+    }
+
+    public static class DispatchJavaComponent implements Callable
+    {
+        @Override
+        public Object onCall(MuleEventContext eventContext) throws Exception
+        {
+            inboundSessionVariableValue = eventContext.getMessage().getProperty("testVariable", SESSION);
+            eventContext.getMuleContext().getClient().dispatch("vm://in", eventContext.getMessage());
```
pablokraan: Please, add a test using send instead of dispatch

dfeist:
@facundovs You need to test the `connectorMessageProcessor` case too, as i'm pretty sure this is going to be affected by the same issue.  

Regarding the fix, this is the correct place now I believe, but before using `RequestContext` can you instead try copying the messaging before creating the event instead, I think this might be a better approach if it works.

PR #3837, opened 2016-12-30
MULE-9771: Add non-blocking support for HTTP extension

afelisatti:


marianogonzalez:
no test case verifying that things actually go non blocking? Wasn't that tested in 3.x?

PR #3838, opened 2016-12-31
MULE-11237: Add language rules into the extension model

marianogonzalez:


modules/extensions-spring-support/src/test/resources/schemas/vegan.xsd
```diff
@@ -405,14 +425,24 @@
...
-                    <xs:element minOccurs="0" maxOccurs="1" ref="mule:abstract-reconnection-strategy"></xs:element>
                     <xs:element minOccurs="0" maxOccurs="1" ref="mule:abstract-redelivery-policy"></xs:element>
+                    <xs:element minOccurs="0" maxOccurs="1" ref="mule:abstract-reconnection-strategy"></xs:element>
                 </xs:sequence>
                 <xs:attribute type="mule:substitutableName" use="optional" name="config-ref">
                     <xs:annotation>
                         <xs:documentation>Specify which configuration to use for this invocation.</xs:documentation>
                     </xs:annotation>
                 </xs:attribute>
+                <xs:attribute type="xs:string" use="optional" name="redeliveryPolicy">
```
alegmarra: with this we enable the user to set the infrastructure parameters as bean refs, which couldn't do before.
the ref'ed mule elements like `reconnect` don't support top level element definitions (it's top substitution group is not `mule:abstract-extension`), so it's not possible to declare the reconnection as a bean.

We should either avoid creating this attributes, or allow the infrastructure parameters to be declared as top level elements

PR #3839, opened 2016-12-31
MULE-11238: Explicit parameter groups in DSL

alegmarra:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationParameterValueResolver.java
```diff
@@ -42,7 +42,7 @@ public Object getParameterValue(String containerName) {
 
   private Optional<ParameterGroupDescriptor> getParameterGroup(String parameterGroupName) {
     return operationModel.getParameterGroupModels().stream()
-        .filter(group -> group.getName().equals(parameterGroupName))
+        .filter(group -> group.getName().equals(parameterGroupName) && !group.isShowInline())
```
marianogonzalez: why this?

alegmarra: when resolving an inline group, we need to handle it as a top level parameter. This avoid the group to be handled as a container of plain attributes and instead fallback to finding the parameter from the `executionContext` using the container name

marianogonzalez: add a comment clarifying this

PR #3845, opened 2017-01-02
MULE-10969: Http UI Improvements

juandesi:


modules/http/src/main/java/org/mule/extension/http/internal/request/RequestConnectionParams.java
```diff
@@ -45,19 +46,11 @@
    */
   @Parameter
   @Optional
+  @Example("http://www.somehost.com")
```
afelisatti: Remove "http://"

modules/http/src/main/java/org/mule/extension/http/internal/request/RequestConnectionParams.java
```diff
@@ -67,6 +60,16 @@
...
 
   /**
+   * Port where the requests will be sent. If the protocol attribute is HTTP (default) then the default value is 80, if the
+   * protocol attribute is HTTPS then the default value is 443.
+   */
+  @Parameter
+  @Optional
+  @Example("80")
+  @Placement(order = 3)
+  private Function<Event, Integer> port;
```
afelisatti: Why move this? It was ordered as it was.

juandesi: I miss pressed something ¯\_(ツ)_/¯ ajajaj

PR #3849, opened 2017-01-02
MULE-11323: Fix: A more declarative exception message has to be added to unsupported methods in DefaultMessageCollection

fsgonz:


core/src/main/java/org/mule/DefaultMessageCollection.java
```diff
@@ -274,7 +274,9 @@ public int size()
         }
         else
         {
-            throw new UnsupportedOperationException("getPayloadAsBytes(), use getPayload(DataType.BYTE_ARRAY_DATA_TYPE)");
+            throw new UnsupportedOperationException(
+                    "getPayloadAsBytes(), There has been an attempt to directly access the payload of a message collection, which is unsupported. " + System.getProperty("line.separator") +
+                                                    "Please retrieve the value from messageList or use getPayload(DataType.BYTE_ARRAY_DATA_TYPE)");
```
pablokraan: What do you mean by "retrieve the value from messageList "?

fsgonz: I mean retrieving the value from the attribute MessageList from DefaultMessageCollection. Another option would be to flatten the message collection. The quid is that message collection consists in a collection of payloads, not  only one payload. The issue is that but after and scatter-and-gather you don't have getPayload supported in the resulting message so you have to add some kind of transformation if you want to get an unified payload and obtain it through getPayload.

I would otherwise change the getPayload method to be supported so that it returns a "flattened" payload from each item of the collection but till now it was supposed to not be supported by design. This latter option would make it unnecessary to add a transformation or a flatting after a scatter gather.

In [SE-3746](https://www.mulesoft.org/jira/browse/SE-3746) it was requested only a change in the exception message.

fsgonz: According to [SE-3746] (https://www.mulesoft.org/jira/browse/SE-3746) the new exception description was accepted.

PR #3852, opened 2017-01-03
MULE-11325: Change policy descriptor to use json instead of propertie…

pablokraan:
…s files

_ Applying changes to match new mule-api
_ Replacing policy’s properties descriptor with a son file


modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/builder/PolicyFileBuilder.java
```diff
@@ -109,24 +137,8 @@ public PolicyFileBuilder definedBy(String configFile) {
   public PolicyFileBuilder usingResource(String resourceFile, String targetFile) {
     checkImmutable();
     checkArgument(!isEmpty(resourceFile), "Resource file cannot be empty");
-    resources.add(new ZipUtils.ZipResource(resourceFile, "classes/" + targetFile));
+    resources.add(new ZipResource(resourceFile, "classes/" + targetFile));
```
fernandezlautaro: can't you use the FileSystemPolicyClassLoaderModelLoader#CLASSES_DIR?

PR #3857, opened 2017-01-03
MULE-11331: Statements leaked when auto key generation fails while ex…

marianogonzalez:
…ecuting stored procedure


modules/db/src/main/java/org/mule/module/db/internal/domain/executor/StoredProcedureExecutor.java
```diff
@@ -40,7 +40,25 @@ protected Object doExecuteQuery(DbConnection connection, Statement statement, Qu
     @Override
     protected Object doExecuteQuery(DbConnection connection, Statement statement, Query query, AutoGeneratedKeyStrategy autoGeneratedKeyStrategy) throws SQLException
     {
-        autoGeneratedKeyStrategy.execute(statement, query.getQueryTemplate());
+        try
+        {
+            autoGeneratedKeyStrategy.execute(statement, query.getQueryTemplate());
```
pablokraan: This try/catch must go inside the different AutoGeneratedKeyStrategy implementatations (maybe defining an abstract class to avoir repeating the code)

modules/db/src/test/java/org/mule/module/db/internal/domain/executor/StoredProcedureExecutorTestCase.java
```diff
@@ -0,0 +1,72 @@
...
+    @Before
+    public void before()
+    {
+        executor = new StoredProcedureExecutor(statementFactory, statementResultHandler);
+    }
+
+    @Test
+    public void statementClosedIfKeyGenerationFails() throws Exception
+    {
+        Statement statement = mock(Statement.class);
```
pablokraan: There are different methods in the AutoGeneratedKeyStrategy, all of them have the same issue, so a test for each method must be added

modules/db/src/main/java/org/mule/module/db/internal/domain/executor/StoredProcedureExecutor.java
anoordover: Is it an idea to implement this using a Decorator/Wrapper pattern around a AutoGeneratedKeyStrategy. You could instantiate this Decorator/Wrapper inside StoredProcedureMessageProcessor.

pablokraan: @anoordover pull requests are welcome

PR #3864, opened 2017-01-04
MULE-10747: Move CXF module to the compatibility module

elrodro83:


modules/cxf/src/main/resources/META-INF/mule-module.properties
```diff
@@ -1,14 +1,14 @@
 module.name=cxf
```
gsfernandes: This file, should be deleted? Now that CXF is no longer to be a module and instead is part of the compatibility plugin.

modules/cxf/src/test/resources/META-INF/mule-module.properties
```diff
@@ -1,14 +1,14 @@
 module.name=cxf-tests
```
gsfernandes: Same here, this file should be deleted.

PR #3866, opened 2017-01-05
MULE-11324: NullPointerException when applying more than one policy

pablokraan:
_ Separating MavenBundleDescriptorLoader from MavenClassLoaderModelLoader
_ Extracting maven utils from MavenClassLoaderModelLoader
_ Adding PropertiesBundleDescriptorLoader to use on artifacts without maven support
_ Updating MulePluginDescriptorGenerator to include bundleDescriptor loader
_ Adding deployment test with multiple policies


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/plugin/MavenClassLoaderModelLoader.java
```diff
@@ -135,7 +108,7 @@ public BundleDescriptor loadBundleDescriptor(File pluginFolder) {
    */
   public ClassLoaderModel loadClassLoaderModel(File pluginFolder,
                                                Map<String, Object> attributes) {
-    final Model model = getPomModel(pluginFolder);
+    final Model model = MavenUtils.getPomModel(pluginFolder);
```
fernandezlautaro: static import

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/plugin/ArtifactPluginDescriptorFactory.java
```diff
@@ -193,9 +195,16 @@ private ArtifactPluginDescriptor loadFromJsonDescriptor(File pluginFolder, File
...
-      descriptor.setBundleDescriptor(mavenClassLoaderModelLoader.loadBundleDescriptor(pluginFolder));
     });
 
+    MuleArtifactLoaderDescriptor bundleDescriptorLoader = mulePluginModel.getBundleDescriptorLoader();
+    if (!bundleDescriptorLoader.getId().equals(MAVEN)) {
+      throw new ArtifactDescriptorCreateException(format("The identifier '%s' for a bundle descriptor loader is not supported (error found while reading plugin '%s')",
+                                                         bundleDescriptorLoader.getId(),
+                                                         pluginFolder.getAbsolutePath()));
+    }
+    descriptor.setBundleDescriptor(new MavenBundleDescriptorLoader().loadBundleDescriptor(pluginFolder));
```
fernandezlautaro: despite not being used, shouldn't we pass the `bundleDescriptorLoader.getAttributes()` to the `#loadBundleDescriptor(..)` method?

Also, as there is two implementations of the "bundle loader", maybe an interface where it gets the File+attributes is needed, and we would implement SPI for the `MAVEN` ID and the `PROPERTIES_BUNDLE_DESCRIPTOR_LOADER_ID` ID (as it happens in PropertiesBundleDescriptorLoader)

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/PropertiesBundleDescriptorLoader.java
```diff
@@ -0,0 +1,46 @@
...
+   * @param attributes attributes defined in the loader.
+   * @return a locator of the coordinates of the current artifact
+   * @throws ArtifactDescriptorCreateException if any bundle descriptor required property is missing on the given attributes.
+   */
+  public BundleDescriptor loadBundleDescriptor(Map<String, Object> attributes) {
+    String version = (String) attributes.get(VERSION);
+    String groupId = (String) attributes.get(GROUP_ID);
+    String artifactId = (String) attributes.get(ARTIFACT_ID);
+    String classifier = (String) attributes.get(CLASSIFIER);
+    String type = (String) attributes.get(TYPE);
```
fernandezlautaro: all of these casts could cause a ClassCastException, as the one writing the JSON file might have put `'version': ['hi']`.

I checked the code we used for the `org.mule.runtime.extension.api.loader.ExtensionModelLoader`'s implementations, but it seems that has been deleted.
Don't you think the attributes should be validated? (maybe we could drop the Map<string,object> and put something that does a get()+cast, throwing a typed exception)

PR #3867, opened 2017-01-05
WIP: Spike: Interception API

gsfernandes:
DO NOT MERGE this PR.
@pablolagreca here is an spike for interception API to be reviewed and discussed.

@damiansima, please take a look to the API and the test to see how it has to be configured an interception callback.

Be aware that still not all the MPs are supported to be intercepted due to they were not migrated to new parsers. We have to have this discussion with @pablolagreca. In particular the one that I'm concern is flow-ref. 
In addition to that I would need to get feedback from @dfeist about the non-blocking interceptor part.
Feedback is welcome guys!

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/execution/interception/InterceptorMessageProcessor.java
+++ b/core/src/main/java/org/mule/runtime/core/execution/interception/InterceptorMessageProcessor.java
@@ -1,0 +1,207 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+
+package org.mule.runtime.core.execution.interception;
+
+import static java.lang.String.valueOf;
+import static java.util.Collections.emptyMap;
+import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
+import static org.mule.runtime.core.api.lifecycle.LifecycleUtils.disposeIfNeeded;
+import static org.mule.runtime.core.api.lifecycle.LifecycleUtils.initialiseIfNeeded;
+import static org.mule.runtime.core.api.lifecycle.LifecycleUtils.startIfNeeded;
+import static org.mule.runtime.core.api.lifecycle.LifecycleUtils.stopIfNeeded;
+import static org.mule.runtime.core.util.rx.Exceptions.checkedFunction;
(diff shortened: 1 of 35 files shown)


core/src/main/java/org/mule/runtime/core/execution/interception/InterceptorMessageProcessor.java
```diff
@@ -0,0 +1,204 @@
...
+        }
+      }
+      return intercepted.process(event);
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Publisher<Event> apply(Publisher<Event> publisher) {
```
dfeist: This will cause all intercepting processors (and the rest of the chain) to block. I assume we don't want this? To avoid this, implement apply, and delegate and block from process.

core/src/main/java/org/mule/runtime/core/execution/interception/InterceptorMessageProcessor.java
```diff
@@ -0,0 +1,207 @@
...
+    return intercepted.getProcessingType();
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Event process(Event event) throws MuleException {
+    final ComponentIdentifier componentIdentifier = (ComponentIdentifier) this.getAnnotations().get(ANNOTATION_NAME);
+    Map<String, Object> parameters = ((Map) this.getAnnotations().getOrDefault(ANNOTATION_PARAMETERS, emptyMap()));
```
marianogonzalez: is it correct to assume that this class is only intended to work with non SDK MPs? Otherwise, this won't interop well.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
```diff
@@ -91,7 +92,8 @@
  *
  * @since 3.7.0
  */
-public class OperationMessageProcessor extends ExtensionComponent implements Processor, EntityMetadataProvider, Lifecycle {
+public class OperationMessageProcessor extends ExtensionComponent implements Processor, EntityMetadataProvider, Lifecycle,
+    Interceptable {
```
marianogonzalez: You shouldn't treat this one the same way to treat the others. Interceptable should probably be more than just a marker interface and give room for different components to be intercepted in their own way.

gsfernandes: Agree

PR #3876, opened 2017-01-05
MULE-11144: Exclusive Parameters validations fails if the parameters …

pablomusumeci:
…uses an Alias


modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/connector/PetStoreExclusiveParameterWithAliasTestCase.java
```diff
@@ -0,0 +1,37 @@
...
+
+  @Override
+  protected String getConfigFile() {
+    return "petstore-exclusive-parameters-with-alias.xml";
+  }
+
+  @Test
+  public void getBreederOperation() throws Exception {
+    ExclusivePetBreeder petBreeder = (ExclusivePetBreeder) runFlow("getBreederOperation").getMessage().getPayload();
+    assertThat(petBreeder.getunaliasedNammals(), is(not(nullValue())));
```
marianogonzalez: why do you assert for not null?? It's better if you assert for the actual `Primate` value that you used in the XML

PR #3877, opened 2017-01-05
MULE-11309: Implement the Extensions Client

juandesi:


modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/ExtensionsClientTestCase.java
```diff
@@ -0,0 +1,257 @@
...
+    assertThat(result.getAttributes().isPresent(), is(true));
+    assertThat(result.getAttributes().get().getValue(), is(0));
+  }
+
+  @Test
+  @Description("Executes a void operation using the client")
+  public void executeVoidOperation() throws MuleException {
+    OperationParameters params = builder().configName(HEISENBERG_CONFIG).build();
+    Result<Object, Attributes> result = client.execute(HEISENBERG_EXT_NAME, "die", params);
+    assertThat(result.getOutput(), is(nullValue()));
```
juandesi: is this OK @marianogonzalez ? not 100% if this is the correct behavior that we want.

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/ExtensionsClientTestCase.java
```diff
@@ -0,0 +1,257 @@
...
+import ru.yandex.qatools.allure.annotations.Description;
+import ru.yandex.qatools.allure.annotations.Features;
+
+@Features("ExtensionsClient")
+public class ExtensionsClientTestCase extends ExtensionFunctionalTestCase {
+
+  @Rule
+  public ExpectedException exception = ExpectedException.none();
+
+  private static final String HEISENBERG_EXT_NAME = HEISENBERG;
```
marianogonzalez: this constant seems redundant

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/ExtensionsClientTestCase.java
marianogonzalez: for now yes. The alternative would be changing the value for an Optional but not sure if that's a good idea to introduce right now

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/DefaultExtensionsClient.java
```diff
@@ -0,0 +1,149 @@
...
+ * @since 4.0
+ */
+public final class DefaultExtensionsClient implements ExtensionsClient, Initialisable {
+
+  @Inject
+  private MuleContext muleContext;
+
+  private ExtensionManagerAdapter extensionManager;
+
+  private final DefaultPolicyManager policyManager = new DefaultPolicyManager();
```
marianogonzalez: don't. PolicyManager needs to be injected as well. You can't just go and create your own

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/DefaultExtensionsClient.java
```diff
@@ -0,0 +1,149 @@
...
+  private OperationMessageProcessor createProcessor(String extensionName, String operationName, OperationParameters parameters) {
+    ExtensionModel extension = findExtension(extensionName);
+    OperationModel operation = findOperation(extension, operationName);
+    ConfigurationProvider config = parameters.getConfigName().map(this::findConfiguration).orElse(null);
+    OperationMessageProcessorFactory factory = new OperationMessageProcessorFactory(extension, operation, config, policyManager,
+                                                                                    toResolvers(parameters), muleContext, "");
+    try {
+      OperationMessageProcessor processor = factory.getOperationMessageProcessor();
+      processor.doStart();
+      processor.doInitialise();
```
marianogonzalez: flipped

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/DefaultExtensionsClient.java
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  private ConfigurationProvider findConfiguration(String configName) {
+    return configurations.computeIfAbsent(configName, name -> extensionManager.getConfigurationProvider(configName)
+        .orElseThrow(() -> new MuleRuntimeException(createStaticMessage("No configuration [" + configName + "] found"))));
+  }
+
+  private OperationModel findOperation(ExtensionModel extensionModel, String operationName) {
+    return operations.computeIfAbsent(operationName, operation -> ImmutableList.<OperationModel>builder()
+        .addAll(extensionModel.getConfigurationModels().stream().flatMap(c -> c.getOperationModels().stream()).collect(toList()))
```
marianogonzalez: This is unneficient. This should use an IdempotentExtensionWalker and stop() when match found

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/DefaultExtensionsClient.java
```diff
@@ -0,0 +1,149 @@
...
+      throw new MuleRuntimeException(createStaticMessage("Could not create Operation Message Processor"), e);
+    }
+  }
+
+  private Map<String, Object> toResolvers(OperationParameters parameters) {
+    return parameters.getParameters().keySet().stream()
+        .collect(toMap(identity(), key -> new StaticValueResolver<>(parameters.getParameters().get(key))));
+  }
+
+  private ConfigurationProvider findConfiguration(String configName) {
```
marianogonzalez: This method is unnecessary. ExtensionManager.getConfigurationProvider() already holds and internal cache which is properly maintained/invalidated/etc. Delete this method and the configurations map

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/DefaultExtensionsClient.java
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  private OperationModel findOperation(ExtensionModel extensionModel, String operationName) {
+    return operations.computeIfAbsent(operationName, operation -> ImmutableList.<OperationModel>builder()
+        .addAll(extensionModel.getConfigurationModels().stream().flatMap(c -> c.getOperationModels().stream()).collect(toList()))
+        .addAll(extensionModel.getOperationModels()).build().stream().filter(o -> o.getName().equalsIgnoreCase(operation))
+        .findAny().orElseThrow(() -> new MuleRuntimeException(createStaticMessage("No Operation [" + operation + "] Found"))));
+  }
+
+  private ExtensionModel findExtension(String extensionName) {
```
marianogonzalez: same as before. ExtensionManager is already smart enoguh at handling this. Delete method and map

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/DefaultExtensionsClient.java
```diff
@@ -0,0 +1,149 @@
...
+        .collect(toMap(identity(), key -> new StaticValueResolver<>(parameters.getParameters().get(key))));
+  }
+
+  private ConfigurationProvider findConfiguration(String configName) {
+    return configurations.computeIfAbsent(configName, name -> extensionManager.getConfigurationProvider(configName)
+        .orElseThrow(() -> new MuleRuntimeException(createStaticMessage("No configuration [" + configName + "] found"))));
+  }
+
+  private OperationModel findOperation(ExtensionModel extensionModel, String operationName) {
+    return operations.computeIfAbsent(operationName, operation -> ImmutableList.<OperationModel>builder()
```
marianogonzalez: the key of the map cannot be the operationName since two different extensions can have operations with the same name. If you want to cache this, the key must be a Pair<ExtensionModel, OperationName>

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/DefaultExtensionsClient.java
```diff
@@ -0,0 +1,149 @@
...
+      processor.doStart();
+      processor.doInitialise();
+      return processor;
+    } catch (Exception e) {
+      throw new MuleRuntimeException(createStaticMessage("Could not create Operation Message Processor"), e);
+    }
+  }
+
+  private Map<String, Object> toResolvers(OperationParameters parameters) {
+    return parameters.getParameters().keySet().stream()
```
marianogonzalez: I don't understand this. I was expecting a ParameterResolverSetResolver to be used here. What was the point of all that refactor?

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/AbstractExtensionObjectFactory.java
```diff
@@ -78,199 +31,35 @@
     implements ObjectTypeParametersResolver {
 
   protected final MuleContext muleContext;
-  private Map<String, Object> parameters = new HashMap<>();
+  protected Map<String, Object> parameters = new HashMap<>();
+  protected ParametersResolverSetResolver parametersResolver;
```
juandesi: it resolves a ResolverSet of parameters :p

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/ExtensionsClientTestCase.java
juandesi: Is for tests readability.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/DefaultExtensionsClient.java
juandesi: All provided parameters don't need to be resolved, they should be passed as they are, the only way of doing this is wrapping them in a static value resolver, this way the ParameterResolver will not try to instantiate those.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/DefaultExtensionsClient.java
juandesi: gotcha

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/AbstractExtensionObjectFactory.java
marianogonzalez: It does quite a lot more than that. It also checks exclusive optionals, etc... My point was that it has the Resolver word twice and thus it's hard to read, pronounce and understand

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/DefaultExtensionsClient.java
marianogonzalez: Why do you use the doLifecycle() methods instead of the original ones. 

1) The initialise method already contains logic for locating the ConfigurationProvider. If you could leverage that, it's one less thing to do here
2) The initialise method also validates that the config is not an expression and other things that are worth reusing
3) The start() method makes sure that the doStart() method is executed with the correct classloader, which is necessary. By skipping that, you're missing that condition here and thus getting exposed to trouble.

I think a best effort should be made so that you use the initialise() and start() methods here

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
```diff
@@ -191,7 +191,7 @@ public Event process(Event event) throws MuleException {
   }
 
   @Override
-  protected void doInitialise() throws InitialisationException {
+  public void doInitialise() throws InitialisationException {
```
marianogonzalez: related to the prior comments, doInitialise() and doStart() should remain protected

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/ExtensionsClientTestCase.java
```diff
@@ -0,0 +1,265 @@
...
+  public void executeFailureOperation() throws MuleException {
+    exception.expectCause(instanceOf(ConnectionException.class));
+    exception.expectMessage("You are not allowed to speak with gus.");
+    OperationParameters params = builder().configName(HEISENBERG_CONFIG).build();
+    client.execute(HEISENBERG_EXT_NAME, "callGusFring", params);
+  }
+
+  @Test
+  @Description("Executes an operation asynchronously using the client and checks the output")
+  public void executeOperationWithNonBlocking() throws MuleException {
```
dfeist: Can this test be paramatized instead?  There are some tests here that only have versions using blocking API such that non-blocking API doesn't have same coverge.

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/ExtensionsClientTestCase.java
```diff
@@ -0,0 +1,265 @@
...
+    success.set(false);
+    OperationParameters params = builder().configName(HEISENBERG_CONFIG)
+        .addParameter("greeting", "jeje")
+        .addParameter("age", 23)
+        .addParameter("name", "Juani")
+        .addParameter("knownAddresses", emptyList())
+        .build();
+
+    Publisher<Result<String, Attributes>> publisher = client.executeAsync(HEISENBERG_EXT_NAME, "alias", params);
+    publisher.subscribe(new Subscriber<Result<String, Attributes>>() {
```
dfeist: No need to create a new subscriber here.  Instead use `Mono.from(publisher).doOnNext(..).doOnError(..).subscribe();`.  We are avoiding using Reactor in API, but fine to use in implementation/tests.

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/ExtensionsClientTestCase.java
```diff
@@ -0,0 +1,265 @@
...
+  }
+
+  @Test
+  @Description("Executes an operation asynchronously that fails using the client and checks the error")
+  public void executeFailureOperationWithNonBlocking() throws MuleException {
+    Reference<Boolean> success = new Reference<>();
+    success.set(false);
+    OperationParameters params = builder().configName(HEISENBERG_CONFIG).build();
+    Publisher<Result<Object, Attributes>> publisher = client.executeAsync(HEISENBERG_EXT_NAME, "callGusFring", params);
+    publisher.subscribe(new Subscriber<Result<Object, Attributes>>() {
```
dfeist: Same here.

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/ExtensionsClientTestCase.java
dfeist: Can use an enum, or use fuction like this:
https://github.com/mulesoft/mule/blob/mule-4.x/core-tests/src/test/java/org/mule/runtime/core/construct/FlowTestCase.java#L77

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/DefaultExtensionsClient.java
```diff
@@ -73,11 +73,10 @@
   public void initialise() throws InitialisationException {
     this.extensionManager = (ExtensionManagerAdapter) muleContext.getExtensionManager();
     try {
-      muleContext.getInjector().inject(this.policyManager);
-    } catch (MuleException e) {
-      throw new InitialisationException(e, this.policyManager);
+      this.policyManager = muleContext.getRegistry().lookupObject(PolicyManager.class);
```
marianogonzalez: why can't you inject this just like you did with the MuleContext?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessorFactory.java
```diff
@@ -29,37 +29,36 @@
  */
 public class OperationMessageProcessorFactory {
```
marianogonzalez: make class final and add private empty constructor

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/DefaultExtensionsClient.java
```diff
@@ -0,0 +1,174 @@
...
+import org.mule.runtime.module.extension.internal.runtime.operation.OperationMessageProcessor;
+import org.mule.runtime.module.extension.internal.runtime.resolver.StaticValueResolver;
+import org.mule.runtime.module.extension.internal.runtime.resolver.ValueResolver;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import javax.inject.Inject;
+
+import javafx.util.Pair;
```
marianogonzalez: don't use this dependency

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/DefaultExtensionsClient.java
```diff
@@ -0,0 +1,182 @@
...
+   * {@inheritDoc}
+   */
+  @Override
+  public <T, A extends Attributes> CompletableFuture<Result<T, A>> executeAsync(String extension, String operation,
+                                                                                OperationParameters parameters)
+    throws MuleException {
+    OperationMessageProcessor processor = createProcessor(extension, operation, parameters);
+    Publisher<Event> publisher = processor.apply(just(getInitialiserEvent(muleContext)));
+    //return from(publisher).map(e -> Result.<T, A>builder(e.getMessage()).build());
+    return null;
```
dfeist: null?  You can use `Mono.from(processor.apply(just(getInitialiserEvent(muleContext)))).toFuture()`

juandesi: Still WIP

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/DefaultExtensionsClient.java
```diff
@@ -0,0 +1,184 @@
...
+  /**
+   * {@inheritDoc}
+   * <p>
+   * Set's up the {@link PolicyManager} and the {@link ExtensionManager} used to execute the operations.
+   */
+  @Override
+  public void initialise() throws InitialisationException {
+    this.extensionManager = (ExtensionManagerAdapter) muleContext.getExtensionManager();
+    try {
+      this.policyManager = muleContext.getRegistry().lookupObject(PolicyManager.class);
```
marianogonzalez: why lookup instead of inject?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/DefaultExtensionsClient.java
```diff
@@ -0,0 +1,184 @@
...
+import org.mule.runtime.module.extension.internal.runtime.resolver.TypeSafeExpressionValueResolver;
+import org.mule.runtime.module.extension.internal.runtime.resolver.ValueResolver;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+
+import javax.inject.Inject;
+
+import javafx.util.Pair;
```
marianogonzalez: don't use this class

PR #3883, opened 2017-01-06
MULE-10366: Code formatter improvements

anosenzo:
Changes to use the Format goal by default and leave the Validate goal to use in the CI Server

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
+++ b/core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
@@ -52,7 +52,7 @@ public class ModuleFlowProcessingPhase
     extends NotificationFiringProcessingPhase<ModuleFlowProcessingPhaseTemplate> {
 
   // TODO MULE-11167 Policies should be non blocking
-  private static final String ENABLE_SOURCE_POLICIES_SYSTEM_PROPERTY = "enableSourcePolicies";
+  public static final String ENABLE_SOURCE_POLICIES_SYSTEM_PROPERTY = "enableSourcePolicies";
 
   private final PolicyManager policyManager;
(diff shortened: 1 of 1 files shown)


anosenzo:
@elrodro83 some comments about why we made the decision to leave the Format goal by default but let me know if you see something that is wrong:
* Now, a PR validation will fail with the formatter validation only in the case someone makes changes without compiling the code in their machines.
* If we remove the VALIDATE goal and use the FORMAT goal by default the devs won't be delayed if they forget formatting their code.
* In both cases the first point is going to continue happening.
* It is not necessary to use the VALIDATE goal in the normal flow because it will never fail as we are doing the FORMAT goal in a step before.
* The formatting in Intellij only happends when you use the shortcut or using the "reformat on commit" (https://intellij-support.jetbrains.com/hc/en-us/community/posts/206229129-Auto-format-in-IntelliJ-IDEA-) but a lot of devs use the console to commit in place of the Intellij UI.
* We want to leave the VALIDATE goal in the CI Server to avoid having changes that nobody is aware that happened.  
* Between doing the FORMAT goal by default or continue using the VALIDATE as default and provide a profile/property to make the formatting these are the pros/cons:
** Using the FORMAT by defeault there won't be delays when you forget to apply the code format and then compile.
** Using VALIDATE by default will force the devs to remember to add a profile/property each time they compile (you can avoid this using an alias in most cases).
** When devs from other teams need to compile Mule and send a PR the won't have the formatting error but they possible won't be aware of the formatting changes in their code.
** Using the FORMAT goal by default will make necessary to change the Jenkins jobs (that is something easy to do using the config slicing plugin) and when someone creates an empty job which uses the formatter plugin should remember to add the profile/property (it shouldn't be a problem because in general everybody makes a copy of an existing job or from a template job).

PR #3884, opened 2017-01-06
MULE-11358: Path Element is not correctly created for elements after a MessageFilter

mbuchwald:


tests/integration/src/test/java/org/mule/context/notification/FlowStackTestCase.java
```diff
@@ -392,4 +392,29 @@ public void subFlowDynamicWithScatterGatherChain() throws Exception
                 isFlowStackElement("subFlow", "/subFlowDynamicWithScatterGatherChain/processors/0/1/0/subFlow/subprocessors/0"),
                 isFlowStackElement("subFlowDynamicWithScatterGatherChain", "/subFlowDynamicWithScatterGatherChain/processors/0/1/0"));
     }
+
+
+    @Test
+    public void flowChainedFilter() throws Exception
+    {
+        muleContext.getClient().send("vm://in-flowChainedFilter", new DefaultMuleMessage("payload", muleContext));
```
pablokraan: instead of "payload" use TEST_MESSAGE

core/src/main/java/org/mule/routing/MessageFilter.java
```diff
@@ -188,4 +190,13 @@ public void dispose()
 
         LifecycleUtils.disposeIfNeeded(filter, LOGGER);
     }
+
+    @Override
+    public void addMessageProcessorPathElements(MessageProcessorPathElement pathElement)
+    {
+        if (!(next instanceof MessageProcessorChain))
```
pablokraan: Why is a message processor chain ignored?

tests/integration/src/test/java/org/mule/context/notification/FlowStackTestCase.java
```diff
@@ -392,4 +392,29 @@ public void subFlowDynamicWithScatterGatherChain() throws Exception
...
+
+        assertThat(FlowStackAsserter.stackToAssert, not(nullValue()));
+
+        assertStackElements(FlowStackAsserter.stackToAssert,
```
pablokraan: static import?

tests/integration/src/test/java/org/mule/context/notification/FlowStackTestCase.java
```diff
@@ -392,4 +392,29 @@ public void subFlowDynamicWithScatterGatherChain() throws Exception
...
+                isFlowStackElement("flow", "/flow/processors/0"),
+                isFlowStackElement("flowChainedFilter", "/flowChainedFilter/processors/0/1"));
+    }
+
+    @Test
+    public void flowForEachFilter() throws Exception
+    {
+        muleContext.getClient().send("vm://in-flowForEachFilter", new DefaultMuleMessage("payload", muleContext));
```
pablokraan: same comments in this test method

core/src/main/java/org/mule/routing/MessageFilter.java
mbuchwald: In the second test case (the _for each_ one) the DefaultMessageProcessorChainBuilder sets the test component as the next operator of the MessageFilter, but as a DefaultMessageProcessorChain. This works fine, but for the path creating, this is what caused the issue for the path creation

pablokraan: Add a comment explaining that

mbuchwald: After thinking about this, I realised that this is not the real fix for the issue. Will reopen the PR after I find out the real fix.

PR #3890, opened 2017-01-09
MULE-11292: Functions, TypedValues and ParameterResolvers are not supported inside a Pojo

estebanwasinger:


extensions/http/src/test/java/org/mule/extension/http/HttpRequestConfigTestCase.java
```diff
@@ -13,17 +13,22 @@
...
 import org.junit.Test;
+import org.mule.tck.junit4.rule.DynamicPort;
 
 public class HttpRequestConfigTestCase extends AbstractHttpTestCase {
 
   private static final String DEFAULT_PORT_HTTP_REQUEST_CONFIG_NAME = "requestConfigHttp";
   private static final String DEFAULT_PORT_HTTPS_REQUEST_CONFIG_NAME = "requestConfigHttps";
 
+  @Rule
+  public DynamicPort clientPort = new DynamicPort("clientPort");
```
marianogonzalez: why is this needed? This test is pre-existing and was working before

tests/http/src/test/java/org/mule/test/module/http/functional/listener/HttpListenerCustomTlsConfigTestCase.java
```diff
@@ -20,11 +20,16 @@
...
   public DynamicPort port1 = new DynamicPort("port1");
+
   @Rule
   public DynamicPort port2 = new DynamicPort("port2");
+
   @Rule
   public DynamicPort port3 = new DynamicPort("port3");
 
+  @Rule
+  public DynamicPort port4 = new DynamicPort("port4");
```
marianogonzalez: why?

afelisatti: Looking at the tests, it seems those ports end up being override by the operation. So instead of this you could have a static port (for request configs it will have no effect) or just avoid the port attribute in the config.

PR #3894, opened 2017-01-09
MULE-11279 Create JMH benchmarks for Message, Event and Flow

dfeist:


tests/performance/pom.xml
```diff
@@ -54,10 +106,15 @@
...
         </dependency>
         <dependency>
-            <groupId>org.databene</groupId>
-            <artifactId>contiperf</artifactId>
-            <version>2.2.0</version>
-            <scope>test</scope>
-        </dependency>    
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-core</artifactId>
+            <version>1.17.3</version>
```
aiannucci: This should be only test scope

tests/performance/src/main/java/org/mule/FlowOneWayBenchmark.java
```diff
@@ -0,0 +1,160 @@
...
+    public void setup() throws Exception
+    {
+        MuleContextFactory muleContextFactory = new DefaultMuleContextFactory();
+        muleContext = muleContextFactory.createMuleContext();
+        muleContext.start();
+
+        source = new TriggerableMessageSource();
+        flow = new Flow("flow", muleContext);
+        List<MessageProcessor> processors = new ArrayList<>();
+        flow.setMessageProcessors(Collections.<MessageProcessor>singletonList(new MessageProcessor()
```
elrodro83: static import

tests/performance/pom.xml
```diff
@@ -14,12 +14,66 @@
...
+            <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
                     <skip>${skipPerformanceTests}</skip>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
```
elrodro83: is this shading really needed?

tests/performance/src/main/java/org/mule/FlowOneWayBenchmark.java
```diff
@@ -0,0 +1,160 @@
...
+    private TriggerableMessageSource source;
+
+    @Param({"org.mule.processor.strategy.AsynchronousProcessingStrategy",
+            "org.mule.processor.strategy.QueuedAsynchronousProcessingStrategy"})
+    public String processingStrategyFactory;
+
+    @Setup
+    public void setup() throws Exception
+    {
+        MuleContextFactory muleContextFactory = new DefaultMuleContextFactory();
```
elrodro83: this seems to be repeated (or a very similar code) in other tests. Extract to a supercalss/utils

tests/performance/src/main/java/org/mule/FlowRequestResponseBenchmark.java
```diff
@@ -0,0 +1,157 @@
...
+/**
+ *
+ */
+@State(Benchmark)
+@Fork(1)
+@BenchmarkMode(AverageTime)
+@OutputTimeUnit(MICROSECONDS)
+@Warmup(iterations = 10)
+@Measurement(iterations = 10)
+public class FlowRequestResponseBenchmark
```
elrodro83: Can't ONE_WAY/REQ_RESP be parameterized?

tests/performance/src/test/java/org/mule/AbstractBenchmarkAssertionTestCase.java
```diff
@@ -0,0 +1,100 @@
...
+import org.openjdk.jmh.results.RunResult;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+/**
+ * Base class for creating benchmark assertion tests allowing JMH benchmark results to be asserted using JUnit tests with Hamcrest.
+ */
+public abstract class AbstractBenchmarkAssertionTestCase extends AbstractMuleTestCase
```
elrodro83: is this class intended to be reused? only one test is extending it in this PR

tests/performance/pom.xml
dfeist: Yes, it's the recommended way to create a uber-jar for running benchmarks on different environments without maven.

tests/performance/src/test/java/org/mule/AbstractBenchmarkAssertionTestCase.java
dfeist: Yes, it's not reused much, because hardly any existing assertions.

tests/performance/pom.xml
dfeist: Why? Anything wrong with putting benchmarks in src/main and tests that assert benchmarks in src/test?

tests/performance/src/main/java/org/mule/FlowRequestResponseBenchmark.java
dfeist: I can't remember why i didn't param this, maybe left over from previous approach. Parameterizing it should be easy with JMH.

tests/performance/src/main/java/org/mule/FlowOneWayBenchmark.java
dfeist: Needed for explicit type.

tests/performance/src/main/java/org/mule/FlowRequestResponseBenchmark.java
dfeist: I remembered now.  ProcessingStrategy param is different in each test. In 4.x this can be one test.

tests/performance/pom.xml
elrodro83: do we want to run this tests outside of maven?

dfeist: Yes, it's easier to push a single jar to a remote environment, also it's best practice to run it standalone to get best results.

PR #3897, opened 2017-01-09
MULE-11342: Add leak detection to unit test scheduler service

elrodro83:


compatibility/core/src/main/java/org/mule/compatibility/core/transport/AbstractConnector.java
```diff
@@ -405,17 +405,18 @@ public final synchronized void start() throws MuleException {
       return;
     }
 
-    scheduler = createScheduler();
-
     if (!isConnected()) {
+      scheduler = createScheduler();
```
marianogonzalez: why should a scheduler only be created if not connected, and how would a connector already be connected on the start() method?

services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultSchedulerService.java
```diff
@@ -85,6 +88,7 @@
   private org.quartz.Scheduler quartzScheduler;
 
   private volatile boolean started = false;
+  private List<Scheduler> activeSchedulers = synchronizedList(new ArrayList<>());
```
marianogonzalez: who clears this list? it looks like you're speculating with schedulers only be disposed when the application is disposed and that's not always true. Sources are a good counter example. Custom extensions is another. These schedulers should be decorated so that they're removed from this list when shutdown

elrodro83: Why would a source not dispose its schedulers when the application it's in is disposed?

compatibility/core/src/main/java/org/mule/compatibility/core/transport/AbstractConnector.java
elrodro83: That scheduler only makes sense to handle situations that happen only when a connection is active. Will rename the variable to reflect this.

marianogonzalez: in that case, who's responsible for disposing that scehduler?

elrodro83: the stop() method

services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultSchedulerService.java
elrodro83: Also, there is a callback provided so these are removed when the scheduler is stopped,

compatibility/core/src/main/java/org/mule/compatibility/core/transport/AbstractConnector.java
```diff
@@ -405,21 +405,29 @@ public final synchronized void start() throws MuleException {
...
         throw me;
       } catch (Exception e) {
+        shutdownInternalScheduler();
         throw new EndpointConnectException(e, this);
       }
     } else {
-      startAfterConnect();
+      try {
+        startAfterConnect();
+      } catch (MuleException me) {
```
dfeist: What about RuntimeException? This is caught on `connect()` but not here.

core/src/main/java/org/mule/runtime/core/api/scheduler/SchedulerService.java
```diff
@@ -125,4 +126,11 @@
    */
   Scheduler customScheduler(SchedulerConfig config, int queueSize);
 
+  /**
+   * Provides a read-only view of all currently active {@link Scheduler}s created through this service.
+   * 
+   * @return all currently active {@link Scheduler}s.
+   */
+  List<Scheduler> getActiveSchedulers();
```
dfeist: Is "Active" prefix needed?  Does scheduler service maintain non-active schedulers?

services/scheduler/src/test/java/org/mule/service/scheduler/internal/DefaultSchedulerQuartzTestCase.java
```diff
@@ -54,7 +54,8 @@
   public void before() throws SchedulerException {
     super.before();
     executor = new DefaultScheduler(DefaultSchedulerQuartzTestCase.class.getSimpleName(), sharedExecutor, 1,
-                                    sharedScheduledExecutor, sharedQuartzScheduler, CUSTOM);
+                                    sharedScheduledExecutor, sharedQuartzScheduler, CUSTOM, sched -> {
```
dfeist: Crate `null` consumer or SchedulerStopper for clarity instead of repeating empty consumer in multiple places?

tests/functional/src/main/java/org/mule/functional/listener/SystemExceptionListener.java
```diff
@@ -21,14 +24,18 @@
...
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 /**
  * Listener for exceptions managed by the {@link org.mule.runtime.core.api.exception.SystemExceptionHandler}.
  */
 public class SystemExceptionListener {
 
+  private transient Logger logger = LoggerFactory.getLogger(SystemExceptionListener.class);
```
dfeist: why transient? why not final?

PR #3898, opened 2017-01-09
MULE-7211: Fix flaky test AbstractWSConsumerFunctionalTestCase (when …

alepulver:
…a global error counter across tests reaches 10, the expected error message is truncated; to avoid it we increase the logging level for the parser)

### DIFF
--- a/modules/ws/src/test/java/org/mule/module/ws/functional/AbstractWSConsumerFunctionalTestCase.java
+++ b/modules/ws/src/test/java/org/mule/module/ws/functional/AbstractWSConsumerFunctionalTestCase.java
@@ -7,6 +7,7 @@
 package org.mule.module.ws.functional;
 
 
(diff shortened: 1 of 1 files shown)


modules/ws/src/test/java/org/mule/module/ws/functional/AbstractWSConsumerFunctionalTestCase.java
```diff
@@ -53,12 +57,17 @@
     protected void doSetUpBeforeMuleContextCreation() throws Exception
     {
         System.setProperty(USE_TRANSPORT_FOR_URIS, Boolean.toString(useTransportForUris));
+
+        final Logger xmlLogger = Logger.getLogger("com.sun.xml.bind");
```
elrodro83: Use a JUnit Rule for this

PR #3903, opened 2017-01-09
MULE-11369: make XMLInputFactory throw an exception again when reject…

alepulver:
…ing internal entities (as other parsers do)


modules/xml/src/test/java/org/mule/transformers/xml/xslt/XsltTransformerBLTestCase.java
```diff
@@ -50,12 +50,11 @@ public void enabled() throws Exception
         assertThat(output, containsString("010101010101010101010101010101010101010101010101"));
     }
 
-    @Test
+    @Test(expected = TransformerMessagingException.class)
```
elrodro83: use ExpectedException rule and assert the message also

PR #3904, opened 2017-01-09
MULE-10939: Cover all DW/MEL scenarios for ExpressionManager in tests

afelisatti:


core-tests/src/test/java/org/mule/runtime/core/el/DefaultExpressionManagerTestCase.java
```diff
@@ -208,68 +142,27 @@ public void parseCompatibility() throws MuleException {
     assertThat(expressionManager.parse("dw:'this is ' ++ payload", testEvent(), mock(FlowConstruct.class)),
                is(String.format("this is %s", TEST_PAYLOAD)));
     expectedException.expect(RuntimeException.class);
-    expressionManager.parse(weavify(expression), testEvent(), mock(FlowConstruct.class));
+    expressionManager.parse("dw:" + expression, testEvent(), mock(FlowConstruct.class));
```
pablolagreca: why are you removing the call to weavify and doing it manually?

core/src/main/java/org/mule/runtime/core/api/el/ExpressionLanguage.java
```diff
@@ -22,6 +22,13 @@
 public interface ExpressionLanguage {
 
   /**
+   * Registers the given {@link BindingContext} as global.
```
pablolagreca: Can you add javadoc about what would happen with a previous global context if another one is registered? Or if once a global context was registered it can/cannot be changed?. What about concurrency against possible expressions under execution?

core-tests/src/test/java/org/mule/runtime/core/el/DefaultExpressionManagerTestCase.java
afelisatti: Because this is the only scenario in which the manager has to do a prefix check so it didn't make sense to have a method for it.

core-tests/src/test/java/org/mule/runtime/core/el/DataWeaveExpressionLanguageTestCase.java
```diff
@@ -21,11 +21,11 @@
 import static org.mockito.Mockito.when;
 import static org.mule.runtime.api.metadata.DataType.OBJECT;
 import static org.mule.runtime.api.metadata.DataType.STRING;
-import static org.mule.runtime.core.el.v2.MuleExpressionLanguage.ATTRIBUTES;
-import static org.mule.runtime.core.el.v2.MuleExpressionLanguage.ERROR;
-import static org.mule.runtime.core.el.v2.MuleExpressionLanguage.FLOW;
-import static org.mule.runtime.core.el.v2.MuleExpressionLanguage.PAYLOAD;
-import static org.mule.runtime.core.el.v2.MuleExpressionLanguage.VARIABLES;
+import static org.mule.runtime.core.el.v2.DataWeaveExpressionLanguage.ATTRIBUTES;
```
afelisatti: We haven't moved the MEL stuff so there may be some conflicts but I'll check.

PR #3910, opened 2017-01-10
MULE-11279 Create JMH benchmarks for Message, Event and Flow.

dfeist:


tests/performance/src/main/java/org/mule/FlowBenchmark.java
```diff
@@ -0,0 +1,78 @@
...
+public class FlowBenchmark extends AbstractBenchmark {
+
+  public static final String TEST_PAYLOAD = "test";
+  public static final String TEST_CONNECTOR = "test";
+
+  private MuleContext muleContext;
+  private Flow flow;
+  private TriggerableMessageSource source;
+
+  @Param({"org.mule.runtime.core.processor.strategy.LegacySynchronousProcessingStrategyFactory",
```
elrodro83: i think i already asked this, but can't you put the class instead of the string?

tests/performance/src/main/java/org/mule/el/mvel/MVELBenchmark.java
```diff
@@ -0,0 +1,81 @@
...
+import org.mule.runtime.core.construct.Flow;
+import org.mule.runtime.core.el.mvel.MVELExpressionLanguage;
+
+import java.util.Random;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.TearDown;
+
+public class MVELBenchmark extends AbstractBenchmark {
```
elrodro83: should we have a compatibility-performance module?

tests/performance/src/main/java/org/mule/FlowBenchmark.java
dfeist: No :-(

tests/performance/src/main/java/org/mule/el/mvel/MVELBenchmark.java
dfeist: think that's overkill for now, this is only a test module.

PR #3922, opened 2017-01-11
MULE-11380: DynamicConfigurations not disposed when start() fails

elrodro83:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DynamicConfigurationProvider.java
```diff
@@ -152,7 +159,12 @@ protected void registerConfiguration(ConfigurationInstance configuration) {
         }
 
         if (lifecycleManager.isPhaseComplete(Startable.PHASE_NAME)) {
-          startConfig(configuration);
+          try {
+            startConfig(configuration);
+          } catch (Exception e) {
+            disposeIfNeeded(configuration, LOGGER);
```
marianogonzalez: shouldn't the same happen if it fails on initialise()?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DynamicConfigurationProvider.java
```diff
@@ -155,13 +156,19 @@ protected void registerConfiguration(ConfigurationInstance configuration) {
...
+            throw e;
+          }
         }
 
         if (lifecycleManager.isPhaseComplete(Startable.PHASE_NAME)) {
           try {
             startConfig(configuration);
           } catch (Exception e) {
+            stopIfNeeded(configuration);
```
marianogonzalez: if stopFails, you need to dispose anyway

elrodro83: i hate lifecycles...

PR #3923, opened 2017-01-11
MULE-11257: Remove Event uses from HTTP extension - Replace Function …

afelisatti:
…uses in request operation.

### DIFF
--- a/modules/http/src/main/java/org/mule/extension/http/internal/request/HttpRequestOperations.java
+++ b/modules/http/src/main/java/org/mule/extension/http/internal/request/HttpRequestOperations.java
@@ -41,8 +41,6 @@ import org.mule.runtime.extension.api.runtime.process.CompletionCallback;
 import org.mule.runtime.module.http.api.HttpConstants;
 import org.mule.runtime.module.http.internal.HttpParser;
 
-import java.util.function.Function;
-
 import javax.inject.Inject;
 
 public class HttpRequestOperations implements Initialisable, Disposable {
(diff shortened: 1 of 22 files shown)


modules/http/src/main/java/org/mule/extension/http/internal/request/HttpRequestOperations.java
```diff
@@ -87,19 +85,19 @@ public void request(@Optional(defaultValue = "/") String path,
       HttpRequesterRequestBuilder resolvedBuilder = requestBuilder != null ? requestBuilder : new HttpRequesterRequestBuilder();
       UriParameters uriParameters = client.getDefaultUriParameters();
 
-      String resolvedHost = resolveIfNecessary(overrides.getHost(), uriParameters.getHost(), muleEvent);
-      Integer resolvedPort = resolveIfNecessary(overrides.getPort(), uriParameters.getPort(), muleEvent);
-      String resolvedBasePath = config.getBasePath().apply(muleEvent);
+      String resolvedHost = resolveIfNecessary(overrides.getHost(), uriParameters.getHost());
```
marianogonzalez: you're removing the usages of event but still receiving the even in the operation

afelisatti: Yes, I'll leave that for another PR since I won't be able to fully remove it from there until I can synchronise with @elrodro83 regarding oauth. Though that might be in the last PR of the series, my next one will add all the payload variables and such to replace it.

PR #3938, opened 2017-01-12
MULE-11404: Apply OperationUX to the WSC

juandesi:


extensions/wsc/src/main/java/org/mule/extension/ws/internal/metadata/HeadersElementResolver.java
```diff
@@ -61,6 +61,7 @@ private MetadataType buildHeaderType(BaseTypeBuilder baseTypeBuilder, TypeLoader
                                        Message message)
       throws MetadataResolvingException {
     ObjectTypeBuilder typeBuilder = baseTypeBuilder.objectType();
+    typeBuilder.id("headers");
```
marianogonzalez: typeId should be `Map.class.getName()`

PR #3939, opened 2017-01-12
MULE-11298: addition of useConnectorToRetrieveWsdl to manage strategies.

fsgonz:


core/src/main/java/org/mule/api/config/MuleProperties.java
```diff
@@ -201,4 +201,5 @@
     public static final String MULE_FLOW_TRACE = SYSTEM_PROPERTY_PREFIX + "flowTrace";
     public static final String CONTENT_TYPE_PROPERTY = "Content-Type";
     public static final String MULE_PGP_ENCRYPTION_ALGORITHM = SYSTEM_PROPERTY_PREFIX + "pgp.encryption.algorithm";
+    public static final String MULE_USE_CONNECTOR_TO_RETRIEVE_WSDL = SYSTEM_PROPERTY_PREFIX + "use.connector.to.retrieve.wsdl";
```
alepulver: I would use something like SYSTEM_PROPERTY_PREFIX + "ws.useConnectorToRetrieveWsdl"

modules/ws/src/main/java/org/mule/module/ws/consumer/WSConsumer.java
```diff
@@ -107,20 +106,35 @@ public MuleEvent process(MuleEvent event) throws MuleException
...
+    private void initializeWsdlRetrievalStrategy() throws InitialisationException
     {
         HttpRequesterConfig httpRequesterConfig = config.getConnectorConfig();
+        String url = config.getWsdlLocation();
+
+        boolean isHttpRequester = verifyIsHttpRequester(url);
+
+        if (httpRequesterConfig == null && config.isUseConnectorToRetrieveWsdl())
+        {
+            throw new InitialisationException(CoreMessages.createStaticMessage("The useConnectorToRetrieveWsdl option requires connectorConfig to work"),
```
alepulver: It would be better to print a warning message here instead of failing because the user might not want to use it everywhere (the system property affects all ws consumers). It's not necessary to test that the warning is shown.

modules/ws/src/test/resources/ws-consumer-wsdl-null-wsdl-location.xml
```diff
@@ -0,0 +1,25 @@
...
+
+	<http:request-config name="HTTP_Request_Configuration"
+		protocol="HTTPS" host="fake.com" port="80">
+	</http:request-config>
+
+
+	<ws:consumer-config name="Web_Service_Consumer"
+		service="TestService" port="TestPort" serviceAddress="http://fake.com/services/Test"
+		connectorConfig="HTTP_Request_Configuration"
+		useConnectorToRetrieveWsdl="true" />
```
alepulver: This was removed and shouldn't be valid. If the test asserts the configuration failed to load, it should make sure it didn't fail because of errors like this one.

PR #3942, opened 2017-01-13
AGW-978 upgrading mockito version

santiagopurucker:


pablokraan:
This PR must linked to a MULE issue, not AGW (which is not public)

PR #3946, opened 2017-01-15
MULE-11359: Adapt container to work with new binary heavy package

pablokraan:
_ Moving plugins/lib to lib/shared folder

### DIFF
--- a/modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/builder/ApplicationFileBuilder.java
+++ b/modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/builder/ApplicationFileBuilder.java
@@ -8,11 +8,13 @@
 package org.mule.runtime.module.deployment.impl.internal.builder;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static org.apache.commons.io.FilenameUtils.getName;
 import static org.apache.commons.lang.StringUtils.isEmpty;
+import static org.mule.runtime.container.api.MuleFoldersUtil.getAppPluginsFolderPath;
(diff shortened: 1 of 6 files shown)


modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/builder/ApplicationFileBuilder.java
```diff
@@ -148,7 +151,7 @@ public ApplicationFileBuilder containingPlugin(ArtifactPluginFileBuilder plugin)
   public ApplicationFileBuilder sharingLibrary(String jarFile) {
     checkImmutable();
     checkArgument(!StringUtils.isEmpty(jarFile), "Jar file cannot be empty");
-    resources.add(new ZipResource(jarFile, "plugins/lib/" + FilenameUtils.getName(jarFile)));
+    resources.add(new ZipResource(jarFile, LIB_FOLDER + separator + SHARED_FOLDER + separator + getName(jarFile)));
```
pablolagreca: Wouldn't be better to have a method in MuleFoldersUtil to get the shared folder name instead?

pablokraan: There is already a method to do that in MuleFoldersUtil, but it provides the folder relative to MULE_HOME (ie: in the container). This folder here, is relative to the folder where the app is being built.

pablolagreca: Can we have both methods? One for getting the relative path and another one for getting the absolute path?.

PR #3951, opened 2017-01-16
MULE-11352: Add Event (or a view of it) and its Builder to mule-api

elrodro83:
First iteration of api cleanup:
* Remove exchangePattern and synchronous flag. Users of those are changed to get that data from the flow.
* Remove old services code from statistics.


dfeist:
Should `org.mule.functional.junit4.TestEventBuilder#withExchangePattern` be removed as part of this?

PR #3953, opened 2017-01-17
MULE-11357:reducing the artifacts when packaging plugins (fix build)

fernandezlautaro:


tests/integration/pom.xml
```diff
@@ -99,27 +99,31 @@
             <groupId>org.mule.modules</groupId>
             <artifactId>mule-module-validation</artifactId>
             <version>${project.version}</version>
+            <classifier>mule-plugin</classifier>
```
elrodro83: shouldn't mule-module-file-extension-common have the classifier too?

fernandezlautaro: it should but as the mule-module-file-extension-common renaming will be part of MULE-11383 (all artifacts will become JAR files) it's left as it is right now

PR #3955, opened 2017-01-17
MULE-11352: Remove MuleEvent from mule-api

elrodro83:
Also remove nonBlocking from FlowRunner


marianogonzalez:
The changes are in order, but consider that this will break the agent and tooling client possibly too. 

Consider that before merging.

PR #3957, opened 2017-01-18
MULE-11255: Add support to use artifact plugins on policies

pablokraan:
_ Adding ExtensionManagerFactory interface+ implementations
_ Converting ApplicationExtensionManagerConfigurationBuilder into ArtifactExtensionManagerConfigurationBuilder
_ Adding basic composite extension manager for policies
_ Adding plugins to DefaultPolicyTemplate
_ Adding plugins to PolicyTemplate
_ Making PolicyTemplateClassLoaderBuilder to return a  MuleDeployableArtifactClassLoader class loader
_ Changing ArtifactClassLoaderFactory to DeployableArtifactClassLoaderFactory on policy factory
_ Updating PolicyFileBuilder to include plugins
_ Adding plugins to PolicyTemplate
_ Making PolicyTemplateClassLoaderBuilder to return a  MuleDeployableArtifactClassLoader class loader
_ Extracting code into BundleDescriptorUtils
_ Adding integration tests

### DIFF
--- a/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/ArtifactExtensionManagerFactory.java
+++ b/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/ArtifactExtensionManagerFactory.java
@@ -4,29 +4,26 @@
  * license, a copy of which has been included with this distribution in the
  * LICENSE.txt file.
  */
-package org.mule.runtime.module.deployment.impl.internal.application;
+
+package org.mule.runtime.module.deployment.impl.internal.policy;
 
 import static java.lang.String.format;
-import static org.mule.runtime.api.util.Preconditions.checkNotNull;
+import static org.mule.runtime.module.deployment.impl.internal.artifact.ArtifactExtensionManagerConfigurationBuilder.META_INF_FOLDER;
 import static org.mule.runtime.module.extension.internal.ExtensionProperties.EXTENSION_MANIFEST_FILE_NAME;
 import static org.mule.runtime.module.extension.internal.loader.java.JavaExtensionModelLoader.TYPE_PROPERTY_NAME;
 import static org.mule.runtime.module.extension.internal.loader.java.JavaExtensionModelLoader.VERSION;
 import static org.slf4j.LoggerFactory.getLogger;
-
 import org.mule.runtime.api.deployment.meta.MulePluginModel;
-import org.mule.runtime.api.lifecycle.InitialisationException;
 import org.mule.runtime.api.meta.model.ExtensionModel;
+import org.mule.runtime.core.DefaultMuleContext;
 import org.mule.runtime.core.api.MuleContext;
-import org.mule.runtime.core.api.config.ConfigurationBuilder;
 import org.mule.runtime.core.api.extension.ExtensionManager;
-import org.mule.runtime.core.config.builders.AbstractConfigurationBuilder;
 import org.mule.runtime.deployment.model.api.plugin.ArtifactPlugin;
 import org.mule.runtime.deployment.model.api.plugin.ArtifactPluginDescriptor;
 import org.mule.runtime.extension.api.loader.ExtensionModelLoader;
 import org.mule.runtime.extension.api.manifest.ExtensionManifest;
 import org.mule.runtime.module.extension.internal.loader.ExtensionModelLoaderRepository;
 import org.mule.runtime.module.extension.internal.loader.java.JavaExtensionModelLoader;
-import org.mule.runtime.module.extension.internal.manager.DefaultExtensionManagerAdapterFactory;
 import org.mule.runtime.module.extension.internal.manager.ExtensionManagerFactory;
 
 import java.net.URL;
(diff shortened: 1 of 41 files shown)


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/ArtifactExtensionManagerFactory.java
```diff
@@ -130,12 +111,11 @@ private void discoverExtensionThroughJsonDescriber(ArtifactPlugin artifactPlugin
...
-    try {
-      return extensionManagerFactory.createExtensionManager(muleContext);
-    } catch (Exception e) {
-      throw new InitialisationException(e, muleContext);
-    }
-  }
+  private ExtensionManager createExtensionManager(MuleContext muleContext) {
+    ExtensionManager extensionManager = extensionManagerFactory.create(muleContext);
 
+    ((DefaultMuleContext) muleContext).setExtensionManager(extensionManager);
```
marianogonzalez: isn't the configuration builder already doing this?

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  @Override
+  public void registerConfigurationProvider(ConfigurationProvider configurationProvider) {
+    throw new UnsupportedOperationException("Composite extension manager cannot register extension providers");
+  }
+
+  @Override
+  public ExtensionManifest parseExtensionManifestXml(URL manifestUrl) {
+    throw new UnsupportedOperationException("Composite extension manager cannot parse extension manifests");
```
marianogonzalez: why can't this just be delegated to the parent one?

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
```diff
@@ -0,0 +1,130 @@
...
+import java.net.URL;
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * Composes {@link ExtensionManager} from a child and a parent artifacts, so child artifact can access extensions provided by the
+ * parent.
+ */
+public class CompositeArtifactExtensionManager implements ExtensionManager {
```
marianogonzalez: This class seems to be explicitly defined for this particular case, but we do know that we'll also need composition for the domains use case. This composite should be defined in a more generic way (maybe even be abstract) and then have a specialization which has the particular rules of the policy scenario.

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
```diff
@@ -0,0 +1,130 @@
...
+
+  private final ExtensionManager parentExtensionManager;
+  private final ExtensionManager childExtensionManager;
```
marianogonzalez: this should be a list in order for this class to support more use cases

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  @Override
+  public void registerExtension(ExtensionModel extensionModel) {
+    throw new UnsupportedOperationException("Composite extension manager cannot register extensions");
+  }
+
+  @Override
+  public Set<ExtensionModel> getExtensions() {
+    return extensionModels;
```
marianogonzalez: i would just collect rather than returning this fixed value.

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/ArtifactExtensionManagerFactory.java
pablokraan: Looks like that is the case, I'll update it

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
pablokraan: Is true, applications will use something similar one domains support plugins. But is kind of useless to create an abstraction that will be used only here and there is no knowledge about which parts should be in that abstraction.

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
pablokraan: Artifacts can have only a parent artifact, I don't see any use case that will require a list of parents. Can you provide one?

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
pablokraan: Extensions are registered on the composed extension managers when they are created/initialized, before being composed. There is no extension registered after that point. Why should this be dynamic?

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
pablokraan: As in explained in the previous point, there is no more extension processing once the composite has been created. Is not going to be used, unless you can provide an scenario I missed

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
marianogonzalez: i meant that the childs should be alist

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
marianogonzalez: I can't provide a scenario, but if there's no impediment to fulfill this API I'd rather do it. The scenario might appear in the future and this becomes a bug we might not catch before hand. Liskov

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
pablokraan: Understood, there was an error on my previous sentence. What I meant was that mule artifact can have only a parent, an because this composite is for a child artifact, is a 1-1 relation. One parent, one child. No need for a list. Unless I missing something

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
pablokraan: There is no reason to implement this method or the other ones that throw UnsupportedOperationException because they will not be used.  The problem here is that this interface is used for two different things, but I won't fix that as part of this PR

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
marianogonzalez: I would very much like to leverage this class to support extenisons in domains. In that case, one parent (domain) can have many childs (apps)

pablokraan: That is the idea, every app will have a CompositeExtensionManager, parent will be the domain, child will be the app.
Maybe the name is not clear, but the javadoc explains it

PR #3958, opened 2017-01-18
MULE-11349 Introduce ChildEventContext to support per-flow completion

dfeist:


core/src/main/java/org/mule/runtime/core/DefaultEventContext.java
```diff
@@ -161,4 +174,91 @@ public void subscribe(Subscriber<? super Event> s) {
     subscribers.add(s);
     monoProcessor.subscribe(s);
   }
+
+  private static class ChildEventContext implements EventContext, Serializable {
+
+    private static final long serialVersionUID = 1054412872901205234L;
+
+    private transient MonoProcessor<Event> monoProcessor = MonoProcessor.create();
+    private List<Subscriber> subscribers = new ArrayList<>();
```
elrodro83: can we enforce somehow that this subscribers are serializable?

tests/integration/src/test/java/org/mule/test/integration/exceptions/ErrorHandlerTestCase.java
```diff
@@ -231,6 +231,9 @@ private void callAndThrowException(Object payload, final Exception exceptionToTh
     public Event process(Event event) throws MuleException {
       Throwable exception = (Throwable) event.getVariable("exception").getValue();
       if (exception instanceof MuleException) {
+        if (exception instanceof MessagingException) {
```
elrodro83: when can it happen that a MessagingException doesn't have the event set?

dfeist: In this test case, it's creating messaging exception manually in test method, passing it as a variable and then launching it from a custom component. When it is launched it has the wrong event in it.

core/src/main/java/org/mule/runtime/core/DefaultEventContext.java
dfeist: No because that would rule out the use of reactor/rxjava as well as be quite limiting.  Also regardless of reactor/rxJava a HTTP callback typically wouldn't be serializable either as it needs to maintain reference to connection.

tests/integration/src/test/java/org/mule/test/integration/exceptions/ErrorHandlerTestCase.java
elrodro83: didn't see it was a test method, ok

PR #3961, opened 2017-01-18
MULE-10809: Support native libraries/required dependencies

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/JavaModelLoaderDelegate.java
```diff
@@ -168,6 +177,34 @@ public ExtensionDeclarer declare(ExtensionLoadingContext context) {
     return declarer;
   }
 
+  protected void parseExternalLibs(WithAnnotations withAnnotations, DeclaresExternalLibraries declarer) {
+    Optional<ExternalLibs> externalLibs = withAnnotations.getAnnotation(ExternalLibs.class);
```
alegmarra: you can use `MuleExtensionAnnotationParser#parseRepeatableAnnotation` like we do for subtypes: 
`List<SubTypeMapping> typeMappings = parseRepeatableAnnotation(type, SubTypeMapping.class, c -> ((SubTypesMapping) c).value());`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/JavaModelLoaderDelegate.java
```diff
@@ -168,6 +177,34 @@ public ExtensionDeclarer declare(ExtensionLoadingContext context) {
...
+        .withDescription(externalLibAnnotation.description());
+
+
+    ifNotBlank(externalLibAnnotation.fileName(), builder::withFileName);
+    ifNotBlank(externalLibAnnotation.requiredClassName(), builder::withRequiredClassName);
+
+    declarer.withExternalLibrary(builder.build());
+  }
+
+  private void ifNotBlank(String value, Consumer<String> consumer) {
```
alegmarra: candidate for `org.mule.runtime.core.util.StringUtils` ?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/JavaModelLoaderDelegate.java
marianogonzalez: looked at it. Not exactly the same. Besides that depends on classes and I'm using Wasinger's type abstraction

PR #3963, opened 2017-01-18
MULE-11339: Configurable AbortPolicy for Schedulers

elrodro83:
Also, made some improvements in the scheduler tests


services/scheduler/src/main/java/org/mule/service/scheduler/internal/executor/WaitPolicy.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A handler for unexecutable tasks that waits until the task can be submitted for execution or times out. Generously snipped from
+ * the jsr166 repository at:
+ * <a href="http://gee.cs.oswego.edu/cgi-bin/viewcvs.cgi/jsr166/src/main/java/util/concurrent/ThreadPoolExecutor.java"></a>.
+ */
+// @Immutable
+public class WaitPolicy implements RejectedExecutionHandler {
```
dfeist: Do we need to two copies of this, what is the difference? Maybe make this clear in javadoc.

services/scheduler/src/main/java/org/mule/service/scheduler/internal/executor/WaitPolicy.java
```diff
@@ -0,0 +1,61 @@
...
+    this.time = (time < 0 ? Long.MAX_VALUE : time);
+    this.timeUnit = timeUnit;
+  }
+
+  @Override
+  @SuppressWarnings("boxing")
+  public void rejectedExecution(Runnable r, ThreadPoolExecutor e) {
+    try {
+      if (e.isShutdown()) {
+        throw new RejectedExecutionException("ThreadPoolExecutor is already shut down");
```
dfeist: Shouldn't this be SchedulerBusyException? Especially if WaitPolicy is Scheduler specific.

services/scheduler/src/main/java/org/mule/service/scheduler/internal/executor/WaitPolicy.java
```diff
@@ -0,0 +1,61 @@
...
+      } else if (!e.getQueue().offer(r, time, timeUnit)) {
+        String message = format("Scheduler did not accept within %1d %2s", time, timeUnit);
+        throw new RejectedExecutionException(message);
```
dfeist: Same here.

services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultSchedulerService.java
```diff
@@ -282,15 +286,15 @@ public void start() throws MuleException {
 
     cpuLightExecutor = new ThreadPoolExecutor(threadPoolsConfig.getCpuLightPoolSize(), threadPoolsConfig.getCpuLightPoolSize(),
                                               0, SECONDS, new LinkedBlockingQueue<>(threadPoolsConfig.getCpuLightQueueSize()),
-                                              new SchedulerThreadFactory(cpuLightGroup));
+                                              new SchedulerThreadFactory(cpuLightGroup), byCallerThreadGroupPolicy);
```
dfeist: The AbortPolicy is not made configurable as part of this PR from what I can see. That might not be needed, but it doesn't match the JIRA issue.

services/scheduler/src/main/java/org/mule/service/scheduler/internal/executor/ByCallerThreadGroupPolicy.java
```diff
@@ -0,0 +1,60 @@
...
+   * @param waitGroups the group of threads for which a {@link WaitPolicy} will be applied. For the rest, an {@link AbortPolicy}
+   *        will be applied.
+   */
+  public ByCallerThreadGroupPolicy(Set<ThreadGroup> waitGroups) {
+    this.waitGroups = unmodifiableSet(waitGroups);
+  }
+
+  @Override
+  public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
+    if (currentThread().getThreadGroup() != null && waitGroups.contains(currentThread().getThreadGroup())) {
```
dfeist: Should custom thread pool always wait, how can that we hard-coded without knowing what custom pool is used for? Remember HTTP selectors will be custom soon.  Also are these rules generic enough to apply to everything do you think? Did we say CpuIntensive -> CpuLight should fail, i'm not 100% convinced it should.  If we want a general rule then slower->faster should block and faster->slower should fail probably.

services/scheduler/src/test/java/org/mule/service/scheduler/internal/executor/WaitPolicyTestCase.java
```diff
@@ -0,0 +1,317 @@
...
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.ReentrantLock;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class WaitPolicyTestCase extends AbstractMuleTestCase {
```
dfeist: Duplicate test cases too (that's ok if duplicate implementations is justified)

core/src/main/resources/META-INF/mule-module.properties
```diff
@@ -41,6 +41,7 @@ artifact.export.classPackages=org.mule.runtime.core,\
                               org.mule.runtime.core.api.routing.filter,\
                               org.mule.runtime.core.api.rx,\
                               org.mule.runtime.core.api.scheduler,\
+                              org.mule.runtime.core.api.scheduler.exception,\
```
elrodro83: It will get to the source, so it will need to be able to load the class for the exception on whatever classloader the source is

tests/integration/src/test/java/org/mule/test/service/scheduler/SchedulerServiceTestCase.java
```diff
@@ -81,16 +83,15 @@ public void schedulerCustomName() {
   @Test
   @Description("Tests that the exception that happens when a thread pool is full is properly handled.")
   public void overloadErrorHandling() throws Exception {
-    // This has to match the default value returned by ThreadPoolsConfig#getIoMaxPoolSize() + ThreadPoolsConfig#getIoQueueSize().
```
elrodro83: this comment is removed in this pr, its no longer relevant

services/scheduler/src/main/java/org/mule/service/scheduler/internal/executor/ByCallerThreadGroupPolicy.java
elrodro83: i'll add the configuration for the custom schedulers.
For the cpu intensive -> cpu light scenario, we shouldavoid blocking the first one since it is cpu bound. Maybe this can be improved when we make cpu intensive into a ForkJoin pool, since that will allow to allocate threads to compensate the blocked ones.

services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultSchedulerService.java
elrodro83: Only for custom schedulers it would be relevant. Added that change

services/scheduler/src/main/java/org/mule/service/scheduler/internal/executor/ByCallerThreadGroupPolicy.java
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Dynamically determines the {@link RejectedExecutionHandler} implementation to use according to the {@link ThreadGroup} of the
+ * current thread.
+ * 
+ * @see AbortPolicy
+ * @see WaitPolicy
+ * 
+ * @since 4.0
+ */
+public final class ByCallerThreadGroupPolicy implements RejectedExecutionHandler {
```
elrodro83: it's in an internal package (not exposed) in a non-api implementation module

core/src/main/java/org/mule/runtime/core/exception/ErrorTypeLocatorFactory.java
```diff
@@ -51,7 +51,7 @@ public static ErrorTypeLocator createDefaultErrorTypeLocator(ErrorTypeRepository
             .addExceptionMapping(RetryPolicyExhaustedException.class, errorTypeRepository.lookupErrorType(RETRY_EXHAUSTED).get())
             .addExceptionMapping(IOException.class, errorTypeRepository.lookupErrorType(CONNECTIVITY).get())
             .addExceptionMapping(SecurityException.class, errorTypeRepository.lookupErrorType(SECURITY).get())
-            .addExceptionMapping(RejectedExecutionException.class, errorTypeRepository.getErrorType(OVERLOAD).get())
+            .addExceptionMapping(SchedulerBusyException.class, errorTypeRepository.getErrorType(OVERLOAD).get())
```
elrodro83: the only way we can realistically determine an OVERLOAD is when the runtime executors are full, that's why i changed the thrown exception to this, so we can accurately determine those cases.

In the case of unmanaged executors, it would be responsability of the users of those to handle the RejectedExecutionException.

services/scheduler/src/main/java/org/mule/service/scheduler/internal/DefaultSchedulerService.java
```diff
@@ -221,6 +235,14 @@ public Scheduler customScheduler(SchedulerConfig config, int queueSize) {
     return customScheduler;
   }
 
+  private ThreadGroup resolveThreadGroupForCustomScheduler(SchedulerConfig config) {
+    if (config.getWaitDispatchingToBusyScheduler() != null && config.getWaitDispatchingToBusyScheduler()) {
```
elrodro83: if i reverse, the case where the value is null with throw a null pointer exception. It is validated in `DefaultSchedulerServiceTestCase`

core/src/main/java/org/mule/runtime/core/api/scheduler/SchedulerBusyException.java
```diff
@@ -0,0 +1,32 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.api.scheduler;
+
+import org.mule.runtime.api.scheduler.Scheduler;
+
+import java.util.concurrent.RejectedExecutionException;
+
+/**
+ * Exception thrown by a {@link Scheduler} when all of its threads are busy and it cannot accept a new task for execution.
```
elrodro83: only for shutdown the RejectedExcecutionException will be thrown.

services/scheduler/src/test/java/org/mule/service/scheduler/internal/executor/WaitPolicyTestCase.java
elrodro83: added same comment as in WaitPolicy

services/scheduler/src/main/java/org/mule/service/scheduler/internal/executor/WaitPolicy.java
```diff
@@ -36,23 +41,22 @@ public WaitPolicy() {
...
+    this.time = (time < 0 ? MAX_VALUE : time);
     this.timeUnit = timeUnit;
   }
 
+  @Override
   @SuppressWarnings("boxing")
   public void rejectedExecution(Runnable r, ThreadPoolExecutor e) {
     try {
-      if (e.isShutdown()) {
-        throw new RejectedExecutionException("ThreadPoolExecutor is already shut down");
```
dfeist: What happened to the shutdown case?

core/src/main/java/org/mule/runtime/core/api/scheduler/SchedulerConfig.java
```diff
@@ -63,4 +85,27 @@ public SchedulerConfig withName(String schedulerName) {
...
+   * Sets the rejection policy to use when dispatching to a busy {@link Scheduler}.
+   * <p>
+   * This is only applicable for <b>custom</b> {@link Scheduler}s. The policy cannot be changed for the runtime managed
+   * {@link Scheduler}.
+   * 
+   * @see SchedulerBusyException
+   * 
+   * @return the updated configuration
+   */
+  public SchedulerConfig withDispatchingRejectionPolicy(DispatchingRejectionPolicy dispatchingRejectionPolicy) {
```
dfeist: Can we drop "Dispatching" and just use `RejectionPolicy?

services/scheduler/src/main/java/org/mule/service/scheduler/internal/executor/WaitPolicy.java
elrodro83: that isn't needed, it is already handled by the executor itself

core/src/main/java/org/mule/runtime/core/api/scheduler/SchedulerConfig.java
```diff
@@ -16,6 +19,24 @@
 public class SchedulerConfig {
 
   /**
+   * Different values on how to handle the scenario where a task is dispatched to a busy {@link Scheduler}.
+   * <p>
+   * A {@link Scheduler} is considered busy when all of its threads are busy and it cannot accept a new task for execution.
+   */
+  public enum DispatchingRejectionPolicy {
+    /**
+     * The actual rejection policy will depend on the type of scheduler the thread is. For cpu-bound threads it will be
```
dfeist: Given pool types are already well-defined in SchedulerService there is no reason not to call them out specifically here by name to make things even clearer.

core/src/main/java/org/mule/runtime/core/api/scheduler/SchedulerConfig.java
elrodro83: mmm, that may bring some confusion with the rejectionhandler of an executor. i wanted to give it a name that clearly differentiates from it.

PR #3967, opened 2017-01-19
MULE-11343: Policies must be registered eagerly to detect invalid ins…

balbifm:
…tances

### DIFF
--- a/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultApplicationPolicyInstance.java
+++ b/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultApplicationPolicyInstance.java
@@ -12,8 +12,11 @@ import static java.util.Optional.of;
 import static org.mule.runtime.core.config.bootstrap.ArtifactType.APP;
 import static org.mule.runtime.module.deployment.impl.internal.artifact.ArtifactContextBuilder.newBuilder;
 import org.mule.runtime.api.exception.MuleException;
+import org.mule.runtime.api.lifecycle.InitialisationException;
+import org.mule.runtime.api.lifecycle.Lifecycle;
 import org.mule.runtime.core.api.config.MuleProperties;
 import org.mule.runtime.core.api.registry.RegistrationException;
+import org.mule.runtime.core.policy.DefaultPolicyInstance;
 import org.mule.runtime.core.policy.Policy;
 import org.mule.runtime.core.policy.PolicyInstance;
 import org.mule.runtime.core.policy.PolicyParametrization;
@@ -71,9 +74,6 @@ public class DefaultApplicationPolicyInstance implements ApplicationPolicyInstan
     this.classLoaderRepository = classLoaderRepository;
     this.artifactPlugins = artifactPlugins;
     this.extensionModelLoaderRepository = extensionModelLoaderRepository;
-
-    initPolicyContext();
-    initPolicyInstance();
   }
 
   private void initPolicyContext() {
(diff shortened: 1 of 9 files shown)


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultApplicationPolicyInstance.java
```diff
@@ -39,18 +39,18 @@
...
   private final ServiceRepository serviceRepository;
   private final ClassLoaderRepository classLoaderRepository;
   private final List<ArtifactPlugin> artifactPlugins;
   private final ExtensionModelLoaderRepository extensionModelLoaderRepository;
+  private ArtifactContext policyContext;
   private PolicyInstance policyInstance;
 
   /**
-   * Creates a new policy instance
+   * Creates and initialises a new policy instance
```
pablokraan: It would be better to remove the initialization logic from the constructor. Add install/init/start/stop methods as in the ones in DeployableArtifact (maybe refactoring and implementing that interface)

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/DeploymentServiceTestCase.java
```diff
@@ -369,6 +369,13 @@ private static File getResourceFile(String resource) {
                                                                          PROPERTIES_BUNDLE_DESCRIPTOR_LOADER_ID))
           .build());
 
+  private final PolicyFileBuilder brokenPolicyFileBuilder =
```
pablokraan: Make this a local variable in the test method

modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/api/application/ApplicationPolicyManager.java
```diff
@@ -16,7 +16,7 @@
 public interface ApplicationPolicyManager {
 
   /**
-   * Add a new policy
+   * Adds and initialises a new policy
```
pablokraan: No need to update this javadoc regarding the initialization. But it should return a boolean indicating if it was added successfully or not, throwing an exception is not good

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultApplicationPolicyInstance.java
```diff
@@ -39,18 +42,18 @@
...
```
pablokraan: Remove "and initialises"

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultApplicationPolicyInstance.java
```diff
@@ -74,31 +77,29 @@ public DefaultApplicationPolicyInstance(Application application, PolicyTemplate
...
+    artifactBuilder.withServiceConfigurator(customizationService -> customizationService
+        .overrideDefaultServiceImpl(MuleProperties.OBJECT_POLICY_MANAGER_STATE_HANDLER,
+                                    application.getMuleContext().getRegistry()
+                                        .lookupObject(MuleProperties.OBJECT_POLICY_MANAGER_STATE_HANDLER)));
+
+    try {
+      policyContext = artifactBuilder.build();
+      policyContext.getMuleContext().start();
+    } catch (MuleException e) {
+      throw new IllegalStateException("Cannot create artifact context for the policy instance", e);
```
pablokraan: Would be better to throw an exception more related to the problem, like a lifecycle exception or similar

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultApplicationPolicyInstance.java
```diff
@@ -145,6 +139,12 @@ public int getOrder() {
   }
 
   @Override
+  public void initialise() {
```
pablokraan: Check that policy was not already initialised

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultApplicationPolicyInstance.java
```diff
@@ -145,6 +143,17 @@ public int getOrder() {
   }
 
   @Override
+  public void initialise() throws InitialisationException {
+    if (policyContext == null) {
```
pablokraan: Seems to me that there is a race condition If two threads call this method

PR #3969, opened 2017-01-19
MULE-11257: Remove Event uses from HTTP extension - Clean up most use…

afelisatti:
…s excluding auth related ones


compatibility/modules/cxf/src/test/resources/http-security-conf-flow-httpn.xml
```diff
@@ -39,7 +39,7 @@
     
     <flow name="secure">
         <httpn:listener path="/services/Echo" config-ref="listenerConfig" allowedMethods="POST" />
-        <httpn:basic-security-filter realm="mule-realm"/>
+        <httpn:basic-security-filter realm="mule-realm" attributes="#[message.attributes]"/>
```
elrodro83: can we infer the attributes? as this is here, it will be very disruptive for the migration

afelisatti: You're right. I'll add this variable to MEL so the default expression works regardless of the EL.

PR #3973, opened 2017-01-19
MULE-11479: HTTP requester ignores socket properties connect timeout

afelisatti:


afelisatti:
No, and I could find a way to force a connect timeout.

PR #3975, opened 2017-01-20
MULE-11307: Separate OAuth extension from the implementation

elrodro83:
* Refactor out Event usages


extensions/oauth/src/main/java/org/mule/extension/oauth2/internal/DeferredExpressionResolver.java
```diff
@@ -29,4 +36,23 @@ public DeferredExpressionResolver(MuleContext muleContext) {
...
+      return null;
+    } else if (!expr.getExpression().isPresent()
+        || !muleContext.getExpressionManager().isExpression(expr.getExpression().get())) {
+      return expr.resolve();
+    } else {
+      BindingContext resultContext = BindingContext.builder()
+          .addBinding("payload",
+                      new TypedValue(result.getOutput(), DataType.builder().fromObject(result.getOutput())
+                          .mediaType(result.getMediaType().orElse(ANY)).build()))
+          .addBinding("attributes",
```
afelisatti: I don't think it will be necessary but the dataType is also mapped now.

extensions/oauth/src/main/java/org/mule/extension/oauth2/internal/MuleEventLogger.java
```diff
@@ -19,27 +17,19 @@
...
+   * Logs the result message payload type and the payload as string.
    * 
-   * @param muleEvent event to log.
+   * @param result result to log.
    */
-  public void logContent(Event muleEvent) {
-    logger.error("Message content type is " + muleEvent.getMessage().getPayload().getDataType().getType());
-    logger.error("Message content is " + muleEvent.getMessage());
-    try {
-      String payloadAsString = muleContext.getTransformationService().getPayloadForLogging(muleEvent.getMessage());
```
afelisatti: Either keep the toString thing or update the javadoc

elrodro83: will update the javadoc

PR #3982, opened 2017-01-23
MULE-11514: Remove org.mule.runtime.core.construct.Flow from Mule API

pablokraan:
_ Renaming org.mule.runtime.core.construct.Flow to org.mule.runtime.core.construct.DefaultFlow
_ Extracting org.mule.runtime.core.api.construct.Flow from org.mule.runtime.core.construct.DefaultFlow
_ Adding org.mule.runtime.core.api.construct.FlowBuilder to hide Flow implementation
_ Removing org.mule.runtime.core.construct from exported packages
_ Moving org.mule.runtime.core.construct to org.mule.runtime.core.internal.construct package


core/src/main/java/org/mule/runtime/core/api/construct/FlowBuilder.java
```diff
@@ -0,0 +1,148 @@
...
+    return this;
+  }
+
+  /**
+   * Configures the exception listener to manage exceptions thrown on the flow execution.
+   *
+   * @param exceptionListener exception listener to use on the flow. Non null.
+   * @return same builder instance
+   */
+  public FlowBuilder managingExceptionsWith(MessagingExceptionHandler exceptionListener) {
```
dfeist: `handleErrorsWith`?  I think handle is clearer than manage, also we'd said we'd start talking about errors rather than exceptions in Mule and seen in other work done for Mule 4 e.g. `OnErrorContinue`

core/src/main/java/org/mule/runtime/core/api/construct/FlowBuilder.java
```diff
@@ -0,0 +1,148 @@
...
+  private ProcessingStrategyFactory processingStrategyFactory;
+  private DefaultFlow flow;
+
+  /**
+   * Creates a new builder
+   *
+   * @param name name of the flow to be created. Non empty.
+   * @param muleContext context where the flow will be registered. Non null.
+   */
+  public FlowBuilder(String name, MuleContext muleContext) {
```
dfeist: Can we use the same pattern as with Event/Message and others, where you create builder from `Flow.builder()` or `Flow.builder(name, muleContet)`?

core/src/main/java/org/mule/runtime/core/internal/construct/DefaultFlow.java
```diff
@@ -39,17 +36,11 @@
...
- * <li>Implements MessagePorcessor allowing direct invocation of the pipeline</li>
- * <li>Supports the optional configuration of a {@link ProcessingStrategy} that determines how message processors are processed.
- * </ul>
+ * Default implementation of {@link org.mule.runtime.core.api.construct.Flow}
  */
-public class Flow extends AbstractPipeline implements Processor {
+public class DefaultFlow extends AbstractPipeline implements org.mule.runtime.core.api.construct.Flow {
 
-  public Flow(String name, MuleContext muleContext) {
+  public DefaultFlow(String name, MuleContext muleContext) {
```
dfeist: Shouldn't this visibility be reduced to ideally force use of builder. This is what we did with Message/Event and others.

core/src/main/java/org/mule/runtime/core/routing/filters/ExpressionFilter.java
```diff
@@ -81,7 +82,7 @@ public boolean accept(InternalMessage message, Event.Builder builder) {
     }
 
     // TODO MULE-9341 Remove Filters. Expression filter will be replaced by something that uses MuleEvent.
-    Flow flowConstruct = new Flow("", muleContext);
+    Flow flowConstruct = new DefaultFlow("", muleContext);
```
dfeist: Why is builder used in some places, but constructor in others?

core/src/main/java/org/mule/runtime/core/api/construct/Flow.java
```diff
@@ -0,0 +1,28 @@
...
+ * <p/>
+ * A flow adds the following behaviour to a {@link Pipeline}
+ * <ul>
+ * <li>Rejects inbound events when Flow is not started</li>
+ * <li>Gathers statistics and processing time data</li>
+ * <li>Implements MessageProcessor allowing direct invocation of the pipeline</li>
+ * <li>Supports the optional configuration of a {@link ProcessingStrategy} that determines how message processors are processed.
+ * </ul>
+ */
+public interface Flow extends AnnotatedObject, FlowConstruct, Lifecycle, Pipeline, Processor {
```
dfeist: No need to implement FlowConstruct if you implement Pipeline.

core/src/main/java/org/mule/runtime/core/api/construct/FlowBuilder.java
```diff
@@ -0,0 +1,148 @@
...
+    return this;
+  }
+
+  /**
+   * Configures the factory used to create processing strategies on the created flow.
+   *
+   * @param processingStrategyFactory factory to create processing strategies. Non null.
+   * @return same builder instance.
+   */
+  public FlowBuilder creatingProcessingStrategiesWith(ProcessingStrategyFactory processingStrategyFactory) {
```
dfeist: Name is too long, why not just  `processingStrategyFactory()`?

core/src/main/java/org/mule/runtime/core/api/construct/FlowBuilder.java
```diff
@@ -0,0 +1,148 @@
...
+    this.muleContext = muleContext;
+  }
+
+  /**
+   * Configures the message source for the flow.
+   *
+   * @param messageSource message source to use. Non null.
+   * @return same builder instance.
+   */
+  public FlowBuilder withMessageSource(MessageSource messageSource) {
```
dfeist: Note that in the other new stuff we've done in Mule 4, we've avoided the fluent prefixes and just used the name of what's being set e.g. `Event.builder().message(foo).build()`. Shouldn't we either do the same here or refactor existing work to be consistent?

core/src/main/java/org/mule/runtime/core/api/construct/FlowBuilder.java
pablokraan: I'll take a look

core/src/main/java/org/mule/runtime/core/api/construct/FlowBuilder.java
pablokraan: I'll use the same format used on the other builders

core/src/main/java/org/mule/runtime/core/api/construct/FlowBuilder.java
pablokraan: This comment is not applicable given that I will remove the fluent API

core/src/main/java/org/mule/runtime/core/api/construct/FlowBuilder.java
pablokraan: same

core/src/main/java/org/mule/runtime/core/internal/construct/DefaultFlow.java
pablokraan: I'll take a look

core/src/main/java/org/mule/runtime/core/routing/filters/ExpressionFilter.java
pablokraan: I did not think that was required to force the usage of the builder, but I can change them (related to previous comment)

core/src/main/java/org/mule/runtime/core/api/construct/Flow.java
pablokraan: good catch

core/src/main/java/org/mule/runtime/core/internal/construct/DefaultFlowBuilder.java
```diff
@@ -0,0 +1,297 @@
...
+  protected final void checkImmutable() {
+    if (flow != null) {
+      throw new IllegalStateException("Cannot change attributes once the flow was built");
+    }
+  }
+
+  /**
+   * Default implementation of {@link Flow}
+   */
+  public static class DefaultFlow extends AbstractPipeline implements Flow {
```
dfeist: Isn't it best for tests to extend Flow to create a builder and create they own implementation of flow that way?  This would mean this class could be protected.

core/src/main/java/org/mule/runtime/core/routing/filters/RegExFilter.java
```diff
@@ -75,7 +74,7 @@ public RegExFilter(String pattern, String value, int flags) {
   @Override
   public boolean accept(InternalMessage message, Event.Builder builder) {
     // TODO MULE-9341 Remove Filters that are not needed
-    Flow flowConstruct = new Flow("", muleContext);
+    Flow flowConstruct = Flow.builder("", muleContext).build();
```
dfeist: Worth replacing this for constant FLOW_NAME?

core/src/main/java/org/mule/runtime/core/internal/construct/DefaultFlowBuilder.java
pablokraan: Flow subclasses override methods from the DefaultFlow class. I prefer to do any extra change in a separated PR in order to minimize the changes that affect MUnit (they subclass our flow too).

PR #3984, opened 2017-01-23
MULE-9649: Check that different application plugins do not export the…

pablokraan:
… same java packages

_ Validating exported packages on application plugins

### DIFF
--- a/modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/internal/plugin/BundlePluginDependenciesResolver.java
+++ b/modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/internal/plugin/BundlePluginDependenciesResolver.java
@@ -20,9 +20,11 @@ import org.mule.runtime.module.artifact.descriptor.ClassLoaderModel.ClassLoaderM
 
 import java.io.File;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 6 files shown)


modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/internal/plugin/BundlePluginDependenciesResolver.java
```diff
@@ -92,6 +102,29 @@ public BundlePluginDependenciesResolver(ArtifactDescriptorFactory<ArtifactPlugin
...
+          exportedPackages.put(packageName, exportedOn);
+        } else {
+          error = true;
+        }
+        exportedOn.add(plugin.getName());
+      }
+    }
+
+    if (error) {
+      throw new DuplicateExportedPackageException(exportedPackages);
```
fernandezlautaro: I believe until [MULE-11464](https://www.mulesoft.org/jira/browse/MULE-11464) is fixed, this PR will break the build as some connectors export other plugins' packages.
Such as the HTTP with Sockets, or anyone relying on the file-extension-common

pablokraan: Is correct that some plugins export the same package which is exported on another plugin, but that won't break the build.
The thing is that the case you mentioned, HTTP exporting the same packages as Sockets, is already managed during the deployment on a "sanitization" process. Every package exported on a plugin A that is also exported on another B, when A depends on B, will be removed from the exported packages of A

fernandezlautaro: I forgot about the sanitization process.
👍

PR #3992, opened 2017-01-23
MULE-11498: don't append "." to paths for listing in SFTP connector

alepulver:


afelisatti:
Is there any way to test this? A mock server or something?

PR #3995, opened 2017-01-24
MULE-10253: support for custom types in smart connectors

fernandezlautaro:
Current smart connectors will have the possibility to define custom types for any input and output parameter.

### DIFF
--- a/modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/loader/XmlExtensionLoaderDelegate.java
+++ b/modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/loader/XmlExtensionLoaderDelegate.java
@@ -11,6 +11,7 @@ import static java.lang.String.format;
 import static java.lang.Thread.currentThread;
 import static java.util.Optional.empty;
 import static java.util.Optional.of;
+import static java.util.Optional.ofNullable;
 import static org.apache.commons.lang3.StringUtils.isEmpty;
 import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
 import static org.mule.runtime.dsl.api.component.config.ComponentIdentifier.Builder;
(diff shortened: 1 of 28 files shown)


modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/loader/XmlExtensionLoaderDelegate.java
```diff
@@ -71,14 +72,23 @@
   protected static final String CONFIG_NAME = "config";
 
   private static final ClassTypeLoader typeLoader = ExtensionsTypeLoaderFactory.getDefault().createTypeLoader();
-  private static final Map<String, MetadataType> types = ImmutableMap.<String, MetadataType>builder()
+  private static final Map<String, MetadataType> defaultInputTypes = ImmutableMap.<String, MetadataType>builder()
```
pablolagreca: Create a variable with the common set of types and the reuse it for defaultInputTypes and defaultOutputTypes

modules/extensions-xml-support/src/test/java/org/mule/runtime/extension/internal/loader/catalog/TypesCatalogLoaderTest.java
```diff
@@ -0,0 +1,32 @@
...
+import org.junit.Test;
+import org.mule.runtime.extension.internal.loader.catalog.loader.xml.TypesCatalogXmlLoader;
+import org.mule.runtime.extension.internal.loader.catalog.model.TypesCatalog;
+
+import java.io.File;
+
+/**
+ * TODO(fernandezlautaro: MULE-11501 this class must be moved to a separate module
+ */
+public class TypesCatalogLoaderTest {
```
pablolagreca: Must end with TestCase and extends AbstractMuleTestCase

modules/extensions-xml-support/src/test/resources/module-custom-types/module-custom-types-catalog.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
pablolagreca: This file has the same content as module-param-custom-types-catalog.xml. Can you use just one of them?

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/loader/XmlExtensionLoaderDelegate.java
fernandezlautaro: done

modules/extensions-xml-support/src/test/java/org/mule/runtime/extension/internal/loader/catalog/TypesCatalogLoaderTest.java
fernandezlautaro: done

modules/extensions-xml-support/src/test/resources/module-custom-types/module-custom-types-catalog.xml
fernandezlautaro: I'll change the one used in module-param-custom-types-catalog.xml, as using the same file will too much of a trouble for the parameterized test class.

PR #4010, opened 2017-01-26
MULE-11518: MuleDeployment Junit Rule: Add the possibility to pass pr…

anosenzo:
…operties that are going to be evaluated in the before method of the Rule

### DIFF
--- a/tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleDeployment.java
+++ b/tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleDeployment.java
@@ -25,6 +25,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleDeployment.java
```diff
@@ -85,6 +85,12 @@
   private List<String> libraries = new ArrayList<>();
   private MuleProcessController mule;
   private Map<String, String> properties = new HashMap<>();
+  private Map<String, StringPropertySupplier> propertiesUsingLambdas = new HashMap<>();
```
gsfernandes: You could just use a 
```Supplier<String>``` instead of creating your own supplier interface

anosenzo: Done

PR #4011, opened 2017-01-26
MULE-11430: Decouple validations module from MuleEvent and MuleMessage

pablomusumeci:


extensions/validation/src/main/java/org/mule/extension/validation/api/ObjectSource.java
```diff
@@ -48,27 +48,18 @@
...
   @Parameter
   @Optional
-  private String ref;
+  @XmlHints(allowInlineDefinition = false)
+  private T ref;
 
   public ObjectSource() {}
 
-  public ObjectSource(String type, String ref) {
+  public ObjectSource(String type, T ref) {
```
marianogonzalez: This should no longer be a T. This should explicitly be a validator.

extensions/validation/src/main/java/org/mule/extension/validation/api/ObjectSource.java
```diff
@@ -48,27 +48,18 @@
...
-    checkArgument(!(hasType && hasRef), "type and ref attributes are mutually exclusive. Please provide only one of them");
-    checkArgument(hasType ^ hasRef, "One of class or ref attributes are required. Please provide one of them");
-
-    if (hasRef) {
-      return doGetByRef(muleContext);
-    } else {
-      return doGetByClassName();
-    }
+  public final T getObject() {
+    return !isBlank(type) ? doGetByClassName() : ref;
```
marianogonzalez: doGetByClassName() should cache the object it returns and not return a new one each time. The getRegistry().applyProcessors() thing should be one in here one the instance is first created and not all the time it is used. That means that doGetByClassName should be synchronized and that this method should receive the muleContext

extensions/validation/src/main/java/org/mule/extension/validation/internal/CustomValidatorOperation.java
```diff
@@ -62,19 +63,24 @@ protected void logSuccessfulValidation(Validator validator) {
...
     }
 
     @Override
     protected Validator doGetByClassName() {
       return class2ValidatorCache.getUnchecked(this);
     }
 
-    private Validator createValidator() {
-      return super.doGetByClassName();
+    private synchronized Validator createValidator() throws MuleException {
```
marianogonzalez: not needed

extensions/validation/src/main/java/org/mule/extension/validation/internal/CustomValidatorOperation.java
```diff
@@ -62,19 +63,24 @@ protected void logSuccessfulValidation(Validator validator) {
     }
   }
 
-  private class ValidatorSource extends ObjectSource<Validator> {
+  private class ValidatorSource extends ObjectSource {
```
marianogonzalez: ValidatorSource no longer needed. Just use ObjectSource

PR #4016, opened 2017-01-26
MULE-10362: provide doc support for smart connectors

fernandezlautaro:
Connectors in XML now expose the category, vendor and documentation.

### DIFF
--- a/modules/spring-config/src/main/resources/META-INF/mule-module.xsd
+++ b/modules/spring-config/src/main/resources/META-INF/mule-module.xsd
@@ -70,7 +70,7 @@
     <xsd:element name="parameter" type="parameterType">
         <xsd:annotation>
             <xsd:documentation>
-                An named element that will be consumed only in the scope of the body element by the list of message processors.
+                A parameter element defines an input value for the operation in which it is define. Such parameter must be defined with a meaningful name, a type which defines the kind of content the parameter must have and optionally a default value that will be used if the invocation to the operation does not defines a value for the parameter. The parameter can be accessed within the body definition of the operation using an expression such as #[mel: param.paramName]
             </xsd:documentation>
         </xsd:annotation>
     </xsd:element>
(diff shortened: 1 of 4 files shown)


modules/spring-config/src/main/resources/META-INF/mule-module.xsd
```diff
@@ -87,6 +67,58 @@
         </xsd:complexContent>
     </xsd:complexType>
 
+    <xsd:element name="parameter" type="parameterType">
+        <xsd:annotation>
+            <xsd:documentation>
+                An named element that will be consumed only in the scope of the body element by the list of message processors.
```
pablolagreca: Doesn't seem to accurate this description.

modules/spring-config/src/main/resources/META-INF/mule-module.xsd
```diff
@@ -133,7 +154,35 @@
...
+                    <xsd:annotation>
+                        <xsd:documentation>
+                            Expected category of the module.
+                        </xsd:documentation>
+                    </xsd:annotation>
+                </xsd:attribute>
+                <xsd:attribute name="vendor" type="xsd:string" default="MuleSoft">
+                    <xsd:annotation>
+                        <xsd:documentation>
+                            Expected category of the module.
```
pablolagreca: wrong doc

modules/spring-config/src/main/resources/META-INF/mule-module.xsd
fernandezlautaro: why not?
a `<parameter>` is the only visible thing any message processor within the `<body>` element can consume as they don't see the event going through

Mind recommending me a different doc for it?

modules/spring-config/src/main/resources/META-INF/mule-module.xsd
fernandezlautaro: done

modules/spring-config/src/main/resources/META-INF/mule-module.xsd
pablolagreca: A parameter element defines an input value for the operation in which it is define. Such parameter must be defined with a meaningful name, a type which defines the kind of content the parameter must have and optionally a default value that will be used if the invocation to the operation does not defines a value for the parameter. The parameter can be accesed within the body definition of the operation using #[..lalalaa..].

PR #4019, opened 2017-01-26
MULE-11535: return 400 instead of 500 when invalid URL encoding in tr…

alepulver:
…ansport

### DIFF
--- a/modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerUriEncodingErrorTestCase.java
+++ b/modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerUriEncodingErrorTestCase.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.module.http.functional.listener;
+
+import static java.lang.String.format;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertThat;
+import static org.mule.module.http.api.HttpConstants.HttpStatus.BAD_REQUEST;
+import static org.mule.module.http.api.HttpConstants.HttpStatus.OK;
+import static org.mule.module.http.api.HttpConstants.ResponseProperties.HTTP_STATUS_PROPERTY;
+import static org.mule.module.http.api.client.HttpRequestOptionsBuilder.newOptions;
+
+import org.mule.api.MuleMessage;
+import org.mule.tck.junit4.FunctionalTestCase;
+import org.mule.tck.junit4.rule.DynamicPort;
+
+import org.junit.Rule;
+import org.junit.Test;
+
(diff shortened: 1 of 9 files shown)


modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerUriEncodingErrorTestCase.java
```diff
@@ -0,0 +1,61 @@
...
+public class HttpListenerUriEncodingErrorTestCase extends FunctionalTestCase
+{
+    @Rule
+    public DynamicPort dynamicPort = new DynamicPort("port");
+
+    @Test
+    public void failsWithAppropriateError() throws Exception
+    {
+        String address = getUri() + "?blah=badcode%2";
+        MuleMessage response = muleContext.getClient().send(address, getTestMuleMessage(), newOptions().disableStatusCodeValidation().build());
```
afelisatti: Can you avoid the mule client? It would be better to cherry-pick to 4.x.

modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerUriEncodingErrorTestCase.java
```diff
@@ -0,0 +1,61 @@
...
+
+        assertThat(response.<Integer>getInboundProperty(HTTP_STATUS_PROPERTY), is(BAD_REQUEST.getStatusCode()));
+        assertThat(response.getPayloadAsString(), containsString("URLDecoder"));
+    }
+
+    @Test
+    public void worksWhenValidUri() throws Exception
+    {
+        String address = getUri() + "?blah=a%20space";
+        MuleMessage response = muleContext.getClient().send(address, getTestMuleMessage());
```
afelisatti: Same as above. Avoiding the mule client will give you less headaches.

transports/http/src/main/java/org/mule/transport/http/HttpMessageProcessTemplate.java
```diff
@@ -192,13 +193,20 @@ private MessagingException getExceptionForCreatingFailureResponse(MessagingExcep
     @Override
     public void afterFailureProcessingFlow(MuleException exception) throws MuleException
     {
+        // MULE-11535: unwrap errors thrown in the initialization phase so that the true cause can be used to generate status codes
+        Throwable error = exception;
+        if (error instanceof DefaultMuleException && error.getCause() instanceof CreateException)
```
afelisatti: Why is the DefaultMuleException check required? Isn't enough to do a null check on the causes and the CreateException one?

afelisatti: Also, is it possible that cause.cause is null?

transports/http/src/main/java/org/mule/transport/http/HttpMuleMessageFactory.java
```diff
@@ -165,8 +168,16 @@ else if (transportMessage instanceof HttpMethod)
...
-        
-        queryParameters.put(HttpConnector.HTTP_QUERY_PARAMS, processQueryParams(uri, encoding));
+
+        try
+        {
+            queryParameters.put(HttpConnector.HTTP_QUERY_PARAMS, processQueryParams(uri, encoding));
+        }
+        catch (IllegalArgumentException e)
+        {
+            Message errorDescription = MessageFactory.createStaticMessage(e.getMessage());
```
afelisatti: Shouldn't you use a custom message saying the query parameters were invalid? The exception message will still be available.

PR #4022, opened 2017-01-26
MULE-11539: No content response includes Transfer-Encoding header

afelisatti:


bobvanderlinden:
Should this also apply for 304? Currently I'm seeing a 304 being passed through a Mule server that results in a 304 in the response with a `Transfer-Encoding: chunked` followed by `0\r\n` in the HTTP body. This seems to cause issues with proxies that presume 304 has no response.

afelisatti:
Which version of Mule are you using? I think Grizzly was handling 304.

PR #4023, opened 2017-01-26
MULE-11467: Mule API Clean up

pablokraan:


compatibility/tests/integration-transports/src/test/java/org/mule/compatibility/config/spring/parsers/specific/ComponentWithBindingsDefinitionParserFlowTestCase.java
```diff
@@ -31,13 +31,13 @@
...
 import org.mule.runtime.core.component.AbstractJavaComponent;
 import org.mule.runtime.core.context.DefaultMuleContextFactory;
-import org.mule.runtime.core.model.resolvers.ArrayEntryPointResolver;
-import org.mule.runtime.core.model.resolvers.CallableEntryPointResolver;
-import org.mule.runtime.core.model.resolvers.DefaultEntryPointResolverSet;
-import org.mule.runtime.core.model.resolvers.ExplicitMethodEntryPointResolver;
-import org.mule.runtime.core.model.resolvers.MethodHeaderPropertyEntryPointResolver;
-import org.mule.runtime.core.model.resolvers.NoArgumentsEntryPointResolver;
-import org.mule.runtime.core.model.resolvers.ReflectionEntryPointResolver;
+import org.mule.runtime.core.api.model.resolvers.ArrayEntryPointResolver;
```
dfeist: Do these need to be in API in order to be used by XML parser? Otherwise why are these API?  Also these may need to go in compatibility module after further cleanup.

pablokraan: Yes, must be API because of the parsers. This is a first iteration, parsers must be updated to there is no need to expose these classes

dfeist: If this is temporary, then are you keeping track of all the package that need to be moved out of API later?  Isn't it better to moved them all a specific `apiconfig` package or something.  Up to you, just as long as we don't get any leftovers in API after.

PR #4025, opened 2017-01-26
MULE-11541: Move HttpConstants and HttpHeaders to http-service-api

elrodro83:


modules/http-api/src/main/java/org/mule/service/http/api/HttpConstants.java
```diff
@@ -0,0 +1,181 @@
...
+    public String getScheme() {
+      return scheme;
+    }
+
+    public int getDefaultPort() {
+      return defaultPort;
+    }
+  }
+
+  public static final class RequestProperties {
```
afelisatti: This should not go here but rather in the transport constants since they refer to the inbound/outbound properties handled by the HTTP module and transport, in fact they are probably only needed in the compatibility module.

modules/http-api/src/main/java/org/mule/service/http/api/HttpConstants.java
```diff
@@ -0,0 +1,181 @@
...
+    public static final String HTTP_REQUEST_URI = HTTP_PREFIX + "request.uri";
+    public static final String HTTP_REMOTE_ADDRESS = HTTP_PREFIX + "remote.address";
+    public static final String HTTP_LISTENER_PATH = HTTP_PREFIX + "listener.path";
+    public static final String HTTP_SCHEME = HTTP_PREFIX + "scheme";
+    public static final String HTTP_CLIENT_CERTIFICATE = HTTP_PREFIX + "client.cert";
+    public static final String HTTP_DISABLE_STATUS_CODE_EXCEPTION_CHECK = HTTP_PREFIX + "disable.status.code.exception.check";
+  }
+
+  public static final class ResponseProperties {
+
```
afelisatti: Same as above, this regards inbound/outbound properties.

modules/http-api/src/main/java/org/mule/service/http/api/HttpConstants.java
```diff
@@ -0,0 +1,181 @@
...
+    public static final String HTTP_STATUS_PROPERTY = RequestProperties.HTTP_STATUS_PROPERTY;
+    public static final String HTTP_REASON_PROPERTY = RequestProperties.HTTP_PREFIX + "reason";
+  }
+
+  public static final String ALL_INTERFACES_IP = "0.0.0.0";
+
+  public enum HttpStatus {
+    CONTINUE(100, "Continue"),
```
afelisatti: Is this the proper way to bypass the formatter? Good to know 👍

PR #4027, opened 2017-01-27
MULE-11218 - DataSense not working for Database insert + select in th…

pablolagreca:
…e same flow


modules/spring-config/src/main/java/org/mule/runtime/config/spring/dsl/model/MinimalApplicationModelGenerator.java
```diff
@@ -88,17 +93,43 @@ public ApplicationModel getMinimalModelByName(String name) {
...
+      String part = parts[i];
+      switch (part) {
+        case "source": {
+          ComponentModel sourceComponentModel = flowModel.getInnerComponents().get(0);
+          checkState(areMatchingTypes(MessageSource.class, sourceComponentModel.getType()),
+                     format(
+                            "Flow path is pointing to the message source of flow %s but it seems that the flow does not have a message source, component identifier of first element is %s",
+                            flowModel.getNameAttribute(), sourceComponentModel.getIdentifier()));
+          Iterator<ComponentModel> iterator = flowModel.getInnerComponents().iterator();
           iterator.next();
```
gsfernandes: Maybe a comment on this line would help why the next is being done before the hasNext.

PR #4043, opened 2017-01-30
MULE-11546: DSL of Map with POJO not generated correctly

alegmarra:


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/schema/builder/CollectionSchemaDelegate.java
```diff
@@ -80,7 +81,7 @@ private LocalComplexType generateCollectionComplexType(DslElementSyntax collecti
        */
       @Override
       public void visitObject(ObjectType objectType) {
-        if (isMap(objectType)) {
+        if (isMap(objectType) && !isInfrastructure(objectType)) {
```
marianogonzalez: remove the second clause in case of buy-in of my comment in the other PR

PR #4044, opened 2017-01-30
MULE-11564: adding password support for <parameter> and <properties>

fernandezlautaro:


modules/extensions-xml-support/src/test/resources/module-documentation/module-documentation.xml
```diff
@@ -12,10 +12,12 @@
            http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd">
 
     <property name="aPropertyWithDoc" type="string" doc:description="Documentation for the property"/>
+    <property name="aHiddenPropertyWithDoc" type="string" password="true" doc:description="Documentation for the hidden property"/>
 
     <operation name="operation-with-doc" doc:description="Documentation for the operation">
         <parameters>
             <parameter name="paramWithDoc" type="string" doc:description="Documentation for the parameter" />
+            <parameter name="hiddenParamWithDoc" type="string" password="true" doc:description="Documentation for the hidden parameter" />
```
pablolagreca: Do you foresee any other type of parameter other than "password"? Like complex parameters, or email parameter or anything like that?. Because in case there may appear other parameters types in the future then it may be better to have different "parameter" elements. Just think about it and let's discuss it offline if necessary.

pablolagreca: Another thing. Review how this is approached in the SDK with annotations and if there are other types for parameters so we align the model of the XML with the model of the SDK if it makes sense.

fernandezlautaro: There might be other types of elements in the future, but TBH I'd say the missing attributes today are ralated to display things, such as: friendly name, tab name, order, and so on.

About complex types of parameters, like _email parameter_ are really ahead in the road and I'd not pay attention at this time.

PR #4047, opened 2017-01-30
MULE-11429: Auto Generate extensions documentation

juandesi:


extensions/pom.xml
```diff
@@ -18,8 +18,10 @@
     <description>Mule extensions that augment the core Mule functionality through the Mule Extension API</description>
 
     <properties>
+        <generateDocs>false</generateDocs>
         <mule.extensions.maven.plugin.version>1.0.0-SNAPSHOT</mule.extensions.maven.plugin.version>
         <mule.metadata.version>1.0.0-SNAPSHOT</mule.metadata.version>
+        <extension.documentation.maven.plugin.version>1.0.0-SNAPSHOT</extension.documentation.maven.plugin.version>
```
marianogonzalez: a) this should be in a constant
b) didn't you talked to me today about this being part of the extensions packager?

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/description/ExtensionDescriptionDocumenter.java
```diff
@@ -69,12 +69,17 @@ private void documentConfigurations(ExtensionDeclaration declaration, TypeElemen
   }
 
   private Optional<ConfigurationDeclaration> findMatchingConfiguration(ExtensionDeclaration declaration, TypeElement element) {
-    return declaration.getConfigurations().stream()
+    List<ConfigurationDeclaration> configurations = declaration.getConfigurations();
+    if (configurations.size() == 1) {
+      return Optional.of(configurations.get(0));
```
marianogonzalez: why doesn't the naming logic apply for this case too?

extensions/pom.xml
juandesi: a) it's a defined property, what do you mean by constant?
b) I've meant to move the code to the mule-extensions-maven-plugin.

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/description/ExtensionDescriptionDocumenter.java
juandesi: if you have only one config you don't need to match anything.

PR #4052, opened 2017-01-31
MULE-11529: Define interception API

elrodro83:
* Fix newly introduced flaky tests

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/processor/chain/AbstractMessageProcessorChain.java
+++ b/core/src/main/java/org/mule/runtime/core/processor/chain/AbstractMessageProcessorChain.java
@@ -45,6 +45,7 @@ import org.mule.runtime.core.util.StringUtils;
 
 import java.util.ArrayList;
 import java.util.Iterator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mule/runtime/core/processor/chain/AbstractMessageProcessorChain.java
```diff
@@ -121,14 +104,47 @@ protected Event doProcess(Event event) throws MuleException {
...
+
     if (flowConstruct instanceof Pipeline) {
-      interceptors.add(0, (processor, next) -> ((Pipeline) flowConstruct).getProcessingStrategy().onProcessor(processor, next));
+      interceptors.add((processor, next) -> ((Pipeline) flowConstruct).getProcessingStrategy().onProcessor(processor, next));
     }
 
-    List<ProcessorInterceptor> interceptionHandlerChain =
-        muleContext.getProcessorInterceptorManager().getInterceptors();
+    // Handle errors
+    interceptors.add((processor, next) -> stream -> from(stream)
```
dfeist: Does it not make sense to keep these 4 as state and add the flowConstruct one to local state rather than instance state in apply()

dfeist: Or alternatively pull this out into a method to reduce size of apply().

PR #4054, opened 2017-01-31
MULE-11420: TypedValue and ParameterResolver can't be used inside a P…

estebanwasinger:
…ojo annotated with @Nullsafe and Content

### DIFF
--- a/modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/typed/value/AbstractTypedValueTestCase.java
+++ b/modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/typed/value/AbstractTypedValueTestCase.java
@@ -20,6 +20,8 @@ import org.mule.runtime.extension.api.annotation.Extension;
 import org.mule.runtime.extension.api.annotation.Operations;
 import org.mule.runtime.extension.api.annotation.Sources;
 import org.mule.runtime.extension.api.annotation.execution.OnSuccess;
+import org.mule.runtime.extension.api.annotation.param.Content;
+import org.mule.runtime.extension.api.annotation.param.NullSafe;
 import org.mule.runtime.extension.api.annotation.param.Optional;
 import org.mule.runtime.extension.api.annotation.param.Parameter;
 import org.mule.runtime.extension.api.annotation.param.UseConfig;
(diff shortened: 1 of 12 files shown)


modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/typed/value/AbstractTypedValueTestCase.java
```diff
@@ -49,9 +52,11 @@ void runAndAssertTypedValue(String flowName, Object payloadValue, MediaType medi
   void assertTypedValue(TypedValue typedValue, Object payloadValue, MediaType mediaType, Charset charset) {
     assertThat(typedValue, is(instanceOf(TypedValue.class)));
     Object value = typedValue.getValue();
-    assertThat(value, is(instanceOf(payloadValue.getClass())));
     assertThat(value, is(payloadValue));
-    assertThat(typedValue.getDataType(), is(like(payloadValue.getClass(), mediaType, charset)));
+    if (value != null) {
```
alegmarra: why is it ok to have a null value? should this be two separate assertions?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ExpressionTypedValueValueResolver.java
```diff
@@ -42,7 +42,7 @@ public ExpressionTypedValueValueResolver(String expression, Class<Object> expect
     } else {
       DataType expectedDataType =
           DataType.builder().type(expectedClass).mediaType(typedValue.getDataType().getMediaType()).build();
-      T result = (T) transform(typedValue, expectedDataType, event);
+      T result = typedValue.getValue() != null ? (T) transform(typedValue, expectedDataType, event) : null;
```
marianogonzalez: it is weird that you check `typedValue.getValue()` to not be null but then transform the entire `typedValue`. It sounds like this is a check that the transform method should do.

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/typed/value/AbstractTypedValueTestCase.java
estebanwasinger: `null` is a valid resolution type, should be handled correctly.

alegmarra: yes, value can be null, but with this assertion we are not testing for expected null vs expected value

PR #4056, opened 2017-01-31
Mule 11571

alepulver:


modules/db/src/test/java/org/mule/module/db/integration/TestDbConfig.java
```diff
@@ -20,7 +20,7 @@
 
     private static boolean USE_DERBY = true;
     private static boolean USE_MYSQL = false;
-    private static boolean USE_ORACLE = false;
+    private static boolean USE_ORACLE = true;
```
pablokraan: Revert this change

modules/db/src/test/java/org/mule/module/db/integration/vendor/oracle/OracleSelectWithTestCase.java
```diff
@@ -0,0 +1,52 @@
...
+    @Override
+    protected String[] getFlowConfigurationResources()
+    {
+        return new String[] {"integration/vendor/oracle/oracle-select-with-config.xml"};
+    }
+
+    @Test
+    public void selectWith() throws Exception
+    {
+        LocalMuleClient client = muleContext.getClient();
```
pablokraan: Would be better if you use flow runner instead of having to declare an inbound endpoint and then using the mule client

PR #4057, opened 2017-01-31
MULE-11583: adding functional tests to ensure smart connectors are va…

fernandezlautaro:
…lidated with the same common-semantic rules as SDK connectors

### DIFF
--- a/modules/extensions-xml-support/src/test/resources/validation/module-repeated-parameters.xml
+++ b/modules/extensions-xml-support/src/test/resources/validation/module-repeated-parameters.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module name="module-repeated-parameters"
+        namespace="http://www.mulesoft.org/schema/mule/module-repeated-parameters"
+        xmlns="http://www.mulesoft.org/schema/mule/module"
+        xmlns:mule="http://www.mulesoft.org/schema/mule/core"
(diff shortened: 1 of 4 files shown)


modules/extensions-xml-support/src/test/resources/validation/module-repeated-parameters.xml
```diff
@@ -0,0 +1,21 @@
...
+        xmlns:mule="http://www.mulesoft.org/schema/mule/core"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="
+           http://www.mulesoft.org/schema/mule/module http://www.mulesoft.org/schema/mule/module/current/mule-module.xsd
+           http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd">
+
+    <operation name="repeated-parameters">
+        <parameters>
+            <parameter name="repeated-parameter" type="string"/>
+            <parameter name="repeated-parameter" type="string"/>
```
alegmarra: check also for parameters with the same name but different type

fernandezlautaro: should it be different output?

fernandezlautaro: I'm not testing *all* validators, just some of them

PR #4055, opened 2017-01-31
MULE-10281: Add error support in HTTP

afelisatti:

PR #4060, opened 2017-01-31
MULE-11470: Make HTTP messages body a TypedValue

afelisatti:


afelisatti:
This and #4055 cannot be merged at once since which ever is merged first will cause the other to require a rebase.

PR #4061, opened 2017-02-01
MULE-11585: smart connector support for @Content-like attributes

fernandezlautaro:


modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/loader/XmlExtensionLoaderDelegate.java
```diff
@@ -80,9 +81,17 @@
   private static final Map<String, MetadataType> defaultOutputTypes = getCommonTypesBuilder()
       .put("void", typeLoader.load(Void.class))
       .build();
+  private static final Map<String, ParameterRole> parameterRoleTypes = ImmutableMap.<String, ParameterRole>builder()
```
pablolagreca: Create an static method to get the enum from an string representation instead of doing this.

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/loader/XmlExtensionLoaderDelegate.java
```diff
@@ -283,16 +292,24 @@ private void extractOperationParameters(OperationDeclarer operationDeclarer, Com
...
     }
   }
 
-  private void extractParameter(ParameterizedDeclarer parameterizedDeclarer, ComponentModel param) {
+  private ParameterRole getRole(ComponentModel param) {
+    return parameterRoleTypes.get(param.getParameters().get(ROLE));
+  }
+
+  private void extractProperty(ParameterizedDeclarer parameterizedDeclarer, ComponentModel param) {
+    extractParameter(parameterizedDeclarer, param, ParameterRole.BEHAVIOUR);
```
pablolagreca: static imports

PR #4063, opened 2017-02-01
MULE-11433: Ignore flacky test

estebanwasinger:


extensions/email/src/test/java/org/mule/extension/email/retriever/IMAPNegativeConnectivityTestCase.java
```diff
@@ -82,6 +83,7 @@ public void configCannotReach() {
     connectivityUtils.assertFailedConnection("configCannotReach", exceptionMatcher, is(errorType(UNKNOWN_HOST)));
   }
 
+  @Ignore
```
afelisatti: Add a reference to an issue about fixing it. Otherwise this will be ignored for ever and you might as well delete it.

PR #4075, opened 2017-02-02
MULE-11591: Improve extensions UX

pablomusumeci:


extensions/db/src/main/java/org/mule/extension/db/api/param/ParameterizedStatementDefinition.java
```diff
@@ -40,7 +40,7 @@
    * The map's values will contain the actual assignation for each parameter.
    */
   @Parameter
-  @Optional(defaultValue = "#[mel:new java.util.HashMap()]")
+  @Optional(defaultValue = "#[{}]")
```
marianogonzalez: Add an `@Example`

PR #4077, opened 2017-02-02
MULE-11313: NPE doing test connection for http request with default v…

estebanwasinger:
…alues

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/utils/ImplicitObjectUtils.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/utils/ImplicitObjectUtils.java
@@ -18,9 +18,7 @@ import org.mule.runtime.module.extension.internal.runtime.resolver.StaticValueRe
 import org.mule.runtime.module.extension.internal.runtime.resolver.TypeSafeExpressionValueResolver;
 import org.mule.runtime.module.extension.internal.runtime.resolver.ValueResolver;
 
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 
 /**
  * Utilities for creating object with implicit values based on a {@link ParameterizedModel}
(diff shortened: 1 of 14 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/utils/ImplicitObjectUtils.java
```diff
@@ -35,13 +42,27 @@ private ImplicitObjectUtils() {}
    */
   public static ResolverSet buildImplicitResolverSet(ParameterizedModel parameterizedModel, MuleContext muleContext) {
     ResolverSet resolverSet = new ResolverSet();
+
+    Map<String, Object> parameterValues = new HashMap<>();
+    for (ParameterModel model : parameterizedModel.getAllParameterModels()) {
```
marianogonzalez: this can be done with a map collector

marianogonzalez: don't do this here. Add a constructor/factory method on ParametersResolver which does this from a ParmeterizedModel

alegmarra:
missing TODO for the complete fix?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/utils/ImplicitObjectUtils.java
estebanwasinger: > this can be done with a map collector
no :( , the collector can't be used if there is null values

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/utils/ImplicitObjectUtils.java
```diff
@@ -40,15 +38,11 @@ private ImplicitObjectUtils() {}
...
     ResolverSet resolverSet = new ResolverSet();
-
-    Map<String, Object> parameterValues = new HashMap<>();
-    for (ParameterModel model : parameterizedModel.getAllParameterModels()) {
-      parameterValues.put(model.getName(), model.getDefaultValue());
-    }
-
-    ParametersResolver parametersResolver = new ParametersResolver(muleContext, parameterValues);
+    ParametersResolver parametersResolver =
+        ParametersResolver.fromParameterizedModelDefaultValues(parameterizedModel, muleContext);
```
marianogonzalez: fromDefaultValues

PR #4082, opened 2017-02-03

pablolagreca:
… each processor

Add new location mechanism
Pending remove old code for setting the processorPath in all the components
More commits will be done under the same issue to complete the task


extensions/http/src/test/java/org/mule/extension/http/policy/HttpListenerPolicyPointcutParametersFactoryTestCase.java
```diff
@@ -43,7 +42,7 @@ public void supportsHttpListener() {
   @Test
   public void doesNotSupportHttpRequester() {
     assertThat(factory
-        .supportsSourceIdentifier(new Builder().withNamespace("http").withName("request").build()),
+        .supportsSourceIdentifier(ComponentIdentifier.builder().withNamespace("http").withName("request").build()),
```
gsfernandes: Static import

pablolagreca: I can't use it because there's already an static import for another buider() function.

PR #4084, opened 2017-02-03
MULE-11439: Different values in a connection don't trigger a new conn…

pablomusumeci:
…ect() for cached providers


tests/http/src/test/java/org/mule/test/module/http/functional/requester/HttpRequestDynamicConfigTestCase.java
```diff
@@ -96,4 +105,34 @@ public void requestsGoThroughClient2() throws Exception {
     assertThat(headers.keys(), hasItem(TRANSFER_ENCODING));
   }
 
+  @Test
+  public void requestWithDynamicConnectionParamUsesDifferentConfigs() throws Exception {
+    expectedError.expectError("HTTPN", HttpError.CONNECTIVITY, ConnectException.class, "Connection refused");
```
afelisatti: Shouldn't this be used right above the second request?

pablomusumeci: This is a test rule, its location has no effect over the code being tested.

PR #4090, opened 2017-02-06
MULE-11432: Disallow operations to receive the current Event and Message

estebanwasinger:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/OperationParametersTypeModelValidator.java
```diff
@@ -0,0 +1,54 @@
...
+
+  @Override
+  public void validate(ExtensionModel extensionModel, ProblemsReporter problemsReporter) {
+    new IdempotentExtensionWalker() {
+
+      @Override
+      protected void onOperation(OperationModel model) {
+        model.getAllParameterModels().forEach(parameterModel -> {
+          if (parameterModel.getType().getMetadataFormat().equals(MetadataFormat.JAVA)) {
+            Class<?> type = getType(parameterModel.getType());
```
alegmarra: I'd avoid getting the Class, instead load the Event and Message MetadataTypes and check if they are equals to the parameter type

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/TypeSafeStaticValueResolver.java
```diff
@@ -0,0 +1,81 @@
...
+import org.mule.runtime.core.api.MuleContext;
+
+/**
+ * {@link ValueResolver} implementation which always returns the same value, but if it is required,
+ * the value will be converted to the required type using the mule transformers.
+ *
+ * @param <T> Type of the value to resolver
+ * @since 4.0
+ */
+public class TypeSafeStaticValueResolver<T> implements ValueResolver<T> {
```
marianogonzalez: This approach seems overkill.. What is the advantage of having a `TypeSafeStaticValueResolver`, plus a `TypeSafeExpressionValueResolver` which both compose a `TypeSafeTransformer` (which is not actually a transformer) vs simply having a `TypeSafeValueResolverDecorator` which can  work with any kind of ValueResolver we may need.

extensions/file/src/main/java/org/mule/extension/file/internal/FileOperations.java
```diff
@@ -150,20 +148,18 @@
    * @param createParentDirectories whether or not to attempt creating any parent directories which don't exists.
    * @param lock                    whether or not to lock the file. Defaults to false
    * @param mode                    a {@link FileWriteMode}. Defaults to {@code OVERWRITE}
-   * @param event                   The current {@link Event}
    * @throws IllegalArgumentException if an illegal combination of arguments is supplied
    */
   @Summary("Writes the given \"Content\" in the file pointed by \"Path\"")
   @Throws(FileWriteErrorTypeProvider.class)
   public void write(@UseConfig FileConnectorConfig config, @Connection FileSystem fileSystem, @Optional String path,
```
marianogonzalez: you need to change the javadoc (and also the spec, but that can wait) ;)

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/OperationParametersTypeModelValidator.java
marianogonzalez: there's no problem in getting the class if you already know that the format is JAVA.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/TypeSafeExpressionValueResolver.java
```diff
@@ -33,55 +27,17 @@
  */
 public class TypeSafeExpressionValueResolver<T> implements ValueResolver<T> {
```
marianogonzalez: why does this class still exists?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ExpressionTypedValueValueResolver.java
```diff
@@ -23,13 +22,15 @@
  * @param <T>
  * @since 4.0
  */
-public class ExpressionTypedValueValueResolver<T> extends TypeSafeExpressionValueResolver<TypedValue<T>> {
+public class ExpressionTypedValueValueResolver<T> extends ExpressionValueResolver<TypedValue<T>> {
```
marianogonzalez: why does this class still exists?

PR #4091, opened 2017-02-06
MULE-11573 Migrate 'synchronous' processing strategy away from using …

dfeist:
…blocking code path.  Optimise to use a stream per source thread.


core/src/main/java/org/mule/runtime/core/processor/strategy/SynchronousProcessingStrategyFactory.java
```diff
@@ -45,4 +59,71 @@ public ProcessingStrategy create(MuleContext muleContext, String schedulersNameP
     return SYNCHRONOUS_PROCESSING_STRATEGY_INSTANCE;
   }
 
+  static class SinkPerThreadSink implements Sink, Disposable {
+
+    private Supplier<Sink> sinkSupplier;
+    private Cache<Thread, Sink> sinkCache =
+        newBuilder().weakValues().removalListener(notification -> disposeIfNeeded(notification.getValue(), NOP_LOGGER)).build();
```
dfeist: Because i'm not interested in logging the failure to dispose a sink during garbage collection.

core/src/main/java/org/mule/runtime/core/processor/strategy/SynchronousProcessingStrategyFactory.java
```diff
@@ -45,4 +59,71 @@ public ProcessingStrategy create(MuleContext muleContext, String schedulersNameP
     return SYNCHRONOUS_PROCESSING_STRATEGY_INSTANCE;
   }
 
+  static class SinkPerThreadSink implements Sink, Disposable {
```
dfeist: Maybe, not as clear though.  It's a Sink that uses a Sink per thread.  Let me see i I can come up with another name...

PR #4093, opened 2017-02-06
MULE-11311: allowInlineDefinition annotation is ignored on packaging

pablomusumeci:


modules/extensions-spring-support/src/test/resources/schemas/vegan.xsd
```diff
@@ -82,18 +82,6 @@
                                 </xs:choice>
                             </xs:complexType>
                         </xs:element>
-                        <xs:element minOccurs="0" maxOccurs="1" name="parameterized-functions">
```
alegmarra: I doubt this is OK, since the definition for this is: 
@Parameter
@Optional
private Function<Event, Map<String, String>> parameterizedFunction;

If we are making the Functions "Expression.REQUIRED" or if we don't care at all because the Functions will be removed, we can accept this. Other way, we have to fix this side effect

marianogonzalez: we just killed functions

tests/test-extensions/vegan-extension/src/main/java/org/mule/test/vegan/extension/VeganFidelityOperation.java
```diff
@@ -16,7 +16,8 @@
 
 public class VeganFidelityOperation {
 
-  public String tryEat(@XmlHints(allowReferences = false) Object food) {
+  public String tryEat(@XmlHints(allowReferences = false) Object food,
+                       @Optional @XmlHints(allowInlineDefinition = false) Menu<String, Integer> menu) {
```
marianogonzalez: why is it necessary to create your own Map implementation in order to test this? Not really sure custom Map types is something we want to support

PR #4094, opened 2017-02-06
MULE-11521: Define interception handlers registration mechanism

elrodro83:


core/src/main/java/org/mule/runtime/core/processor/interceptor/DefaultProcessorInterceptorManager.java
```diff
@@ -19,23 +22,38 @@
...
+    interceptorsOrder = asList(packagesOrder);
+  }
 
-  /**
-   * {@inheritDoc}
-   */
   @Override
-  public void addInterceptor(ProcessorInterceptor interceptor) {
+  public void addInterceptor(ProcessorInterceptorFactory interceptor) {
     checkNotNull(interceptor, "interceptionHandler cannot be null");
```
gsfernandes: change parameter name and null check message.

PR #4102, opened 2017-02-07
MULE-11176: Default Scheduler is running tasks with the wrong

elrodro83:
classloader
* Fix for scheduled tasks


services/scheduler/src/main/java/org/mule/service/scheduler/internal/RunnableRepeatableFutureDecorator.java
```diff
@@ -90,6 +97,8 @@ public void run() {
       if (logger.isTraceEnabled()) {
         logger.trace("Task " + this.toString() + " finished after " + (System.nanoTime() - startTime) + " nanoseconds");
       }
+
+      currentThread.setContextClassLoader(currentClassLoader);
```
pablokraan: Thread context classLoader  must be updated inside a try/finally

PR #4103, opened 2017-02-07
MULE-11620: Exceptions in interceptionAPI future from around hang

elrodro83:
execution


core/src/main/java/org/mule/runtime/core/processor/interceptor/ReactiveInterceptorAdapter.java
```diff
@@ -116,6 +115,9 @@ private boolean implementsAround(ProcessorInterceptor interceptor) {
     final ReactiveInterceptionAction reactiveInterceptionAction = new ReactiveInterceptionAction(interceptionEvent, next);
     return interceptor.around(resolveParameters(event, component, dslParameters), interceptionEvent,
                               reactiveInterceptionAction)
+        .exceptionally(t -> {
+          throw new CompletionException(new MessagingException(event, t.getCause()));
```
dfeist: can you set the failing processor here, you have it it seems?

PR #4104, opened 2017-02-07
MULE-11422: Prepare FakeMuleServer to support policies and plugins

balbifm:


tests/infrastructure/src/main/java/org/mule/test/infrastructure/deployment/FakeMuleServer.java
```diff
@@ -227,6 +230,8 @@ private void setMuleFolders() throws IOException {
     serverPluginsDir = createFolder("server-plugins");
     servicesDir = createFolder(SERVICES_FOLDER);
     domainsDir = createFolder(DOMAINS_FOLDER);
+    pluginsDir = createFolder(PLUGINS_FOLDER);
```
aiannucci: As plugins will live inside each application, why is this necessary?

PR #4106, opened 2017-02-07
MULE-11753: Remove hardcoded versions of maven project.version from extensions-support module tests.

anosenzo:


modules/extensions-support/pom.xml
```diff
@@ -174,6 +174,36 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
alegmarra: let's move this to plugin management in mule

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/java/JavaDeclarationDelegateTestCase.java
```diff
@@ -158,7 +159,7 @@
   private static final String GET_MEDICAL_HISTORY = "getMedicalHistory";
   private static final String APPROVE_INVESTMENT = "approve";
   private static final String IGNORED_OPERATION = "ignoredOperation";
-  private static final String EXTENSION_VERSION = "4.0-SNAPSHOT";
+  private static final String EXTENSION_VERSION = getProperty("maven_project_version");
```
alegmarra: util method for rehuse

PR #4112, opened 2017-02-08
MULE-11766: DefaultExtensionManager's method attemptToCreateImplicitConfigurationInstance should be synchronized

mbuchwald:


modules/extensions-support/src/main/java/org/mule/module/extension/internal/manager/DefaultExtensionManager.java
```diff
@@ -144,7 +144,7 @@ else if (matches > 1)
         }
     }
 
-    private void attemptToCreateImplicitConfigurationInstance(Extension extension, OperationContext operationContext)
+    private synchronized void attemptToCreateImplicitConfigurationInstance(Extension extension, OperationContext operationContext)
     {
         ConfigurationInstanceHolder configurationInstanceHolder = implicitConfigurationFactory.createImplicitConfigurationInstance(extension, operationContext, this);
```
pablokraan: Can't you synchronize on the extension instance to avoid locking the whole manager?

PR #4113, opened 2017-02-08
MULE-11459: Introduce CursorStreams - fixing some tests since I'm ino…

marianogonzalez:
…cent from most


core/src/main/resources/META-INF/mule-module.properties
```diff
@@ -85,6 +85,7 @@ artifact.export.classPackages=org.mule.runtime.core,\
                               org.mule.runtime.core.internal.metadata,\
                               org.mule.runtime.core.internal.registry,\
                               org.mule.runtime.core.internal.streaming,\
+                              org.mule.runtime.core.internal.streaming.object,\
```
pablokraan: At this point of the development, we should take care of adding real API to this file, and having something internal does not seem to be API

PR #4125, opened 2017-02-10
MULE-11459: Introduce Cursor Streams - correcting fixes

marianogonzalez:


core/src/main/java/org/mule/runtime/core/internal/streaming/bytes/DefaultByteStreamingManager.java
```diff
@@ -11,20 +11,19 @@
...
 import org.mule.runtime.core.api.EventContext;
 import org.mule.runtime.core.api.MuleContext;
-import org.mule.runtime.core.internal.streaming.bytes.factory.FileStoreCursorStreamProviderFactory;
-import org.mule.runtime.core.internal.streaming.bytes.factory.InMemoryCursorStreamProviderFactory;
-import org.mule.runtime.core.internal.streaming.bytes.factory.NullCursorStreamProviderFactory;
 import org.mule.runtime.core.streaming.bytes.ByteStreamingStatistics;
 import org.mule.runtime.core.streaming.bytes.CursorStreamProviderFactory;
 import org.mule.runtime.core.streaming.bytes.FileStoreCursorStreamConfig;
 import org.mule.runtime.core.streaming.bytes.InMemoryCursorStreamConfig;
+import org.mule.runtime.core.internal.streaming.bytes.factory.FileStoreCursorStreamProviderFactory;
```
dfeist: Why is order of imports now inverts instead of alphabetical?

core/src/main/java/org/mule/runtime/core/AbstractEventContext.java
```diff
@@ -80,15 +80,7 @@ public final void error(Throwable throwable) {
...
-    if (completed) {
-      if (!childContexts.isEmpty()) {
-        return childContexts.stream().filter(c -> !c.isTerminated()).count() == 0;
-      }
-
-      return true;
-    }
-
-    return false;
+    return completed ? childContexts.stream().allMatch(EventContext::isTerminated) : false;
```
dfeist: In line 51, shouldn't you be setting boolean after completing mono maybe?

core/src/main/java/org/mule/runtime/core/internal/streaming/bytes/DefaultByteStreamingManager.java
marianogonzalez: don't know. INtelliJ stuff. Code formatter passes

core/src/main/java/org/mule/runtime/core/AbstractEventContext.java
marianogonzalez: what would the practical differnece be?

PR #4127, opened 2017-02-10
MULE-11750: Add URL option to HTTP request

afelisatti:
DO NOT MERGE: This is a spike that must be reviewed and it's not final.

### DIFF
--- a/extensions/http/src/main/java/org/mule/extension/http/internal/request/ConfigurationOverrides.java
+++ b/extensions/http/src/main/java/org/mule/extension/http/internal/request/ConfigurationOverrides.java
@@ -21,22 +21,6 @@ import org.mule.runtime.extension.api.annotation.param.display.Placement;
 public final class ConfigurationOverrides {
 
   /**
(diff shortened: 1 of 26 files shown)


modules/http/src/main/java/org/mule/extension/http/internal/request/ConfigurationOverrides.java
```diff
@@ -21,22 +21,6 @@
 public final class ConfigurationOverrides {
 
   /**
-   * Host where the requests will be sent.
```
pablolagreca: Why remove this ones? This are still meant to be used

pablolagreca: Are this cover by the uri settings?

afelisatti: The use case for overriding host and port is kind of obsolete to me when you can override the entire url.

PR #4130, opened 2017-02-10
MULE-10545: Change http selectors/workers to use a connector pool from

elrodro83:
the service
* Fix tests that use http over the testing scheduler service

### DIFF
--- a/tests/unit/src/main/java/org/mule/tck/SimpleUnitTestSupportScheduler.java
+++ b/tests/unit/src/main/java/org/mule/tck/SimpleUnitTestSupportScheduler.java
@@ -38,7 +38,7 @@ public class SimpleUnitTestSupportScheduler extends ScheduledThreadPoolExecutor 
 
   @Override
   public void stop(long gracefulShutdownTimeoutSecs, TimeUnit unit) {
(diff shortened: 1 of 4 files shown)


tests/unit/src/main/java/org/mule/tck/SimpleUnitTestSupportScheduler.java
```diff
@@ -38,7 +38,7 @@ public SimpleUnitTestSupportScheduler(int corePoolSize,
 
   @Override
   public void stop(long gracefulShutdownTimeoutSecs, TimeUnit unit) {
-    // Nothing to do. The lifecycle of this pool is managed by the UnitTestSchedulerService that instantiated this
+    shutdownNow();
```
marianogonzalez: you used to be fine with dong nothing. Won't there be any problems if you shutdown twice (once here and another in the owning lifecycle)?

elrodro83: Added the check on SimpleUnitTestSupportSchedulerService#stop to avoid problems.

PR #4138, opened 2017-02-11

pablolagreca:
… each processor - fix test cases


core/src/main/java/org/mule/runtime/core/processor/ReferenceProcessor.java
```diff
@@ -47,6 +48,11 @@ public Event process(Event event) throws MuleException {
     return referencedProcessor.process(event);
   }
 
+  @Override
+  public Publisher<Event> apply(Publisher<Event> publisher) {
```
dfeist: Why is this an `InterceptingMessageProcessor` if it never ever uses `next` via `processNext` or `applyNext`?

pablolagreca: Dan it is used by the process method which is delgated to the underalying (referencedProcessor)

PR #4141, opened 2017-02-13

pablolagreca:
… each processor - Updating ConfigurationComponentLocator to use new location path mechanism

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/locator/Location.java
+++ b/core/src/main/java/org/mule/runtime/core/api/locator/Location.java
@@ -1,0 +1,186 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.api.locator;
+
+import static java.lang.Integer.parseInt;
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/mule/runtime/core/api/locator/Location.java
```diff
@@ -0,0 +1,186 @@
...
+
+/**
+ * A location allows to define the position of a certain component in the configuration.
+ * 
+ * Only components contained within global components with name can be referenced using a {@link Location} instance.
+ * 
+ * It is expected that the string representation of the object defined the serialized for of the location which consist of the
+ * global element name and the parts separated by an slash character.
+ * 
+ * @since 1.0
```
marianogonzalez: 4.0

core/src/main/java/org/mule/runtime/core/internal/locator/DefaultConfigurationComponentLocator.java
```diff
@@ -6,82 +6,39 @@
...
-  }
-
-  //TODO MULE-10751 - add complete support for componentPath
-  @Override
-  public Object findByPath(String componentPath) {
-    String[] parts = componentPath.split("/");
-    if (parts.length == 0) {
-      throw new IllegalArgumentException("Incomplete component path " + componentPath);
+  public Optional<Object> find(Location location) {
+    for (ComponentLocation componentLocation : componentMap.keySet()) {
```
marianogonzalez: this is easily replaced by a findAny

PR #4142, opened 2017-02-13
MULE-11825: Template-query isn't validating the definition of params …

facundovs:
…used in query text.

If a param is referenced in the query text, It must be defined as a in-param in the db:template.
db:insert (or other operation) only can override params that have already defined in the db:template.
If a parameter referenced in the query text wasn't defined an exception should be triggered.


modules/db/src/test/java/org/mule/module/db/integration/template/TemplateQueryConfigTestCase.java
```diff
@@ -132,23 +132,6 @@ public void usesDynamicQueryWithPlaceholder() throws Exception
     }
 
     @Test
-    public void usesNullDefaultParamValue() throws Exception
```
facundovs: I removed this case because It was testing the behavior that I'm fixing.
In the config, text query references a param that isn't defined through a db-in-param element.
It must trigger an exception.
Currently, It is setting the param value as null.
If It's important setting a parameter with a null value, please let me know and I'll add this feature.

modules/db/src/test/resources/integration/select/select-unresolved-named-param-config.xml
```diff
@@ -8,6 +8,7 @@
     <flow name="unresolvedNamedParam">
         <db:select>
             <db:parameterized-query>select * from Planet where name = :name</db:parameterized-query>
+            <db:in-param name="name" value="#[payload]"/>
```
facundovs: This case is testing the definition of a db:in-param element with a name that isn't referenced by the query text.
Due to the changes, As text query parameter references an inexistent db-in-param element, this exception is triggered first.
So I added a db:in-param with name "name" to can test this feature.

modules/db/src/main/java/org/mule/module/db/internal/config/domain/query/ParameterizedQueryTemplateFactoryBean.java
```diff
@@ -143,4 +147,10 @@ public boolean isSingleton()
     {
         return false;
     }
+
+    private String buildNotDefinedInParamErrorMessage(String name)
+    {
+        return "Parameter with name " + name + ", used in the query text, does not match any defined query parameter name.";
```
pablokraan: Use String format and change the string to be like "Parameter with name 'xxx', used ..."

modules/db/src/main/java/org/mule/module/db/internal/config/domain/query/ParameterizedQueryTemplateFactoryBean.java
```diff
@@ -54,7 +54,11 @@ public QueryTemplate getObject() throws Exception
         {
             QueryParam param = findOverriddenParam(templateParam.getName(), queryParams);
 
-            if (param == null)
+            if (param == null && ((DefaultInputQueryParam) templateParam).isDbInParam())
```
pablokraan: Why do you need to check whether or not the param is an input parameter?

modules/db/src/main/java/org/mule/module/db/internal/domain/param/DefaultInputQueryParam.java
```diff
@@ -30,4 +32,18 @@ public Object getValue()
...
+    /**
+     * @return true if the InputQueryParam references a param that must be defined through a <db:in-param> element in the configuration file.
+     * false if the InputQueryParam is defined through a literal or a MEL expression
+     */
+    public boolean isDbInParam()
+    {
+        return isDbParam;
+    }
+
+    public void setDbParam(boolean dbParam)
```
pablokraan: This class was supposed to be immutable. Find a different way to create it with the right value

modules/db/src/test/java/org/mule/module/db/integration/select/SelectUnresolvedTextQueryParam.java
```diff
@@ -0,0 +1,29 @@
...
+ * LICENSE.txt file.
+ */
+
+package org.mule.module.db.integration.select;
+
+import org.mule.module.db.integration.config.AbstractConfigurationErrorTestCase;
+
+import org.junit.Test;
+
+public class SelectUnresolvedTextQueryParam extends AbstractConfigurationErrorTestCase
```
pablokraan: Test class must have TestCase suffix

modules/db/src/test/java/org/mule/module/db/integration/template/TemplateQueryConfigTestCase.java
pablokraan: Re-add the test, but define the parameter with null value

modules/db/src/main/java/org/mule/module/db/internal/config/domain/query/ParameterizedQueryTemplateFactoryBean.java
facundovs: Because If It's a literal or a MEL expression, Validation isn't necessary (They doesn't refer to a db:in-param config).

modules/db/src/main/java/org/mule/module/db/internal/config/domain/query/ParameterizedQueryTemplateFactoryBean.java
```diff
@@ -136,6 +137,14 @@ public boolean evaluate(Object object)
         return null;
     }
 
+    private Object checkNullValue(Object value)
```
facundovs: I added this method to set with a NULL value, literal "null" string.
I amn't sure, It was the expected behavior.

modules/db/src/test/resources/integration/template/template-config.xml
```diff
@@ -55,6 +55,11 @@
         <db:dynamic-query>${dynamicQuery}</db:dynamic-query>
     </db:template-query>
 
+    <db:template-query name="testNullParamsQuery">
+        <db:parameterized-query>SELECT * FROM PLANET WHERE POSITION = :position</db:parameterized-query>
+        <db:in-param name="position" defaultValue="null"/>
```
facundovs: I suppoused It must be a literal string value (because It's testing TemplateQuery, and It doesn't resolve MEL expressions.

modules/db/src/test/java/org/mule/module/db/integration/template/TemplateQueryConfigTestCase.java
```diff
@@ -132,6 +132,23 @@ public void usesDynamicQueryWithPlaceholder() throws Exception
...
+
+        QueryTemplate queryTemplate = (QueryTemplate) queryTemplateBean;
+
+        assertFalse(queryTemplate.isDynamic());
+        assertEquals("SELECT * FROM PLANET WHERE POSITION = ?", queryTemplate.getSqlText());
+        assertEquals(1, queryTemplate.getInputParams().size());
+        InputQueryParam param1 = queryTemplate.getInputParams().get(0);
+        assertEquals(UnknownDbType.getInstance(), param1.getType());
+        assertEquals("position", param1.getName());
+        assertEquals(null, param1.getValue());
```
facundovs: Should I compare with null value or with "null" literal Value?
What is the expected behavior?
Look at checkNullValue method.

modules/db/src/main/java/org/mule/module/db/internal/config/domain/query/ParameterizedQueryTemplateFactoryBean.java
pablokraan: Is weird that If something is a DefaultInputQueryParam then you ask for isDbInParam.
I would rename it to hasValue or similar, and must be defined in the InputQueryParam interface. (also note that you can't just cast to DefaultInputQueryParam, you should check that the cast is valid before doing it).

modules/db/src/main/java/org/mule/module/db/internal/config/domain/query/ParameterizedQueryTemplateFactoryBean.java
pablokraan: If you can define a NULL value on a template, then it makes sense to be able to set NULL on a db:input-param/db:inout-param.
I would only accept NULL to represent a null.
Also, note that this change must be properly documented on the migration guide.

modules/db/src/main/java/org/mule/module/db/internal/config/domain/query/ParameterizedQueryTemplateFactoryBean.java
```diff
@@ -143,4 +158,10 @@ public boolean isSingleton()
     {
         return false;
     }
+
+    private String buildNotDefinedInParamErrorMessage(String name)
+    {
+        return format("Parameter with name %s, used in the query text, does not match any defined query parameter name.", name);
```
pablokraan: Parameter with name %s, used in the query text, does not match any defined query parameter name >
Parameter with name '%s', used in the query text, does not match any defined query parameter name defined in the query template

modules/db/src/main/java/org/mule/module/db/internal/domain/param/DefaultInputQueryParam.java
```diff
@@ -14,20 +14,38 @@
...
-        this(index, type, value, null);
+        this(index, type, value, null, false);
     }
 
     public DefaultInputQueryParam(int index, DbType type, Object value, String name)
     {
+        this(index, type, value, name, false);
+    }
+
+    public DefaultInputQueryParam(int index, DbType type, Object value, String name, boolean isDbParam)
```
pablokraan: Instead of making this constructor public, add another constructor that does not receives the value and then make this one private

modules/db/src/main/java/org/mule/module/db/internal/domain/param/DefaultInputQueryParam.java
```diff
@@ -14,20 +14,38 @@
...
     }
 
     @Override
     public Object getValue()
     {
         return value;
     }
+
+    /**
+     * @return true if the InputQueryParam references a param that must be defined through a <db:in-param> element in the configuration file.
```
pablokraan: Rename the method as mentioned before. Javadoc must say something like "indicates if the parameter has a defined value...... This is useful when a paremter is defined in a query template...."

modules/db/src/main/java/org/mule/module/db/internal/domain/param/DefaultInputQueryParam.java
facundovs: But I used this contructor to set the property with the value "true" in SimpleQueryTemplateParser

distributions/standalone/src/main/resources/MIGRATION.txt
```diff
@@ -290,3 +290,4 @@ MULE-7524: Encoded mule expressions in dynamic endpoints address components are
 MULE-7546: Implementations of org.mule.api.MuleMessage now need to implement clearAttachments() which was added to the interface. Implementations that extend org.mule.el.context.AbstractMapContext must now implement clear() given this method has been removed from the abstract implementation.
 MULE-7612: maps returned by queries using the Database connector can have different keys as now column's aliases are used instead of column's names.
 MULE-8066: Since 3.6 MuleClient uses the new HTTP module for processing HTTP URLs. To use the previous behaviour use the attribute 'useTransportForUris' in the http:config  element that's located inside the mule configuration element. You can also set this globally by using the system property 'mule.http.useTransportForUris'.
+MULE-11825: In a db template query, to set a db param with the null value, you can use the "null" literal value. For example: <db:in-param name="name" defaultValue="null"/>
```
pablokraan: db > DB (abbreviations must be in upper case)

pablokraan: My previous comment was "I would only accept NULL to represent a null.", any reason you choose "null" instead?

modules/db/src/main/java/org/mule/module/db/internal/domain/param/DefaultInOutQueryParam.java
```diff
@@ -14,10 +14,17 @@
...
 
+    private final boolean hasValue;
+
     public DefaultInOutQueryParam(int index, DbType type, String name, Object value)
     {
-        super(index, type, name);
+        this(index,type, name, value, value != null);
+    }
 
+    private DefaultInOutQueryParam(int index, DbType type, String name, Object value, boolean hasValue)
```
pablokraan: Why do you have this private constructor?

modules/db/src/main/java/org/mule/module/db/internal/config/domain/query/ParameterizedQueryTemplateFactoryBean.java
```diff
@@ -54,7 +55,11 @@ public QueryTemplate getObject() throws Exception
         {
             QueryParam param = findOverriddenParam(templateParam.getName(), queryParams);
 
-            if (param == null)
+            if (param == null && templateParam instanceof InputQueryParam && ! ( (InputQueryParam) templateParam).hasValue())
```
pablokraan: Fix code format

modules/db/src/main/java/org/mule/module/db/internal/config/domain/query/ParameterizedQueryTemplateFactoryBean.java
```diff
@@ -96,16 +101,16 @@ private QueryParam overrideParam(QueryParam templateParam, QueryParam queryParam
         DbType paramType = templateParam.getType();
         if (!(queryParam.getType() instanceof UnknownDbType))
         {
-           paramType = queryParam.getType();
+            paramType = queryParam.getType();
```
pablokraan: Looks like this extra character is not needed

modules/db/src/main/java/org/mule/module/db/internal/config/domain/query/ParameterizedQueryTemplateFactoryBean.java
```diff
@@ -143,4 +158,10 @@ public boolean isSingleton()
     {
         return false;
     }
+
+    private String buildNotDefinedInParamErrorMessage(String name)
+    {
+        return format("Parameter with name %s, used in the query text, does not match any defined query parameter name defined in the query template.", name);
```
pablokraan: Fix error message as commented in a previous comment

modules/db/src/main/java/org/mule/module/db/internal/domain/param/DefaultInputQueryParam.java
```diff
@@ -14,20 +14,35 @@
...
     {
         this(index, type, value, null);
     }
 
     public DefaultInputQueryParam(int index, DbType type, Object value, String name)
     {
+        this(index, type, value, name, value != null);
+    }
+
+    private DefaultInputQueryParam(int index, DbType type, Object value, String name, boolean hasValue)
```
pablokraan: I don't understand why do you need this private constructor. In a previous version of your code, you were calling this, in that that time, public constructor with different boolean values. But now, there is no need to have it as it is used only from one constructor.
In a previous comment, the suggestion was to add a constructor without the "value" parameter, so the caller could invoke different ones depending on the given situation, and it that scenario a private constructor was useful. But now I'm lost

modules/db/src/main/java/org/mule/module/db/internal/domain/param/InputQueryParam.java
```diff
@@ -14,4 +14,10 @@
 {
 
     Object getValue();
+
+    /**
+     * @return true if the parameter has a defined value in the query template.
```
pablokraan: query template. > query template, false otherwise.

modules/db/src/test/java/org/mule/module/db/integration/select/SelectUnresolvedTextQueryParamTestCase.java
```diff
@@ -0,0 +1,29 @@
...
+{
+
+    @Override
+    protected String getConfigFile()
+    {
+        return "integration/select/select-unresolved-text-query-param-config.xml";
+    }
+
+    @Test
+    public void requires() throws Exception
```
pablokraan: requires what?

modules/db/src/test/resources/integration/template/template-config.xml
```diff
@@ -57,6 +57,7 @@
 
     <db:template-query name="testNullParamsQuery">
         <db:parameterized-query>SELECT * FROM PLANET WHERE POSITION = :position</db:parameterized-query>
+        <db:in-param name="position" defaultValue="null"/>
```
pablokraan: In a previous comment I mentioned that if you are going to use  "NULL" to set a null value then you hsould also support setting "NULL" when a template is used on a DB message processor (or when the query is defined implicitly inside the message processor). Please add another test to cover those scenarios.

modules/db/src/main/java/org/mule/module/db/internal/config/domain/query/ParameterizedQueryTemplateFactoryBean.java
```diff
@@ -132,6 +137,16 @@ public boolean evaluate(Object object)
         return null;
     }
 
+    private Object checkNullValue(Object value)
+    {
+        if (value instanceof String && ("null".equals(value)))
```
pablokraan: Add a constant for the value that represents null.

pablokraan: instanceOf is redundant

modules/db/src/main/java/org/mule/module/db/internal/util/ValueUtils.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+
+package org.mule.module.db.internal.util;
+
+public class ValueUtils
```
pablokraan: Add javadoc

modules/db/src/main/java/org/mule/module/db/internal/util/ValueUtils.java
```diff
@@ -0,0 +1,29 @@
...
+{
+
+    private static final String NULL_VALUE = "NULL";
+
+    private ValueUtils()
+    {
+    }
+
+    public static Object setNullValue(Object value)
```
pablokraan: This is NOT a setNullValue, as it not setting the value on a field as expected

pablokraan: Add javadoc

PR #4143, opened 2017-02-14
MULE-11829: When using maxRedeliveryAttempts="0" in a rollbackexception strategy, it is assumed that redeliveryAttempts is set.

fsgonz:


core/src/main/java/org/mule/exception/RollbackMessagingExceptionStrategy.java
```diff
@@ -47,7 +47,7 @@ public Integer getMaxRedeliveryAttempts()
 
     public boolean hasMaxRedeliveryAttempts()
     {
-        return this.maxRedeliveryAttempts != null;
+        return this.maxRedeliveryAttempts != null && this.maxRedeliveryAttempts != 0;
```
pablolagreca: This is wrong. hasMaxRedeliveryAttempts() must always return true if there's an explicit configuration on it.

When the flow has a rollback-exception-strategy with a maxRedeliveryAttempts configured on it, then after the flow an IdempotentRedeliveryPolicy is added after the source. This class lock messages that have the same ID to prevent processing the same event concurrently so redelivery can be handle correctly. By default, the IdempotentRedeliveryPolicy is configured to generate the unique message id using a hash over the content of the message. If all the messages have the same content then the same id is generated and it's to be expected that the messages will all be executed synchronously.

Verify:
 - If all messages have the same generated ID, then the user should define an idempotent-redelivery-policy after the source and customize the message id.
- if no redelivery is expected, then just don't configure maxRedeliveryAttempts
 - If all messages are not processed until the first one is processed, then further investigation is required.

fsgonz:
I close this pull request. First I'm running tests involving different kind of code change proposals.

PR #4144, opened 2017-02-14

pablolagreca:
… each processor - fix tests


modules/spring-config/src/main/java/org/mule/runtime/config/spring/factories/AsyncMessageProcessorsFactoryBean.java
```diff
@@ -59,7 +59,7 @@ public AsyncDelegateMessageProcessor getObject() throws Exception {
 
   @Override
   public boolean isSingleton() {
-    return false;
+    return true;
```
marianogonzalez: why is this? If I remember correctly  this has always been like this. Wouldn't this affect unrelated functionality?

PR #4151, opened 2017-02-16
MULE-11835: RSS parser isn't parsing elements with namespaces.

facundovs:
Use of Namespaces in RSS was added in RSS 1.0.
Currently, We used the Rome API, that doesn't implement the parsing of elements that have namespaces.
However, Rome API allows us to implement our own parser.
An own parser should be created to parse elements with namespaces.

### DIFF
--- a/distributions/standalone/assembly-whitelist.txt
+++ b/distributions/standalone/assembly-whitelist.txt
@@ -305,7 +305,7 @@
 /mule-standalone-${productVersion}/lib/opt/jbossjta-4.15.0.Final.jar
 /mule-standalone-${productVersion}/lib/opt/jcifs-1.3.3.jar
 /mule-standalone-${productVersion}/lib/opt/jcodings-1.0.16.jar
-/mule-standalone-${productVersion}/lib/opt/jdom-2.0.2.jar
+/mule-standalone-${productVersion}/lib/opt/jdom2-2.0.6.jar
 /mule-standalone-${productVersion}/lib/opt/jersey-client-2.11.jar
 /mule-standalone-${productVersion}/lib/opt/jersey-common-2.11.jar
 /mule-standalone-${productVersion}/lib/opt/jersey-guava-2.11.jar
--- a/distributions/standalone/assembly-whitelist.txt
+++ b/distributions/standalone/assembly-whitelist.txt
@@ -305,7 +305,7 @@
 /mule-standalone-${productVersion}/lib/opt/jbossjta-4.15.0.Final.jar
 /mule-standalone-${productVersion}/lib/opt/jcifs-1.3.3.jar
 /mule-standalone-${productVersion}/lib/opt/jcodings-1.0.16.jar
-/mule-standalone-${productVersion}/lib/opt/jdom-1.1.3.jar
+/mule-standalone-${productVersion}/lib/opt/jdom-2.0.2.jar
 /mule-standalone-${productVersion}/lib/opt/jersey-client-2.11.jar
 /mule-standalone-${productVersion}/lib/opt/jersey-common-2.11.jar
 /mule-standalone-${productVersion}/lib/opt/jersey-guava-2.11.jar
@@ -379,7 +379,8 @@
 /mule-standalone-${productVersion}/lib/opt/reflections-0.9.9.jar
 /mule-standalone-${productVersion}/lib/opt/relaxngDatatype-20020414.jar
 /mule-standalone-${productVersion}/lib/opt/rhino-1.7R4.jar
-/mule-standalone-${productVersion}/lib/opt/rome-0.9.jar
+/mule-standalone-${productVersion}/lib/opt/rome-1.5.0.jar
+/mule-standalone-${productVersion}/lib/opt/rome-utils-1.5.0.jar
 /mule-standalone-${productVersion}/lib/opt/Saxon-HE-9.6.0-7.jar
 /mule-standalone-${productVersion}/lib/opt/Saxon-HE-9.6.0-7-xqj.jar
 /mule-standalone-${productVersion}/lib/opt/signpost-core-1.2.1.2.jar
--- a/distributions/standalone/assembly-whitelist.txt
+++ b/distributions/standalone/assembly-whitelist.txt
@@ -305,7 +305,7 @@
 /mule-standalone-${productVersion}/lib/opt/jbossjta-4.15.0.Final.jar
 /mule-standalone-${productVersion}/lib/opt/jcifs-1.3.3.jar
 /mule-standalone-${productVersion}/lib/opt/jcodings-1.0.16.jar
-/mule-standalone-${productVersion}/lib/opt/jdom2-2.0.6.jar
+/mule-standalone-${productVersion}/lib/opt/jdom-2.0.2.jar
 /mule-standalone-${productVersion}/lib/opt/jersey-client-2.11.jar
 /mule-standalone-${productVersion}/lib/opt/jersey-common-2.11.jar
 /mule-standalone-${productVersion}/lib/opt/jersey-guava-2.11.jar
--- a/modules/rss/pom.xml
+++ b/modules/rss/pom.xml
@@ -31,12 +31,6 @@
                 <groupId>com.rometools</groupId>
                 <artifactId>rome</artifactId>
                 <version>1.5.0</version>
-            <exclusions>
-                <exclusion>
-                    <groupId>org.jdom</groupId>
-                    <artifactId>jdom</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.jdom</groupId>
--- a/modules/rss/pom.xml
+++ b/modules/rss/pom.xml
@@ -28,12 +28,12 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
-            <groupId>rome</groupId>
-            <artifactId>rome</artifactId>
-            <version>0.9</version>
+                <groupId>com.rometools</groupId>
+                <artifactId>rome</artifactId>
+                <version>1.5.0</version>
             <exclusions>
                 <exclusion>
-                    <groupId>jdom</groupId>
+                    <groupId>org.jdom</groupId>
                     <artifactId>jdom</artifactId>
                 </exclusion>
             </exclusions>
--- a/modules/rss/pom.xml
+++ b/modules/rss/pom.xml
@@ -30,17 +30,17 @@
         <dependency>
                 <groupId>com.rometools</groupId>
                 <artifactId>rome</artifactId>
-                <version>1.5.0</version>
+                <version>1.7.1</version>
             <exclusions>
                 <exclusion>
                     <groupId>org.jdom</groupId>
-                    <artifactId>jdom</artifactId>
+                    <artifactId>jdom2</artifactId>
                 </exclusion>
             </exclusions>
         </dependency>
         <dependency>
             <groupId>org.jdom</groupId>
-            <artifactId>jdom</artifactId>
+            <artifactId>jdom2</artifactId>
             <version>${jdomVersion}</version>
         </dependency>
 
--- a/modules/rss/pom.xml
+++ b/modules/rss/pom.xml
@@ -30,17 +30,17 @@
         <dependency>
                 <groupId>com.rometools</groupId>
                 <artifactId>rome</artifactId>
-                <version>1.7.1</version>
+                <version>1.5.0</version>
             <exclusions>
                 <exclusion>
                     <groupId>org.jdom</groupId>
-                    <artifactId>jdom2</artifactId>
+                    <artifactId>jdom</artifactId>
                 </exclusion>
             </exclusions>
         </dependency>
         <dependency>
             <groupId>org.jdom</groupId>
-            <artifactId>jdom2</artifactId>
+            <artifactId>jdom</artifactId>
             <version>${jdomVersion}</version>
         </dependency>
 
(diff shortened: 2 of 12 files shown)


modules/rss/src/main/java/org/mule/module/rss/transformers/RssParser.java
```diff
@@ -0,0 +1,229 @@
...
+    @Override
+    public Item parseItem(final Element rssRoot, final Element itemElement, final Locale locale)
+    {
+
+        final Item item = super.parseItem(rssRoot, itemElement, locale);
+
+        if (item.getTitle() == null)
+        {
+            final Element title = parseChildWithNamespaces(itemElement, "title");
+            if (item.getTitle() == null)
```
pablokraan: Check for item.getTitle == null is redundant at this point. Probably you want to check title element instead

modules/rss/src/main/java/org/mule/module/rss/transformers/RssParser.java
```diff
@@ -0,0 +1,229 @@
...
+            {
+                item.setTitle(title.getText());
+            }
+        }
+
+        if (item.getLink() == null)
```
pablokraan: Try to extract utility methods for each block that manages an attribute

modules/rss/src/test/java/org/mule/module/rss/transformers/RssParserTestCase.java
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+
+package org.mule.module.rss.transformers;
+
+
```
pablokraan: Remove extra empty lines

modules/rss/src/test/java/org/mule/module/rss/transformers/RssParserTestCase.java
```diff
@@ -0,0 +1,72 @@
...
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.InputStream;
+import java.util.List;
+
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class RssParserTestCase
```
pablokraan: Extend from AbstractMuleTestCase

modules/rss/src/test/java/org/mule/module/rss/transformers/RssParserTestCase.java
```diff
@@ -0,0 +1,72 @@
...
+{
+    private static List<SyndEntry> entries = null;
+    @BeforeClass
```
pablokraan: Add extra line

modules/rss/src/test/java/org/mule/module/rss/transformers/RssParserTestCase.java
```diff
@@ -0,0 +1,72 @@
...
+import java.io.FileInputStream;
+import java.io.InputStream;
+import java.util.List;
+
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class RssParserTestCase
+{
+    private static List<SyndEntry> entries = null;
```
pablokraan: Why is this field static?

distributions/standalone/assembly-whitelist.txt
```diff
@@ -379,7 +379,8 @@
 /mule-standalone-${productVersion}/lib/opt/reflections-0.9.9.jar
 /mule-standalone-${productVersion}/lib/opt/relaxngDatatype-20020414.jar
 /mule-standalone-${productVersion}/lib/opt/rhino-1.7R4.jar
-/mule-standalone-${productVersion}/lib/opt/rome-0.9.jar
+/mule-standalone-${productVersion}/lib/opt/rome-1.5.0.jar
```
pablokraan: Why is this library updated?

pablokraan: Remember to validate license and vulnerabilities with QA

modules/rss/pom.xml
```diff
@@ -28,12 +28,12 @@
...
-            <groupId>rome</groupId>
-            <artifactId>rome</artifactId>
-            <version>0.9</version>
+                <groupId>com.rometools</groupId>
+                <artifactId>rome</artifactId>
+                <version>1.5.0</version>
             <exclusions>
                 <exclusion>
-                    <groupId>jdom</groupId>
+                    <groupId>org.jdom</groupId>
```
pablokraan: Why is this dependency excluded here but updated on the whitelist? Who is adding the updated dependency?

distributions/standalone/assembly-whitelist.txt
facundovs: I need to update the library to able to implement my own parser.
Previous versions don't allow that.

modules/rss/pom.xml
facundovs: I updated the version of the library in the parent pom

modules/rss/src/test/java/org/mule/module/rss/transformers/RssParserTestCase.java
facundovs: Because It's initialized in beforeClass static method.

modules/rss/pom.xml
pablokraan: Your answer is contradictory. If you needed to update the dependency because rome needs it, then why are you excluding that dependency here? Am I missing something?

modules/rss/src/main/java/org/mule/module/rss/transformers/RssParser.java
```diff
@@ -0,0 +1,272 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.Locale;
+
+import org.jdom2.Document;
+import org.jdom2.Element;
+import org.jdom2.Namespace;
+
+
+public class RssParser extends RSS20Parser implements WireFeedParser
```
pablokraan: Add javadocs

modules/rss/src/main/java/org/mule/module/rss/transformers/RssParser.java
```diff
@@ -0,0 +1,272 @@
...
+    @Override
+    protected Namespace getRSSNamespace()
+    {
+        return NO_NAMESPACE;
+    }
+
+    @Override
+    public Item parseItem(final Element rssRoot, final Element itemElement, final Locale locale)
+    {
+
```
pablokraan: Do not add empty lines just because it is free. Add them only when it makes the code easier to read or to separate field/method/etc. Do not add empty lines at the beginning of a method body

modules/rss/src/main/java/org/mule/module/rss/transformers/RssParser.java
```diff
@@ -0,0 +1,272 @@
...
+        parseContent(item, itemElement);
+        parseCategories(item, itemElement);
+
+        return item;
+    }
+
+    private Element parseChildWithNamespaces(Element parentElement, String childName)
+    {
+        Element element = null;
+        Iterator itr = namespaces.iterator();
```
pablokraan: Use meaningful manes for variables. Rename itr

pablokraan: Why do you need an iterator if you can iterate on the list using a foreach?

modules/rss/src/main/java/org/mule/module/rss/transformers/RssParser.java
```diff
@@ -0,0 +1,272 @@
...
+            element = parentElement.getChild(childName, namespace);
+        }
+
+        return element;
+    }
+
+    private List<Element> parseChildrenWithNamespaces(Element parentElement, String childName)
+    {
+        List<Element> elements = new ArrayList<>();
+        Iterator itr = namespaces.iterator();
```
pablokraan: Same here

modules/rss/src/test/java/org/mule/module/rss/transformers/RssParserTestCase.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.InputStream;
+import java.util.List;
+
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class RssParserTestCase extends AbstractMuleTestCase
```
pablokraan: Add @SmallTest

modules/rss/src/test/java/org/mule/module/rss/transformers/RssParserTestCase.java
```diff
@@ -0,0 +1,74 @@
...
+{
+
+    private static List<SyndEntry> entries = null;
+
+    @BeforeClass
+    public static void setup() throws Exception
+    {
+        ObjectToRssFeed objectToRssFeed = new ObjectToRssFeed();
+        InputStream fileInputStream;
+        fileInputStream = new FileInputStream(new File("src/test/resources/rss-test.xml"));
```
pablokraan: Load the file as a resource using the classloader

modules/rss/src/test/java/org/mule/module/rss/transformers/RssParserTestCase.java
```diff
@@ -0,0 +1,74 @@
...
+        SyndFeed feed = (SyndFeed) objectToRssFeed.doTransform(fileInputStream, null);
+        entries = feed.getEntries();
+    }
+
+    @Test
+    public void testEntriesSize() throws Exception
+    {
+        assertThat(entries.size(), is(2));
```
pablokraan: What is the purpose of having separated tests that always run against the same shared state?

modules/rss/src/main/java/org/mule/module/rss/transformers/RssParser.java
facundovs: Because I also evaluate other conditions and I always try to avoid "breaks".

modules/rss/src/main/java/org/mule/module/rss/transformers/RssParser.java
```diff
@@ -0,0 +1,274 @@
...
+import java.util.List;
+import java.util.Locale;
+
+import org.jdom2.Document;
+import org.jdom2.Element;
+import org.jdom2.Namespace;
+
+/**
+ * Parses RSS item elements and its children.
+ * Considers the use of namespaces that were added in RSS 1.0
```
pablokraan: Can you clarify why we need to add something into the parser for RSS2.0 that was added on RSS 1.0?

modules/rss/src/main/java/org/mule/module/rss/transformers/RssParser.java
```diff
@@ -0,0 +1,274 @@
...
+ */
+public class RssParser extends RSS20Parser implements WireFeedParser
+{
+
+    private List<Namespace> namespaces;
+
+    @Override
+    public String getType()
+    {
+        return "rss_2.0";
```
pablokraan: Is not this a duplicate of the method in the parent class?

modules/rss/src/main/java/org/mule/module/rss/transformers/RssParser.java
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    @Override
+    public boolean isMyType(Document document)
+    {
+        return true;
```
pablokraan: Why is this method returning true? Can't you use the method from the parent class instead?

modules/rss/src/test/java/org/mule/module/rss/transformers/RssParserTestCase.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Test
+    public void testParsedEntries() throws Exception
+    {
+        assertValues(entries.get(0));
+        assertValues(entries.get(1));
+        assertThat(entries.size(), is(2));
+    }
+
+    public void assertValues(SyndEntry entry)
```
pablokraan: private

modules/rss/src/test/java/org/mule/module/rss/transformers/RssParserTestCase.java
```diff
@@ -0,0 +1,65 @@
...
+        InputStream fileInputStream;
+        fileInputStream = getClass().getResourceAsStream("/rss-test.xml");
+        SyndFeed feed = (SyndFeed) objectToRssFeed.doTransform(fileInputStream, null);
+        entries = feed.getEntries();
+    }
+
+    @Test
+    public void testParsedEntries() throws Exception
+    {
+        assertValues(entries.get(0));
```
pablokraan: Add some error message to use when the assertion fail or extract each line to a method so we can understand what are you asserting without having to read the RRS file

modules/rss/src/main/java/org/mule/module/rss/transformers/RssParser.java
facundovs: According the Rome Documentation: "If a plugin list definition contains more than one plugin with the same primary key, the latter one is the one that will be used. In the case or parsers, generators and converters the primary key is the type of feed they handle. " 
The RSS20Parser (a Rome class) doesn't consider the use of namespaces in the parsing. Therefore, I need to duplicate the method. In that way, my Parser is loaded and instantiated first. As you can see, in each method, first I call the parent method because I only added the extra feature of contempling namespaces.

modules/rss/src/main/java/org/mule/module/rss/transformers/RssParser.java
facundovs: According to the documentation this is the way to use this parser as the default. If I use the method of the parent class, the RSS 1.0 documents won't be parsed by this class (however, It is necessary because in RSS 1.0, namespaces were added).

PR #4156, opened 2017-02-20
MULE-11843: Http Server scheduler is shutdown before the connections are

elrodro83:
closed

### DIFF
--- a/modules/http/src/main/java/org/mule/runtime/module/http/internal/listener/grizzly/GrizzlyHttpServer.java
+++ b/modules/http/src/main/java/org/mule/runtime/module/http/internal/listener/grizzly/GrizzlyHttpServer.java
@@ -23,9 +23,6 @@ import java.util.Collection;
 import java.util.concurrent.ExecutorService;
 import java.util.function.Supplier;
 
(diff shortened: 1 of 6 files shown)


modules/http/src/main/java/org/mule/runtime/module/http/internal/listener/grizzly/GrizzlyHttpServer.java
```diff
@@ -36,23 +39,33 @@
...
+    this.schedulerDisposer = schedulerDisposer;
   }
 
   @Override
   public synchronized void start() throws IOException {
     this.scheduler = schedulerSource != null ? schedulerSource.get() : null;
     serverConnection = transport.bind(serverAddress.getIp(), serverAddress.getPort());
+    serverConnection.addCloseListener((CloseListener<Closeable, ICloseType>) (closeable, type) -> {
+      try {
+        scheduler.stop(5000, MILLISECONDS);
```
pablolagreca: The timeout for this should be related to the graceful shutdown of mule. Can you get that value from the mule context and use it here somehow?. 

If it's not possible then we should add a todo at least.

PR #4157, opened 2017-02-20
MULE-11847: support for prefix and namespace for smart connectors

fernandezlautaro:
Do not merge until https://github.com/mulesoft/mule-extensions-api/pull/267 is accepted and merged.

### DIFF
--- a/modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/loader/XmlExtensionLoaderDelegate.java
+++ b/modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/loader/XmlExtensionLoaderDelegate.java
@@ -18,6 +18,8 @@ import static org.apache.commons.lang3.StringUtils.isEmpty;
 import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
 import static org.mule.runtime.api.meta.model.display.LayoutModel.builder;
 import static org.mule.runtime.api.meta.model.parameter.ParameterRole.BEHAVIOUR;
+import static org.mule.runtime.extension.api.util.XmlModelUtils.createXmlLanguageModel;
+import com.google.common.collect.ImmutableMap;
 import org.mule.metadata.api.ClassTypeLoader;
 import org.mule.metadata.api.model.MetadataType;
 import org.mule.runtime.api.component.ComponentIdentifier;
@@ -47,8 +49,7 @@ import org.mule.runtime.extension.api.declaration.type.ExtensionsTypeLoaderFacto
 import org.mule.runtime.extension.api.loader.ExtensionLoadingContext;
 import org.mule.runtime.extension.internal.loader.catalog.loader.xml.TypesCatalogXmlLoader;
 import org.mule.runtime.extension.internal.loader.catalog.model.TypesCatalog;
-
-import com.google.common.collect.ImmutableMap;
+import org.w3c.dom.Document;
 
 import java.io.IOException;
 import java.net.URL;
@@ -61,8 +62,6 @@ import java.util.Optional;
 import java.util.Properties;
 import java.util.stream.Collectors;
 
-import org.w3c.dom.Document;
-
 /**
  * Describes an {@link ExtensionModel} by scanning an XML provided in the constructor
  *
@@ -74,6 +73,7 @@ final class XmlExtensionLoaderDelegate {
   private static final String PARAMETER_DEFAULT_VALUE = "defaultValue";
   private static final String TYPE_ATTRIBUTE = "type";
   private static final String MODULE_NAME = "name";
+  private static final String MODULE_PREFIX_ATTRIBUTE = "prefix";
   private static final String MODULE_NAMESPACE_ATTRIBUTE = "namespace";
   private static final String MODULE_NAMESPACE_NAME = "module";
   protected static final String CONFIG_NAME = "config";
@@ -92,6 +92,7 @@ final class XmlExtensionLoaderDelegate {
 
   private static final String CATEGORY = "category";
   private static final String VENDOR = "vendor";
+  private static final String MIN_MULE_VERSION = "minMuleVersion";
   private static final String DOC_DESCRIPTION = "doc:description";
   private static final String PASSWORD = "password";
   private static final String ROLE = "role";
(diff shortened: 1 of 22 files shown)


modules/spring-config/src/main/resources/META-INF/mule-module.xsd
```diff
@@ -150,24 +150,42 @@
...
+                            Min Mule version that the extension requires to work correctly.
+                        </xsd:documentation>
+                    </xsd:annotation>
+                </xsd:attribute>
+                <!-- DSL XML attributes -->
+                <xsd:attribute name="namespace" use="optional">
+                    <xsd:annotation>
+                        <xsd:documentation>
+                            Expected namespace of the module to look for when generating the schemas. If left empty it will
+                            default to http://www.mulesoft.org/schema/mule/[prefix], where [prefix] is the attribute defined below.
```
pablolagreca: where [prefix] is the attribute defined below. -> where [prefix] is the attribute prefix attribute value.

This documentation may be shown in Studio independently from this context.

modules/spring-config/src/main/resources/META-INF/mule-module.xsd
```diff
@@ -150,24 +150,42 @@
...
+                            Expected category of the module.
                         </xsd:documentation>
                     </xsd:annotation>
                 </xsd:attribute>
-                <xsd:attribute name="category" type="categoryType" default="COMMUNITY">
+                <xsd:attribute name="vendor" type="xsd:string" default="MuleSoft">
                     <xsd:annotation>
                         <xsd:documentation>
-                            Expected category of the module.
+                            Expected vendor of the module.
```
pablolagreca: Can you explain a little bit more what vendor is?. If I a new developer and just do an extension for me or my company, what should I put here?

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/loader/XmlExtensionLoaderDelegate.java
```diff
@@ -249,6 +243,12 @@ private void loadModuleExtension(ExtensionDeclarer declarer, ComponentModel modu
     }
   }
 
+  private XmlDslModel getXmlDslModel(ComponentModel moduleModel, String name, String version) {
```
marianogonzalez: just read identical code in JavaXmlDeclarationEnricher.java. Doesn't look like doing the same here is necessary at all

modules/spring-config/src/main/resources/META-INF/mule-module.xsd
```diff
@@ -150,24 +150,42 @@
...
+                <!-- DSL XML attributes -->
+                <xsd:attribute name="namespace" use="optional">
+                    <xsd:annotation>
+                        <xsd:documentation>
+                            Expected namespace of the module to look for when generating the schemas. If left empty it will
+                            default to http://www.mulesoft.org/schema/mule/[prefix], where [prefix] is the attribute defined below.
+                        </xsd:documentation>
+                    </xsd:annotation>
+                </xsd:attribute>
+                <xsd:attribute name="prefix" use="optional">
```
marianogonzalez: This is inconsistent with the default behaviour in java based connectors (where the concept of prefix doesn't exist at all). Why should smart connector have a different way of configuring this? Why don't they configure the same stuff as in the @Xml annotation?

modules/spring-config/src/main/resources/META-INF/mule-module.xsd
fernandezlautaro: I believe it's the owner of the connector
@gonzalez: have a more clear definition for it?

modules/spring-config/src/main/resources/META-INF/mule-module.xsd
fernandezlautaro: I believe the naming in the @Xml annotation is wrong:  check https://www.w3schools.com/xml/xml_namespaces.asp (that's the best I can do from a :iPhone:)

Prefix is the `sfdc` and namespace the value you expect to be in the XSD under targetNamespace

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/loader/XmlExtensionLoaderDelegate.java
fernandezlautaro: it's similar, but not the same. 
The @Xml annotation might not be present, that's why it has a different way of accessing the valuea

fernandezlautaro:
@pablolagreca @marianogonzalez can you review the comments and approves if apply?

modules/spring-config/src/main/resources/META-INF/mule-module.xsd
marianogonzalez: From the jdoc:

`
  /**
   * Returns the name of the extension's vendor
   * This name is used to:
   * <ul>
   * <li>Represent the extension's vendor</li>
   * <li>Differentiate different extensions with the same name, to give the possibility of having two connectors
   * with the same name, for example, one made by MuleSoft and the second one by a third party</li>
   * </ul>
   *
   * @return the name of the extension's vendor
   */`

modules/spring-config/src/main/resources/META-INF/mule-module.xsd
```diff
@@ -150,24 +150,42 @@
...
+                <!-- DSL XML attributes -->
+                <xsd:attribute name="namespace" use="optional">
+                    <xsd:annotation>
+                        <xsd:documentation>
+                            Expected namespace of the module to look for when generating the schemas. If left empty it will
+                            default to http://www.mulesoft.org/schema/mule/[prefix], where [prefix] is the attribute prefix attribute value.
+                        </xsd:documentation>
+                    </xsd:annotation>
+                </xsd:attribute>
+                <xsd:attribute name="prefix" use="optional">
```
marianogonzalez: This should not be prefix. There's no point in specifying a prefix here since the extension user is free to put whatever prefix he wants.

In the Java based SDK, there's an `@Xml` annotation in which the developer provides **fixed** attributes such as namespace, url location, etc. From there the SDK automatically infers a default prefix, knowing that the final user may change it.

fernandezlautaro: what the @Xml has is two attributes: `namespace` and `namespaceLocation` that maps to the prefix and namespace.
To my understanding, there's no `namespaceLocation` and it's confusing as it's being used as the namespace of an extension. From wikipedia: `...A namespace name is a uniform resource identifier (URI). ...`, which means having something like [`mulesoft`](https://github.com/mulesoft/mule-extensions-api/blob/master/mule-extensions-api/src/main/java/org/mule/runtime/extension/api/annotation/dsl/xml/Xml.java#L32) does not comply to a URI, but the [`schemaLocation`](https://github.com/mulesoft/mule-extensions-api/blob/master/mule-extensions-api/src/main/java/org/mule/runtime/extension/api/annotation/dsl/xml/Xml.java#L41) does.

marianogonzalez: ok, then change `@Xml` so that it says `prefix` instead of namespace. The thing is that the two experiences should match

PR #4166, opened 2017-02-22
MULE-11850: Paths for processors in dynamically referenced subflows are not recalculated

mbuchwald:


tests/integration/src/test/java/org/mule/context/notification/FlowStackTestCase.java
```diff
@@ -298,6 +320,18 @@ public void subFlowDynamicWithChoice() throws Exception
     }
 
     @Test
+    public void recursiveSubFlow() throws Exception
```
elrodro83: add a test case for recursive dynamic

core/src/main/java/org/mule/util/NotificationUtils.java
```diff
@@ -53,22 +51,19 @@ public String resolvePath(MessageProcessor processor)
             {
                 return path;
             }
-            else
+
+            for (Entry<MessageProcessor, String> flowMapEntries : flowMap.entrySet())
```
elrodro83: change to just use the keyset()

PR #4168, opened 2017-02-22
MULE-11865: Sftp endpoints does not preserve session serialized variables.

fsgonz:


transports/sftp/src/main/java/org/mule/transport/sftp/SftpMessageDispatcher.java
```diff
@@ -229,6 +231,15 @@ private String buildFilename(MuleEvent event)
     protected MuleMessage doSend(MuleEvent event) throws Exception
     {
         doDispatch(event);
+        addSerializedSessionAsInboundProperty(event);
         return event.getMessage();
     }
+
+    private void addSerializedSessionAsInboundProperty(MuleEvent event)
+    {
+        if (event.getMessage().getOutboundProperty(MuleProperties.MULE_SESSION_PROPERTY) != null)
```
elrodro83: static imports

PR #4172, opened 2017-02-23
MULE-11502:support for synthetic 'target' parameter in smort connectors

fernandezlautaro:


core/src/main/java/org/mule/runtime/core/processor/chain/ModuleOperationMessageProcessorChainBuilder.java
```diff
@@ -125,16 +131,23 @@ protected Event doProcess(Event event) throws MuleException {
...
       } else {
         return publisher;
       }
     }
 
     private Event createNewEventFromJustMessage(Event request, Event response) {
-      return Event.builder(request).message(builder(response.getMessage()).build()).build();
+      final Event.Builder builder = Event.builder(request);
+      if (target.isPresent()) {
+        builder.addVariable(target.get(), response.getMessage().getPayload().getValue(),
```
pablolagreca: The whole message needs to go as value for the target variable, not just the payload

modules/extensions-xml-support/src/test/java/org/mule/test/functional/ModuleSimpleTestCase.java
```diff
@@ -73,4 +75,22 @@ public void testSetPayloadUsingUndefinedParam() throws Exception {
...
     assertThat(muleEvent.getMessage().getPayload().getValue(), Is.is(nullValue()));
   }
+
+  @Test
+  public void testSetPayloadHardcodedFlowWithTarget() throws Exception {
+    Event event = flowRunner("testSetPayloadHardcodedFlowWithTarget").run();
+    assertThat(event.getMessage().getPayload().getValue(), nullValue());
+    final TypedValue<Object> targetVariable = event.getVariable("target-variable");
+    assertThat(targetVariable, notNullValue());
+    assertThat(targetVariable.getValue(), Is.is("hardcoded value"));
```
marianogonzalez: static import

PR #4175, opened 2017-02-23
MULE-11875: Race condition when putting an object in the registry

elrodro83:
asynchronously and disposing the muleContext at the same time

Also includes:
* Fix for the issue that prevented running integration tests alone
* Name of the lifecycle manager was not being properly generated

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/LifecycleAwareConfigurationProvider.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/LifecycleAwareConfigurationProvider.java
@@ -6,6 +6,7 @@
  */
 package org.mule.runtime.module.extension.internal.runtime.config;
 
+import static java.lang.String.format;
 import static org.mule.runtime.core.api.lifecycle.LifecycleUtils.disposeIfNeeded;
 import static org.mule.runtime.core.api.lifecycle.LifecycleUtils.initialiseIfNeeded;
 import static org.mule.runtime.core.api.lifecycle.LifecycleUtils.startIfNeeded;
@@ -13,6 +14,7 @@ import static org.mule.runtime.core.api.lifecycle.LifecycleUtils.stopIfNeeded;
 import static org.mule.runtime.core.util.ClassUtils.withContextClassLoader;
 import static org.mule.runtime.module.extension.internal.util.MuleExtensionUtils.getClassLoader;
 import static org.slf4j.LoggerFactory.getLogger;
+
 import org.mule.runtime.api.exception.MuleException;
 import org.mule.runtime.api.lifecycle.Disposable;
 import org.mule.runtime.api.lifecycle.Initialisable;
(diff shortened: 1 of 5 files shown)


modules/spring-config/src/test/java/org/mule/runtime/config/spring/SpringRegistryTestCase.java
```diff
@@ -71,4 +90,65 @@ private void createSpringRegistryWithParentContext() {
...
+    springRegistry.initialise();
+    springRegistry.getLifecycleManager().fireLifecycle(Startable.PHASE_NAME);
+
+    final Startable asyncStartableBean = mock(Startable.class, withSettings().extraInterfaces(Disposable.class));
+
+    final Latch startingLatch = new Latch();
+
+    doAnswer(invocation -> {
+      startingLatch.countDown();
+      sleep(1000);
```
marianogonzalez: this sleep should be replaced with a second latch. This is a potential source of flakyness

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/LifecycleAwareConfigurationProvider.java
```diff
@@ -56,8 +58,7 @@
   private final ConfigurationModel configurationModel;
   private final List<ConfigurationInstance> configurationInstances = new LinkedList<>();
   private final ClassLoader extensionClassLoader;
-  protected SimpleLifecycleManager lifecycleManager =
-      new DefaultLifecycleManager<>(String.format("%s-%s", getClass().getName(), getName()), this);
+  protected final SimpleLifecycleManager lifecycleManager;
```
marianogonzalez: what do you gain from this change?

tests/integration/pom.xml
```diff
@@ -92,6 +92,7 @@
             <groupId>org.mule.modules</groupId>
             <artifactId>mule-module-file-extension-common</artifactId>
             <version>${project.version}</version>
+            <classifier>mule-plugin</classifier>
```
marianogonzalez: how is this related?

modules/spring-config/src/test/java/org/mule/runtime/config/spring/SpringRegistryTestCase.java
elrodro83: a latch is for waiting for sonething to happen. here i have to check that something has not happenned after some time.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/LifecycleAwareConfigurationProvider.java
elrodro83: when this was being executed, the name was not set still and the string was ended with `-null`

tests/integration/pom.xml
elrodro83: i needed to fix this in order to run the integration tests and analyze it.

modules/spring-config/src/test/java/org/mule/runtime/config/spring/SpringRegistryTestCase.java
marianogonzalez: "waiting for some time" is the definition of flaky. In the bug you're trying to fix, why was this object being wrongfully disposed? what was triggering that disposal? You can add a second latch which is fired by that condition and THEN you verify that it was never disposed.

PR #4179, opened 2017-02-23
MULE-11821: support for optional parameters in smart connectors

fernandezlautaro:


modules/extensions-xml-support/src/test/resources/modules/module-param-default-types.xsd
```diff
@@ -61,6 +61,20 @@
...
     </xs:complexType>
+    <xs:element xmlns="http://www.mulesoft.org/schema/mule/module-param-default-types" xmlns:mule="http://www.mulesoft.org/schema/mule/core" type="Optional-parameter-operationType" substitutionGroup="mule:abstract-operator" name="optional-parameter-operation"></xs:element>
+    <xs:complexType name="Optional-parameter-operationType">
+        <xs:complexContent>
+            <xs:extension xmlns:mule="http://www.mulesoft.org/schema/mule/core" base="mule:abstractOperatorType">
+                <xs:sequence minOccurs="0" maxOccurs="1"></xs:sequence>
+                <xs:attribute type="mule:expressionString" use="optional" name="optionalParameter"></xs:attribute>
+                <xs:attribute type="xs:string" use="optional" name="target">
+                    <xs:annotation>
+                        <xs:documentation>The name of a variable on which the operation's output will be placed</xs:documentation>
```
pablolagreca: Fix javadoc

modules/spring-config/src/main/resources/META-INF/mule-module.xsd
```diff
@@ -219,4 +224,20 @@
             <xsd:enumeration value="PRIMARY"/>
         </xsd:restriction>
     </xsd:simpleType>
+
+    <xsd:simpleType name="useType">
+        <xsd:annotation>
+            <xsd:documentation>
+                Set of defined uses for a given property/parameter.
+                REQUIRED marks the attribute as required in the XSD, failing if leaved empty when consuming the parameter/property. It can not be REQUIRED if the parameter/property has a defaultValue;
```
pablolagreca: I would remove the XSD part of the description. This is meant to be used within Studio as inline doc. when you are using the attribute in the UI. I don't think that it matters to the user that it will be required in the XSD. I think they would like to just understand the meaning of what's going to happen when working with the operation they created. What do you think?.

Same with OPTIONAL

modules/extensions-xml-support/src/test/resources/modules/module-param-default-types.xsd
fernandezlautaro: that's auto generated by the SDK and I believe it's ok.
What would you change?

PR #4195, opened 2017-03-02
MULE-10651: Remove HTTP module

elrodro83:
* Cut dependency between the module and the extension


elrodro83:
Do not merge directly, build pending

PR #4202, opened 2017-03-03
MULE-11909 Log Exceptions

rscadrde:
https://www.mulesoft.org/jira/browse/MULE-11909

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/context/notification/Sender.java
+++ b/core/src/main/java/org/mule/runtime/core/context/notification/Sender.java
@@ -8,12 +8,16 @@ package org.mule.runtime.core.context.notification;
 
 import org.mule.runtime.core.api.context.notification.ServerNotification;
 import org.mule.runtime.core.routing.filters.WildcardFilter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * This does the work necessary to deliver events to a particular listener. It is generated for a particular {@link Configuration}
  * and stored in a {@link org.mule.runtime.core.context.notification.Policy}.
  */
 class Sender {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(Sender.class);
 
   private ListenerSubscriptionPair pair;
   private WildcardFilter subscriptionFilter;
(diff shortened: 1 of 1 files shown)


rscadrde:
Because of 

https://dev.eclipse.org/sonar/coding_rules#rule_key=squid%3AS2221|s=createdAt|asc=false

Catching Exception seems like an efficient way to handle multiple possible exceptions. Unfortunately, it traps all exception types and so both checked and runtime exceptions, casting too broad a net. Indeed, was it really the intention of developers to also catch runtime exceptions? To prevent any misunderstanding, if both checked and runtime exceptions are really expected to be caught, they should be explicitly listed in the catch clause.
Noncompliant Code Example
try {
  // do something that might throw an UnsupportedDataTypeException or UnsupportedEncodingException
} catch (Exception e) { // Noncompliant
  // log exception ...
}

core/src/main/java/org/mule/runtime/core/context/notification/Sender.java
```diff
@@ -29,8 +33,12 @@ public void dispatch(ServerNotification notification, NotifierCallback notifier)
         || (null != notification.getResourceIdentifier() && subscriptionFilter.accept(notification.getResourceIdentifier()))) {
       try {
         notifier.notify(pair.getListener(), notification);
-      } catch (Exception e) {
+      } catch (RuntimeException e) {
```
rscadrde: Because of 

https://dev.eclipse.org/sonar/coding_rules#rule_key=squid%3AS2221|s=createdAt|asc=false

_Catching Exception seems like an efficient way to handle multiple possible exceptions. Unfortunately, it traps all exception types and so both checked and runtime exceptions, casting too broad a net. Indeed, was it really the intention of developers to also catch runtime exceptions? To prevent any misunderstanding, if both checked and runtime exceptions are really expected to be caught, they should be explicitly listed in the catch clause.
Noncompliant Code Example
try {
  // do something that might throw an UnsupportedDataTypeException or UnsupportedEncodingException
} catch (Exception e) { // Noncompliant
  // log exception ...
}_

aiannucci:
Closing this because it's outdated, Sender is internal and not catching exceptions anymore.

PR #4205, opened 2017-03-03
MULE-11917: Error message evaluating a Dataweave Expression is not shown in the app log.

anosenzo:


core/src/main/resources/META-INF/services/org/mule/runtime/core/i18n/core-messages.properties
```diff
@@ -278,7 +278,7 @@
 332=The Object Store named "{0}" is not registered with Mule.
 333=The Session Property "{0}" is not serializable, it will not be preserved as part of the MuleSession.
 334=The Session Property "{0}" is not serializable and will be lost when MuleSession is serialized or propagated over a transport.
-335=Execution of the expression "{0}" failed.
+335=Execution of the expression "{0}" failed with this error: "{1}".
```
elrodro83: `"{1}" evaluating expression "{0}"`

elrodro83:
Add a test case for this too.

PR #4206, opened 2017-03-03
MULE-11612: ImplicitConnectionProviderValueResolver doesn't communica…

pablomusumeci:
…te the dynamism correctly


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultImplicitConnectionProviderFactory.java
```diff
@@ -12,47 +12,67 @@
...
+      synchronized (this) {
+        if (resolverSet == null) {
+          connectionProviderModel = getFirstImplicit(getAllConnectionProviders(extensionModel, configurationModel));
+
+          if (connectionProviderModel == null) {
+            throw new IllegalStateException(format(
+                                                   "Configuration '%s' of extension '%s' does not define a connection provider and none can be created automatically. Please define one.",
+                                                   configurationModel.getName(), configurationModel.getName()));
+          }
+          resolverSet = buildImplicitResolverSet(connectionProviderModel, muleContext);
```
marianogonzalez: after this line, add `resolverSetProvider = () -> resolverSet so that the synchronized block doesn't add no longer needed contention

PR #4207, opened 2017-03-03
MULE-11918: Credentials are shown in generic db connection when an exception is raised

fsgonz:


core/src/main/java/org/mule/api/util/CredentialsMaskUtil.java
```diff
@@ -0,0 +1,75 @@
...
+
+package org.mule.api.util;
+
+import static java.lang.String.format;
+import static java.util.regex.Pattern.compile;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class CredentialsMaskUtil
```
marianogonzalez: jdoc in class and public methods

PR #4211, opened 2017-03-06
MULE-11924: After multiform request, Mule Message has an invalid data type

facundovs:
… type.

When a multiform request is sent through the HTTP Requester, the payload is set to null, but the data type is set to "multiform data". It is a issue, because when the Http response is built in the HTTP Listener, It takes the multiform data type from the mule message, but without the expected boundary that is necesary to process the attachment. Therefore, an missing boundary exception is triggered.
If a Payload is set to NullPayload, content-type of transports should be ignored in the mule message.

### DIFF
--- a/modules/http/src/test/java/org/mule/module/http/functional/HttpInboundAttachmentTestCase.java
+++ b/modules/http/src/test/java/org/mule/module/http/functional/HttpInboundAttachmentTestCase.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+
+package org.mule.module.http.functional;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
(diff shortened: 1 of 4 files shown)


elrodro83:
Shouldn't the response have a `204 - No Content` status when it is built for a massage with null payload?

modules/http/src/test/java/org/mule/module/http/functional/HttpInboundAttachmentTestCase.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Test
+    public void testStateOfMessageAfterMultiFormRequest() throws Exception
+    {
+        runFlow("testFlowRequester");
+        MuleClient client = muleContext.getClient();
+        MuleMessage result = client.request("vm://out", RECEIVE_TIMEOUT);
+        assertThat(result.getPayload(), is((Object) NullPayload.getInstance()));
+        assertThat(result.getDataType().getType(), is(equalTo((Class)NullPayload.class)));
```
elrodro83: use an `InstaceOf` matcher

facundovs: I tried it, but getType() returns java.lang.class, not a object that is instance of this class.

facundovs:
Regarding to the HTTP Status, I'll create [MULE-11926](https://www.mulesoft.org/jira/browse/MULE-11926) to consider it in future versions.

PR #4212, opened 2017-03-06
MULE-11849 Convert ExpressionExecutor to Service

machaval:


distributions/standalone/assembly.xml
```diff
@@ -16,6 +16,13 @@
                 <exclude>lib/boot/exec/**</exclude>
             </excludes>
         </fileSet>
+        <fileSet>
```
elrodro83: why do you need this? there is already a dependency set for services, around line 170

distributions/standalone/pom.xml
```diff
@@ -47,14 +46,32 @@
...
                             </includeGroupIds>
                             <includeClassifiers>mule-plugin</includeClassifiers>
                             <excludeArtifactIds>mule-service-http</excludeArtifactIds>
                         </configuration>
                     </execution>
+                    <execution>
+                        <id>copy-additional-services</id>
+                        <phase>generate-sources</phase>
+                        <goals>
+                            <goal>copy</goal>
```
elrodro83: isn't this copy done by the assembly already?

distributions/standalone/pom.xml
```diff
@@ -190,14 +207,7 @@
...
         <!-- The scope provided is to avoid adding its dependencies in lib/opt -->
-        <dependency>
-            <groupId>com.mulesoft.weave</groupId>
-            <artifactId>mule-plugin-weave</artifactId>
-            <version>${weaveVersion}</version>
-            <classifier>dist</classifier>
-            <type>zip</type>
-            <scope>provided</scope>
-        </dependency>
+
```
elrodro83: remove the comments associated with the dependency too

PR #4215, opened 2017-03-06
MULE-11823: disableValidation parameter fails to be declared in XML

alegmarra:


tests/dsl/src/test/resources/integration-multi-config-dsl-app.xml
```diff
@@ -68,6 +68,13 @@
                 #[mel:['description' : payload]]
             </db:input-parameters>
         </db:insert>
+
+        <sockets:send-and-receive target="myVar">
+         <repeatable-in-memory-stream bufferSizeIncrement="256"
```
marianogonzalez: this definition is using all default values. Use unlikely values to make sure that the test doesn't pass "by accident" due to the default values

alegmarra: initialBufferSize default is 256, and the test overrides it to 512

PR #4217, opened 2017-03-07
MULE-9039 NPE when doing multiple http outbound calls.

dfeist:


modules/http/src/main/java/com/ning/http/client/providers/grizzly/HttpTransactionContext.java
```diff
@@ -0,0 +1,315 @@
...
+import org.glassfish.grizzly.websockets.HandShake;
+import org.glassfish.grizzly.websockets.ProtocolHandler;
+
+/**
+ *
+ * @author Grizzly team
+ *
+ * This is a modified version from the original file from AHC.
+ */
+public final class HttpTransactionContext {
```
elrodro83: Can you mark with a comment what you changed here from the original code?

PR #4238, opened 2017-03-08
MULE-11940: Fix For each collection doesn't work with iterator.

facundovs:
According to Mule Documentation, It is posible to combine foreach element with split DataWeave function.
However, Currently ForEach isn't supporting iterators as Input Type.


elrodro83:
It seems the commit with the changes doesn't have the correct base branch

PR #4249, opened 2017-03-09
MULE-11947: It is not possible to create global sftp endpoints with same config and different names.

fsgonz:


transports/sftp/src/main/java/org/mule/transport/sftp/SftpConnector.java
```diff
@@ -638,4 +639,11 @@ public void disconnect() throws Exception
     {
         setConnected(false);
     }
+    
+    @Override
+    protected Object getReceiverKey(FlowConstruct flowConstruct, InboundEndpoint endpoint)
```
elrodro83: can you check this is not required in other transports?

fsgonz:
A more general PR that changes the keyReceiver implementation in the AbstractConnector class will be used.

PR #4251, opened 2017-03-10
MULE-11418: Improve JMS Extension acknowledgement mechanism

estebanwasinger:


estebanwasinger:
Also fixes:
https://www.mulesoft.org/jira/browse/MULE-11855
https://www.mulesoft.org/jira/browse/MULE-11533

extensions/jms/src/main/java/org/mule/extensions/jms/JmsExtension.java
```diff
@@ -47,6 +49,7 @@
 @Xml(prefix = "jmsn")
```
marianogonzalez: old transport is already out of the distribution. Confirm with @elrodro83 if it's ok to change this to simply `jms`

extensions/jms/src/main/java/org/mule/extensions/jms/JmsSessionManager.java
```diff
@@ -0,0 +1,128 @@
...
+   * @param jmsLock the optional {@link JmsListenerLock} to be able to unlock the {@link JmsListener}
+   * @throws IllegalArgumentException if no Session was registered with the given AckId
+   */
+  public void registerMessageForAck(String ackId, Message message, Session session, JmsListenerLock jmsLock) {
+    if (!pendingSessions.containsKey(ackId)) {
+      pendingSessions.put(ackId, new SessionInformation(message, session, jmsLock));
+    }
+
+    if (LOGGER.isDebugEnabled()) {
+      LOGGER.debug(format("Registered Message for Session AckId [%s]", ackId));
```
marianogonzalez: use simple concatenation for simple messages like this. Just found out that concatenation is faster that format() in JDK8 (at least for simple cases)

extensions/jms/src/main/java/org/mule/extensions/jms/JmsSessionManager.java
```diff
@@ -0,0 +1,128 @@
...
+   * identified by the {@code ackId}
+   *
+   * @param ackId the id associated to the {@link Session} that should be ACKed
+   * @throws JMSException if an error occurs during the ack
+   */
+  public void doAck(String ackId) throws JMSException {
+    SessionInformation sessionInformation = pendingSessions.get(ackId);
+
+    if (sessionInformation == null) {
+      throw new JmsAckException(format("No pending acknowledgement with ackId [%s] exists in this Connection", ackId));
```
marianogonzalez: should this be an exception or simply a log?

extensions/jms/src/main/java/org/mule/extensions/jms/JmsSessionManager.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  /**
+   * Executes the {@link Message#acknowledge} on the latest {@link Message} associated to the {@link Session}
+   * identified by the {@code ackId}
+   *
+   * @param ackId the id associated to the {@link Session} that should be ACKed
+   * @throws JMSException if an error occurs during the ack
+   */
+  public void doAck(String ackId) throws JMSException {
```
marianogonzalez: why `doAck()` instead of simply `ack()`. The `do` prefix is often use to indicate a template method implementation

extensions/jms/src/main/java/org/mule/extensions/jms/JmsSessionManager.java
```diff
@@ -0,0 +1,128 @@
...
+    SessionInformation sessionInformation = pendingSessions.get(ackId);
+
+    if (sessionInformation == null) {
+      throw new JmsAckException(format("No pending acknowledgement with ackId [%s] exists in this Connection", ackId));
+    }
+
+    sessionInformation.getMessage().acknowledge();
+    if (LOGGER.isDebugEnabled()) {
+      LOGGER.debug(format("Acknowledged Message for Session with AckId [%s]", ackId));
```
marianogonzalez: formats... all over

extensions/jms/src/main/java/org/mule/extensions/jms/JmsSessionManager.java
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  /**
+   * Executes the {@link Session#recover()} over the {@link Session} identified by the {@code ackId}
+   *
+   * @param ackId the id associated to the {@link Session} used to create the {@link Message}
+   * @throws JMSException if an error occurs during recovering the session
+   */
+  public void recoverSession(String ackId) throws JMSException {
+    SessionInformation sessionInformation = pendingSessions.get(ackId);
```
marianogonzalez: thrid time i see this code. Make a private getPendingSession() method which does the get and the exception check

extensions/jms/src/main/java/org/mule/extensions/jms/api/config/JmsConfig.java
```diff
@@ -33,24 +37,31 @@
...
   @Expression(NOT_SUPPORTED)
   private String encoding;
 
   /**
-   * the content type of the {@link Message} {@code body}
+   * The default content type of the {@link Message} {@code body} to consume if the message doesn't communicate it
    */
   @Parameter
   @Expression(NOT_SUPPORTED)
-  @Optional(defaultValue = "*/*")
```
marianogonzalez: why no longer optional

extensions/jms/src/main/java/org/mule/extensions/jms/api/config/JmsConfig.java
```diff
@@ -90,4 +104,7 @@ public JmsProducerConfig getProducerConfig() {
     return producerConfig;
   }
 
+  public String getDefaultEncoding() {
+    return muleContext.getConfiguration().getDefaultEncoding();
```
marianogonzalez: why don't you use the encoding parameter you already defined above?

extensions/jms/src/main/java/org/mule/extensions/jms/api/config/JmsConfig.java
```diff
@@ -73,6 +84,9 @@
   @Placement(tab = "Producer")
   private JmsProducerConfig producerConfig;
 
+  public JmsSessionManager getSessionManager() {
+    return sessionManager;
```
marianogonzalez: why this getter? whoever needs this can simply inject it

extensions/jms/src/main/java/org/mule/extensions/jms/api/operation/JmsAcknowledge.java
```diff
@@ -8,57 +8,80 @@
...
    * to a session of the current connection
    */
   @Throws(JmsAckErrorTypeProvider.class)
-  public void ack(@Connection JmsConnection connection, @Summary("The AckId of the Message to ACK") String ackId) {
-
+  public void ack(@Summary("The AckId of the Message to ACK") String ackId) {
     try {
       if (LOGGER.isDebugEnabled()) {
-        LOGGER.debug("Performing ACK on session: " + ackId);
+        LOGGER.error("Performing ACK on session: " + ackId);
```
marianogonzalez: debug

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/DefaultJmsListenerLock.java
```diff
@@ -0,0 +1,71 @@
...
+ * Default implementation of {@link JmsListenerLock} based in the usage of {@link Semaphore} to support
+ * the capability of locking and unlocking the listener from different {@link Thread threads}
+ *
+ * @since 4.0
+ * @see JmsListenerLock
+ * @see JmsListener
+ */
+public class DefaultJmsListenerLock implements JmsListenerLock {
+
+  private Semaphore semaphore = new Semaphore(0);
```
marianogonzalez: why a semaphore? Looks like a simple ReentrantLock is enough for this

extensions/jms/src/test/resources/mimeType/jms-mime-type-propagation.xml
```diff
@@ -0,0 +1,51 @@
...
+        <jmsn:publish-consume destination="${destination}">
+            <jmsn:message-builder >
+                <jmsn:reply-to destination="${destination}" destinationType="QUEUE"/>
+            </jmsn:message-builder>
+        </jmsn:publish-consume>
+    </flow>
+
+    <flow name="publishConsumer-with-override">
+        <jmsn:publish-consume destination="${destination}" contentType="text/json" encoding="US-ASCII">
+            <jmsn:message-builder >
```
marianogonzalez: consider this message builder to be `showInDsl=false`, so that this is consistent with the changes that http recently went through

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsListener.java
```diff
@@ -116,26 +112,36 @@
 
   @Parameter
   @Optional
+  @Example(EXAMPLE_CONTENT_TYPE)
   @Summary("The content type of the message body")
```
marianogonzalez: all of these fields are missing jdocs. Summary is completementary to the Jdoc, but not a replacement of it. Otherwise the generated docs will suck

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsListener.java
```diff
@@ -116,26 +112,36 @@
...
   private String contentType;
 
   @Parameter
   @Optional
+  @Example(EXAMPLE_ENCODING)
   @Summary("The encoding of the message body")
   private String encoding;
 
+  @Parameter
+  @Optional(defaultValue = "true")
```
marianogonzalez: if this a good default value? This means that by default you get crappy which only makes sense if what's to gain from that is needed in most scenarios

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/DefaultJmsListenerLock.java
```diff
@@ -0,0 +1,71 @@
...
+  private boolean isFailure = false;
+  private Throwable cause;
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public void lock() {
+    semaphore.acquireUninterruptibly();
```
marianogonzalez: why not simply acquire()?  This should react to Thread interruption, otherwise it can cause problems with a graceful shutdown.

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/DefaultJmsListenerLock.java
marianogonzalez: why default value of 0? I would expect 1 as a default value

extensions/jms/src/main/java/org/mule/extensions/jms/JmsExtension.java
elrodro83: go ahead and change it to just `jms`

extensions/jms/src/main/java/org/mule/extensions/jms/JmsSessionManager.java
estebanwasinger: I think that is severe enough to raise an exception, the whole action could not be executed.

extensions/jms/src/main/java/org/mule/extensions/jms/api/config/JmsConfig.java
estebanwasinger: still optional, now It doesn't have a default value, due that is not required

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/DefaultJmsListenerLock.java
estebanwasinger: The lock should always be start blocked because since the first message the listener should wait to the flow processing finish.

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsListener.java
alegmarra: sync is less performant but more predictable and maintains 3.x-like behaviour. The user can always enable async if they don't care about mixing messages ack's

extensions/jms/src/main/java/org/mule/extensions/jms/JmsSessionManager.java
```diff
@@ -0,0 +1,129 @@
...
+   *
+   * @param ackId the id associated to the {@link Session} that should be ACKed
+   * @throws JMSException if an error occurs during the ack
+   */
+  public void ack(String ackId) throws JMSException {
+    SessionInformation sessionInformation = getSessionInformation(ackId)
+        .orElseThrow(() -> new JmsAckException(format("No pending acknowledgement with ackId [%s] exists in this Connection",
+                                                      ackId)));
+
+    sessionInformation.getMessage().acknowledge();
```
alegmarra: shouldn't we remove the session from the pending sessions?

extensions/jms/src/main/java/org/mule/extensions/jms/JmsSessionManager.java
```diff
@@ -0,0 +1,129 @@
...
+
+    Session session = sessionInformation.getSession();
+
+    sessionInformation.getJmsListenerLock().ifPresent(lock -> {
+      if (lock.isLocked()) {
+        lock.unlock();
+      }
+    });
+
+    session.recover();
```
alegmarra: probably we should recover before unlocking the listener. Are the messages that were "locked" delivered when the lock is released? If that happens, we'll be delivering messages that will be redelivered later because of the recover (since locked listener messages are not ack'd)

alegmarra: also, we have to remove the recovered session from the "pending" sessions, since no message should be ack'd if it was already recovered... right? probably get's a little bit messy in the async mode

extensions/jms/src/main/java/org/mule/extensions/jms/api/operation/JmsAcknowledge.java
```diff
@@ -8,57 +8,80 @@
...
-  public void ack(@Connection JmsConnection connection, @Summary("The AckId of the Message to ACK") String ackId) {
-
+  public void ack(@Summary("The AckId of the Message to ACK") String ackId) {
     try {
       if (LOGGER.isDebugEnabled()) {
         LOGGER.debug("Performing ACK on session: " + ackId);
       }
 
-      connection.doAck(ackId);
+      sessionManager.ack(ackId);
```
alegmarra: check for blank ackId before invoking the session manager, so the error message can be more accurate

extensions/jms/src/main/java/org/mule/extensions/jms/api/operation/JmsAcknowledge.java
```diff
@@ -8,57 +8,80 @@
...
+   * @param ackId The AckId of the Message Session to recover
+   */
+  @Throws(JmsSessionRecoverErrorTypeProvider.class)
+  public void recoverSession(String ackId) {
+    try {
+      if (LOGGER.isDebugEnabled()) {
+        LOGGER.debug("Recovering session: " + ackId);
+      }
+
+      sessionManager.recoverSession(ackId);
```
alegmarra: same for blank check

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsListener.java
```diff
@@ -172,23 +178,35 @@ public void onStop() {
     if (LOGGER.isDebugEnabled()) {
       LOGGER.debug("Stopping JMSSubscriber source");
     }
+
+    jmsLock.unlock();
+
+    try {
+      if (ackMode.equals(AUTO) || ackMode.equals(DUPS_OK) || ackMode.equals(MANUAL)) {
```
alegmarra: if we don't want to recover only for NONE, then is more explicit to check for that instead all the other options

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsListener.java
```diff
@@ -172,23 +178,35 @@ public void onStop() {
...
+    }
   }
 
   @OnSuccess
   public void onSuccess(@Optional @NullSafe JmsListenerResponseBuilder response,
                         SourceCallbackContext callbackContext) {
-
-    ackOriginalMessage(callbackContext);
-
+    jmsLock.unlock();
```
alegmarra: here we completed the processing of the flow, but I don't get where is the AUTO ack performed. Are you doing that behind the curtains?

extensions/jms/src/main/java/org/mule/extensions/jms/JmsSessionManager.java
estebanwasinger: no, because the ackId will be repeated between messages, so if the listener is configured in an async way, we could remove the `SessionInformation` of a session that currently have in-flight messages.

extensions/jms/src/main/java/org/mule/extensions/jms/api/operation/JmsConsume.java
```diff
@@ -79,8 +86,8 @@
                                                @Optional @Summary("The Type of the Consumer that should be used for the provided destination") ConsumerType consumerType,
                                                @Optional @Summary("The Session ACK mode to use when consuming a message") AckMode ackMode,
```
alegmarra: I thought we were going to restrict the AckModes for Consume, since we are not going to intercept the end of the flow for auto ACK'in

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsListener.java
estebanwasinger: The Auto ACK is done automatically by the JMS Client

extensions/jms/src/main/java/org/mule/extensions/jms/JmsSessionManager.java
marianogonzalez: I think @alegmarra has a point here. I think this sessions should have some kind of timeout, and the SessionInformation instances should not be stored in a Map, they should be stored in a Guava cache which expiration is set to such timeout.

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsListener.java
```diff
@@ -172,23 +178,35 @@ public void onStop() {
     if (LOGGER.isDebugEnabled()) {
       LOGGER.debug("Stopping JMSSubscriber source");
     }
+
+    jmsLock.unlock();
+
+    try {
+      if (ackMode.equals(AUTO) || ackMode.equals(DUPS_OK) || ackMode.equals(MANUAL)) {
+        session.get().recover();
```
marianogonzalez: the old transport had the concept of how many consumers to have on each destination. When the source was stopped, the stop logic had to be done on each of those consumers. Where is that concept in this connector? Shouldn't the ACK logic be done for each of them?

estebanwasinger: > the old transport had the concept of how many consumers to have on each destination. When the source was stopped, the stop logic had to be done on each of those consumers

In this new JMS Listener, a listener means only one JMS consumer, therefore, there is no need to stop multiple consumers.
But also, the stop of the consumers/producers is done when returning the listener's connection `BaseConnectionProvider#onReturn()`

> Shouldn't the ACK logic be done for each of them?

Not sure to understand, but if the Listeners is being stopped, we can not guarantee that the in-flight messages have been correctly consumed, so we can't execute an Acknowledgement over the session.

extensions/jms/src/main/java/org/mule/extensions/jms/JmsSessionManager.java
alegmarra: We have to manually handle the removal of the sessions, since no eviction policy would be completely valid for that session cache. Also, we need to handle the removal for session's recovery, so the problem would still be there.

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsListener.java
alegmarra: forgot about that =)

extensions/jms/src/main/java/org/mule/extensions/jms/api/operation/JmsConsume.java
estebanwasinger: Ticket: https://www.mulesoft.org/jira/browse/MULE-11963

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsListener.java
estebanwasinger: This is for `NONE` and `TRANSACTED`

PR #4256, opened 2017-03-10
MULE-11762: Wrong implicit config created when using an exclusive ope…

pablomusumeci:
…ration

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/extension/ExtensionManager.java
+++ b/core/src/main/java/org/mule/runtime/core/api/extension/ExtensionManager.java
@@ -6,6 +6,7 @@
  */
 package org.mule.runtime.core.api.extension;
 
+import org.mule.runtime.api.meta.model.ComponentModel;
 import org.mule.runtime.api.meta.model.ExtensionModel;
 import org.mule.runtime.api.meta.model.config.ConfigurationModel;
 import org.mule.runtime.core.api.Event;
@@ -78,14 +79,14 @@ public interface ExtensionManager {
    * Because no {@link ConfigurationProvider} is specified, the following algorithm will be applied to try and determine the
    * instance to be returned:
    * <ul>
-   * <li>If <b>one</b> {@link ConfigurationProvider} is registered, capable of handing configurations of the given
-   * {@code extensionModel}, then that provider is used</li>
+   * <li>If <b>one</b> (and only one) {@link ConfigurationProvider} is registered, capable of handing configurations of the given
+   * {@code componentModel}, then that provider is used</li>
    * <li>If more than one {@link ConfigurationProvider} meeting the criteria above is found, then a {@link IllegalStateException}
    * is thrown</li>
    * <li>If no such {@link ConfigurationProvider} is registered, then an attempt will be made to locate a
    * {@link ConfigurationModel} for which an implicit configuration can be inferred. A model can be considered implicit if all its
-   * parameters are either optional or provide a default value. If such a model is found, then a {@link ConfigurationProvider} is
-   * created and registered for that model.</li>
+   * parameters are either optional or provide a default value. If such a model is found and it is unique, then a
+   * {@link ConfigurationProvider} is created and registered for that model.</li>
    * <li>If none of the above conditions is met, then an {@link IllegalStateException} is thrown</li>
    * </ul>
    * <p>
@@ -93,22 +94,12 @@ public interface ExtensionManager {
    * updated for the returned {@link ConfigurationInstance}
    *
    * @param extensionModel the {@link ExtensionModel} for which a configuration is wanted
-   * @param event the current Event
+   * @param componentModel the {@link ComponentModel} associated to a {@link ConfigurationInstance}
+   * @param muleEvent the current Event
    * @return a {@link ConfigurationInstance}
    * @throws IllegalStateException if none or too many {@link ConfigurationProvider} are found to be suitable
    */
-  ConfigurationInstance getConfiguration(ExtensionModel extensionModel, Event event);
-
-  /**
-   * Locates and returns the {@link ConfigurationProvider} which would serve an invocation to the
-   * {@link #getConfiguration(ExtensionModel, Event)} method.
-   * <p>
-   * This means that the returned provider will be located using the same set of rules as the aforementioned method
-   *
-   * @param extensionModel the {@link ExtensionModel} for which a configuration is wanted
-   * @return an {@link Optional} {@link ConfigurationProvider}
-   */
-  Optional<ConfigurationProvider> getConfigurationProvider(ExtensionModel extensionModel);
+  ConfigurationInstance getConfiguration(ExtensionModel extensionModel, ComponentModel componentModel, Event muleEvent);
 
   /**
    * Locates and returns the {@link ConfigurationProvider} which would serve an invocation to the
@@ -122,8 +113,18 @@ public interface ExtensionManager {
   Optional<ConfigurationProvider> getConfigurationProvider(String configurationProviderName);
 
   /**
+   * Locates and returns (if there is any) a suitable {@link ConfigurationProvider} for the given {@link ComponentModel}.
+   * 
+   * @param extensionModel the {@link ExtensionModel} for which a configuration is wanted
+   * @param
+   * @return an {@link Optional} {@link ConfigurationProvider}
+   */
+  Optional<ConfigurationProvider> getConfigurationProvider(ExtensionModel extensionModel,
+                                                           ComponentModel componentModel);
+
+  /**
    * Registered the given {@link ConfigurationProvider} which should be later be used to serve invocations to
-   * {@link #getConfigurationProvider(ExtensionModel)} and {@link #getConfiguration(String, Event)}
+   * {@link #getConfigurationProvider(ExtensionModel, ComponentModel)} and {@link #getConfiguration(String, Event)}
    *
    * @param configurationProvider a {@link ConfigurationProvider}
    */
--- a/core/src/main/java/org/mule/runtime/core/api/extension/ExtensionManager.java
+++ b/core/src/main/java/org/mule/runtime/core/api/extension/ExtensionManager.java
@@ -72,9 +72,8 @@ public interface ExtensionManager {
    */
   ConfigurationInstance getConfiguration(String configurationProviderName, Event event);
 
-
   /**
-   * Returns a {@link ConfigurationInstance} for the given {@code extensionModel}.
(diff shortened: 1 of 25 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/utils/ImplicitObjectUtils.java
```diff
@@ -81,7 +81,14 @@ public static ResolverSet buildImplicitResolverSet(ParameterizedModel parameteri
         .orElse(null);
   }
 
-  private static boolean canBeUsedImplicitly(ParameterizedModel parameterizedModel) {
+  /**
+   * Returns whether the given model can be used implicitly. That means that all of its parameters which are non optional have
+   * associated a default value.
```
alegmarra: `That means that all of its parameters which are non optional have associated a default value.`
makes no sense, only optional parameters can have a default value.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/utils/ImplicitObjectUtils.java
```diff
@@ -81,7 +81,14 @@ public static ResolverSet buildImplicitResolverSet(ParameterizedModel parameteri
...
+   * 
+   * @param parameterizedModel model
+   * @return whether the given model can be used implicitly or not.
+   */
+  public static boolean canBeUsedImplicitly(ParameterizedModel parameterizedModel) {
     return parameterizedModel.getAllParameterModels().stream().noneMatch(p -> p.isRequired() && p.getDefaultValue() == null);
```
alegmarra: we should check for required only, having a default on the optional or not having it is irrelevant for implicit creation of the element. Also, a required with a non null default value should be an error in the model.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/DefaultExtensionManager.java
```diff
@@ -137,37 +139,43 @@ public void registerConfigurationProvider(ConfigurationProvider configurationPro
...
+                                                                  ConfigurationModel configurationModel) {
+    Collection<ConfigurationProvider> providers = extensionRegistry.getConfigurationProviders(extensionModel, configurationModel);
 
     int matches = providers.size();
 
     if (matches == 1) {
-      return Optional.of(providers.get(0));
+      return providers.stream().findFirst();
     } else if (matches > 1) {
       throw new TooManyConfigsException("Too many configs found", extensionModel, matches);
```
alegmarra: declare for which configurationModel there were too many options

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/ExtensionRegistry.java
```diff
@@ -50,13 +52,16 @@
...
 
-  private final LoadingCache<ExtensionModel, List<ConfigurationProvider>> providersByExtension =
-      CacheBuilder.newBuilder().build(new CacheLoader<ExtensionModel, List<ConfigurationProvider>>() {
+  private final LoadingCache<ExtensionModel, Multimap<ConfigurationModel, ConfigurationProvider>> providersByExtension =
+      CacheBuilder.newBuilder().build(new CacheLoader<ExtensionModel, Multimap<ConfigurationModel, ConfigurationProvider>>() {
 
         @Override
-        public List<ConfigurationProvider> load(ExtensionModel key) throws Exception {
-          return registry.lookupObjects(ConfigurationProvider.class).stream()
+        public Multimap<ConfigurationModel, ConfigurationProvider> load(ExtensionModel key) throws Exception {
```
alegmarra: when do we have more than one configurationProvider for the same ConfigurationModel?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultImplicitConfigurationProviderFactory.java
```diff
@@ -37,10 +36,9 @@
    */
   @Override
   public ConfigurationProvider createImplicitConfigurationProvider(ExtensionModel extensionModel,
+                                                                   ConfigurationModel implicitConfigurationModel,
                                                                    Event event,
                                                                    MuleContext muleContext) {
-    ConfigurationModel implicitConfigurationModel = getFirstImplicit(extensionModel.getConfigurationModels());
-
     if (implicitConfigurationModel == null) {
```
marianogonzalez: should also check that the model is in fact implicit. The method in the ExtnesionManager which checks that should be given a better name and moved into a utils class. Then, reuse it here.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultImplicitConfigurationProviderFactory.java
```diff
@@ -40,12 +42,14 @@ public ConfigurationProvider createImplicitConfigurationProvider(ExtensionModel
                                                                    Event event,
                                                                    MuleContext muleContext) {
     if (implicitConfigurationModel == null) {
-      throw new IllegalStateException(String.format(
-                                                    "Could not find a config for extension '%s' and none can be created automatically. Please define one",
-                                                    extensionModel.getName()));
+      throw new IllegalStateException("Could not find a config for extension '" + extensionModel.getName()
+          + "' and none can be created automatically. Please define one");
+    } else if (!canBeUsedImplicitly(implicitConfigurationModel)) {
+      throw new IllegalStateException("Configuration model '" + extensionModel.getName()
```
marianogonzalez: From the end user POV it doesn't make sense to talk about implicit configuration models. The error should be the same as in line 45

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
```diff
@@ -338,8 +342,15 @@ protected CursorStreamProviderFactory getCursorStreamProviderFactory() {
...
   }
 
+  /**
+   * Returns a (if any) {@link ConfigurationModel} that is suitable for the {@link ComponentModel}. If there are more than one
+   * {@link ConfigurationModel} that can be used by this component, the ones that can be used implicitly (according to
+   * {@link ExtensionModelUtils#canBeUsedImplicitly(ParameterizedModel)}) will be prioritized.
+   */
   private Optional<ConfigurationModel> computeConfig() {
-    return getConfigurationForComponent(extensionModel, componentModel);
+    return getConfigurationForComponent(extensionModel, componentModel).stream().sorted(reverseOrder((x, y) -> Boolean
```
marianogonzalez: Can you explain why you need this reverse sort thing instead of simply filtering for implicit and then just find first?

pablomusumeci: From the available ConfigurationModels, if there is any model that can be used implicitly pick that one (that's where the ordering is needed, and the reverse thing is because ordering booleans is the same as ordering 0 and 1). But if there are no implicit models available, just pick any other configuration model for that component.

marianogonzalez: That is the same as filter(is implicit).findFirst () only that way more inefficient. Change it

pablomusumeci: It's not the same. I still want to receive a ConfigurationModel even if is is not implicit (in the case that there is no implicit config available). Filtering would remove these non-implicit models. This way prioritizes implicits, but if there is no implicit and one non-implicit I still get that last model.

marianogonzalez: then it's even worst! You must NOT prioritize implicit over not implicit or anything. The logic is way simpler: 

1) i f the component doesn't require a config you do nothing
2) if a config is specified, you use it
3) If no config is specified and one is required, then you pick the first implicit available

You're fundamentally breaking the original behaviour. You cannot fix the bug at expense of that behaviour

alegmarra: I don't agree with "then you pick the first implicit available", if there is more than one implicit we should fail. 
Selecting one implicit over another is a different story, the one where the extension developer declares a Config as a default if more than one option is available. Otherwise we are depending on how we load the configs in the extension model for selecting a config, implying that we can change the behaviour/break an application when any change in that order changes.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
```diff
@@ -311,46 +308,48 @@ private MetadataContext getMetadataContext() throws MetadataResolvingException,
    * @return a configuration instance for the current component with a given {@link Event}
    */
   protected Optional<ConfigurationInstance> getConfiguration(Event event) {
-    if (!configurationModel.get().isPresent()) {
```
marianogonzalez: you seem to have lost this case

core/src/main/java/org/mule/runtime/core/api/extension/ExtensionManager.java
```diff
@@ -124,6 +124,25 @@
...
+   * implicitly, that will be returned. If there are more than one, an {@link IllegalStateException} will be thrown.
+   *
+   * If none of the previous conditions could be fulfilled, {@link Optional#empty()} will be returned.
+   * 
+   * @param extensionModel the {@link ExtensionModel} for which a configuration is wanted
+   * @param assignableConfigurationModels a {@link Set} of all the possible {@link ConfigurationModel} that can be used for a
+   *        given component.
+   * @return an {@link Optional} {@link ConfigurationModel}
+   */
+  Optional<ConfigurationModel> getConfigurationModelForExtension(ExtensionModel extensionModel,
```
marianogonzalez: this method should receive the ComponentModel and calculate the assignable configurations. Determining the list of assignable configs is part of the solution that this method should provide

marianogonzalez: I don't thin this method should be exposed here. It makes sense for the ExtnesionManager to know how to do this, but it shouldn't expose that behaviour. This is very internal and this interface is API

PR #4258, opened 2017-03-10
MULE-11761: Create the new Soap Service.

juandesi:


services/pom.xml
```diff
@@ -21,5 +21,6 @@
         <module>scheduler</module>
         <module>oauth</module>
         <module>services-all</module>
+        <module>soap</module>
```
elrodro83: keep services-all always last

elrodro83:
You're missing adding the service to the distro

PR #4260, opened 2017-03-12
MULE-9139 - revampo poll message source - rename poll to scheduler, r…

pablolagreca:
…emove watermark feature


core-tests/src/test/java/org/mule/runtime/core/source/polling/SchedulerMessageSourceTestCase.java
```diff
@@ -0,0 +1,101 @@
...
+  public void emptyStringResponseFromNestedMP() throws Exception {
+
+    SchedulerMessageSource schedulerMessageSource = createMessageSource();
+
+    SensingNullMessageProcessor flow = getSensingNullMessageProcessor();
+    schedulerMessageSource.setListener(flow);
+
+    schedulerMessageSource.poll();
+
+    assertNotNull(flow.event);
```
marianogonzalez: the assertion doesn't match the test description.

marianogonzalez: hamcrest

core-tests/src/test/java/org/mule/runtime/core/source/polling/SchedulerMessageSourceTestCase.java
```diff
@@ -0,0 +1,101 @@
...
+  private SchedulerMessageSource schedulerMessageSource;
+
+  @After
+  public void after() throws MuleException {
+    stopIfNeeded(schedulerMessageSource);
+    disposeIfNeeded(schedulerMessageSource, logger);
+  }
+
+  @Test
+  public void setExecutionClassLoader() throws Exception {
```
marianogonzalez: i don't see how this test is any different from `emptyStringResponseFromNestedMP()`. Which assertion depends on the execution classloader?

modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -2707,20 +2707,19 @@
...
+    <xsd:complexType name="schedulerMessageSourceType">
         <xsd:complexContent>
             <xsd:extension base="abstractMessageSourceType">
                 <xsd:sequence>
                     <xsd:element ref="abstract-scheduler" minOccurs="0" maxOccurs="1"/>
-                    <xsd:element ref="abstract-poll-override" minOccurs="0" maxOccurs="1"/>
-                    <xsd:group ref="messageProcessorOrMixedContentMessageProcessor"/>
+                    <xsd:element ref="abstract-scheduler-override" minOccurs="0" maxOccurs="1"/>
                 </xsd:sequence>
                 <xsd:attribute name="frequency"  type="substitutableLong" default="1000">
```
dfeist: Make sense to clean things up and introduce a simple-scheduler child element with frequency, to avoid needing the exclusiveness checks to ensure `frequency` fails if used at same time as a scheduler

modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -820,7 +820,7 @@
     </xsd:complexType>
 
 
-    <xsd:element name="watermark" type="watermarkType" substitutionGroup="abstract-poll-override">
+    <xsd:element name="watermark" type="watermarkType" substitutionGroup="abstract-scheduler-override">
```
dfeist: Why is this still in schema if watermark was removed?  Do we still need scheduler overrides even now?

core-tests/src/test/java/org/mule/runtime/core/source/polling/SchedulerMessageSourceTestCase.java
marianogonzalez: you forgot this one on the last commit

PR #4264, opened 2017-03-13
MULE-11961: Prepare HTTP service for extension and fix tests cases

afelisatti:


services/http/src/main/java/org/mule/services/http/impl/service/HttpServiceImplementation.java
```diff
@@ -32,7 +32,7 @@
 
   private static final Logger logger = LoggerFactory.getLogger(HttpServiceImplementation.class);
 
-  private final SchedulerService schedulerService;
+  protected final SchedulerService schedulerService;
```
elrodro83: why do you need this change?

afelisatti: So that it's visible to subclasses.

PR #4271, opened 2017-03-13
MULE-11966: JmxAgent registers objects using the object name and not the original registry name

fsgonz:


modules/management/src/test/java/org/mule/management/mbeans/issues/MBeanRegistrationPreservesRegistryNameTestCase.java
```diff
@@ -0,0 +1,81 @@
...
+        jmxSupport = jmxAgent.getJmxSupportFactory().getJmxSupport();
+        domainName = jmxSupport.getDomainName(muleContext);
+        mBeanServer = jmxAgent.getMBeanServer();
+    }
+
+
+    @Test
+    public void whenAnMBeanIsRegisteredThenNameIsOriginalRegistryName() throws Exception
+    {
+        Flow flow = muleContext.getRegistry().get(FLOW_ORIGINAL_REGISTRY_KEY);
```
elrodro83: this doesn't seem to be right. The property is resolved and the resolved value ends up in the name, whether it's the flow name of the bean name.

modules/management/src/main/java/org/mule/module/management/agent/AbstractJmxAgent.java
```diff
@@ -427,9 +427,9 @@ protected void registerServiceServices() throws NotCompliantMBeanException, MBea
     protected void registerFlowConstructServices() throws NotCompliantMBeanException, MBeanRegistrationException,
         InstanceAlreadyExistsException, MalformedObjectNameException
     {
-        for (AbstractFlowConstruct flowConstruct : muleContext.getRegistry().lookupObjects(AbstractFlowConstruct.class))
+        for (String  rawName : muleContext.getRegistry().lookupByType(AbstractFlowConstruct.class).keySet())
```
elrodro83: Wouldn't this happen in other cases apart from flows?

modules/management/src/test/java/org/mule/management/mbeans/issues/MBeanRegistrationPreservesRegistryNameTestCase.java
fsgonz: Spring considers name and id as identifiers, so it does not register the beans with a placeholder resolved. A flow is registered as a spring bean when the xml is read and the bean definition name is used as a key and that it is not resolved. The java bean attributes placeholder are resolved but not the definition name which is the one to use if one wants to retrieve from the springregistry (beanFactory().getBean(key))

modules/management/src/main/java/org/mule/module/management/agent/AbstractJmxAgent.java
fsgonz: Currently that is not a problem as only the managed beans for the flow type tries to retrieve something from the spring registry in its postconstruct. Anyway, I've pushed changes so that the case of connectors and model/services use the spring bean definition name as the name of the jmx managed bean. This is for coherence purposes. Connectors cannot set placeholders in its names as the name property in mule.xsd is of substitutableName type which has a restriction propertyPlaceHolderType.

PR #4272, opened 2017-03-14
MULE-11603 Add DW performance tests. Add single very limited test to …

dfeist:
…get initial number.

### DIFF
--- a/tests/performance/pom.xml
+++ b/tests/performance/pom.xml
@@ -111,6 +111,12 @@
             <classifier>mule-service</classifier>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mule.services</groupId>
+            <artifactId>mule-service-weave</artifactId>
+            <classifier>mule-service</classifier>
+            <version>${project.version}</version>
+        </dependency>
 
         <!-- Plugins -->
         <dependency>
--- a/tests/performance/src/main/java/org/mule/el/SimpleExpressionBenchmark.java
+++ b/tests/performance/src/main/java/org/mule/el/SimpleExpressionBenchmark.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.el;
+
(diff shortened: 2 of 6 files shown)


tests/performance/pom.xml
```diff
@@ -111,6 +111,12 @@
             <classifier>mule-service</classifier>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mule.services</groupId>
+            <artifactId>mule-service-weave</artifactId>
+            <classifier>mule-service</classifier>
+            <version>${project.version}</version>
+        </dependency>
```
pablokraan: provided?

tests/performance/src/main/java/org/mule/el/SimpleExpressionBenchmark.java
```diff
@@ -0,0 +1,53 @@
...
+  @Param({"mel:payload", "payload"})
+  private String expression;
+
+  private MuleContext muleContext;
+  private Event event;
+
+  @Setup
+  public void setup() throws MuleException {
+    muleContext = createMuleContextWithServices();
+    event = Event.builder(DefaultEventContext.create(createFlow(muleContext), ""))
```
pablokraan: Use static import

tests/performance/src/test/java/org/mule/test/core/el/ExpressionBenchmarkAssertionTestCase.java
```diff
@@ -0,0 +1,39 @@
...
+  public void evaluatePayloadMEL() {
+    runAndAssertBenchmark(SimpleExpressionBenchmark.class, "evaluatePayload", 1, singletonMap(EXPRESSION_PARAM,
+                                                                                              new String[] {
+                                                                                                  "mel:payload"}),
+                          700, NANOSECONDS, 2000);
+  }
+
+  @Test
+  public void evaluatePayloadDW() {
+    // DW numbers are significantly below MEL currently but by adding assertions we avoid any regressions
```
pablokraan: There should be a reference to the jira tracking this problem here.

dfeist: Yep, I'll check with @afelisatti  to see if we have one..

tests/performance/pom.xml
dfeist: If i put it provided then it's not included in the benchmark uber jar.

PR #4277, opened 2017-03-14
MULE-11973: PGP secret key should not be needed for encryption

fsgonz:


modules/pgp/src/test/java/org/mule/module/pgp/PGPWithoutSecretKeyTestCase.java
```diff
@@ -0,0 +1,89 @@
...
+        assertThat(message, is(notNullValue()));
+
+    }
+
+    @Test
+    public void whenNoSecretKeyFilenameProvidedAndDecryptThenFlowError() throws Exception
+    {
+        MuleClient client = muleContext.getClient();
+
+        client.send("vm://inDecrypt", new DefaultMuleMessage("encryptedtext", muleContext));
```
alepulver: Use runFlow instead of a vm inbound endpoint and MuleClient.

modules/pgp/src/test/resources/file-decryption-config-no-secret-key.xml
```diff
@@ -0,0 +1,55 @@
...
+		<pgp:keybased-encryption-strategy
+			name="keyBasedEncryptionStrategy" credentialsAccessor-ref="fakeCredentialAccessor"
+			keyManager-ref="pgpKeyManager" />
+	</pgp:security-manager>
+
+	<flow name="pgpDecryptProcessor">
+
+		<vm:inbound-endpoint path="inDecrypt" />
+
+		<vm:outbound-endpoint path="outDecrypt">
```
alepulver: Can you avoid using vm outbound endpoint by using the returned payload? Maybe even in the custom processor.

PR #4278, opened 2017-03-14
DEVKIT-2238-Added the possibility to use a system variable instead of…

emil-sv-mule:
… the constant


pablokraan:
@emil-sv-mule thanks for contributing. First: please read the contribution guide and sign up the contribution agreement.
Second, we are working on a fix for this in order to make the separator configurable per connector, ie: no system property. Is Ok if we close this PR?

PR #4285, opened 2017-03-14
MULE-10651: Remove HTTP module - Delete obsolete classes, move used o nes and leave empty module.

afelisatti:


afelisatti:
jenkins validate

PR #4289, opened 2017-03-15
MULE-11980: Apply package renaming from compatibility plugin

pablokraan:


modules/spring-config/src/main/java/org/mule/runtime/config/spring/parsers/specific/NotificationDefinitionParser.java
```diff
@@ -51,7 +51,7 @@
   private static final String ENDPOINT_MESSAGE_NOTIFICATION_CLASS =
       "org.mule.compatibility.core.context.notification.EndpointMessageNotification";
```
elrodro83: rename this too

PR #4293, opened 2017-03-15
MULE-11983: Upgrade JSCH to version 0.154

anosenzo:


pom.xml
```diff
@@ -193,7 +193,7 @@
         <jdomVersion>2.0.2</jdomVersion>
         <jettyVersion>9.0.7.v20131107</jettyVersion>
         <jodaTimeVersion>2.9.1</jodaTimeVersion>
-        <jschVersion>0.1.51</jschVersion>
+        <jschVersion>0.1.54</jschVersion>
```
marianogonzalez: need to add this on the changelog as well. You also need a EE PR in which the EE changelog is also modified

PR #4294, opened 2017-03-15
MULE-11398: remove an incorrect validation previously added

alepulver:


modules/spring-config/src/main/java/org/mule/runtime/config/spring/dsl/spring/MapEntryBeanDefinitionCreator.java
```diff
@@ -76,8 +74,6 @@ boolean handleRequest(CreateBeanDefinitionRequest createBeanDefinitionRequest) {
       value = getValue(objectTypeVisitor, componentModel, componentBuildingDefinition);
     } else if (componentModel.getParameters().get(ENTRY_TYPE_VALUE_REF_PARAMETER_NAME) != null) {
       value = new RuntimeBeanReference(componentModel.getParameters().get(ENTRY_TYPE_VALUE_REF_PARAMETER_NAME));
-    } else {
-      throw new MuleRuntimeException(createStaticMessage("Unknown bean type for map entry"));
```
marianogonzalez: this seems to be attacking the symptom rather than the problem. Isn't this actually an exceptional condition? Why does the code enter this path in the first place?

PR #4296, opened 2017-03-15
MULE-11946: references to global components work properly when macro expanded

fernandezlautaro:


modules/extensions-xml-support/pom.xml
```diff
@@ -137,6 +139,13 @@
             <classifier>mule-plugin</classifier>
         </dependency>
         <dependency>
+            <groupId>org.mule.modules</groupId>
+            <artifactId>mule-module-jms</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
```
pablolagreca: Should plugins be provided or compile in this scope?

fernandezlautaro: actually, they should be `<scope>test</scope>`. But the runner ask them to be `<scope>provided</scope>` or it fails

PR #4302, opened 2017-03-16
MULE-10651: Remove HTTP module entirely

afelisatti:


afelisatti:
DO NOT MERGE - We need an empty module jar deployed before removing it.

PR #4304, opened 2017-03-16
MULE-11989: Generic DB Config should accept user and password as attributes.

fsgonz:


modules/db/src/test/resources/integration/config/generic-db-config.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<mule xmlns="http://www.mulesoft.org/schema/mule/core"
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xmlns:db="http://www.mulesoft.org/schema/mule/db"
+      xsi:schemaLocation="http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd
+            http://www.mulesoft.org/schema/mule/db http://www.mulesoft.org/schema/mule/db/current/mule-db.xsd">
+
+     <db:generic-config name="dbConfig" user="userTest" password="passwordTest" url="jdbc:derby:muleEmbeddedDB;sql.enforce_strict_size=true" driverClassName="org.apache.derby.jdbc.EmbeddedDriver"/>
```
pablolagreca: should this be user or username?

pablolagreca: How do other -config elements provide username and password?

fsgonz: The other config use the attribute "user". Actually, instead of using the type DatabaseConfigType for the generic config, I've changed the xsd so that it uses abstractUserAndPasswordDatabaseConfigType, on which the other configs (oracle, derby, etc.) are based (abstractUserAndPasswordDatabaseConfigType is the xml schema type that adds "user" and "password" elements).

PR #4311, opened 2017-03-17
MULE-11992: An Exception is raised if reconnect-notifier is used within a global configuration.

fsgonz:


modules/spring-config/src/main/java/org/mule/config/spring/parsers/specific/RetryNotifierDefinitionParser.java
```diff
@@ -19,5 +23,29 @@ public RetryNotifierDefinitionParser(Class clazz)
...
+    }
+
+    private boolean isConfigElement(Element element)
+    {
+        return getAncestorBeanName(element).equals(OBJECT_MULE_CONFIGURATION);
+    }
+
+    private String getAncestorBeanName(Element element)
+    {
+        return ((Element) ((Element) element.getParentNode()).getParentNode()).getAttribute(ATTRIBUTE_NAME);
```
elrodro83: could there be more than 2 levels of ancestry?

modules/spring-config/src/main/java/org/mule/config/spring/parsers/specific/RetryNotifierDefinitionParser.java
```diff
@@ -19,5 +23,29 @@ public RetryNotifierDefinitionParser(Class clazz)
     {
         super("notifier", clazz);
     }
+    
+    @Override
+    protected boolean isChild(Element element, ParserContext parserContext, BeanDefinitionBuilder builder)
+    {
+        if (isConfigElement(element))
```
elrodro83: make this in one line

tests/integration/src/test/java/org/mule/config/spring/handlers/RetryInGlobalConfigurationTestCase.java
```diff
@@ -0,0 +1,49 @@
...
+        Connector c = muleContext.getRegistry().lookupConnector("testConnector");
+        assertThat(c, notNullValue());
+
+        RetryPolicyTemplate rpf = c.getRetryPolicyTemplate();
+        assertThat(rpf, notNullValue());
+        RetryNotifier rn = rpf.getNotifier();
+        assertThat(rn, notNullValue());
+        assertThat(rn, instanceOf(ConnectNotifier.class));
+
+        assertThat(c.isConnected(), equalTo(TRUE));
```
elrodro83: why the constant instead of just `true`?

modules/spring-config/src/main/java/org/mule/config/spring/parsers/specific/RetryNotifierDefinitionParser.java
fsgonz: No. In case it is in the global configuration, this should have this form:

<configuration>
<reconnect frequency="500">
<reconnect-notifier/>
</reconnect>
</configuration>

A same strategy is used to determine if it is a global configuration for the scripting parser.

tests/integration/src/test/java/org/mule/config/spring/handlers/RetryInGlobalConfigurationTestCase.java
fsgonz: I've changed it. A legacy when boxing/autoboxing was not automatically performed. It has no sense now (it had no sense then either for this particular testCase as the method return primitives and not the wrapper type).

PR #4315, opened 2017-03-17
MULE-11996: Error with Dataweave transformation with XML and then use…

pablokraan:
… a Logger

_ Exporting additional packages on XML module


modules/xml/src/main/resources/META-INF/mule-module.properties
```diff
@@ -10,7 +10,22 @@ artifact.export.classPackages=org.mule.runtime.module.xml.config,\
                               org.mule.runtime.module.xml.transformer.wire,\
                               org.mule.runtime.module.xml.util,\
                               org.mule.runtime.module.xml.xpath,\
+                              com.ctc.wstx.api,\
```
elrodro83: should we also be exporting the packages from the woodstox dependencies? (stax2-api)

PR #4317, opened 2017-03-17
MULE-10363: remove 'oauthContext()' MEL expression in favour of provi…

fernandezlautaro:
…ding operations


fernandezlautaro:
@pablolagreca @elrodro83 
Not sure how are we handling documentation for existing stuff in Mule4, but https://docs.mulesoft.com/mule-user-guide/v/3.6/authentication-in-http-requests#access-authorization-information-through-the-token-manager will be deprecated once this PR gets merged.

extensions/oauth/src/test/resources/operations/operations-config.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<mule xmlns="http://www.mulesoft.org/schema/mule/core"
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xmlns:oauth="http://www.mulesoft.org/schema/mule/oauth"
+      xsi:schemaLocation="
+       http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd
+       http://www.mulesoft.org/schema/mule/oauth http://www.mulesoft.org/schema/mule/oauth/current/mule-oauth.xsd">
+
+    <flow name="accessTokenFlow">
```
elrodro83: You could use the extensions client to call on operation rather than declaring a flow for that.

modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/state/package.java
```diff
@@ -1,11 +0,0 @@
...
- * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
- * The software in this package is published under the terms of the CPAL v1.0
- * license, a copy of which has been included with this distribution in the
- * LICENSE.txt file.
- */
-
-/**
- * This package contains classes that may be persisted with an application classloader, and because of that have to exported.
- */
-package org.mule.runtime.oauth.api.state;
```
elrodro83: why remove this?

extensions/oauth/src/test/resources/operations/operations-config.xml
elrodro83: If you do this,you do not need to make the array change.

fernandezlautaro: not sure what `extensions client` means, but I replicated what I saw here https://github.com/mulesoft/mule/blob/mule-4.x/extensions/oauth/src/test/java/org/mule/test/oauth2/internal/tokenmanager/InvalidateOauthContextTestCase.java#L39 (seems more like a functional test, as you are using the complete model: xml -> parsers -> extension -> operation).

modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/state/package.java
fernandezlautaro: isn't this some error when moving classes around?

extensions/oauth/src/test/resources/operations/operations-config.xml
```diff
@@ -0,0 +1,29 @@
...
+        <oauth:access-token tokenManager="tokenManagerConfig" resourceOwnerId="#[variables.resourceOwnerId]"/>
+    </flow>
+
+    <flow name="refreshTokenFlow">
+        <oauth:refresh-token tokenManager="tokenManagerConfig" resourceOwnerId="#[variables.resourceOwnerId]"/>
```
pablolagreca: We need to find a better name. Something like "retrieve-refresh-token", if not it would seem like this operation is doing a "refresh token". Once we decide the name for this operation, it should be apply to everything else (retrieve-access-token, retrieve-expires-in, etc)

fernandezlautaro: `retrieve-` is perfect. I'll apply the changes

extensions/oauth/src/test/resources/operations/operations-config.xml
fernandezlautaro: Tried, but there's an error in the extensions client call code (or I'm calling it with the wrong set of parameters)
Raised https://www.mulesoft.org/jira/browse/MULE-12009 for it (and talked with @pablomusumeci about it)

PR #4318, opened 2017-03-18
MULE-8990: State param encoding in OAuth support not compatible with …

facundovs:
…certain services like Concur

Concur (and others new connectors) complains about getting characters like '<'  '>' in the URL parameters, claiming it is a dangerous request.
Currently, We used it in the constant variable EVENT_STATE_TEMPLATE.
Devkit needs a mechanism to replace this character when they generate code.
With this fix, Devkit will able to implement the getSuffix method to override the respective character.
If the implemented method isn't overridden, the default ">>" prefix will be used.
"<<" Prefix will be removed because It's not used.

### DIFF
--- a/modules/devkit-support/src/main/java/org/mule/security/oauth/OAuthProperties.java
+++ b/modules/devkit-support/src/main/java/org/mule/security/oauth/OAuthProperties.java
@@ -12,12 +12,17 @@ package org.mule.security.oauth;
 public interface OAuthProperties
 {
 
-    public static final String VERIFIER = "_oauthVerifier";
-    public static final String ACCESS_TOKEN_ID = "OAuthAccessTokenId";
-    public static final String EVENT_STATE_TEMPLATE = "<<MULE_EVENT_ID=%s>>";
-    public static final String HTTP_STATUS = "http.status";
-    public static final String CALLBACK_LOCATION = "Location";
-    public static final String AUTHORIZATION_EVENT_KEY_TEMPLATE = "%s-authorization-event";
+    String VERIFIER = "_oauthVerifier";
+    String ACCESS_TOKEN_ID = "OAuthAccessTokenId";
+    String BASE_EVENT_STATE_TEMPLATE = "MULE_EVENT_ID=%s";
+    String DEFAULT_EVENT_STATE_TEMPLATE_SUFFIX = ">>";
+    String DEFAULT_EVENT_STATE_TEMPLATE = BASE_EVENT_STATE_TEMPLATE + DEFAULT_EVENT_STATE_TEMPLATE_SUFFIX;
+    String EVENT_ID_REGEX = "MULE_EVENT_ID=([\\w-]*)";
+    String ORIGINAL_STATE_REGEX = "MULE_EVENT_ID=[\\w-]*";
+    String EVENT_STATE_REGEX_SUFFIX = "(.*)";
+    String HTTP_STATUS = "http.status";
+    String CALLBACK_LOCATION = "Location";
+    String AUTHORIZATION_EVENT_KEY_TEMPLATE = "%s-authorization-event";
 }
 
 
--- a/modules/devkit-support/src/main/java/org/mule/security/oauth/processor/BaseOAuth2AuthorizeMessageProcessor.java
+++ b/modules/devkit-support/src/main/java/org/mule/security/oauth/processor/BaseOAuth2AuthorizeMessageProcessor.java
@@ -7,6 +7,10 @@
 
 package org.mule.security.oauth.processor;
 
+import static java.lang.System.getProperty;
+import static org.mule.api.config.MuleProperties.SYSTEM_PROPERTY_PREFIX;
+import static org.mule.security.oauth.OAuthProperties.BASE_EVENT_STATE_TEMPLATE;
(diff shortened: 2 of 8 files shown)


modules/devkit-support/src/test/java/org/mule/security/oauth/FetchAccessTokenTestCase.java
```diff
@@ -46,7 +46,7 @@ protected void doSetUp() throws Exception
         this.objectStore = new InMemoryObjectStore<Serializable>();
 
         this.event = getTestEvent("");
-        event.getMessage().setProperty("state", "<<MULE_EVENT_ID=whatever>>", PropertyScope.INBOUND);
+        event.getMessage().setProperty("state", "MULE_EVENT_ID=whatever>>", PropertyScope.INBOUND);
```
marianogonzalez: This makes no sense. What's the point of putting that suffix if there's no prefix? It doesn't serve it's purpose. Either replace both suffix and prefix for non conflictive characters or remove both.

modules/devkit-support/src/main/java/org/mule/security/oauth/OAuthProperties.java
```diff
@@ -12,12 +12,17 @@
...
-    public static final String VERIFIER = "_oauthVerifier";
-    public static final String ACCESS_TOKEN_ID = "OAuthAccessTokenId";
-    public static final String EVENT_STATE_TEMPLATE = "<<MULE_EVENT_ID=%s>>";
-    public static final String HTTP_STATUS = "http.status";
-    public static final String CALLBACK_LOCATION = "Location";
-    public static final String AUTHORIZATION_EVENT_KEY_TEMPLATE = "%s-authorization-event";
+    String VERIFIER = "_oauthVerifier";
+    String ACCESS_TOKEN_ID = "OAuthAccessTokenId";
+    String BASE_EVENT_STATE_TEMPLATE = "MULE_EVENT_ID=%s";
+    String DEFAULT_EVENT_STATE_TEMPLATE_SUFFIX = ">>";
```
marianogonzalez: should not exist

modules/devkit-support/src/main/java/org/mule/security/oauth/processor/BaseOAuth2AuthorizeMessageProcessor.java
```diff
@@ -177,4 +181,13 @@ private void setState(Map<String, String> extraParameters, MuleEvent event)
         }
     }
 
+    /**
+     * Override this method when it's necessary override the separator of Ids.
+     * @return the value of CUSTOM_SUFFIX_PROPERTY if it was defined, DEFAULT_EVENT_STATE_SUFFIX otherwise.
+     */
+    protected String getSuffix()
```
marianogonzalez: delete this

modules/devkit-support/src/main/java/org/mule/security/oauth/OAuthProperties.java
```diff
@@ -12,12 +12,17 @@
...
+    String DEFAULT_EVENT_STATE_TEMPLATE = BASE_EVENT_STATE_TEMPLATE + DEFAULT_EVENT_STATE_TEMPLATE_SUFFIX;
+    String EVENT_ID_REGEX = "MULE_EVENT_ID=([\\w-]*)";
+    String ORIGINAL_STATE_REGEX = "MULE_EVENT_ID=[\\w-]*";
+    String EVENT_STATE_REGEX_SUFFIX = "(.*)";
+    String HTTP_STATUS = "http.status";
```
elrodro83: there is already a constant for this in the http module

modules/devkit-support/src/main/java/org/mule/security/oauth/OAuthProperties.java
```diff
@@ -12,12 +12,17 @@
...
-    public static final String EVENT_STATE_TEMPLATE = "<<MULE_EVENT_ID=%s>>";
-    public static final String HTTP_STATUS = "http.status";
-    public static final String CALLBACK_LOCATION = "Location";
-    public static final String AUTHORIZATION_EVENT_KEY_TEMPLATE = "%s-authorization-event";
+    String VERIFIER = "_oauthVerifier";
+    String ACCESS_TOKEN_ID = "OAuthAccessTokenId";
+    String BASE_EVENT_STATE_TEMPLATE = "MULE_EVENT_ID=%s";
+    String DEFAULT_EVENT_STATE_TEMPLATE_SUFFIX = ">>";
+    String DEFAULT_EVENT_STATE_TEMPLATE = BASE_EVENT_STATE_TEMPLATE + DEFAULT_EVENT_STATE_TEMPLATE_SUFFIX;
+    String EVENT_ID_REGEX = "MULE_EVENT_ID=([\\w-]*)";
```
elrodro83: extract `MULE_EVENT_ID` to its own constant

modules/devkit-support/src/main/java/org/mule/security/oauth/OAuthProperties.java
marianogonzalez: this is legacy devkit support and should not depend on http

estebanwasinger:
> With this fix, Devkit will able to implement the getSuffix method to override the respective character
I don't know if DevKit will use `getSuffix()` because using this method will make newer connectors incompatible with older versions of Mule.

PR #4324, opened 2017-03-21
MULE-8990: State param encoding in OAuth support not compatible with …

facundovs:
…certain services like Concur.

The best solution for this issue is to remove the EVENT_STATE_TEMPLATE template.
It will be possible to separate the eventId and some other extraId, because an EventId has a known length: a UUID of 36 characters.
It's important to know that this UUID is concatenated with the ClusterNodeId. When It's the default It is a zero. However, As It is customizable by System Property, we must check if this property is defined to know the variable length of the EventId.


core/src/test/java/org/mule/util/UUIDTestCase.java
```diff
@@ -57,6 +59,16 @@ public int compare(Object[] o1, Object[] o2)
         }
     }
 
+    @Test
+    public void testUUUIDLength() throws Exception
+    {
+        for (int i = 0; i < 1000; i++)
```
marianogonzalez: doing it once is enough

core/src/main/java/org/mule/DefaultMuleEvent.java
```diff
@@ -69,6 +70,8 @@
 
     private static Log logger = LogFactory.getLog(DefaultMuleEvent.class);
 
+    public static final int DEFAULT_LENGTH_MULE_EVENT_ID = UUID_LENGTH + 2 ;
```
marianogonzalez: why + 2?

modules/devkit-support/src/main/java/org/mule/security/oauth/processor/OAuth2FetchAccessTokenMessageProcessor.java
```diff
@@ -212,6 +214,32 @@ public void set(String key, String value)
...
+        if (clusterId != null)
+        {
+            return state.length() >= UUID_LENGTH + clusterId.length() ? state.substring(0, UUID_LENGTH + clusterId.length() + 1) : null;
+        }
+
+        return state.length() >= DEFAULT_LENGTH_MULE_EVENT_ID ? state.substring(0, DEFAULT_LENGTH_MULE_EVENT_ID) : null;
+
+    }
+
+    private String getExtraId(String state)
```
marianogonzalez: getExtraId -> getOriginalState()

core/src/main/java/org/mule/DefaultMuleEvent.java
facundovs: because In DefaultMuleEvent.generateEventId() -> DefaultMuleContext.getUniqueIdString(), UUID is concatenated with default clusterNodeId (zero) and "-" character.

marianogonzalez: add a comment explaining that

marcosnc: Could clusterNodeId have 2 digits?

facundovs: By default It's a zero. However, It can have a customizable value. 
I'm checking that [here](https://github.com/mulesoft/mule/pull/4324/files#diff-d43440932499749e49e2c518b38914efR219)

core/src/main/java/org/mule/DefaultMuleEvent.java
```diff
@@ -69,6 +70,9 @@
 
     private static Log logger = LogFactory.getLog(DefaultMuleEvent.class);
 
+    /** EventId is concatenated with a dash character and a default clusterNodeId (a zero) .**/
+    public static final int DEFAULT_LENGTH_MULE_EVENT_ID = UUID_LENGTH + 2 ;
```
marianogonzalez: you need two digits for the cluster id

PR #4328, opened 2017-03-21
MULE-12020: Fix HTTP requester grizzly configuration

afelisatti:


modules/http/src/test/java/org/mule/module/http/functional/requester/HttpRequestTlsConnectionCloseTestCase.java
```diff
@@ -38,12 +41,20 @@ protected void writeResponse(HttpServletResponse response) throws IOException
     {
         super.writeResponse(response);
         response.addHeader(CONNECTION, CLOSE);
+        // Avoid closing the connection until the response is received
+        response.flushBuffer();
+        try {
+            latch.await(1, SECONDS);
```
elrodro83: use the receive timeout constant we already have in a superclass

PR #4333, opened 2017-03-22
MULE-12023: In HttpMultipartMuleMessageFactory, multiple threads use …

facundovs:
…instance variable without syncronization

In HttpMultipartMuleMessageFactory, multiple threads access to parts instance variable without syncronization. It's causing multiple exceptions are
triggered, such as NPE and ConcurrentModificationException.
extractPayloadFromHttpRequest, convertMultiPartHeaders and addAttachments methods should be executed atomically.

### DIFF
--- a/core/src/main/java/org/mule/transport/AbstractMuleMessageFactory.java
+++ b/core/src/main/java/org/mule/transport/AbstractMuleMessageFactory.java
@@ -63,21 +63,35 @@ public abstract class AbstractMuleMessageFactory implements MuleMessageFactory
         return doCreate(transportMessage, null, encoding, muleContext);
     }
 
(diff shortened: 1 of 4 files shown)


transports/http/src/main/java/org/mule/transport/http/HttpMuleMessageFactory.java
```diff
@@ -71,6 +72,27 @@ public HttpMuleMessageFactory(MuleContext context)
...
+        if (transportMessage == null)
+        {
+            return new DefaultMuleMessage(NullPayload.getInstance(), muleContext);
+        }
+
+        validateMessage(transportMessage);
+
+        synchronized (this)
+        {
+            Object payload = extractPayload(transportMessage, encoding);
```
alepulver: How about just calling super here to avoid duplicating the same code again?

PR #4336, opened 2017-03-22
MULE-12054: Add after method to clear static notification list

abrden:


afelisatti:
Please create a specific ticket for this and reference it here.

PR #4339, opened 2017-03-22
MULE-11558: Remove XML module

elrodro83:


elrodro83:
For the xml module, this is just building an empty jar so it is uploaded to nexus and make a fail-fast on any remaining users. After a build runs with this change, it will be removed

modules/pom.xml
```diff
@@ -44,6 +43,7 @@
         <module>tls</module>
         <module>tomcat</module>
         <module>tooling-support</module>
+        <!-- TODO Remove this module -->
         <module>xml</module>
```
pablokraan: Is this module left here in order to update the repository with an empty jar?

services/soap/pom.xml
```diff
@@ -100,7 +111,6 @@
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
-            <version>${guavaVersion}</version>
```
pablokraan: There are many dependencies in poms where the version attribute was removed. I guess, because the dependency is managed in the parent pom. This particular change is in a service. Services should have their own dependencies as they will be included inside the service artifact. (Probably a change that goes beyond this PR, but something we need to take in account)

services/soap/src/main/java/org/mule/services/soap/impl/xml/util/XMLUtils.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package org.mule.services.soap.impl.xml.util;
+
+import org.dom4j.io.DOMReader;
+import org.w3c.dom.Document;
+
+/**
+ * General utility methods for working with XML.
+ */
+public class XMLUtils extends org.mule.runtime.core.util.XMLUtils {
```
pablokraan: Could XmlUtils be in a separated artifact and just re-use that jar when needed instead of copying classes? Also the XMLUtils from core should be probably not exposed on the container

modules/pom.xml
elrodro83: exactly

services/soap/src/main/java/org/mule/services/soap/impl/xml/util/XMLUtils.java
elrodro83: in this case, this is just the code needed by the soap service and the wsc. I'll create a ticket so this can be tracked separately from those artifacts (which are related.)

elrodro83: This is already being tackled by @pablomusumeci

PR #4341, opened 2017-03-23
MULE-12033: Set the stage for introducing object streaming

marianogonzalez:


modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -5221,6 +5221,79 @@
...
     </xsd:complexType>
 
+    <!--===================-->
+    <!-- Objects Streaming -->
+    <!--===================-->
+
+    <xsd:element name="abstract-object-streaming-strategy" abstract="true">
+        <xsd:annotation>
+            <xsd:documentation>
+                A placeholder for a object streaming strategy element. This element allows to configure repeatable streams behaviour
```
elrodro83: `a object` -> `an object`

PR #4349, opened 2017-03-23
MULE-12041: Globalfunctions are overriden each time they are defined.

fsgonz:


tests/integration/src/test/resources/org/mule/test/el/expression-language-config.xml
```diff
@@ -6,6 +6,10 @@
        http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd
        http://www.mulesoft.org/schema/mule/test http://www.mulesoft.org/schema/mule/test/current/mule-test.xsd">
 
+	<spring:beans>
```
elrodro83: can you just include another `expression-language` in the `configuration` in this file?

fsgonz: Configuration only lets you define one expression language as defined in mule.xsd.

PR #4361, opened 2017-03-27
MULE-11942: Missing information about metadata key when resolving ope…

pablomusumeci:
…ration metadata

### DIFF
--- a/modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/metadata/MetadataExtensionFunctionalTestCase.java
+++ b/modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/metadata/MetadataExtensionFunctionalTestCase.java
@@ -32,7 +32,6 @@ import org.mule.runtime.api.component.location.Location;
 import org.mule.runtime.api.meta.Typed;
 import org.mule.runtime.api.meta.model.ComponentModel;
 import org.mule.runtime.api.meta.model.OutputModel;
-import org.mule.runtime.api.meta.model.operation.OperationModel;
 import org.mule.runtime.api.metadata.MetadataKey;
 import org.mule.runtime.api.metadata.MetadataKeysContainer;
 import org.mule.runtime.api.metadata.MetadataService;
@@ -53,13 +52,14 @@ import java.lang.reflect.Type;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Set;
+import java.util.function.BiConsumer;
 
 import org.junit.Before;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 
 @RunWith(Parameterized.class)
-public abstract class MetadataExtensionFunctionalTestCase extends ExtensionFunctionalTestCase {
+public abstract class MetadataExtensionFunctionalTestCase<T extends ComponentModel<T>> extends ExtensionFunctionalTestCase {
 
   protected static final String METADATA_TEST = "metadata-tests.xml";
   protected static final String DSQL_QUERY = "dsql:SELECT id FROM Circle WHERE (diameter < 18)";
@@ -116,16 +116,11 @@ public abstract class MetadataExtensionFunctionalTestCase extends ExtensionFunct
   protected final static MetadataKey PERSON_METADATA_KEY = newKey(PERSON).build();
   protected static final MetadataKey CAR_KEY = newKey(CAR).build();
   protected static final MetadataKey LOCATION_MULTILEVEL_KEY =
-      PartAwareMetadataKeyBuilder.newKey(AMERICA, CONTINENT).withChild(PartAwareMetadataKeyBuilder.newKey(USA, COUNTRY)
-          .withChild(PartAwareMetadataKeyBuilder.newKey(SAN_FRANCISCO, CITY))).build();
+      MultilevelMetadataKeyBuilder.newKey(AMERICA, CONTINENT).withChild(MultilevelMetadataKeyBuilder.newKey(USA, COUNTRY)
+          .withChild(MultilevelMetadataKeyBuilder.newKey(SAN_FRANCISCO, CITY))).build();
 
   protected final static NullMetadataKey NULL_METADATA_KEY = new NullMetadataKey();
   protected final static ClassTypeLoader TYPE_LOADER = ExtensionsTestUtils.TYPE_LOADER;
-
-  private static final MetadataComponentDescriptorProvider<OperationModel> explicitMetadataResolver =
-      MetadataService::getOperationMetadata;
-  private static final MetadataComponentDescriptorProvider<OperationModel> dslMetadataResolver =
-      (metadataService, componentId, key) -> metadataService.getOperationMetadata(componentId);
 
   protected MetadataType personType;
   protected Location location;
@@ -133,10 +128,13 @@ public abstract class MetadataExtensionFunctionalTestCase extends ExtensionFunct
   protected MetadataService metadataService;
   protected ClassTypeLoader typeLoader = ExtensionsTypeLoaderFactory.getDefault().createTypeLoader();
   protected BaseTypeBuilder typeBuilder = BaseTypeBuilder.create(JAVA);
-  protected MetadataComponentDescriptorProvider provider;
+  protected MetadataComponentDescriptorProvider<T> provider;
 
-  @Parameterized.Parameter
-  public ResolutionType resolutionType;
+  protected ResolutionType resolutionType;
+
+  MetadataExtensionFunctionalTestCase(ResolutionType resolutionType) {
+    this.resolutionType = resolutionType;
+  }
 
   @Override
   protected Class<?>[] getAnnotatedExtensionClasses() {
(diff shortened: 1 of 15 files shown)


modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/metadata/MetadataExtensionFunctionalTestCase.java
```diff
@@ -230,6 +249,27 @@ protected void assertExpectedType(Typed typedModel, MetadataType expectedType, b
...
 
+  protected <T extends ComponentModel> void assertResolvedKey(MetadataResult<ComponentMetadataDescriptor<T>> result,
+                                                              MetadataKey metadataKey) {
+    assertThat(result.get().getMetadataAttributes().getKey().isPresent(), is(true));
+    MetadataKey resultKey = result.get().getMetadataAttributes().getKey().get();
+    assertSameKey(metadataKey, resultKey);
+
+    MetadataKey child = metadataKey.getChilds().stream().findFirst().orElseGet(() -> null);
+    MetadataKey otherChild = resultKey.getChilds().stream().findFirst().orElseGet(() -> null);
+    while (child != null && otherChild != null) {
```
alegmarra: All this should not be necessary in order to compare if two keys are the same. 
If current `equals` doesn't work like this, we should update it.

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/metadata/MetadataExtensionFunctionalTestCase.java
```diff
@@ -110,6 +114,11 @@
   protected static final String CITY = "city";
 
   protected final static MetadataKey PERSON_METADATA_KEY = newKey(PERSON).build();
+  protected static final MetadataKey CAR_KEY = newKey(CAR).build();
+  protected static final MetadataKey LOCATION_MULTILEVEL_KEY =
+      PartAwareMetadataKeyBuilder.newKey(AMERICA, CONTINENT).withChild(PartAwareMetadataKeyBuilder.newKey(USA, COUNTRY)
```
marianogonzalez: PartAwareMetadataKeyBuilder -> MultilevelMetadataKeyBuilder

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/metadata/SourceMetadataTestCase.java
```diff
@@ -33,16 +34,27 @@
...
+  private static final MetadataComponentDescriptorProvider<SourceModel> dslMetadataResolver =
+      (metadataService, componentId, key) -> metadataService.getSourceMetadata(componentId);
 
   @Override
   protected String getConfigFile() {
     return METADATA_TEST;
   }
 
+  @Override
+  protected void setProvider() {
```
marianogonzalez: already saw this piece of code above

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataKeyIdObjectResolver.java
```diff
@@ -127,6 +132,58 @@ public Object resolve() throws MetadataResolvingException {
...
+            }
+          } catch (Exception e) {
+            exceptionReference
+                .set(new MetadataResolvingException("Could not construct Metadata Key part for parameter " + p.getName(),
+                                                    FailureCode.INVALID_METADATA_KEY, e));
+          }
+        });
+
+    if (exceptionReference.get() != null) {
+      throw exceptionReference.get();
```
alegmarra: don't use streams if you have to do this

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataKeyIdObjectResolver.java
```diff
@@ -127,6 +132,58 @@ public Object resolve() throws MetadataResolvingException {
...
+  MetadataKey reconstructKeyFromType(Object resolvedKey) throws MetadataResolvingException {
+    if (isKeyLess()) {
+      return new NullMetadataKey();
+    }
+
+    if (keyParts.size() == 1) {
+      return newKey(valueOf(resolvedKey)).build();
+    }
+
+    Reference<MetadataKeyBuilder> rootBuilder = new Reference<>();
```
marianogonzalez: 3 reference objects just to be able to use a stream is not worth it. Do old school iteration instead and have a cleaner code

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataKeyIdObjectResolver.java
```diff
@@ -262,6 +319,11 @@ public Object getResultId() throws MetadataResolvingException {
   private List<ParameterModel> getMetadataKeyParts(ComponentModel componentModel) {
     return componentModel.getAllParameterModels().stream()
         .filter(p -> p.getModelProperty(MetadataKeyPartModelProperty.class).isPresent())
+        .sorted((p1, p2) -> {
+          Optional<MetadataKeyPartModelProperty> mk1 = p1.getModelProperty(MetadataKeyPartModelProperty.class);
+          Optional<MetadataKeyPartModelProperty> mk2 = p2.getModelProperty(MetadataKeyPartModelProperty.class);
+          return mk1.get().getOrder() - mk2.get().getOrder();
```
marianogonzalez: what if either one is empty?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataKeyIdObjectResolver.java
```diff
@@ -262,6 +319,11 @@ public Object getResultId() throws MetadataResolvingException {
   private List<ParameterModel> getMetadataKeyParts(ComponentModel componentModel) {
     return componentModel.getAllParameterModels().stream()
         .filter(p -> p.getModelProperty(MetadataKeyPartModelProperty.class).isPresent())
+        .sorted((p1, p2) -> {
+          Optional<MetadataKeyPartModelProperty> mk1 = p1.getModelProperty(MetadataKeyPartModelProperty.class);
```
alegmarra: .filter()
.map(p-> p.getModelProperty.get)
.sorted(k1, k2 -> k1.getOder - k2.getOrder)

marianogonzalez: this would return a sorted list of model properties. What he wants is the sorted list of parameters.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataMediator.java
```diff
@@ -121,41 +120,55 @@ public MetadataMediator(T componentModel) {
...
+      Object keyValue;
+      MetadataResult keyValueResult = getMetadataKeyObjectValue(metadataKeyResolver);
+      if (!keyValueResult.isSuccess()) {
+        return keyValueResult;
+      } else {
+        keyValue = keyValueResult.get();
+      }
+
+      MetadataAttributes.MetadataAttributesBuilder builder = MetadataAttributes.builder();
+      if (keyValue != null) {
```
alegmarra: why is this a valid scenario? can we really continue with a null key? 
the current TODO in that line should be fixed by this PR and this if should not be necessary any more.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataMediator.java
```diff
@@ -121,41 +120,55 @@ public MetadataMediator(T componentModel) {
...
+        builder.withKey(keyIdObjectResolver.reconstructKeyFromType(keyValue));
+      }
+
+      return getMetadata(context, keyValue, builder);
```
alegmarra: is the `null` key being propagated all the way to the Typeresolver the developer implemented?

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/metadata/MetadataExtensionFunctionalTestCase.java
```diff
@@ -139,8 +146,8 @@
   @Parameterized.Parameters(name = "{0}")
   public static Collection<Object[]> data() {
     return Arrays.asList(new Object[][] {
-        {EXPLICIT_RESOLUTION, explicitMetadataResolver},
```
alegmarra: revert, this is better than having the `setProvider` method in the setup

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/metadata/MetadataExtensionFunctionalTestCase.java
```diff
@@ -168,6 +184,9 @@ ComponentMetadataDescriptor getSuccessComponentDynamicMetadata() {
     MetadataResult<ComponentMetadataDescriptor<T>> componentMetadata = getComponentDynamicMetadata(key);
     String msg = componentMetadata.getFailures().stream().map(f -> "Failure: " + f.getMessage()).collect(joining(", "));
     assertThat(msg, componentMetadata.isSuccess(), is(true));
+    if (!key.getId().equals(NULL_METADATA_KEY.getId())) {
```
alegmarra: don't use the NULL_KEY to handle the assertion, if we have cases where the key has to be validated then handle it in each test, or create a `withKey` helper method that always asserts this

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataKeyIdObjectResolver.java
alegmarra: 🤔 true that

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataKeyIdObjectResolver.java
alegmarra: already filtered them out

PR #4364, opened 2017-03-27
MULE-12063: Add create=true to create muleEmbeddedDB if it doesn't exist

abrden:


modules/db/src/test/java/org/mule/module/db/integration/config/GenericDbConfigTestCase.java
```diff
@@ -27,7 +27,7 @@
 {
 
     private static final String EXPECTED_DRIVER_CLASSNAME = "org.apache.derby.jdbc.EmbeddedDriver";
-    private static final String EXPECTED_URL = "jdbc:derby:muleEmbeddedDB;sql.enforce_strict_size=true";
```
estebanwasinger: Also, I would change the database to `target/muleEmbeddedDB` so derby files are automatically cleaned when executing a `clean` with maven.

estebanwasinger: https://github.com/mulesoft/mule/blob/mule-4.x/extensions/db/src/test/resources/integration/config/derby-datasource.xml#L9

abrden: Great idea, I moved it for all the xmls in the db module

PR #4366, opened 2017-03-28
MULE-11978 CursorStreamProvider is closed too early

dfeist:


core/src/main/java/org/mule/runtime/core/DefaultEventContext.java
```diff
@@ -135,7 +145,9 @@ public ProcessorsTrace getProcessorsTrace() {
    * @param correlationId the correlation id that was set by the {@link MessageSource} for the first {@link Event} of this
    *        context, if available.
    */
-  private DefaultEventContext(FlowConstruct flow, String connectorName, String correlationId) {
+  private DefaultEventContext(FlowConstruct flow, String connectorName, String correlationId,
+                              Publisher<Void> completionCallback) {
```
marianogonzalez: add param jdoc

PR #4370, opened 2017-03-28
MULE-11221: Support Transactions in Sources

estebanwasinger:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/ExtensionsConnectionAdapter.java
```diff
@@ -56,13 +54,36 @@
...
         ? getTransactedConnectionHandler(executionContext, executionContext.getTransactionConfig().get())
         : getTransactionlessConnectionHandler(executionContext);
   }
 
+  private <T extends TransactionalConnection> ConnectionHandler<T> getTransactedConnectionHandler(
+                                                                                                  ExecutionContextAdapter<? extends ComponentModel> executionContext,
+                                                                                                  TransactionConfig transactionConfig)
+      throws ConnectionException, TransactionException {
+
+    if (transactionConfig.getAction() == ACTION_NOT_SUPPORTED || transactionConfig.getAction() == ACTION_NONE) {
```
alegmarra: `ACTION_NOT_SUPPORTED` it's a lie, and not even a white lie. Not_supported should be used only for non-transactional components

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/ExtensionsConnectionAdapter.java
```diff
@@ -56,13 +54,36 @@
...
+      return getTransactionlessConnectionHandler(executionContext);
+    }
+
+    ConfigurationInstance configuration = executionContext.getConfiguration()
+        .orElseThrow(() -> new IllegalStateException(format(
+                                                            "Operation '%s' of extension '%s' cannot participate in a transaction because it doesn't have a config",
```
alegmarra: `Operation` -> `Component`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -252,7 +269,7 @@ public Scheduler getFlowExecutionExecutor() {
 
       @Override
       public TransactionConfig getTransactionConfig() {
-        return null;
+        return sourceModel.isTransactional() ? buildTransactionConfig() : null;
```
alegmarra: Optional<TransactionConfig> getTransactionConfig()

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/ExtensionsConnectionAdapter.java
```diff
@@ -46,6 +42,8 @@
   @Inject
   private ConnectionManager connectionManager;
 
+  private final TransactionBindDelegate transactionBindDelegate = new TransactionBindDelegate();
```
marianogonzalez: TransactionBindingDelegate

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/DefaultSourceCallback.java
```diff
@@ -34,6 +45,9 @@
  */
 class DefaultSourceCallback<T, A extends Attributes> implements SourceCallback<T, A> {
 
+  private static final String UNABLE_TO_START_TX_ERROR_MSG_TEMPLATE = "Unable to start a transaction from the Source '%s' of the extension '%s' with out a %s";
```
marianogonzalez: with out -> without?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -345,4 +373,13 @@ protected void doInitialise() throws InitialisationException {
       throw new InitialisationException(e, this);
     }
   }
+
+  private byte getSourceTransactionalAction() {
+    try {
+      return toActionCode(sourceAdapter.getTransactionalAction());
+    } catch (MuleException e) {
+      LOGGER.error("An error occurred obtaining the Source Transactional Action, defaulting to [NONE]");
```
marianogonzalez: why would this fail? why default to NONE (and basically disobey the user) instead of bubbling the exception?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -345,4 +373,13 @@ protected void doInitialise() throws InitialisationException {
...
+      return ACTION_NONE;
```
alegmarra: why are we doing this instead of failing? the cases where an exception is thrown inside the `getTransactionalAction` occur because the value assigned was invalid, and we (should) always fail when that happens. Here we are saying "You used an expression, that's not supported, so we'll just ignore the transactional part"

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceAdapter.java
```diff
@@ -82,13 +87,14 @@ public SourceAdapter(ExtensionModel extensionModel, SourceModel sourceModel,
                        Source source,
                        Optional<ConfigurationInstance> configurationInstance,
                        SourceCallbackFactory sourceCallbackFactory,
-                       ResolverSet successCallbackParameters,
+                       ResolverSet nonCallbackParameters, ResolverSet successCallbackParameters,
```
marianogonzalez: one argument per line seems to be the rule of the land

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceAdapter.java
```diff
@@ -261,6 +275,21 @@ public Source getDelegate() {
     return source;
   }
 
+  public SourceTransactionalAction getTransactionalAction() throws MuleException {
+    ValueResolver valueResolver = nonCallbackParameters.getResolvers().get(TRANSACTIONAL_ACTION_PARAMETER_NAME);
+
+    if (valueResolver.isDynamic()) {
```
marianogonzalez: this seems overly defensive or you're hiding a bug elsewhere. You should not worry about this at this point since the parameter model doesn't accept expressions. If you ever got a dynamic resolver here, then that's a bug you need to investigate

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/transaction/TransactionBindDelegate.java
```diff
@@ -0,0 +1,94 @@
...
+   * @param extensionModel model of the current extension
+   * @param transactionConfig given transaction config
+   * @param txKey the transaction key
+   * @param currentTx current transaction
+   * @param connectionHandlerSupplier {@link Supplier} to get the {@link ConnectionHandler} of the current component
+   * @return The {@link ConnectionHandler} that has be bound to the transaction.
+   * @throws ConnectionException if a problem occurred retrieving the {@link ConnectionHandler}
+   * @throws TransactionException if the connection could not be bound to the current transaction
+   */
+  public <T extends TransactionalConnection> ConnectionHandler<T> bindResource(ComponentModel componentModel,
```
marianogonzalez: too many arguments are bad for your health. Apply the command pattern

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/transaction/TransactionBindDelegate.java
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  private ExtensionTransactionalResource createTransactionalResource(Transaction currentTx, ConnectionHandler connectionHandler,
+                                                                     Object connection) {
+    return connection instanceof XATransactionalConnection
+        ? new XAExtensionTransactionalResource((XATransactionalConnection) connection, connectionHandler, currentTx)
+        : new ExtensionTransactionalResource((TransactionalConnection) connection, connectionHandler, currentTx);
+  }
+
+  public interface ConnectionSupplier<T> {
```
marianogonzalez: `@FunctionalInterface`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/transaction/TransactionBindDelegate.java
```diff
@@ -0,0 +1,94 @@
...
+   * @throws TransactionException if the connection could not be bound to the current transaction
+   */
+  public <T extends TransactionalConnection> ConnectionHandler<T> bindResource(ComponentModel componentModel,
+                                                                               ExtensionModel extensionModel,
+                                                                               TransactionConfig transactionConfig,
+                                                                               ExtensionTransactionKey txKey,
+                                                                               Transaction currentTx,
+                                                                               ConnectionSupplier<ConnectionHandler<T>> connectionHandlerSupplier)
+      throws ConnectionException, TransactionException {
+    if (currentTx != null) {
```
alegmarra: is this really a valid scenario? if this happens we are not binding anything just using the supplier as a fallback. Looks like if we don't have a transaction we should never try to bind a resource in the first place

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/transaction/TransactionSourceBinder.java
```diff
@@ -0,0 +1,60 @@
...
+                                                            componentModel.getName(),
+                                                            extensionModel.getName())));
+
+    final ExtensionTransactionKey txKey = new ExtensionTransactionKey(configuration);
+    final Transaction currentTx = TransactionCoordination.getInstance().getTransaction();
+    transactionBindDelegate.bindResource(componentModel, extensionModel, transactionConfig, txKey, currentTx,
+                                         () -> connectionHandler);
+  }
+
+  private boolean shouldBeBound(TransactionConfig transactionConfig) {
```
marianogonzalez: shouldBeBound -> isNotTransactional

tests/test-extensions/transactional-extension/src/main/java/org/mule/test/transactional/TransactionalOperations.java
```diff
@@ -31,6 +35,12 @@ public void verifyTransactionRolledback(@Connection TestTransactionalConnection
     checkState(connection.isTransactionRolledback(), "transaction not rolled back");
   }
 
+  public void verifySameConnection(@Connection TestTransactionalConnection connection,
+                                   @Content TestTransactionalConnection transactionalMessage) {
+    checkState(Objects.equals(connection.getConnectionId(), transactionalMessage.getConnectionId()),
```
marianogonzalez: if you're asserting for the same, then use `==`. Don't speculate on the equals() implementation

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/transaction/TransactionBindDelegate.java
```diff
@@ -0,0 +1,94 @@
...
+      ExtensionTransactionalResource<T> txResource = createTransactionalResource(currentTx, connectionHandler, connection);
+      boolean bound = false;
+      try {
+        if (currentTx.supports(txKey, txResource)) {
+          currentTx.bindResource(txKey, txResource);
+          bound = true;
+          return new TransactionalConnectionHandler(txResource);
+        } else if (transactionConfig.isTransacted()) {
+          throw new TransactionException(createStaticMessage(format("%s '%s' of extension '%s' uses a transactional connection '%s', but the current transaction "
+                                             + "doesn't support it and could not being bound",
```
alegmarra: `being` -> `be` ?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/transaction/TransactionSourceBinder.java
```diff
@@ -0,0 +1,60 @@
...
+
+  public void bindToTransaction(ComponentModel componentModel,
+                                ExtensionModel extensionModel,
+                                TransactionConfig transactionConfig,
+                                ConfigurationInstance configurationInstance,
+                                ConnectionHandler connectionHandler)
+
+      throws ConnectionException, TransactionException {
+
+    if (shouldBeBound(transactionConfig)) {
```
alegmarra: `shouldNotBeBound` or `!shouldBeBound` (changing the impl of course =p)

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/transaction/TransactionalSourceTestCase.java
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  @Override
+  protected String getConfigFile() {
+    return "source-transaction-config.xml";
+  }
+
+  @After
+  public void tearDown() {
+    MessageStorage.messages = new ConcurrentLinkedQueue<>();
```
marianogonzalez: MessageStore should have a reset() method. You need to do this on the doSetup() as well

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/transaction/TransactionSourceBinder.java
```diff
@@ -0,0 +1,60 @@
...
+                                                            extensionModel.getName())));
+
+    final ExtensionTransactionKey txKey = new ExtensionTransactionKey(configuration);
+    final Transaction currentTx = TransactionCoordination.getInstance().getTransaction();
+    transactionBindDelegate.bindResource(componentModel, extensionModel, transactionConfig, txKey, currentTx,
+                                         () -> connectionHandler);
+  }
+
+  private boolean shouldBeBound(TransactionConfig transactionConfig) {
+    return transactionConfig.getAction() == ACTION_NOT_SUPPORTED || transactionConfig.getAction() == ACTION_NONE;
```
alegmarra: seen this twice already, isn't `TransactionConfig::isTransacted` equivalent for this? It adds 'ACTION_NEVER` to the resolution, but does it change anything for this two use cases?

alegmarra: if not, do we need a different concept of `isXXXXXX` in the TransactionalConfig?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/ExtensionsConnectionAdapter.java
marianogonzalez: it's not a lie. It's a very bad case of naming which we're keeping because of trying to playing nice with JTA. Not steve's fault and a decision was made to roll with this

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/transaction/TransactionBindDelegate.java
marianogonzalez: yes, this is valid

alegmarra: I don't think we should call a "bind resource" over a null Transaction. Seems like this should be handled in the outside

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/ExtensionsConnectionAdapter.java
pablomusumeci: https://github.com/mulesoft/mule-extensions-api/blob/master/mule-extensions-api/src/main/java/org/mule/runtime/extension/api/util/NameUtils.java#L301

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/transaction/TransactionSourceBinder.java
estebanwasinger: 👏 I didn't saw that

core/src/main/java/org/mule/runtime/core/execution/AsyncResponseFlowProcessingPhase.java
```diff
@@ -49,8 +49,8 @@ public void runPhase(final AsyncResponseFlowProcessingPhaseTemplate template, fi
           TransactionalErrorHandlingExecutionTemplate transactionTemplate =
               createMainExecutionTemplate(messageProcessContext.getFlowConstruct().getMuleContext(),
                                           messageProcessContext.getFlowConstruct(),
-                                          (messageProcessContext.getTransactionConfig() == null ? new MuleTransactionConfig()
-                                              : messageProcessContext.getTransactionConfig()),
+                                          (messageProcessContext.getTransactionConfig().isPresent()
```
marianogonzalez: orElse();

core/src/main/java/org/mule/runtime/core/execution/FlowProcessingPhase.java
```diff
@@ -48,8 +46,8 @@ public void runPhase(final FlowProcessingPhaseTemplate flowProcessingPhaseTempla
           TransactionalErrorHandlingExecutionTemplate transactionTemplate =
               createMainExecutionTemplate(messageProcessContext.getFlowConstruct().getMuleContext(),
                                           messageProcessContext.getFlowConstruct(),
-                                          (messageProcessContext.getTransactionConfig() == null ? new MuleTransactionConfig()
-                                              : messageProcessContext.getTransactionConfig()),
+                                          (messageProcessContext.getTransactionConfig().isPresent()
```
marianogonzalez: orElse

core/src/main/java/org/mule/runtime/core/execution/FlowProcessingPhase.java
```diff
@@ -48,8 +46,8 @@ public void runPhase(final FlowProcessingPhaseTemplate flowProcessingPhaseTempla
...
+                                              ? messageProcessContext.getTransactionConfig().get() : new MuleTransactionConfig()),
```
marianogonzalez: orElse()!!

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
```diff
@@ -265,8 +265,8 @@ private Processor createFlowExecutionProcessor(MessageSource messageSource, Mess
         TransactionalErrorHandlingExecutionTemplate transactionTemplate =
             createMainExecutionTemplate(messageProcessContext.getFlowConstruct().getMuleContext(),
                                         messageProcessContext.getFlowConstruct(),
-                                        (messageProcessContext.getTransactionConfig() == null ? new MuleTransactionConfig()
-                                            : messageProcessContext.getTransactionConfig()),
+                                        (messageProcessContext.getTransactionConfig().isPresent()
+                                            ? messageProcessContext.getTransactionConfig().get() : new MuleTransactionConfig()),
```
marianogonzalez: orEse()

PR #4373, opened 2017-03-28
MULE-12068: add more security configuration to XML related factories

alepulver:


core/src/main/java/org/mule/util/xmlsecurity/DefaultXMLSecureFactories.java
```diff
@@ -69,4 +74,66 @@ public static XMLInputFactory createXmlInputFactory(Boolean externalEntities, Bo
...
+        return factory;
+    }
+
+    public static void configureTransformerFactory(Boolean externalEntities, Boolean expandEntities, TransformerFactory factory)
+    {
+        if (!externalEntities && !expandEntities)
+        {
+            try
+            {
+                factory.setAttribute(XMLConstants.ACCESS_EXTERNAL_STYLESHEET, "");
```
elrodro83: static imports for all XmlConstants

core/src/main/java/org/mule/util/xmlsecurity/DefaultXMLSecureFactories.java
```diff
@@ -69,4 +74,66 @@ public static XMLInputFactory createXmlInputFactory(Boolean externalEntities, Bo
...
+                factory.setAttribute(XMLConstants.ACCESS_EXTERNAL_DTD, "");
+            }
+            catch (Exception e)
+            {
+                logWarning("TransformerFactory", factory.getClass().getName());
+            }
+        }
+    }
+
+    public static void configureSchemaFactory(Boolean externalEntities, Boolean expandEntities, SchemaFactory factory)
```
elrodro83: shouldn't there also be a createTransformerFactory method?

modules/json/src/main/java/org/mule/module/json/transformers/TransformerInputs.java
```diff
@@ -101,6 +103,8 @@ public static TransformerFactory createTransformerFactory()
             transformerFactory = TransformerFactory.newInstance();
         }
 
+        XMLSecureFactories.createDefault().configureTransformerFactory(transformerFactory);
```
elrodro83: move this inside the try, and in the fallback call `createTransformerFactory`

modules/json/src/main/java/org/mule/module/json/transformers/TransformerInputs.java
```diff
@@ -120,6 +124,8 @@ public static SchemaFactory createSchemaFactory(String schemaLanguage)
             schemaFactory = SchemaFactory.newInstance(schemaLanguage);
         }
 
+        XMLSecureFactories.createDefault().configureSchemaFactory(schemaFactory);
```
elrodro83: shouldn't there also be a createSchemaFactory method?

modules/json/src/main/java/org/mule/module/json/transformers/TransformerInputs.java
```diff
@@ -120,6 +124,8 @@ public static SchemaFactory createSchemaFactory(String schemaLanguage)
...
+
         System.out.println(schemaFactory.getClass());
```
elrodro83: remove this sysout now that you're at it

core/src/main/java/org/mule/util/xmlsecurity/DefaultXMLSecureFactories.java
alepulver: As different implementations are used, it's less repetitive to provide a configuration method instead of making a new static method for each case.

modules/json/src/main/java/org/mule/module/json/transformers/TransformerInputs.java
alepulver: If we don't configure both possible factories, the other implementation would be vulnerable.

modules/json/src/main/java/org/mule/module/json/transformers/TransformerInputs.java
alepulver: Same as createTransformerFactory. It also includes an additional parameter for constructing.

core/src/main/java/org/mule/util/xmlsecurity/DefaultXMLSecureFactories.java
elrodro83: then we should also provide configuration methods for other things too, to keep everything consistent

alepulver: We don't want them to configure manually others because it would cause performance issues (configuration was the expensive operation, more than creating factories).

PR #4382, opened 2017-03-29
MULE-12076: TLS Context is not replicated in wsdl retrieval when importing xsds.

fsgonz:


modules/ws/src/main/java/org/mule/module/ws/consumer/MuleWSDLLocator.java
```diff
@@ -0,0 +1,163 @@
...
+    @Override
+    public String getLatestImportURI()
+    {
+        return latestImportedURI;
+    }
+
+    @Override
+    public void close()
+    {
+        if (useConnectorToRetrieveWsdl)
```
alepulver: Move this to a clean() method inside the strategies to avoid these checks (which are also incomplete since it depends on the URL too).

modules/ws/src/main/java/org/mule/module/ws/consumer/MuleWSDLLocator.java
```diff
@@ -0,0 +1,163 @@
...
+                stream.close();
+            }
+            catch (IOException e)
+            {
+                logger.warn("Error closing stream during WSDL retrieval");
+            }
+        }
+    }
+
+    private boolean verifyIsHttpRequester(String url)
```
alepulver: name -> isHttpUrl ?

modules/ws/src/main/java/org/mule/module/ws/consumer/MuleWSDLLocator.java
```diff
@@ -0,0 +1,163 @@
...
+    private boolean useConnectorToRetrieveWsdl;
+    private HttpRequesterWsdlRetrieverStrategy httpRetrieverStrategy;
+    private URLRetrieverStrategy urlRetrieverStrategy;
+    private Collection<InputStream> streams = new ArrayList<InputStream>();
+    
+    public MuleWSDLLocator(MuleWSDLLocatorConfig config) throws MuleException
+    {
+        this.baseURI = config.getBaseURI();
+        this.useConnectorToRetrieveWsdl = config.isUseConnectorToRetrieveWsdl();
+        createURLRetrieverStrategy();
```
alepulver: Don't create here, see below.

modules/ws/src/main/java/org/mule/module/ws/consumer/MuleWSDLLocator.java
```diff
@@ -0,0 +1,163 @@
...
+
+    }
+
+    private InputSource getInputSource(String url) throws WSDLException
+    {
+        boolean isHttpRequester = verifyIsHttpRequester(url);
+
+        InputStream resultStream = null;
+
+        if (useConnectorToRetrieveWsdl && isHttpRequester)
```
alepulver: Create the corresponding strategy here and use interface to call retrieveWsdl.

fsgonz: Now it is arranged through an abstract pattern. I get a concrete factory according to the strategy to use, create the strategy and then I use the interface method to retrieveWsdl.

modules/ws/src/main/java/org/mule/module/ws/consumer/MuleWSDLLocator.java
fsgonz: I move the clean logic to the strategy. As I create the strategy each time the WSDLLocator method to create a InputSource is invoked, the httpClient is not reused

PR #4385, opened 2017-03-30
MULE-11939: HTTP extension does not leverage cursor streaming - chang…

marianogonzalez:
…e approach


modules/extensions-spring-support/src/test/resources/schemas/marvel.xsd
```diff
@@ -33,8 +33,8 @@
         <xs:complexContent>
             <xs:extension xmlns:mule="http://www.mulesoft.org/schema/mule/core" base="mule:abstractMessageSourceType">
                 <xs:sequence minOccurs="0" maxOccurs="1">
-                    <xs:element minOccurs="0" maxOccurs="1" ref="mule:abstract-byte-streaming-strategy"></xs:element>
                     <xs:element minOccurs="0" maxOccurs="1" ref="mule:abstract-redelivery-policy"></xs:element>
+                    <xs:element minOccurs="0" maxOccurs="1" ref="mule:abstract-byte-streaming-strategy"></xs:element>
```
alegmarra: was this order change intentional or a side effect? is not important right now, but after beta we would be breaking backwards.

marianogonzalez: this is because in my prior PR i changed the source to not be configless. I'm actually fixing the build with this (shame shame shame shame)

PR #4389, opened 2017-03-30
MULE-12013: FTP copy doesn't work if you want to copy the file in the…

pablomusumeci:
… same folder with other name

### DIFF
--- a/extensions/file/src/main/java/org/mule/extension/file/internal/FileOperations.java
+++ b/extensions/file/src/main/java/org/mule/extension/file/internal/FileOperations.java
@@ -162,14 +162,15 @@ public final class FileOperations extends BaseFileSystemOperations {
    * @param targetPath              the target directory where the file is going to be copied
    * @param createParentDirectories whether or not to attempt creating any parent directories which don't exists.
    * @param overwrite               whether or not overwrite the file if the target destination already exists.
+   * @param renameTo                the new file name, {@code null} if the file doesn't need to be renamed
    * @throws IllegalArgumentException if an illegal combination of arguments is supplied
    */
   @Summary("Copies a file in another directory")
   @Throws(FileCopyErrorTypeProvider.class)
   public void copy(@UseConfig FileConnectorConfig config, @Connection FileSystem fileSystem, String sourcePath,
                    String targetPath, @Optional(defaultValue = "true") boolean createParentDirectories,
-                   @Optional(defaultValue = "false") boolean overwrite) {
-    super.doCopy(config, fileSystem, sourcePath, targetPath, createParentDirectories, overwrite);
+                   @Optional(defaultValue = "false") boolean overwrite, @Optional String renameTo) {
+    super.doCopy(config, fileSystem, sourcePath, targetPath, createParentDirectories, overwrite, renameTo);
   }
 
   /**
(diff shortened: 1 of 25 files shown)


extensions/file/src/main/java/org/mule/extension/file/internal/FileOperations.java
```diff
@@ -162,14 +162,15 @@ public void write(@UseConfig FileConnectorConfig config, @Connection FileSystem
    * @param targetPath              the target directory where the file is going to be copied
    * @param createParentDirectories whether or not to attempt creating any parent directories which don't exists.
    * @param overwrite               whether or not overwrite the file if the target destination already exists.
+   * @param renameTo                the new file name, {@code null} if the file doesn't need to be renamed
```
marianogonzalez: we need to shift the wording on the jdocs to be less technical and more user oriented. Rephrase this explaining that if provided, the copy will have this new name. Otherwise, the current name will be honoured.

extensions/file/src/main/java/org/mule/extension/file/internal/FileOperations.java
```diff
@@ -162,14 +162,15 @@ public void write(@UseConfig FileConnectorConfig config, @Connection FileSystem
...
    * @throws IllegalArgumentException if an illegal combination of arguments is supplied
    */
   @Summary("Copies a file in another directory")
```
marianogonzalez: change this. The whole point of this change is that you can copy it in the same directory.

extensions/file/src/main/java/org/mule/extension/file/internal/FileOperations.java
```diff
@@ -191,14 +192,15 @@ public void copy(@UseConfig FileConnectorConfig config, @Connection FileSystem f
    * @param targetPath              the target directory
    * @param createParentDirectories whether or not to attempt creating any parent directories which don't exists.
    * @param overwrite               whether or not overwrite the file if the target destination already exists.
+   * @param renameTo                the new file name, {@code null} if the file doesn't need to be renamed
```
marianogonzalez: idem wording

extensions/file/src/main/java/org/mule/extension/file/internal/FileOperations.java
```diff
@@ -191,14 +192,15 @@ public void copy(@UseConfig FileConnectorConfig config, @Connection FileSystem f
...
    * @throws IllegalArgumentException if an illegal combination of arguments is supplied
    */
   @Summary("Moves a file to another directory")
```
marianogonzalez: idem

extensions/file/src/main/java/org/mule/extension/file/internal/command/AbstractLocalCopyCommand.java
```diff
@@ -47,26 +47,29 @@
    * @param target the path to the target destination
    * @param overwrite whether to overwrite existing target paths
    * @param createParentDirectory whether to create the target's parent directory if it doesn't exists
+   * @param renameTo the new file name, {@code null} if the file doesn't need to be renamed
```
marianogonzalez: just in case, this jdoc is NOT reached by the documentation generator so it's ok to stay technical here

extensions/file/src/main/java/org/mule/extension/file/internal/command/LocalWriteCommand.java
```diff
@@ -64,7 +65,7 @@ public void write(String filePath, InputStream content, FileWriteMode mode,
       throw new FileAccessDeniedException(format("Could not write to file '%s' because access was denied by the operating system",
                                                  path),
                                           e);
-    } catch (org.mule.extension.file.common.api.exceptions.FileAlreadyExistsException e) {
```
marianogonzalez: why this? Is there a test which verifies that the associated error type is still catchable?

extensions/ftp/src/main/java/org/mule/extension/ftp/internal/FtpOperations.java
```diff
@@ -162,14 +162,15 @@ public void write(@UseConfig FileConnectorConfig config, @Connection FileSystem
    * @param targetPath              the target directory where the file is going to be copied
    * @param createParentDirectories whether or not to attempt creating any parent directories which don't exists.
    * @param overwrite               whether or not overwrite the file if the target destination already exists.
+   * @param renameTo                the new file name, {@code null} if the file doesn't need to be renamed
```
marianogonzalez: jdoc

PR #4390, opened 2017-03-30
MULE-10617 Migrate Error Handling API to use Non-Blocking API

dfeist:
PENDING GREEN BUILD.

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/exception/ErrorHandler.java
+++ b/core/src/main/java/org/mule/runtime/core/exception/ErrorHandler.java
@@ -31,6 +31,7 @@ import org.mule.runtime.core.message.DefaultExceptionPayload;
 import org.mule.runtime.core.processor.AbstractMuleObjectOwner;
 
 import org.reactivestreams.Publisher;
+import reactor.core.Exceptions;
 import reactor.core.publisher.Mono;
 
 /**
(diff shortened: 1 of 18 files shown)


tests/integration/src/test/resources/non-blocking-fully-supported-test-config.xml
```diff
@@ -229,11 +229,30 @@
         <error-handler>
             <on-error-continue>
                 <test:assert-intercepting responseSameTask="false" responseExpression="#[mel:assertPayload()]"/>
-                <set-variable variableName="beforeThread" value="#[mel:Thread.currentThread()]"/>
```
elrodro83: doesn't this apply anymore?

core/src/main/java/org/mule/runtime/core/exception/ErrorHandler.java
```diff
@@ -54,18 +59,24 @@ public void setGlobalName(String globalName) {
...
+            .exceptionPayload(new DefaultExceptionPayload(exception)).build())
         .build();
     if (isCriticalException(exception)) {
-      return event;
+      return just(event);
     }
     for (MessagingExceptionHandlerAcceptor exceptionListener : exceptionListeners) {
       if (exceptionListener.accept(event)) {
-        return exceptionListener.handleException(exception, event);
+        return exceptionListener.apply(new MessagingException(event, exception));
```
elrodro83: why build a new exception?

PR #4402, opened 2017-03-31
MULE-11998: It should be possible to define endpoints with different names and the same address for stp/ftp connectors

fsgonz:


core/src/main/java/org/mule/endpoint/DefaultEndpointFactory.java
```diff
@@ -34,19 +34,30 @@
 
     protected MuleContext muleContext;
 
+    @Override
+    public InboundEndpoint getInboundEndpoint(String uri, String name) throws MuleException
```
alepulver: It appears to be used in tests only. If that's the case, it shouldn't be here.

core/src/main/java/org/mule/api/endpoint/EndpointFactory.java
```diff
@@ -30,6 +30,18 @@
      * MULE-2292
      * 
      * @param uri endpoint identifier or uri
+     * @param name name of the endpoint
+     * @throws MuleException
+     */
+    InboundEndpoint getInboundEndpoint(String uri, String name) throws MuleException;
```
alepulver: See comment below, also applies here.

PR #4406, opened 2017-03-31
MULE-11885 - Add support to trigger a poll execution

pablolagreca:


tests/integration/src/test/java/org/mule/test/integration/SchedulerManagementTestCase.java
```diff
@@ -0,0 +1,118 @@
...
+  @Description("scheduler start twice does not fail")
+  @Test
+  public void startTwiceDoesNotFail() throws MuleException {
+    SchedulerMessageSource scheduler = (DefaultSchedulerMessageSource) muleContext.getConfigurationComponentLocator()
+        .find(builder().globalName("schedulerControlledFromSameFlow").addSourcePart().build()).get();
+    scheduler.start();
+    scheduler.start();
+  }
+
+  @Description("scheduler start twice does not fail")
```
afelisatti: stopped

core/src/main/java/org/mule/runtime/core/source/scheduler/DefaultSchedulerMessageSource.java
```diff
@@ -85,18 +88,31 @@ public String getPollingUniqueName() {
...
+  @Override
+  public boolean isStarted() {
+    return started;
   }
 
   /**
    * Checks whether polling should take place on this instance.
    */
-  public final void performPoll() {
+  public final void run() {
```
dfeist: `implements Runnable`?

core/src/main/java/org/mule/runtime/core/source/scheduler/DefaultSchedulerMessageSource.java
```diff
@@ -85,18 +88,31 @@ public String getPollingUniqueName() {
...
+    }
     // Stop the scheduler to address the case when the flow is stop but not the application
     if (schedulingJob != null) {
       schedulingJob.cancel(false);
       schedulingJob = null;
     }
+    this.started = false;
+  }
+
+  public void trigger() {
```
dfeist: Can you explain in javadoc the difference between run() and trigger() given both are public.

modules/schedulers/src/test/java/org/mule/test/modules/schedulers/cron/CronsSchedulerTestCase.java
```diff
@@ -15,7 +15,7 @@
 import org.mule.runtime.api.exception.MuleException;
```
dfeist: Not sure when this was committed, but it should be singular. `CronsSchedulerTestCase.java` -> `CronSchedulerTestCase.java`

modules/schedulers/src/test/java/org/mule/test/modules/schedulers/cron/StoppedCronSchedulerTestCase.java
```diff
@@ -60,8 +60,8 @@ private void runSchedulersOnce() throws Exception {
     flow.start();
     try {
       MessageSource flowSource = flow.getMessageSource();
-      if (flowSource instanceof SchedulerMessageSource) {
-        ((SchedulerMessageSource) flowSource).performPoll();
+      if (flowSource instanceof DefaultSchedulerMessageSource) {
+        ((DefaultSchedulerMessageSource) flowSource).run();
```
dfeist: Any reason to use `run()` and the the more direct `trigger()` here? (assuming i understood difference)

tests/integration/src/test/resources/org/mule/test/integration/locator/component-locator-config.xml
```diff
@@ -14,5 +14,10 @@
         </async>
     </flow>
 
+    <flow name="anotherFlow">
+        <scheduler/>
```
dfeist: Is this valid?  I'm not convinced having default scheduler is a good idea, how does this work?

pablolagreca: it works as it used to work with <poll> I don't why it would be a problem to have a default value for the frequency.

modules/schedulers/src/test/java/org/mule/test/modules/schedulers/cron/CronsSchedulerTestCase.java
pablolagreca: done

tests/integration/src/test/resources/org/mule/test/integration/locator/component-locator-config.xml
dfeist: I'm not convinced about default frequency, but it makes some sense.  The concern here is the default/implicit scheduler, this is using fixed-freq vs. cron or other scheduler but this is not at all clear.

core/src/main/java/org/mule/runtime/core/source/scheduler/DefaultSchedulerMessageSource.java
pablolagreca: Doesn't make much sense to use the class a runnable

core/src/main/java/org/mule/runtime/core/source/scheduler/DefaultSchedulerMessageSource.java
pablolagreca: I just make run private and use the performPoll underneeth.

PR #4408, opened 2017-03-31
MULE-10412 Adding support for missing variables app, server, mule.

machaval:
Migrating extension services from mel to dw

### DIFF
--- a/core-tests/src/test/java/org/mule/runtime/core/el/DataWeaveExpressionLanguageTestCase.java
+++ b/core-tests/src/test/java/org/mule/runtime/core/el/DataWeaveExpressionLanguageTestCase.java
@@ -7,6 +7,7 @@
 package org.mule.runtime.core.el;
 
 import static java.util.Optional.empty;
+import static org.apache.commons.lang.SystemUtils.FILE_SEPARATOR;
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.hasEntry;
@@ -222,7 +223,7 @@ public class DataWeaveExpressionLanguageTestCase extends AbstractMuleContextTest
     Event event = testEvent();
     muleContext.getRegistry().registerObject("myBean", new MyBean("DataWeave"));
     TypedValue evaluate = expressionLanguage.evaluate("server.fileSeparator", event, BindingContext.builder().build());
-    assertThat(evaluate.getValue(), is(SystemUtils.FILE_SEPARATOR));
+    assertThat(evaluate.getValue(), is(FILE_SEPARATOR));
   }
 
   @Test
--- a/core-tests/src/test/java/org/mule/runtime/core/el/DataWeaveExpressionLanguageTestCase.java
+++ b/core-tests/src/test/java/org/mule/runtime/core/el/DataWeaveExpressionLanguageTestCase.java
@@ -30,6 +30,7 @@ import static org.mule.runtime.core.el.DataWeaveExpressionLanguage.FLOW;
 import static org.mule.runtime.core.el.DataWeaveExpressionLanguage.PAYLOAD;
 import static org.mule.runtime.core.el.DataWeaveExpressionLanguage.VARIABLES;
 
+import org.junit.Before;
 import org.mule.runtime.api.el.BindingContext;
 import org.mule.runtime.api.exception.MuleException;
 import org.mule.runtime.api.message.Attributes;
@@ -42,10 +43,10 @@ import org.mule.runtime.core.api.Event;
 import org.mule.runtime.core.api.construct.FlowConstruct;
 import org.mule.runtime.core.api.expression.ExpressionRuntimeException;
 import org.mule.runtime.core.api.message.InternalMessage;
+import org.mule.runtime.core.config.MuleManifest;
 import org.mule.runtime.core.message.BaseAttributes;
-import org.mule.tck.junit4.AbstractMuleTestCase;
-
-import com.mulesoft.weave.el.WeaveExpressionExecutor;
+import org.mule.runtime.core.util.SystemUtils;
+import org.mule.tck.junit4.AbstractMuleContextTestCase;
 
 import com.google.common.collect.Sets;
 
@@ -62,13 +63,17 @@ import ru.yandex.qatools.allure.annotations.Stories;
 
 @Features("Expression Language")
 @Stories("Support DW")
-public class DataWeaveExpressionLanguageTestCase extends AbstractMuleTestCase {
+public class DataWeaveExpressionLanguageTestCase extends AbstractMuleContextTestCase {
 
-  private DataWeaveExpressionLanguage expressionLanguage =
-      new DataWeaveExpressionLanguage(new WeaveExpressionExecutor());
+  private DataWeaveExpressionLanguage expressionLanguage;
 
   @Rule
   public ExpectedException expectedEx = ExpectedException.none();
+
+  @Before
+  public void setUp() {
+    expressionLanguage = new DataWeaveExpressionLanguage(muleContext);
+  }
 
   @Test
   public void stringExpression() throws Exception {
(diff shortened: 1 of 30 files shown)


core-tests/src/test/java/org/mule/runtime/core/el/DataWeaveExpressionLanguageTestCase.java
```diff
@@ -205,6 +210,30 @@ public void variablesCannotOverrideEventBindings() throws MuleException {
...
+    TypedValue evaluate = expressionLanguage.evaluate("app.registry.myBean.name", event, BindingContext.builder().build());
+    assertThat(evaluate.getValue(), is("DataWeave"));
+  }
+
+  @Test
+  public void accessServerFileSeparator() throws MuleException {
+    Event event = testEvent();
+    muleContext.getRegistry().registerObject("myBean", new MyBean("DataWeave"));
+    TypedValue evaluate = expressionLanguage.evaluate("server.fileSeparator", event, BindingContext.builder().build());
+    assertThat(evaluate.getValue(), is(SystemUtils.FILE_SEPARATOR));
```
afelisatti: Static import

core/src/main/java/org/mule/runtime/core/el/DataWeaveExpressionLanguage.java
```diff
@@ -45,14 +51,24 @@
...
+  private MuleContext muleContext;
 
   @Inject
-  public DataWeaveExpressionLanguage(ExpressionExecutor expressionExecutor) {
-    this.expressionExecutor = expressionExecutor;
+  public DataWeaveExpressionLanguage(MuleContext muleContext) {
+    try {
+      this.expressionExecutor = muleContext.getRegistry().lookupObject(ExpressionExecutor.class);
+    } catch (RegistrationException e) {
+      throw new MuleRuntimeException(e);
```
afelisatti: Add a descriptive message saying the executor could not be obtained or something

core/src/main/java/org/mule/runtime/core/el/DataWeaveExpressionLanguage.java
```diff
@@ -123,14 +139,19 @@ private void addEventBindings(Event event, BindingContext.Builder contextBuilder
...
-                                new TypedValue(unmodifiableMap(flowVars), fromType(flowVars.getClass())));
+                                new TypedValue<>(unmodifiableMap(flowVars), fromType(flowVars.getClass())));
       Message message = event.getMessage();
       Attributes attributes = message.getAttributes();
-      contextBuilder.addBinding(ATTRIBUTES, new TypedValue(attributes, fromType(attributes.getClass())));
+      contextBuilder.addBinding(ATTRIBUTES, new TypedValue<>(attributes, fromType(attributes.getClass())));
       contextBuilder.addBinding(PAYLOAD, message.getPayload());
-      contextBuilder.addBinding(DATA_TYPE, new TypedValue(message.getPayload().getDataType(), fromType(DataType.class)));
+      contextBuilder.addBinding(DATA_TYPE, new TypedValue<>(message.getPayload().getDataType(), fromType(DataType.class)));
+      contextBuilder.addBinding(MULE,
```
afelisatti: This means the mule, server and app variables will be added per Event but they are not related. They should be global bindings instead.

core/src/main/java/org/mule/runtime/core/el/context/ExtendedServerContext.java
```diff
@@ -0,0 +1,47 @@
...
+ * <li><b>isAfter(DateTimeContext date)</b> <i>Boolean. Returns true if the date parameter is after the current DateTime.</i>
+ * <li><b>addSeconds(int seconds)</b> <i>Add n seconds to the current DateTime. Returns DateTime for method chaining.</i>
+ * <li><b>addMinutes(int minutes)</b> <i>Add n minutes to the current DateTime. Returns DateTime for method chaining.</i>
+ * <li><b>addHours(int hours)</b> <i>Add n hours to the current DateTime. Returns DateTime for method chaining.</i>
+ * <li><b>addDay(int days)</b> <i>Add n days to the current DateTime. Returns DateTime for method chaining.</i>
+ * <li><b>addWeeks(int weeks)</b> <i>Add n weeks to the current DateTime. Returns DateTime for method chaining.</i>
+ * <li><b>addMonths(int months)</b> <i>Add n months to the current DateTime. Returns DateTime for method chaining.</i>
+ * <li><b>addYears(int years)</b> <i>Add n years to the current DateTime. Returns DateTime for method chaining.</i>
+ */
+public class ExtendedServerContext extends ServerContext {
```
afelisatti: Could you add that this is to support Mule 3 and the new approach (dw)?

core/src/main/java/org/mule/runtime/core/el/DataWeaveExpressionLanguage.java
```diff
@@ -123,14 +139,19 @@ private void addEventBindings(Event event, BindingContext.Builder contextBuilder
...
+                                new TypedValue<>(new MuleInstanceContext(muleContext), fromType(MuleInstanceContext.class)));
+      contextBuilder.addBinding(SERVER, new TypedValue<>(new ServerContext(), fromType(ServerContext.class)));
+      contextBuilder.addBinding(APP, new TypedValue<>(new AppContext(muleContext), fromType(AppContext.class)));
```
afelisatti: AFAIK, we don't want to expose the whole registry to users. I think we should discuss having a different view for the AppContext.

core/src/main/java/org/mule/runtime/core/el/DataWeaveExpressionLanguage.java
machaval: All this is going to be changed once I make it a server then mule app and server are going to be global

core/src/main/java/org/mule/runtime/core/el/DataWeaveExpressionLanguage.java
machaval: The AppContext is what we used to expose in mvel. And currently there is no way to filter. Pablo LG said that for now we do this and later when spring is revisited they are going to filter it

core/src/main/java/org/mule/runtime/core/el/context/ExtendedServerContext.java
machaval: What do you mean?

afelisatti: I'm guessing this is meant to separate the date stuff which should now be done through DW from the rest, right? So I think we should clear that up in the javadoc so that's it's easy to understand the division.

core/src/main/java/org/mule/runtime/core/el/DataWeaveExpressionLanguage.java
afelisatti: Go it.

core/src/main/java/org/mule/runtime/core/el/DataWeaveExpressionLanguage.java
afelisatti: What do you mean? I though since these only depend on the muleContext you could add them as globals when the language is created/initialised.

core/src/main/java/org/mule/runtime/core/el/context/ExtendedServerContext.java
machaval: Ok I will add a comment to the javadoc specifing that this is for compatibilty only

core/src/main/java/org/mule/runtime/core/el/DataWeaveExpressionLanguage.java
machaval: exactly but currently (in this pull request) is still a singleton (service) so I can not add them as global. I need to add all as local bindings. Upcoming pull request will fix this.

afelisatti: Damn service...

afelisatti:
@machaval there's a compilation failure due to DataWeaveExpressionLanguage.java

afelisatti:
jenkins validate

afelisatti:
jenkins validate

PR #4412, opened 2017-03-31
MULE-12112: Database Connector unable to handle CLOB during sql query

fsgonz:


modules/db/src/main/java/org/mule/module/db/internal/domain/type/MetadataDbTypeManager.java
```diff
@@ -101,7 +101,16 @@ protected void initialise(DbConnection connection)
                 Number data_type = (Number) typeRecord.get(METADATA_TYPE_ID_COLUMN);
                 String type_name = (String) typeRecord.get(METADATA_TYPE_NAME_COLUMN);
 
-                ResolvedDbType resolvedDbType = new ResolvedDbType(data_type.intValue(), type_name);
+                ResolvedDbType resolvedDbType = null;
+
+                if (data_type.intValue() == Types.CLOB)
```
pablokraan: I'm not sure that this change is OK. This kind of types should be being managed by StaticDbTypeManager. Did you review that class and how is being used?

fsgonz: In case the type is not resolved until it is instantiated (as it happens in  this case), the composite manager tries to resolve the type using the metadata manager by means of the id and the name of the class. As CLOB is a type retrieved from metadata, it is resolved as a ResolvedDbType and the staticDbTypeManager is not invoked because the metadata resolver has a higher priority. Furthermore, in case a type is not resolved in metadata, for thess casesm, it is not resolved from the static resolver either as the static resolver cannot handle type resolution by id and name (it handles it only by name). If it cannot be resolved a ResolvedDbType is returned (which would also result in an error, as the conversion from string would fail). As a ClobResolvedDbType is needed to handle the conversion from String to Clob automatically, I think the MetadataDbTypeManager should be used and it has to return a ClobResolvedDbType.

PR #4415, opened 2017-04-03
MULE-11070: Remove legacy processing strategies

elrodro83:
* Fix flaky tests


core-tests/src/test/java/org/mule/runtime/core/processor/AsyncDelegateMessageProcessorTestCase.java
```diff
@@ -94,8 +96,12 @@ public void process() throws Exception {
     // Block until async completes, not just target processor.
     from(target.sensedEvent.getContext().getCompletionPublisher()).block(ofMillis(BLOCK_TIMEOUT));
     assertThat(target.sensedEvent, notNullValue());
-    assertCompletionDone(target.sensedEvent.getContext());
-    assertCompletionDone(request.getContext());
+
```
dfeist: You can use block() with timeout to wait for completion, no need to poll.

PR #4416, opened 2017-04-03
MULE-11642: OAuth config is not initialized at application startup

elrodro83:


extensions/oauth/src/main/java/org/mule/extension/oauth2/internal/clientcredentials/ClientCredentialsGrantType.java
```diff
@@ -27,7 +28,11 @@
 
 /**
  * Authorization element for client credentials oauth grant type
+ * <p>
+ * When used as a top-level element, any initialization (e.g., first token refresh) will be done when the application is
```
pablolagreca: Is first token refresh done on start method?

extensions/oauth/src/test/resources/authorization-code/authorization-code-localcallbackref-config.xml
```diff
@@ -14,22 +14,21 @@
 
     <oauth:token-manager-config name="tokenManagerConfig"/>
 
+    <oauth:authorization-code-grant-type
```
pablolagreca: Do we still support this config inside http:request-connection?. And if so, why do you had to move it to a top level config?

marianogonzalez: I find it very weird that one annotation is the only change in the productive code. Seems to me like you're fixing the bug by simply defining the grant type as a top level element. If this is so, then we need to:

* Have @pablolagreca to validate that is fine
* Define the grant type not only as supports top level definition, you must also set it as doesn't support inline definition

If this is not the case, then you're missing tests in which you use it inline.

elrodro83: @marianogonzalez  How can you define it isn allowed to be inline?

elrodro83: @marianogonzalez added the parameter no not allow these elements inline. Still need to confirm with @pablolagreca if the usability for this is fine. If not, we'll have to review the lifecycles of beans nested in dynamic configs again.

marianogonzalez: The problem with that approach is that Mozart won't be able to use it since they don't handle top level elements.

That been said, the OAuth module does require the manager and other global stuff to be registered and it's not among the most common use cases. So maybe this is fine and Mozart will have to eventually overcome this limitation. 

We can also review the lifecycle again or find a third alternative. Awaiting for @pablolagreca 's input

elrodro83:
Fixed by https://github.com/mulesoft/mule/pull/4500

PR #4419, opened 2017-04-03
MULE-11617: Add missing extensions descriptions

juandesi:


marianogonzalez:
NOT. That description needs to go away and use the jdoc instead

PR #4420, opened 2017-04-03
MULE-12068: improvements to use cache in TransformerFactory

alepulver:


modules/xml/src/test/java/org/mule/util/XmlLibrariesVerificationTestCase.java
```diff
@@ -58,4 +59,11 @@ public void transformer()
         TransformerFactory factory = TransformerFactory.newInstance();
         assertThat(factory, instanceOf(org.apache.xalan.processor.TransformerFactoryImpl.class));
     }
+
+    @Test
+    public void schema()
```
elrodro83: can you confirm the Validator implementation won't change based on how the classpath is fromed?

modules/json/src/main/java/org/mule/module/json/transformers/TransformerInputs.java
```diff
@@ -96,15 +96,14 @@ public static TransformerFactory createTransformerFactory()
         {
             // Create a factory we know to be STAX-compliant
             transformerFactory = (TransformerFactory) Class.forName(PREFERRED_TRANSFORMATION_FACTORY_CLASS_NAME).newInstance();
+            XMLSecureFactories.createDefault().configureTransformerFactory(transformerFactory);
         }
         catch (Exception ex)
         {
             // Fall back to default factory
-            transformerFactory = TransformerFactory.newInstance();
+            transformerFactory = XMLSecureFactories.createDefault().getTransformerFactory();
```
elrodro83: Have you checked the other modules for occurrences of this code to migrate?

modules/xml/src/test/java/org/mule/util/XmlLibrariesVerificationTestCase.java
alepulver: It is generated from the SchemaLoader instance, so there is a direct dependency.

modules/json/src/main/java/org/mule/module/json/transformers/TransformerInputs.java
alepulver: For now, AES has uses of TransformerFactory.

PR #4430, opened 2017-04-03
MULE-11951 - ArtifactInfo should know about log4j.xml file - revertin…

pablolagreca:
…g dependency removed

### DIFF
--- a/embedded/embedded-impl/pom.xml
+++ b/embedded/embedded-impl/pom.xml
@@ -34,6 +34,12 @@
             <artifactId>mule-core</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <!-- TODO MULE-11458 - this dependency is not downloaded by maven because the zip extension -->
+        <dependency>
+            <groupId>log4j</groupId>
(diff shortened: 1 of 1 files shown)


embedded/embedded-impl/pom.xml
```diff
@@ -34,6 +34,12 @@
             <artifactId>mule-core</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <!-- TODO MULE-11458 - this dependency is not downloaded by maven because the zip extension -->
+        <dependency>
+            <groupId>log4j</groupId>
+            <artifactId>log4j</artifactId>
+            <version>1.2.17</version>
```
marianogonzalez: why this instead of the slf4j bridge?

pablolagreca: This is revert of a previous code. For some reason the services are using it.

marianogonzalez: I'll merge it, but with ethical concerns. Will talk to @gsfernandes about this

PR #4439, opened 2017-04-04
MULE-12145: Sort URLs used to create the container classloader create…

pablokraan:
…d on the test runner

_ Temporal workaround


pablokraan:
jenkins validate

PR #4440, opened 2017-04-04
MULE-12147: add proxy support for SFTP connector

alepulver:


transports/sftp/src/main/java/org/mule/transport/sftp/SftpClient.java
```diff
@@ -214,6 +220,19 @@ public void login(String user, String identityFile, String passphrase) throws IO
         }
     }
 
+    private void configureProxy(Session session)
+    {
+        if (proxyConfig != null)
+        {
+            ProxyHTTP proxy = new ProxyHTTP(proxyConfig.getHost(), proxyConfig.getPort());
```
marianogonzalez: ProxyHTTP is not the most common use case. User should be able to choose from SocksS5, SocksS4 and Http

transports/sftp/src/main/java/org/mule/transport/sftp/config/SftpProxyConfig.java
```diff
@@ -0,0 +1,58 @@
...
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.transport.sftp.config;
+
+/**
+ * A Proxy configuration for the SFTP connector.
+ */
+public class SftpProxyConfig
```
marianogonzalez: `@since`

transports/sftp/src/test/resources/sftp-proxy-config.xml
```diff
@@ -0,0 +1,26 @@
...
+    </spring:beans>
+
+    <sftp:connector name="proxyConnector">
+        <sftp:proxy-config host="localhost" port="${proxyPort}"/>
+    </sftp:connector>
+
+    <flow name="sftpThroughProxy">
+        <!-- Use an invalid address so that it will fail if it doesn't go through the proxy -->
+        <sftp:inbound-endpoint
+            address="sftp://${USER1_NAME}:${USER1_PASSWORD}@127.0.0.100:${SFTP_PORT}/~/testdir"
```
marianogonzalez: shouldn't you get an error if you provide address and proxy at the same time?

alepulver: No, the address is the real one. The proxy is just the middle man connecting to the address.
It's probably confusing because the test proxy ignores the CONNECT argument, but a real proxy connects to whatever address one tells it to do.

PR #4441, opened 2017-04-05
MULE-11767: Improve TLS API

afelisatti:


modules/tls/src/main/java/org/mule/runtime/module/tls/internal/DefaultTlsContextFactory.java
```diff
@@ -213,12 +218,24 @@ public SSLContext createSslContext() throws KeyManagementException, NoSuchAlgori
...
+  public SSLServerSocketFactory getServerSocketFactory() throws KeyManagementException, NoSuchAlgorithmException {
+    return new RestrictedSSLServerSocketFactory(createSslContext(), getEnabledCipherSuites(), getEnabledProtocols());
+  }
+
+  @Override
   public String[] getEnabledCipherSuites() {
-    if (isUseDefaults(enabledCipherSuites)) {
-      return tlsConfiguration.getEnabledCipherSuites();
+    String[] enabledCipherSuites;
+    if (isUseDefaults(this.enabledCipherSuites)) {
```
pablolagreca: Question: When FIPS is enabled in the runtime, does it means that not all cipher suites can be used?. If that's the case, and the user configures a cipher suite not supported by FIPS, are we failing in that case?

modules/tls/src/main/java/org/mule/runtime/module/tls/internal/MuleTlsContextFactoryBuilder.java
```diff
@@ -6,44 +6,15 @@
...
-import org.mule.runtime.core.api.context.MuleContextAware;
-import org.mule.runtime.api.lifecycle.Initialisable;
-import org.mule.runtime.api.lifecycle.InitialisationException;
-import org.mule.runtime.module.tls.api.DefaultTlsContextFactoryBuilder;
 
-import java.util.concurrent.atomic.AtomicBoolean;
-
-@DefaultTlsContextFactoryBuilder
-public class MuleTlsContextFactoryBuilder implements TlsContextFactoryBuilder, Initialisable, MuleContextAware {
+public class MuleTlsContextFactoryBuilder implements TlsContextFactoryBuilder {
```
pablolagreca: Seems we are using "Default" everywhere instead of "Mule"

modules/tls/src/main/java/org/mule/runtime/module/tls/internal/DefaultTlsContextFactory.java
afelisatti: Yes, there's a check on initialise() where (if set) the protocols and ciphers are intersected with the global ones to check for differences, failing if that is the case.

PR #4442, opened 2017-04-05
MULE-12143 Exception raised in Mule Service is wrapped in UndeclaredThrowableException

dfeist:
…DW expression to throw error of type UNKNOWN instead of EXPRESSION


pablokraan:
jenkins validate

PR #4445, opened 2017-04-06
MULE-11838: Support flow paths in ElementDeclaration

alegmarra:


modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -2034,7 +2034,7 @@
                             </xsd:complexType>
                         </xsd:element>
                     </xsd:choice>
-                    <xsd:choice>
+                    <xsd:choice minOccurs="0" maxOccurs="1">
```
marianogonzalez: this requires a full build to be run before merging

PR #4446, opened 2017-04-06
MULE-12153: WS RequestBodyGenerator fails and generates a warning when imports/includes demands a TLS context.

fsgonz:


marianogonzalez:
Please check with @juandesi if this is an issue in Mule 4 too

aiannucci:
The whitelist must be updated now that the new version of XML Beans is being pulled by mule-common

PR #4447, opened 2017-04-06
MULE-12068: add JDK5 TransformerFactory fix to XMLSecureFactories

alepulver:


elrodro83:
Also, please add a test case of the scenario this is fixing

PR #4454, opened 2017-04-07
MULE-9122: Remove management module

mbuchwald:


pablolagreca:
Don't you need to remove this module and it's dependencies from assembly.xml and assembly-whitelist.txt?

PR #4459, opened 2017-04-07
MULE-12165: Echo application is blocked reserving and unreserving dir…

marianogonzalez:
…ect memory, which also causes OOM.


core/src/main/java/org/mule/runtime/core/internal/streaming/bytes/AbstractInputStreamBuffer.java
```diff
@@ -32,13 +27,12 @@
  */
 public abstract class AbstractInputStreamBuffer extends AbstractStreamingBuffer implements InputStreamBuffer {
 
-  private final ByteBufferManager bufferManager;
+  protected final ByteBufferManager bufferManager;
```
dfeist: Why not keep private with getter?

core/src/main/java/org/mule/runtime/core/internal/streaming/object/iterator/ConsumerStreamingIterator.java
```diff
@@ -20,7 +20,7 @@
  *
  * @since 4.0
  */
-public final class ConsumerStreamingIterator<T> implements StreamingIterator<T> {
+public class ConsumerStreamingIterator<T> implements StreamingIterator<T> {
```
dfeist: Can't this stay final, don't see any extension..

core/src/main/java/org/mule/runtime/core/internal/streaming/bytes/AbstractInputStreamBuffer.java
marianogonzalez: gives no value since the field is final and all these classes are internal

core/src/main/java/org/mule/runtime/core/internal/streaming/object/iterator/ConsumerStreamingIterator.java
marianogonzalez: there're test in EE which are broken because they try to mock this class

PR #4461, opened 2017-04-07
MULE-12118: Add support for multiple consumers for JMS Listener

estebanwasinger:


extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsListener.java
```diff
@@ -142,45 +141,47 @@
   @Optional(defaultValue = "true")
   private boolean synchronous;
 
-  private JmsListenerLock jmsLock;
+  /**
+   * The number of concurrent consumers that will be used to receive JMS Messages
+   */
+  @Parameter
+  @Optional(defaultValue = "4")
```
estebanwasinger: Using the same default value as JMS for 3.x

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsListener.java
```diff
@@ -142,45 +141,47 @@
...
+
+    numberOfConsumers = getValidNumberOfConsumers(numberOfConsumers);
 
     try {
-      session = connection.createSession(ackMode, consumerType.isTopic());
-      jmsSupport = connection.getJmsSupport();
-      final Destination jmsDestination = jmsSupport.createDestination(session.get(), destination, consumerType.isTopic());
-      final JmsMessageConsumer consumer = connection.createConsumer(session.get(), jmsDestination, selector, consumerType);
+      for (int i = 0; i < numberOfConsumers; i++) {
+        LOGGER.debug("Starting JMS Listener with [" + numberOfConsumers + "] consumers");
```
estebanwasinger: move outside of the `for()`

extensions/jms/src/main/java/org/mule/extensions/jms/JmsSessionManager.java
```diff
@@ -108,13 +110,18 @@ public void recoverSession(String ackId) throws JMSException {
...
       }
     }
-
   }
 
   private Optional<SessionInformation> getSessionInformation(String ackId) {
     return ofNullable(pendingSessions.remove(ackId));
   }
 
+  public void unlockAll() {
```
estebanwasinger: This should only unlock the ones from the same connection

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsListener.java
```diff
@@ -318,9 +268,25 @@ private boolean replyDestinationIsTopic(Destination destination) {
...
+    return synchronous ? new DefaultJmsListenerLock() : new NullJmsListenerLock();
+  }
+
+  private int getValidNumberOfConsumers(int numberOfConsumers) {
+    if (numberOfConsumers > 1 && consumerType.isTopic()) {
+      TopicConsumer topicConsumer = (TopicConsumer) consumerType;
+
+      if (!isCapableOfMultiConsumersOnTopic(topicConsumer)) {
+        LOGGER.warn("Destination [" + destination + "] is a topic, but " + numberOfConsumers
+            + " receivers have been requested. Will configure only 1");
```
alegmarra: let's also hint that this is only possible if the topic consumer is Shared:
"Destination [" + destination + "] is a topic, but " + numberOfConsumers  + " receivers have been requested. This is only possible for 'shared' topic consumers. Only one will be configured instead."

extensions/jms/src/main/java/org/mule/extensions/jms/api/operation/JmsPublishConsume.java
```diff
@@ -71,38 +73,39 @@
...
    * @throws JmsExtensionException if an error occurs
    */
   @OutputResolver(output = JmsOutputResolver.class)
   @Throws(JmsPublishConsumeErrorTypeProvider.class)
-  public Result<Object, JmsAttributes> publishConsume(@UseConfig JmsConfig config, @Connection JmsConnection connection,
-                                                      @Placement(order = 0) @XmlHints(
-                                                          allowReferences = false) @Summary("The name of the Destination where the Message should be sent") String destination,
+  public Result<Object, JmsAttributes> publishConsume(@UseConfig JmsConfig config,
+                                                      @Connection JmsConnection connection,
+                                                      @Placement(order = 0) @XmlHints(allowReferences = false) String destination,
```
marianogonzalez: a string parameter doesn't accept references to begin with

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsListener.java
```diff
@@ -190,47 +202,58 @@ public void onStart(SourceCallback<Object, JmsAttributes> sourceCallback) throws
...
-      LOGGER.debug("Stopping JMSSubscriber source");
+      LOGGER.debug("Stopping JMS Message Listener");
     }
 
-    jmsLock.unlock();
-
-    try {
-      if (ackMode.equals(AUTO) || ackMode.equals(DUPS_OK) || ackMode.equals(MANUAL)) {
-        session.get().recover();
+    if (synchronous) {
```
marianogonzalez: why do you still need this synchronous thing? Isn't this achieved by simply setting only one consumer?

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsListener.java
```diff
@@ -318,9 +288,25 @@ private boolean replyDestinationIsTopic(Destination destination) {
...
+  private JmsListenerLock createJmsLock() {
+    return synchronous ? new DefaultJmsListenerLock() : new NullJmsListenerLock();
+  }
+
+  private int getValidNumberOfConsumers(int numberOfConsumers) {
+    if (numberOfConsumers > 1 && consumerType.isTopic()) {
+      TopicConsumer topicConsumer = (TopicConsumer) consumerType;
+
+      if (!isCapableOfMultiConsumersOnTopic(topicConsumer)) {
+        LOGGER.warn("Destination [" + destination + "] is a topic, but " + numberOfConsumers
```
marianogonzalez: no extension should ever decide to ignore a user setting. If this is invalid, then throw an exception. Or even better, make it so that it's not possible to have and invalid config. Isn't it possible to have the destination as a sub type so that when it's topic there's no option to set consumers?

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsMessageListener.java
```diff
@@ -0,0 +1,148 @@
...
+    produceMessageResult(message, context, resolveEncoding(message), resolveContentType(message));
+    waitForMessageToBeProcessed(jmsLock);
+  }
+
+  private void evaluateAckAction(Message message) {
+    try {
+      evaluateMessageAck(ackMode, session, message, sessionManager, jmsLock);
+    } catch (JMSException e) {
+      LOGGER.error("An error occurred while processing an incoming message: ", e);
+      sourceCallback.onSourceException(e);
```
marianogonzalez: this isn't a source exception.

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsMessageListener.java
```diff
@@ -0,0 +1,148 @@
...
+        context.addVariable(REPLY_TO_DESTINATION_VAR, replyTo);
+      }
+    } catch (JMSException e) {
+      LOGGER.error("An error occurred while obtaining the ReplyTo destination: ", e);
+      sourceCallback
+          .onSourceException(new JmsExtensionException(e, "An error occurred while obtaining the ReplyTo destination: "));
+    }
+  }
+
+  private void produceMessageResult(Message message, SourceCallbackContext context, String encoding, String contentType) {
```
marianogonzalez: non descriptive name. Use something like dispatchMEssage or anything which actually gives the impression that a mule message is being pushed to the pipeline

extensions/jms/src/test/java/org/mule/extensions/jms/test/multiconsumer/AbstractJmsMultiConsumerTestCase.java
```diff
@@ -0,0 +1,65 @@
...
+    flowRunner("recoverSession").withPayload(ackId).run();
+  }
+
+  void publishTo(int amount, String destination, DestinationType destinationType) throws Exception {
+    for (int i = 0; i < amount; i++) {
+      publish("message", destination, ImmutableMap.of("destinationType", destinationType));
+    }
+  }
+
+  <T, S> Stream<T> aFor(S initialState, Function<S, Boolean> condition, Function<S, S> updater, Function<S, T> supplier) {
```
marianogonzalez: condition should be a Predicate

marianogonzalez: unclear method name, with unclear parameter names. Improve that + jdoc

extensions/jms/src/test/java/org/mule/extensions/jms/test/multiconsumer/Jms1xMultiConsumerTestCase.java
```diff
@@ -0,0 +1,94 @@
...
+  protected String[] getConfigFiles() {
+    return new String[] {"multiconsumer/jms-1.x-multi-consumers.xml", "config/activemq/activemq-default.xml"};
+  }
+
+  @Test
+  public void multiConsumersConsumeMessagesInParallel() throws Exception {
+    publishTo(NUMBER_OF_MESSAGES, destination.getValue(), QUEUE);
+
+    long distinctAckIds = aFor(0, i -> i < NUMBER_OF_MESSAGES, i -> ++i, i -> pollMessage())
+        .map(Result::getAttributes)
```
marianogonzalez: Abuse

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsListener.java
estebanwasinger: > Isn't it possible to have the destination as a subtype so that when it's topic there's no option to set consumers?
No, because in JMS 2 with a shared connection is possible to have multiple consumers for topics

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsMessageListener.java
estebanwasinger: Why is this not a source exception? I'm communicating that a error occurred trying to process an incoming message.

extensions/jms/src/main/java/org/mule/extensions/jms/api/source/JmsListener.java
```diff
@@ -153,7 +153,7 @@
    * The number of concurrent consumers that will be used to receive JMS Messages
    */
   @Parameter
-  @Optional(defaultValue = "4")
+  @Optional(defaultValue = "1")
```
marianogonzalez: 4. And put a todo to fix the API

PR #4464, opened 2017-04-07
MULE-11866: Add callbacks before and after authorization-code authorize

elrodro83:


modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/state/DefaultResourceOwnerOAuthContext.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+package org.mule.runtime.oauth.api.state;
+
+import java.io.Serializable;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.locks.Lock;
+
+/**
+ * OAuth state for a particular resource owner which typically represents an user.
```
marianogonzalez: `@since`

modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/builder/OAuthAuthorizationCodeDancerBuilder.java
```diff
@@ -115,4 +120,26 @@
...
+   * Allows custom code to be run just before doing the request to the provided {@code tokenUrl}.
+   * <p>
+   * The map returned by the provided function will be passed to the {@link #afterDanceCallback(BiConsumer)}, if set.
+   * 
+   * @param callback a {@link Function} that receives the parameters that will be used in the executing dance and returns a map to
+   *        be then passed to the {@link #afterDanceCallback(BiConsumer)}. Not null.
+   * 
+   * @return this builder
+   */
+  OAuthAuthorizationCodeDancerBuilder beforeDanceCallback(Function<AuthorizationCodeRequest, Map<String, TypedValue>> callback);
```
marianogonzalez: this is very weird. Why would I return a map that is passed to the after but is ignored by all other components? Doesn't this overlap with the state parameter?

What's the use case for this? Please clarify because otherwise, it seems like Function is a too generic contract for whatever is it you're trying to achieve here.

modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/state/DefaultResourceOwnerOAuthContext.java
elrodro83: done

modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/builder/AuthorizationCodeDanceCallbackContext.java
```diff
@@ -0,0 +1,19 @@
...
+
+import org.mule.runtime.oauth.api.AuthorizationCodeOAuthDancer;
+
+/**
+ * Implementations keep context information between the before and after callbacks set for an
+ * {@link AuthorizationCodeOAuthDancer}.
+ * 
+ * @since 4.0
+ */
+public interface AuthorizationCodeDanceCallbackContext {
```
marianogonzalez: this interface is worthless as it is. It should have an `Optional<Object> getParameter(String)` or something of that sort.

elrodro83: added

PR #4465, opened 2017-04-10
MULE-12075 Serialization issues in EventContext

dfeist:


core-tests/src/test/java/org/mule/runtime/core/DefaultEventContextTestCase.java
```diff
@@ -421,6 +421,57 @@ public void externalCompletionWithChild() throws Exception {
     assertCompletionDone(parent);
   }
 
+  @Test
+  @Description("When a child event context is serialized the parent context no longer waits for completion of child context.")
```
marianogonzalez: This is wrong. Just because the event is serialized in  MP-N, it doesn't mean that MP-N+1 will continue processing it and the parent does need to wait for its completion.

core-tests/src/test/java/org/mule/runtime/core/DefaultEventContextTestCase.java
```diff
@@ -421,6 +421,57 @@ public void externalCompletionWithChild() throws Exception {
...
+    Event event = testEvent();
+
+    parent.success(event);
+
+    assertResponse(parent, event);
+    assertCompletionDone(parent);
+  }
+
+  @Test
+  @Description("When a child event context is de-serialized it completion works in the same way.")
```
marianogonzalez: `in the same way` is not clear

core/src/main/java/org/mule/runtime/core/DefaultEventContext.java
```diff
@@ -225,6 +227,15 @@ public String toString() {
     public Optional<EventContext> getParentContext() {
       return of(parent);
     }
+
+    private void writeObject(ObjectOutputStream out) throws IOException {
+      // Un-register from parent when child is serialized
+      if (parent instanceof AbstractEventContext) {
+        ((AbstractEventContext) parent).removeChildContext(this);
```
marianogonzalez: i think this is wrong

PR #4466, opened 2017-04-10
MULE-12164: Extensions documenter doesn't document correctly Message …

estebanwasinger:
…Sources


estebanwasinger:
https://github.com/mulesoft/mule-extensions-maven-plugin/pull/60

PR #4470, opened 2017-04-10
MULE-12051: All functional tests from CE must use the isolated test

pablokraan:
_ Migrating ExtensionFunctionalTestCase tests to use AbstractExtensionFunctionalTestCase
_ Fixing tests that compare generated resources after changes in the extensions
_ Adding plugin classifier for test extensions
_ Removing mule-module.properties from test extensions
_ Extending pet store extension with failing source
_ Extending pet store extension with failing operation
_ Expanding pet store extension to include more operations and providers
_ Externalizing parameter resolver extension
_ Externalizing implicit config extension
_ Externalizing some extension
_ Externalizing type value extension
_ Externalizing multi implicit config extension
_ Updating implicit config extension to work on another test
_ Externalizing implicit exclusive config extension


modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/AbstractExtensionFunctionalTestCase.java
```diff
@@ -0,0 +1,38 @@
...
+
+@ArtifactClassLoaderRunnerConfig(
+    plugins = {"org.mule.tests:mule-heisenberg-extension", "org.mule.tests:mule-vegan-extension",
+        "org.mule.tests:mule-petstore-extension", "org.mule.tests:mule-metadata-extension",
+        "org.mule.tests:mule-marvel-extension", "org.mule.tests:mule-subtypes-extension",
+        "org.mule.tests:mule-tx-extension", "org.mule.tests:mule-some-extension",
+        "org.mule.tests:mule-implicit-config-extension", "org.mule.tests:mule-typed-value-extension",
+        "org.mule.tests:mule-parameter-resolver-extension", "org.mule.tests:mule-implicit-exclusive-config-extension",
+        "org.mule.tests:mule-multi-implicit-config-extension"},
+    testExclusions = {"org.mule.tests:mule-heisenberg-extension:*:*:*", "org.mule.tests:mule-vegan-extension:*:*:*",
```
gsfernandes: Just to confirm, are these artifacts coming as transitive dependency from a test artifact?
The only cases where we needed to exclude things using this method was related to dependencies declared from mule-tests-functional artifact to mule artifacts and not declared as provided. For instance, mule-tests-functional depends on mule-core as compile therefore we needed to exclude that dependency when resolving test scope dependencies.

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/metadata/SourceMetadataTestCase.java
```diff
@@ -32,7 +31,7 @@
 
   public SourceMetadataTestCase(ResolutionType resolutionType) {
     super(resolutionType);
-    this.provider = resolutionType == EXPLICIT_RESOLUTION ? MetadataService::getSourceMetadata
+    this.provider = resolutionType == ResolutionType.EXPLICIT_RESOLUTION ? MetadataService::getSourceMetadata
```
gsfernandes: static import

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/AbstractExtensionFunctionalTestCase.java
pablokraan: I had to add this to make it work, I can try removing it and changing the scope of these dependencies to provided

PR #4479, opened 2017-04-11
MULE-12183: AbstractAsyncRequestReplyRequester should not add correlation sequence to correlationID

mbuchwald:


core/src/test/java/org/mule/routing/requestreply/AsyncRequestReplyRequesterTestCase.java
```diff
@@ -268,6 +268,47 @@ public void release()
...
+
+        asyncMP.setListener(target);
+        asyncReplyMP.setListener(asyncMP);
+        asyncReplyMP.setReplySource(target.getMessageSource());
+
+        MuleEvent event = getTestEvent(TEST_MESSAGE, getTestService(),
+                getTestInboundEndpoint(MessageExchangePattern.ONE_WAY));
+        event.getMessage().setCorrelationSequence(3);
+
+        asyncReplyMP.process(event);
```
gsfernandes: I don't see any asserts here, shouldn't you validate something here after calling process?

PR #4480, opened 2017-04-11
MULE-10291: suport <operation>'s parametrization in Mule Event

fernandezlautaro:


core/src/main/java/org/mule/runtime/core/api/Event.java
```diff
@@ -82,6 +83,20 @@
   Set<String> getVariableNames();
 
   /**
+   * Returns the complete set of properties in the current event
```
pablolagreca: Please add more info about what properties are used for

core/src/main/java/org/mule/runtime/core/api/Event.java
```diff
@@ -82,6 +83,20 @@
...
+   * @return a {@link Map} containing the properties' name as key and the {@link TypedValue} as value
+   */
+  Map<String, TypedValue<Object>> getProperties();
+
+  /**
+   * Returns the complete set of parameters in the current event
+   *
+   * @return a {@link Map} containing the parameters' name as key and the {@link TypedValue} as value
+   */
+  Map<String, TypedValue<Object>> getParameters();
```
pablolagreca: same here.

core/src/main/java/org/mule/runtime/core/el/DataWeaveExpressionLanguageAdaptor.java
```diff
@@ -53,6 +50,8 @@
   public static final String ATTRIBUTES = "attributes";
   public static final String ERROR = "error";
   public static final String VARIABLES = "variables";
+  public static final String PROPERTY = "property";
+  public static final String PARAM = "param";
```
pablolagreca: let's use parameters since we don't have any other binding without the complete name

fernandezlautaro: if we go with `parameters` rather than `param` (which is consistent with `variables`), then instead of `property` it should be `properties` as well.

Could you confirm?

PR #4487, opened 2017-04-12
MULE-11871: Add a method in the dancer to get the current state for a resourceOwner

elrodro83:
* Remove unneeded polimorphism for dancers

### DIFF
--- a/services/oauth/src/main/java/org/mule/services/oauth/internal/AbstractOAuthDancer.java
+++ b/services/oauth/src/main/java/org/mule/services/oauth/internal/AbstractOAuthDancer.java
@@ -6,9 +6,6 @@
  */
 package org.mule.services.oauth.internal;
 
-import static java.lang.String.format;
-import static java.util.concurrent.CompletableFuture.completedFuture;
-import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
 import static org.mule.runtime.api.metadata.MediaType.ANY;
 import static org.mule.runtime.api.metadata.MediaType.parse;
 import static org.mule.service.http.api.HttpConstants.HttpStatus.BAD_REQUEST;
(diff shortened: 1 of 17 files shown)


services/oauth/src/main/java/org/mule/services/oauth/internal/AbstractOAuthDancer.java
```diff
@@ -238,7 +223,7 @@ protected TokenResponse invokeTokenUrl(String tokenUrl, Map<String, String> toke
    * @param resourceOwnerId id of the user.
    * @return oauth state
    */
-  protected DefaultResourceOwnerOAuthContext getContextForResourceOwner(final String resourceOwnerId) {
+  public ResourceOwnerOAuthContext getContextForResourceOwner(final String resourceOwnerId) {
```
afelisatti: Why is this in the abstract class if only authorisation code needs it according to it's api?

elrodro83: client credentials reuses this, putting its only context witha constant as a key, here: https://github.com/mulesoft/mule/pull/4487/files#diff-88a5f3c13546df357111856a8829052cR138

When using just the api interfaces, this is not exposef for client credentials

services/oauth/src/test/java/org/mule/test/oauth/internal/OAuthContextTestcase.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.Map;
+import java.util.concurrent.TimeoutException;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import ru.yandex.qatools.allure.annotations.Features;
+
+@Features("OAuth Service")
+public class OAuthContextTestcase extends AbstractMuleContextTestCase {
```
afelisatti: TestCase

PR #4490, opened 2017-04-12
MULE-11868: Add invalidate() method in OAuthDancer

afelisatti:


modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/AuthorizationCodeOAuthDancer.java
```diff
@@ -42,6 +42,14 @@
   CompletableFuture<Void> refreshToken(String resourceOwner);
 
   /**
+   * Clears the oauth context for a given user.
+   *
+   * @param resourceOwnerId id of the user.
+   * @return a completable future that is complete when the context has been cleared.
+   */
+  CompletableFuture<Void> invalidateContext(String resourceOwnerId);
```
elrodro83: you don need to return a Completable Future here. In the other methods it is there to support non-blocking http calls eventually.

PR #4491, opened 2017-04-12
MULE-10833 - Make ComponentDefinitionBuildingProviders independent fr…

pablolagreca:
…om MuleContext


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/ReflectiveMethodComponentExecutor.java
```diff
@@ -75,7 +75,12 @@ public Object execute(ExecutionContext<M> executionContext) throws Exception {
   }
 
   private Object[] getParameterValues(ExecutionContext<M> executionContext, Class<?>[] parameterTypes) {
-    return argumentResolverDelegate.resolve(executionContext, parameterTypes);
+    // try {
```
marianogonzalez: cleanup

PR #4498, opened 2017-04-17
MULE-12121: Replace XMLGenerationService from tooling client to Artif…

alegmarra:
…actSerializationService


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/resources/MuleExtensionModelProvider.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Utility class to access the {@link ExtensionModel} definition for Mule's Runtime
+ *
+ * @since 4.0
+ */
+public class MuleExtensionModelProvider {
+
+  private static final String MODEL_JSON = "META-INF/mule-extension-model.json";
+  private static final ExtensionModel EXTENSION_MODEL = new ExtensionModelJsonSerializer(false)
```
marianogonzalez: although we should have test coverage to ensure that a deserialized model is usable, we shouldn't internally obtain it this way. We will most likely want those models to have model properties and other herbs  which don't make it to the serialized model

alegmarra: Agree in that is no ideal, but this change is better than what we had and should be more than enough since this is only used by tooling, which already works with all deserialized plugins.
It sounds simple to provide this model without using the json, it implies creating the whole declaration manually, going through the factory, etc.
Not using the Json sounds like a different ticket to me. We can always change later how we create the model.

PR #4499, opened 2017-04-17
MULE-12119: Add the capability to inject the Transactional Action

estebanwasinger:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/InfrastructureTypeMapping.java
```diff
@@ -60,11 +66,11 @@
   }
 
   public static Optional<QNameModelProperty> getQName(String name) {
-    return Optional.of(QNAMES.get(name));
+    return Optional.ofNullable(QNAMES.get(name));
```
marianogonzalez: this seems like hiding someething.. @alegmarra thoughts?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/InfrastructureTypeMapping.java
```diff
@@ -60,11 +66,11 @@
...
   }
 
   public static Optional<ParameterDslConfiguration> getDslConfiguration(String name) {
-    return Optional.of(DSL_CONFIGURATIONS.get(name));
+    return Optional.ofNullable(DSL_CONFIGURATIONS.get(name));
```
marianogonzalez: idem.. 👈 @alegmarra

PR #4505, opened 2017-04-17
MULE-12232: Introduce the concept of FlowListener

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultFlowListener.java
```diff
@@ -0,0 +1,105 @@
...
+    onComplete = handler;
+  }
+
+  private void onTerminate(Event event, Throwable error) {
+    try {
+      if (event != null && successConsumer != null) {
+        try {
+          successConsumer.accept(event.getMessage());
+        } catch (Exception e) {
+          LOGGER.warn("Exception found while execution onSuccess FlowListener", e);
```
alegmarra: we could receive the name and namespace of the component that receives this instance of FlowListener in order to improve these error messages. Think of the logs when multiple components fail for the same flow termination.

PR #4506, opened 2017-04-17
MULE-12218: Reduce overhead of debug log statements by using isDebugEnabled

afelisatti:


afelisatti:
There are no more logger uses here nor in Message or Event.

PR #4507, opened 2017-04-18
MULE-12216 Eliminate thread contention due to use of java.lang.ClassL…

dfeist:
…oader#getSystemClassLoader in ArtifactAwareContextSelector

### DIFF
--- a/modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/ArtifactAwareContextSelector.java
+++ b/modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/ArtifactAwareContextSelector.java
@@ -49,7 +49,7 @@ import org.apache.logging.log4j.status.StatusLogger;
  */
 class ArtifactAwareContextSelector implements ContextSelector, Disposable {
 
-  static final StatusLogger STATUS_LOGGER = StatusLogger.getLogger();
+  static final StatusLogger LOGGER = StatusLogger.getLogger();
   private static final ClassLoader SYSTEM_CLASSLOADER = getSystemClassLoader();
 
   private LoggerContextCache cache = new LoggerContextCache(this, getClass().getClassLoader());
--- a/modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/ArtifactAwareContextSelector.java
+++ b/modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/ArtifactAwareContextSelector.java
@@ -6,6 +6,8 @@
  */
 package org.mule.runtime.module.launcher.log4j2;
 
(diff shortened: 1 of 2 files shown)


modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/ArtifactAwareContextSelector.java
```diff
@@ -47,7 +49,8 @@
  */
 class ArtifactAwareContextSelector implements ContextSelector, Disposable {
 
-  static final StatusLogger logger = StatusLogger.getLogger();
+  static final StatusLogger STATUS_LOGGER = StatusLogger.getLogger();
```
pablokraan: Can you use anothe name instead of STATUS_LOGGER?

pablokraan: If I'm correct, these changes are required on mule 3 too. Can you check that?

modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/MuleLoggerContextFactory.java
```diff
@@ -85,22 +81,21 @@ private URI getArtifactLoggingConfig(ArtifactClassLoader muleCL) {
...
+                         e.getClass().getName(), e.getMessage());
       appLogConfig = getLogConfig(muleCL);
     }
 
-    if (appLogConfig != null && ArtifactAwareContextSelector.logger.isInfoEnabled()) {
-      ArtifactAwareContextSelector.logger.info("Found logging config for application '{}' at '{}'", muleCL.getArtifactId(),
-                                               appLogConfig);
+    if (appLogConfig != null && STATUS_LOGGER.isInfoEnabled()) {
+      STATUS_LOGGER.info("Found logging config for application '{}' at '{}'", muleCL.getArtifactId(),
+                         appLogConfig);
```
pablokraan: these lines can be merge in a single one

modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/ArtifactAwareContextSelector.java
dfeist: Any suggestion? It was simply "logger" before..  I'm not 100% clear on what role of this static logger is.

pablokraan: You used status logger because that is the name of the class that owns it... but that is not clear. I prefer just logger as before, otherwise use a name that describe why it must be used from outside this class

PR #4510, opened 2017-04-18
MULE-10646: Add performance validation fixes to HTTP streaming solution.

afelisatti:


modules/http/src/main/java/org/mule/module/http/internal/request/DefaultHttpRequesterConfig.java
```diff
@@ -42,6 +42,7 @@
     private static final int UNLIMITED_CONNECTIONS = -1;
     private static final int DEFAULT_CONNECTION_IDLE_TIMEOUT = 30 * 1000;
     private static final String THREAD_NAME_PREFIX_PATTERN = "%shttp.requester.%s";
+    private static final int DEFAULT_RESPONSE_BUFFER_SIZE = 10240;
```
dfeist: this is slightly clearer if you do `=10 * 1024`

modules/http/src/main/java/org/mule/module/http/internal/request/grizzly/GrizzlyHttpClient.java
```diff
@@ -627,7 +622,7 @@ public InputStream sendAndReceiveInputStream(HttpRequest request, int responseTi
     {
         Request grizzlyRequest = createGrizzlyRequest(request, responseTimeout, followRedirects, authentication);
         PipedOutputStream outPipe = new PipedOutputStream();
-        PipedInputStream inPipe = new PipedInputStream(outPipe);
+        PipedInputStream inPipe = new PipedInputStream(outPipe, responseBufferSize);
```
dfeist: Probably fine, but not clear to me why PipedInputStream is created in 3 different places.

modules/http/src/test/java/org/mule/module/http/internal/request/grizzly/FlowWorkManagerIOStrategyTestCase.java
```diff
@@ -1,133 +0,0 @@
...
-import org.junit.rules.ExpectedException;
-import org.junit.runner.RunWith;
-import org.mockito.Answers;
-import org.mockito.Mock;
-import org.mockito.Mockito;
-import org.mockito.runners.MockitoJUnitRunner;
-
-@RunWith(MockitoJUnitRunner.class)
-@SmallTest
-public class FlowWorkManagerIOStrategyTestCase extends AbstractMuleTestCase
```
dfeist: Update test case rather than delete.

modules/http/src/main/java/org/mule/module/http/internal/request/grizzly/IOStrategyTransportCustomizer.java
```diff
@@ -40,7 +40,7 @@ public IOStrategyTransportCustomizer(String threadNamePrefix, Integer maxWorkerP
     @Override
     public void customize(TCPNIOTransport transport, FilterChainBuilder filterChainBuilder)
     {
-        transport.setIOStrategy(FlowWorkManagerIOStrategy.getInstance());
+        transport.setIOStrategy(WorkerThreadIOStrategy.getInstance());
```
dfeist: - Where is the size of this pool configured, whats default?
- We should still hand off to Flow pool at some point.

modules/http/src/main/java/org/mule/module/http/internal/request/grizzly/GrizzlyHttpClient.java
afelisatti: Because it's 3 different send methods, blocking, non blocking and some stuff for ws consumer.

modules/http/src/main/java/org/mule/module/http/internal/request/grizzly/IOStrategyTransportCustomizer.java
afelisatti: The pool in configured right under this line using Grizzly defaults, just like selectors. They do twice the cores count. That hand off is within the AsyncHandler for the NB send() and not necessary for the other methods because they are blocking.

modules/http/src/main/java/org/mule/module/http/internal/request/grizzly/GrizzlyHttpClient.java
dfeist: ok

modules/http/src/main/java/org/mule/module/http/internal/request/grizzly/IOStrategyTransportCustomizer.java
dfeist: ok

PR #4511, opened 2017-04-18
MULE-10646: Add performance validation fixes to HTTP streaming solution.

afelisatti:


extensions/http/src/main/java/org/mule/extension/http/internal/request/RequestConnectionParams.java
```diff
@@ -25,6 +25,9 @@
  */
 public final class RequestConnectionParams {
 
+  public static final String DEFAULT_MAX_CONNECTIONS = "-1";
+  public static final String DEFAULT_CONNECTION_IDLE_TIMEOUT = "30000";
+  public static final String DEFAULT_RESPONSE_BUFFER_SIZE = "10240";
```
dfeist: Easier to read with `10 * 1024`

extensions/http/src/main/java/org/mule/extension/http/internal/request/RequestConnectionParams.java
```diff
@@ -116,6 +128,10 @@ public Integer getConnectionIdleTimeout() {
     return connectionIdleTimeout;
   }
 
+  public Integer getResponseBufferSize() {
```
dfeist: Can't `int` be used?

extensions/http/src/main/java/org/mule/extension/http/internal/request/RequestConnectionParams.java
```diff
@@ -144,6 +160,10 @@ public void setConnectionIdleTimeout(Integer connectionIdleTimeout) {
     this.connectionIdleTimeout = connectionIdleTimeout;
   }
 
+  public void setResponseBufferSize(Integer responseBufferSize) {
```
dfeist: Even if `int` can't be used for getter, surely it can be used for setter?

modules/http-api/src/main/java/org/mule/service/http/api/client/HttpClientConfiguration.java
```diff
@@ -84,6 +91,7 @@ public String getOwnerName() {
     private int maxConnections = -1;
     private boolean usePersistentConnections = true;
     private int connectionIdleTimeout = 30000;
+    private int responseBufferSize = 10240;
```
dfeist: Why is default defined twice?

afelisatti: one is the service default and another the extension's, they could potentially be different

extensions/http/src/main/java/org/mule/extension/http/internal/request/RequestConnectionParams.java
afelisatti: I'll check, I think it's just following the other fields but we may be able to change it.

modules/http-api/src/main/java/org/mule/service/http/api/client/HttpClientConfiguration.java
dfeist: mmm, can this use constants e.g. `DEFAULT_RESPONSE_BUFFER_SIZE` as defaults instead of literals, or are they not visible?

extensions/http/src/main/java/org/mule/extension/http/internal/request/RequestConnectionParams.java
afelisatti: This is a String though...

afelisatti:
jenkins validate

afelisatti:
jenkins validate

PR #4512, opened 2017-04-18
MULE-12040: Don't use application log after the application is undepl…

facundovs:
…oyed.

The DefaultResourceReleaser's logger gets resolved to the application's logger, but it's called from AbstractArtifactClassLoader#dispose after the logging context is removed.
So a new MuleLoggerContext is created, locking the application's log file in Windows and keeping alive the JMS connection if logging through a JMSAppender. This stays that way after the application is undeployed.
To fix this we should either change the order of the logging context disposing.
In addition,we should also change the keys that are used in LoggerContextCache, because two differents instances of the same classloader are generating two LoggerContexts (HashCodes of classloaders are being used as keys). If we use the name of the application as key, we can guarantee only a loggerContext per application be created.
This fix is necessary but not enough to give a solution to this bug.
I have found a bug in log4j side, which is causing that some FileDescriptors aren't being closed, and I'll create a PR to fix it.

### DIFF
--- a/modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
+++ b/modules/launcher/src/main/java/org/mule/module/launcher/artifact/AbstractArtifactClassLoader.java
@@ -61,6 +61,22 @@ public abstract class AbstractArtifactClassLoader extends FineGrainedControlClas
     @Override
     public void dispose()
     {
+
+        try
+        {
+            createResourceReleaserInstance().release();
+        }
+        catch (Exception e)
+        {
+            logger.error(e);
+        }
+
+        super.dispose();
+        shutdownListeners();
+    }
+
+    private void shutdownListeners ()
+    {
         for (ShutdownListener listener : shutdownListeners)
         {
             try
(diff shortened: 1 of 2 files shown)


pablokraan:
"In addition,we should also change the keys that are used in LoggerContextCache, because two instances of the same classloader are generating two LoggerContexts (HashCodes of classloaders are being used as keys). If we use the name of the application as key, we can guarantee only a loggerContext per application be created."
I don't understand how, if the key is the classloader's hashcode, that two logger contexts are generated for the same instance. Can you  explain that?

pablokraan:
"I have found a bug in log4j side, which is causing that some FileDescriptors aren't being closed, and I'll create a PR to fix it."
Check that the issue you mentioned is not already fixed on the latest log4j version, IIRC, there were some leaks fixed on 2.7

modules/launcher/src/main/java/org/mule/module/launcher/log4j2/LoggerContextCache.java
```diff
@@ -255,12 +257,16 @@ public void run()
         }
     }
 
-    private int computeKey(ClassLoader classLoader)
+    private String computeKey(ClassLoader classLoader)
     {
-        return classLoader.hashCode();
+        if (classLoader instanceof ApplicationClassLoader)
```
pablokraan: What happens when an application is redeployed? and what happens with domains or any other type or artifacT?

facundovs:
@pablokraan, Regarding to your first comment, I mean two different instances of the same classloader.
Regarding to the log4j version, I've checked it in the last version, and It hasn't fixed yet.

facundovs:
@pablokraan, I've removed the changes in the cache keys.

PR #4516, opened 2017-04-18
Mule 11526

pablolagreca:


tests/integration/src/test/java/org/mule/test/integration/interception/ProcessorInterceptorFactoryTestCase.java
```diff
@@ -86,7 +102,7 @@ public ProcessorInterceptor get() {
 
   public static class HasInjectedAttributesInterceptor implements ProcessorInterceptor {
 
-    public static final AtomicBoolean intercepted = new AtomicBoolean(false);
+    private static final LinkedList<InterceptionParameters> interceptionParameters = new LinkedList<>();
```
elrodro83: change declaration to just `List`. Only `addLast` is being used, can be changed to just `add`

PR #4517, opened 2017-04-18
MULE-12239 When objects form a plugin are being registered, the context classloader is the one from the app instead of the plugin

elrodro83:


elrodro83:
jenkins validate

elrodro83:
jenkins validate

modules/spring-config/src/main/java/org/mule/runtime/config/spring/SpringXmlConfigurationBuilder.java
```diff
@@ -141,12 +146,21 @@ private MuleArtifactContext createApplicationContext(MuleContext muleContext) th
   protected MuleArtifactContext doCreateApplicationContext(MuleContext muleContext, ConfigResource[] artifactConfigResources,
                                                            ArtifactDeclaration artifactDeclaration,
                                                            OptionalObjectsController optionalObjectsController) {
+    List<ClassLoader> artifactPluginClassLoaders;
```
pablolagreca: This logic seems duplicate. Can you move it to an utility method?

PR #4525, opened 2017-04-19
MULE-10355: ExtensionModel generation under deployment module is ordered

fernandezlautaro:


modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/DeploymentServiceTestCase.java
```diff
@@ -1939,7 +1960,7 @@ public void deploysMultiPluginVersionLib() throws Exception {
   }
 
   @Test
-  public void deploysApplicationWithPluginDependingOnPlugin() throws Exception {
+  public void deploysApplicationWithPluginDependingOnPlugin() throws Exception { //TODO lautaro este puede ser uno
```
pablolagreca: ???

fernandezlautaro: jaja, la cagué.

PR #4529, opened 2017-04-19
MULE-12239: When objects form a plugin are being registered, the context

elrodro83:
classloader is the one from the app instead of the plugin
* Fix test runner to properly build an app classloader


elrodro83:
DO NOT MERGE YET!! BUILD PENDING!!

PR #4532, opened 2017-04-20
Wrong comparison in QueueConfiguration.equals

canoasan:


core/src/main/java/org/mule/util/queue/objectstore/QueueConfiguration.java
```diff
@@ -59,15 +59,7 @@ public boolean equals(Object obj)
...
-        if (capacity != other.capacity)
-        {
-            return false;
-        }
-        if (!objectStore.equals(objectStore))
-        {
-            return false;
-        }
-        return true;
+        return capacity == other.capacity && objectStore.equals(other);
```
elrodro83: `objectStore.equals(other.objectStore)`

PR #4534, opened 2017-04-20
MULE-12268: Add test to verify the reception of multiple headers in the HTTP request

afelisatti:


modules/http/src/test/java/org/mule/module/http/functional/requester/HttpRequestMultipleValueHeadersTestCase.java
```diff
@@ -36,11 +39,13 @@
     public SystemProperty host = new SystemProperty("host", "localhost");
     @Rule
     public SystemProperty encoding = new SystemProperty("encoding" , CHUNKED);
+    @Rule
+    public DynamicPort port = new DynamicPort("port");
 
     @Override
     protected String getConfigFile()
     {
-        return "http-request-headers-config.xml";
```
elrodro83: is this file still being used elsewhere? if not delete it

afelisatti: Yes, this test was actually reusing the config file of the general org.mule.module.http.functional.requester.HttpRequestHeadersTestCase so a created a specific one for it.

PR #4539, opened 2017-04-21
MULE-12271 Refactor use of EventContext to simplify

dfeist:


core/src/main/java/org/mule/runtime/core/api/processor/MessageProcessors.java
```diff
@@ -96,7 +96,17 @@ public static MessageProcessorChain newExplicitChain(List<Processor> processors)
...
-      return just(event).transform(processor).otherwise(Exceptions.EventDroppedException.class, ede -> empty()).block();
+      return just(event).transform(processor).otherwise(EventDroppedException.class, mde -> empty()).block();
+    } catch (Throwable e) {
+      throw rxExceptionToMuleException(e);
+    }
+  }
+
+  public static Event processToApply(Event event, Flow flow) throws MuleException {
+    try {
+      just(event).transform(flow).doOnError(throwable -> {
```
elrodro83: isn't this just ignoring the error?

core/src/main/java/org/mule/runtime/core/processor/strategy/StreamPerEventSink.java
```diff
@@ -39,7 +39,7 @@ public StreamPerEventSink(Function<Publisher<Event>, Publisher<Event>> processor
 
   @Override
   public void accept(Event event) {
-    eventConsumer.accept(event);
-    just(event).transform(processor).subscribe();
+    just(event).doOnNext(request -> eventConsumer.accept(request)).transform(processor).doOnError(throwable -> {
```
elrodro83: why is the doOnError lambda empty?

modules/spring-config/src/test/java/org/mule/runtime/config/spring/factories/FlowRefFactoryBeanTestCase.java
```diff
@@ -126,12 +127,12 @@ public void dynamicFlowRefSubFlowConstructAware() throws Exception {
     FlowConstruct flowConstruct = mock(FlowConstruct.class);
     Event event = testEvent();
     FlowConstructAware targetSubFlowConstructAware = mock(FlowConstructAware.class, INITIALIZABLE_MESSAGE_PROCESSOR);
-    when(((Processor) targetSubFlowConstructAware).process(event)).thenReturn(result);
+    when(((Processor) targetSubFlowConstructAware).process(any(Event.class))).thenReturn(result);
```
elrodro83: so we are always creating a new event, then?

core/src/main/java/org/mule/runtime/core/api/processor/MessageProcessors.java
dfeist: Error is handled in pipeline and event callback is called there. This is needed because reactor fails when there is an error and no onError. Will explain in comment, also see if all of them are really needed.

modules/spring-config/src/test/java/org/mule/runtime/config/spring/factories/FlowRefFactoryBeanTestCase.java
dfeist: Yes, because each iteration/request needs a new eventContext else responses get mixed up.

PR #4547, opened 2017-04-21
MULE-11083: Classloading: implement classLoader LoaderDescriptor for …

gsfernandes:
…plugins in the mule-plugin.json - Removed plugin.properties support


pablokraan:
Changes are good, but is not clear to me why the commit message refers to a closed jira. Is that OK?

gsfernandes:
Created https://github.com/mulesoft/mule/pull/4553

PR #4548, opened 2017-04-22
MULE-12273: Add validations in PGP Module to avoid NPE and improve traceability

facundovs:
Currently, Mule PGP module doesn't have sufficient validations, so NPExceptions are triggered in the following cases:

- Not found Private Key file.
- Not found Public Key file.
- Incorrect Principle.
- Not defined secretPassphrase.
- Unreadable files.

Validations must be added to have a better traceability of the errors.
AmbiguousPGPPrincipalException was added to validate repeated usersId in different public keys.
When no valid principal is set, in the exception will be suggested a list of available principals order by keyIds.
Besides, CryptoFailureException is not propagating the exception message, causing traceability failures.


modules/pgp/src/main/java/org/mule/module/pgp/i18n/PGPMessages.java
```diff
@@ -50,6 +50,31 @@ public static Message noSecretKeyFoundButAvailable(String availableKeys)
         return factory.createMessage(BUNDLE_PATH, 7, availableKeys);
     }
 
+    public static Message noPublicKeyForPrinciple(String principalId, String availablePrincipals)
```
alepulver: noPublicKeyForPrinciple -> noPublicKeyForPrincipal

modules/pgp/src/main/resources/META-INF/services/org/mule/i18n/pgp-messages.properties
```diff
@@ -1,7 +1,12 @@
...
 5=No encryption strategy has been set. You need to set either the "strategyName" to lookup or an EncryptionStrategy object "strategy"
 6=PGP public key has expired
-7=No Secret Key found but the following are available: {0}
\ No newline at end of file
+7=No Secret Key found but the following are available: {0}
+8=No public key known for principal: {0}. Available principals: {1}
+9=No key file found in: {0}
+10=No secret key ring filename defined
+11=No secretPassphrase attribute defined
+12=Principal with value "{0}" exists in keys: "{1}" and "{2}", but It should be unique.
```
alepulver: It -> it

PR #4550, opened 2017-04-23
MULE-12101 Replace use of EventDroppedException for child context

dfeist:


core/src/main/java/org/mule/runtime/core/api/processor/MessageProcessors.java
```diff
@@ -6,23 +6,27 @@
  */
 package org.mule.runtime.core.api.processor;
 
+import static org.mule.runtime.core.DefaultEventContext.child;
+import static org.mule.runtime.core.api.Event.builder;
 import static org.mule.runtime.core.api.rx.Exceptions.rxExceptionToMuleException;
 import static org.mule.runtime.core.processor.chain.ExplicitMessageProcessorChainBuilder.*;
```
elrodro83: no *

core/src/main/java/org/mule/runtime/core/internal/util/rx/Operators.java
```diff
@@ -30,13 +29,13 @@ private Operators() {}
...
-        if (r != null) {
-          sink.next(r);
+    return (event, sink) -> {
+      if (event != null) {
+        Event result = mapper.apply(event);
+        if (result != null) {
+          sink.next(result);
         } else {
-          sink.error(newEventDroppedException(t));
+          event.getContext().success();
```
elrodro83: is `success` correct here? previously it was an exception

dfeist: This is exactly what this PR is doing:
Removing the use of an exception to represent a dropped message, which then handled at the end of the flow with ` event.getContext().success(); to instead complete with no response immediately.  The reason I used exception in the first place was to handle the enricher/response-adaptor scenarios where we need to handle the null rather than respond null, but I'm now using a different approach to do this using a child context and can therefore clean out all use of ExceptionDroppedException.

PR #4553, opened 2017-04-24
MULE-11092: Classloading: plugin.properties should be dropped in favo…

gsfernandes:
…ur of mule-plugin.json


gsfernandes:
Jenkins validate

PR #4557, opened 2017-04-24
MULE-10936: Allow the SchedulerService to create pools per app rather

elrodro83:
that per container - split scheduler service

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/scheduler/SchedulerPoolsConfig.java
+++ b/core/src/main/java/org/mule/runtime/core/api/scheduler/SchedulerPoolsConfig.java
@@ -6,7 +6,6 @@
  */
 package org.mule.runtime.core.api.scheduler;
 
-import java.util.Properties;
 import java.util.concurrent.Executor;
 
 /**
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/mule/runtime/core/api/scheduler/SchedulerPoolsConfig.java
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * @return the size of the queue to use for holding {@code cpu_intensive} tasks before they are executed.
+   */
+  int getCpuIntensiveQueueSize();
+
+  /**
+   * @param name the name of the owner of the quartz scheduler
+   * @return the properties to provide the quartz scheduler
+   */
+  Properties defaultQuartzProperties(String name);
```
dfeist: We shouldn't really be coupling to quartz here, even if not in compilation. Is there no way to avoid this?

services/scheduler/src/main/java/org/mule/service/scheduler/internal/threads/SchedulerThreadPools.java
```diff
@@ -0,0 +1,390 @@
...
+import org.slf4j.Logger;
+
+/**
+ * {@link Scheduler}s provided by this implementation of {@link SchedulerService} use a shared single-threaded
+ * {@link ScheduledExecutorService} for scheduling work. When a scheduled tasks is fired, they are executed using the
+ * {@link Scheduler}'s own executor.
+ * 
+ * @since 4.0
+ */
+public class SchedulerThreadPools {
```
dfeist: Was this moved or renamed? I see this added, but nothing removed..

elrodro83: The code here was removed from ´DefaultSchedulerService´

PR #4561, opened 2017-04-25
MULE-12077: Allow operations to override parameters from config

alegmarra:


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/source/SourceExtensionObjectFactory.java
```diff
@@ -44,7 +44,7 @@
  *
  * @since 4.0
  */
-public class ExtensionSourceObjectFactory extends AbstractExtensionObjectFactory<ExtensionMessageSource> {
+public class SourceExtensionObjectFactory extends AbstractExtensionObjectFactory<ExtensionMessageSource> {
```
marianogonzalez: Bad rename. The term `Extension` is qualifying the `Source` not the `Source` qualifying the `Extension`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/ModelValidationUtils.java
```diff
@@ -0,0 +1,45 @@
...
+import org.mule.runtime.extension.api.loader.ProblemsReporter;
+
+import java.util.List;
+
+/**
+ * Utility class for {@link ExtensionModelValidator}s
+ *
+ * @since 4.0
+ */
+class ModelValidationUtils {
```
marianogonzalez: final.. add empty private constructor

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ValueResolver.java
```diff
@@ -26,6 +29,19 @@
...
   /**
+   * Resolves a value from the given {@code event} and the {@link ConfigurationInstance config} associated
+   * to the component, if one is present.
+   *
+   * @param event a {@link Event}
+   * @param config an {@link Optional} {@link ConfigurationInstance}
+   * @return a resolved value
+   * @throws MuleException if the resolution of the value fails
+   */
+  default T resolve(Event event, Optional<ConfigurationInstance> config) throws MuleException {
```
marianogonzalez: i don't like this method to begin with. But assuming you end up convincing me of it, it should be the other way around. `T resolve(Event event)` should have a default impl which calls this one with an empty optional. You should force ValueResolver developers to consider the possibility of the config being present.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/IntrospectionUtils.java
```diff
@@ -399,6 +402,44 @@ public static Object getFieldValue(Object object, String fieldName) throws Illeg
...
+   * Retrieves the value of the parameter of name {@code parameterName} obtained from the {@code delegate} or
+   * from {@link ConfigurationInstance#getValue() config instance} if the {@code delegate} produces a
+   * {@code null} value.
+   *
+   * @param config the {@link ConfigurationInstance config} from where the parameter value will be obtained
+   * @param parameterName the name parameter to resolve
+   * @return the value of the parameter with name {@code parameterName} obtained from the {@code delegate} or
+   * from {@link ConfigurationInstance#getValue() config instance} if the {@code delegate} produces a {@code null} value.
+   */
+  public static Object resolveConfigOverrideParameter(Supplier delegate, Optional<ConfigurationInstance> config,
```
marianogonzalez: Several things are wrong with this method:

* This doesn't belong here in the first place. IntrospectionUtils should at most have logic for locating the field/argument which is annotated, but going as far as actually evaluating and returning a value is not something that should happen here
* Fail fast. This performs a lot of checks which could fail but the supplier is evaluated before hand. You might incur into an expensive computation for no gain
* This method is being constantly called from the ValueResolver. This constant introspection of the model, etc should be done on the value resolver constructor and the field cached. Otherwise this is a bottleneck

PR #4564, opened 2017-04-25
MULE-12189: Region class loader must check that different artifacts a…

pablokraan:
…re not exporting the same package

_ Updating ClassLoaderLookupPolicy to provide getClassLookupStrategy and getPackageLookupStrategy methods
_ Updating RegionClassLoader to verify that no classloader can export anything that is already provided from the parent classloader lookup policy
_ Updating RegionClassLoader to verify that multiple classloaders in the region do not export the same package
_ Moving integration tests that where located in the wrong package
_ Updating AbstractArtifactClassLoaderBuilder to sanitize exported packages for region artifacts
_ Updating IsolatedClassLoaderFactory to sanitize test packages exported on the test classloader
_ Updating IsolatedClassLoaderFactory to sanitize packages exported by plugins used on the tests


tests/runner/src/main/java/org/mule/test/runner/classloader/IsolatedClassLoaderFactory.java
```diff
@@ -88,6 +93,8 @@
    */
   public ArtifactClassLoaderHolder createArtifactClassLoader(List<String> extraBootPackages,
                                                              ArtifactsUrlClassification artifactsUrlClassification) {
+    JarInfo appJarInfo = getTestJarInfo(artifactsUrlClassification);
```
elrodro83: test vs app, can the naming be made consistent?

tests/runner/src/main/java/org/mule/test/runner/classloader/IsolatedClassLoaderFactory.java
```diff
@@ -235,23 +241,79 @@ private boolean hasPrivilegedApiAccess(PluginUrlClassification pluginUrlClassifi
...
+   * <p/>
+   * A similar sanitization is done for packages that are system packages, as child artifacts cannot redefine them.
+   *
+   * @param productionPackages all packages from the module under test's production code.
+   * @param testPackages all packages from the module under test's test code
+   * @return sanitized packages to export on the test class loader.
+   */
+  private Set<String> sanitizeTestExportedPackages(Set<String> productionPackages, Set<String> testPackages) {
+    Set<String> sanitizedTestPackages = new HashSet<>(testPackages);
+    removePackagesFromTestClassLoader(sanitizedTestPackages, ContainerClassLoaderFactory.SYSTEM_PACKAGES);
```
elrodro83: static import

PR #4566, opened 2017-04-25
MULE-12107: Add Dynamic Metadata resolution support in Soap Connect.

juandesi:


modules/extensions-spring-support/pom.xml
```diff
@@ -172,12 +172,7 @@
             <artifactId>cxf-rt-frontend-jaxws</artifactId>
             <version>${cxfVersion}</version>
             <scope>test</scope>
-            <exclusions>
```
marianogonzalez: why remove this? IIRC this was added to fix other issues

modules/extensions-spring-support/pom.xml
```diff
@@ -312,5 +308,16 @@
...
+        <dependency>
+            <groupId>org.skyscreamer</groupId>
+            <artifactId>jsonassert</artifactId>
+            <version>1.1.1</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mule.runtime</groupId>
+            <artifactId>mule-service-http-api</artifactId>
+            <version>${project.version}</version>
```
marianogonzalez: this is either missing the test scope or should be defined in a prior section

modules/extensions-spring-support/pom.xml
juandesi: my bad, I wanted to copy it and I cut it 😄

PR #4569, opened 2017-04-25
MULE-6129 - Change default behavior for Choice when no otherwise Proc…

pablolagreca:
…essor is set

### DIFF
--- a/tests/integration/src/test/resources/org/mule/test/integration/routing/choice-router-config.xml
+++ b/tests/integration/src/test/resources/org/mule/test/integration/routing/choice-router-config.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<mule xmlns="http://www.mulesoft.org/schema/mule/core" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xmlns:spring="http://www.springframework.org/schema/beans"
+      xmlns:test="http://www.mulesoft.org/schema/mule/test"
+      xsi:schemaLocation="
+               http://www.mulesoft.org/schema/mule/test http://www.mulesoft.org/schema/mule/test/current/mule-test.xsd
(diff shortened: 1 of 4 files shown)


tests/integration/src/test/resources/org/mule/test/integration/routing/choice-router-config.xml
```diff
@@ -0,0 +1,19 @@
...
+      xmlns:spring="http://www.springframework.org/schema/beans"
+      xmlns:test="http://www.mulesoft.org/schema/mule/test"
+      xsi:schemaLocation="
+               http://www.mulesoft.org/schema/mule/test http://www.mulesoft.org/schema/mule/test/current/mule-test.xsd
+               http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-current.xsd
+               http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd">
+
+    <flow name="flow">
+        <choice>
+            <when expression="#[payload == 'A']">
```
pablokraan: Does not make more sense to implement the proposal defined at https://www.mulesoft.org/jira/browse/MULE-6129?focusedCommentId=80558&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-80558 instead of supporting a choice with not default?

pablolagreca: I really don't think it makes sense to have two different routers that are practically the same.
Pros:
- We have a single router instead of two for conditionals simplifying the core routers
- If an additional route needs to be added later on, the user doesn't have to change from <choice> to <if>, they would just add a single router
Cons:
- An <if> it's more clear from the DSL standpoint if there's a single route, but from the graphical editor I think it would be pretty much the same.

PR #4572, opened 2017-04-26
MULE-10394 - using the classloader of the service instead of the curr…

pablolagreca:
…ent one and fixing test case to discover artifacts from mulesoft public repo

### DIFF
--- a/modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/application/ApplicationDescriptorFactoryTestCase.java
+++ b/modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/application/ApplicationDescriptorFactoryTestCase.java
@@ -84,11 +84,11 @@ public class ApplicationDescriptorFactoryTestCase extends AbstractMuleTestCase {
   public static final String JAR_FILE_NAME = "test.jar";
 
   @Rule
-  public SystemProperty remoteMavenRepository = new SystemProperty("muleRuntimeConfig.maven.repositoryLocation",
-                                                                   discoverProvider(ApplicationDescriptorFactoryTestCase.class
-                                                                       .getClassLoader()).getLocalRepositorySuppliers()
-                                                                           .environmentMavenRepositorySupplier().get()
-                                                                           .getAbsolutePath());
+  public SystemProperty repositoryLocation = new SystemProperty("muleRuntimeConfig.maven.repositoryLocation",
+                                                                discoverProvider(ApplicationDescriptorFactoryTestCase.class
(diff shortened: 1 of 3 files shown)


modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/application/ApplicationDescriptorFactoryTestCase.java
```diff
@@ -84,11 +84,16 @@ private static File getResourceFile(String resource) {
...
-                                                                           .environmentMavenRepositorySupplier().get()
-                                                                           .getAbsolutePath());
+  public SystemProperty repositoryLocation = new SystemProperty("muleRuntimeConfig.maven.repositoryLocation",
+                                                                discoverProvider(ApplicationDescriptorFactoryTestCase.class
+                                                                    .getClassLoader()).getLocalRepositorySuppliers()
+                                                                        .environmentMavenRepositorySupplier().get()
+                                                                        .getAbsolutePath());
+
+  @Rule
+  public SystemProperty remoteMavenRepository =
```
pablokraan: Can't you use M2_REPO instead or the remote one? If it is required to go remote to find an artifact, then that is wrong to me, every artifact required on the test should be already built or downloaded as part of another module build.

gsfernandes: I think that there is another problem here, as the code before this change was always getting the artifacts resolved from the local repository. Where there is a logic to resolve it by searching first for the "localRepository" system property (set by Maven) or https://github.com/mulesoft/mule-maven-client/blob/master/mule-maven-client-impl/src/main/java/org/mule/maven/client/internal/DefaultLocalRepositorySupplierFactory.java#L69
I'm wondering if this test was failing as it was trying to resolve a plugin that was not build yet (an extension for instance) but as it is deployment module it cannot depend on it.
If that's the case we should create a test extension/plugin on this module to test this logic instead of depending on a real one.
I just found that this test is trying to validate an scenario where the application descriptor to be created has http extension.
This is a problem as it will never be built before this module is build.
I will keep this "fix" and create another task to remove these kind of tests to another module.

pablokraan: If that is the problem, then it should be enough to declare a dependency on the deployment module pom to the HTTP extension. I don't think that that will introduce a circular dependency

gsfernandes:
@pablokraan, changes as we discussed. Added the dependency to http-ext as test and removed the settings for remote repositories.
Please review it.

PR #4573, opened 2017-04-26
MULE-12219 Nested MessageProcessors post notifications are triggered within the error message

dfeist:
PENDING BUILD


core/src/main/java/org/mule/runtime/core/internal/util/rx/IgnoreSubscriber.java
```diff
@@ -0,0 +1,41 @@
...
+ * Singleton {@link Subscriber} that request unlimited demand and ignores errors.
+ * 
+ * @param <T>
+ */
+public class IgnoreSubscriber<T> extends BaseSubscriber<T> {
+
+  private static Subscriber INSTANCE = new IgnoreSubscriber<>();
+
+  public static <T> Subscriber<T> get() {
+    return new IgnoreSubscriber<>();
```
alepulver: Shouldn't this return INSTANCE?

core/src/main/java/org/mule/runtime/core/processor/chain/AbstractMessageProcessorChain.java
```diff
@@ -103,16 +109,55 @@ protected Event doProcess(Event event) throws MuleException {
...
+            .onProcessor(processorFunction);
+      }
+
+      ReactiveProcessor finalProcessorFunction = processorFunction;
+      stream = from(stream).flatMap(event -> {
+        try {
+          return Flux.just(event)
+              .transform(finalProcessorFunction)
+              .onErrorResumeWith(MessagingException.class, handleError(event.getContext()));
+        } catch (Throwable throwable) {
```
alepulver: Is this catch ever executed? If the flux is being used like a "builder" here, and not run (with force/subscribe/etc), exceptions thrown in finalProcessorFunction or handleError won't be caught here.

core/src/main/java/org/mule/runtime/core/internal/util/rx/IgnoreSubscriber.java
dfeist: Oops yes.

core/src/main/java/org/mule/runtime/core/processor/chain/AbstractMessageProcessorChain.java
dfeist: This needs to be removed, forgot to remove it.

PR #4577, opened 2017-04-26
MULE-11752: WSC doesnt pick WSDL files from the current classpath

juandesi:


extensions/wsc/src/main/java/org/mule/extension/ws/internal/connection/SoapClientConnectionProvider.java
```diff
@@ -120,4 +140,9 @@ public ConnectionValidationResult validate(SoapClient client) {
     // TODO MULE-12036
     return ConnectionValidationResult.success();
   }
+
+  private String getWsdlLocation(String wsdlLocation) {
+    URL resource = currentThread().getContextClassLoader().getResource(wsdlLocation);
```
marianogonzalez: have you tried this on an actual application running on an actual mule? In the test case this will always work because the extension's classloader will include this test resource

PR #4578, opened 2017-04-26
MULE-11192: Tooling Runtime Client handle multiple configurations whe…

gsfernandes:
…n building ApplicationModel for DataSense


modules/spring-config/src/main/java/org/mule/runtime/config/spring/dsl/model/ApplicationModel.java
```diff
@@ -472,7 +472,12 @@ private void convertConfigFileToComponentModel(ArtifactConfig artifactConfig) {
       ComponentModel componentModel =
           componentModelReader.extractComponentDefinitionModel(configFile.getConfigLines().get(0), configFile.getFilename());
       if (isMuleConfigFile(configFile)) {
-        muleComponentModels.add(componentModel);
+        if (muleComponentModels.isEmpty()) {
+          muleComponentModels.add(componentModel);
+        } else {
+          // Only one componentModel as Root should be set, therefore componentModel from other configFiles are added as innerComponents
+          muleComponentModels.get(0).getInnerComponents().addAll(componentModel.getInnerComponents());
```
alegmarra: should we copy the parameters and custom attributes from one root to the other?

PR #4583, opened 2017-04-27
MULE-10936: Allow the SchedulerService to create pools per app rather

elrodro83:
that per container
* review

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/scheduler/SchedulerPoolsConfig.java
+++ b/core/src/main/java/org/mule/runtime/core/api/scheduler/SchedulerPoolsConfig.java
@@ -19,48 +19,48 @@ public interface SchedulerPoolsConfig {
    * @return the maximum time (in milliseconds) to wait until all tasks in all the runtime thread pools have completed execution
    *         when stopping the scheduler service.
    */
-  long getGracefulShutdownTimeout();
+  Long getGracefulShutdownTimeout();
 
   /**
    * @return the number of threads to keep in the {@code cpu_lite} pool, even if they are idle.
    */
-  int getCpuLightPoolSize();
+  Integer getCpuLightPoolSize();
 
   /**
    * @return the size of the queue to use for holding {@code cpu_lite} tasks before they are executed.
    */
-  int getCpuLightQueueSize();
+  Integer getCpuLightQueueSize();
 
   /**
    * @return the number of threads to keep in the {@code I/O} pool.
    */
-  int getIoCorePoolSize();
+  Integer getIoCorePoolSize();
 
   /**
    * @return the maximum number of threads to allow in the {@code I/O} pool.
    */
-  int getIoMaxPoolSize();
+  Integer getIoMaxPoolSize();
 
   /**
    * @return the size of the queue to use for holding {@code I/O} tasks before they are executed.
    */
-  int getIoQueueSize();
+  Integer getIoQueueSize();
 
   /**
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/mule/runtime/core/api/scheduler/SchedulerPoolsConfig.java
```diff
@@ -19,48 +19,48 @@
    * @return the maximum time (in milliseconds) to wait until all tasks in all the runtime thread pools have completed execution
    *         when stopping the scheduler service.
    */
-  long getGracefulShutdownTimeout();
+  Long getGracefulShutdownTimeout();
```
dfeist: Why not Long -> OptionalInt?

services/scheduler/src/main/java/org/mule/service/scheduler/internal/config/ContainerThreadPoolsConfig.java
```diff
@@ -163,83 +169,83 @@ private ContainerThreadPoolsConfig() {
   }
 
   @Override
-  public long getGracefulShutdownTimeout() {
+  public Long getGracefulShutdownTimeout() {
     return gracefulShutdownTimeout;
   }
 
-  private void setGracefulShutdownTimeout(long gracefulShutdownTimeout) {
+  private void setGracefulShutdownTimeout(Long gracefulShutdownTimeout) {
```
dfeist: Why do setters need to use a nullable value?

services/scheduler/src/main/java/org/mule/service/scheduler/internal/threads/SchedulerThreadPools.java
```diff
@@ -107,20 +107,21 @@ public SchedulerThreadPools(String name, SchedulerPoolsConfig threadPoolsConfig)
...
-        new ThreadPoolExecutor(threadPoolsConfig.getCpuLightPoolSize(), threadPoolsConfig.getCpuLightPoolSize(),
-                               0, SECONDS, new LinkedBlockingQueue<>(threadPoolsConfig.getCpuLightQueueSize()),
+        new ThreadPoolExecutor(threadPoolsConfig.getCpuLightPoolSize().getAsInt(),
+                               threadPoolsConfig.getCpuLightPoolSize().getAsInt(),
+                               0, SECONDS, new LinkedBlockingQueue<>(threadPoolsConfig.getCpuLightQueueSize().getAsInt()),
                                new SchedulerThreadFactory(cpuLightGroup), byCallerThreadGroupPolicy);
     ioExecutor =
-        new ThreadPoolExecutor(threadPoolsConfig.getIoCorePoolSize(), threadPoolsConfig.getIoMaxPoolSize(),
-                               threadPoolsConfig.getIoKeepAlive(), MILLISECONDS,
+        new ThreadPoolExecutor(threadPoolsConfig.getIoCorePoolSize().getAsInt(), threadPoolsConfig.getIoMaxPoolSize().getAsInt(),
```
dfeist: What about default if not set?

elrodro83: we'll handle that when we make values not required

dfeist: ok, at least it with throw exception, and not just return null.

PR #4588, opened 2017-04-27
MULE-12335: Database Connector unable to handle BLOB conversion.

fsgonz:


modules/db/src/main/java/org/mule/module/db/internal/config/domain/database/DbConfigDefinitionParser.java
```diff
@@ -123,9 +123,9 @@ else if (id == Types.STRUCT)
                         customDbTypes.add(new StructDbType(id, name));
                     }
                 }
-                else if (id == Types.CLOB)
+                else if (types.contains(new ResolvedDbType(id, name)))
                 {
-                    customDbTypes.add(new ClobResolvedDataType(id, name));
+                    customDbTypes.add(types.get(types.indexOf(new ResolvedDbType(id, name))));
```
pablokraan: Should not use Clob/Blob ResolvedDataType here?

modules/db/src/main/java/org/mule/module/db/internal/domain/type/AbstractDbType.java
```diff
@@ -48,4 +48,30 @@ public String toString()
     {
         return name;
     }
+
+    @Override
+    public boolean equals(Object obj)
+    {
+        if (this == obj)
```
pablokraan: Fix code style

modules/db/src/test/java/org/mule/module/db/internal/domain/type/LobResolvedDataTypeTestCase.java
```diff
@@ -29,34 +32,60 @@
 import org.junit.rules.ExpectedException;
 
 @SmallTest
-public class ClobResolvedDataTypeTestCase extends AbstractMuleTestCase {
+public class LobResolvedDataTypeTestCase extends AbstractMuleTestCase {
```
pablokraan: Used separated classes to test clboa and blob types. If you need to reuse, then create an abstract superclass

modules/db/src/main/java/org/mule/module/db/internal/domain/type/AbstractDbType.java
```diff
@@ -48,4 +48,30 @@ public String toString()
...
+        if (name == null)
+        {
+            if (other.name != null)
+                return false;
+        }
+        else if (!name.equals(other.name))
+            return false;
+        return true;
+    }
+
```
pablokraan: Remove extra lines

modules/db/src/main/java/org/mule/module/db/internal/config/domain/database/DbConfigDefinitionParser.java
fsgonz: This was done in order to always retrieve the subclass of ResolvedDataType which is declared in JdbcDataType. When you use 

types.get(types.indexOf(new ResolvedDbType(id, name))) 

as the equals is overriden and it only takes into account the id and the name, it will get ClobResolvedDataType or BlobResolvedDataType according to the id and name attributes of the ResolvedDbType used to get the class in JdbcTypes.

PR #4591, opened 2017-04-28
MULE-10394: Added support for setting remote repositories with order

gsfernandes:


gsfernandes:
@pablolagreca this is partial fix as the framework that merges the configs does not support this kind of feature. I have made the decision that if there is no position it will after the ones positioned.
Still if we want later to persist the config to the json file it will no write the position value.

modules/global-config/src/main/java/org/mule/runtime/globalconfig/internal/MavenConfigBuilder.java
```diff
@@ -104,6 +107,16 @@ public static MavenConfiguration buildMavenConfig(Config mavenConfig) {
     }
   }
 
+  private static Comparator<Map.Entry<String, Object>> remoteRepositoriesComparator() {
+    return (firstEntry, secondEntry) -> {
+      Integer firstPosition = Integer
+          .valueOf(((Map<String, String>) firstEntry.getValue()).getOrDefault(POSITION, String.valueOf(Integer.MAX_VALUE)));
```
pablokraan: Static imports

tests/unit/src/main/java/org/mule/tck/MuleTestUtils.java
```diff
@@ -86,6 +88,38 @@ public static void testWithSystemProperty(String propertyName, String propertyVa
     }
   }
 
+  /**
+   * Execute callback with a given system properties set and replaces the system property with it's original value once done. Useful
```
pablokraan: Execute -> Executes

PR #4592, opened 2017-04-28
MULE-10394: Support for SNAPSHOT versions and timestamp resolutions

gsfernandes:


gsfernandes:
Depends on https://github.com/mulesoft/mule-maven-client/pull/2

PR #4597, opened 2017-05-01
MULE-12345: Maven dependency resolution doesn't output logs

elrodro83:
* Bump slf4j version
* Configure logging for embedded as early as possible, to get the
logging of the dependencies resolving


embedded/embedded-test/src/test/java/org/mule/runtime/module/embedded/EmbeddedContainerTestCase.java
```diff
@@ -218,4 +221,9 @@ private void doWithinApplication(String applicaitonFolder, Consumer<Integer> por
   private URI getClasspathResourceAsUri(String resource) throws URISyntaxException {
     return getClass().getClassLoader().getResource(resource).toURI();
   }
+
+  @Override
+  public int getTestTimeoutSecs() {
+    return 20 * super.getTestTimeoutSecs();
```
dfeist: Is timeout of 20 minutes good idea?

elrodro83: this test was currently downloading all of its dependencies from a remote repo. It took 11 minutes on my computer to run.

PR #4600, opened 2017-05-02
MULE-12050: Export default services providers from the JRE

pablokraan:
_ Adding JreExplorer to detect packages, resources and services from JRE
_ Removing jre-pacakges.properties
_ Removing some boot packages


pablokraan:
I just checked this on a standalone distribution, all jars on the endorsed lib folder are processed first and then, services from those libs are put at the beginning of the exported services list

PR #4601, opened 2017-05-02
MULE-10619: FilteringArtifactClassLoader is not being used by RegionC…

pablokraan:
…lassLoader when loading classes and resources instead it uses the ArtifactClassLoader

### DIFF
--- a/modules/artifact/src/main/java/org/mule/runtime/module/artifact/classloader/RegionClassLoader.java
+++ b/modules/artifact/src/main/java/org/mule/runtime/module/artifact/classloader/RegionClassLoader.java
@@ -240,9 +240,9 @@ public class RegionClassLoader extends MuleDeployableArtifactClassLoader {
     } catch (Exception e) {
       final String message = "Error disposing classloader for '{}'. This can cause a memory leak";
       if (logger.isDebugEnabled()) {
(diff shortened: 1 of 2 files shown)


modules/artifact/src/main/java/org/mule/runtime/module/artifact/classloader/RegionClassLoader.java
```diff
@@ -233,23 +226,27 @@ public final URL findResource(final String name) {
...
 
+  private void disposeClassLoader(ArtifactClassLoader classLoader) {
+    try {
+      classLoader.dispose();
+    } catch (Exception e) {
+      final String message = "Error disposing classloader for '{}'. This can cause a memory leak";
+      if (logger.isDebugEnabled()) {
+        logger.debug(message, getArtifactDescriptor().getName(), e);
+      } else {
+        logger.error(message, getArtifactDescriptor().getName());
```
elrodro83: should somehow indicate which call to this method (which inner classlaoder) failed

pablokraan: getArtifactDescriptor().getName() tells you wich classloader had the problem

PR #4612, opened 2017-05-02
Mule 11749

alepulver:
DO NOT MERGE YET

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/rx/Exceptions.java
+++ b/core/src/main/java/org/mule/runtime/core/api/rx/Exceptions.java
@@ -6,11 +6,15 @@
  */
 package org.mule.runtime.core.api.rx;
 
+import static reactor.core.Exceptions.isBubbling;
+import static reactor.core.Exceptions.propagate;
 import static reactor.core.Exceptions.unwrap;
 
 import org.mule.runtime.api.exception.MuleException;
+import org.mule.runtime.api.exception.MuleRuntimeException;
 import org.mule.runtime.core.api.DefaultMuleException;
 import org.mule.runtime.core.exception.MessagingException;
+import org.mule.runtime.core.exception.MuleFatalException;
 import org.mule.runtime.core.util.func.CheckedBiConsumer;
 import org.mule.runtime.core.util.func.CheckedBiFunction;
 import org.mule.runtime.core.util.func.CheckedBiPredicate;
@@ -153,4 +157,31 @@ public class Exceptions {
     }
   }
 
+  /**
+   * Propagate an exception through streams even if it's considered fatal by Reactor.
+   *
+   * @param t throwable.
+   * @return exception wrapped by Reactor, after possibly wrapping with {@link MuleFatalException}.
+   */
+  public static RuntimeException propagateWrappingFatal(Throwable t) {
+    return propagate(wrapFatal(t));
+  }
+
+  /**
+   * Wrap an exception with {@link MuleFatalException} so that Reactor can handle it safely instead of always bubbling it up.
+   *
+   * @param t throwable.
+   * @return possibly wrapped exception.
+   */
+  public static Throwable wrapFatal(Throwable t) {
+    if (t instanceof LinkageError) {
+      return new MuleFatalException(t);
+    } else if (t instanceof VirtualMachineError) {
+      return new MuleFatalException(t);
+    } else if (isBubbling(t)) {
+      return new MuleRuntimeException(unwrap(t));
+    } else {
+      return t;
+    }
+  }
 }
--- a/core/src/main/java/org/mule/runtime/core/exception/Errors.java
+++ b/core/src/main/java/org/mule/runtime/core/exception/Errors.java
@@ -7,6 +7,8 @@
 package org.mule.runtime.core.exception;
 
 import static org.mule.runtime.api.component.ComponentIdentifier.builder;
+import static org.mule.runtime.core.exception.Errors.Identifiers.FATAL_ERROR_IDENTIFIER;
+import static org.mule.runtime.internal.dsl.DslConstants.CORE_PREFIX;
 import static org.mule.runtime.core.exception.Errors.Identifiers.ANY_IDENTIFIER;
 import static org.mule.runtime.core.exception.Errors.Identifiers.CONNECTIVITY_ERROR_IDENTIFIER;
 import static org.mule.runtime.core.exception.Errors.Identifiers.CRITICAL_IDENTIFIER;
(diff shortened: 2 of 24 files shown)


core/src/main/java/org/mule/runtime/core/exception/Errors.java
```diff
@@ -91,8 +95,10 @@
      */
     public static final String OVERLOAD_ERROR_IDENTIFIER = "OVERLOAD";
 
-    public static final String STREAM_MAXIMUM_SIZE_EXCEEDED_ERROR_IDENTIFIER = "STREAM_MAXIMUM_SIZE_EXCEEDED";
-
+    /**
+     * Indicates that a fatal JVM error occurred (such as stack overflow). Cannot be handled.
+     */
+    public static final String FATAL_JVM_ERROR_IDENTIFIER = "FATAL_JVM_ERROR";
```
dfeist: Any reason not to use just `FATAL`?

core/src/main/java/org/mule/runtime/core/util/ExceptionUtils.java
```diff
@@ -275,4 +278,23 @@ public static Throwable getRootCauseException(Throwable exception) {
     return exception instanceof ErrorMessageAwareException ? ((ErrorMessageAwareException) exception).getRootCause() : exception;
   }
 
+  /**
+   * Propagate an exception through streams, may wrap it to avoid throwing from Reactor code.
+   *
+   * @param t throwable.
+   * @return possibly wrapped exception to avoid Reactor stopping the stream.
+   */
+  public static RuntimeException propagateReactor(Throwable t) {
```
dfeist: I've been putting exceptions stuff in `Exceptions`, maybe that wasn't a good idea in hindsight, but thats where all the rx exception stuff is currently.

core/src/main/java/org/mule/runtime/core/util/ExceptionUtils.java
```diff
@@ -275,4 +278,23 @@ public static Throwable getRootCauseException(Throwable exception) {
...
+    return propagate(wrapReactor(t));
+  }
+
+  public static Throwable wrapReactor(Throwable t) {
```
dfeist: Javadoc. Also consider is there is a clearer name we can give this method, otherwise doesn't matter as long as javadoc is clear.

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/HeisenbergOperations.java
```diff
@@ -266,6 +266,10 @@ public void ignoredOperation() {
 
   }
 
+  public void throwError() {
+    throw new AssertionError();
```
dfeist: This isn't fatal error as far as reactor is concerned, should we test with a fatal JVM error too? e.g. StackOverflow?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/ReactiveOperationExecutionWrapper.java
```diff
@@ -64,6 +65,8 @@ public ReactiveOperationExecutionWrapper(OperationExecutor delegate) {
         delegate.execute(executionContext);
       } catch (Exception e) {
         sink.error(e);
+      } catch (Throwable t) {
+        sink.error(propagateReactor(t));
```
dfeist: Why propagate() and not just wrap()?

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/HeisenbergOperations.java
alepulver: It should be LinkageError, I was testing and tried both. The AssertionError was not previously handled but it is from this change on.

core/src/main/java/org/mule/runtime/core/api/rx/Exceptions.java
```diff
@@ -153,4 +155,29 @@ public static MuleException rxExceptionToMuleException(Throwable throwable)
     }
   }
 
+  /**
+   * Propagate an exception through streams even if it's considered fatal by Reactor.
+   *
+   * @param t throwable.
+   * @return exception wrapped by Reactor, after possibly wrapping with {@link MuleFatalException}.
+   */
+  public static RuntimeException propagateFatal(Throwable t) {
```
dfeist: How about just `propagate` because this is not always propagating fatal

core/src/main/java/org/mule/runtime/core/api/rx/Exceptions.java
```diff
@@ -153,4 +155,29 @@ public static MuleException rxExceptionToMuleException(Throwable throwable)
...
+   *
+   * @param t throwable.
+   * @return possibly wrapped exception.
+   */
+  public static Throwable wrapFatal(Throwable t) {
+    if (t instanceof LinkageError) {
+      return new MuleFatalException(t);
+    } else if (t instanceof VirtualMachineError) {
+      return new MuleFatalException(t);
+    } else {
```
dfeist: Should we also catch BubblingException here?  Maybe catch it, unwrap it and throw fatal with cause?

core/src/main/java/org/mule/runtime/core/exception/MuleFatalException.java
```diff
@@ -0,0 +1,46 @@
...
+import org.mule.runtime.api.exception.MuleRuntimeException;
+import org.mule.runtime.api.i18n.I18nMessage;
+
+/**
+ * {@code MuleFatalException} Is the runtime exception used to wrap fatal errors such as {@code VirtualMachineError}. This
+ * is to avoid Reactor intercepting them in a stream and propagating outside to arbitrary threads.
+ *
+ * @since 4.0
+ */
+public class MuleFatalException extends MuleRuntimeException {
```
dfeist: final?

core/src/main/java/org/mule/runtime/core/api/rx/Exceptions.java
dfeist: Maybe `propagateWrapFatal` `propagateWrappingFatal` then? Just because in most cases this will not actually propagate a fatal

core/src/main/java/org/mule/runtime/core/api/rx/Exceptions.java
alepulver: They are private so we can't catch them. We could always try to unwrap them, though. But if it was a ReactiveException we wouldn't be wrapping it again in wrapFatal, only in propagateWrappingFatal.

Would it be OK to just use propagateWrappingFatal instead of wrapFatal?

core/src/main/java/org/mule/runtime/core/api/rx/Exceptions.java
alepulver: Ok.

core/src/main/java/org/mule/runtime/core/exception/MuleFatalException.java
alepulver: Ok.

PR #4613, opened 2017-05-02
MULE-12369: Runtime should know how to transform java 8 dates to java…

marianogonzalez:
… 7 dates


core-tests/src/test/java/org/mule/runtime/core/internal/transformer/simple/LocalDateToDateTestCase.java
```diff
@@ -0,0 +1,49 @@
...
+import org.mule.tck.junit4.AbstractMuleContextTestCase;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.ZoneOffset;
+import java.util.Date;
+
+import org.junit.Test;
+
+public class LocalDateToDateTestCase extends AbstractMuleContextTestCase {
```
dfeist: No value in extending AbstractTransformerTestCase, like other transformer tests?

marianogonzalez: no

PR #4617, opened 2017-05-03
MULE-10394 - updating test case to use mulesoft public repository for…

pablolagreca:
… maven artifacts

### DIFF
--- a/modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/application/ApplicationDescriptorFactoryTestCase.java
+++ b/modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/application/ApplicationDescriptorFactoryTestCase.java
@@ -91,6 +91,11 @@ public class ApplicationDescriptorFactoryTestCase extends AbstractMuleTestCase {
                                                                         .getAbsolutePath());
 
   @Rule
+  public SystemProperty remoteMavenRepository =
+      new SystemProperty("muleRuntimeConfig.maven.repositories.mulesoftPublic.url",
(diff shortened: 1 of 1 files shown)


modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/application/ApplicationDescriptorFactoryTestCase.java
```diff
@@ -91,6 +91,11 @@ private static File getResourceFile(String resource) {
                                                                         .getAbsolutePath());
 
   @Rule
+  public SystemProperty remoteMavenRepository =
+      new SystemProperty("muleRuntimeConfig.maven.repositories.mulesoftPublic.url",
```
pablokraan: I have some kind of deja-vu with this change... Why a unit test has to download artifacts from a external repo?

pablolagreca: The only thing about this test that make it seem like an unit test is that extends AbstractMuleTestCase. But is actually testing the whole creation of an application descriptor from an application folder structure.

pablokraan: but why it has to download the dependencies instead of using the ones that are built on the same reactor?

pablolagreca: Closing this PR for now since making the repo private and per executor in CB seems to make it work.

PR #4618, opened 2017-05-03
MULE-12376: Remove transaction constraint classes

pablokraan:


core/src/main/resources/META-INF/mule-module.properties
```diff
@@ -124,7 +124,6 @@ artifact.export.classPackages=org.mule.runtime.core,\
                               org.mule.runtime.core.streaming.bytes,\
                               org.mule.runtime.core.streaming.object,\
                               org.mule.runtime.core.transaction,\
-                              org.mule.runtime.core.transaction.constraints,\
```
elrodro83: image

PR #4620, opened 2017-05-03
MULE-12366: Temporary Queues aren't deleted when JMS is used with BTM

facundovs:
When JMS is used with BTM and a request-response exchange is applied, temporary queues aren't deleted.
The problem is that the consumer of the temporary queue is never closed.
This consumer is instance of MessageConsumerWrapper (Bitronix), and its close() method is empty.
The solution is similar to the following mule code:
https://github.com/mulesoft/mule-ee/blob/mule-3.x/modules/bti/src/main/java/com/mulesoft/mule/bti/jms/BitronixMessageConsumerInvocationHandler.java#L35
i.e close the consumer using the getMessageConsumer method.
BTM-132 issue mentioned in the code comment is lost, and the bitronix project looks discontinued.

### DIFF
--- a/transports/jms/src/main/java/org/mule/transport/jms/JmsConnector.java
+++ b/transports/jms/src/main/java/org/mule/transport/jms/JmsConnector.java
@@ -43,6 +43,7 @@ import org.mule.transport.jms.redelivery.RedeliveryHandlerFactory;
 import org.mule.util.BeanUtils;
 import org.mule.util.concurrent.ThreadNameHelper;
 
+import java.lang.reflect.Method;
 import java.text.MessageFormat;
 import java.util.Map;
 import java.util.Timer;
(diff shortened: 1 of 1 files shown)


transports/jms/src/main/java/org/mule/transport/jms/JmsConnector.java
```diff
@@ -854,6 +856,30 @@ public void closeQuietly(MessageConsumer consumer)
     }
 
     /**
+     * This method is to close a consumer instance of bitronix.tm.resource.jms.MessageWrapperConsumer due to consumer.close() method is empty in this case.
+     * @param consumer
+     * @throws JMSException
+     */
+    private void closeWrapperConsumer(MessageConsumer consumer) throws JMSException
```
pablolagreca: Since we have the BTI fork can't we fix this bug over there?

PR #4621, opened 2017-05-03
MULE-11184 Processor error hook is not applied to any Publisher

dfeist:
REVIEW ONLY. PENDING BUILD.


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
```diff
@@ -163,7 +163,14 @@ public Event process(Event event) throws MuleException {
       ExecutionContextAdapter<OperationModel> operationContext =
           createExecutionContext(configuration, operationParameters, event);
 
-      // TODO: MULE-11184 - it shouldn't be necessary to create the MessagingException here.
+      // While a hook in reactor is used to map Throwable to MessagingException when an error occurs this does not cover the case
+      // where an error is explicitly triggered via a Sink such as such as when using Mono.create in ReactorCompletionCallback
+      // rather than being thrown by a reactor operator. Although changes could be made to Mule to cater for this in
+      // AbstractMessageProcessorChain, this is not trivial given processor interceptors and a potent performance overhead
+      // associated with the addition of many additional flatMaps. It might be slightly create this to create this
```
marianogonzalez: `It might be slightly create this to create this` -> :o

PR #4624, opened 2017-05-03
MULE-11977: Metadata keys cannot be resolve when the application cont…

gsfernandes:
…ains invalid config-ref


gsfernandes:
Not a valid fix, it has to be more complex the logic to remove a componentModel from the graph when it cannot be resolved due a reference is wrong.

PR #4625, opened 2017-05-03
MULE-11693: Migrate idempotent message filter

elrodro83:
* Make object store timeouts long instead of int


core/src/main/java/org/mule/runtime/core/routing/IdempotentMessageFilter.java
```diff
@@ -38,18 +41,15 @@
  * <b>EIP Reference:</b> <a href="http://www.eaipatterns.com/IdempotentReceiver.html">
  * http://www.eaipatterns.com/IdempotentReceiver.html</a>
  */
-public class IdempotentMessageFilter extends AbstractFilteringMessageProcessor implements Initialisable, Disposable {
+public class IdempotentMessageFilter extends AbstractInterceptingMessageProcessor implements Initialisable, Disposable {
```
pablolagreca: Is it possible to use what MG used for making DB not intercepting anymore?. This will bring problems with the interception API and debugger.

core/src/main/java/org/mule/runtime/core/routing/IdempotentMessageFilter.java
```diff
@@ -134,6 +128,27 @@ protected boolean accept(Event event, Event.Builder builder) {
     }
   }
 
+  @Override
+  public final Event process(Event event) throws MuleException {
+    if (accept(event)) {
+      return processNext(event);
+    } else {
+      event.getContext().success();
+      return null;
```
pablolagreca: One of the things we are considering is to not return null anymore and fail always that the filter does not accept the event. That way we manage that case through the error handler and we avoid weird scenarios with request-response sources and also with some places where the filter is placed and only filters a chain but not the entire flow.

pablolagreca:
This issue description says that an error type for filters must be created for when the filter filters the message

core/src/main/java/org/mule/runtime/core/routing/IdempotentMessageFilter.java
dfeist: If we no longer implemented `process()` then we could clean up all of this significantly. But given we still implement `process()` this processor still needs to intercept the flow and therefore needs to control if next is invoked or not.  The only other way of control if next processor in the chain is invoked would be to return `null` or some special `Event` from this processor which the chain that uses to interrupt execution, but I'm not sure it's worth changing this, and not part of this PR for sure.

core/src/main/java/org/mule/runtime/core/routing/IdempotentMessageFilter.java
dfeist: Whatever we do with other filters and their migration to validators, this processor still needs to be able to drop events.  The mechanism by which this is done in blocking `process` model is by returning null. Echoing the event is not the same behave.

core/src/main/java/org/mule/runtime/core/routing/IdempotentMessageFilter.java
```diff
@@ -134,6 +128,27 @@ protected boolean accept(Event event, Event.Builder builder) {
     }
   }
 
+  @Override
+  public final Event process(Event event) throws MuleException {
+    if (accept(event)) {
+      return processNext(event);
+    } else {
+      event.getContext().success();
```
dfeist: The chain should do this in `process` code path.  I need to check if that is correctly implemented, but this shouldn't be needed here.

tests/integration/src/test/resources/org/mule/test/integration/notifications/message-processor-notification-test-flow.xml
```diff
@@ -224,16 +224,12 @@
...
-        </idempotent-message-filter>
+        <idempotent-message-filter idExpression="#[id]" objectStore-ref="objectStore"/>
         <logger level="INFO"/>
     </flow>
 
     <flow name="idempotent-secure-hash-msg-filter">
-        <idempotent-secure-hash-message-filter idExpression="#[mel:message:id]">
-            <in-memory-store entryTTL="1000" expirationInterval="2000" maxEntries="3000"/>
-        </idempotent-secure-hash-message-filter>
+        <idempotent-secure-hash-message-filter idExpression="#[id]" objectStore-ref="objectStore"/>
```
dfeist: Do we still need both these implementations? Is there anyway to combine these two into a single component?

elrodro83: Until we have a way to write this as a nice expression in DW (https://github.com/mulesoft/data-weave/issues/121) we should keep this.

core-tests/src/test/java/org/mule/runtime/core/routing/IdempotentMessageFilterMule6079TestCase.java
```diff
@@ -24,74 +25,65 @@
 import java.util.Map;
 import java.util.TreeMap;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.junit.Test;
-import org.mockito.Mockito;
 
 public class IdempotentMessageFilterMule6079TestCase extends AbstractMuleContextTestCase {
```
dfeist: Update test name, if this test has been migrated to use validator?

core/src/main/java/org/mule/runtime/core/api/routing/filter/FilteredException.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.mule.runtime.api.exception.MuleException;
+
+/**
+ * Indicates that the execution of the current event is stopped. This exception is thrown to indicate this condition to the source
+ * of the flow.
+ * 
+ * @since 4.0
+ */
+public class FilteredException extends MuleException {
```
dfeist: Not sure I understand this Exception.  Aren't there exceptions already used by validators that can be used? If a specific exception is needed for this validator then surely it should be "DuplicateMessageException" and not "FiltereredException", especially as the idea of filtering is going away.

core-tests/src/test/java/org/mule/runtime/core/routing/IdempotentMessageFilterMule6079TestCase.java
dfeist: Update test name, if this test has been migrated to use validator?

core/src/main/java/org/mule/runtime/core/api/routing/filter/FilteredException.java
pablolagreca: This should be ValidationException and the error type VALIDATION. We need to talk to MG so he start using that same type as parent for the validations of the validation module.

core/src/main/java/org/mule/runtime/core/exception/ErrorTypeRepositoryFactory.java
```diff
@@ -8,6 +8,7 @@
 
 import static org.mule.runtime.core.exception.Errors.ComponentIdentifiers.CONNECTIVITY;
 import static org.mule.runtime.core.exception.Errors.ComponentIdentifiers.EXPRESSION;
+import static org.mule.runtime.core.exception.Errors.ComponentIdentifiers.FILTERED;
```
dfeist: Again, FILTERED -> Either something reused from validations or DUPLICATE_MESSAGE if we need something specific.

tests/integration/src/test/resources/org/mule/test/integration/routing/inbound/idempotent-router-with-filter-flow.xml
```diff
@@ -11,7 +11,7 @@
         <message-filter>
             <payload-type-filter expectedType="java.lang.String"/>
         </message-filter>
-        <idempotent-message-filter/>
+        <idempotent-message-validator/>
```
dfeist: I don't see any test cases where DUPLICATED_MESSAGE (FILTERED) error type is being thrown, and then handled in an error-handler.  Isn't this important given it's the use case where this would typically be being used?

elrodro83: added a case for that

PR #4626, opened 2017-05-03
MULE-11870: support certificate revocation checking in TLS context

alepulver:
DO NOT MERGE YET


core/src/main/java/org/mule/api/security/tls/TlsConfiguration.java
```diff
@@ -112,7 +126,8 @@
 {
     public static final String DEFAULT_KEYSTORE = ".keystore";
     public static final String DEFAULT_KEYSTORE_TYPE = KeyStore.getDefaultType();
-    public static final String DEFAULT_KEYMANAGER_ALGORITHM = KeyManagerFactory.getDefaultAlgorithm();
+    // Don't use default as in Java 7 it's SunX509 or implementation dependant, starting from Java 8 it's PKIX
```
afelisatti: Are you sure about this? While doing some debugging recently with Java 8 I kept getting SunX509 as default.

transports/ssl/src/main/resources/META-INF/mule-tls.xsd
```diff
@@ -137,23 +137,43 @@
...
-                        <xsd:element ref="trust-store" minOccurs="0" maxOccurs="1" />
-                        <xsd:element ref="key-store" minOccurs="1" maxOccurs="1" />
-                    </xsd:sequence>
-                    <xsd:sequence>
-                        <xsd:element ref="trust-store" minOccurs="1" maxOccurs="1" />
-                        <xsd:element ref="key-store" minOccurs="1" maxOccurs="1" />
-                    </xsd:sequence>
-                </xsd:choice>
+                <xsd:sequence minOccurs="1">
+                    <xsd:element ref="trust-store" minOccurs="0" maxOccurs="1" />
```
afelisatti: I'm not sure about this. The other config was to support having nothing, just one of each or all definitions. Please review this with @pablolagreca because there was a reason for having that separate.

transports/ssl/src/main/resources/META-INF/mule-tls.xsd
```diff
@@ -265,4 +283,42 @@
...
+            </xsd:annotation>
+        </xsd:attribute>
+    </xsd:complexType>
+
+    <xsd:complexType name="tlsContextCrlFileRevocationCheckType">
+        <xsd:attribute name="path" type="xsd:string">
+            <xsd:annotation>
+                <xsd:documentation>
+                    The path to a CRL (Certificate Revocation List) file to be used for this trust store.
+                    A certificate mentioned there will not be accepted for authentication.
```
afelisatti: What would "there" be? Need to make this clearer.

PR #4632, opened 2017-05-04
MULE-12385: Fix: Some endpoints allow to define a reconnection strategy

facundovs:
In mule 2.x, it was allowed to define reconnection strategies within endpoints.
In mule 3.x, Reconnection strategies only must be defined in connectors or globally.
However, currently, in some connectors, such as HTTP, FTP, AMQP, Mail, so on, it's still possible to define reconnection strategies within endpoints (in the xml configuration, but it doesn't work).
A possible solution is to modify the scheme.
This change only be applied to 3.x.

### DIFF
--- a/distributions/standalone/src/main/resources/MIGRATION.txt
+++ b/distributions/standalone/src/main/resources/MIGRATION.txt
@@ -16,7 +16,7 @@ MULE-10686: XML entity expansion in Jersey is now disabled by default because it
 MULE-10979: The default response timeout and default transaction timeout can't be configured using system properties on the command line or in the wrapper.conf file anymore. In replacement, use the configuration element. For example: <configuration defaultResponseTimeout="20000"  defaultTransactionTimeout="40000"/>.
 MULE-11118: The HTTP listener now replies with status code 503 when the thread pool is exhausted (and poolExhaustedAction="ABORT") instead of closing the socket.
 MULE-11825: In a DB template query, to set a DB param with the null value, you can use the "NULL" literal value. For example: <db:in-param name="name" defaultValue="NULL"/>
-MULE-12385: Reconnection Strategies can only be defined in connector components or globally (using <configuration> element). Defining reconnection strategies in endpoints isn't already supported.
+MULE-12385: Reconnection Strategies can only be defined in connector components or globally (using <configuration> element). In mule 3.x, defining reconnection strategies was being supported by the XSD, but ignored by Runtime. Now, the XSD was changed to not allow to use this invalid configuration.
 
 Migration changes from 3.7.x to 3.8.x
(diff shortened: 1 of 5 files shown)


distributions/standalone/src/main/resources/MIGRATION.txt
```diff
@@ -16,6 +16,7 @@ MULE-10686: XML entity expansion in Jersey is now disabled by default because it
 MULE-10979: The default response timeout and default transaction timeout can't be configured using system properties on the command line or in the wrapper.conf file anymore. In replacement, use the configuration element. For example: <configuration defaultResponseTimeout="20000"  defaultTransactionTimeout="40000"/>.
 MULE-11118: The HTTP listener now replies with status code 503 when the thread pool is exhausted (and poolExhaustedAction="ABORT") instead of closing the socket.
 MULE-11825: In a DB template query, to set a DB param with the null value, you can use the "NULL" literal value. For example: <db:in-param name="name" defaultValue="NULL"/>
+MULE-12385: Reconnection Strategies can only be defined in connector components or globally (using <configuration> element). Defining reconnection strategies in endpoints isn't already supported.
```
pablokraan: "Defining reconnection strategies in endpoints isn't already supported." I would rephrase this, to indicate that the XSD supported to configure it, but that it was not supported on runtime or that the configuration was not used

modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -3670,7 +3670,6 @@
     <xsd:group name="defaultEndpointElements">
         <xsd:sequence>
             <xsd:group ref="commonDefaultEndpointElements"/>
-            <xsd:element ref="abstract-reconnection-strategy" minOccurs="0"/>
```
pablokraan: Review this change with Studio guys

tests/integration/src/test/java/org/mule/test/reconnection/InvalidReconnectionStrategyTestCase.java
```diff
@@ -0,0 +1,31 @@
...
+package org.mule.test.reconnection;
+
+import org.junit.Test;
+import org.mule.config.spring.AbstractSchemaValidationTestCase;
+import org.xml.sax.SAXException;
+
+public class InvalidReconnectionStrategyTestCase extends AbstractSchemaValidationTestCase
+{
+
+    @Test(expected = SAXException.class)
```
pablokraan: This exception is too generic. Can you check the error message?

modules/spring-config/src/main/resources/META-INF/mule.xsd
facundovs: I was talking to Pablo Cabrera and he asks me to create a JIRA to notify these changes.

PR #4641, opened 2017-05-04
MULE-12398 - Fix test cases using maven artifact resolution - fix emb…

pablolagreca:
…edded test cases - take 1


embedded/embedded-api/src/main/java/org/mule/runtime/module/embedded/internal/MavenContainerClassLoaderFactory.java
```diff
@@ -8,6 +8,7 @@
 package org.mule.runtime.module.embedded.internal;
 
 import static java.util.stream.Collectors.toList;
+import static org.codehaus.plexus.util.FileUtils.toFile;
```
gsfernandes: Perhaps we could use https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FileUtils.html#toFile(java.net.URL) instead of plexus utility code here.

PR #4649, opened 2017-05-05
MULE-10582 NPE when an operation throws exception inside a foreach. A…

dfeist:
…dd test cases.


dfeist:
@afelisatti I couldn't reproduce the issue described in the JIRA, but have added tests and resolved another minor issue.

PR #4657, opened 2017-05-05
MULE-12298: Heavyweight deployment fails when no Maven repository is

anosenzo:


tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/Controller.java
```diff
@@ -69,19 +72,30 @@ public Controller(String muleHome, int timeout) {
...
     if (error != 0) {
       throw new MuleControllerException("The mule instance couldn't be started");
     }
   }
 
+  protected void checkRepositoryLocationAndUpdateInternalRepoPropertyIfPresent(String... args) {
+    Optional<String> repoVar =
+        Arrays.asList(args).stream().filter(arg -> arg.contains("-M-DmuleRuntimeConfig.maven.repositoryLocation=")).findFirst();
+    if (repoVar.isPresent()) {
+      this.internalRepository = new File(repoVar.get().split("=")[1]);
```
aiannucci: Why are you setting internalRepository variable with an external repository location?

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/Controller.java
```diff
@@ -69,19 +72,30 @@ public Controller(String muleHome, int timeout) {
...
+    checkRepositoryLocationAndUpdateInternalRepoPropertyIfPresent(args);
     int error = runSync("start", args);
     if (error != 0) {
       throw new MuleControllerException("The mule instance couldn't be started");
     }
   }
 
+  protected void checkRepositoryLocationAndUpdateInternalRepoPropertyIfPresent(String... args) {
+    Optional<String> repoVar =
+        Arrays.asList(args).stream().filter(arg -> arg.contains("-M-DmuleRuntimeConfig.maven.repositoryLocation=")).findFirst();
```
elrodro83: static import

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/Controller.java
```diff
@@ -260,7 +274,7 @@ protected File getArtifactInternalRepository(String artifactName) {
    * @return the directory of the internal repository for the Mule runtime.
    */
   protected File getRuntimeInternalRepository() {
-    return new File(muleHome, "repository");
+    return this.internalRepository;
```
aiannucci: This method is now returning something that could not be an internal repository

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/Controller.java
anosenzo: The property repositoryLocation is to use another path for the repo of the distro. So in place of use the repository folder inside the distro you can use for example the /user/.m2/repository folder.

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/Controller.java
anosenzo: Check the comment above.

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/Controller.java
aiannucci: Then the variable name should be something like *mavenRepository*, because it won't be always internal

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/Controller.java
anosenzo: done

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/Controller.java
anosenzo: I think that the internal is something to note that it is not the remote repositories. So maybe it has sense to keep it in the name. @gsfernandes What do you think??
Anyway a possible name without internal could be runtimeRepository to differentiate it from the app maven repository.

aiannucci: This is a variable of the *Controller* class, so it is pretty clear that it belongs to the runtime and not to any particular application. But calling it internal makes you think it will retrieve the value of the repository that lives inside the distribution

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/Controller.java
```diff
@@ -60,6 +63,7 @@
   protected File domainsDir;
   protected File appsDir;
   protected File libsDir;
+  protected File internalRepository;
```
aiannucci: internalRepository -> mavenRepository

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/Controller.java
anosenzo: but as I said in the previous comment the confusion could come between the runtimeRepository and the artifactRepository please check this two methods:

   /**
    * @param artifactName
    * @return the directory of the internal repository for the artifact with the given name.
    */
   protected File getArtifactInternalRepository(String artifactName) {
     return new File(new File(appsDir, artifactName), "repository");
   }
 
   /**
     * @return the directory of the internal repository for the Mule runtime.
     */
    protected File getRuntimeInternalRepository() {
 -    return new File(muleHome, "repository");
 +    return this.internalRepository;

PR #4663, opened 2017-05-05
MULE-11567: Resolve operation parameters during interception

elrodro83:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/ReactiveOperationExecutionWrapper.java
```diff
@@ -96,4 +100,10 @@ public void dispose() {
   public void setMuleContext(MuleContext muleContext) {
     this.muleContext = muleContext;
   }
+
+  @Override
+  public Function<ExecutionContext<OperationModel>, Map<String, Object>> createArgumentResolver(OperationModel operationModel) {
```
pablolagreca: Can you return Optional instead of the value since the method may return null?

PR #4665, opened 2017-05-08
MULE-12322 Flow should not continue processing in non-blocking proces…

dfeist:
…sor response thread. PENDING BUILD

### DIFF
--- a/core-tests/src/test/java/org/mule/runtime/core/processor/strategy/ProactorProcessingStrategyTestCase.java
+++ b/core-tests/src/test/java/org/mule/runtime/core/processor/strategy/ProactorProcessingStrategyTestCase.java
@@ -6,20 +6,23 @@
  */
 package org.mule.runtime.core.processor.strategy;
 
+import static java.util.Arrays.asList;
+import static org.hamcrest.CoreMatchers.allOf;
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.instanceOf;
+import static org.hamcrest.CoreMatchers.not;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.greaterThanOrEqualTo;
+import static org.hamcrest.Matchers.lessThanOrEqualTo;
+import static org.hamcrest.Matchers.startsWith;
+import static org.hamcrest.core.IsCollectionContaining.hasItem;
+import static org.junit.internal.matchers.ThrowableMessageMatcher.hasMessage;
 import static org.mule.runtime.core.processor.strategy.AbstractProcessingStrategy.TRANSACTIONAL_ERROR_MESSAGE;
 import static org.mule.runtime.core.processor.strategy.AbstractRingBufferProcessingStrategyFactory.DEFAULT_BUFFER_SIZE;
 import static org.mule.runtime.core.processor.strategy.AbstractRingBufferProcessingStrategyFactory.DEFAULT_SUBSCRIBER_COUNT;
 import static org.mule.runtime.core.processor.strategy.AbstractRingBufferProcessingStrategyFactory.DEFAULT_WAIT_STRATEGY;
 import static org.mule.test.allure.AllureConstants.ProcessingStrategiesFeature.PROCESSING_STRATEGIES;
 import static org.mule.test.allure.AllureConstants.ProcessingStrategiesFeature.ProcessingStrategiesStory.PROACTOR;
-import static java.util.Arrays.asList;
-import static org.hamcrest.CoreMatchers.allOf;
-import static org.hamcrest.CoreMatchers.equalTo;
-import static org.hamcrest.CoreMatchers.instanceOf;
-import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.greaterThanOrEqualTo;
-import static org.hamcrest.Matchers.lessThanOrEqualTo;
-import static org.junit.internal.matchers.ThrowableMessageMatcher.hasMessage;
 
 import org.mule.runtime.core.api.DefaultMuleException;
 import org.mule.runtime.core.api.MuleContext;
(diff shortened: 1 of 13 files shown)


core-tests/src/test/java/org/mule/runtime/core/processor/strategy/ReactorProcessingStrategyTestCase.java
```diff
@@ -151,14 +151,27 @@ public void tx() throws Exception {
...
+  @Description("When the ReactorProcessingStrategy is configured any async processing will be returned to CPU_LIGHT thread. "
+      + "This helps avoid deadlocks when there are reduced number of threads used by async processor.")
+  public void asyncCpuLightConcurrent() throws Exception {
+    super.asyncCpuLightConcurrent();
+    assertThat(threads.size(), between(2, 3));
+    assertThat(threads.stream().filter(name -> name.startsWith(CPU_LIGHT)).count(), between(2l, 3l));
     assertThat(threads.stream().filter(name -> name.startsWith(IO)).count(), equalTo(0l));
     assertThat(threads.stream().filter(name -> name.startsWith(CPU_INTENSIVE)).count(), equalTo(0l));
-    assertThat(threads.stream().filter(name -> name.startsWith(CUSTOM)).count(), equalTo(1l));
+    assertThat(threads.stream().filter(name -> name.startsWith(CUSTOM)).count(), equalTo(0l));
```
elrodro83: there are already many assertions like this, but you may use `not(IsCollectionContaining.hasItem(Matchers.startsWith(CUSTOM)))` instead of doing the stream/filter thing

core/src/main/java/org/mule/runtime/core/processor/strategy/WorkQueueProcessingStrategyFactory.java
```diff
@@ -100,6 +101,15 @@ public ReactiveProcessor onPipeline(ReactiveProcessor pipeline) {
     }
 
     @Override
+    public ReactiveProcessor onProcessor(ReactiveProcessor processor) {
+      if (processor.getProcessingType() == ASYNC) {
+        return publisher -> from(publisher).transform(processor).publishOn(fromExecutorService(ioScheduler));
```
elrodro83: according to the description of `ASYNC`, shouldn't the publishing be on a cpu light rather than io?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
```diff
@@ -303,7 +305,14 @@ protected ParameterValueResolver getParameterValueResolver() {
 
   @Override
   public ProcessingType getProcessingType() {
-    return asProcessingType(operationModel.getExecutionType());
+    ProcessingType processingType = asProcessingType(operationModel.getExecutionType());
+    if (processingType == CPU_LITE && !operationModel.isBlocking()) {
+      // If processing type is CPU_LITE and operation is non-blocking then use ASYNC processing type so that the Flow can return
```
elrodro83: Is this something we want to do? didn't we originally intend to avoid the thread change when going back to the flow after a cpu-light process?

core-tests/src/test/java/org/mule/runtime/core/processor/strategy/ProactorProcessingStrategyTestCase.java
```diff
@@ -63,8 +66,9 @@ public void singleCpuLight() throws Exception {
     super.singleCpuLight();
     assertThat(threads.size(), equalTo(1));
     assertThat(threads.stream().filter(name -> name.startsWith(CPU_LIGHT)).count(), equalTo(1l));
```
elrodro83: change asserts. sames as the line below, but without the `not`. Check other occurrences

core-tests/src/test/java/org/mule/runtime/core/processor/strategy/ProactorProcessingStrategyTestCase.java
```diff
@@ -74,8 +78,9 @@ public void singleCpuLightConcurrent() throws Exception {
     super.singleCpuLightConcurrent();
     assertThat(threads.size(), allOf(greaterThanOrEqualTo(1), lessThanOrEqualTo(2)));
```
elrodro83: use IsCollectionWithSize instead of asserting like this

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
```diff
@@ -307,9 +307,9 @@ protected ParameterValueResolver getParameterValueResolver() {
   public ProcessingType getProcessingType() {
     ProcessingType processingType = asProcessingType(operationModel.getExecutionType());
     if (processingType == CPU_LITE && !operationModel.isBlocking()) {
-      // If processing type is CPU_LITE and operation is non-blocking then use ASYNC processing type so that the Flow can return
+      // If processing type is CPU_LITE and operation is non-blocking then use CPU_LITE_ASYNC processing type so that the Flow can return
```
elrodro83: how about operations that are not IO? for instance, if a router were an operation

extensions/http/src/main/java/org/mule/extension/http/internal/request/HttpRequester.java
```diff
@@ -98,26 +98,24 @@ private ResponseHandler createResponseHandler(MuleContext muleContext,
 
       @Override
       public void onCompletion(HttpResponse response) {
-        //TODO: MULE-11310 - Make HTTP request scheduling fine grained
-        scheduler.execute(() -> {
```
elrodro83: do you still need the `scheduler` attribute in this class?

PR #4670, opened 2017-05-08
MULE-12363: Missing CDATA on import of ArtifactDeclaration

alegmarra:


modules/spring-config/src/main/java/org/mule/runtime/config/spring/dsl/declaration/DefaultXmlArtifactDeclarationLoader.java
```diff
@@ -450,11 +453,21 @@ private void declareTransform(ComponentModel model,
...
+      } else {
+        value = cdata(groupConfig.getTextContent());
+      }
+      objectBuilder.withParameter(SCRIPT, value);
     }
     return objectBuilder;
   }
 
+  private boolean isCData(ConfigLine config) {
+    return config.getCustomAttributes().get(IS_CDATA) != null;
```
marianogonzalez: sounds like get(String) should return Optional. Evaluate if this will become a backwards concern and is something that can be changed later

modules/spring-config/src/main/java/org/mule/runtime/config/spring/dsl/model/internal/InfrastructureElementModelDelegate.java
```diff
@@ -88,12 +89,12 @@ public void addParameter(String parameterName, ParameterValue value,
...
           @Override
-          public void visitSimpleValue(String value) {
-            parentConfig.withParameter(parameterName, value);
+          public void visitSimpleValue(ParameterSimpleValue text) {
+            parentConfig.withParameter(parameterName, text.getValue());
             parentElement.containing(DslElementModel.builder()
                 .withModel(parameterModel)
                 .withDsl(paramDsl)
-                .withValue(value)
+                .withValue(text.getValue()) //FIXME
```
marianogonzalez: fix what?

PR #4673, opened 2017-05-08
MULE-10923: Add ShutdownHook to MuleDeployment rule

anosenzo:


tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleDeployment.java
```diff
@@ -358,12 +363,26 @@ public String describeFailure() {
...
-        logger.info("Stopping Mule Server");
-        mule.stop();
-      }
+      stopMule();
       super.after();
     }
+    removeShutdownHooks();
+  }
+
+  protected void stopMule() {
```
aiannucci: Change to private

anosenzo: Done

PR #4674, opened 2017-05-08
MULE-12408 - Support EL split in Foreach - initial commit, making for…

pablolagreca:
…each work with string which content are iterable thinks


extensions/db/src/test/resources/integration/storedprocedure/stored-procedure-streaming-returning-resultsets-config.xml
```diff
@@ -10,14 +10,14 @@
             <db:sql>{ call getSplitTestRecords() }</db:sql>
         </db:stored-procedure>
 
-        <set-variable variableName="records" value="#[[]]"/>
+        <set-variable variableName="records" value="#[[]]" mimeType="application/java"/>
```
dfeist: Unrelated to this PR but this should be `mediaType` to match API.

core/src/main/java/org/mule/runtime/core/routing/ExpressionSplitterStrategy.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class ExpressionSplitterStrategy {
+
+  private ExpressionManager expressionManager;
+
+  public ExpressionSplitterStrategy(ExpressionManager expressionManager) {
+    this.expressionManager = expressionManager;
+  }
+
+  public List<?> splitMessage(Event event, String expression) {
```
dfeist: If this is a strategy implementation it should be implementing an interface. Should this be implementing SplittingStrategy?

core/src/main/java/org/mule/runtime/core/routing/ExpressionSplitterStrategy.java
```diff
@@ -0,0 +1,39 @@
...
+import org.mule.runtime.api.el.BindingContext;
+import org.mule.runtime.api.metadata.TypedValue;
+import org.mule.runtime.core.api.Event;
+import org.mule.runtime.core.api.el.ExpressionManager;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+
+public class ExpressionSplitterStrategy {
```
dfeist: Some names use `SplittingStrategy` and others user "SpliiterStrategy`

core-tests/src/test/java/org/mule/runtime/core/routing/ExpressionSplitterIteratorTestCase.java
```diff
@@ -38,41 +37,42 @@
...
+    assertListValues(values);
   }
 
-  private List<Integer> createListOfIntegers() {
-    List<Integer> integers = new ArrayList<>(3);
+  private List<TypedValue<?>> createListOfIntegers() {
+    List<TypedValue<?>> integers = new ArrayList<>(3);
     for (int i = 0; i < 3; i++) {
-      integers.add(i);
+      integers.add(new TypedValue<>(i, DataType.fromType(Integer.class)));
```
dfeist: Static

extensions/db/src/test/resources/integration/storedprocedure/stored-procedure-streaming-returning-resultsets-config.xml
pablolagreca: I think since this is going to end up in compatibility, it doesn't even worth to change it. WDYT?

dfeist: Maybe not, but as i said unrelated to this commit.

PR #4682, opened 2017-05-09
SE-712: flow initialState should support a property placeholder

GustavoARod:
* Updating mule XSD for changing the restriction base type of the initialState properties of the baseFlowConstructType and flowType
* This change was based on the selector property (mule.xsd:884) of the watermarkType which currently have the desired behavior

### DIFF
--- a/modules/spring-config/src/main/resources/META-INF/mule.xsd
+++ b/modules/spring-config/src/main/resources/META-INF/mule.xsd
@@ -2437,7 +2437,7 @@
                         </xsd:documentation>
                     </xsd:annotation>
                     <xsd:simpleType>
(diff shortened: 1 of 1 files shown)


pablokraan:
You must add a functional test case that test the change using a property place holder

dfeist:
Also commits to CE (this `mule` project) should be made with a `MULE` JIRA issue.

GustavoARod:
According to the comment's I'll be adding functional test and also creating the right Jira issue. Taking that into account it's easier for me to close this PR and create a new one with the suggestions already applied. 

Anyways thank you very much @pablokraan and @dfeist

PR #4689, opened 2017-05-09
MULE-12111: Soap Connect: give the capability to create use a custom transport

juandesi:


modules/extensions-soap-support/src/main/java/org/mule/runtime/module/extension/soap/internal/loader/SoapModelLoaderDelegate.java
```diff
@@ -57,14 +62,24 @@
    */
   public ExtensionDeclarer declare(ExtensionLoadingContext context) {
     final SoapExtensionTypeWrapper<?> extension = getSoapExtensionType(this.extensionType);
+    List<MessageDispatcherProviderTypeWrapper> customTransportProviders = extension.getDispatcherProviders();
     ExtensionDeclarer extensionDeclarer = getExtensionDeclarer(context);
+    declareSubtypes(extensionDeclarer, customTransportProviders);
```
alegmarra: do this in `SubTypesDeclarationEnricher`

PR #4690, opened 2017-05-09
MULE-12334: Deserialisation breaks for inner class

pablokraan:
_ Reverting the refactoring on DefaultComponentLocation
_ Fixing the bug
_ Using org.mule.test.integration.exceptions.ExceptionHandlingTestCase#testUntilSuccessfulInTransactionalScope to test that the fix works


modules/dsl-api/src/main/java/org/mule/runtime/dsl/api/component/config/DefaultComponentLocation.java
```diff
@@ -171,6 +171,90 @@ public DefaultComponentLocation appendRoutePart() {
         .add(new DefaultLocationPart("route", empty(), empty(), empty())).build());
   }
 
+  /**
+   * A location part represent an specific location of a component within another component.
+   * 
+   * @since 4.0
+   */
+  public static class DefaultLocationPart implements LocationPart, Serializable {
```
afelisatti: Why is the serialVersionUID no longer here?

pablokraan: good catch!

PR #4695, opened 2017-05-10
MULE-12429: Streaming buffer is wrongfully exceeded

marianogonzalez:


core/src/main/java/org/mule/runtime/core/internal/streaming/bytes/InMemoryStreamBuffer.java
```diff
@@ -127,36 +126,41 @@ public void doClose() {
...
    * @throws StreamingBufferSizeExceededException if the buffer is not big enough and cannot be expanded
    */
   @Override
   public int consumeForwardData() throws IOException {
-    if (!buffer.hasRemaining()) {
-      buffer = expandBuffer(buffer.capacity());
-    }
+    ByteBuffer readBuffer = buffer.hasRemaining()
+        ? buffer
+        : allocate(bufferSizeIncrement > 0 ? bufferSizeIncrement : 10);
```
afelisatti: Can you make the 10 a constant at least?

PR #4699, opened 2017-05-10
MULE-6619: flow initialState should support a property placeholder

GustavoARod:
* Updating mule XSD for allowing a property placeholder on the initialState property of the flowType elemen

### DIFF
--- a/tests/integration/pom.xml
+++ b/tests/integration/pom.xml
@@ -64,6 +64,8 @@
             <groupId>org.mule.modules</groupId>
             <artifactId>mule-module-db</artifactId>
             <version>${project.version}</version>
+            <type>test-jar</type>
+            <scope>test</scope>
         </dependency>
         <!-- Include all Mule Transports -->
         <dependency>
--- a/tests/integration/pom.xml
+++ b/tests/integration/pom.xml
@@ -60,13 +60,6 @@
             <type>pom</type>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.mule.modules</groupId>
-            <artifactId>mule-module-db</artifactId>
-            <version>${project.version}</version>
-            <type>test-jar</type>
-            <scope>test</scope>
-        </dependency>
         <!-- Include all Mule Transports -->
         <dependency>
             <groupId>org.mule.transports</groupId>
--- a/tests/integration/src/test/java/org/mule/test/components/InvalidFlowStatePlaceholderTestCase.java
+++ b/tests/integration/src/test/java/org/mule/test/components/InvalidFlowStatePlaceholderTestCase.java
@@ -6,53 +6,23 @@
  */
 package org.mule.test.components;
 
-import static org.hamcrest.CoreMatchers.containsString;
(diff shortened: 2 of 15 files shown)


tests/integration/src/test/java/org/mule/test/components/InvalidFlowStatePlaceholderTestCase.java
```diff
@@ -0,0 +1,58 @@
...
+import org.mule.context.DefaultMuleContextFactory;
+import org.mule.tck.junit4.AbstractMuleTestCase;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+public class InvalidFlowStatePlaceholderTestCase extends AbstractMuleTestCase
```
pablokraan: Extend from AbstractConfigurationErrorTestCase instead. Check the examples in the code

tests/integration/src/test/resources/org/mule/test/components/invalid-flow-initial-state.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<mule xmlns="http://www.mulesoft.org/schema/mule/core" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:spring="http://www.springframework.org/schema/beans"
+	xsi:schemaLocation="
+       http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-3.0.xsd
+       http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd">
+	<flow name="invalidPlaceholderStateFlow" initialState="${state}]">
+		<inbound-endpoint address="test://in.placeholder" />
```
pablokraan: Is the endpoint really needed?

tests/integration/src/test/java/org/mule/test/components/InvalidFlowStatePlaceholderTestCase.java
GustavoARod: Hey Pablo, I realized that the `AbstractConfigurationErrorTestCase` is under the `mule-module-db` which is not a dependency of the `mule-test-integration`yet. I'm not quite sure if it worths or makes sense to add that dependency for the purpose we're trying to achieve.

tests/integration/src/test/resources/org/mule/test/components/invalid-flow-initial-state.xml
GustavoARod: No, It's not

tests/integration/pom.xml
```diff
@@ -64,6 +64,8 @@
             <groupId>org.mule.modules</groupId>
             <artifactId>mule-module-db</artifactId>
             <version>${project.version}</version>
+            <type>test-jar</type>
```
pablokraan: Instead of adding this dependency, move the AbstractConfigurationErrorTestCase class to the test functional module inside the org.mule.tck package

PR #4707, opened 2017-05-10
MULE-12436, expose package org.mule.runtime.core.internal.construct

damiansima:


core/src/main/resources/META-INF/mule-module.properties
```diff
@@ -204,6 +204,7 @@ artifact.export.classPackages=org.mule.runtime.core,\
                               javax.xml.stream.events, \
                               javax.xml.parsers, \
                               javax.xml.transform,\
+                              org.mule.runtime.core.internal.construct,\
```
pablokraan: This must be in the privileged packages

PR #4710, opened 2017-05-10
MULE-12433: CursorStream should throw IOException

marianogonzalez:


core/src/main/java/org/mule/runtime/core/internal/streaming/bytes/AbstractCursorStream.java
```diff
@@ -76,8 +75,10 @@ public CursorStreamProvider getProvider() {
 
   protected abstract void doRelease();
 
-  protected void assertNotDisposed() {
-    checkState(!released, "Stream is closed");
+  protected void assertNotDisposed() throws IOException {
```
dfeist: No test case/method?

marianogonzalez: i think it's fine with just having the existing test not breaking

marianogonzalez: added a test just because I love to make you happy

core-tests/src/test/java/org/mule/runtime/core/internal/streaming/bytes/CursorStreamProviderTestCase.java
```diff
@@ -275,6 +275,13 @@ public void dataLengthMatchesMaxBufferSizeExactly() throws Exception {
     withCursor(cursor -> assertThat(IOUtils.toString(cursor), equalTo(data)));
   }
 
+  @Test(expected = IOException.class)
```
dfeist: Should really use ExpectedException here.

PR #4732, opened 2017-05-12
MULE-12459: OAuth extension fails if it doesn't provide default set o…

marianogonzalez:
…f scopes

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/oauth/DefaultExtensionsOAuthManager.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/oauth/DefaultExtensionsOAuthManager.java
@@ -11,6 +11,7 @@ import static java.util.Arrays.asList;
 import static java.util.Optional.empty;
 import static java.util.Optional.of;
 import static java.util.stream.Collectors.toMap;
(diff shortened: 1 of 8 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/oauth/DefaultExtensionsOAuthManager.java
```diff
@@ -192,7 +193,13 @@ private AuthorizationCodeOAuthDancer createDancer(OAuthConfig config) throws Mul
         .responseExpiresInExpr(grantType.getExpirationRegex())
         .responseRefreshTokenExpr(grantType.getRefreshTokenExpr())
         .responseAccessTokenExpr(grantType.getAccessTokenExpr());
-    dancerBuilder.scopes(authCodeConfig.getScope().orElse(grantType.getDefaultScope()));
+
+    String scopes = authCodeConfig.getScope()
+        .orElseGet(() -> isBlank(grantType.getDefaultScope()) ? null : grantType.getDefaultScope());
+
+    if (scopes != null) {
```
alegmarra: `String scopes = authCodeConfig.getScope().orElseGet(() -> grantType.getDefaultScope());`
`if (!isBlank(scopes)) {`

PR #4735, opened 2017-05-12
MULE-11538: Add DataType for Map objects

afelisatti:


core-tests/src/test/java/org/mule/runtime/core/transformer/types/DataTypeBuilderTestCase.java
```diff
@@ -149,6 +171,52 @@ public void templateTypedCollection() {
   }
 
   @Test
+  public void templateMap() {
+    final DataType template = DataType.builder().type(HashMap.class).mediaType("text/plain;charset=ASCII").build();
```
elrodro83: add cases where the media type is set on the key and the values

PR #4742, opened 2017-05-15
MULE-12434: Poll extension model is invalid

pablomusumeci:


modules/schedulers/src/main/resources/META-INF/mule-schedulers-extension-model.json
```diff
@@ -0,0 +1,134 @@
+{
+  "name": "Mule Schedulers",
```
marianogonzalez: I am willing to accept that there's a need for this to exist in json format, but the source of it has to be a Declarer. This is impossible to maintain otherwise

PR #4746, opened 2017-05-15
MULE-11603: Add DW performance tests

LucianoGandini:
Also added elastic-search plugin to upload jmh results.


tests/performance/pom.xml
```diff
@@ -150,6 +151,76 @@
             </build>
         </profile>
         <profile>
+            <id>micro-benchmark</id>
+            <build>
+                <plugins>
+                    <plugin>
```
dfeist: Why is this needed? Are we running JMH tests and also running benchmark tests cases with same profile?

tests/performance/pom.xml
```diff
@@ -150,6 +151,76 @@
...
+                                <goals>
+                                    <goal>exec</goal>
+                                </goals>
+                                <configuration>
+                                    <classpathScope>test</classpathScope>
+                                    <executable>java</executable>
+                                    <workingDirectory>${basedir}/target</workingDirectory>
+                                    <arguments>
+                                        <argument>-classpath</argument>
+                                        <classpath/>
```
dfeist: Can this line be removed?

tests/performance/pom.xml
```diff
@@ -150,6 +151,76 @@
...
+                                        <mainClass>org.openjdk.jmh.Main</mainClass>
+                                        <!--Create json report-->
+                                        <argument>-rf</argument>
+                                        <argument>json</argument>
+                                        <!--Enable gc profiler-->
+                                        <argument>-prof</argument>
+                                        <argument>gc</argument>
+                                    </arguments>
+                                    <!--<commandlineArgs>-rf json -prof gc</commandlineArgs>-->
```
dfeist: Remove?

tests/performance/src/main/java/org/mule/el/ExpressionBenchmark.java
```diff
@@ -0,0 +1,75 @@
...
+import org.mule.runtime.api.exception.MuleException;
+import org.mule.runtime.core.api.Event;
+import org.mule.runtime.core.api.EventContext;
+import org.mule.runtime.core.api.MuleContext;
+import org.mule.runtime.core.api.scheduler.SchedulerService;
+
+import org.openjdk.jmh.annotations.*;
+
+@OutputTimeUnit(NANOSECONDS)
+@Warmup(iterations = 1)
```
dfeist: Why only 1 warmup iteration. Wouldn't it be best to define this is pom.xml?

tests/performance/src/test/java/org/mule/AbstractBenchmarkAssertionTestCase.java
```diff
@@ -116,7 +116,7 @@ protected void runAndAssertBenchmark(Class clazz, String testName, int threads,
   protected void runAndAssertBenchmark(Class clazz, String testName, int threads, Map<String, String[]> params, TimeUnit timeUnit,
                                        boolean profileGC, Consumer<RunResult> assertions) {
     try {
-      if (getProperty(ENABLE_PERFORMANCE_TESTS_SYSTEM_PROPERTY) != null) {
+      if (Boolean.getBoolean(getProperty(ENABLE_PERFORMANCE_TESTS_SYSTEM_PROPERTY))) {
```
dfeist: Static import.

tests/performance/pom.xml
```diff
@@ -150,6 +151,76 @@
...
+            <id>micro-benchmark</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <systemPropertyVariables>
+                                <enablePerformanceTests>false</enablePerformanceTests>
+                                <contiperf.active>false</contiperf.active>
```
dfeist: Contiperf still used?

LucianoGandini: I just replicate the existing profiles, if it was removed, we should delete it.

tests/performance/pom.xml
LucianoGandini: Its a new profile to run the jmh tests as it was from the compiled jar and then upload the output to elastisearch.

tests/performance/pom.xml
LucianoGandini: No, this injects the classpath that the exec plugin needs to run the java class.

tests/performance/pom.xml
LucianoGandini: will do

PR #4754, opened 2017-05-16
MULE-11615: Change File Commons Version

estebanwasinger:


distributions/standalone/pom.xml
```diff
@@ -184,7 +184,7 @@
         <dependency>
```
marianogonzalez: verify if this is necessary at all. Doesn't seem so

PR #4764, opened 2017-05-17
MULE-10056: Maven module plugin is not properly detecting transitive …

pablokraan:
…exported dependencies

_ Removing un-existent packages
_ Adding missing exported packages
_ Adding optional packages
_ Removing non-required dependency from UUID artifact

### DIFF
--- a/modules/extensions-spring-support/src/main/resources/META-INF/mule-module.properties
+++ b/modules/extensions-spring-support/src/main/resources/META-INF/mule-module.properties
@@ -4,6 +4,8 @@ artifact.export.classPackages=org.mule.runtime.module.extension.internal.capabil
                               org.mule.runtime.module.extension.internal.capability.xml.schema.builder,\
                               org.mule.runtime.module.extension.internal.capability.xml.schema.model,\
                               org.mule.runtime.module.extension.internal.config,\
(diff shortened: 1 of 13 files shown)


modules/extensions-spring-support/src/main/resources/META-INF/mule-module.properties
```diff
@@ -4,14 +4,20 @@ artifact.export.classPackages=org.mule.runtime.module.extension.internal.capabil
...
+                              org.mule.runtime.module.extension.internal.runtime.operation,\
                               org.mule.runtime.module.extension.internal.resources,\
                               javax.annotation.processing,\
                               javax.xml.bind.annotation,\
                               javax.xml.bind.annotation.adapters,\
                               javax.lang.model,\
                               javax.lang.model.element,\
                               javax.xml.bind,\
-                              org.mule.runtime.module.extension.internal.runtime.operation
+                              reactor.core,\
```
elrodro83: do we have an issue to track these packages that shouldn't be exported and clean them up at some point?

pablokraan: nope. That is part of the BIG api clean up task. I'll start to search and track those packages soon

PR #4787, opened 2017-05-19
MULE-12402: until-successful retries fail after application restart d…

pablokraan:
…ue to inner flow still stopped


core/src/main/java/org/mule/runtime/core/routing/UntilSuccessful.java
```diff
@@ -91,12 +91,13 @@ public void initialise() throws InitialisationException {
...
 
-    if (untilSuccessfulStrategy instanceof Initialisable) {
-      ((Initialisable) untilSuccessfulStrategy).initialise();
-    }
     if (untilSuccessfulStrategy instanceof MuleContextAware) {
       ((MuleContextAware) untilSuccessfulStrategy).setMuleContext(muleContext);
     }
+
+    if (untilSuccessfulStrategy instanceof Initialisable) {
+      ((Initialisable) untilSuccessfulStrategy).initialise();
```
elrodro83: why not call `LifecycleUtils#initializeIfNeeded()`? It has a method that receives the muleContext to set.

pablokraan: Just moved the code, but I can replace that

PR #4793, opened 2017-05-19
MULE-12528: Race condition when stopping an ExtensionSource while it's

elrodro83:
being recconnected


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -179,9 +183,16 @@ private void notifyExceptionAndShutDown(Throwable exception) {
   }
 
   private void restart() throws MuleException {
-    stopSource();
-    disposeSource();
-    startSource();
+    if (started.get()) {
+      synchronized (started) {
```
marianogonzalez: synchronize before the if

PR #4797, opened 2017-05-22
MULE-12323 - Review wrapped mule exceptions and error handling

pablolagreca:


core-tests/src/test/java/org/mule/runtime/core/execution/ExceptionToMessagingExceptionExecutionInterceptorTestCase.java
```diff
@@ -54,23 +54,29 @@
   private MuleException mockMuleException;
   @Mock
   private ErrorTypeLocator mockErrorTypeLocator;
+  @Mock
+  private org.mule.runtime.core.exception.ErrorTypeRepository mockErrorTypeRepository;
```
afelisatti: Is the static import possible?

core/src/main/java/org/mule/runtime/core/util/ExceptionUtils.java
```diff
@@ -255,6 +265,87 @@ public static Event createErrorEvent(Event currentEvent, Object annotatedObject,
...
+   * @return the found exception or empty.
+   */
+  private static Optional<Exception> findRootExceptionForErrorHandling(Exception exception, ErrorTypeLocator errorTypeLocator,
+                                                                       ErrorTypeRepository errorTypeRepository) {
+    List<Throwable> causesAsList = getExceptionsAsList(exception);
+    for (Throwable cause : causesAsList) {
+      if (cause instanceof TypedException) {
+        return of((Exception) cause);
+      }
+      if (cause instanceof MuleException || cause instanceof MuleRuntimeException) {
```
afelisatti: Does this mean that the first MuleException/MuleRuntimeException will be used? What if there are several levels of these?

core/src/main/java/org/mule/runtime/core/util/ExceptionUtils.java
```diff
@@ -255,6 +265,87 @@ public static Event createErrorEvent(Event currentEvent, Object annotatedObject,
...
+        ErrorType errorType = errorTypeLocator.lookupErrorType(cause);
+        if (!errorType.equals(errorTypeRepository.getErrorType(buildFromStringRepresentation(UNKNOWN_ERROR_IDENTIFIER)))) {
+          return of((Exception) cause);
+        }
+      }
+    }
+    Optional<MessagingException> messagingExceptionOptional = getDeepestExceptionType(MessagingException.class, exception);
```
afelisatti: I'm not sure why the order here is not inverted, shouldn't we check for an existing error first and then try to resolve it?

PR #4803, opened 2017-05-22
MULE-12538: Configure Garbage Collection parameters for JDK 8 in Mule 4

LucianoGandini:


dfeist:
We also need to adjust heap size to match min requirements.

PR #4807, opened 2017-05-22
MULE-12473: email config elements should have the -config suffix

pablomusumeci:


modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/streaming/BytesStreamingExtensionTestCase.java
```diff
@@ -203,7 +203,7 @@ private ParameterModel getStreamingStrategyParameterModel(Supplier<Parameterized
 
   private ConfigurationModel getConfigurationModel() {
     return getExtensionModel("Marvel")
-        .map(extension -> extension.getConfigurationModel("dr-strange").get())
+        .map(extension -> extension.getConfigurationModel("dr-strange-config").get())
```
marianogonzalez: doesn't this mean that you should change the marvel extension test json as well?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/ConfigModelLoaderDelegate.java
```diff
@@ -31,6 +31,7 @@
 final class ConfigModelLoaderDelegate extends AbstractModelLoaderDelegate {
 
   private static final String CONFIGURATION = "Configuration";
+  private static final String CUSTOM_CONFIG_SUFFIX = "-" + DEFAULT_CONFIG_NAME;
```
marianogonzalez: do not tie the suffix to the default config name. simply use the `-config` literal

PR #4850, opened 2017-05-26
MULE-11982: Boot packages needs to be removed and modules must explic…

pablokraan:
…itly define what they export

_ Embedded should use host application classloader instead of the JDK-only classloader


distributions/embedded/embedded-api/src/main/java/org/mule/runtime/module/embedded/internal/DefaultEmbeddedContainerBuilder.java
```diff
@@ -62,7 +62,7 @@
 
   @Override
   public EmbeddedContainer.EmbeddedContainerBuilder withApplicationConfiguration(ApplicationConfiguration applicationConfigruation) {
```
elrodro83: the param has the same typo

PR #4861, opened 2017-05-27
MULE-12453: NPE on lifecycle stop phase after deployment error

elrodro83:
* Better fix
* Fix for another case also

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/management/stats/DefaultProcessingTimeWatcher.java
+++ b/core/src/main/java/org/mule/runtime/core/management/stats/DefaultProcessingTimeWatcher.java
@@ -18,6 +18,7 @@ import java.lang.ref.ReferenceQueue;
 import java.lang.ref.WeakReference;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Future;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mule/runtime/core/management/stats/DefaultProcessingTimeWatcher.java
```diff
@@ -58,12 +63,14 @@ public void setMuleContext(MuleContext muleContext) {
     @Override
     public void run() {
       try {
-        ProcessingTimeReference ref = (ProcessingTimeReference) queue.remove();
-        refs.remove(ref);
-
-        FlowConstructStatistics stats = ref.getStatistics();
-        if (stats.isEnabled()) {
-          stats.addCompleteFlowExecutionTime(ref.getAccumulator().longValue());
+        while (true) {
```
pablolagreca: If you depending on some part of your internal code to throw InterruptedException, wouldn't be better to check for interrput explicilty? What if the code you are calling at some point changes in a way that won't throw InterruptedException? For instance, it does not use any synchornization mechanism anymore

elrodro83: Agree, We should change this in 3.x too

PR #4865, opened 2017-05-29
MULE-12617: Prefetch size cannot be set for activemq-connector to guarantee order

fsgonz:


transports/jms/src/main/java/org/mule/transport/jms/activemq/ActiveMQJmsConnector.java
```diff
@@ -53,6 +53,15 @@ protected void applyVendorSpecificConnectionFactoryProperties(ConnectionFactory
     {
         try
         {
+            Method getPrefetchPolicy = connectionFactory.getClass().getMethod("getPrefetchPolicy");
```
elrodro83: why do you need to do this reflection stuff? this method is public (http://activemq.apache.org/maven/5.13.4/apidocs/org/apache/activemq/ActiveMQConnectionFactory.html#getPrefetchPolicy()) and this class is already coupled to ActiveMQ

fsgonz: I need to use reflection because ConnectionFactory is the javax class. it's the same reason why I must use reflection to get getRedeliveryPolicy.

fsgonz: It can be cast as you told me. I'll change it.

fsgonz: Actually, what happens is that activemq dependency is available only for testing. ActiveMQ has to be added by the user to their apps but is not included in the mule distribution.

PR #4868, opened 2017-05-29
MULE-12595: HTTP transport endpoint >30% performance drop on Mule 3.8.5.

facundovs:
Syncronized block is limited to only HTTP Multipart Requests.

### DIFF
--- a/transports/http/src/main/java/org/mule/transport/http/HttpMuleMessageFactory.java
+++ b/transports/http/src/main/java/org/mule/transport/http/HttpMuleMessageFactory.java
@@ -144,7 +144,10 @@ public class HttpMuleMessageFactory extends AbstractMuleMessageFactory
             httpVersion = httpRequest.getRequestLine().getHttpVersion();
             uri = httpRequest.getRequestLine().getUri();
             headers = convertHeadersToMap(httpRequest.getHeaders(), uri);
-            convertMultiPartHeaders(headers);
+            if (isHttpMultipartRequest(httpRequest))
+            {
+                convertMultiPartHeaders(headers);
+            }
         }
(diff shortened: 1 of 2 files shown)


transports/http/src/main/java/org/mule/transport/http/HttpMultipartMuleMessageFactory.java
```diff
@@ -25,8 +25,16 @@
     private Collection<Part> parts;
 
     @Override
-    protected synchronized MuleMessage doCreate(Object transportMessage, MuleMessage previousMessage, String encoding, MuleContext muleContext) throws Exception
+    protected MuleMessage doCreate(Object transportMessage, MuleMessage previousMessage, String encoding, MuleContext muleContext) throws Exception
     {
+        if (transportMessage instanceof HttpRequest && isHttpMultipartRequest((HttpRequest) transportMessage))
```
afelisatti: What happens if this does not apply?

afelisatti: Never mind. The last return was not visible.

transports/http/src/main/java/org/mule/transport/http/HttpMuleMessageFactory.java
```diff
@@ -455,4 +458,9 @@ public void setExchangePattern(MessageExchangePattern mep)
     {
         exchangePattern = mep;
     }
+
+    protected boolean isHttpMultipartRequest (HttpRequest httpRequest)
+    {
+        return httpRequest.getContentType().contains("multipart/form-data");
```
afelisatti: Are you sure only form-data applies here? What about mixed?

PR #4874, opened 2017-05-29
MULE-12583: Fix flaky DataTypeBuilderTestCase.cacheClean

elrodro83:
Also fixes flakies of expiration and processing strategy by increasing the timeouts

### DIFF
--- a/core-tests/src/test/java/org/mule/runtime/core/transformer/types/DataTypeBuilderTestCase.java
+++ b/core-tests/src/test/java/org/mule/runtime/core/transformer/types/DataTypeBuilderTestCase.java
@@ -24,6 +24,7 @@ import static org.mule.runtime.api.metadata.DataType.NUMBER;
 import static org.mule.runtime.api.metadata.DataType.OBJECT;
 import static org.mule.runtime.api.metadata.DataType.STRING;
 import static org.mule.runtime.core.util.IOUtils.toByteArray;
+
 import org.mule.runtime.api.el.BindingContext;
 import org.mule.runtime.api.el.ExpressionFunction;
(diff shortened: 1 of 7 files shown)


core-tests/src/test/java/org/mule/runtime/core/transformer/types/DataTypeBuilderTestCase.java
```diff
@@ -309,7 +314,7 @@ public void cacheClean() throws InterruptedException, ClassNotFoundException {
     DataType.builder().type(custom.loadClass(Message.class.getName())).build();
     custom = null;
 
-    new PollingProber().check(new JUnitLambdaProbe(() -> {
+    new PollingProber(10000, 100).check(new JUnitLambdaProbe(() -> {
```
dfeist: Can constants be used here? Are there any existing contasts in abstract test cases?

tests/unit/src/main/java/org/mule/tck/junit4/AbstractMuleContextTestCase.java
```diff
@@ -127,7 +127,7 @@
   /**
    * Default timeout used when blocking on {@link org.reactivestreams.Publisher} completion.
    */
-  public static final int BLOCK_TIMEOUT = 200;
+  public static final int BLOCK_TIMEOUT = 500;
```
dfeist: This shouldn't need to be so high, at least not where I've used it most. Do we know 500ms is needed on build server?

elrodro83: i saw some failures caused because of this, yes

PR #4876, opened 2017-05-29
MULE-11982: Boot packages needs to be removed and modules must explic…

pablokraan:
…itly define what they export

_ Updating test to ensure that all the dependencies are read from the embedded container classloader
_ Removing dead code from EmbeddedController (services are not zips anymore)

### DIFF
--- a/distributions/embedded/embedded-impl/src/main/java/org/mule/runtime/module/embedded/impl/EmbeddedController.java
+++ b/distributions/embedded/embedded-impl/src/main/java/org/mule/runtime/module/embedded/impl/EmbeddedController.java
@@ -32,7 +32,6 @@ import java.io.File;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URL;
-import java.net.URLClassLoader;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


distributions/embedded/embedded-impl/src/main/java/org/mule/runtime/module/embedded/impl/EmbeddedController.java
```diff
@@ -187,18 +186,6 @@ private void setUpEnvironment() {
     setProperty(MULE_HOME_DIRECTORY_PROPERTY, containerInfo.getContainerBaseFolder().getPath());
     getDomainsFolder().mkdirs();
     getServicesFolder().mkdirs();
-    URLClassLoader urlClassLoader = (URLClassLoader) Thread.currentThread().getContextClassLoader();
-    URL[] urLs = urlClassLoader.getURLs();
-    for (URL urL : urLs) {
-      File dependencyFile = new File(urL.getPath());
-      if (dependencyFile.getName().toLowerCase().endsWith("zip")) {
```
elrodro83: @gsfernandes what kind of sorcery is embedded doing to load the services?

PR #4878, opened 2017-05-30
MULE-10646: Adapt response buffer to 3.8.5 by making it a system property

afelisatti:


modules/http/src/test/java/org/mule/module/http/functional/requester/HttpRequestConfigTestCase.java
```diff
@@ -40,7 +37,6 @@ protected String getConfigFile()
   public void verifyConfig() throws Exception
   {
     DefaultHttpRequesterConfig requestConfig = muleContext.getRegistry().lookupObject("requestConfig");
-    assertThat(requestConfig.getResponseBufferSize(), equalTo(RESPONSE_BUFFER));
```
pablolagreca: I don't get how you are testing the system property

afelisatti: You are right, I'll change the test instead of removing that part. The streamResponse property tests depend on the buffer size so those are using it but only in the default way.

PR #4898, opened 2017-05-31
Mule 12653: Reduce performance build time and PS iterations.

LucianoGandini:
Changes added makes performance builds successful (https://muleesb.ci.cloudbees.com/view/Mule%204/job/Mule-CE-Performance-Runner/31/).

### DIFF
--- a/tests/performance/src/main/java/org/mule/FlowNullProcessorBenchmark.java
+++ b/tests/performance/src/main/java/org/mule/FlowNullProcessorBenchmark.java
@@ -25,7 +25,7 @@ public class FlowNullProcessorBenchmark extends AbstractFlowBenchmark {
 
   @Override
   protected int getStreamIterations() {
-    return 100;
+    return 1000;
   }
 
   @Benchmark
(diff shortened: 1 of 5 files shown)


tests/performance/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,29 @@
...
+        <AsyncLogger name="org.apache" level="ERROR"/>
+
+        <!-- Reduce startup noise -->
+        <AsyncLogger name="org.springframework.beans.factory" level="WARN"/>
+
+        <!-- Mule classes -->
+        <AsyncLogger name="org.mule" level="WARN"/>
+        <AsyncLogger name="com.mulesoft" level="WARN"/>
+
+        <AsyncRoot level="INFO">
```
dfeist: Do we need info?

tests/performance/src/main/java/org/mule/FlowNullProcessorBenchmark.java
```diff
@@ -25,7 +25,7 @@
 
   @Override
   protected int getStreamIterations() {
-    return 1000;
+    return 100;
```
dfeist: Leave iterations at 1000 if problem strategy is commented out anyway

LucianoGandini: Result numbers were really high and you mentioned that JMH has issues with it.

tests/performance/src/main/resources/log4j2.xml
LucianoGandini: It replicates test configuration.

PR #4902, opened 2017-06-01
MULE-12656: Descriptions are not populated in the extension model

juandesi:
No tests has been added since the current ones still apply, the problem was that the test configuration was named "anotherConfig" which ends up hyphenized as `another-config`, this way the documenter could always find the config because the `-config` suffix was not added by the SDK, the config has been renamed to just `another` so the SDK will add the suffix when creating the extension making the test plausible

### DIFF
--- a/modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/capability/xml/TestAnotherDocumentedConfig.java
+++ b/modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/capability/xml/TestAnotherDocumentedConfig.java
@@ -14,7 +14,7 @@ import org.mule.runtime.extension.api.annotation.param.Parameter;
 /**
  * This is some Another Config documentation.
  */
-@Configuration(name = "anotherConfig")
+@Configuration(name = "another")
 @ConnectionProviders(TestAnotherDocumentedProvider.class)
 @Operations({TestDocumentedExtensionOperations.class})
 public class TestAnotherDocumentedConfig {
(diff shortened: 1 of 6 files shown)


modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/capability/xml/TestAnotherDocumentedConfig.java
```diff
@@ -14,7 +14,7 @@
 /**
  * This is some Another Config documentation.
  */
-@Configuration(name = "anotherConfig")
+@Configuration(name = "another")
```
marianogonzalez: I don't understand how this is a fix. ConfigLoaderDelegate only appends the -config if the name doesn't already end in config

modules/extensions-spring-support/src/test/resources/META-INF/documentation-extension-descriptions.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <extension-documentation>
-    <elements name="anotherConfig">
+    <elements name="another-config">
```
marianogonzalez: does this name match the name on the model or the name in the DSL? it should be the first

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/capability/xml/TestAnotherDocumentedConfig.java
marianogonzalez: it also seems to me that part of the problem is that ConfigLoaderDelegate is appending `-config` instead of appending `Config`

modules/extensions-spring-support/src/test/resources/META-INF/documentation-extension-descriptions.xml
juandesi: the name in the model

PR #4903, opened 2017-06-01
MULE-12663: In Message Filter, When unaccepted events are handled, execution of flow caller is not stopped.

facundovs:

### DIFF
--- a/transports/rmi/src/test/java/org/mule/transport/AbstractFunctionalTestCase.java
+++ b/transports/rmi/src/test/java/org/mule/transport/AbstractFunctionalTestCase.java
@@ -10,9 +10,12 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
+import static org.mule.tck.AbstractServiceAndFlowTestCase.ConfigVariant.SERVICE;
 
+import org.mule.api.MuleEventContext;
 import org.mule.api.MuleMessage;
 import org.mule.api.client.MuleClient;
+import org.mule.api.lifecycle.Callable;
 import org.mule.api.transport.DispatchException;
 import org.mule.config.i18n.Message;
 import org.mule.tck.AbstractServiceAndFlowTestCase;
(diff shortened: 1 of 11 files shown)


transports/rmi/src/test/java/org/mule/transport/issues/RmiMethodTypeMule1857TestCase.java
```diff
@@ -31,3 +31,4 @@ public RmiMethodTypeMule1857TestCase(ConfigVariant variant, String configResourc
     }      
     
 }
+//
```
dfeist: revert

transports/rmi/src/test/resources/jnp-functional-test-flow.xml
```diff
@@ -43,6 +43,7 @@
     <sub-flow name="Sender2SubFlow">
         <payload-type-filter expectedType="java.lang.String"/>
         <outbound-endpoint ref="Sender2"/>
+        <component class="org.mule.transport.AbstractFunctionalTestCase$TestJavaComponent"/>
```
dfeist: Is this for same issue, it wasn't in 3.8 PR

transports/rmi/src/test/java/org/mule/transport/AbstractFunctionalTestCase.java
```diff
@@ -46,13 +52,28 @@ public void testCase() throws Exception
 
         // send String
         message = client.send("vm://testin", "test matching component first time", null);
-        assertNotNull(message);
-        assertEquals(message.getPayload(), "emit tsrif tnenopmoc gnihctam tset");
+        if(variant.equals(SERVICE))
```
dfeist: Is this for same issue, it wasn't in 3.8 PR

facundovs: Yes It is. 
As in 3.x the default behavior changes, I had to modify some tests.

transports/rmi/src/test/resources/jnp-functional-test-flow.xml
facundovs: Yes It is.
As in 3.x the default behavior changes, I had to modify some tests.

PR #4910, opened 2017-06-01
MULE-12667: When the replyTo consumer in JMSMessageDispatcher is created the correlation id is not appropriately resolved

fsgonz:


transports/jms/src/main/java/org/mule/transport/jms/JmsMessageDispatcher.java
```diff
@@ -605,6 +605,14 @@ protected MessageConsumer createReplyToConsumer(Message jmsMessage, MuleEvent ev
                                                         connector.isNoLocal(), null, topic, endpoint);
     }
 
+    protected String encodeVendorSpecificID(String jmsCorrelationID)
+    {
+        // By default, no encoding is performed. In case an encoding is
```
pablokraan: Put this as the method javadoc

PR #4914, opened 2017-06-02
MULE-12656: Descriptions are not populated in the extension model

juandesi:


modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/capability/xml/TestAnotherDocumentedConfig.java
```diff
@@ -14,7 +14,7 @@
 /**
  * This is some Another Config documentation.
  */
-@Configuration(name = "anotherConfig")
+@Configuration(name = "another")
```
alegmarra: we should leave at least one case that tests for `another-config-config` not happening

modules/extensions-spring-support/src/test/resources/models/ram.json
```diff
@@ -365,6 +365,56 @@
                     },
                     {
                       "key": {
+                        "name": "namedParts"
```
alegmarra: o.0 what's this?

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/capability/xml/TestAnotherDocumentedConfig.java
marianogonzalez: you should be able to leave this name as is and it should still work.

juandesi: it will, There is another case where config is in the name of the configuration I just changed this one so we have one name without the config suffix

PR #4918, opened 2017-06-02
MULE-11527: Parameters should be resolved only once if before doesn't make any changes while intercepting the processor

elrodro83:


core-tests/src/test/java/org/mule/runtime/core/processor/interceptor/ReactiveInterceptorAdapterTestCase.java
```diff
@@ -381,61 +395,33 @@ public void interceptorMutatesEventAroundAfterFailWithCause() throws Exception {
...
-  @Test
-  public void interceptorThrowsExceptionBeforeNoMock() throws Exception {
-    RuntimeException expectedException = new RuntimeException("Some Error");
-    ProcessorInterceptor interceptor = spy(new ProcessorInterceptor() {
 
       @Override
-      public void before(ComponentLocation location, Map<String, Object> parameters, InterceptionEvent event) {
-        throw expectedException;
+      public void after(ComponentLocation location, InterceptionEvent event, Optional<Throwable> thrown) {
+        // TODO Auto-generated method stub
```
gsfernandes: Remove TODO

core/src/main/java/org/mule/runtime/core/processor/chain/AbstractMessageProcessorChain.java
```diff
@@ -225,7 +227,7 @@ public ProcessingType getProcessingType() {
       };
     } else {
       return messagingException -> Mono.from(getMessagingExceptionHandler().apply(messagingException))
-          .flatMap(handled -> {
+          .flatMapMany(handled -> {
```
gsfernandes: @dfeist please check this change. This should also be changed in other places as the method was deprecated.

PR #4924, opened 2017-06-05
MULE-12608: Remove com.google.common from exported packages

pablokraan:


core/src/main/java/org/mule/runtime/core/util/collection/ImmutableListCollector.java
```diff
@@ -23,29 +23,29 @@
  * @param <T> the generic type of the elements in the list
  * @since 4.0
  */
-public class ImmutableListCollector<T> implements Collector<T, ImmutableList.Builder<T>, List<T>> {
+public class ImmutableListCollector<T> implements Collector<T, ImmutableListBuilder<T>, List<T>> {
```
elrodro83: Maybe we should have an `ImmutableCollectors` utility class to build these Immutable Collection Collectors objects. That way, only that utility will need to be exposed and all these stuff would be internal.

PR #4926, opened 2017-06-05
MULE-12606: Remove org.apache.commons from exported packages

pablokraan:


core/src/main/java/org/mule/runtime/core/util/CaseInsensitiveHashMap.java
```diff
@@ -29,42 +31,15 @@
  *
  * @since 3.0.0
  */
-// TODO(pablo.kraan): MULE-12606 - remove this class from the API or avoid exposing apache classes
-public class CaseInsensitiveHashMap extends AbstractHashedMap implements Serializable {
+public class CaseInsensitiveHashMap<K, V> implements Map<K, V>, Serializable {
```
elrodro83: add the *serialVersionUID*

PR #4928, opened 2017-06-05
MULE-12241 - Embedded container should run based on an app

pablolagreca:


pablolagreca:
DO NOT MERGE

distributions/embedded/embedded-impl/src/main/java/org/mule/runtime/module/embedded/impl/EmbeddedController.java
```diff
@@ -146,29 +140,13 @@ public void executeWithinContainerClassLoader(ContainerTask task) {
...
-    return applicationFolder;
-  }
-
   /**
    * Invoked by reflection
    */
   public void stop() {
     executeWithinContainerClassLoader(() -> {
       deleteTree(new File(containerInfo.getContainerBaseFolder().getPath()));
+      Domain domain = application.getDomain();
```
gsfernandes: Did you forget to do something here with the domain?

PR #4930, opened 2017-06-05
EE-5513: Create Integration Poll + Object Store (Store) + Object Stor…

pablomusumeci:
…e (Retrieve)


tests/allure/src/main/java/org/mule/test/allure/AllureConstants.java
```diff
@@ -184,6 +184,15 @@
 
   }
 
+  interface ObjectStoreFeature {
+
+    String OS_EXTENSION = "ObjectStore Extension";
```
aiannucci: ObjectStore Connector

PR #4931, opened 2017-06-05
MULE-12671: WSC is not providing a way to pass transport headers

juandesi:


modules/soap-api/src/main/java/org/mule/runtime/soap/api/message/SoapMultipartPayload.java
```diff
@@ -19,33 +22,55 @@
...
+  }
+
+  @Override
+  public List<String> getPartNames() {
+    return multipart.getPartNames();
+  }
+
+  @Override
+  public Message getPart(String partName) {
+    return multipart.getPart(partName);
```
alegmarra: org.mule.runtime.core.message.DefaultMultiPartPayload#getPart does a findFirst().get(), is it possible to ask for a part that is not there? if that happens, everything blows up

PR #4932, opened 2017-06-05
MULE-12608: Remove com.google.common from exported packages

pablokraan:
_ Adding collectors utils


core/src/main/java/org/mule/runtime/core/util/collection/Collectors.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collector;
+
+/**
+ * Provides different implementations of {@link Collector}
+ */
```
elrodro83: add `@since`

PR #4935, opened 2017-06-05
MULE-12673: fix cases when an exception occurs in an error handler

alepulver:
do not merge yet

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/exception/MessagingException.java
+++ b/core/src/main/java/org/mule/runtime/core/exception/MessagingException.java
@@ -54,6 +54,7 @@ public class MessagingException extends MuleException {
 
   private boolean causeRollback;
   private boolean handled;
+  private boolean inErrorHandler;
   private transient Processor failingMessageProcessor;
 
   /**
@@ -318,6 +319,24 @@ public class MessagingException extends MuleException {
    */
   public boolean handled() {
     return handled;
+  }
+
+  /**
+   * Marks the exception so that onTerminate is called on the source instead of onError if not handled
+   *
+   * @param inErrorHandler true if the exception has occurred in an error handler, false otherwise
+   */
+  public void setInErrorHandler(boolean inErrorHandler) {
+    this.inErrorHandler = inErrorHandler;
+  }
+
+  /**
+   * Signals if the exception occurred in an error handler
+   *
+   * @return true if happened inside error handler
+   */
+  public boolean inErrorHandler() {
+    return inErrorHandler;
   }
 
   /**
--- a/core/src/main/java/org/mule/runtime/core/exception/TemplateOnErrorHandler.java
+++ b/core/src/main/java/org/mule/runtime/core/exception/TemplateOnErrorHandler.java
@@ -13,7 +13,6 @@ import static org.mule.runtime.api.component.ComponentIdentifier.buildFromString
 import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
 import static org.mule.runtime.core.api.context.notification.EnrichedNotificationInfo.createInfo;
 import static org.mule.runtime.core.api.processor.MessageProcessors.newChain;
-import static org.mule.runtime.core.api.processor.MessageProcessors.processToApply;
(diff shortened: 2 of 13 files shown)


core/src/main/java/org/mule/runtime/core/exception/MessagingException.java
```diff
@@ -321,6 +322,24 @@ public boolean handled() {
   }
 
   /**
+   * Marks the exception so that onTerminate is called on the source instead of onError if not handled
+   *
+   * @param inErrorHandler true if the exception has occurred in an error handler, false otherwise
+   */
+  public void setInErrorHandler(boolean inErrorHandler) {
```
dfeist: I don't see why it is necessary to add state here to resolve this issue.

core/src/main/java/org/mule/runtime/core/exception/TemplateOnErrorHandler.java
```diff
@@ -111,15 +109,11 @@ public Event process(Event event) throws MuleException {
...
-            // Use child context if HandleExceptionInterceptor is being used to avoid response being completed twice..
-            return Mono.from(publisher).flatMapMany(event -> processWithChildContext(event, p -> from(p)
-                .flatMapMany(childEvent -> Mono.from(routeAsync(childEvent, exception)))));
-          } else {
-            return from(publisher).then(event -> from(routeAsync(event, exception)));
-          }
+          // TODO MULE-11023 Migrate transaction execution template mechanism to use non-blocking API
+          // Use child context if HandleExceptionInterceptor is being used to avoid response being completed twice,
+          // and nested outer "try" blocks being skipped.
+          return Mono.from(publisher).flatMapMany(event -> processWithChildContext(event, p -> from(p)
```
dfeist: We discussed making process() blocking instead of doing this.  Just need to ensure tests pass.

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
```diff
@@ -142,25 +144,26 @@ public void runPhase(final ModuleFlowProcessingPhaseTemplate template, final Mes
...
-                }
-              } else {
-                try {
+            .doOnSuccess(successConsumer)
+            .onErrorResume(MessagingException.class, me -> {
+              try {
+                if (me.getCause() instanceof SourceErrorException
+                    && sourceResponseErrorTypeMatcher.match(((SourceErrorException) me.getCause()).getErrorType())) {
+                  handleSourceError(exceptionHandler, errorConsumer, me, terminateConsumer);
+                } else if (me.inErrorHandler()) {
```
dfeist: IMO rather than have a different code path for handling error handler exceptions like this, all exceptions from Flow should go via same (onError) route, and if required onError can do conditional behaviour based of if error was thrown by error handler or not by matching ErrorType.

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
```diff
@@ -222,10 +231,22 @@ public void runPhase(final ModuleFlowProcessingPhaseTemplate template, final Mes
...
+    errorConsumer.accept(exception);
+    onTerminate(terminateConsumer, right(me.getEvent()));
+  }
+
+  private void handleErrorHandlingException(Consumer<MessagingException> errorConsumer,
+                                            Consumer<Either<MessagingException, Event>> terminateConsumer, MessagingException me,
+                                            PhaseResultNotifier phaseResultNotifier) {
+    phaseResultNotifier.phaseFailure((Exception) me.getCause());
+    SourceErrorException see = new SourceErrorException(me.getEvent(), sourceErrorResponseGenerateErrorType, me.getCause(), me);
+    onTerminate(terminateConsumer, left(see.toMessagingException()));
```
dfeist: This may be required, but it's not clear to me why this type of error should be handled via onTerminate, while other errors are handler via errorConsumer.  If there are things like parameter resosolution that shouldn't happen in the case of an error in error handler then shouldn't that be a condition (based on errorType) in error consumer?

PR #4937, opened 2017-06-05
Mule-12678: smart connectors evaluater parameter/properties properly …

fernandezlautaro:
…by bypassing the DataType to the ExpressionManager

### DIFF
--- a/modules/spring-config/src/main/java/org/mule/runtime/config/spring/dsl/model/extension/xml/MacroExpansionModuleModel.java
+++ b/modules/spring-config/src/main/java/org/mule/runtime/config/spring/dsl/model/extension/xml/MacroExpansionModuleModel.java
@@ -10,9 +10,9 @@ import static java.lang.String.format;
 import static java.util.Collections.emptyMap;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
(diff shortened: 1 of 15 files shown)


modules/spring-config/src/main/java/org/mule/runtime/config/spring/dsl/model/extension/xml/MacroExpansionModuleModel.java
```diff
@@ -52,11 +52,11 @@
  */
 public class MacroExpansionModuleModel {
 
-  /**
-   * literal that represents the name of the global element for any given module. If the module's name is math, then the value of
-   * this field will name the global element as <math:config ../>
-   */
-  private static final String MODULE_CONFIG_GLOBAL_ELEMENT_NAME = "config";
+  //  /**
```
pablolagreca: Do you need to remove this?

fernandezlautaro: I moved it to `org/mule/runtime/core/processor/chain/ModuleOperationMessageProcessorChainBuilder.java` so that there's one place where the default config is defined.

PR #4941, opened 2017-06-06
MULE-12663: In Message Filter, When unaccepted events are handled, execution of flow caller is not stopped.

facundovs:
In a subflow with a filter, if the filter condition is not satisfied and the unaccepted event is handled, then the flow that called this subflow must
stop its execution.

System property was added to keep backward compatibility.

### DIFF
--- a/core/src/main/java/org/mule/processor/AbstractFilteringMessageProcessor.java
+++ b/core/src/main/java/org/mule/processor/AbstractFilteringMessageProcessor.java
@@ -6,7 +6,6 @@
  */
 package org.mule.processor;
 
-import org.mule.VoidMuleEvent;
 import org.mule.api.MessagingException;
 import org.mule.api.MuleEvent;
 import org.mule.api.MuleException;
--- a/core/src/main/java/org/mule/processor/AbstractFilteringMessageProcessor.java
+++ b/core/src/main/java/org/mule/processor/AbstractFilteringMessageProcessor.java
@@ -27,7 +27,7 @@ import org.mule.config.i18n.CoreMessages;
 public abstract class AbstractFilteringMessageProcessor extends AbstractInterceptingMessageProcessor  implements NonBlockingSupported
 {
 
-    public static final String FILTERS_STOP_ALL_FLOW_CALLERS = SYSTEM_PROPERTY_PREFIX + "filterOnUnacceptedStopsParentFlow";
+    public static final String FILTER_ON_UNACCEPTED_STOPS_PARENT_FLOW = SYSTEM_PROPERTY_PREFIX + "filterOnUnacceptedStopsParentFlow";
 
     /** 
      * Throw a FilterUnacceptedException when a message is rejected by the filter? 
@@ -125,6 +125,6 @@ public abstract class AbstractFilteringMessageProcessor extends AbstractIntercep
 
     private boolean shouldFiltersStopAllFlowCallers()
     {
-        return getBoolean(FILTERS_STOP_ALL_FLOW_CALLERS)? true : false;
+        return getBoolean(FILTER_ON_UNACCEPTED_STOPS_PARENT_FLOW);
     }
 }
--- a/core/src/main/java/org/mule/processor/AbstractFilteringMessageProcessor.java
+++ b/core/src/main/java/org/mule/processor/AbstractFilteringMessageProcessor.java
@@ -7,6 +7,7 @@
 package org.mule.processor;
 
 import static java.lang.System.getProperty;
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/mule/processor/AbstractFilteringMessageProcessor.java
```diff
@@ -108,4 +120,9 @@ public void setThrowOnUnaccepted(boolean throwOnUnaccepted)
     {
         this.throwOnUnaccepted = throwOnUnaccepted;
     }
+
+    private boolean shouldFiltersStopAllFlowCallers()
+    {
+        return "true".equals(getProperty(FILTERS_STOP_ALL_FLOW_CALLERS))? true : false;
```
dfeist: Use java.lang.Boolean#getBoolean

core/src/main/java/org/mule/processor/AbstractFilteringMessageProcessor.java
```diff
@@ -60,10 +65,17 @@ public MuleEvent process(MuleEvent event) throws MuleException
...
 
     protected MuleEvent handleUnaccepted(MuleEvent event) throws MuleException
-    {        
+    {
         if (unacceptedMessageProcessor != null)
         {
-            return unacceptedMessageProcessor.process(event);
+            if (shouldFiltersStopAllFlowCallers())
+            {
+                return null;
```
dfeist: Shouldn't this call `unacceptedMessageProcessor` before returning null?

tests/integration/src/test/resources/message-filter-config.xml
```diff
@@ -0,0 +1,25 @@
...
+    </flow>
+
+    <sub-flow name="SubFlow_Level1" >
+        <message-filter onUnaccepted="SubFlow_level2">
+            <expression-filter expression="#[flowVars.flag == true]"/>
+        </message-filter>
+    </sub-flow>
+
+    <sub-flow name="SubFlow_level2">
+        <echo-component/>
```
dfeist: Consider using `assert` elements in flow.

tests/integration/src/test/java/org/mule/routing/outbound/MessageFilterFunctionalTestCase.java
```diff
@@ -0,0 +1,51 @@
...
+    @Override
+    protected String getConfigFile()
+    {
+        return "message-filter-config.xml";
+    }
+
+    @Test
+    public void testFlowCallerStopsAfterUnacceptedEvent() throws Exception
+    {
+        runFlow("MainFlow");
```
dfeist: Need to assert that `SubFlow_level2` is invoked. Consider using `assert` elements, make this easier and easy to read.  You can avoid use of teat component as well with this i think.

core/src/main/java/org/mule/processor/AbstractFilteringMessageProcessor.java
```diff
@@ -23,6 +25,9 @@
  */
 public abstract class AbstractFilteringMessageProcessor extends AbstractInterceptingMessageProcessor  implements NonBlockingSupported
 {
+
+    public static final String FILTERS_STOP_ALL_FLOW_CALLERS = SYSTEM_PROPERTY_PREFIX + "filter.stop.all.flow.callers";
```
dfeist: camel case e.g. `mule.filterOnUnacceptedStopsParentFlow`

core/src/main/java/org/mule/processor/AbstractFilteringMessageProcessor.java
```diff
@@ -23,6 +26,9 @@
  */
 public abstract class AbstractFilteringMessageProcessor extends AbstractInterceptingMessageProcessor  implements NonBlockingSupported
 {
+
+    public static final String FILTERS_STOP_ALL_FLOW_CALLERS = SYSTEM_PROPERTY_PREFIX + "filterOnUnacceptedStopsParentFlow";
```
dfeist: Can we update constant to match name (minor)?

core/src/main/java/org/mule/processor/AbstractFilteringMessageProcessor.java
```diff
@@ -108,4 +122,9 @@ public void setThrowOnUnaccepted(boolean throwOnUnaccepted)
     {
         this.throwOnUnaccepted = throwOnUnaccepted;
     }
+
+    private boolean shouldFiltersStopAllFlowCallers()
+    {
+        return getBoolean(FILTERS_STOP_ALL_FLOW_CALLERS)? true : false;
```
dfeist: Why is `? true : false` needed?

PR #4947, opened 2017-06-06
MULE-12702: Add the possibility to provide a Charset when consuming a…

pablokraan:
… stream from the IOUtils

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/util/IOUtils.java
+++ b/core/src/main/java/org/mule/runtime/core/api/util/IOUtils.java
@@ -31,6 +31,7 @@ import java.io.Reader;
 import java.io.Writer;
 import java.net.MalformedURLException;
 import java.net.URL;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mule/runtime/core/api/util/IOUtils.java
```diff
@@ -179,6 +180,18 @@ public static String toString(InputStream input) {
   }
 
   /**
+   * This method wraps {@link org.apache.commons.io.IOUtils}' <code>toString(InputStream, Charset)</code> method but catches any
+   * {@link IOException} and wraps it into a {@link RuntimeException}.
+   */
+  public static String toString(InputStream input, Charset encoding) {
```
elrodro83: should we deprecate `toString(CursorStreamProvider cursorStreamProvider)` in favor of this method?

PR #4949, opened 2017-06-06
EE-5482: Fix ee:transform operation model - fix tests

pablomusumeci:


modules/spring-config/src/main/java/org/mule/runtime/config/spring/dsl/model/internal/DefaultXmlDslElementModelConverter.java
```diff
@@ -197,7 +197,6 @@ private Element createElement(String name, String prefix, String namespace) {
     } else {
       doc.getDocumentElement().setAttributeNS("http://www.w3.org/2000/xmlns/",
                                               "xmlns", CORE_NAMESPACE);
-      addNamespaceDeclarationIfNeeded(CORE_PREFIX, namespace, buildSchemaLocation(prefix, namespace));
```
alegmarra: we want to make sure the schema is added, just don't add the extra prefix

PR #4958, opened 2017-06-07
MULE-12713: Default Error Handler not consistently initialized

elrodro83:


dfeist:
Take a look at `TransactionalInterceptingMessageProcessor` and MULE-12697 because there are other places that perform lifecycle on default exception handler.

PR #4960, opened 2017-06-07
MULE-12717: Add status parameter to mule.bat

facundovs:
The status parameter is absent in mule.bat but it exists in .sh script.


aiannucci:
Have you tested this in Windows?

facundovs:
Yes. I was coding and testing it in Windows.

PR #4961, opened 2017-06-07
MULE-12588: Support Sources that send asynchronous responses

marianogonzalez:
``

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
+++ b/core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
@@ -140,6 +140,7 @@ public class ModuleFlowProcessingPhase
             .doOnNext(request -> fireNotification(messageProcessContext.getMessageSource(), request,
                                                   messageProcessContext.getFlowConstruct(),
                                                   MESSAGE_RECEIVED))
+            // switchIfEmpty in case the flow was filtered
             .then(request -> from(template.routeEventAsync(request)).switchIfEmpty(fromCallable(() -> emptyEvent(templateEvent))))
             .then(onSuccess(messageSource, templateEvent, messageProcessContext, phaseResultNotifier, template,
                             terminateConsumer))
--- a/core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
+++ b/core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
@@ -143,17 +143,12 @@ public class ModuleFlowProcessingPhase
             .then(request -> from(template.routeEventAsync(request)).switchIfEmpty(fromCallable(() -> emptyEvent(templateEvent))))
             .then(onSuccess(messageSource, templateEvent, messageProcessContext, phaseResultNotifier, template,
                             terminateConsumer))
+            .onErrorMap(MessagingException.class, me -> me.getCause() instanceof SourceErrorException ? me.getCause() : me)
             .onErrorResume(SourceErrorException.class,
                            see -> onSourceException(exceptionHandler, errorHandler, terminateConsumer, see))
-            .onErrorResume(MessagingException.class, me -> {
-              if (me.getCause() instanceof SourceErrorException) {
-                return onSourceException(exceptionHandler, errorHandler, terminateConsumer, (SourceErrorException) me.getCause());
-              } else {
-                return from(errorHandler.apply(me))
-                    .doOnSuccess(v -> onTerminate(terminateConsumer, right(me)))
-                    .doOnError(e -> onTerminate(terminateConsumer, right(e)));
-              }
-            })
+            .onErrorResume(MessagingException.class, me -> from(errorHandler.apply(me))
+                .doOnSuccess(v -> onTerminate(terminateConsumer, right(me)))
+                .doOnError(e -> onTerminate(terminateConsumer, right(e))))
             .doAfterTerminate((event, throwable) -> responseCompletion.onComplete())
             .subscribe();
       } else {
@@ -221,10 +216,10 @@ public class ModuleFlowProcessingPhase
     }
   }
 
-  private Mono<? extends Void> onSourceException(MessagingExceptionHandler exceptionHandler,
-                                                 Function<MessagingException, Publisher<Void>> errorHandler,
-                                                 Consumer<Either<Event, MessagingException>> terminateConsumer,
-                                                 SourceErrorException see) {
+  private Mono<Void> onSourceException(MessagingExceptionHandler exceptionHandler,
+                                       Function<MessagingException, Publisher<Void>> errorHandler,
+                                       Consumer<Either<Event, MessagingException>> terminateConsumer,
+                                       SourceErrorException see) {
     if (sourceResponseErrorTypeMatcher.match(see.getErrorType())) {
       return from(handleSourceError(exceptionHandler, errorHandler, see))
           .doOnSuccess(v -> onTerminate(terminateConsumer, left(see.getEvent())))
--- a/core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
+++ b/core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
@@ -26,7 +26,9 @@ import static org.mule.runtime.core.execution.TransactionalErrorHandlingExecutio
 import static org.mule.runtime.core.internal.util.FunctionalUtils.safely;
 import static org.mule.runtime.core.util.message.MessageUtils.toMessage;
 import static org.mule.runtime.core.util.message.MessageUtils.toMessageCollection;
+import static reactor.core.publisher.Mono.error;
 import static reactor.core.publisher.Mono.from;
+import static reactor.core.publisher.Mono.fromCallable;
 import static reactor.core.publisher.Mono.just;
 import org.mule.runtime.api.component.location.ComponentLocation;
 import org.mule.runtime.api.exception.MuleException;
@@ -51,18 +53,19 @@ import org.mule.runtime.core.policy.PolicyManager;
 import org.mule.runtime.core.policy.SourcePolicy;
 import org.mule.runtime.core.policy.SuccessSourcePolicyResult;
 import org.mule.runtime.core.transaction.MuleTransactionConfig;
-import org.mule.runtime.core.api.util.func.CheckedConsumer;
+import org.mule.runtime.core.api.util.func.CheckedFunction;
 import org.mule.runtime.extension.api.runtime.operation.Result;
-
-import org.reactivestreams.Publisher;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import reactor.core.publisher.MonoProcessor;
 
 import java.util.Collection;
 import java.util.Map;
 import java.util.function.Consumer;
 import java.util.function.Function;
+
+import org.reactivestreams.Publisher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import reactor.core.publisher.Mono;
+import reactor.core.publisher.MonoProcessor;
 
 /**
  * This phase routes the message through the flow.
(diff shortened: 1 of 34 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ModuleFlowProcessingTemplate.java
```diff
@@ -88,12 +88,10 @@ public void sendAfterTerminateResponseToClient(Either<Event, MessagingException>
...
   }
 
-  private void runAndNotify(CheckedRunnable runnable, Event event, ResponseCompletionCallback responseCompletionCallback) {
-    try {
-      runnable.run();
-      responseCompletionCallback.responseSentSuccessfully();
-    } catch (Exception e) {
-      responseCompletionCallback.responseSentWithFailure(new MessagingException(event, e), event);
-    }
+  private Mono<Void> runAndNotify(CheckedSupplier<Publisher<Void>> supplier, Event event,
```
dfeist: For reuse like this, I find this approach to be clearer, it's also the same as we are doing everywhere else:
```
  @Override
  public Publisher<Void> sendResponseToClient(Event event, Map<String, Object> parameters,
                                              Function<Event, Map<String, Object>> errorResponseParametersFunction,
                                              ResponseCompletionCallback responseCompletionCallback) {
    return from(completionHandler.onCompletion(event, parameters)).transform(notifyCompletion(event, responseCompletionCallback));
  }

  private Function<Publisher<Void>, Publisher<Void>> notifyCompletion(Event event,
                                                                      ResponseCompletionCallback responseCompletionCallback) {
    return publisher -> from(publisher)
        .doOnSuccess(v -> responseCompletionCallback.responseSentSuccessfully())
        .doOnError(e -> responseCompletionCallback.responseSentWithFailure(new MessagingException(event, e), event));
  }
```

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/DefaultSourceCallback.java
```diff
@@ -216,11 +216,9 @@ private void executeFlowTransactionally(SourceCallbackContext context, MessagePr
   }
 
   private void executeFlow(SourceCallbackContext context, MessageProcessContext messageProcessContext, Message message) {
-    messageProcessingManager.processMessage(new ModuleFlowProcessingTemplate(message,
-                                                                             listener,
-                                                                             completionHandlerFactory
-                                                                                 .createCompletionHandler(context),
-                                                                             messageProcessContext),
+    messageProcessingManager.processMessage(
```
dfeist: We should take a closer look at this and see if messageProcessingManager is really needed, or just add a layer of abstraction/complexity (different issue of course)

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
```diff
@@ -134,36 +137,28 @@ public void runPhase(final ModuleFlowProcessingPhaseTemplate template, final Mes
...
-                try {
-                  handleSourceError(exceptionHandler, errorConsumer, (SourceErrorException) me.getCause());
-                  onTerminate(terminateConsumer, left(me.getEvent()));
-                } catch (Exception e) {
-                  onTerminate(terminateConsumer, right(e));
-                }
+            .doOnNext(request -> fireNotification(messageProcessContext.getMessageSource(), request,
+                                                  messageProcessContext.getFlowConstruct(),
+                                                  MESSAGE_RECEIVED))
+            .then(request -> from(template.routeEventAsync(request)).switchIfEmpty(fromCallable(() -> emptyEvent(templateEvent))))
```
dfeist: Why is this `switchIfEmpty` needed here? Is it related to this issue or a piggy-back fix.  I assume this is to handle the case where there is a filter in the flow that drop the event to ensure onError/onSuccess/onTerminate behaves consistently in this case too, but a comment to explain this would help.

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
```diff
@@ -390,11 +411,11 @@ public Event responseSentWithFailure(final MessagingException e, final Event eve
 
   /**
    * This method will not throw any {@link Exception}.
-   * 
+   *
    * @param terminateConsumer the action to perform on the transformed result.
-   * @param result the outcome of trying to send the response of the source through the source. In the case of error, only
-   *        {@link MessagingException} or {@link SourceErrorException} are valid values on the {@code right} side of this
-   *        parameter.
+   * @param result            the outcome of trying to send the response of the source through the source. In the case of error, only
```
dfeist: extra spaces?

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
```diff
@@ -219,11 +221,24 @@ public void runPhase(final ModuleFlowProcessingPhaseTemplate template, final Mes
...
 
-  private void handleSourceError(final MessagingExceptionHandler exceptionHandler, Consumer<MessagingException> errorConsumer,
-                                 SourceErrorException see) {
+  private Mono<? extends Void> onSourceException(MessagingExceptionHandler exceptionHandler,
+                                                 Function<MessagingException, Publisher<Void>> errorHandler,
+                                                 Consumer<Either<Event, MessagingException>> terminateConsumer,
+                                                 SourceErrorException see) {
+    if (sourceResponseErrorTypeMatcher.match(see.getErrorType())) {
+      return from(handleSourceError(exceptionHandler, errorHandler, see))
+          .doOnSuccess(v -> onTerminate(terminateConsumer, left(see.getEvent())))
```
dfeist: `onTerminate` should be being invoked from main Mono, not in onSucess/onError independently.

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
```diff
@@ -134,36 +137,28 @@ public void runPhase(final ModuleFlowProcessingPhaseTemplate template, final Mes
...
+                return onSourceException(exceptionHandler, errorHandler, terminateConsumer, (SourceErrorException) me.getCause());
               } else {
-                try {
-                  errorConsumer.accept(me);
-                  onTerminate(terminateConsumer, right(me));
-                } catch (Exception e) {
-                  onTerminate(terminateConsumer, right(e));
-                }
+                return from(errorHandler.apply(me))
+                    .doOnSuccess(v -> onTerminate(terminateConsumer, right(me)))
```
dfeist: onTerminate should be being invoked from main Mono, not in onSucess/onError and here independently.  This should simply be:
`. onErrorResume(me -> errorHandler.apply(me))`

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
```diff
@@ -219,11 +221,24 @@ public void runPhase(final ModuleFlowProcessingPhaseTemplate template, final Mes
     }
   }
 
-  private void handleSourceError(final MessagingExceptionHandler exceptionHandler, Consumer<MessagingException> errorConsumer,
-                                 SourceErrorException see) {
+  private Mono<? extends Void> onSourceException(MessagingExceptionHandler exceptionHandler,
```
dfeist: Why not just `Mono<Void>`?

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
```diff
@@ -134,36 +137,28 @@ public void runPhase(final ModuleFlowProcessingPhaseTemplate template, final Mes
...
+            .doOnNext(request -> fireNotification(messageProcessContext.getMessageSource(), request,
+                                                  messageProcessContext.getFlowConstruct(),
+                                                  MESSAGE_RECEIVED))
+            .then(request -> from(template.routeEventAsync(request)).switchIfEmpty(fromCallable(() -> emptyEvent(templateEvent))))
+            .then(onSuccess(messageSource, templateEvent, messageProcessContext, phaseResultNotifier, template,
+                            terminateConsumer))
+            .onErrorResume(SourceErrorException.class,
+                           see -> onSourceException(exceptionHandler, errorHandler, terminateConsumer, see))
+            .onErrorResume(MessagingException.class, me -> {
+              if (me.getCause() instanceof SourceErrorException) {
```
dfeist: This can be removed, if error is mapped above I believe.

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
```diff
@@ -134,36 +137,28 @@ public void runPhase(final ModuleFlowProcessingPhaseTemplate template, final Mes
...
-                  onTerminate(terminateConsumer, right(me));
-                } catch (Exception e) {
-                  onTerminate(terminateConsumer, right(e));
-                }
+                return from(errorHandler.apply(me))
+                    .doOnSuccess(v -> onTerminate(terminateConsumer, right(me)))
+                    .doOnError(e -> onTerminate(terminateConsumer, right(e)));
               }
             })
             .doAfterTerminate((event, throwable) -> responseCompletion.onComplete())
```
dfeist: `onTerminate` should be called here.. just before `responseCompletion.onComplete()` shouldn't it?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceCallbackExecutor.java
```diff
@@ -23,11 +25,11 @@
   /**
    * Executes the callback
    *
-   * @param event the result {@link Event}
+   * @param event      the result {@link Event}
```
marianogonzalez: rodro's formatter

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/DefaultSourceCallback.java
marianogonzalez: agreed, but out of the scope of this PR

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
marianogonzalez: without this , a test of a filtered flow with http:listener fails

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
marianogonzalez: rodro

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
alepulver: Doesn't our error hook map thrown exceptions into MessagingExceptions if they aren't? In that case onErrorResume shouldn't be able to handle other things (except they are put in the stream with error(e) instead of throw e). @dfeist ?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ModuleFlowProcessingTemplate.java
marianogonzalez: done

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
```diff
@@ -134,36 +137,28 @@ public void runPhase(final ModuleFlowProcessingPhaseTemplate template, final Mes
...
-                  onTerminate(terminateConsumer, left(me.getEvent()));
-                } catch (Exception e) {
-                  onTerminate(terminateConsumer, right(e));
-                }
+            .doOnNext(request -> fireNotification(messageProcessContext.getMessageSource(), request,
+                                                  messageProcessContext.getFlowConstruct(),
+                                                  MESSAGE_RECEIVED))
+            .then(request -> from(template.routeEventAsync(request)).switchIfEmpty(fromCallable(() -> emptyEvent(templateEvent))))
+            .then(onSuccess(messageSource, templateEvent, messageProcessContext, phaseResultNotifier, template,
+                            terminateConsumer))
```
marianogonzalez: good one... done

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
marianogonzalez: done

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
marianogonzalez: no... notice that depending on the success/failure of this new mono, the second argument of the onTerminate changes. If I don't do it like this in here, I loose one of those exceptions

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
marianogonzalez: not, because of what explained above

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
marianogonzalez: no reason. Just intelliJ over refactoring. Changed

core/src/main/java/org/mule/runtime/core/execution/ModuleFlowProcessingPhase.java
marianogonzalez: not, because of reasons already explained

PR #4969, opened 2017-06-08
MULE-12696 Try should support non-blocking execution when it has no T…

dfeist:
…X config


core/src/main/java/org/mule/runtime/core/processor/TryMessageProcessor.java
```diff
@@ -0,0 +1,159 @@
...
+      }
+    }
+  }
+
+  @Override
+  public Publisher<Event> apply(Publisher<Event> publisher) {
+    if (nestedChain == null) {
+      return publisher;
+    } else if (isTransactionActive() || transactionConfig.getAction() != ACTION_INDIFFERENT) {
+      return Processor.super.apply(publisher);
```
marianogonzalez: why super.apply() instead of this.process() ?

modules/spring-config/src/main/java/org/mule/runtime/config/spring/factories/TryProcessorFactoryBean.java
```diff
@@ -55,32 +51,12 @@ public void setMessageProcessors(List messageProcessors) {
...
   @Override
   public Object getObject() throws Exception {
-    DefaultMessageProcessorChainBuilder builder = new DefaultMessageProcessorChainBuilder();
-    builder.setName("'transaction' child processor chain");
-    TransactionalInterceptingMessageProcessor txProcessor = new TransactionalInterceptingMessageProcessor();
+    TryMessageProcessor txProcessor = new TryMessageProcessor();
     txProcessor.setAnnotations(getAnnotations());
     txProcessor.setExceptionListener(this.exceptionListener);
     txProcessor.setTransactionConfig(createTransactionConfig(this.transactionalAction, this.transactionType));
-    builder.chain(txProcessor);
```
marianogonzalez: where has this logic moved to?

core/src/main/java/org/mule/runtime/core/processor/TryMessageProcessor.java
```diff
@@ -0,0 +1,159 @@
...
+    return transactionConfig;
+  }
+
+  /**
+   * Configure the nested {@link Processor}'s that error handling and transactional behaviour should be applied to.
+   * 
+   * @param processors
+   */
+  public void setMessageProcessors(List<Processor> processors) {
+    this.processors = processors;
```
afelisatti: You should create the chain here, I don't see why you'd need the processors referenced by themselves.

modules/spring-config/src/main/java/org/mule/runtime/config/spring/factories/TryProcessorFactoryBean.java
dfeist: Some isn't required, the rest is moved to TryMessageProcessor which becomes a scope instead of an interceptor.

core/src/main/java/org/mule/runtime/core/processor/TryMessageProcessor.java
dfeist: Because `super.apply(publisher)`, handles null, exceptions etc.

PR #4971, opened 2017-06-08
MULE-11799: Prevent triple state booleans

pablomusumeci:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/ParameterModelsLoaderDelegate.java
```diff
@@ -107,8 +109,8 @@ public ParameterModelsLoaderDelegate(List<ParameterDeclarerContributor> contribu
       if (extensionParameter.isRequired()) {
         parameter = groupDeclarer.withRequiredParameter(extensionParameter.getAlias());
       } else {
-        parameter = groupDeclarer.withOptionalParameter(extensionParameter.getAlias())
-            .defaultingTo(extensionParameter.defaultValue().isPresent() ? extensionParameter.defaultValue().get() : null);
+        OptionalParameterDeclarer optionalParameter = groupDeclarer.withOptionalParameter(extensionParameter.getAlias());
```
marianogonzalez: completely revert the changes on this class. This should be handled on an enricher of the JavaModelLoader.
As you implemented this here, this concern is scattered around the whole class. I prefer to have a BooleanParametersEnricher in which the whole logic is encapsulated and can be unit tested.

Plus, when the parameter is a native boolean, it's not only about setting the default to false, the parameter must also be marked as not required. The enricher is also the ideal place to make that switch. Finally, consider that when switching the parameter to not required, you will need to update all xsd and jsons

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/BooleanParameterDeclarationEnricher.java
```diff
@@ -0,0 +1,55 @@
...
+      @Override
+      protected void onParameter(ParameterizedDeclaration owner, ParameterGroupDeclaration parameterGroup,
+                                 ParameterDeclaration declaration) {
+        declaration.getType().accept(new MetadataTypeVisitor() {
+
+          @Override
+          public void visitBoolean(BooleanType booleanType) {
+            declaration.setRequired(false);
+            if (declaration.getDefaultValue() == null && !declaration.isConfigOverride()) {
+              declaration.setDefaultValue(valueOf(FALSE));
```
marianogonzalez: just use plain old "false". As a String you're adding an unnecessary transformation

PR #4972, opened 2017-06-08
MULE-12472: RejectionPolicy configuration is confusing

elrodro83:
* Replace `RejectionPolicy` with a `waitAllowed` flag
* Add logging of failures in test schedulers

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/scheduler/SchedulerConfig.java
+++ b/core/src/main/java/org/mule/runtime/core/api/scheduler/SchedulerConfig.java
@@ -106,7 +106,7 @@ public class SchedulerConfig {
   }
 
   /**
-   * whether the threads of the target custom {@link Scheduler} may block to wait when dispatching to a busy {@link Scheduler}.
+   * Whether the threads of the target custom {@link Scheduler} may block to wait when dispatching to a busy {@link Scheduler}.
    * <p>
    * This is only applicable for <b>custom</b> {@link Scheduler}s. This behaviour cannot be changed for the runtime managed
    * {@link Scheduler}.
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/mule/runtime/core/api/scheduler/SchedulerConfig.java
```diff
@@ -128,25 +106,23 @@ public boolean hasName() {
   }
 
   /**
-   * Sets the rejection policy to use when dispatching to a busy {@link Scheduler}.
+   * whether the threads of the target custom {@link Scheduler} may block to wait when dispatching to a busy {@link Scheduler}.
```
dfeist: Capiitalize.

core/src/main/java/org/mule/runtime/core/api/scheduler/SchedulerConfig.java
```diff
@@ -53,23 +31,23 @@ public static SchedulerConfig config() {
...
     this.schedulerPrefix = null;
     this.schedulerName = null;
-    this.rejectionAction = DEFAULT;
+    this.waitAllowed = null;
     this.shutdownTimeoutMillis = () -> null;
   }
 
   private SchedulerConfig(Integer maxConcurrentTasks, String schedulerPrefix, String schedulerName,
-                          RejectionAction rejectionAction, Supplier<Long> shutdownTimeoutMillis) {
+                          Boolean waitAllowed, Supplier<Long> shutdownTimeoutMillis) {
```
dfeist: No way to use boolean or Optional here if really required instead of nullable Boolean?

PR #4973, opened 2017-06-08
MULE-12477: add system property to set memory cap

marianogonzalez:


core/src/main/java/org/mule/runtime/core/api/config/MuleProperties.java
```diff
@@ -195,6 +195,7 @@
   public static final String MULE_SECURITY_SYSTEM_PROPERTY = SYSTEM_PROPERTY_PREFIX + "security.model";
   public static final String MULE_SECURITY_PROVIDER_PROPERTY = SYSTEM_PROPERTY_PREFIX + "security.provider";
   public static final String MULE_STREAMING_BUFFER_SIZE = SYSTEM_PROPERTY_PREFIX + "streaming.bufferSize";
+  public static final String MULE_MAX_STREAMING_MEMORY = SYSTEM_PROPERTY_PREFIX + "max.streaming.memory";
```
dfeist: -> mule.streaming.maxMemory

dfeist: -> MULE_STREAMING_MAX_MEMORY
-> mule.streaming.maxMemory

PR #4980, opened 2017-06-09
MULE-12739: Set Drools Assert Behavior Option to EQUALITY to avoid du…

facundovs:
…plicated facts be saved.

From Drools 3.0, Drools allows to customize how checks if an object already exists in the working memory.
By default AssertBehaviorOption is set to IDENTITY that ignores hashCode() and equals() method to compare objects.
This is causing that duplicated objects (facts) are saved in the working memory causing a memory leak.
We need to set this property, by default, to EQUALITY. In this way, equals() and hashCode() method will be invoked in the objects comparison.

### DIFF
--- a/modules/drools/src/main/java/org/mule/module/drools/Drools.java
+++ b/modules/drools/src/main/java/org/mule/module/drools/Drools.java
@@ -41,7 +41,7 @@ import org.slf4j.LoggerFactory;
 public class Drools implements RulesEngine
 {
 
-    public static final String equalityAssertBehavior= SYSTEM_PROPERTY_PREFIX  + "droolsEqualityAssertBehavior";
+    public static final String USE_EQUALITY_ASSERT_BEHAVIOR = SYSTEM_PROPERTY_PREFIX  + "drools.useEqualityAssert";
 
     /** An optional logical name for the Rules Engine. */
     private String name;
@@ -73,7 +73,7 @@ public class Drools implements RulesEngine
         KnowledgeBaseConfiguration conf = 
             KnowledgeBaseFactory.newKnowledgeBaseConfiguration(null, Thread.currentThread().getContextClassLoader());
 
-        if (isEqualityAssertBehavior())
+        if (useEqualityAssertBehavior())
         {
             conf.setOption(EQUALITY);
         }
@@ -192,9 +192,9 @@ public class Drools implements RulesEngine
         return name;
     }
 
-    private boolean isEqualityAssertBehavior ()
+    private boolean useEqualityAssertBehavior ()
     {
-        return getBoolean(equalityAssertBehavior);
+        return getBoolean(USE_EQUALITY_ASSERT_BEHAVIOR);
     }
 }
 
--- a/modules/drools/src/main/java/org/mule/module/drools/Drools.java
+++ b/modules/drools/src/main/java/org/mule/module/drools/Drools.java
@@ -6,6 +6,8 @@
  */
 package org.mule.module.drools;
 
+import static org.drools.conf.AssertBehaviorOption.EQUALITY;
+
 import org.drools.conf.AssertBehaviorOption;
 import org.mule.api.config.ConfigurationException;
 import org.mule.config.i18n.CoreMessages;
@@ -67,7 +69,7 @@ public class Drools implements RulesEngine
         KnowledgeBaseConfiguration conf = 
             KnowledgeBaseFactory.newKnowledgeBaseConfiguration(null, Thread.currentThread().getContextClassLoader());
 
-        conf.setOption(AssertBehaviorOption.EQUALITY);
+        conf.setOption(EQUALITY);
 
         if (rules.getConfiguration() != null)
         {
--- a/modules/drools/src/main/java/org/mule/module/drools/Drools.java
+++ b/modules/drools/src/main/java/org/mule/module/drools/Drools.java
@@ -6,9 +6,10 @@
  */
 package org.mule.module.drools;
 
+import static java.lang.Boolean.getBoolean;
 import static org.drools.conf.AssertBehaviorOption.EQUALITY;
+import static org.mule.api.config.MuleProperties.SYSTEM_PROPERTY_PREFIX;
 
-import org.drools.conf.AssertBehaviorOption;
 import org.mule.api.config.ConfigurationException;
 import org.mule.config.i18n.CoreMessages;
 import org.mule.config.i18n.MessageFactory;
@@ -39,6 +40,9 @@ import org.slf4j.LoggerFactory;
 
 public class Drools implements RulesEngine
 {
+
+    public static final String equalityAssertBehavior= SYSTEM_PROPERTY_PREFIX  + "droolsEqualityAssertBehavior";
+
     /** An optional logical name for the Rules Engine. */
     private String name;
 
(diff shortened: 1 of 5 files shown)


modules/drools/src/main/java/org/mule/module/drools/Drools.java
```diff
@@ -65,6 +66,9 @@ public Object createSession(Rules rules) throws Exception
 
         KnowledgeBaseConfiguration conf = 
             KnowledgeBaseFactory.newKnowledgeBaseConfiguration(null, Thread.currentThread().getContextClassLoader());
+
+        conf.setOption(AssertBehaviorOption.EQUALITY);
```
pablokraan: I guess that different modes exist for a reason... woyld be good to have a way to configure drools using tje old mode (which could be better for perfoema ce reasons umder some scenarios)

pablokraan: Add stitic import

modules/drools/src/test/java/org.mule.module.drools/DroolsTestCase.java
```diff
@@ -0,0 +1,86 @@
...
+import org.mule.module.bpm.Rules;
+
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Mockito.RETURNS_DEEP_STUBS;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class DroolsTestCase {
```
pablokraan: Extend from AbstractMuleTestCase

modules/drools/src/test/java/org.mule.module.drools/DroolsTestCase.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Test
+    public void testMemoryLeakCausedBySaveDuplicatedObjects() throws Exception
+    {
+        Object handle1 = drools.assertFact(rules, new TestFact("idTest", "descriptionTest"));
+        Object handle2 = drools.assertFact(rules, new TestFact("idTest", "descriptionTest"));
+        assertThat(handle1, is(handle2));
+    }
+
+    public class TestFact {
```
pablokraan: Can be lrivate

facundovs: I reference this class in rulesFile.drl

modules/drools/src/main/java/org/mule/module/drools/Drools.java
```diff
@@ -36,6 +40,9 @@
 
 public class Drools implements RulesEngine
 {
+
+    public static final String equalityAssertBehavior= SYSTEM_PROPERTY_PREFIX  + "droolsEqualityAssertBehavior";
```
pablokraan: Define constants in upper case

pablokraan: "droolsEqualityAssertBehavior" > "drools.useEqualityAssert"

modules/drools/src/main/java/org/mule/module/drools/Drools.java
```diff
@@ -178,6 +191,11 @@ public String getName()
     {
         return name;
     }
+
+    private boolean isEqualityAssertBehavior ()
```
pablokraan: Rename method to match the system property change

modules/drools/src/test/java/org.mule.module.drools/DroolsTestCase.java
```diff
@@ -0,0 +1,93 @@
...
+
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Mockito.RETURNS_DEEP_STUBS;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static org.mule.module.drools.Drools.equalityAssertBehavior;
+
+public class DroolsTestCase extends AbstractMuleTestCase
```
pablokraan: Use a name more specific to the problem you are fixing than DroolsTestCase

modules/drools/src/test/java/org.mule.module.drools/DroolsTestCase.java
```diff
@@ -0,0 +1,93 @@
...
+{
+    @Rule
+    public SystemProperty equalityAssertBehaviorSystemProperty = new SystemProperty(equalityAssertBehavior, "true");
```
pablokraan: Both true and false scenarios must be tested

modules/drools/src/test/java/org.mule.module.drools/DroolsAssertBehaviorTestCase.java
```diff
@@ -37,16 +39,27 @@ public void setUp() throws Exception
...
+        Object handle1 = drools.assertFact(rules, new TestFact("idTest", "descriptionTest"));
+        Object handle2 = drools.assertFact(rules, new TestFact("idTest", "descriptionTest"));
+        assertThat(handle1, is(handle2));
     }
 
     @Test
-    public void testMemoryLeakCausedBySaveDuplicatedObjects() throws Exception
+    public void testIdentityAssertBehaviour() throws Exception
     {
+        clearProperty(USE_EQUALITY_ASSERT_BEHAVIOR);
```
pablokraan: You cannot ensure that this method will be executed before the other one. The right way to test this is to use a parameterized test that receives the property value and a function to assert the configuration, and then use org.mule.tck.MuleTestUtils.testWithSystemProperty

modules/drools/src/test/java/org.mule.module.drools/DroolsAssertBehaviorTestCase.java
```diff
@@ -0,0 +1,130 @@
...
+import static org.mockito.Mockito.RETURNS_DEEP_STUBS;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static org.mule.module.drools.Drools.USE_EQUALITY_ASSERT_BEHAVIOR;
+import static org.mule.tck.MuleTestUtils.testWithSystemProperty;
+import static org.mule.tck.MuleTestUtils.TestCallback;
+import static org.junit.runners.Parameterized.Parameters;
+
+
+@RunWith(Parameterized. class)
```
pablokraan: Remove extra space

modules/drools/src/test/java/org.mule.module.drools/DroolsAssertBehaviorTestCase.java
```diff
@@ -0,0 +1,130 @@
...
+        public void setDescription(String description)
+        {
+            this.description = description;
+        }
+    }
+
+    static TestCallback identityBehaviourCallback =  new TestCallback(){
+        @Override
+        public void run() throws Exception {
+            setUp();
```
pablokraan: Why don't you use @BeforeClass?

facundovs: Because I need to execute createSession() method before each test.

pablokraan: create different methods then, one for the class setup and another one for the instance setup

modules/drools/src/test/java/org.mule.module.drools/DroolsAssertBehaviorTestCase.java
pablokraan: I was referring to the space on @RunWith(Parameterized. class)

PR #4984, opened 2017-06-09
MULE-12745: Set HeapMemoryManager as Default Grizzly Memory Manager.

facundovs:
PooledMemoryManager is causing unexpected behaviors in HTTP Requester. Additionaly, this change apparently impacted in memory consumption.
We should use HeapMemoryManager by default, that works as expected in previous mule versions, at least until Grizzly gives an answer to the reported issues.


afelisatti:
This is not enough IMO. It will only make standalone use that memory manager but our tests will still use the default one.

modules/http/src/main/java/org/mule/module/http/internal/request/grizzly/MemoryManagerTransportCustomizer.java
```diff
@@ -0,0 +1,24 @@
...
+
+import org.glassfish.grizzly.filterchain.FilterChainBuilder;
+import org.glassfish.grizzly.memory.HeapMemoryManager;
+import org.glassfish.grizzly.nio.transport.TCPNIOTransport;
+
+public class MemoryManagerTransportCustomizer implements TransportCustomizer
+{
+
+    @Override
+    public void customize(TCPNIOTransport tcpnioTransport, FilterChainBuilder filterChainBuilder)
```
afelisatti: `tcpNioTransport` or just `transport`

modules/http/src/main/java/org/mule/module/http/internal/request/grizzly/GrizzlyHttpClient.java
```diff
@@ -88,6 +88,8 @@
 
     private static final int MAX_CONNECTION_LIFETIME = 30 * 60 * 1000;
 
+    public static String GRIZZLY_MEMORY_MANAGER_SYSTEM_PROPERTY = "org.glassfish.grizzly.DEFAULT_MEMORY_MANAGER";
```
afelisatti: This must already be defined in grizzly, I think it's DMM_PROP_NAME.

facundovs: Yes, but It is private.

distributions/standalone/src/main/resources/conf/wrapper.conf
```diff
@@ -39,6 +39,9 @@ wrapper.java.additional.15=-Dorg.quartz.scheduler.skipUpdateCheck=true
 # For large maps (>512) make the hash function depend on the JVM instance to prevent collision attacks (see MULE-11206)
 wrapper.java.additional.16=-Djdk.map.althashing.threshold=512
 
+# Set Default Grizzly Memory Manager.
+# wrapper.java.additional.17=-Dorg.glassfish.grizzly.DEFAULT_MEMORY_MANAGER=org.glassfish.grizzly.memory.HeapMemoryManager
```
afelisatti: Do not set the number if it's commented, and change the explanation to "Allows modifying Grizzly's memory manager"

PR #4994, opened 2017-06-12
MULE-12750: adding validation when an operation of a smart connector …

fernandezlautaro:
…is missing the 'config-ref' attribute


modules/spring-config/src/main/java/org/mule/runtime/config/spring/dsl/model/extension/xml/MacroExpansionModuleModel.java
```diff
@@ -16,6 +16,7 @@
 import static org.mule.runtime.internal.dsl.DslConstants.CORE_PREFIX;
 import static org.mule.runtime.internal.dsl.DslConstants.KEY_ATTRIBUTE_NAME;
 import static org.mule.runtime.internal.dsl.DslConstants.VALUE_ATTRIBUTE_NAME;
+import org.apache.commons.lang.StringUtils;
```
elrodro83: we use commons-lang 3 now

PR #4998, opened 2017-06-13
MULE-12718: Performance degradation in MEL caused by MULE-11274

fsgonz:


tests/performance/src/main/java/org/mule/el/ExpressionBenchmark.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Benchmark
+    public Object melFlowVars()
+    {
+        return muleContext.getExpressionManager().evaluate("flowVars['foo']=='bar'", event);
+    }
+
+    @Benchmark
+    public Object dwFlowVars()
```
LucianoGandini: This is not a dw expression, can we rename both methods or delete one if the test does the same?

PR #5007, opened 2017-06-14
MULE-12562 - Reduce AbstractResourceConfigurationBuilder logging leve…

canoasan:
…l to DEBUG


core/src/main/java/org/mule/runtime/core/config/builders/AbstractResourceConfigurationBuilder.java
```diff
@@ -74,7 +74,7 @@ public void configure(MuleContext muleContext) throws ConfigurationException {
 
     super.configure(muleContext);
 
-    logger.info(CoreMessages.configurationBuilderSuccess(this, createConfigResourcesString()).toString());
+    logger.debug(CoreMessages.configurationBuilderSuccess(this, createConfigResourcesString()).toString());
```
elrodro83: static import

PR #5022, opened 2017-06-15
MULE-11220: Review the HttpEntity interface and its Implementations.

afelisatti:


afelisatti:
Tests currently running on CI.

modules/http-api/src/main/java/org/mule/runtime/http/api/domain/entity/HttpEntity.java
```diff
@@ -6,11 +6,70 @@
...
+   * Provides the entity's content as a stream. All streaming entities should provide their stream.
+   *
+   * @return an {@code InputStream} representing this entity's content or {@code null} if such representation is not possible
+   */
+  InputStream getContent();
+
+  /**
+   * Provides the entity's content as bytes. If the entity is stream based, then the stream will be consumed as a consequence.
+   *
+   * @return a byte array representing this entity's content or {@code null} if such representation is not possible
```
elrodro83: `or {@code null} if such representation is not possible` -> isn't the implementation returning an emty array in this case?

afelisatti: I'll check. If I remember correctly the only entity which did not allow stream not byte representation was MultipartHttpEntity. The rest, even EmptyHttpEntity, have ways of doing so.

afelisatti: Yes, MultipartHttpEntity signals with null that it cannot handle that representation.

PR #5023, opened 2017-06-15
Mule 9865

alepulver:
do not merge yet, tests running


modules/spring-config/src/main/java/org/mule/runtime/config/spring/MuleCustomEntityResolver.java
```diff
@@ -0,0 +1,127 @@
...
+    ClassLoader classLoaderToUse = classLoader;
+    if (classLoaderToUse == null) {
+      classLoaderToUse = ClassUtils.getDefaultClassLoader();
+    }
+    Enumeration<URL> urls =
+        (classLoaderToUse != null ? classLoaderToUse.getResources(resourceName) : ClassLoader.getSystemResources(resourceName));
+    Properties props = new Properties();
+    while (urls.hasMoreElements()) {
+      URL url = urls.nextElement();
+      URLConnection con = url.openConnection();
```
pablolagreca: Verify that this doesn't point to an HTTP location

alepulver: It doesn't, only loads mule.schemas (which are file:/)

PR #5029, opened 2017-06-16
MULE-12832: Avoid generating stacktraces for exceptions when verbose is

elrodro83:
disabled
* Add generics to exception info
* Migrate uses of deprecated `SystemUtils.LINE_SEPARATOR`
* Move test case from integration tests repo


core/src/main/java/org/mule/runtime/core/api/message/ExceptionPayload.java
```diff
@@ -19,8 +19,6 @@
 @Deprecated
 public interface ExceptionPayload extends Serializable {
 
-  int getCode();
```
pablokraan: Check if META-INF/org/mule/runtime/core/config/mule-exception-codes.properties and META-INF/org/mule/runtime/core/config/mule-exception-config.properties can be deleted too

PR #5013, opened 2017-06-14
MULE-12804: Improve WSC performance

juandesi:

PR #5031, opened 2017-06-16
MULE-12804: Improve WSC performance (#5013)

juandesi:


modules/extensions-soap-support/src/main/java/org/mule/runtime/module/extension/soap/internal/runtime/connection/ForwardingSoapClientConnectionProvider.java
```diff
@@ -33,7 +41,11 @@
  *
  * @since 4.0
  */
-public class ForwardingSoapClientConnectionProvider implements PoolingConnectionProvider<ForwardingSoapClient> {
+public class ForwardingSoapClientConnectionProvider implements PoolingConnectionProvider<ForwardingSoapClient>, Lifecycle {
```
dfeist: Still not sure why `PoolingConnectionProvider` is used for WSC, did you look into this?

PR #5038, opened 2017-06-19
MULE-11723: Flow should be immutable

elrodro83:


core/src/main/java/org/mule/runtime/core/api/processor/strategy/ProcessingStrategyFactory.java
```diff
@@ -19,4 +19,9 @@
    * @return a new {@link ProcessingStrategy}.
    */
   ProcessingStrategy create(MuleContext muleContext, String schedulersNamePrefix);
+
+  default Class<? extends ProcessingStrategy> getProcessingStrategyClass() {
```
dfeist: javadoc

core-tests/src/test/java/org/mule/runtime/core/api/processor/LoggerMessageProcessorTestCase.java
```diff
@@ -35,7 +37,7 @@
 
   @Before
   public void before() throws RegistrationException {
-    flow = builder("flow", mockContextWithServices()).build();
+    flow = builder("flow", mockContextWithServices()).messageProcessors(emptyList()).build();
```
dfeist: `messageProcessors` or `processors`, same with source?  Why is emptyList needed? Does builder not use empty list by default?

core/src/main/java/org/mule/runtime/core/api/processor/strategy/ProcessingStrategyFactory.java
```diff
@@ -19,4 +19,12 @@
    * @return a new {@link ProcessingStrategy}.
    */
   ProcessingStrategy create(MuleContext muleContext, String schedulersNamePrefix);
+
+  /**
+   * @return a reference to the concrete implememntation of the {@link ProcessingStrategy} this facotry will create.
+   */
+  default Class<? extends ProcessingStrategy> getProcessingStrategyClass() {
```
dfeist: - Minor: but `getProcessingStrategyType` would be slight better.
- Isn't it better to force implementations to implement this, given it's used in conditions in Flow/Pipeline?

PR #5055, opened 2017-06-21
MULE-12844 - Embedded container cannot delete folder when stopping ap…

pablolagreca:
…p in Windows


distributions/embedded/embedded-impl/src/main/java/org/mule/runtime/module/embedded/impl/EmbeddedController.java
```diff
@@ -158,6 +157,11 @@ public void stop() {
       } catch (Exception e) {
         LOGGER.debug("failure disposing application", e);
       }
+      try {
+        deleteTree(new File(containerInfo.getContainerBaseFolder().getPath()));
+      } catch (Exception e) {
+        LOGGER.debug("failure deleting container directory " + containerInfo.getContainerBaseFolder().getPath(), e);
```
elrodro83: error/warn instead of debug?

PR #5067, opened 2017-06-21
MULE-12888 Processor chain performs thread hand-off from async proces…

dfeist:
…sor response after executing interceptors.


core-tests/src/test/java/org/mule/runtime/core/processor/strategy/AbstractProcessingStrategyTestCase.java
```diff
@@ -600,5 +609,29 @@ protected boolean matchesSafely(MessagingException item) {
     };
   }
 
+  class AnnotatedAyncProcessor extends AbstractAnnotatedObject implements AnnotatedProcessor
```
afelisatti: Async

PR #5068, opened 2017-06-21
Mule 12320: App that fails deploying take too much time to show the error

lucianoRM:


modules/spring-config/src/main/java/org/mule/runtime/config/spring/ModuleDelegatingEntityResolver.java
```diff
@@ -93,7 +95,9 @@ public InputSource resolveEntity(String publicId, String systemId) throws SAXExc
     if (inputSource == null) {
       inputSource = springEntityResolver.resolveEntity(publicId, systemId);
     }
-
+    if (inputSource == null && !extensions.isEmpty()) {
```
alepulver: Add comment explaining that the code is sometimes called before loading extensions and after, so we have to allow a repetition in that case.

modules/spring-config/src/test/java/org/mule/runtime/config/spring/XmlConfigurationDocumentLoaderTestCase.java
```diff
@@ -54,6 +56,30 @@ public void testWellformedXml() {
...
+      assertThat(getStackTrace(e),
+                 StringContains.containsString("Can't resolve  http://www.mulesoft.org/schema/mule/co"));
+    }
+  }
+
+  @Test
+  public void testMalformedXmlWithWrongSchemaLocationAndNoExtensionsRaisesException() {
+    try {
+      getDocument("mule-wrong-schema-location.xml");
+      fail("Should not have reach here as the document is malformed and an exception should have been thrown using the default constructor");
```
alepulver: reach -> reached

PR #5069, opened 2017-06-21
MULE-12609: Every exported Mule package must contain api

pablokraan:


core/src/main/resources/META-INF/mule-module.properties
```diff
@@ -47,6 +51,7 @@ artifact.export.classPackages=org.mule.runtime.core,\
                               org.mule.runtime.core.api.security,\
                               org.mule.runtime.core.api.security.tls,\
                               org.mule.runtime.core.api.serialization,\
+                              org.mule.runtime.core.api.session,\
```
elrodro83: Shouldn't all session stuff be privileged api?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -37,9 +37,9 @@
 import org.mule.runtime.core.api.source.MessageSource;
 import org.mule.runtime.core.api.transaction.TransactionConfig;
 import org.mule.runtime.core.exception.ErrorTypeLocator;
-import org.mule.runtime.core.execution.ExceptionCallback;
-import org.mule.runtime.core.execution.MessageProcessContext;
-import org.mule.runtime.core.execution.MessageProcessingManager;
+import org.mule.runtime.core.internal.execution.ExceptionCallback;
```
elrodro83: shouldn't ExtensionMessageSource be api?

core/src/main/resources/META-INF/mule-module.properties
pablokraan: It must be api, functional tests depend on it

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
pablokraan: Is not required at API level, is only used by extension support modules

PR #5073, opened 2017-06-22

facundovs:


core/src/main/java/org/mule/runtime/core/processor/AbstractFilteringMessageProcessor.java
```diff
@@ -132,4 +148,9 @@ public boolean isThrowOnUnaccepted() {
   public void setThrowOnUnaccepted(boolean throwOnUnaccepted) {
     this.throwOnUnaccepted = throwOnUnaccepted;
   }
+
+  private boolean shouldNotFilterStopParentFlow() {
+    return getBoolean(FILTER_ON_UNACCEPTED_NOT_STOP_PARENT_FLOW);
```
dfeist: I don't think we should conserve this in 4.x.  What do you think @pablolagreca ?

PR #5077, opened 2017-06-22
MULE-12869: Failure to run SOAP-based apps: 'basic-connection' does n…

gsfernandes:
…ot have an associated Bean Definition Parser - Refactor code to extract logic to another class so it can be reused

### DIFF
--- a/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/artifact/ExtensionModelDiscoverer.java
+++ b/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/artifact/ExtensionModelDiscoverer.java
@@ -1,0 +1,141 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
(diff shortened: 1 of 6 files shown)


gsfernandes:
@pablolagreca @juandesi please review this refactor.

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/artifact/ExtensionModelDiscoverer.java
```diff
@@ -0,0 +1,141 @@
...
+      } else {
+        URL manifest = artifactPlugin.getSecond().findResource(META_INF_FOLDER + "/" + EXTENSION_MANIFEST_FILE_NAME);
+        if (manifest != null) {
+          //TODO: Remove when MULE-11136
+          if (LOGGER.isDebugEnabled()) {
+            LOGGER.debug("Discovered extension " + artifactName);
+          }
+          discoverExtensionThroughManifest(extensions, artifactClassloader, manifest);
+        } else {
+          LOGGER.warn("Extension [" + artifactName + "] could not being discovered");
```
elrodro83: being -> be

PR #5086, opened 2017-06-23
MULE-12923: Add a way to support custom notifications from plugins

elrodro83:


core/src/main/java/org/mule/runtime/core/api/context/notification/NotificationsProvider.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @since 4.0
+ */
+public interface NotificationsProvider {
+
+  /**
+   * The key of the returned map is the string representation of the notification. It must be in the format
+   * {@code [plugin]:[NOTIFICATION]}
+   * 
+   * @return the mapping of the notification id to the concrete types of {@link ServerNotification} and
```
pablokraan: id -> ID

core/src/main/java/org/mule/runtime/core/api/context/notification/NotificationsProvider.java
```diff
@@ -0,0 +1,29 @@
...
+   *         {@link ServerNotificationListener} it represents.
+   */
+  Map<String, Pair<Class<? extends ServerNotification>, Class<? extends ServerNotificationListener>>> getEventListenerMapping();
```
pablokraan: How do this work when plugins are installed in the domain instead on in the app? (or inside a policy)

modules/spring-config/src/main/java/org/mule/runtime/config/spring/NotificationConfig.java
```diff
@@ -128,50 +121,35 @@ public void setInterfaseName(String interfaseName) {
     this.interfaseName = interfaseName;
   }
 
-  public void setEventClass(Class eventClass) {
+  public String getInterfaseName() {
+    return interfaseName;
```
pablokraan: interfaseName > interfaceName

modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -1221,6 +1217,18 @@
         </xsd:restriction>
     </xsd:simpleType>
 
+    <xsd:simpleType name="extendedNotificationTypes">
+        <xsd:restriction base="xsd:NMTOKEN">
+            <xsd:pattern value="[a-z]+:[A-Z\-]+">
+                <xsd:annotation>
+                    <xsd:documentation>
+                        Allows plugins to define their own notifications, in format "[plugin]:[NOTIFICATION]".
```
pablokraan: Not sure if there will be other artifact providing notifications... but what aoubt using "artifactID" instead of "plugin"?

core/src/main/java/org/mule/runtime/core/api/context/notification/NotificationsProvider.java
elrodro83: for domains, the `ServerNotificationManagerConfigurator` that is initialized for the application will get the `NotificationsProvider` instance from the plugins in the domains.

PR #5093, opened 2017-06-26
MULE-12929: Mule Core Extensions aren't being stopped if RuntimeExcep…

facundovs:
…tions are triggered.

If a Runtime Exception is triggered in some extension.stop() execution (for instance NPE), Mule doesn't keep stopping the remaining extensions and not
finish the MuleContainer#stop() execution.
Among other consequences, this is causing shutdown time is not being honored.

### DIFF
--- a/modules/launcher/src/main/java/org/mule/module/launcher/coreextension/DefaultMuleCoreExtensionManager.java
+++ b/modules/launcher/src/main/java/org/mule/module/launcher/coreextension/DefaultMuleCoreExtensionManager.java
@@ -106,7 +106,7 @@ public class DefaultMuleCoreExtensionManager implements MuleCoreExtensionManager
             {
                 extension.stop();
             }
-            catch (Exception e)
+            catch (Throwable e)
             {
                 logger.warn("Error stopping core extension: " + extension.getName(), e);
             }
--- a/modules/launcher/src/main/java/org/mule/module/launcher/coreextension/DefaultMuleCoreExtensionManager.java
+++ b/modules/launcher/src/main/java/org/mule/module/launcher/coreextension/DefaultMuleCoreExtensionManager.java
@@ -106,7 +106,7 @@ public class DefaultMuleCoreExtensionManager implements MuleCoreExtensionManager
             {
                 extension.stop();
             }
-            catch (MuleException e)
+            catch (Exception e)
             {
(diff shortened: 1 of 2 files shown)


modules/launcher/src/main/java/org/mule/module/launcher/coreextension/DefaultMuleCoreExtensionManager.java
```diff
@@ -106,7 +106,7 @@ public void stop() throws MuleException
             {
                 extension.stop();
             }
-            catch (MuleException e)
+            catch (Exception e)
```
pablokraan: If you want to be sure that core extensions are stopped, you must catch Throwable

modules/launcher/src/test/java/org/mule/module/launcher/coreextension/DefaultMuleCoreExtensionManagerTestCase.java
```diff
@@ -211,6 +212,28 @@ public void resolvesCoreExtensionDependencies() throws Exception
         verify(coreExtensionDependencyResolver).resolveDependencies(extensions);
     }
 
+    @Test
+    public void testAllCoreExtensionsAreStoppedAfterRuntimeException() throws Exception
+    {
+        TestDeploymentServiceAwareExtension extensionFailsStops = mock(TestDeploymentServiceAwareExtension.class);
+        TestDeploymentServiceAwareExtension extension = mock(TestDeploymentServiceAwareExtension.class);
```
pablokraan: Make clear in the names that on extesion fails, then other works OK

modules/launcher/src/test/java/org/mule/module/launcher/coreextension/DefaultMuleCoreExtensionManagerTestCase.java
```diff
@@ -211,6 +212,28 @@ public void resolvesCoreExtensionDependencies() throws Exception
...
+        extensions.add(extension);
+        extensions.add(extensionFailsStops);
+        coreExtensionManager.initialise();
+        try
+        {
+            coreExtensionManager.stop();
+        }
+        finally
+        {
+            verify(extension).stop();
```
pablokraan: You must assert that the stop methods are invoked in order

PR #5094, opened 2017-06-26
MULE-11739: Make ImportedTypes not dependent on Extension name

pablomusumeci:


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/schema/ClasspathBasedDslContext.java
```diff
@@ -45,7 +46,7 @@
   private final ClassLoader classLoader;
   private final Map<String, Class<?>> extensionsByName = new HashMap<>();
   private final Map<String, ExtensionModel> resolvedModels = new HashMap<>();
-  private TypeCatalog typeCatalog;
+  private LazyValue<TypeCatalog> typeCatalog = new LazyValue<>(() -> TypeCatalog.getDefault(getExtensions()));
```
marianogonzalez: not against this.. but why?

modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/capability/xml/SchemaGeneratorTestCase.java
```diff
@@ -177,6 +178,16 @@ public void generate() throws Exception {
     }
 
     @Override
+    public Optional<ExtensionModel> getExtensionForType(String typeId) {
+      Optional<String> extension = getTypeCatalog().getDeclaringExtension(typeId);
+      if (!extension.isPresent()) {
```
marianogonzalez: .map(this::getExtension)

modules/extensions-spring-support/src/test/resources/models/subtypes.json
```diff
@@ -49,7 +49,25 @@
     {
       "baseType": {
         "format": "java",
-        "type": "@ref:org.mule.test.heisenberg.extension.model.Weapon"
+        "type": "Object",
+        "annotations": {
+          "typeId": "org.mule.test.heisenberg.extension.model.Weapon",
```
marianogonzalez: this should still be a ref. Otherwise there could be an inconsistency between what's described here and what's on the original extension

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/schema/ClasspathBasedDslContext.java
pablomusumeci: Actually, the value was lazy because it started as null and was calculated once you called `getTypeCatalog()`. I just made it explicitly lazy.

modules/extensions-spring-support/src/test/resources/models/subtypes.json
pablomusumeci: This happened because we are no longer adding the imported types to the extension's type catalog, so the type is defined multiple times (in the original extension's catalog and all the places where it's used in the other extensions that import the same type)

marianogonzalez: doesn't matter. Still needs to be a ref

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/schema/ClasspathBasedDslContext.java
marianogonzalez: actually... how will this work if it's no longer on the extension's catalog? How will the DSlContext get it? Did you try the app model -> xml test cases with this change?

pablomusumeci: TypeCatalog will be filled with all the extensions declared types. DslContext will search any typeId and find the extension which is the owner (aka the one who contributed with) that type.

PR #5097, opened 2017-06-26
MULE-12919: Improve FTP Receiver Reconnection.

facundovs:
In FTP Polling, Reconnection policy is applied in each poll. However, if a InterruptedIOException is triggered, Mule Logs shows the following message:
 "Process was interrupted (InterruptedException), ceasing process".

If in the next polling, Connection is successful, logs only shows it in debug mode.
This is causing confussion in some customers that after the failure message, think that the polling was cancelled.
Additionally, it is needed to call FTPMessageReceiver#filesToFTPArray(), only if the client creation was succesful to avoid NPE.

### DIFF
--- a/core/src/main/java/org/mule/retry/policies/AbstractPolicyTemplate.java
+++ b/core/src/main/java/org/mule/retry/policies/AbstractPolicyTemplate.java
@@ -37,6 +37,8 @@ public abstract class AbstractPolicyTemplate implements RetryPolicyTemplate, Mul
 
     private MuleContext muleContext;
 
+    private boolean lastRetryInterrupted = false;
+
     protected transient final Log logger = LogFactory.getLog(getClass());
 
     public void setMuleContext(MuleContext context)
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mule/retry/policies/AbstractPolicyTemplate.java
```diff
@@ -63,6 +65,13 @@ public RetryContext execute(RetryCallback callback, WorkManager workManager) thr
                     {
                         notifier.onSuccess(context);
                     }
+                    // Needed for Polling Connectors as FTP.
+                    if(lastRetryInterrupted)
```
pablokraan: Fix format

core/src/main/java/org/mule/retry/policies/AbstractPolicyTemplate.java
```diff
@@ -37,6 +37,8 @@
 
     private MuleContext muleContext;
 
+    private boolean lastRetryInterrupted = false;
```
pablokraan: Add volatile

PR #5109, opened 2017-06-27
MULE-9422: Remove <message-filter> functionality overlap

elrodro83:


core/src/main/java/org/mule/runtime/core/routing/Foreach.java
```diff
@@ -194,4 +203,7 @@ public void setCounterVariableName(String counterVariableName) {
     this.counterVariableName = counterVariableName;
   }
 
+  public void setIgnoreErrorType(String ignoreErrorType) {
```
dfeist: jdoc

modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -4117,6 +4072,13 @@
             <xsd:complexContent>
                 <xsd:extension base="baseSplitterType">
                 <xsd:attributeGroup ref="optionalExpressionAttributes"/>
+                <xsd:attribute name="filterOnErrorType" type="errorIdentifierType" use="optional">
+                    <xsd:annotation>
+                        <xsd:documentation>
+                            The full name of the error types or a comma separated list of full names, to filter out of the resulting payload when it occurs. This is useful to use validations inside this component.
```
dfeist: Improve with something like: A comma separated list error types for that should be handled by dropping the split part, such that it won't be included in the aggregated response collection, rather than propagating the error.

modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -4228,6 +4190,13 @@
                         </xsd:documentation>
                     </xsd:annotation>
                 </xsd:attribute>
+                <xsd:attribute name="ignoreErrorType" type="errorIdentifierType" use="optional">
+                    <xsd:annotation>
+                        <xsd:documentation>
+                            The full name of the error types or a comma separated list of full names, to ignore when processing an item. This is useful to use validations inside this component.
```
dfeist: Improve with something like: A comma separated list of error types for that should be handled by filtering the collection item being processed, such that it is excluded in the resulting collection, rather than propagating the error.

modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -4228,6 +4190,13 @@
                         </xsd:documentation>
                     </xsd:annotation>
                 </xsd:attribute>
+                <xsd:attribute name="ignoreErrorType" type="errorIdentifierType" use="optional">
```
dfeist: not `filterOnErrorType`?  Isn't it filtering it from resulting collection? Or did you determine that it's read-only with respect to the original collection in the end?

core/src/main/java/org/mule/runtime/core/routing/outbound/AbstractMessageSequenceSplitter.java
```diff
@@ -34,7 +37,7 @@
  * @see AbstractSplitter
  */
 public abstract class AbstractMessageSequenceSplitter extends AbstractInterceptingMessageProcessor
-    implements MuleContextAware {
+    implements MuleContextAware, Acceptor {
```
dfeist: Compose instead of implement, because it's confusing that the splitter implement this, when in fact it is used to handle results from executed paths instead.

core/src/main/java/org/mule/runtime/core/routing/Splitter.java
```diff
@@ -43,9 +51,26 @@ public void initialise() throws InitialisationException {
...
+    if (filterOnErrorTypeMatcher == null) {
+      filterOnErrorTypeMatcher = createErrorType(muleContext.getErrorTypeRepository(), filterOnErrorType);
+    }
   }
 
   public void setExpression(String expression) {
     this.config.setExpression(expression);
   }
+
+  public void setFilterOnErrorType(String filterOnErrorType) {
```
dfeist: jdoc?

modules/spring-config/src/main/resources/META-INF/mule.xsd
elrodro83: no, after the foreach the original collection is the one that continues processing. i checked this

modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -4193,7 +4193,7 @@
                 <xsd:attribute name="ignoreErrorType" type="errorIdentifierType" use="optional">
                     <xsd:annotation>
                         <xsd:documentation>
-                            The full name of the error types or a comma separated list of full names, to ignore when processing an item. This is useful to use validations inside this component.
+                            A comma separated list of error types that should be handled by dropping the split part. These errors are handled so that items that cause them when being processed are ignored, rather than propagating the error. This is useful to use validations inside this component.
```
dfeist: delete `by dropping the split part. These errors are handled`

modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -4075,7 +4075,7 @@
                 <xsd:attribute name="filterOnErrorType" type="errorIdentifierType" use="optional">
                     <xsd:annotation>
                         <xsd:documentation>
-                            The full name of the error types or a comma separated list of full names, to filter out of the resulting payload when it occurs. This is useful to use validations inside this component.
+                            A comma separated list of error types that should be handled by dropping the split part. These errors are handled so that items that cause them when being processed won't be included in the aggregated response collection, rather than propagating the error. This is useful to use validations inside this component.
```
dfeist: Still not that clear... 
`A comma separated list of error types that should be handled by filtering the split part, rather than propagating the error.  Filtered parts are not included in the aggregated response collection. This is useful to use validations inside this component.`

PR #5111, opened 2017-06-28
MULE-12931: Cannot create application model for an app with APIKit

gsfernandes:


gsfernandes:
@alegmarra @marianogonzalez please check this, I ended up creating a private ModelProperty to identify if the extensionModel has be registered within a ComponentBuildingDefinition.

PR #5112, opened 2017-06-28
MULE-10351: WSC Should use http requester to resolve WSDL if specified.

juandesi:


modules/soap-api/src/main/java/org/mule/runtime/soap/api/transport/HttpResourceLocator.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+
+/**
```
marianogonzalez: remove

modules/soap-api/src/main/java/org/mule/runtime/soap/api/transport/NullTransportResourceLocator.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ */
+package org.mule.runtime.soap.api.transport;
+
+import java.io.InputStream;
+
+/**
+ * Null implementation of {@link TransportResourceLocator}. It does not handle any resource.
+ *
+ * @since 1.0
```
marianogonzalez: 4.0

PR #5119, opened 2017-06-29
MULE-12946: ClassLoader leak on DeclarationEnrichers generate an OOM …

alegmarra:
…Metaspace when calculating Metadata for DataSense

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/DynamicMetadataDeclarationEnricher.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/DynamicMetadataDeclarationEnricher.java
@@ -6,7 +6,6 @@
  */
 package org.mule.runtime.module.extension.internal.loader.enricher;
 
-import static java.lang.Thread.currentThread;
 import static org.mule.runtime.api.meta.model.display.LayoutModel.builderFrom;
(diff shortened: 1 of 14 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/DynamicMetadataDeclarationEnricher.java
```diff
@@ -56,17 +56,29 @@
...
+    private ClassTypeLoader typeLoader;
 
+    public DynamicMetadataDeclarer(ExtensionLoadingContext extensionLoadingContext,
+                                   ImplementingTypeModelProperty implementingType) {
+      this.extensionLoadingContext = extensionLoadingContext;
+      this.extensionType = implementingType.getType();
+    }
+
+    private void apply() {
+      typeLoader = ExtensionsTypeLoaderFactory.getDefault().createTypeLoader(currentThread().getContextClassLoader());
```
gsfernandes: Perhaps you could get the class loader from the context instead of assuming that is is set in the TCCL?
 https://github.com/mulesoft/mule-extensions-api/blob/master/mule-extensions-api/src/main/java/org/mule/runtime/extension/api/loader/ExtensionLoadingContext.java#L117

alegmarra: yup, missed that one

PR #5127, opened 2017-06-29
MULE-9825 - Add support for property placeholders in mule configuration

pablolagreca:


core/src/main/resources/META-INF/mule-module.properties
```diff
@@ -137,7 +138,23 @@ artifact.export.classPackages=org.mule.runtime.core,\
                               org.mule.mvel2.optimizers.impl.refl.nodes,\
                               org.mule.mvel2.util,\
                               org.mule.runtime.core.internal.enricher,\
-                              org.mule.runtime.core.internal.exception
+                              org.mule.runtime.core.internal.exception,\
+                              org.yaml.snakeyaml.parser,\
```
gsfernandes: Would it be possible to avoid exposing this packages as yaml would be an implementation of the configuration properties provider/reader?

gsfernandes: Ok, it is a limitation due to spring already is exposed and now these packages have to be exported.

PR #5134, opened 2017-06-30
MULE-12973: Add support for start parameter in http multipart/related…

facundovs:
… response.

According to [this RFC article](https://tools.ietf.org/html/rfc2387), in multipart/related content type:

The start parameter, if given, is the content-ID of the compound object's "root".
If not present the "root" is the first body part in the Multipart/Related entity.

Additionally, we shoud also honor the following restrictions:
- Type parameter is mandatory in multipart/related.
- The value of the content-type parameter must be the same that the root body part's content-type.

### DIFF
--- a/modules/http/src/main/java/org/mule/module/http/internal/multipart/HttpMultipartEncoder.java
+++ b/modules/http/src/main/java/org/mule/module/http/internal/multipart/HttpMultipartEncoder.java
@@ -6,8 +6,12 @@
  */
 package org.mule.module.http.internal.multipart;
 
+import static org.mule.config.i18n.MessageFactory.createStaticMessage;
 import static org.mule.module.http.api.HttpHeaders.Names.CONTENT_DISPOSITION;
+import static org.mule.module.http.api.HttpHeaders.Names.CONTENT_ID;
 import static org.mule.module.http.api.HttpHeaders.Names.CONTENT_TYPE;
+import static org.mule.module.http.internal.HttpParser.getContentTypeParameter;
+import org.mule.api.MuleRuntimeException;
 import org.mule.module.http.internal.HttpParser;
 import org.mule.module.http.internal.domain.MultipartHttpEntity;
 import org.mule.util.IOUtils;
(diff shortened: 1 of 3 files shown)


modules/http/src/main/java/org/mule/module/http/internal/multipart/HttpMultipartEncoder.java
```diff
@@ -50,20 +74,45 @@ public static MimeMultipart createMultpartContent(MultipartHttpEntity body, Stri
...
             }
         }
+
+        try
+        {
+            verifyNoAmbiguousMultipartRelatedType(contentTypeSubType, contentType, mimeMultipartContent);
+        }
+        catch (MessagingException e)
+        {
+            throw new RuntimeException(e);
```
afelisatti: Shouldn't this be MuleRuntimeException?

modules/http/src/main/java/org/mule/module/http/internal/multipart/HttpMultipartEncoder.java
```diff
@@ -29,10 +33,30 @@
...
+    }
+
+    public static MimeMultipart createContent(MultipartHttpEntity body, String contentType)
+    {
+
         String contentTypeSubType = HttpParser.getContentTypeSubType(contentType);
-        MimeMultipart mimeMultipartContent = new HttpMimeMultipart(contentType, contentTypeSubType);
+        MimeMultipart mimeMultipartContent ;
+
+        if (contentTypeSubType.equals(RELATED) && getContentTypeParameter(contentType, "type") == null)
```
afelisatti: Extract "type" and "start" to constants.

modules/http/src/main/java/org/mule/module/http/internal/multipart/HttpMultipartEncoder.java
```diff
@@ -85,10 +134,22 @@ private static String getContentDisposition(HttpPart part)
...
     {
-        MimeMultipart mimeMultipartContent = HttpMultipartEncoder.createMultpartContent(multipartEntity, contentType);
+        MimeMultipart mimeMultipartContent = HttpMultipartEncoder.createContent(multipartEntity, contentType);
         final ByteArrayOutputStream byteArrayOutputStream;
         byteArrayOutputStream = new ByteArrayOutputStream();
         mimeMultipartContent.writeTo(byteArrayOutputStream);
         return byteArrayOutputStream.toByteArray();
     }
+
+    private static void verifyNoAmbiguousMultipartRelatedType(String subContentType, String contentType, MimeMultipart mimeMultipartContent) throws MessagingException
```
afelisatti: Why extract this? if it was inline you could reuse the type parameter.

modules/http/src/test/java/org/mule/module/http/internal/multipart/HttpMultipartEncoderTestCase.java
```diff
@@ -0,0 +1,121 @@
...
+        HttpPart thirdPart = new HttpPart("thirdPart", THIRD_PART_CONTENT.getBytes(), "text/plain", THIRD_PART_CONTENT.getBytes().length);
+        httpParts.add(thirdPart);
+
+        when(multipartHttpEntity.getParts()).thenReturn(httpParts);
+    }
+
+    @Test
+    public void createMultipartRelatedContentWithStartParameter() throws Exception
+    {
+        MimeMultipart mimeMultipart = createContent(multipartHttpEntity, "multipart/related; type=\"text/plain\"; start=thirdPart");
```
afelisatti: Why is there no boundary here?

modules/http/src/test/java/org/mule/module/http/internal/multipart/HttpMultipartEncoderTestCase.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Test
+    public void createMultipartRelatedContentWithoutMandatoryTypeParameter() throws Exception
+    {
+        try
+        {
+            createContent(multipartHttpEntity, "multipart/related");
+            fail("Exception caused by no present type should be triggered");
+
```
afelisatti: Extra space

PR #5138, opened 2017-06-30
MULE-9249: Create @Path to allow users to use file/directory pickers as editors for their mule extensions

juandesi:


modules/extensions-spring-support/src/test/resources/models/marvel.json
```diff
@@ -1,194 +1,641 @@
 {
-  "name": "Marvel",
-  "description": "",
+  "name": "petstore",
```
marianogonzalez: something tells me you didn't run the tests

juandesi: I did run the tests, I misscopied the content :D

PR #5144, opened 2017-07-02
MULE-10401: Remove HTTP parsing mechanism.

afelisatti:


tests/component-plugin/src/main/java/org/mule/functional/config/TestComponentBuildingDefinitionProvider.java
```diff
@@ -88,6 +91,13 @@ public void init() {
         .build());
 
     componentBuildingDefinitions.add(baseDefinition.copy()
+        .withIdentifier("exception-strategy")
```
elrodro83: why readd this?

tests/component-plugin/src/main/java/org/mule/functional/config/TestComponentBuildingDefinitionProvider.java
```diff
@@ -149,6 +159,12 @@ public void init() {
         .withTypeDefinition(fromType(SharedConfig.class))
         .build());
 
+    componentBuildingDefinitions
```
elrodro83: why this change?

tests/component-plugin/src/main/resources/META-INF/mule-test.xsd
```diff
@@ -301,4 +301,40 @@
         </xsd:complexContent>
     </xsd:complexType>
 
+    <xsd:element name="set-payload" type="setPayloadTransformerType" substitutionGroup="mule:abstract-message-processor">
```
elrodro83: why this?

afelisatti: Since `set-payload` is going to become unavailable here soon and `transform` won't be available in CE I thought we should have a test version to make tests (and their migration) easier. When the `set-payload` task is done the Processor will just need to be moved to the test plugin.

PR #5152, opened 2017-07-03
MULE-11876: Provide a standard way to build a BindingContext from an

elrodro83:
InterceptionEvent

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/el/BindingContextUtils.java
+++ b/core/src/main/java/org/mule/runtime/core/el/BindingContextUtils.java
@@ -63,8 +63,6 @@ public class BindingContextUtils {
 
     BindingContext.Builder contextBuilder = BindingContext.builder(baseContext);
 
-    contextBuilder.addBinding(ID, new TypedValue<>(event.getContext().getId(), STRING));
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/mule/runtime/core/el/BindingContextUtils.java
```diff
@@ -0,0 +1,93 @@
...
+   */
+  public static BindingContext addEventBindings(Event event, BindingContext baseContext) {
+    requireNonNull(event);
+    requireNonNull(baseContext);
+
+    BindingContext.Builder contextBuilder = BindingContext.builder(baseContext);
+
+    contextBuilder.addBinding(ID, new TypedValue<>(event.getContext().getId(), STRING));
+    contextBuilder.addBinding(CORRELATION_ID, new TypedValue<>(event.getContext().getCorrelationId(), STRING));
+    Map<String, TypedValue> flowVars = new HashMap<>();
```
afelisatti: If I remember correctly there's an issue regarding what to do when the user tries to set a variable with the same name as ours, until that is done we should keep the behaviour of setting the flow vars first and then all of our bindings, so they are overridden. Can you fix that for ID and CORRELATION ID?

PR #5156, opened 2017-07-04
MULE-11167 Policies should be non blocking

dfeist:


core-tests/src/test/java/org/mule/runtime/core/processor/MessageProcessorsTestCase.java
```diff
@@ -0,0 +1,265 @@
...
+
+  @Test
+  public void processToApplyRespondAndMapInFlow() throws Exception {
+    assertThat(processToApply(input, createFlow(respondAndMap)).getMessage(), is(response.getMessage()));
+    assertThat(from(input.getContext().getResponsePublisher()).block(), is(response));
+  }
+
+  @Test
+  public void processToApplyError() throws Exception {
+    try {
```
elrodro83: use the expected rule. Do the last assertion in a finally block

PR #5167, opened 2017-07-05
EE-5564: AbstractReturnDelegate should return message with correct DataType for collections

mbuchwald:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/AbstractReturnDelegate.java
```diff
@@ -74,9 +74,11 @@ protected Message toMessage(Object value, ExecutionContextAdapter operationConte
       return MessageUtils.toMessage((Result) value, mediaType, cursorProviderFactory, event);
     } else {
       if (value instanceof Collection && returnsListOfMessages) {
-        value = toMessageCollection((Collection<Result>) value, cursorProviderFactory, event);
+        Collection<Message> messageCollection = toMessageCollection((Collection<Result>) value, cursorProviderFactory, event);
+        value = Message.builder().collectionPayload(messageCollection, Message.class).build();
```
marianogonzalez: shouldn't you also provide information on the Message itself? It's useless to the user to just say this is a list of messages if you don't also give information on the payload and attribute data types

PR #5181, opened 2017-07-06
MULE-9925 - Remove default app from distribution

pablolagreca:


distributions/standalone/assembly-whitelist.txt
```diff
@@ -14,20 +14,6 @@
 /mule-standalone-${productVersion}/MIGRATION.txt
 
 #
-# /apps
-#
-/mule-standalone-${productVersion}/apps/default
```
pablokraan: Should not include /mule-enterprise-standalone-${productVersion}/apps in the list?

distributions/standalone/src/main/resources/apps/IGNORE
```diff
@@ -0,0 +1,3 @@
+Git does not allow versioning of an empty folder so we add this file here to create the folder in git.
```
pablokraan: This file should not go into the distribution. If there is not way to remove the file and keep the folder in the distro, the rename the file to readme.txt of something like that

pablolagreca: This file doesn't go into the distro. Is the same as we do in domains. We put it here so git can track the folder but then we ignore it in assembly.xml

distributions/standalone/assembly-whitelist.txt
pablolagreca: Is in the list above

PR #5185, opened 2017-07-06
MULE-11575 Migrate use of blocking API by transports to use non-block…

dfeist:
…ing API that supports new interceptors. Step 2. REVIEW ONLY.


core/src/main/java/org/mule/runtime/core/processor/chain/ModuleOperationMessageProcessorChainBuilder.java
```diff
@@ -146,21 +147,11 @@ protected MessageProcessorChain createInterceptingChain(Processor head, List<Pro
       return result;
     }
 
-    /**
+    /*
```
elrodro83: ?

PR #5202, opened 2017-07-07
MULE-12826: ParameterModel should be able to point to a ComponentModel

juandesi:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/ParameterTypeModelValidator.java
```diff
@@ -39,12 +43,35 @@ public void validate(ExtensionModel extensionModel, ProblemsReporter problemsRep
...
+        validateParameterType(model, problemsReporter);
+        validateParameterWithReferencesType(model, problemsReporter);
       }
     }.walk(extensionModel);
   }
 
-  private void validateParameter(ParameterModel parameter, ProblemsReporter problemsReporter) {
+  private void validateParameterWithReferencesType(ParameterModel parameter, ProblemsReporter problemsReporter) {
+    if (!parameter.getElementReferences().isEmpty()) {
+      parameter.getType().accept(new MetadataTypeVisitor() {
```
estebanwasinger: This can be simply done with `if(!(parameter.getType() instanceOf StringType)) addError`

PR #5203, opened 2017-07-07
MULE-11599: Support defining DW functions

alegmarra:


modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/config/SingleConfigParserTestCase.java
```diff
@@ -25,11 +25,6 @@
 
 public class SingleConfigParserTestCase extends AbstractConfigParserTestCase {
 
-  @Override
-  protected boolean isDisposeContextPerClass() {
```
marianogonzalez: can you bring these back now that the leak was spotted?

PR #5213, opened 2017-07-09
MULE-11686: Move component and it's dependencies to compatibility module

elrodro83:
* JavaComponent moved


core/src/main/java/org/mule/runtime/core/api/component/Component.java
```diff
@@ -7,18 +7,22 @@
...
  * A <code>Component</code> component processes a {@link Event} by invoking the component instance that has been configured,
  * optionally returning a result.
  * <p/>
  * Implementations of <code>Component</code> can use different types of component implementation, implement component instance
  * pooling or implement <em>bindings</em> which allow for service composition.
+ * 
+ * @deprecated Transport infrastructure is deprecated.
  */
+@Deprecated
 public interface Component extends Processor, FlowConstructAware {
```
pablokraan: Can this class be deleted or moved to privileged API?

elrodro83: Not without some other refactoring. It is used in `org.mule.runtime.core.internal.lifecycle.phases.MuleContext*Phase` to configure the lifecycle of the components in the registry (same as `LegacyConnector`).
I created https://www.mulesoft.org/jira/browse/MULE-13023 to track this.

PR #5228, opened 2017-07-11
MULE-11575 Migrate use of blocking API by transports to use non-blocking API that supports new interceptors. Step 2.

dfeist:


core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
```diff
@@ -317,15 +318,22 @@ public Event process(Event event) throws MuleException {
           .fireNotification(new PipelineMessageNotification(createInfo(event, null, AbstractPipeline.this), AbstractPipeline.this,
                                                             PROCESS_START));
 
+      long startTime = System.currentTimeMillis();
```
elrodro83: static import

core/src/main/java/org/mule/runtime/core/internal/exception/AbstractExceptionListener.java
```diff
@@ -119,15 +118,6 @@ private Throwable getCause(Exception ex) {
     return ex.getCause() instanceof TypedException ? ex.getCause().getCause() : ex.getCause();
   }
 
-  protected void closeStream(Message message) {
```
elrodro83: is this functionality needed elsewhere?

core/src/main/java/org/mule/runtime/core/internal/exception/ErrorHandler.java
```diff
@@ -78,9 +78,10 @@ public Event handleException(MessagingException exception, Event event) {
       return error(exception);
     } else {
       Event event = addExceptionPayload(exception, exception.getEvent());
+      exception.setProcessedEvent(event);
       for (MessagingExceptionHandlerAcceptor exceptionListener : exceptionListeners) {
         if (exceptionListener.accept(event)) {
-          return exceptionListener.apply(new MessagingException(event, exception));
+          return exceptionListener.apply(exception);
```
elrodro83: is this change needed? we should aim to make `MessagingException` immutable eventually

core/src/main/java/org/mule/runtime/core/internal/exception/TemplateOnErrorHandler.java
```diff
@@ -64,24 +63,64 @@
...
-        sink.next(event);
-      } else {
-        exception.setProcessedEvent(event);
-        sink.error(exception);
-      }
-    });
+    return applyInternal(exception, exception.getEvent());
+  }
+
+  private Publisher<Event> applyInternal(final MessagingException exception, Event event) {
```
elrodro83: @afelisatti i need your help to review this

core/src/main/java/org/mule/runtime/core/internal/exception/TemplateOnErrorHandler.java
```diff
@@ -64,24 +63,64 @@
...
+    return just(event)
+        .doOnNext(request -> {
```
elrodro83: can some of these blocks be refactored to different methods for better readability?

core/src/main/java/org/mule/runtime/core/internal/policy/AbstractCompositePolicy.java
```diff
@@ -67,7 +67,7 @@ public AbstractCompositePolicy(List<Policy> policies,
    */
   public final Publisher<Event> processPolicies(Event operationEvent) {
     return just(operationEvent)
-        .flatMapMany(event -> processWithChildContext(event, new AbstractCompositePolicy.NextOperationCall()));
+        .flatMapMany(event -> processWithChildContext(event, new AbstractCompositePolicy.NextOperationCall(), null));
```
elrodro83: isn't there a ComponentLocation available here? @pablolagreca ?

core/src/main/java/org/mule/runtime/core/internal/policy/CompositeSourcePolicy.java
```diff
@@ -82,7 +82,7 @@ public CompositeSourcePolicy(List<Policy> parameterizedPolicies,
   @Override
   protected Publisher<Event> processNextOperation(Event event) {
     return just(event)
-        .then(request -> from(processWithChildContext(request, flowExecutionProcessor)))
+        .then(request -> from(processWithChildContext(request, flowExecutionProcessor, null)))
```
elrodro83: same as above

core/src/main/java/org/mule/runtime/core/internal/policy/AbstractCompositePolicy.java
pablolagreca: There may be something wrong with that API. If there are use cases where the processWithChildContext method may be used outside of an element in the config, then it may not be a component location. In this case there's not since the AbstractCompositePolicy represents the chain of ALL the policies that applied to a source or an operation and that concept is not reflected in the XML.

core/src/main/java/org/mule/runtime/core/internal/exception/ErrorHandler.java
dfeist: Without this change, MessageRedeliveredException gets wrapped and tests fail as they test for specific exception instance. (MessageRedeliveredException extends MessagingException)

core/src/main/java/org/mule/runtime/core/internal/exception/AbstractExceptionListener.java
dfeist: StreamCloser is still used in batch, so i left infrastructure in place. Also when we implement auto-closing of streams for non-repeatable streams we may want to use StreamCloserService.

core/src/main/java/org/mule/runtime/core/AbstractEventContext.java
```diff
@@ -143,6 +145,21 @@ public final void error(Throwable throwable) {
     }
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Publisher<Void> error(MessagingException messagingException, MessagingExceptionHandler handler) {
```
dfeist: This is part of a wider change, if we decide to not propagate MessagingExceptionHandler down though MP's.  If we do this it will be in a separate PR.

core/src/main/java/org/mule/runtime/core/api/rx/Exceptions.java
```diff
@@ -158,6 +158,19 @@ public static MuleException rxExceptionToMuleException(Throwable throwable)
   }
 
   /**
+   * Unwrap a particular {@code Throwable}.
+   * 
+   * @param throwable the exception to wrap
+   * @return the unwrapped exception
+   */
+  public static Throwable unwrap(Throwable throwable) {
+    while (throwable.getClass().getName().equals(REACTIVE_EXCEPTION_CLASS_NAME)) {
```
dfeist: Before we used an implementation of unwrap exposed by reactor, but there are class-loader issues which means that sometimes `instanceof` fails and the exception does not get unwrapped. I discussed this with @pablokraan

core/src/main/java/org/mule/runtime/core/api/util/ExceptionUtils.java
```diff
@@ -48,13 +51,32 @@
...
 
+
+  /**
+   * Null {@link MessagingExceptionHandler} which can we used to configure a {@link MessageProcessorChain} to not handle errors.
+   */
+  public static final MessagingExceptionHandler NULL_ERROR_HANDLER = new MessagingExceptionHandler() {
+
+    @Override
+    public Event handleException(MessagingException exception, Event event) {
+      throw new RuntimeException(exception);
```
dfeist: This matches behaviour in `TemplateOnErrorHandler.java`. Why it throws `RuntimeException` i'm not sure. @afelisatti ?

core/src/main/java/org/mule/runtime/core/internal/policy/AbstractCompositePolicy.java
dfeist: I added `ComponentLocation` initially just for logging as when a child context is used it's valuable to know who create it (i.e. what component). I was expecting that in some cases it wouldn't be available.  Is there a NullComponentLocation that can be used here instead?

core/src/main/java/org/mule/runtime/core/AbstractEventContext.java
elrodro83: Create a Jira and add a TODO

core/src/main/java/org/mule/runtime/core/internal/exception/ErrorHandler.java
elrodro83: fila e jira and add a TODO to track this

dfeist: This isn't a bug IMO, this was a regression in a change i made before.  We shouldn't blindly wrap with MessagingException when we have other implementations.

PR #5231, opened 2017-07-11
MULE-12785: Remove the @DataTypeParameters annotation

juandesi:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/MimeTypeParametersDeclarationEnricher.java
```diff
@@ -0,0 +1,132 @@
...
+      }
+
+      @Override
+      public void visitBinaryType(BinaryType binaryType) {
+        holder.set(true);
+      }
+
+      @Override
+      public void visitObject(ObjectType objectType) {
+        holder.set(getId(objectType).equals(Object.class.getName()));
```
marianogonzalez: why this? Doesn't look good to me

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/MimeTypeParametersDeclarationEnricher.java
```diff
@@ -0,0 +1,132 @@
...
+  private final ClassTypeLoader typeLoader = ExtensionsTypeLoaderFactory.getDefault().createTypeLoader();
+
+  @Override
+  public void enrich(ExtensionLoadingContext extensionLoadingContext) {
+    final ExtensionDeclaration declaration = extensionLoadingContext.getExtensionDeclarer().getDeclaration();
+    new IdempotentDeclarationWalker() {
+
+      @Override
+      protected void onOperation(OperationDeclaration declaration) {
+        getImplementingMethod(declaration).ifPresent(method -> {
```
marianogonzalez: You should not need the implementing method for this. In the metadata type you have all the necessary information to do this. This enricher should be in the ExtensionFactory, not a custom one

PR #5237, opened 2017-07-12
MULE-12994 Support flag -Dmule.timeout.disable=true in HTTP Listeners

dfmacias:


core/src/main/java/org/mule/runtime/core/api/config/MuleProperties.java
```diff
@@ -208,4 +208,5 @@
   public static final String MULE_FLOW_TRACE = SYSTEM_PROPERTY_PREFIX + "flowTrace";
   public static final String MULE_LOG_VERBOSE_CLASSLOADING = SYSTEM_PROPERTY_PREFIX + "classloading.verbose";
   public static final String MULE_MEL_AS_DEFAULT = SYSTEM_PROPERTY_PREFIX + "test.mel.default";
+  public static final String MULE_DISABLE_RESPONSE_TIMEOUT = MuleProperties.SYSTEM_PROPERTY_PREFIX + "timeout.disable";
```
pablokraan: Use static import

PR #5246, opened 2017-07-13
MULE-13010: SDK should no longer deal in terms of Attributes

alegmarra:


core-tests/src/test/java/org/mule/runtime/core/message/DefaultMuleMessageBuilderTestCase.java
```diff
@@ -42,52 +43,51 @@
...
 
-/**
- *
- */
 public class DefaultMuleMessageBuilderTestCase extends AbstractMuleTestCase {
 
   private static final String NEW_PAYLOAD = "new payload";
-  private static final Attributes TEST_ATTR = NULL_ATTRIBUTES;
-  private static final Attributes TEST_ATTR_2 = new BaseAttributes() {};
+  private static final Object BASE_ATTRIBUTES = new BaseAttributes() {};
```
marianogonzalez: this BaseAttributes class should not be used inside extensions. It's in core-api which extensions should not see.

PR #5247, opened 2017-07-13
MULE-12995: SSLSession is not being propagated to jaxws service.

fsgonz:


fsgonz:
I cancel this PR. This needs a change in strategy for the serialization of the SSLContext.

PR #5253, opened 2017-07-13
MULE-12785: Remove the @DataTypeParameters annotation

juandesi:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/MimeTypeParametersDeclarationEnricher.java
```diff
@@ -64,60 +51,41 @@ public void enrich(ExtensionLoadingContext extensionLoadingContext) {
 
       @Override
       protected void onOperation(OperationDeclaration declaration) {
-        getImplementingMethod(declaration).ifPresent(method -> {
-          if (!isVoid(method) && acceptsMimeTypeParameters(getMethodReturnType(method, typeLoader))) {
-            declareMimeTypeParameters(declaration);
-          }
-        });
+        declareMimeTypeParameters(declaration);
```
marianogonzalez: this is good but incomplete. This enricher needs to move to ext-api

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/AbstractJavaExtensionModelLoader.java
```diff
@@ -24,7 +24,7 @@
 import org.mule.runtime.module.extension.internal.loader.enricher.BooleanParameterDeclarationEnricher;
 import org.mule.runtime.module.extension.internal.loader.enricher.ConfigNameDeclarationEnricher;
 import org.mule.runtime.module.extension.internal.loader.enricher.ConnectionDeclarationEnricher;
-import org.mule.runtime.module.extension.internal.loader.enricher.MimeTypeParametersDeclarationEnricher;
+import org.mule.runtime.extension.internal.loader.enricher.MimeTypeParametersDeclarationEnricher;
```
marianogonzalez: should be removed from here

PR #5257, opened 2017-07-14
MULE-12949: Refresh socket extension

pablomusumeci:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/NullSafeValueResolverWrapper.java
```diff
@@ -109,8 +112,9 @@ public void visitObject(ObjectType objectType) {
           }
 
           Optional<String> defaultValue = getDefaultValue(objectField);
-          if (defaultValue.isPresent()) {
-            resolver = getFieldDefaultValueValueResolver(objectField, muleContext);
+          if (defaultValue.isPresent() || field.getAnnotation(DefaultEncoding.class) != null) {
+            resolver = getDefaultValueResolver(field.getAnnotation(DefaultEncoding.class) != null, muleContext,
```
marianogonzalez: i am willing to accept this for the time being in order to move forward, but this is a bad solution. We need to have a new component which allows us to pile up all this types of resolvers and apply them anywhere. 

We can't keep adding these ifs everywhere. All of that needs to be centralized. File a jira for that and add a todo

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ParametersResolver.java
```diff
@@ -192,15 +192,16 @@ private ResolverSet getResolverSet(ParameterizedModel model, List<ParameterModel
...
-          Object defaultValue = p.getDefaultValue();
-          if (defaultValue instanceof String) {
-            return getExpressionBasedValueResolver((String) defaultValue, p, muleContext);
-          } else if (defaultValue != null) {
-            return new StaticValueResolver<>(defaultValue);
-          }
-          return null;
-        });
+        resolver =
+            getDefaultValueResolver(p.getModelProperty(DefaultEncodingModelProperty.class).isPresent(), muleContext, () -> {
```
marianogonzalez: same todo here as well

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ResolverUtils.java
```diff
@@ -65,6 +67,18 @@ private ResolverUtils() {
...
   }
 
+  /**
+   * Gets a {@link ValueResolver} for the parameter if it has an associated a default value or encoding.
+   *
+   * @param hasDefaultEncoding whether the parameter has to use runtime's default encoding or not
+   * @return {@link Supplier} for obtaining the the proper {@link ValueResolver} for the default value, {@code null} if there is
+   *         no default.
+   */
+  static ValueResolver<?> getDefaultValueResolver(boolean hasDefaultEncoding, MuleContext muleContext,
```
marianogonzalez: same todo... this method should not exists.

PR #5259, opened 2017-07-14
MULE-12858: Implement Parameter Options Resolver in Tooling Client's …

estebanwasinger:
…Side

### DIFF
--- a/tests/unit/src/main/java/org/mule/tck/junit4/matcher/ValueMatcher.java
+++ b/tests/unit/src/main/java/org/mule/tck/junit4/matcher/ValueMatcher.java
@@ -9,6 +9,7 @@ package org.mule.tck.junit4.matcher;
 import static java.lang.String.format;
 import static java.util.Arrays.stream;
 import static java.util.stream.Collectors.joining;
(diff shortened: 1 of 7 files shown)


tests/unit/src/main/java/org/mule/tck/junit4/matcher/ValueMatcher.java
```diff
@@ -96,16 +123,18 @@ public ValueMatcher withPartName(String partName) {
...
         .map(Matcher::toString)
         .collect(joining(", "))));
     return this;
   }
 
-  private void validateEquals(String actual, String expected) {
-    if (actual != null && !expected.equals(actual)) {
-      throw new RuntimeException(String.format("Assertion Error - Actual: '%s' Expected: '%s'", actual, expected));
+  private <T> void assertThat(T expected, Matcher<T> actual) {
+    if (actual != null) {
```
estebanwasinger: Wrong name, `expected` should be `actual` and vice versa.
And the lack of the `Matcher` means that that value should not be compared

PR #5263, opened 2017-07-15
MULE-13068: Dispose through tooling client does not call onUndeployme…

gsfernandes:
…ntSuccess causing a Memory leak on the Metaspace.


gsfernandes:
Also PR on beta branch https://github.com/mulesoft/mule/pull/5264

PR #5281, opened 2017-07-17
MULE-11808: Implement SDK based VM connector

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceAdapter.java
```diff
@@ -229,6 +231,21 @@ public void start() throws MuleException {
     }
   }
 
+  private void injectFlowInfo() {
+    List<Field> fields = getFieldsOfType(source.getClass(), FlowInfo.class);
+    if (fields.isEmpty()) {
+      return;
+    } else if (fields.size() > 1) {
+      throw new MuleRuntimeException(createStaticMessage(format(
```
alegmarra: java model validator for this?

PR #5286, opened 2017-07-18
MULE-13104 Rework error handler invocation to be event context based …

dfeist:
…rather than rely on handler propagation.


core/src/main/java/org/mule/runtime/core/AbstractEventContext.java
```diff
@@ -118,7 +123,6 @@ public final void success(Event event) {
     synchronized (this) {
       if (responseProcessor.isTerminated()) {
         LOGGER.info(this + " response was already completed, ignoring.");
-        return;
```
elrodro83: is this removal ok? the method above has it...

core/src/main/java/org/mule/runtime/core/DefaultEventContext.java
```diff
@@ -126,7 +126,33 @@ public static EventContext create(String id, String serverId, ComponentLocation
...
+    return child(parent, componentLocation, NULL_ERROR_HANDLER);
+  }
+
+  /**
+   * Builds a new child execution context from a parent context. A child context delegates all getters to the parent context but
+   * has it's own completion lifecycle. Completion of the child context will not cause the parent context to complete. This is
+   * typically used in {@code flow-ref} type scenarios where a the referenced Flow should complete the child context, but should
+   * not complete the parent context
+   * <p/>
+   * This implementation performs its own error-handling using the closet available error handler in parent contexts and should be
```
elrodro83: closet -> closest. Wouldn't it be better as `most recent`?

PR #5302, opened 2017-07-20
MULE-12937: Artifacts using heavy package must should not depend on m…

pablokraan:
…aven for classloader model resolution


pom.xml
```diff
@@ -281,7 +281,7 @@
         <mule.extensions.maven.plugin.version>1.0.0-SNAPSHOT</mule.extensions.maven.plugin.version>
         <mule.app.plugins.maven.plugin.version>1.0.0-SNAPSHOT</mule.app.plugins.maven.plugin.version>
         <mule.jmh.elasticsearch.maven.plugin.version>1.0.0-SNAPSHOT</mule.jmh.elasticsearch.maven.plugin.version>
-
+        <mule-packager-version>3.0.0-SNAPSHOT</mule-packager-version>
```
pablolagreca: We are using mule.packager.version or mulePackagerVersion but never mule-packager-version

PR #5303, opened 2017-07-20
MULE-13140: As XPathExpression is not thread safe, it should be acces…

facundovs:
…sed with synchronization.

According to the [Oracle Documentation](https://docs.oracle.com/javase/7/docs/api/javax/xml/xpath/XPathExpression.htm):

"An XPath expression is not thread-safe and not reentrant. In other words, it is the application's responsibility to make sure that one XPathExpression object is not used from more than one thread at any given time, and while the evaluate method is invoked, applications may not recursively call the evaluate method."

Regarding to its implementation of XPathExpression, [Saxon documentation](http://www.saxonica.com/html/documentation/javadoc/net/sf/saxon/xpath/XPathExpressionImpl.html) also states:

"The class (XPathExpressionImpl) also includes some methods retained from Saxon's original XPath API. When these methods are used, the object contains the context node and other state, so it is not thread-safe."

Therefore, we must synchronize the use of XPathExpression to avoid concurrency issues.


modules/xml/src/main/java/org/mule/module/xml/xpath/JaxpXPathEvaluator.java
```diff
@@ -64,20 +60,15 @@
...
 {
 
+    private static final int MIN_IDLE_XPATH_EXPRESSIONS = 1;
+    private static final int MAX_IDLE_XPATH_EXPRESSIONS = 32;
+    private static final int MAX_ACTIVE_XPATH_EXPRESSIONS = MAX_IDLE_XPATH_EXPRESSIONS;
+
     private final XPathFactory xpathFactory;
     private final Map<String, String> prefixToNamespaceMap = new HashMap<>();
     private final ThreadLocal<MuleEvent> evaluationEvent = new ThreadLocal<>();
+    private Map<String, GenericObjectPool<XPathExpression>> xpathExpressionPools = new HashMap<>();
```
elrodro83: keep using the loading cache here, since it better handles concurrency, so you can avoid having to synchronixe on this map

modules/xml/src/main/java/org/mule/module/xml/xpath/XPathExpressionFactory.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.apache.commons.pool.BasePoolableObjectFactory;
+
+import javax.xml.namespace.NamespaceContext;
+import javax.xml.xpath.XPath;
+import javax.xml.xpath.XPathExpression;
+import javax.xml.xpath.XPathFactory;
+import javax.xml.xpath.XPathVariableResolver;
+
+public class XPathExpressionFactory extends BasePoolableObjectFactory <XPathExpression>
```
elrodro83: make package access

PR #5304, opened 2017-07-20

facundovs:

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/processor/AbstractFilteringMessageProcessor.java
+++ b/core/src/main/java/org/mule/runtime/core/processor/AbstractFilteringMessageProcessor.java
@@ -6,6 +6,10 @@
  */
 package org.mule.runtime.core.processor;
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mule/runtime/core/processor/AbstractFilteringMessageProcessor.java
```diff
@@ -27,6 +31,9 @@
  */
 public abstract class AbstractFilteringMessageProcessor extends AbstractInterceptingMessageProcessor {
 
+  public static final String FILTER_ON_UNACCEPTED_NOT_STOP_PARENT_FLOW =
```
dfeist: Why add this flag in 4.x?

facundovs: This flag was added to keep backwards compatibility.
Do you prefer remove it and to take this behaviour as the default and unique in mule-4.x?

PR #5313, opened 2017-07-21
MULE-10633: Support SDK extensions on domains

estebanwasinger:


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/artifact/ConnectionManagerConfigurationBuilder.java
```diff
@@ -0,0 +1,48 @@
...
+import org.mule.runtime.core.api.config.ConfigurationBuilder;
+import org.mule.runtime.core.api.config.ConfigurationException;
+import org.mule.runtime.core.api.registry.RegistrationException;
+import org.mule.runtime.core.internal.connection.CompositeConnectionManager;
+import org.mule.runtime.core.internal.connection.DefaultConnectionManager;
+import org.mule.runtime.deployment.model.api.DeployableArtifact;
+
+/**
+ *
+ *
```
afelisatti: Javadoc... Kidding

core/src/main/java/org/mule/runtime/core/api/context/DefaultMuleContextBuilder.java
```diff
@@ -106,9 +108,13 @@ public MuleContext buildMuleContext() {
         .setBootstrapServiceDiscoverer(injectMuleContextIfRequired(getBootstrapPropertiesServiceDiscoverer(), muleContext));
 
     getObjectSerializer(muleContext);
-    ErrorTypeRepository defaultErrorTypeRepository = createDefaultErrorTypeRepository();
-    muleContext.setErrorTypeRepository(defaultErrorTypeRepository);
-    muleContext.setErrorTypeLocator(createDefaultErrorTypeLocator(defaultErrorTypeRepository));
+
+    if (errorTypeRepository == null) {
+      errorTypeRepository = createDefaultErrorTypeRepository();
+    }
```
afelisatti: I think you are missing the else here where you create it passing along the existing one.

core/src/main/java/org/mule/runtime/core/api/exception/CompositeErrorTypeRepository.java
```diff
@@ -0,0 +1,111 @@
...
+    return childErrorTypeRepository.addInternalErrorType(errorTypeIdentifier, parentErrorType);
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Optional<ErrorType> lookupErrorType(ComponentIdentifier errorTypeComponentIdentifier) {
+    Optional<ErrorType> errorType = childErrorTypeRepository.lookupErrorType(errorTypeComponentIdentifier);
+    if (!errorType.isPresent()) {
```
marianogonzalez: orElseGet()

core/src/main/java/org/mule/runtime/core/api/exception/CompositeErrorTypeRepository.java
```diff
@@ -0,0 +1,111 @@
...
+    return errorType;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Optional<ErrorType> getErrorType(ComponentIdentifier errorTypeIdentifier) {
+    Optional<ErrorType> errorType = childErrorTypeRepository.getErrorType(errorTypeIdentifier);
+    if (!errorType.isPresent()) {
```
marianogonzalez: orElseGet()

tests/allure/src/main/java/org/mule/test/allure/AllureConstants.java
```diff
@@ -256,6 +256,15 @@
 
   }
 
+  interface VMFeature {
```
marianogonzalez: why here instead of in the VM extension?

core/src/main/java/org/mule/runtime/core/api/exception/CompositeErrorTypeRepository.java
estebanwasinger: "The problem" is that doing that `orElseGet()` I need to unwrap the Optional to wrap it again to be able to return it and it get's a little weird.
``` ofNullable(childErrorTypeRepository.lookupErrorType(errorTypeComponentIdentifier).orElseGet(() -> parentErrorTypeRepository.lookupErrorType(errorTypeComponentIdentifier).orElse(null)));```

tests/allure/src/main/java/org/mule/test/allure/AllureConstants.java
estebanwasinger: - VM is not a proper maven test dependency
- The convention seems to be to have all these connector constants here.

PR #5321, opened 2017-07-24
MULE-13147: Avoid having spring-module as privileged plugin

elrodro83:
* Make SecurityFilterMessageProcessor not an intercepting MP
* Remove mutators from `AuthenticationFilter` and `SecurityFilter`


modules/spring-config/src/main/resources/META-INF/mule-module.properties
```diff
@@ -1,6 +1,7 @@
 module.name=spring-config
 
-artifact.export.classPackages=org.mule.runtime.config.spring.artifact,\
+artifact.export.classPackages=org.mule.runtime.config.spring.api.dsl.processor,\
```
pablokraan: The fix should be to avoid exposing the classes... exposing them for everyone instead of just to privileged plugins is not better than the current approach, don't you think?

PR #5324, opened 2017-07-24

facundovs:


core/src/main/java/org/mule/runtime/core/privileged/processor/AbstractFilteringMessageProcessor.java
```diff
@@ -75,7 +77,9 @@ public Event process(Event event) throws MuleException {
           if (accept(event, builder)) {
             return just(event).transform(applyNext());
           } else {
-            return just(event).transform(unacceptedMessageProcessor);
+            just(event).transform(unacceptedMessageProcessor).subscribe(requestUnbounded());
```
dfeist: Se que no esta hecho asi ya en varios otros lados, pero seria un poco mejor hacer:
`just(event).transform(unacceptedMessageProcessor).doFinally(() -> event.getContext().success()).materialize().then(empty());`

PR #5329, opened 2017-07-24
MULE-13169: Only create sessions when JMS connection is not being clo…

facundovs:
…sed to aviod deadlock.

### DIFF
--- a/transports/jms/src/main/java/org/mule/transport/jms/JmsConnector.java
+++ b/transports/jms/src/main/java/org/mule/transport/jms/JmsConnector.java
@@ -679,7 +679,7 @@ public class JmsConnector extends AbstractConnector implements ExceptionListener
     {
         if(stopping)
         {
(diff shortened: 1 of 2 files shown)


transports/jms/src/main/java/org/mule/transport/jms/JmsConnector.java
```diff
@@ -675,9 +677,12 @@ public Session getSession(boolean transacted, boolean topic) throws JMSException
 
     private Session createSession(boolean transacted, boolean topic) throws JMSException
     {
-        Session session;
-        session = jmsSupport.createSession(connection, topic, transacted, acknowledgementMode, noLocal);
-        return session;
+        if(stopping)
+        {
+            throw new RuntimeException("It is not possible to create a session since connection is being stopped.");
```
dfeist: Sorry didn't see this before, but can we use a more specific exception here. Either MuleRuntimeException or a more specific java exception like IllegalStateException or something..

PR #5330, opened 2017-07-24
Mule 13170: runExpectingException() should handle cases where exception is thrown from a source different than the expected one

lucianoRM:


tests/functional/src/main/java/org/mule/functional/junit4/FlowRunner.java
```diff
@@ -211,6 +196,38 @@ public void dispatchAsync() throws Exception {
   }
 
   /**
+   * Exdecutes the flow without verification
+   * @return
+   * @throws Exception
+   */
+  private Event runWithoutVerify() throws Exception {
```
elrodro83: isn't there a method that does this already?

PR #5338, opened 2017-07-25
MULE-11675 - Support DI for mule objects into spring beans

pablolagreca:


modules/spring-config/src/main/java/org/springframework/beans/factory/support/DefaultListableBeanFactory.java
```diff
@@ -1327,7 +1325,7 @@ public boolean isRequired() {
         }
       };
       descriptorToUse.increaseNestingLevel();
-      return Optional.fromNullable(doResolveDependency(descriptorToUse, beanName, null, null));
+      return Optional.ofNullable(doResolveDependency(descriptorToUse, beanName, null, null));
```
elrodro83: static import

PR #5344, opened 2017-07-25
MULE-12872 Add payload length to TypedValue API

dfeist:


core/src/main/java/org/mule/runtime/core/internal/message/DefaultMessageBuilder.java
```diff
@@ -56,16 +61,13 @@
 import org.slf4j.LoggerFactory;
 
 public class DefaultMessageBuilder
-    implements InternalMessage.Builder, InternalMessage.PayloadBuilder, InternalMessage.AttributesBuilder,
-    InternalMessage.CollectionBuilder {
+    implements InternalMessage.Builder, InternalMessage.PayloadBuilder, InternalMessage.CollectionBuilder {
```
afelisatti: This change is kinda odd. Can you check it wasn't a bad rebase?

dfeist: Updated mule-api to match.

core/src/main/java/org/mule/runtime/core/internal/message/DefaultMessageBuilder.java
```diff
@@ -157,170 +155,170 @@ public DefaultMessageBuilder(org.mule.runtime.api.message.Message message) {
...
       throw new IllegalStateException("Item MediaType cannot be set, because payload is not a collection");
     }
     return this;
   }
 
   @Override
-  public InternalMessage.Builder attributes(TypedValue<?> typedValue) {
-    this.attributes = typedValue.getValue();
-    this.attributesDataType = typedValue.getDataType();
+  public InternalMessage.CollectionBuilder attributes(TypedValue<?> attributes) {
```
afelisatti: @elrodro83 can you give your take on this? At this point I'm inclined to say we should just have Builder if it gets too complicated. I'd expect CollectionBuilder to only be returned in the `collectionValue` methods, to force `itemMediaType` or something. Everywhere else it should be Builder.

dfeist: This `CollectionBuilder` is only returned when `collection` methods are used in the interface. Here, in the builder implementation all methods return the most generic builder simply the implementation given the user is using it via interface.

It's not more complex than before, but rather following the same approach with bug fixed. In fact i remove the AttributesBuilder which wasn't being used.

elrodro83: Would it be possible for DW to build a collection message with this?

dfeist: @afelisatti had said there wasn't an issue with this as long as it wasn't so restrictive that it becomes unusable without IDE and she decided to leave these builders in place.  Here i'm fixing bug, rather than changing anything in the API.

core/src/main/java/org/mule/runtime/core/api/util/IOUtils.java
```diff
@@ -328,7 +330,8 @@ public static Message toMuleMessagePart(String name, Object object, MediaType co
     final Message.Builder builder;
 
     if (object instanceof File) {
-      builder = Message.builder().value(new FileInputStream((File) object));
+      builder = Message.builder()
+          .payload(new TypedValue(new FileInputStream((File) object), fromObject(object), of(((File) object).length())));
```
elrodro83: extract `(File) object` to a variable?

core/src/main/java/org/mule/runtime/core/el/DefaultExpressionManager.java
```diff
@@ -165,7 +165,7 @@ private TypedValue handleStreaming(TypedValue value, Event event) {
 
     Object payload = value.getValue();
     if (payload instanceof CursorProvider) {
-      value = new TypedValue<>(streamingManager.manage((CursorProvider) payload, event), value.getDataType());
+      value = new TypedValue<>(streamingManager.manage((CursorProvider) payload, event), value.getDataType(), value.getLength());
```
elrodro83: is this code only here?

core/src/main/java/org/mule/runtime/core/internal/message/DefaultMessageBuilder.java
afelisatti: DW does reflection, so it's not limited by this since it will be dealing with the implementation that's all builders at once.

core/src/main/java/org/mule/runtime/core/api/util/StreamingUtils.java
```diff
@@ -136,4 +138,46 @@ private static void handlePossibleException(Reference<Throwable> exception) thro
...
+  /**
+   * Updates the value a given {@link TypedValue} instance by replacing it with a {@link CursorProvider}.
+   *
+   * @param value the typed value to update
+   * @param event the current event
+   * @param streamingManager the streaming manager
+   * @return updated {@link TypedValue instance}
+   */
+  public static TypedValue updateTypedValueForStreaming(TypedValue value, Event event, StreamingManager streamingManager) {
+    if (event == null) {
```
elrodro83: is this case valid? shouldn't it fail?

core/src/main/java/org/mule/runtime/core/api/util/StreamingUtils.java
```diff
@@ -136,4 +138,46 @@ private static void handlePossibleException(Reference<Throwable> exception) thro
...
+      return value;
+    } else {
+      Object payload = value.getValue();
+      if (payload instanceof CursorProvider) {
+        value =
```
elrodro83: early return to avoid reusing the variable `value`

core/src/main/java/org/mule/runtime/core/api/util/StreamingUtils.java
dfeist: This is existing code refactored, prefer not to change behaviour in something @marianogonzalez  implemented as part of a simple refactor to improve code resuse.

PR #5347, opened 2017-07-25
MULE-12202: Don't map variables as single bindings

afelisatti:


tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/model/HealthStatus.java
```diff
@@ -7,7 +7,7 @@
 package org.mule.test.heisenberg.extension.model;
 
 public enum HealthStatus {
-  HEALTHY("Healthy"), CANCER("Cancer"), DEAD("Dead");
+  HEALTHY("HEALTHY"), CANCER("CANCER"), DEAD("DEAD");
```
pablolagreca: TODO

PR #5368, opened 2017-07-27
MULE-12831: Complete ObjectStore connector

marianogonzalez:


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/parameter/TopLevelParameterObjectFactory.java
```diff
@@ -45,6 +59,34 @@ public TopLevelParameterObjectFactory(ObjectType type, ClassLoader classLoader,
   }
 
   @Override
+  public void initialise() throws InitialisationException {
```
alegmarra: isn't initialise called before doGetObject?

marianogonzalez: surprisingly not

PR #5370, opened 2017-07-27
MULE-13209 Cleanup scatter-gather configuration

dfeist:
REVIEW ONLY (tests running)


modules/spring-config/src/main/resources/META-INF/mule.xsd
```diff
@@ -3179,22 +3192,8 @@
...
                     <xsd:sequence>
-                        <xsd:element name="custom-aggregation-strategy" type="aggregationStrategyType" minOccurs="0" maxOccurs="1">
-                            <xsd:annotation>
-                                <xsd:documentation>
-                                    Allows customizing the logic used for aggregation the events back together.
-                                </xsd:documentation>
-                            </xsd:annotation>
-                        </xsd:element>
-                        <xsd:group ref="messageProcessorOrMixedContentMessageProcessor" minOccurs="2" maxOccurs="unbounded"/>
+                        <xsd:group ref="routeOrProcessorChain" minOccurs="2" maxOccurs="unbounded"/>
```
marianogonzalez: I think routes should have a mandatory id. @pablolagreca what's your opinion on this?

dfeist: I don't agree that this needs to be mandatory, an optional id in the future would be good though. Anyhow, we cannot implement this in Mule 4 due to issues with tooling.

pablolagreca: The ID thing will come from the DF spec. As DF says, this PR is for cleanup. Let's discuss id attribute in the spec.

PR #5372, opened 2017-07-28
MULE-11501: catalog manager support for smart connectors

fernandezlautaro:
depends on https://github.com/mulesoft/metadata-model-api/pull/88


modules/extensions-xml-support/src/test/resources/compatibility/modules/module-xsd-custom-types.xml
```diff
@@ -0,0 +1,32 @@
...
+    <operation name="operation-with-custom-types">
+        <parameters>
+            <parameter name="value" type="XsdType1"/>
+            <!--extra parameter without targetNamespace in the schema under XsdType3-->
+            <parameter name="value2" type="XsdType3"/>
+        </parameters>
+        <body>
+            <mule:set-payload value="#[parameters.value]"/>
+        </body>
+        <output type="XsdType2"/>
```
pablolagreca: Do we have an issue to work on allowing here to define the type of the attributes as well?

modules/extensions-xml-support/src/test/resources/modules/module-json-custom-types-catalog.xml
```diff
@@ -1,11 +1,11 @@
...
-            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-            xmlns:types="http://www.mulesoft.org/schema/mule/types">
+<catalogs xmlns="http://www.mulesoft.org/schema/mule/types" >
 
-    <types:catalog>
-        <types:type name="Type1" location="./module-json-custom-types-type1-schema.json"/>
-        <types:type name="Type2" location="./module-json-custom-types-type2-schema.json"/>
-    </types:catalog>
+    <catalog name="Type1" format="application/json">
+        <schema format="application/json+schema" location="./module-json-custom-types-type1-schema.json" />
```
pablolagreca: what's the purpose of ./ ?

pablolagreca: We have a feature in the runtime that allows to put a file inplace in the config and it should be use the general purpose tool for picking up values from a file. So I would be able to do: <schema format="application/json+schema"> MY_JSON_SCHEMA_HERE </schema> or do:
<schema format="application/json+schema">${file::MY_JSON_SCHEMA_LOCATION_HERE}</schema>

modules/extensions-xml-support/src/test/resources/compatibility/modules/module-xsd-custom-types.xml
fernandezlautaro: we do: https://www.mulesoft.org/jira/browse/MULE-12928
It will be an extra element after `<output>`, optional, called `<output-attribute type="bleh"/>` (by default, will be void)

modules/extensions-xml-support/src/test/resources/modules/module-json-custom-types-catalog.xml
fernandezlautaro: no purpose, just a way to make clear it's a local reference. Without the `./` works fine, as it relies in the plugin classloader to look for resources.

fernandezlautaro: >So I would be able to do: MY_JSON_SCHEMA_HERE or do:
${file::MY_JSON_SCHEMA_LOCATION_HERE}

we could test it, but probably won't work for the types.. it should work for the rest of the `<module>` though.. as smart connectors relies in the runtime to load the XMLs ( org.mule.runtime.config.spring.XmlConfigurationDocumentLoader )

PR #5373, opened 2017-07-28
MULE-13218: JUnit rule for executing groovy scripts on system tests

GustavoARod:
- Adding new GroovyScriptExecutor class and groovy dependency on pom.xml

### DIFF
--- a/tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/GroovyScriptExecutor.java
+++ b/tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/GroovyScriptExecutor.java
@@ -34,7 +34,7 @@ import groovy.lang.GroovyShell;
  *
  *   &#064;Test
  *   public void useApplication() throws IOException {
- *     // Groovy scrip is executed
+ *     // Groovy script is executed
  *     // This code exercises the application
  *   }
  * }
--- a/tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/GroovyScriptExecutor.java
+++ b/tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/GroovyScriptExecutor.java
@@ -1,0 +1,95 @@
+/*
+ * (c) 2003-2017 MuleSoft, Inc. This software is protected under international copyright
+ * law. All use of this software is subject to MuleSoft's Master Subscription Agreement
+ * (or other master license agreement) separately entered into in writing between you and
+ * MuleSoft. If such an agreement is not in place, you may not use the software.
+ */
+package org.mule.test.infrastructure.process.rules;
+
+import org.mule.runtime.core.api.util.IOUtils;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Supplier;
+
+import org.junit.rules.ExternalResource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import groovy.lang.GroovyShell;
+
+/**
+ * Junit rule to run Groovy scripts.
+ */
+public class GroovyScriptExecutor extends ExternalResource {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(MuleDeployment.class);
+
+  private final String scriptPath;
+  private Map<String, String> properties = new HashMap<>();
+  private Map<String, Supplier<String>> propertiesUsingLambdas = new HashMap<>();
+
+  public static class Builder {
+
+    GroovyScriptExecutor executor;
+
+    Builder(String scriptPath) {
+      executor = new GroovyScriptExecutor(scriptPath);
+    }
+
+    public Builder withProperty(String property, String value) {
+      executor.properties.put(property, value);
+      return this;
+    }
+
+    public Builder withPropertyUsingLambda(String property, Supplier<String> propertySupplier) {
+      executor.propertiesUsingLambdas.put(property, propertySupplier);
+      return this;
+    }
+
+    public Builder withProperties(Map<String, String> properties) {
+      executor.properties.putAll(properties);
+      return this;
+    }
+
+    public GroovyScriptExecutor build() {
+      return executor;
+    }
+  }
+
+  public static GroovyScriptExecutor.Builder builder(String scriptPath) {
+    return new GroovyScriptExecutor.Builder(scriptPath);
+  }
+
+  protected GroovyScriptExecutor(String scriptPath) {
+    this.scriptPath = scriptPath;
+  }
+
+  @Override
+  protected void before() throws Throwable {
+    executeGroovyScript();
+  }
+
+  private void executeGroovyScript() {
+    try {
+      GroovyShell shell = new GroovyShell();
+
+      resolvePropertiesUsingLambdas();
+      properties.forEach((key, value) -> shell.setProperty(key, value));
+
+      shell.evaluate(IOUtils.getResourceAsString(scriptPath, GroovyScriptExecutor.class));
+
+      LOGGER.info("Groovy script executed");
+
+    } catch (Exception e) {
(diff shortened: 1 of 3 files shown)


tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/GroovyScriptExecutor.java
```diff
@@ -0,0 +1,95 @@
+/*
+ * (c) 2003-2017 MuleSoft, Inc. This software is protected under international copyright
```
aiannucci: Wrong license header, I'm pretty sure this isn't compiling with Maven.

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/GroovyScriptExecutor.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.function.Supplier;
+
+import org.junit.rules.ExternalResource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import groovy.lang.GroovyShell;
+
+/**
+ * Junit rule to run Groovy scripts.
```
aiannucci: Add a proper javadoc

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/GroovyScriptExecutor.java
```diff
@@ -0,0 +1,95 @@
...
+ */
+public class GroovyScriptExecutor extends ExternalResource {
```
aiannucci: This shouldn't extend ExternalResource

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/GroovyScriptExecutor.java
```diff
@@ -0,0 +1,95 @@
...
+    public Builder withProperties(Map<String, String> properties) {
+      executor.properties.putAll(properties);
+      return this;
+    }
+
+    public GroovyScriptExecutor build() {
+      return executor;
+    }
+  }
+
```
aiannucci: Add javadocs for all public elements of this class

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/GroovyScriptExecutor.java
```diff
@@ -0,0 +1,95 @@
...
+import org.slf4j.LoggerFactory;
+
+import groovy.lang.GroovyShell;
+
+/**
+ * Junit rule to run Groovy scripts.
+ */
+public class GroovyScriptExecutor extends ExternalResource {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(MuleDeployment.class);
```
aiannucci: Wrong class for logger

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/GroovyScriptExecutor.java
```diff
@@ -0,0 +1,95 @@
...
+
+  private final String scriptPath;
+  private Map<String, String> properties = new HashMap<>();
```
aiannucci: I'm not sure about the "properties" concept here. If the purpose is to run a script, you may want to pass *parameters* to it, instead of properties.

aiannucci: Meaning that, parameters should be an array instead of a map

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/GroovyScriptExecutor.java
```diff
@@ -0,0 +1,95 @@
...
+      resolvePropertiesUsingLambdas();
+      properties.forEach((key, value) -> shell.setProperty(key, value));
+
+      shell.evaluate(IOUtils.getResourceAsString(scriptPath, GroovyScriptExecutor.class));
+
+      LOGGER.info("Groovy script executed");
+
+    } catch (Exception e) {
+      String errorMessage = "Error executing Groovy script: " + scriptPath;
+      LOGGER.error(errorMessage, e);
```
aiannucci: It seems unnecessary to log an error message if you are then throwing a runtime exception that will also be logged.

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/GroovyScriptExecutor.java
```diff
@@ -0,0 +1,95 @@
...
+      GroovyShell shell = new GroovyShell();
+
+      resolvePropertiesUsingLambdas();
+      properties.forEach((key, value) -> shell.setProperty(key, value));
+
+      shell.evaluate(IOUtils.getResourceAsString(scriptPath, GroovyScriptExecutor.class));
+
+      LOGGER.info("Groovy script executed");
+
+    } catch (Exception e) {
```
aiannucci: Try not to catch *Exception* right away, instead, try to catch more specific exceptions and throw different errors if the script file was not found, if the script failed, etc.

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/GroovyScriptExecutor.java
GustavoARod: Good point. Nevertheless, "properties" is the concept used by the `GroovyShell` class. Take a look at the method used for this matter. [`public void setProperty(String property, Object newValue) `](https://github.com/groovy/groovy-core/blob/master/src/main/groovy/lang/GroovyShell.java#L137). So, taking that into account, I mean, the fact that properties for the Groovy shell are name and value pair, I think we'd change this part. What do you think?

aiannucci: Fair enough. Congrats, you just earned the Groovy expert title

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/GroovyScriptExecutor.java
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * <pre>
+ * public class MuleApplicationTestCase {
+ *
+ *   &#064;ClassRule
+ *   public static GroovyScriptExecutor executor = builder(&quot;/path/to/script.groovy&quot;).withProperty("property", "value");
+ *
+ *   &#064;Test
+ *   public void useApplication() throws IOException {
+ *     // Groovy scrip is executed
```
aiannucci: scrip -> script

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleDeployment.java
```diff
@@ -47,8 +47,7 @@
  * public class MuleApplicationTestCase {
  * 
  *   &#064;ClassRule
- *   public static MuleDeployment deployment =
- *       builder().withApplications(&quot;/path/to/application.zip&quot;).withProperty("-M-Dproperty", "value").timeout(120).deploy();
+ *   public static MuleDeployment deployment = builder(&quot;/path/to/script.groovy&quot;).withProperty("property", "value");
```
aiannucci: is this right?

GustavoARod: :O This was committed by mistake. Sorry

aiannucci:
jenkins validate

aiannucci:
jenkins validate

PR #5377, opened 2017-07-28
MULE-11955: Be able to retrieve protocol, host, port and path from HT…

marianogonzalez:
…TP Listener


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -378,4 +386,20 @@ protected void doInitialise() throws InitialisationException {
...
+  public Map<String, Object> getInitialisationParameters() {
+    final ValueResolvingContext ctx = from(MuleExtensionUtils.getInitialiserEvent());
+    Map<String, ValueResolver<?>> resolvers = sourceAdapterFactory.getSourceParameters().getResolvers();
+    return resolvers.entrySet().stream()
+        .collect(toImmutableMap(entry -> entry.getKey(), entry -> {
+          try {
+            return entry.getValue().resolve(ctx);
+          } catch (MuleException e) {
+            throw new MuleRuntimeException(createStaticMessage(format(
+                                                                      "Could not resolve parameter '%s'", entry.getKey())),
```
alegmarra: add info about the source that failed, model name, namespace, etc. maybe even location?

PR #5385, opened 2017-07-31
MULE-1683: When the quartz connector is stopped, standby() should be …

facundovs:
…invoked instead of shutdown() method.

### DIFF
--- a/transports/quartz/src/main/java/org/mule/transport/quartz/QuartzConnector.java
+++ b/transports/quartz/src/main/java/org/mule/transport/quartz/QuartzConnector.java
@@ -244,7 +244,8 @@ public class QuartzConnector extends AbstractConnector
     }
 
     /**
-     * creates a SchedulerFactory used to create quartz scheduler instances.
+     * Creates a {@link SchedulerFactory} used to create Quartz scheduler instances
+     *
      * @return a SchedulerFactory.
      * @throws SchedulerException
      */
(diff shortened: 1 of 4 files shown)


transports/quartz/src/main/java/org/mule/transport/quartz/QuartzConnector.java
```diff
@@ -243,6 +243,11 @@ TimeZone resolveTimeZone(String timeZone, String name)
         return resolvedTimeZone;
     }
 
+    protected SchedulerFactory createSchedulerFactory() throws SchedulerException
```
pablokraan: Add javadoc

transports/quartz/src/main/java/org/mule/transport/quartz/QuartzConnector.java
```diff
@@ -243,6 +243,11 @@ TimeZone resolveTimeZone(String timeZone, String name)
         return resolvedTimeZone;
     }
 
+    /**
+     * creates a SchedulerFactory used to create quartz scheduler instances.
```
pablokraan: Creates a {@link SchedulerFactory} used to create Quartz scheduler instances

pablokraan: Add an extra line between method and parameter descriptions

PR #5387, opened 2017-07-31
MULE-13032: Review all interfaces builders to match DW standard

afelisatti:


core/src/main/java/org/mule/runtime/core/internal/message/DefaultMessageBuilder.java
```diff
@@ -115,7 +115,12 @@ public DefaultMessageBuilder(org.mule.runtime.api.message.Message message) {
 
   @Override
   public InternalMessage.Builder payload(Object payload) {
-    this.payload = payload;
+    if (payload instanceof TypedValue<?>) {
```
dfeist: This is bad idea IMO, what if I want to create a payload that as a TypedValue value? We either need to make payload() always receive a `TypedValue` (preferred) or differentiate this with a different method. Why not rename payload() to value(Object) and use payload(TypedValue), that is probably better than payload(Object) and typedPayload(TypedValue)

core/src/main/java/org/mule/runtime/core/internal/message/DefaultMessageBuilder.java
```diff
@@ -115,7 +115,12 @@ public DefaultMessageBuilder(org.mule.runtime.api.message.Message message) {
...
+      this.payload = ((TypedValue) payload).getValue();
+      this.dataType = ((TypedValue) payload).getDataType();
```
dfeist: Internally the builder should really be using typedValue as state, such that if we add a field to TypedValue we don't need to edit this builder in order to ensure it gets propagated.

core/src/main/java/org/mule/runtime/core/internal/message/DefaultMessageBuilder.java
afelisatti: Sounds good. You'll have many more PRs to review :)

core/src/main/java/org/mule/runtime/core/internal/message/DefaultMessageBuilder.java
afelisatti: True, but that's beyond this PR IMO because we'd also have to avoid creating a new TypedValue  every time a new part is changed.

PR #5389, opened 2017-08-01
MULE-13230: MetadataOutputDelegate generating metadata type with wron…

pablomusumeci:
…g type annotation

### DIFF
--- a/modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/metadata/MetadataOperationTestCase.java
+++ b/modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/metadata/MetadataOperationTestCase.java
@@ -14,12 +14,14 @@ import static org.hamcrest.Matchers.hasItem;
 import static org.hamcrest.Matchers.hasItems;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.isEmptyString;
 import static org.hamcrest.Matchers.isOneOf;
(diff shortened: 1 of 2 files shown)


modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/metadata/MetadataOperationTestCase.java
```diff
@@ -569,6 +571,7 @@ private MetadataType getResolvedTypeFromList() {
     ComponentMetadataDescriptor descriptor = result.get();
     MetadataType param = descriptor.getModel().getOutput().getType();
     assertThat(param, is(instanceOf(ArrayType.class)));
+    assertThat(getId(param), is(isEmptyString()));
```
marianogonzalez: why should the id BE an empty string?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataOutputDelegate.java
```diff
@@ -188,7 +188,7 @@ private MetadataType adaptToListIfNecessary(MetadataType resolvedType, Object ke
       resolvedType = wrapInMessageType(resolvedType, key, metadataContext, message.getAttributesType());
     }
 
-    return metadataContext.getTypeBuilder().arrayType().id(typeId).of(resolvedType).build();
+    return metadataContext.getTypeBuilder().arrayType().of(resolvedType).build();
```
marianogonzalez: have you run all test cases to verify that the SDK doesn't try to obtain the list type?

PR #5393, opened 2017-08-02
MULE-13215 Define ForkJoinStrategy API and implement must-have strate…

dfeist:
…gies for RC

### DIFF
--- a/core-tests/src/test/java/org/mule/runtime/core/el/context/MessageContextTestCase.java
+++ b/core-tests/src/test/java/org/mule/runtime/core/el/context/MessageContextTestCase.java
@@ -31,6 +31,9 @@ import org.mule.runtime.core.api.message.GroupCorrelation;
 import org.mule.runtime.core.internal.context.notification.DefaultFlowCallStack;
 import org.mule.runtime.core.internal.message.InternalMessage;
 
+import java.util.Optional;
+import java.util.OptionalInt;
+
 import org.junit.Before;
 import org.junit.Test;
 
@@ -49,7 +52,7 @@ public class MessageContextTestCase extends AbstractELTestCase {
     when(event.getFlowCallStack()).thenReturn(new DefaultFlowCallStack());
     when(event.getError()).thenReturn(empty());
     message = spy(Message.of(null));
-    when(event.getGroupCorrelation()).thenReturn(mock(GroupCorrelation.class));
+    when(event.getGroupCorrelation()).thenReturn(empty());
     when(event.getMessage()).thenAnswer(invocation -> message);
   }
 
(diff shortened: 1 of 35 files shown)


core/src/main/java/org/mule/runtime/core/internal/routing/forkjoin/AbstractForkJoinStrategy.java
```diff
@@ -0,0 +1,141 @@
...
+import java.util.function.Function;
+
+import org.reactivestreams.Publisher;
+
+/**
+ * Abstract {@link org.mule.runtime.core.api.routing.ForkJoinStrategy} that provides the base behavior for strategies that will
+ * perform parallel invocation of {@link org.mule.runtime.core.api.routing.ForkJoinStrategy.RoutingPair}'s that wish to use the
+ * following common behaviour:
+ * <ul>
+ * <li>Emit result even only once all routes are complete.
```
alepulver: Maybe "Emit result event only once, after all routes are complete." or similar?

core-tests/src/test/java/org/mule/runtime/core/internal/routing/forkjoin/CollectMapForkJoinStrategyTestCase.java
```diff
@@ -0,0 +1,61 @@
...
+
+    Event result =
+        from(strategy.forkJoin(original, fromIterable(asList(pair1, pair2, pair3)),
+                               processingStrategy, 1, true)).block();
+
+    assertThat(result.getMessage().getPayload().getValue(), instanceOf(Map.class));
+    Map<String, Message> resultMap = (Map<String, Message>) result.getMessage().getPayload().getValue();
+    assertThat(resultMap.entrySet(), hasSize(3));
+    assertThat(resultMap.keySet(), hasItems("0", "1", "2"));
+    assertThat(resultMap.values(), hasItems(route1Result, route2Result, route3Result));
```
alepulver: Shouldn't this test that resultMap.get("0") == route1Result instead of keys and values  being correct but not necessarily matching?

core-tests/src/test/java/org/mule/runtime/core/el/context/MessageContextTestCase.java
```diff
@@ -75,15 +78,15 @@ public void correlationId() throws Exception {
...
     assertEquals(4, evaluate("message.correlationSequence", event));
     assertFinalProperty("message.correlationSequence=2", event);
   }
 
   @Test
   public void correlationGroupSize() throws Exception {
-    when(event.getGroupCorrelation()).thenReturn(new GroupCorrelation(null, 4));
-    assertEquals(4, evaluate("message.correlationGroupSize", event));
+    when(event.getGroupCorrelation()).thenReturn(Optional.of(GroupCorrelation.of(0, 2)));
+    assertEquals(2, evaluate("message.correlationGroupSize", event));
```
alepulver: Why did the number change?

dfeist: Because previously there was a bug in implementation, and sequenceId/groupSize were mixed up. Best way to avoid this is to use different test values in test case.

PR #5423, opened 2017-08-06
MULE-13256: fix test

alepulver:


modules/spring-config/src/main/java/org/mule/runtime/config/spring/internal/MuleDocumentLoader.java
```diff
@@ -58,7 +60,48 @@ public void handle(short operation, String key, Object data, Node src, Node dst)
...
+          factory.setNamespaceAware(true);
+          try {
+            factory.setAttribute(SCHEMA_LANGUAGE_ATTRIBUTE, XSD_SCHEMA_LANGUAGE);
+          } catch (IllegalArgumentException ex) {
+            ParserConfigurationException pcex = new ParserConfigurationException(
+                                                                                 "Unable to validate using XSD: Your JAXP provider ["
+                                                                                     + factory +
+                                                                                     "] does not support XML Schema. Are you running on Java 1.4 with Apache Crimson? "
+                                                                                     +
+                                                                                     "Upgrade to Apache Xerces (or Java 1.5) for full XSD support.");
```
elrodro83: Mule 4 requieres java 8. This message Is confusing.

modules/spring-config/src/main/java/org/mule/runtime/config/spring/internal/MuleDocumentLoader.java
```diff
@@ -58,7 +60,48 @@ public void handle(short operation, String key, Object data, Node src, Node dst)
...
+   * JAXP attribute used to configure the schema language for validation.
+   */
+  private static final String SCHEMA_LANGUAGE_ATTRIBUTE = "http://java.sun.com/xml/jaxp/properties/schemaLanguage";
+
+  /**
+   * JAXP attribute value indicating the XSD schema language.
+   */
+  private static final String XSD_SCHEMA_LANGUAGE = "http://www.w3.org/2001/XMLSchema";
+
+  private final DocumentLoader defaultLoader = new DefaultDocumentLoader() {
```
elrodro83: Doesn't Spring document loader support this?

elrodro83:
I think the current error message is better that the one expected in the test. It points specifically to the missing namespace instead of pointing to an element

modules/spring-config/src/main/java/org/mule/runtime/config/spring/internal/MuleDocumentLoader.java
alepulver: It doesn't allow passing a factory as argument, that's what I'm changing (instead of `DefaultBuilderFactory.newInstance()` it uses `XMLSecureFactories.createDefault().getDocumentBuilderFactory()`). This is because including xerces in the classpath changes its behavior, as with many other libraries.

modules/spring-config/src/main/java/org/mule/runtime/config/spring/internal/MuleDocumentLoader.java
alepulver: The code is copied from Spring but changing the factory construction, I could but it might be incompatible if something depends on the error message.

alepulver:
I agree. But we are using a different parser now, and we wanted to use the JVM one.

PR #5428, opened 2017-08-07
MULE-11553 - Create an API to create beans that behave like flows

pablolagreca:


core/src/main/java/org/mule/runtime/core/internal/message/DefaultEventBuilder.java
```diff
@@ -481,29 +533,18 @@ private void setMessage(Message message) {
...
 
     @Override
-    public Map<String, TypedValue<Object>> getProperties() {
-      return properties;
-    }
-
-    @Override
-    public Map<String, TypedValue<Object>> getParameters() {
-      return parameters;
+    public Map<String, TypedValue<?>> getParameters() {
```
dfeist: Close https://www.mulesoft.org/jira/browse/MULE-12998 (
Event API is inconsistent and implementation is not immutable after addition of properties and parameters.) when this is merged.

core/src/main/java/org/mule/runtime/core/internal/routing/ExpressionSplittingStrategy.java
```diff
@@ -6,7 +6,7 @@
  */
 package org.mule.runtime.core.internal.routing;
 
-import static org.mule.runtime.core.el.BindingContextUtils.NULL_BINDING_CONTEXT;
+import static org.mule.runtime.internal.el.BindingContextUtils.NULL_BINDING_CONTEXT;
```
afelisatti: Wasn't this moved to api?

core/src/main/java/org/mule/runtime/core/privileged/processor/CompositeProcessorChainRouter.java
```diff
@@ -0,0 +1,86 @@
...
+
+import javax.inject.Inject;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Component to be used that supports a collection of {@link MessageProcessorChain} and executes them in order.
+ * <p/>
+ * Meant to be used be runtime privileged extensions that needs to construct top level chains of execution.
```
afelisatti: `be` -> `by`, `needs` -> `need`

core/src/main/java/org/mule/runtime/core/privileged/processor/CompositeProcessorChainRouter.java
```diff
@@ -0,0 +1,86 @@
...
+  public Event process(Event event) {
+    org.mule.runtime.core.api.Event.Builder builder =
+        org.mule.runtime.core.api.Event.builder(DefaultEventContext.create(getUUID(), muleContext.getId(), getLocation()));
+    org.mule.runtime.core.api.Event defaultEvent = builder.from(event).build();
+    try {
+      for (MessageProcessorChain processorChain : processorChains) {
+        defaultEvent = processorChain.process(defaultEvent);
+      }
+    } catch (MuleException e) {
+      throw new MuleRuntimeException(e);
```
afelisatti: Why throw a runtime exception? Shouldn't there be a specific exception or just propagate it?

core/src/main/java/org/mule/runtime/core/privileged/processor/objectfactory/MessageProcessorChainObjectFactory.java
```diff
@@ -0,0 +1,63 @@
...
+  protected List processors;
+  protected String name;
+
+
+  public void setMessageProcessors(List processors) {
+    this.processors = processors;
+  }
+
+  @Override
+  public Object doGetObject() throws Exception {
```
afelisatti: Isn't this logic repeated in the `initialise()` method of the referenced object?

tests/parsers-plugin/src/main/java/org.mule.tests.parsers.api/LifecycleSensingObjectFactory.java
```diff
@@ -0,0 +1,55 @@
...
+import org.mule.runtime.api.lifecycle.Lifecycle;
+import org.mule.runtime.core.api.processor.Processor;
+import org.mule.runtime.dsl.api.component.AbstractAnnotatedObjectFactory;
+
+import java.util.LinkedList;
+import java.util.List;
+
+public class LifecycleSensingObjectFactory extends AbstractAnnotatedObjectFactory<Processor> implements Lifecycle
+
+{
```
afelisatti: Format

core/src/main/java/org/mule/runtime/core/privileged/processor/objectfactory/MessageProcessorChainObjectFactory.java
pablolagreca: I removed the set annotations part that I saw repetead. Not sure what else do you saw repeated.

core/src/main/java/org/mule/runtime/core/api/Event.java
```diff
@@ -48,73 +47,20 @@
...
-  /**
-   * Returns the variable registered under the given {@code key}
-   *
-   * @param key the name or key of the variable. This must be non-null.
-   * @param <T> the type of the variable value.
-   * @return a {@link TypedValue} containing the variable's value and {@link DataType}
-   * @throws java.util.NoSuchElementException if the flow variable does not exist.
    */
-  <T> TypedValue<T> getVariable(String key);
+  Map<String, ?> getInternalParameters();
```
dfeist: Isn't it best to start using `TypedValue` more consistently in core rather than introducing a second set of API methods to support `Object` instead of `TypedValue`? Or did i misunderstand this and they really are two different distinct sets of parameters?

pablolagreca: In this case we don't need TypeValue (for internalParameters). InternalParameter was created to use it for internal purposes like InterceptionAPI. InterceptionAPI was using the parameters field which was wrong since their purpose is another.

PR #5429, opened 2017-08-07
MULE-13280: In File Endpoints, FileAge is not always honored.

facundovs:
If I have an FileEndpoint that defines a FileAge and refers to a File Connector that doesn't define a fileAge, so the fileAge attribute is ignored.

### DIFF
--- a/transports/file/src/test/java/org/mule/transport/file/FileAgeTestCase.java
+++ b/transports/file/src/test/java/org/mule/transport/file/FileAgeTestCase.java
@@ -9,9 +9,14 @@ package org.mule.transport.file;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
 import static java.lang.Long.parseLong;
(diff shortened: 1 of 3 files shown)


transports/file/src/test/java/org/mule/transport/file/FileAgeTestCase.java
```diff
@@ -9,36 +9,83 @@
...
     public void testFileAgeConnectorValue()
     {
         FileConnector fileConnector = muleContext.getRegistry().lookupObject("fileConnector");
         assertThat(fileConnector.getFileAge(), is(parseLong(CONNECTOR_FILE_AGE)));
     }
 
+    @Test
+    public void fileAgeInheritedFromConnectorIsHonored() throws Exception
+    {
+        File createdFile = createDataFile(temporaryDirectory, TEST_MESSAGE, "UTF-8");
```
pablokraan: Use StandardCharsets#UTF_8.name()

PR #5446, opened 2017-08-08
MULE-11804: Define APIs to dispatch notifications and register listeners

elrodro83:
* Encapsulate notifications related stuff from MuleContext
* Definenew objects accessible from the registry to just dispatch
notifications and manage listeners.
* Fix issue in default value calculation in parsers
* Avoid passing the MuleContext around just to work with
notifications/listeners


core-tests/src/test/java/org/mule/runtime/core/api/context/MuleContextLifecycleTestCase.java
```diff
@@ -49,6 +57,7 @@
   private MuleContextBuilder ctxBuilder;
   private SensingLifecycleManager lifecycleManager;
   private MuleContext ctx;
+  private NotificationListenerRegistry notificationRegistrer;
```
dfeist: rename field?

PR #5449, opened 2017-08-08
MULE-13257 Support maintaining successful route pairs when errors occ…

dfeist:
…ur in ForkJoinStrategy implementations


core-tests/src/test/java/org/mule/runtime/core/internal/routing/forkjoin/AbstractForkJoinStrategyTestCase.java
```diff
@@ -28,307 +29,383 @@
...
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.TimeoutException;
-import java.util.function.Function;
-
-import org.junit.After;
-import org.junit.Before;
-import org.junit.Test;
-
 
 public abstract class AbstractForkJoinStrategyTestCase extends AbstractMuleContextTestCase {
```
afelisatti: Allure annotations are inherited now, so you can add them here.

core-tests/src/test/java/org/mule/runtime/core/internal/routing/forkjoin/AbstractForkJoinStrategyTestCase.java
```diff
@@ -28,307 +29,383 @@
...
+
+    expectedException.expect(instanceOf(MessagingException.class));
+    expectedException.expectCause(instanceOf(CompositeRoutingException.class));
+
+    invokeStrategyBlocking(strategy, testEvent(), asList(createRoutingPairWithSleep(of(1), 20), pair2),
+                           throwable -> {
+                             verify(pair2Processor, times(1)).process(any(Event.class));
+                             CompositeRoutingException compositeRoutingException = assertCompositeRoutingException(throwable, 1);
+                             RoutingResult routingResult = assertRoutingResult(compositeRoutingException, 1, 1);
+                             assertThat(routingResult.getFailures().get(Integer.toString(0)).getCause(),
```
afelisatti: Why are you doing this conversions?

core/src/main/java/org/mule/runtime/core/DefaultMuleContext.java
```diff
@@ -266,9 +266,9 @@
       }
     });
 
-    // Log dropped events/errors rather than blow up which causes cryptic timeouts and stack traces.
-    Hooks.onErrorDropped(error -> logger.error("ERROR DROPPED UNEXPECTEDLY " + error));
-    Hooks.onNextDropped(event -> logger.error("EVENT DROPPED UNEXPECTEDLY " + event));
+    // Log dropped events/errors
+    Hooks.onErrorDropped(error -> logger.debug("ERROR DROPPED " + error));
```
afelisatti: Is this stable enough to be debug?

core/src/main/java/org/mule/runtime/core/api/exception/Errors.java
```diff
@@ -107,6 +108,11 @@
     public static final String NOT_PERMITTED_ERROR_IDENTIFIER = "NOT_PERMITTED";
 
     /**
+     * Indicates that a timeout during processing.
```
afelisatti: `a timeout occurred`

core/src/main/java/org/mule/runtime/core/internal/routing/forkjoin/AbstractForkJoinStrategyFactory.java
```diff
@@ -0,0 +1,166 @@
...
+      final Event.Builder resultBuilder = Event.builder(original);
+      return from(routingPairs)
+          .map(addSequence(count))
+          .flatMap(processRoutePair(processingStrategy, maxConcurrency, delayErrors, timeout, reatorTimeoutScheduler,
+                                    timeoutErrorType),
+                   maxConcurrency)
+          .collectList()
+          .doOnNext(checkedConsumer(list -> {
+            if (list.stream().anyMatch(event -> event.getError().isPresent())) {
+              throw new MessagingException(original, createCompositeRoutingException(list));
```
afelisatti: Is this necessary? ME creation should be limited.

afelisatti: What happens if you throw the composite one directly?

core/src/main/java/org/mule/runtime/core/privileged/processor/chain/AbstractMessageProcessorChain.java
```diff
@@ -175,7 +175,14 @@ public ProcessingType getProcessingType() {
                                                                     processor, muleContext)))
                            .then(Mono.empty()))
         .onErrorResume(MessagingException.class,
-                       throwable -> Mono.from(event.getContext().error(throwable)).then(Mono.empty()))));
+                       throwable -> {
+                         if (!throwable.getEvent().getError().isPresent()) {
+                           // just in case anything in processor chain itself fails. We still have to create Error earlier on
+                           // though, so it's available in interceptors.
+                           throwable = updateMessagingException(processor).apply(throwable);
```
afelisatti: Have you run all tests? Changes like this usually break everything :)

dfeist: Will do, when build is green :-( Anyhow this only covers a very specific scenario where updating the payload for streaming within the chain interceptors themselves result in a NPE, otherwise error should never be null here.

core/src/main/java/org/mule/runtime/core/DefaultMuleContext.java
dfeist: Yes, there is logging en EventContext anyway, for the majority of use cases.  We cannot leave these `error` because when using `timeout` dropping the event is valid, and it's then replaced with another one.

PR #5454, opened 2017-08-09
MULE-11759: DB XA transactional connections are no supported in dynam…

estebanwasinger:
…ic configs

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/transaction/xa/XaTransaction.java
+++ b/core/src/main/java/org/mule/runtime/core/api/transaction/xa/XaTransaction.java
@@ -6,18 +6,14 @@
  */
 package org.mule.runtime.core.api.transaction.xa;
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mule/runtime/core/api/transaction/xa/XaTransaction.java
```diff
@@ -387,7 +387,7 @@ public Object getResource() {
 
     @Override
     public int hashCode() {
-      return System.identityHashCode(resourceFactory);
+      return Objects.hashCode(resourceFactory);
```
marianogonzalez: why?

estebanwasinger: because the `resourceFactory` will a `Reference` referencing a `ConnectionProvider`, due that different `Reference` instances pointing to the same `ConnectionProvider` could appear, the `identityHashCode` is useless

PR #5455, opened 2017-08-09
MULE-13286: FTP is not honoring connection timeout in some particular scenarios.

facundovs:
Ftp connection timeout is only being taken into account when the connection is stablished.
However we should use the same value to set the defaultTimeout, and the socketTimeout attributes in FTPClient. Otherwise, a socket can wait forever if server doesn't reply.


facundovs:
Maybe we should set a default value for socket timeouts because the default value is 0 that waits forever.

core/src/main/java/org/mule/retry/policies/AbstractPolicyTemplate.java
```diff
@@ -100,13 +100,14 @@ public RetryContext execute(RetryCallback callback, WorkManager workManager) thr
             }
             while (status.isOk());
 
+            context.setFailed(cause);
```
pablokraan: Why is this required?

transports/ftp/src/main/java/org/mule/transport/ftp/FtpConnectionFactory.java
```diff
@@ -77,11 +78,12 @@ public boolean validateObject(Object obj)
         FTPClient client = (FTPClient) obj;
         try
         {
+            client.setSoTimeout(connectionTimeout);
```
pablokraan: Don't we have a responseTimeout?

transports/ftp/src/main/java/org/mule/transport/ftp/FtpConnector.java
```diff
@@ -580,7 +590,15 @@ public Connector getWorkOwner()
             {
                 try
                 {
-                    getRetryPolicyTemplate().execute(callbackReconnection, muleContext.getWorkManager());
+                    RetryContext context = getRetryPolicyTemplate().execute(callbackReconnection, muleContext.getWorkManager());
+                    if(client[0] != null)
```
pablokraan: Fix code style

transports/ftp/src/main/java/org/mule/transport/ftp/FtpMessageReceiver.java
```diff
@@ -154,12 +154,19 @@ public Connector getWorkOwner()
         {
             try
             {
-                retryTemplate.execute(callbackReconnection, this.connector.getMuleContext().getWorkManager());
-                return filesToFTPArray(client[0]);
+                RetryContext retryContext = retryTemplate.execute(callbackReconnection, this.connector.getMuleContext().getWorkManager());
+                if(client[0] != null)
```
pablokraan: Fix code style

transports/ftp/src/test/java/org/mule/transport/ftp/AbstractFtpServerTestCase.java
```diff
@@ -210,4 +211,9 @@ protected File createDataFile(File folder, String encoding, final String testMes
 
         return target;
     }
+
+    protected Ftplet createFtpLet ()
```
pablokraan: Can be private?

transports/ftp/src/test/java/org/mule/transport/ftp/FtpConnectionSocketTimeoutTestCase.java
```diff
@@ -0,0 +1,207 @@
...
+            sleep();
+            return null;
+        }
+    };
+
+    private static void sleep()
+    {
+        try
+        {
+            Thread.sleep(SERVER_SLEEP);
```
pablokraan: Avoid the usage of sleep, try to use a latch or something like that

core/src/main/java/org/mule/retry/policies/AbstractPolicyTemplate.java
facundovs: When InterruptedException or InterruptedIOException are triggered, they are catched by the AbstractPolicy that after returns a RetryContext. In these scenarios, the code caller doesn't have any way to propagate the exception or to have information about this.

facundovs: In FTP Reconnection Strategy Scenario, we need to know if an InterruptedException was thrown and the FTPClient wasn't able to be created. As the exception is not propagated by the PolicyTemplate, al least it should to return this in the context

transports/ftp/src/test/java/org/mule/transport/ftp/AbstractFtpServerTestCase.java
facundovs: I created this method to able to override the FTPLet in subclasses.

transports/ftp/src/test/java/org/mule/transport/ftp/FtpConnectionSocketTimeoutTestCase.java
```diff
@@ -161,14 +160,11 @@ public FtpletResult onConnect(FtpSession session) throws FtpException, IOExcepti
 
     private static void sleep()
     {
-        try
+        long now = currentTimeMillis();
+        long stopTime = now + SERVER_SLEEP;
+        while (now < stopTime)
```
pablokraan: When I requested to avoid Thread.sleep I mean to use a latch or something that does not involves wasting time. Here you are using busy waiting, which is worst than the sleep because it has all the same problems and it also spends additional processing time

transports/ftp/src/test/java/org/mule/transport/ftp/FtpConnectionSocketTimeoutTestCase.java
```diff
@@ -0,0 +1,215 @@
...
+            sleep();
+            return null;
+        }
+    };
+
+    private static void sleep()
+    {
+        try
+        {
+            serverLatch.await(SERVER_SLEEP, TimeUnit.MILLISECONDS);
```
pablokraan: You must assert against the result of the await to be sure that the timeout was reached.

transports/ftp/src/test/java/org/mule/transport/ftp/FtpConnectionSocketTimeoutTestCase.java
```diff
@@ -0,0 +1,215 @@
...
+
+    private static String CONNECTION_TIMEOUT = "1000";
+
+    private static int TEST_DELTA = 2000;
+
+    private static int SERVER_DELTA = 3000;
+
+    private static long WAIT_TIMEOUT = parseLong(CONNECTION_TIMEOUT) + TEST_DELTA;
+
+    private static long SERVER_SLEEP = parseLong(CONNECTION_TIMEOUT) + SERVER_DELTA;
```
pablokraan: Now that you are using a latch, check if you can use smaller timeout values

PR #5456, opened 2017-08-09
MULE-13290: Extension model is missing types and extensions declare own any

afelisatti:


modules/extensions-spring-support/src/test/resources/models/marvel.json
```diff
@@ -3560,23 +3556,11 @@
...
       "type": "CUSTOM_ERROR",
       "namespace": "MARVEL",
       "parent": {
         "type": "ANY",
-        "namespace": "MARVEL",
-        "parent": {
-          "type": "ANY",
-          "namespace": "MULE"
-        }
+        "namespace": "MULE"
```
alegmarra: why are we losing the MARVEL:ANY here?

core/src/main/java/org/mule/runtime/core/api/exception/Errors.java
```diff
@@ -151,7 +149,14 @@
     public static final String SOURCE_ERROR_RESPONSE_GENERATE_ERROR_IDENTIFIER = "SOURCE_ERROR_RESPONSE_GENERATE";
 
     /**
-     * Indicates that a severe error occurred. Cannot be handled. Other unhandleable errors should go under it.
+     * Indicates that an unknown and unexpected error occurred. Cannot be handled directly, only through ANY.
+     */
+    public static final String UNKNOWN_ERROR_IDENTIFIER = "UNKNOWN";
+
+    // UNHANDLEABLE
```
alegmarra: probably better to communicate this in a more explicit way

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/ErrorsModelFactory.java
```diff
@@ -132,9 +133,12 @@ private String getErrorNamespace(ErrorTypeDefinition errorType) {
   private void addType(ErrorTypeDefinition<?> errorType,
                        Graph<ErrorTypeDefinition, Pair<ErrorTypeDefinition, ErrorTypeDefinition>> graph) {
     graph.addVertex(errorType);
-    ErrorTypeDefinition parentErrorType = errorType.getParent().orElse(ANY);
-    graph.addVertex(parentErrorType);
-    graph.addEdge(errorType, parentErrorType);
+    String type = errorType.getType();
+    if (!ANY.name().equals(type) && !CRITICAL.name().equals(type)) {
```
alegmarra: any reason not to allow ANY in other namespaces?

afelisatti: It messes up with the hierarchy, connectivity and retry are mapped to Mule's types, but any others to their own ANY. This makes it consistent. If we want a way to match all errors from an extension then we can add a special "namespace" matcher in the future. Today you could use the "when" attribute and use a custom expression.

core/src/main/java/org/mule/runtime/core/api/exception/Errors.java
afelisatti: Identifiers are rarely used alone. I'll add a distinction within the ComponentIdentifiers.

PR #5462, opened 2017-08-10
MULE-13233 - TLS configuration fails because of global element order

pablolagreca:


modules/spring-config/src/test/java/org/mule/runtime/config/spring/ApplicationContextsTestCase.java
```diff
@@ -1,156 +0,0 @@
...
-import org.mule.tck.testmodels.fruit.Seed;
-
-import org.junit.After;
-import org.junit.Rule;
-import org.junit.Test;
-import org.springframework.beans.factory.NoSuchBeanDefinitionException;
-import org.springframework.context.ApplicationContext;
-import org.springframework.context.support.ClassPathXmlApplicationContext;
-
-public class ApplicationContextsTestCase extends AbstractMuleTestCase {
```
afelisatti: Why is this test no longer relevant?

PR #5463, opened 2017-08-10
MULE-13154: delay logger creation after the flow is actually ran

alepulver:


modules/artifact/src/main/java/org/mule/module/artifact/classloader/ErrorHooksConfiguration.java
```diff
@@ -45,7 +44,15 @@
...
 
     // Log dropped events/errors rather than blow up which causes cryptic timeouts and stack traces.
-    Hooks.onErrorDropped(error -> logger.error("ERROR DROPPED UNEXPECTEDLY " + error));
-    Hooks.onNextDropped(event -> logger.error("EVENT DROPPED UNEXPECTEDLY " + event));
+    Hooks.onErrorDropped(error -> logError("ERROR DROPPED UNEXPECTEDLY " + error));
+    Hooks.onNextDropped(event -> logError("EVENT DROPPED UNEXPECTEDLY " + event));
+  }
+
+  private static void logError(String message) {
+    if (logger == null) {
```
pablokraan: Synchronize this

PR #5474, opened 2017-08-11
SEC-306: Add support for PGP Binary Encryption.

facundovs:


modules/pgp/src/main/java/org/mule/module/pgp/config/PGPMode.java
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+
+package org.mule.module.pgp.config;
+
+
+public enum PGPMode
```
alepulver: This could be renamed to PGPOutputMode or similar. Also related getters/setters.

PR #5479, opened 2017-08-11
MULE-12992: error declaration support for smart connectors

fernandezlautaro:


modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/loader/XmlExtensionLoaderDelegate.java
```diff
@@ -475,4 +487,30 @@ private MetadataType extractType(String receivedType) {
...
+            throw new IllegalModelDefinitionException(format("The operation [%s] cannot have an <error> with an empty 'type' attribute",
+                                                             operationName));
+          }
+          if (typeName.contains(NAMESPACE_SEPARATOR)) {
+            throw new IllegalModelDefinitionException(format("The operation [%s] cannot have an <error> [%s] that contains a reserved character [%s]",
+                                                             operationName, typeName,
+                                                             NAMESPACE_SEPARATOR));
+          }
+          operationDeclarer.withErrorModel(ErrorModelBuilder.newError(typeName, namespace)
+              .withParent(ErrorModelBuilder.newError(Errors.ComponentIdentifiers.ANY).build())
```
afelisatti: Does this mean you won't be supporting parents for now?

fernandezlautaro: it wont. 
It's messy to use the declarative part of a connector to define a hierarchy (discussed already with PLG)

Both smart connectors and runtime should have a place to declare the errors, and their hierarchy, and then use them as references.

PR #5482, opened 2017-08-14
MULE-11553 - fixing test cases

pablolagreca:


core-tests/src/test/java/org/mule/runtime/core/internal/construct/DefaultFlowBuilderTestCase.java
```diff
@@ -71,7 +74,7 @@ public void buildsSimpleFlow() throws Exception {
     assertThat(flow.getMuleContext(), is(muleContext));
     assertThat(flow.getProcessors(), is(empty()));
     assertThat(flow.getSource(), is(nullValue()));
-    assertThat(flow.getExceptionListener(), sameInstance(muleContext.getDefaultErrorHandler()));
+    assertThat(flow.getExceptionListener(), not(sameInstance(muleContext.getDefaultErrorHandler(Optional.empty()))));
```
dfeist: Why has this changed?

dfeist: Why has this changed? Even if `getDefaultErrorHandler` has changed internally, this still needs to test that Flow uses default error handler when one isn't set/

core-tests/src/test/java/org/mule/runtime/core/internal/routing/requestreply/AsyncRequestReplyRequesterTestCase.java
```diff
@@ -44,18 +44,18 @@
 import org.mule.tck.probe.JUnitLambdaProbe;
 import org.mule.tck.probe.PollingProber;
 
+import org.junit.After;
```
dfeist: minor: imports in wrong order, should be mule, java, others

core-tests/src/test/java/org/mule/runtime/core/privileged/processor/chain/DefaultMessageProcessorChainTestCase.java
```diff
@@ -73,10 +73,6 @@
 import org.mule.tck.junit4.AbstractReactiveProcessorTestCase;
 import org.mule.tck.size.SmallTest;
 
-import java.util.Collection;
```
dfeist: again, ideally fix you import order in IDE

core/src/main/java/org/mule/runtime/core/api/processor/MessageProcessors.java
```diff
@@ -41,64 +44,68 @@ private MessageProcessors() {
...
   /**
    * Creates a new explicit {@link MessageProcessorChain} from a {@link List} of {@link Processor}'s. Note that this performs
-   * chains construction but wil not inject {@link MuleContext} or {@link FlowConstruct} or perform any lifecycle.
+   * chains construction but wil not inject {@link MuleContext} or perform any lifecycle.
    *
    * @param processors list of processors to construct chains from.
    * @return new {@link MessageProcessorChain} instance.
    */
-  public static MessageProcessorChain newExplicitChain(List<Processor> processors) {
+  public static MessageProcessorChain newExplicitChain(Optional<ProcessingStrategy> processingStrategy,
```
dfeist: Q: Do we still 'explicit chains'? This was added to be able to have different flow paths/component location depending on is chain was internal or created by user. Unsure if this is still needed though.

core/src/main/java/org/mule/runtime/core/api/routing/SelectiveRouter.java
```diff
@@ -18,11 +18,11 @@
  */
 public interface SelectiveRouter extends Router {
 
-  void addRoute(String expression, MessageProcessorChain processor);
+  void addRoute(String expression, Processor processor);
```
dfeist: Why? Isn't it best that routers work with child chains rather than processors?  If they use child chains then we know processing strategy and notifications will be fired/used correctly, whereas with Processor the router than needs to create these chains internally.

core/src/main/java/org/mule/runtime/core/internal/enricher/MessageEnricher.java
```diff
@@ -116,9 +117,15 @@ protected InternalEvent enrich(final InternalEvent event, InternalEvent eventToE
     return eventToEnrich;
   }
 
+  @Override
+  public void initialise() throws InitialisationException {
+    enrichmentProcessor = newChain(MessageProcessors.getProcessingStrategy(muleContext, getRootContainerName()),
```
dfeist: static import?

core/src/main/java/org/mule/runtime/core/internal/exception/RedeliveryExceeded.java
```diff
@@ -30,7 +32,7 @@
 
   @Override
   public void initialise() throws InitialisationException {
-    configuredMessageProcessors = newChain(messageProcessors);
+    configuredMessageProcessors = newChain(empty(), messageProcessors);
```
dfeist: why empty processing strategy?

core/src/main/java/org/mule/runtime/core/internal/exception/TemplateOnErrorHandler.java
```diff
@@ -263,4 +279,10 @@ public void setHandleException(boolean handleException) {
   public void setErrorType(String errorType) {
     this.errorType = errorType;
   }
+
+  public void setRootContainerName(String rootContainerName) {
```
dfeist: Think it's seen this code (or very similar) in multiple places, ideally put the body in a static helper and reuse.

core/src/main/java/org/mule/runtime/core/internal/processor/ResponseMessageProcessorAdapter.java
```diff
@@ -20,9 +20,11 @@
 import org.mule.runtime.api.lifecycle.Stoppable;
 import org.mule.runtime.core.api.InternalEvent;
 import org.mule.runtime.core.api.MuleContext;
+import org.mule.runtime.core.api.construct.FlowConstruct;
```
dfeist: Unused imports?

core/src/main/java/org/mule/runtime/core/privileged/processor/chain/DefaultMessageProcessorChainBuilder.java
```diff
@@ -159,4 +177,54 @@ protected DefaultMessageProcessorChain(String name, Optional<ProcessingStrategy>
...
+      public void dispose() {
+        disposeIfNeeded(processor, LOGGER);
+      }
+
+      public void stop() throws MuleException {
+        stopIfNeeded(processor);
+      }
+
+      @Override
+      public InternalEvent process(InternalEvent event) throws MuleException {
```
dfeist: Implement `apply()` also, else this will fallback to blocking when delegating.

modules/spring-config/src/main/java/org/mule/runtime/config/spring/internal/factories/ResponseMessageProcessorsFactoryBean.java
```diff
@@ -6,35 +6,42 @@
...
 
   public void setMessageProcessors(List messageProcessors) {
     this.messageProcessors = messageProcessors;
   }
 
   @Override
-  public Object getObject() throws Exception {
+  public ResponseMessageProcessorAdapter doGetObject() throws Exception {
     DefaultMessageProcessorChainBuilder builder = new DefaultMessageProcessorChainBuilder();
+    //builder.setProcessingStrategy(flowConstruct.getProcessingStrategy());
```
dfeist: remove?

dfeist: remove

modules/spring-config/src/main/java/org/mule/runtime/config/spring/internal/factories/ResponseMessageProcessorsFactoryBean.java
```diff
@@ -46,20 +53,32 @@ public Object getObject() throws Exception {
       }
     }
     ResponseMessageProcessorAdapter responseAdapter = new ResponseMessageProcessorAdapter();
-    responseAdapter.setProcessor(builder.build());
+    responseAdapter.setProcessor(new Processor() {
+
+      private Processor processor;
+
+      @Override
+      public InternalEvent process(InternalEvent event) throws MuleException {
```
dfeist: Need to implement apply() also to delegate non-blocking invocation if this is required.

modules/spring-config/src/main/java/org/mule/runtime/config/spring/internal/factories/ResponseMessageProcessorsFactoryBean.java
```diff
@@ -46,20 +53,32 @@ public Object getObject() throws Exception {
...
+        if (processor == null) {
+          FlowConstruct flowConstruct = (FlowConstruct) registry.lookupByName(getRootContainerName()).get();
```
dfeist: Isn't there a static utility to lookup flowConstruct?

core-tests/src/test/java/org/mule/runtime/core/internal/construct/DefaultFlowBuilderTestCase.java
pablolagreca: No. Each flow needs it's own copy of the error handler. Take into account that error handler and the processors inside it support lifecycle. If each of them have the same copy then it means that if flow A and B are using the default error handling, doing stop over A will stop MPs inside B error handler

core/src/main/java/org/mule/runtime/core/api/processor/MessageProcessors.java
pablolagreca: Probably not. We would have to remove it and see what breaks

core/src/main/java/org/mule/runtime/core/api/routing/SelectiveRouter.java
pablolagreca: Discussed offline. Basically the processing strategy is created within a factory bean, so it's not possible to get a reference to it from other factory beans. That's why the chain needs to be created during initialise

core/src/main/java/org/mule/runtime/core/api/processor/MessageProcessors.java
elrodro83: Create an issue for this

modules/spring-config/src/main/java/org/mule/runtime/config/spring/internal/dsl/model/ComponentLocationVisitor.java
```diff
@@ -145,6 +150,12 @@ public void accept(ComponentModel componentModel) {
     componentModel.setComponentLocation(componentLocation);
   }
 
+  private boolean isRoute(ComponentModel componentModel) {
+    return componentModel.getIdentifier().equals(buildFromStringRepresentation("mule:route"))
```
elrodro83: use constansts for these.

PR #5485, opened 2017-08-14
MULE-11383: Packaging: plugins, services, policies, and so on, must be packaged as JAR files

pablokraan:


modules/service/pom.xml
```diff
@@ -52,5 +58,24 @@
             <version>${aetherProviderVersion}</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.mule.tools.maven</groupId>
+            <artifactId>mule-packager</artifactId>
+            <exclusions>
+                <!--
+                Excluding dependencies that are required to generate the classloader-model.json that are not required on runtime.
```
elrodro83: The functionality that requires these dependencies should be on a separate module so they don't have to be excluded by its clients. Create a JIRA to track this.

modules/service/src/main/java/org/mule/runtime/module/service/FileSystemServiceProviderDiscoverer.java
```diff
@@ -37,15 +38,18 @@
 
   private final ArtifactClassLoader apiClassLoader;
   private final ArtifactClassLoaderFactory<ServiceDescriptor> serviceClassLoaderFactory;
+  private final DescriptorLoaderRepository descriptorLoaderRepository;
 
   /**
    * Creates a new instance.
-   *
-   * @param containerClassLoader container artifact classLoader. Non null.
+   *  @param containerClassLoader container artifact classLoader. Non null.
```
elrodro83: odd identation

modules/service/src/main/resources/META-INF/mule-module.properties
```diff
@@ -1,3 +1,4 @@
 module.name=services
 
-artifact.export.classPackages=org.mule.runtime.module.service
+artifact.export.classPackages=org.mule.runtime.module.service,\
+  org.mule.runtime.module.artifact.internal.descriptor
```
elrodro83: can youy avoid exposing this internal package? or should it be renamed?

modules/service/src/main/java/org/mule/runtime/module/service/ServiceClassLoaderModelLoader.java
```diff
@@ -0,0 +1,101 @@
...
+  @Override
+  public ClassLoaderModel load(File artifactFile, Map<String, Object> attributes, ArtifactType artifactType)
+      throws InvalidDescriptorLoaderException {
+    if (artifactFile == null || !artifactFile.exists()) {
+      throw new IllegalArgumentException("Service folder does not exists: "
+          + (artifactFile != null ? artifactFile.getName() : null));
+    }
+
+    ClassLoaderModel.ClassLoaderModelBuilder classLoaderModelBuilder = new ClassLoaderModel.ClassLoaderModelBuilder();
+    for (URL url : getServiceUrls(artifactFile)) {
```
elrodro83: should services allow for lightweight packaging too?

pablokraan: No

PR #5486, opened 2017-08-14
MULE-9563: Support nonProxyHosts in http:proxy

BlockLucas:


modules/http-api/src/main/java/org/mule/runtime/http/api/client/proxy/ProxyConfig.java
```diff
@@ -41,6 +41,11 @@ static ProxyConfigBuilder builder() {
   String getPassword();
 
   /**
+   * @return A list of hosts separated by ",", which specifies that the proxy must not be used against those
```
afelisatti: `A list of comma separated hosts against which the proxy should not be used`

PR #5487, opened 2017-08-14
MULE-11804: Define APIs to dispatch notifications and register listen…

gsfernandes:
…ers - Fixed default values for configuration parameters when xsd validations are disabled


modules/spring-config/src/main/java/org/mule/runtime/config/spring/internal/dsl/model/CoreComponentBuildingDefinitionProvider.java
```diff
@@ -726,8 +727,13 @@ public void init() {
   }
 
   private TypeConverter<String, Predicate<? extends Notification>> getNotificationSubscriptionConverter() {
-    return subscription -> (notification -> subscription != null ? subscription
-        .equals(((AbstractServerNotification) notification).getResourceIdentifier()) : true);
+    return subscription -> {
+      if (subscription == ANY_SELECTOR_STRING) {
```
elrodro83: .equals

PR #5488, opened 2017-08-14
MULE-13312: XML validations might be disabled for ExtensionModel gene…

fernandezlautaro:
…ration


modules/extensions-xml-support/src/test/java/org/mule/runtime/extension/internal/loader/XmlExtensionLoaderTestCase.java
```diff
@@ -313,10 +338,66 @@ public void testModuleDocumentation() throws IOException {
     assertThat(modelProperty.get().getBodyComponentModel().getInnerComponents().size(), is(1));
   }
 
+  @Test
+  public void testModuleWrongBodyContent() {
```
pablolagreca: doesn't make sense to have this parameterized if the content of the test changes completely based on the input. Perhaps two methods are better?

PR #5491, opened 2017-08-15
MULE-11246: Improve PGP Module.

facundovs:
Currently the PGP module forces the user to provide a secureAliasId (the private key ID in decimal notation, something that isn't readily available from GnuPG for instance).
This is in principle not required as an encrypted message knows which key ID is required to decrypt it. By silently ignoring this information we make it difficult to detect configuration errors (entering the wrong key ID) and make it impossible to decrypt messages encrypted for different recipients.
A short term solution is to fail if the selected key ID doesn't match the one reported by the encrypted message, and a long term solution is to remove the secretAliasId altogether.
Additionally, we should consider PGP Decryption is not propagating the exceptions as expected. Error messages are not clear in the following scenarios:

- Incorrect PassPhrase.
- Invalid secretAliasId key.


distributions/standalone/src/main/resources/MIGRATION.txt
```diff
@@ -20,6 +20,7 @@ MULE-11825: In a DB template query, to set a DB param with the null value, you c
...
 MULE-12612: As FTP reconnection is at operation level, FTP connector does not support asynchronous reconnection strategies because it only makes sense if reconnection takes place during the start phase of the connector lifecycle. In case you use this kind of reconnection, please change them as follows: <reconnect blocking="true"/> inside FTP Connector, or just remove the blocking attribute.
 MULE-13164: Inconsistent null return from one-way VM inbound endpoint. From Mule 3.9 one-way inbound VM endpoints will now consistently return null to Flow using a request-response outbound endpoint or Mule Client using send(). (In previous versions a successful response led to null return yet an error resulted in a message being returned.)
+MULE-11246: The secretAliasId isn't already a mandatory parameter. If not given, Mule will take the secretAliasId from the message to decrypt. Additionally, from 3.9.x the secretAliasId must be a hexadecimal value.
```
alepulver: The secretAliasId isn't already a mandatory parameter. -> The secretAliasId parameter isn't mandatory anymore.
a hexadecimal -> an hexadecimal

PR #5497, opened 2017-08-15
MULE-12066: Support XA transactions in SDK extensions at source level

estebanwasinger:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -277,7 +277,18 @@ private TransactionConfig buildTransactionConfig() {
     MuleTransactionConfig transactionConfig = new MuleTransactionConfig();
     transactionConfig.setAction(toActionCode(sourceAdapter.getTransactionalAction()));
     transactionConfig.setMuleContext(muleContext);
-    transactionConfig.setFactory(transactionFactory);
+
+    TransactionType transactionalType = sourceAdapter.getTransactionalType();
+    switch (transactionalType) {
```
marianogonzalez: why is this different? The lookup logic should be able to pick both kinds of transaction factories

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceAdapter.java
```diff
@@ -351,32 +355,53 @@ public Source getDelegate() {
...
+      object = valueResolver.resolve(from(getInitialiserEvent(muleContext)));
     } catch (MuleException e) {
-      throw new MuleRuntimeException(createStaticMessage("Unable to get the Transactional Action value for Message Source"), e);
+      throw new MuleRuntimeException(createStaticMessage("Unable to get the " + type.getSimpleName()
+          + " value for Message Source"), e);
     }
 
-    if (!(transactionalAction instanceof SourceTransactionalAction)) {
-      throw new IllegalStateException("The resolved value is not a Transactional Action");
+    if (!(type.isAssignableFrom(object.getClass()))) {
```
marianogonzalez: type.isInstance

tests/test-extensions/transactional-extension/src/main/java/org/mule/test/transactional/TransactionalSource.java
```diff
@@ -9,35 +9,56 @@
...
 @MetadataScope(outputResolver = TransactionalMetadataResolver.class)
 public class TransactionalSource extends Source<TestTransactionalConnection, Object> {
 
+  private static final String IS_XA = "isXa";
+
+  public static Boolean isSuccess;
+  public static DummyXaResource xaResource;
+
+  @Parameter
+  TransactionType algoType;
```
marianogonzalez: algoType?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/InfrastructureTypeMapping.java
```diff
@@ -36,6 +36,7 @@
       .put(TlsContextFactory.class, new InfrastructureType(TLS_PARAMETER_NAME, 8))
       .put(SourceTransactionalAction.class, new InfrastructureType(TRANSACTIONAL_ACTION_PARAMETER_NAME, 6))
       .put(OperationTransactionalAction.class, new InfrastructureType(TRANSACTIONAL_ACTION_PARAMETER_NAME, 7))
+      .put(TransactionType.class, new InfrastructureType(TRANSACTIONAL_TYPE_PARAMETER_NAME, 9))
```
marianogonzalez: you also need to add tests in the Dsl and application model apis to make sure that we're able to serialize/deserialize an application using this feature to/from the DSL. Look at https://github.com/mulesoft/mule/commit/9c2efa45d7f4d30a942c17cf83bff7b7a32feafa for examples

estebanwasinger: I don't find any serialization/deserialization test case example in the referenced commit.
Also, when I implemented `OperationTransactionalAction` and `SourceTransactionalAction` I asked Ale if I needed to something else and he told that for this cases I was not necessary.

core/src/main/java/org/mule/runtime/core/internal/transaction/TransactionFactoryLocator.java
```diff
@@ -0,0 +1,56 @@
...
+   * @param type The {@link TransactionType} that the {@link TransactionFactory} should handle.
+   * @return An {@link Optional} {@link TransactionFactory}
+   */
+  public Optional<TransactionFactory> lookUpTransactionFactory(TransactionType type) {
+    return ofNullable(factories.computeIfAbsent(type, this::getTransactionFactory));
+  }
+
+  @Override
+  public void dispose() {
+    factories = null;
```
marianogonzalez: .clear(), or have the factories initialised in an `initialise()` method

core/src/main/java/org/mule/runtime/core/internal/transaction/TransactionFactoryLocator.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  private TransactionFactory getTransactionFactory(TransactionType transactionType) {
+    return getAvailableFactories().get(transactionType);
+  }
+
+  private Map<TransactionType, TypedTransactionFactory> getAvailableFactories() {
```
marianogonzalez: this is still inefficient. With this code, the SPI search will be performed at least twice. Why can't you just do it on initialise() and keep that map?

core/src/main/java/org/mule/runtime/core/internal/transaction/TransactionFactoryLocator.java
```diff
@@ -36,12 +37,16 @@
    * @return An {@link Optional} {@link TransactionFactory}
    */
   public Optional<TransactionFactory> lookUpTransactionFactory(TransactionType type) {
+    if (!initialized) {
```
marianogonzalez: I don't understand why you need this flag instead of just having an initialise phase. Assuming there's a good reason for that, you need to reset this on Dispose and use an AtomicBoolean instead, since I'm not sure this class won't be used concurrently

PR #5503, opened 2017-08-15
MULE-13009: Inconsistent error propagation in extension operations depending on operation type

juandesi:


core/src/main/java/org/mule/runtime/core/api/exception/NullExceptionHandler.java
```diff
@@ -0,0 +1,39 @@
...
+import org.mule.runtime.core.api.processor.MessageProcessorChain;
+import org.reactivestreams.Publisher;
+import reactor.core.publisher.Mono;
+
+/**
+ * Null {@link MessagingExceptionHandler} which can be used to configure a {@link MessageProcessorChain} to not handle errors.
+ *
+ * @since 4.0
+ */
+public class NullExceptionHandler implements MessagingExceptionHandler {
```
marianogonzalez: final

core/src/main/java/org/mule/runtime/core/api/util/MessagingExceptionResolver.java
```diff
@@ -0,0 +1,142 @@
...
+   * This method will pick the FIRST cause exception that has a mule or extension KNOWN error as the real cause, if there is not
+   * an exception in the causes that match with an Known error type then this method will try to find the error that the current
+   * {@link Event} is carrying.
+   * <p>
+   * When there are multiple exceptions that contains the same root error type, then this method will wrap the one that has
+   * highest position in the causes list
+   *
+   * @return a {@link MessagingException} with the proper {@link Error} associated to it's {@link InternalEvent}
+   */
+  public static MessagingException resolve(Processor processor, MessagingException me, MuleContext context) {
```
marianogonzalez: having this as a static method might be convenient, but prevents this from being mock and build tests in which a specific exception is resolved. 

Consider turning this into an instance method. If you have a good reason not to, then this class needs to be final and have a private constructor

core/src/main/java/org/mule/runtime/core/api/util/MessagingExceptionResolver.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Tries to resolve a new {@link MessagingException}s with the real cause of the problem from an incoming
+ * {@link MessagingException} that carries a list of causes inside it so it can be thrown and no information is lost.
+ *
+ * @since 4.0
+ */
+public class MessagingExceptionResolver {
```
marianogonzalez: what if an extension needs to use this? Shouldn't this be mule api?

juandesi: no way, this is 100% concern of the runtime

core/src/main/java/org/mule/runtime/core/api/util/ExceptionUtils.java
```diff
@@ -212,246 +147,114 @@ public static String getFullStackTraceWithoutMessages(Throwable throwable) {
...
-        .filter(error -> errorCauseMatchesException(causeException, error)
-            || messagingException.causedExactlyBy(error.getCause().getClass()))
-        .isPresent()) {
-      Error newError = getErrorFromFailingProcessor(annotatedObject, causeException, errorTypeLocator);
-      InternalEvent event = InternalEvent.builder(messagingException.getEvent()).error(newError).build();
-      messagingException.setProcessedEvent(event);
-      return event;
-    } else {
-      return currentEvent;
+  static List<ErrorMapping> getErrorMappings(Object processor) {
```
afelisatti: protected?

core/src/main/java/org/mule/runtime/core/api/util/ExceptionUtils.java
```diff
@@ -212,246 +147,114 @@ public static String getFullStackTraceWithoutMessages(Throwable throwable) {
...
-    if (failing == null) {
-      failing = processor;
-      exception = new MessagingException(createStaticMessage(rootException.getMessage()), exception.getEvent(),
-                                         rootException instanceof MessagingException ? rootException.getCause() : rootException,
-                                         processor);
-    }
-    exception.setProcessedEvent(createErrorEvent(exception.getEvent(), processor, exception, errorTypeLocator));
-    return putContext(exception, failing, exception.getEvent(), muleContext);
+  static boolean isUnknownMuleError(ErrorType type) {
+    return type.getIdentifier().equals(UNKNOWN_ERROR_IDENTIFIER);
```
afelisatti: Consider namespace as well.

core/src/main/java/org/mule/runtime/core/api/util/ExceptionUtils.java
```diff
@@ -212,246 +147,114 @@ public static String getFullStackTraceWithoutMessages(Throwable throwable) {
...
-    }
-
-    if (errorMappings != null && !errorMappings.isEmpty()) {
-      Optional<ErrorMapping> matchedErrorMapping = errorMappings.stream().filter(mapping -> mapping.match(errorType)).findFirst();
-      if (matchedErrorMapping.isPresent()) {
-        return matchedErrorMapping.get().getTarget();
-      }
-    }
-    return errorType;
+  private static ErrorType getErrorTypeFromFailingProcessor(Object processor, Throwable exception,
```
afelisatti: Do these have javadocs? hard to see in the diff...

core/src/main/java/org/mule/runtime/core/api/util/MessagingExceptionResolver.java
```diff
@@ -0,0 +1,142 @@
...
+
+  /**
+   * Resolves a new {@link MessagingException} with the real cause of the problem based on the content of an
+   * Incoming {@link MessagingException} with a chain of causes inside it and the current event that the exception is carrying.
```
afelisatti: incoming

core/src/main/java/org/mule/runtime/core/api/util/MessagingExceptionResolver.java
```diff
@@ -0,0 +1,142 @@
...
+
+    if (!rootCause.isPresent()) {
+      return updateCurrent(me, processor, context);
+    }
+
+    Throwable root = rootCause.get().getFirst();
+    ErrorType rootErrorType = rootCause.get().getSecond();
+    Processor failingProcessor = getFailingProcessor(me, root).orElse(processor);
+
+    ErrorType errorType = getErrorMappings(processor)
```
afelisatti: Why are mappings applied twice? Here and in `ExceptionUtils`.

core/src/main/java/org/mule/runtime/core/internal/exception/ErrorTypeLocatorFactory.java
```diff
@@ -85,6 +85,7 @@ public static ErrorTypeLocator createDefaultErrorTypeLocator(ErrorTypeRepository
                                  errorTypeRepository.lookupErrorType(STREAM_MAXIMUM_SIZE_EXCEEDED).get())
             .addExceptionMapping(MuleFatalException.class, errorTypeRepository.getErrorType(FATAL).get())
             .addExceptionMapping(CompositeRoutingException.class, errorTypeRepository.getErrorType(COMPOSITE_ROUTING).get())
+            .addExceptionMapping(LinkageError.class, errorTypeRepository.getErrorType(FATAL).get())
```
afelisatti: Why was this necessary? AFAIK that error should be wrapped to a MuleFatalException.

afelisatti:
Also, make sure you have tests cases for the UNKNOWN case, when there's nothing but that to be found.

core/src/main/java/org/mule/runtime/core/api/util/ExceptionUtils.java
juandesi: nope, those are trivial private methods

afelisatti:
@juandesi just let me know when the rebase is done and I'll approve it again.

PR #5510, opened 2017-08-17
Mule 11952

pablolagreca:


core/src/main/java/org/mule/runtime/core/internal/construct/DefaultFlowBuilder.java
```diff
@@ -239,6 +242,18 @@ public InternalEvent process(final InternalEvent event) throws MuleException {
           });
     }
 
+    @Override
+    protected CompletableFuture<Event> executeEvent(InternalEvent event) {
+      return Mono.just(event).transform(this).onErrorResume(throwable -> {
```
dfeist: Use `onErrorResume(MessagingException.class, me ->` instead.

PR #5517, opened 2017-08-17
MULE-13096: Improve version checker to use a library that supports se…

gsfernandes:
…mantic versioning

### DIFF
--- a/modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/descriptor/BundleDescriptorUtils.java
+++ b/modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/descriptor/BundleDescriptorUtils.java
@@ -7,12 +7,13 @@
 
 package org.mule.runtime.module.artifact.api.descriptor;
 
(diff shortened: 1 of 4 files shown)


modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/descriptor/BundleDescriptorUtils.java
```diff
@@ -36,33 +37,24 @@ public static boolean isCompatibleVersion(String availableVersion, String expect
       return true;
     }
 
-    Version available = getBundleVersion(availableVersion);
-    Version expected = getBundleVersion(expectedVersion);
+    Semver available = getBundleVersion(availableVersion);
```
marianogonzalez: what about MuleVersion class? Shouldn't it also use this?

gsfernandes: This is for plugins not for Mule artifacts. On plugins is where we support semantic versioning.

PR #5518, opened 2017-08-17
MULE-13018 - Container should not prepend mule/ to configuration files

pablolagreca:


modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/DeploymentServiceTestCase.java
```diff
@@ -868,7 +867,8 @@ public void redeploysExplodedAppOnStartup() throws Exception {
 
     reset(applicationDeploymentListener);
 
-    File configFile = new File(appsDir + "/" + dummyAppDescriptorFileBuilder.getDeployedPath(), MULE_CONFIG_XML_FILE);
+    File configFile = new File(appsDir + "/" + dummyAppDescriptorFileBuilder.getDeployedPath(),
```
pablokraan: "/"  > File.separator?

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/DefaultMuleApplication.java
```diff
@@ -337,7 +340,10 @@ public String getArtifactId() {
 
   @Override
   public File[] getResourceFiles() {
-    return descriptor.getConfigResourcesFile();
+    return descriptor.getConfigResources().stream()
+        .map(configFile -> new File(getLocation(), Paths.get("classes", configFile).toString()))
```
pablokraan: I guess there is a CLASSES_DIR constant somewhere

PR #5528, opened 2017-08-18
MULE-11383: Packaging: plugins, services, policies, and so on, must b…

pablokraan:
…e packaged as JAR files

_ Removing classloadermodel created for test services
_ Using same classloader model for services and server plugins

### DIFF
--- a/modules/service/src/main/java/org/mule/runtime/module/service/LibFolderClassLoaderModelLoader.java
+++ b/modules/service/src/main/java/org/mule/runtime/module/service/LibFolderClassLoaderModelLoader.java
@@ -8,6 +8,7 @@
 package org.mule.runtime.module.service;
 
 import static java.lang.String.format;
+import static java.util.Arrays.asList;
 import static org.mule.runtime.core.api.config.bootstrap.ArtifactType.SERVER_PLUGIN;
 import static org.mule.runtime.core.api.config.bootstrap.ArtifactType.SERVICE;
 import org.mule.runtime.core.api.config.bootstrap.ArtifactType;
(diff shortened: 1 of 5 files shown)


modules/service/src/main/java/org/mule/runtime/module/service/ServiceDescriptorFactory.java
```diff
@@ -78,7 +78,7 @@ public ServiceDescriptor create(File artifactFolder) throws ArtifactDescriptorCr
 
   private ClassLoaderModel createClassLoaderModel(File artifactFolder) {
     try {
-      return new ServiceClassLoaderModelLoader().load(artifactFolder, Collections.emptyMap(), ArtifactType.SERVICE);
+      return new LibFolderClassLoaderModelLoader().load(artifactFolder, Collections.emptyMap(), ArtifactType.SERVICE);
```
elrodro83: static import

modules/service/src/main/java/org/mule/runtime/module/service/LibFolderClassLoaderModelLoader.java
```diff
@@ -109,6 +111,6 @@ private void addDirectoryToClassLoader(List<URL> urls, File classesFolder) {
 
   @Override
   public boolean supportsArtifactType(ArtifactType artifactType) {
-    return SERVICE.equals(artifactType);
+    return SERVICE.equals(artifactType) || SERVER_PLUGIN.equals(artifactType);
```
elrodro83: put a static set of supported typed and check with `contains`

PR #5530, opened 2017-08-18
MULE-13317: support calling operations defined in the same module

fernandezlautaro:


modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/loader/XmlExtensionLoaderDelegate.java
```diff
@@ -272,7 +281,54 @@ private Document getModuleDocument(ExtensionLoadingContext context, URL resource
...
+        new XmlApplicationParser(new SpiServiceRegistry(), singletonList(currentThread().getContextClassLoader()));
+    Optional<ConfigLine> parseModule = xmlApplicationParser.parse(moduleDocument.getDocumentElement());
+    if (!parseModule.isPresent()) {
+      // This happens in org.mule.runtime.config.spring.dsl.processor.xml.XmlApplicationParser.configLineFromElement()
+      throw new IllegalArgumentException(format("There was an issue trying to read the stream of '%s'", resource.getFile()));
+    }
+    ComponentModelReader componentModelReader =
+        new ComponentModelReader(new DefaultConfigurationPropertiesResolver(empty(),
+                                                                            new SystemPropertiesConfigurationProvider()));
+    // TODO MULE-12291: we would, ideally, leave either the relative path (modulePath) or the URL to the current config file,
```
pablolagreca: This was already fix for application configuration files, can you fix it for modules?

PR #5533, opened 2017-08-18
CHHYBRID-3119: Added deploy/redeploy endpoints to support Configuration Management.

fpiagent:


pablokraan:
You need to create a JIRA in Mule so users in the community can see why you are changing this

PR #5535, opened 2017-08-18
MULE-13217 Rework foreach to be non-blocking and decoupled from splitter

dfeist:


core/src/main/java/org/mule/runtime/core/api/routing/ForkJoinStrategy.java
```diff
@@ -39,8 +39,20 @@
  */
 public interface ForkJoinStrategy {
 
+  /**
+   * Processes n routing pairs and aggregate the result as defined in the strategy implementation.
```
elrodro83: n -> `{@code n}`
aggregate -> aggregates

core/src/main/java/org/mule/runtime/core/api/routing/ForkJoinStrategy.java
```diff
@@ -39,8 +39,20 @@
...
+   * 
+   * @param original the original event
+   * @param routingPairs the routing paris to be
```
elrodro83: to be ?

core/src/main/java/org/mule/runtime/core/internal/routing/Foreach.java
```diff
@@ -60,98 +61,104 @@
...
+              responseBuilder.addVariable(counterVariableName, previousCounterVar);
             } else {
-              return e;
+              responseBuilder.removeVariable(counterVariableName);
             }
-          }).block();
-    } catch (Throwable throwable) {
-      throw rxExceptionToMuleException(throwable);
-    }
+            if (previousRootMessageVar != null) {
```
elrodro83: are this properties set again even in the case of errors? are there test cases for that scenario?

core/src/main/java/org/mule/runtime/core/internal/routing/Foreach.java
```diff
@@ -60,98 +61,104 @@
...
-    Object payload = message.getPayload().getValue();
-    if (payload instanceof Document || payload.getClass().getName().startsWith("org.dom4j.")) {
-      return message;
-    } else {
-      return muleContext.getTransformationService().internalTransform(message, DataType.fromType(Document.class));
-    }
+  private Flux<InternalEvent> splitAndProcess(InternalEvent request) {
+    AtomicInteger count = new AtomicInteger();
+    return fromIterable(() -> splittingStrategy.split(request))
+        .transform(p -> batchSize > 1 ? from(p).buffer(batchSize).map(typedValueListToTypedValue()) : p)
```
elrodro83: explain with comments for normal humans to understand

core/src/main/java/org/mule/runtime/core/internal/routing/ScatterGatherRouter.java
```diff
@@ -47,12 +48,15 @@
...
-        propagateWrappingFatal(new RoutePathNotFoundException(noEndpointsForRouter(), null));
-      }
-    };
+    return event -> validateMessageIsNotConsumable(event.getMessage());
+  }
+
+  @Override
+  public void initialise() throws InitialisationException {
+    super.initialise();
+    if (routes.size() < 2) {
```
elrodro83: previously only one route was valid. Is there any reason to don't allow that anymore?
I know it doesn't make much sense, just asking from a migration PoV

dfeist: This was a bug in my implementation, In 3.x minimum was two, also schema enforces 2.

core/src/main/java/org/mule/runtime/core/internal/routing/Foreach.java
dfeist: Good point.  If an error with splitting occurs then this is not an issue but if error processing a part occurrs then need to take this into account, even if it was ignored in 3.x.

PR #5557, opened 2017-08-23
MULE-12066: Support XA transactions in SDK extensions at source level…

estebanwasinger:
… - Fix possible NPE


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/DefaultSourceCallbackContext.java
```diff
@@ -69,9 +70,14 @@ public TransactionHandle bindConnection(Object connection) throws ConnectionExce
                                                                     sourceCallback.getConfigurationInstance(),
                                                                     connectionHandler);
 
-      transactionHandle = connection instanceof XATransactionalConnection
-          ? new XaTransactionHandle(muleContext.getTransactionManager())
-          : new DefaultTransactionHandle((TransactionalConnection) connection);
+      if (connection instanceof XATransactionalConnection) {
+        TransactionManager transactionManager = muleContext.getTransactionManager();
```
estebanwasinger: I retrieve the TX Manager inside the `If` so I don't have to go to the registry, only when the given connection can be part of an XA TX

PR #5573, opened 2017-08-25
MULE-13386: Credentials mask fails when the URI has special characters.

fsgonz:


core/src/test/java/org/mule/api/util/CredentialsMaskUtilTestCase.java
```diff
@@ -20,11 +24,23 @@
     private static final String URL_TEST = "jdbc:sqlserver://1.1.1.1:1443;databaseName=STAGING;user=mulesoftuser;password=mulesoftpass;";
 
     private static final String EXPECTED_URL_TEST_MASKED = "jdbc:sqlserver://1.1.1.1:1443;databaseName=STAGING;user=<<user>>;password=<<credentials>>;";
+    
+    private static final String URI_SPECIAL_CHARACTERS_AUTHORITY = "sftp://user:Deloitte@alskdalaks@@hol@localhost:22/home/user/";
```
elrodro83: is this a valid url? don't the `@` in the password have to be masked?

elrodro83: Use a more generic credentials

fsgonz:
You're right. This is not an issue. This does not affect the sftp connector when trying to mask the URI. Sorry for your time.

PR #5575, opened 2017-08-25
MULE-13009: Inconsistent error propagation in extension operations depending on operation type

juandesi:


juandesi:
Waiting for the pipeline job to validate the changes

PR #5583, opened 2017-08-27
MULE-13393: Fix DeployableFileBuilder is including repository folder …

balbifm:
…in artifact jar even with light package

### DIFF
--- a/modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/DeploymentServiceTestCase.java
+++ b/modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/DeploymentServiceTestCase.java
@@ -2011,25 +2011,6 @@ public class DeploymentServiceTestCase extends AbstractMuleTestCase {
   }
 
   @Test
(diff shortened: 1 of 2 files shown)


modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/DeploymentServiceTestCase.java
```diff
@@ -2011,25 +2011,6 @@ public void deploysApplicationWithPluginDependingOnPlugin() throws Exception {
   }
 
   @Test
-  public void deploysLightApplicationWithPluginDependingOnPlugin() throws Exception {
```
balbifm: Removing this test because with an actual light JAR, it fails when trying to get the plugin from maven since is not there

pablokraan: Do not remove it. Create a jira to fix it properly and then ignore it referencing the issue in the description

PR #5586, opened 2017-08-28
MULE-13382: JMS Message Listener doesn't release listener if an error…

estebanwasinger:
… occurs or the source stops

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
@@ -58,10 +58,21 @@ public class SourceConnectionManager {
    *
    * @param connection the connection to be released
    */
-  void disconnect(Object connection) {
+  void release(Object connection) {
     ConnectionHandler<Object> connectionHandler = connections.remove(new Reference<>(connection));
     if (connectionHandler != null) {
       connectionHandler.release();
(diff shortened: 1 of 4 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
```diff
@@ -59,9 +59,10 @@
    * @param connection the connection to be released
    */
   void disconnect(Object connection) {
-    ConnectionHandler<Object> connectionHandler = connections.remove(connection);
+    ConnectionHandler<Object> connectionHandler = connections.remove(new Reference<>(connection));
     if (connectionHandler != null) {
       connectionHandler.release();
+      connectionHandler.invalidate();
```
marianogonzalez: invalidate should only be called when there was an error.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
```diff
@@ -58,14 +58,26 @@
...
       connectionHandler.release();
     }
   }
 
   /**
+   * Disconnects the given connection
+   *
+   * @param connection the connection to be released
+   */
+  void disconnect(Object connection) {
```
marianogonzalez: rename to invalidate and update jdoc

PR #5589, opened 2017-08-28
MULE-13398: Propagation of SSL prevents Jackson serialization of InboundProperties.

fsgonz:


modules/http/src/main/java/org/mule/module/http/internal/listener/properties/SSLSessionProperties.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+package org.mule.module.http.internal.listener.properties;
+
+import java.io.Serializable;
+
+import javax.net.ssl.SSLSession;
+
+/**
+ * Wrapper to contain SSL context and avoid errors due to serialization
+ *
```
elrodro83: add `@since`

PR #5590, opened 2017-08-28
MULE-13399: ExtensionManager not stopped while undeploying operation

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/ExtensionActivator.java
```diff
@@ -134,5 +134,6 @@ public void stop() throws MuleException {
       stopIfNeeded(element);
     }
     lifecycleAwareElements.clear();
+    enumTypes.clear();
```
pablokraan: This change is not related to the issue being fixed and it does not have a test

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/ExtensionActivatorTestCase.java
```diff
@@ -0,0 +1,38 @@
...
+import static org.hamcrest.Matchers.hasSize;
+import static org.junit.Assert.assertThat;
+import org.mule.functional.junit4.ExtensionFunctionalTestCase;
+import org.mule.runtime.module.extension.internal.manager.DefaultExtensionManager;
+import org.mule.runtime.module.extension.internal.manager.ExtensionActivator;
+import org.mule.test.heisenberg.extension.HeisenbergExtension;
+
+import org.junit.Test;
+
+public class ExtensionActivatorTestCase extends ExtensionFunctionalTestCase {
```
pablokraan: ExtensionFunctionalTestCase is not supposed to be used anymore

PR #5602, opened 2017-08-30
MULE-12703: Expression language does not handle context properties

afelisatti:


core/src/main/java/org/mule/runtime/core/el/function/PropertyAccessFunction.java
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.el.function;
```
pablolagreca: shouldn't this belong to package internal?

core/src/main/java/org/mule/runtime/core/el/function/MuleFunctionsBindingContextProvider.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.el.function;
```
pablolagreca: shouldn't this belong to package internal?

PR #5608, opened 2017-08-31
MULE-13426: Add cluster support for API-Gateway policies

mtqp:
We are testing *DefaultApplicationPolicyInstance* in API-Gateway side using integration tests for standalone and cluster scenarios.

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/time/LocalTimeSupplier.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/time/LocalTimeSupplier.java
@@ -6,6 +6,8 @@
  */
 package org.mule.runtime.core.internal.time;
 
+import static java.lang.System.currentTimeMillis;
+
 import org.mule.runtime.core.api.time.TimeSupplier;
 
 import java.util.function.Supplier;
@@ -24,6 +26,6 @@ public class LocalTimeSupplier implements TimeSupplier {
    */
   @Override
   public Long get() {
-    return System.currentTimeMillis();
+    return currentTimeMillis();
   }
 }
--- a/core/src/main/java/org/mule/runtime/core/internal/time/LocalTimeSupplier.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/time/LocalTimeSupplier.java
@@ -4,16 +4,18 @@
  * license, a copy of which has been included with this distribution in the
  * LICENSE.txt file.
  */
-package org.mule.runtime.core.api.time;
+package org.mule.runtime.core.internal.time;
+
+import org.mule.runtime.core.api.time.TimeSupplier;
 
 import java.util.function.Supplier;
 
 /**
- * A {@link Supplier} which provides a unified time in milliseconds.
+ * A {@link Supplier} which provides the current system time in milliseconds.
  *
  * @since 4.0
  */
-public interface TimeSupplier extends Supplier<Long> {
+public class LocalTimeSupplier implements TimeSupplier {
 
   /**
    * Returns {@link System#currentTimeMillis()}
@@ -21,5 +23,7 @@ public interface TimeSupplier extends Supplier<Long> {
    * @return the current time in milliseconds
    */
   @Override
-  Long get();
+  public Long get() {
+    return System.currentTimeMillis();
+  }
 }
--- a/modules/container/src/main/java/org/mule/runtime/container/api/MuleInvocationHandler.java
+++ b/modules/container/src/main/java/org/mule/runtime/container/api/MuleInvocationHandler.java
@@ -21,35 +21,36 @@ import java.util.LinkedList;
 import java.util.List;
 
 /**
- * Extends {@link InvocationHandler} to provide and expose metadata about the inner {@link Service} implementation.
+ * Extends {@link InvocationHandler} to provide and expose metadata about the inner object implementation.
  * <p>
- * This allows for nested {@link Service} {@link Proxy}ies to work as expected.
+ * This allows for nested object {@link Proxy}ies to work as expected.
  * 
  * @since 4.0
  */
-public abstract class ServiceInvocationHandler implements InvocationHandler {
+public abstract class MuleInvocationHandler<T> implements InvocationHandler {
 
-  private final Service service;
+  private final T innerObject;
 
   /**
    * Creates a new proxy for the provided service instance.
    *
-   * @param service service instance to wrap. Non null.
+   * @param innerObject object instance to wrap. Non null.
    */
-  protected ServiceInvocationHandler(Service service) {
-    checkArgument(service != null, "service cannot be null");
-    this.service = service;
+  protected MuleInvocationHandler(T innerObject) {
+    checkArgument(innerObject != null, "service cannot be null");
+    this.innerObject = innerObject;
   }
 
   /**
    * @return the methods declared in the implementation of the proxied service.
    */
-  protected Method[] getServiceImplementationDeclaredMethods() {
-    if (isProxyClass(getService().getClass()) && getInvocationHandler(getService()) instanceof ServiceInvocationHandler) {
-      return ((ServiceInvocationHandler) getInvocationHandler(getService())).getServiceImplementationDeclaredMethods();
+  protected Method[] getImplementationDeclaredMethods() {
+    if (isProxyClass(getProxiedObject().getClass())
+        && getInvocationHandler(getProxiedObject()) instanceof MuleInvocationHandler) {
+      return ((MuleInvocationHandler) getInvocationHandler(getProxiedObject())).getImplementationDeclaredMethods();
     } else {
       List<Method> methods = new LinkedList<>();
-      Class<?> clazz = getService().getClass();
+      Class<?> clazz = getProxiedObject().getClass();
       while (clazz != Object.class) {
         methods.addAll(asList(clazz.getDeclaredMethods()));
         clazz = clazz.getSuperclass();
@@ -65,11 +66,12 @@ public abstract class ServiceInvocationHandler implements InvocationHandler {
    * See {@link InvocationHandler#invoke(Object, Method, Object[])}
    */
   protected Object doInvoke(Object proxy, Method method, Object[] args) throws Throwable {
-    if (isProxyClass(getService().getClass()) && getInvocationHandler(getService()) instanceof ServiceInvocationHandler) {
-      return ((ServiceInvocationHandler) getInvocationHandler(getService())).invoke(getService(), method, args);
+    if (isProxyClass(getProxiedObject().getClass())
+        && getInvocationHandler(getProxiedObject()) instanceof MuleInvocationHandler) {
+      return getInvocationHandler(getProxiedObject()).invoke(getProxiedObject(), method, args);
     } else {
       try {
-        return method.invoke(getService(), args);
+        return method.invoke(getProxiedObject(), args);
       } catch (InvocationTargetException ite) {
         // Unwrap target exception to ensure InvocationTargetException (in case of unchecked exceptions) or
         // UndeclaredThrowableException (in case of checked exceptions) is not thrown by Service instead of target exception.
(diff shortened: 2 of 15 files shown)


modules/container/src/main/java/org/mule/runtime/container/api/MuleInvocationHandler.java
```diff
@@ -21,35 +21,36 @@
...
- * Extends {@link InvocationHandler} to provide and expose metadata about the inner {@link Service} implementation.
+ * Extends {@link InvocationHandler} to provide and expose metadata about the inner object implementation.
  * <p>
- * This allows for nested {@link Service} {@link Proxy}ies to work as expected.
+ * This allows for nested object {@link Proxy}ies to work as expected.
  * 
  * @since 4.0
  */
-public abstract class ServiceInvocationHandler implements InvocationHandler {
+public abstract class MuleInvocationHandler<T> implements InvocationHandler {
```
pablokraan: Can you find a name that is more descriptive than MuleInvocationHandler based on what this class does?

modules/container/src/main/java/org/mule/runtime/container/api/MuleInvocationHandler.java
```diff
@@ -65,11 +66,12 @@ protected ServiceInvocationHandler(Service service) {
    * See {@link InvocationHandler#invoke(Object, Method, Object[])}
    */
   protected Object doInvoke(Object proxy, Method method, Object[] args) throws Throwable {
-    if (isProxyClass(getService().getClass()) && getInvocationHandler(getService()) instanceof ServiceInvocationHandler) {
-      return ((ServiceInvocationHandler) getInvocationHandler(getService())).invoke(getService(), method, args);
+    if (isProxyClass(getProxiedObject().getClass())
+        && getInvocationHandler(getProxiedObject()) instanceof MuleInvocationHandler) {
```
pablokraan: This condition seems to be duplicated on lines 48-49

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultApplicationPolicyInstance.java
```diff
@@ -90,11 +97,26 @@ private void initPolicyContext() throws InitialisationException {
...
                                                                                      new DefaultExtensionManagerFactory()));
 
-    artifactBuilder.withServiceConfigurator(customizationService -> customizationService
-        .overrideDefaultServiceImpl(MuleProperties.OBJECT_POLICY_MANAGER_STATE_HANDLER,
-                                    application.getMuleContext().getRegistry()
-                                        .lookupObject(MuleProperties.OBJECT_POLICY_MANAGER_STATE_HANDLER)));
-
+    artifactBuilder.withServiceConfigurator(customizationService -> {
+      MuleRegistry applicationRegistry = application.getMuleContext().getRegistry();
+      customizationService.overrideDefaultServiceImpl(OBJECT_LOCK_PROVIDER,
```
pablokraan: Why don't you use createLifecycleFilterProxy for these objects?

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/proxy/LifecycleFilterProxy.java
```diff
@@ -0,0 +1,61 @@
...
+import org.mule.runtime.container.api.MuleInvocationHandler;
+
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+
+/**
+ * Proxies an object instance to filter invocations (without raising an exception) of lifecycle methods from {@link Startable},
+ * {@link Stoppable} and {@link Disposable} interfaces.
+ */
+public class LifecycleFilterProxy<T> extends MuleInvocationHandler<T> {
```
pablokraan: Add @since 4.0

modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/proxy/LifecycleFilterProxyTestCase.java
```diff
@@ -0,0 +1,91 @@
...
+
+import org.mule.runtime.api.exception.MuleException;
+import org.mule.runtime.api.lifecycle.Disposable;
+import org.mule.runtime.api.lifecycle.Startable;
+import org.mule.runtime.api.lifecycle.Stoppable;
+
+import org.junit.Before;
+import org.junit.Test;
+
+public class LifecycleFilterProxyTestCase {
```
pablokraan: Extend from AbstractMuleTestCase

modules/container/src/main/java/org/mule/runtime/container/api/MuleInvocationHandler.java
mtqp: I agree with you, had trouble choosing one as _InvocationHandler_ was already taken by the interface, so I took the lazy road... As it adds a reflective capabilities (reading all methods from the inner object), what do you think about *ReflectiveInvocationHandler*? If you come up with anything better please let me know!

modules/container/src/main/java/org/mule/runtime/container/api/MuleInvocationHandler.java
mtqp: That is correct, already extracted into a method, thanks!

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/proxy/LifecycleFilterProxy.java
mtqp: Done

modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/proxy/LifecycleFilterProxyTestCase.java
mtqp: Done

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultApplicationPolicyInstance.java
mtqp: The @Inject in our extension failed if I proxied them. I checked though, and no dispose method was executed on those without the proxy so the policy lifecycle shouldn't be altering anything from app's context.
Nevertheless, do you see a way to overcome the @Inject error so that we can proxy everything?

pablokraan: Can you add more context about why the extension failed? If failed because there was no available object to inject, then you probably you need to add another interface to the proxy in order to satisfy that requirement

modules/container/src/main/java/org/mule/runtime/container/api/MuleInvocationHandler.java
pablokraan: "Reflective"  is redeundant as every invocation handler uses reflection. Based on the class javadoc, I would use MetadataInvocationHandler or something similar

modules/container/src/main/java/org/mule/runtime/container/api/ReflectiveInvocationHandler.java
```diff
@@ -21,35 +21,35 @@
...
-  protected ServiceInvocationHandler(Service service) {
-    checkArgument(service != null, "service cannot be null");
-    this.service = service;
+  protected ReflectiveInvocationHandler(T innerObject) {
+    checkArgument(innerObject != null, "service cannot be null");
+    this.innerObject = innerObject;
   }
 
   /**
    * @return the methods declared in the implementation of the proxied service.
```
pablokraan: Fix all the javadocs to use obejct instead of service

core/src/main/java/org/mule/runtime/core/internal/time/LocalTimeSupplier.java
```diff
@@ -0,0 +1,29 @@
...
+public class LocalTimeSupplier implements TimeSupplier {
+
+  /**
+   * Returns {@link System#currentTimeMillis()}
+   *
+   * @return the current time in milliseconds
+   */
+  @Override
+  public Long get() {
+    return System.currentTimeMillis();
```
pablokraan: add static import

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultApplicationPolicyInstance.java
```diff
@@ -110,7 +111,8 @@ private void initPolicyContext() throws InitialisationException {
       customizationService.overrideDefaultServiceImpl(OBJECT_POLICY_MANAGER_STATE_HANDLER,
```
pablokraan: Add a comment explaining why this object is not proxied

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultImplicitConfigurationProviderFactory.java
```diff
@@ -78,7 +79,7 @@ public ConfigurationProvider createImplicitConfigurationProvider(ExtensionModel
         return new DynamicConfigurationProvider(providerName, extensionModel, implicitConfigurationModel,
                                                 resolverSet,
                                                 implicitConnectionProviderValueResolver,
-                                                ImmutableExpirationPolicy.getDefault(new TimeSupplier()), muleContext);
+                                                ImmutableExpirationPolicy.getDefault(new LocalTimeSupplier()), muleContext);
```
pablokraan: What happens when a extension is used in a cluster? Should not use the same time supplier used in the provided muleContext?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/LifecycleAwareConfigurationInstance.java
```diff
@@ -347,7 +348,7 @@ public ConfigurationState getState() {
 
   private void initStats() {
     if (timeSupplier == null) {
-      timeSupplier = new TimeSupplier();
+      timeSupplier = new LocalTimeSupplier();
```
pablokraan: Same here

core/src/main/java/org/mule/runtime/core/internal/time/LocalTimeSupplier.java
mtqp: Great, done.

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultApplicationPolicyInstance.java
mtqp: Done.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultImplicitConfigurationProviderFactory.java
mtqp: I double checked with the SDK team, and this is only used for the expiration of the configuration.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/LifecycleAwareConfigurationInstance.java
mtqp: Same.

PR #5609, opened 2017-08-31
MULE-7211: Fix Flaky Tests in Mule

elrodro83:
* Attempt at fix for expiration test cases
* Troubleshooting info for nonblocking failures


core/src/main/java/org/mule/runtime/core/api/util/monitor/ExpiryMonitor.java
```diff
@@ -116,28 +116,32 @@ public void resetExpirable(Expirable expirable) {
...
-        // TODO MULE-863: What should we really do?
-        logger.debug(e.getMessage());
+    synchronized (monitors) {
+      for (Object element : monitors.values()) {
+        holder = (ExpirableHolder) element;
+        removeExpirable(holder.getExpirable());
+        try {
+          holder.getExpirable().expired();
+        } catch (Exception e) {
+          // TODO MULE-863: What should we really do?
```
pablokraan: Does this TODO come from mule 1.x? The jira number is prety old. If that is correct, remove it

PR #5611, opened 2017-08-31
MULE-13422: SDK register invalid errors for extension sources

juandesi:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/ExtensionErrorsRegistrant.java
```diff
@@ -93,7 +93,7 @@ protected void onOperation(OperationModel model) {
 
   private ErrorType getErrorType(ErrorModel errorModel) {
     ComponentIdentifier identifier = createIdentifier(errorModel.getType(), errorModel.getNamespace());
-    Optional<ErrorType> optionalError = errorTypeRepository.lookupErrorType(identifier);
+    Optional<ErrorType> optionalError = errorTypeRepository.getErrorType(identifier);
```
afelisatti: Modify the addErrorType method of the repository as well to include a check of the internal map.

core/src/main/java/org/mule/runtime/core/api/exception/DefaultErrorTypeRepository.java
```diff
@@ -88,27 +90,38 @@ public DefaultErrorTypeRepository() {
...
+    return errorType;
   }
 
   /**
    * {@inheritDoc}
    */
   @Override
   public ErrorType addInternalErrorType(ComponentIdentifier errorTypeIdentifier, ErrorType parentErrorType) {
-    return addErrorTypeTo(errorTypeIdentifier, parentErrorType, this.internalErrorTypes);
+    checkIfErrorIsAlreadyRegistered(errorTypeIdentifier, internalErrorTypes);
```
afelisatti: Shouldn't we check both maps each time? We won't need collapsing the maps into one then as above.

PR #5615, opened 2017-08-31
MULE-13094: Embedded Container: If app fails to deploy, trying to deploy same app again ends instantly - Refactor

lucianoRM:


modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/descriptor/ArtifactDescriptor.java
```diff
@@ -58,6 +58,14 @@ public void setRootFolder(File rootFolder) {
   }
 
   /**
+   * Returns a {@link File} representing the descriptor file
+   * @return the descriptor file
+   */
+  public File getDescriptorFile() {
```
pablokraan: Move this to DeployableArtifactDescriptor

modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/DefaultArchiveDeployer.java
```diff
@@ -303,7 +303,8 @@ public void deployArtifact(T artifact) throws DeploymentException {
   private void addZombieApp(Artifact artifact) {
     if (allResourcesExist(artifact.getResourceFiles())) {
       try {
-        artifactZombieMap.put(artifact.getArtifactName(), new ZombieArtifact(artifact.getResourceFiles()));
+        artifactZombieMap.put(artifact.getArtifactName(),
```
pablokraan: When the artifact has redeploymentEnabled=false do not include the descriptor in the list

modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/DefaultArchiveDeployer.java
```diff
@@ -303,7 +307,13 @@ public void deployArtifact(T artifact) throws DeploymentException {
   private void addZombieApp(Artifact artifact) {
```
pablokraan: Change parameter type from Artifact to DeployableArtifact and remove redundant casts

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/DeploymentServiceTestCase.java
```diff
@@ -633,12 +636,8 @@ public void deploysBrokenAppZipOnStartup() throws Exception {
...
     assertApplicationAnchorFileDoesNotExists(brokenAppFileBuilder.getId());
 
-    final Map<URI, Long> zombieMap = deploymentService.getZombieApplications();
-    assertEquals("Wrong number of zombie apps registered.", 1, zombieMap.size());
-    final Map.Entry<URI, Long> zombie = zombieMap.entrySet().iterator().next();
-    assertEquals("Wrong URL tagged as zombie.", brokenAppFileBuilder.getDeployedPath(),
-                 new File(zombie.getKey()).getName());
-    assertTrue("Invalid lastModified value for file URL.", zombie.getValue() != -1);
+    assertZombieApplication(brokenAppFileBuilder.getDeployedPath(), 1, false);
+
```
pablokraan: Do not add redundant lines

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/DeploymentServiceTestCase.java
```diff
@@ -4205,12 +4123,43 @@ private void executeApplicationFlow(String flowName) throws MuleException {
...
+    final Map<String, Map<URI, Long>> zombieMap = deploymentService.getZombieApplications();
+    assertZombie(zombieMap, appId, expectedZombieCount, parent);
+  }
+
+  private void assertZombieDomain(String domainId, int expectedZombieCount, boolean parent, boolean expectName) {
+    final Map<String, Map<URI, Long>> zombieMap = deploymentService.getZombieDomains();
+    assertZombie(zombieMap, domainId, expectedZombieCount, parent);
+  }
+
+  private void assertZombie(Map<String, Map<URI, Long>> zombieMap, String appId, int expectedZombieCount, boolean parent) {
```
pablokraan: Is not clear what the "parent" parameter does. Can you split the method and make it more clear in the name which are the differences usages?

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/DeploymentServiceTestCase.java
```diff
@@ -4205,12 +4123,43 @@ private void executeApplicationFlow(String flowName) throws MuleException {
...
+      assertThat("Invalid lastModified value for file URL.", zombieMap.get(appId), not(hasValue(-1)));
+    }
+  }
+
+  private Map.Entry<URI, Long> getZombieFromMap(Predicate<Map.Entry> filter, Map<String, Map<URI, Long>> zombieMap) {
+    Map.Entry<URI, Long> zombieEntry = null;
+    //For every app in zombieMap
+    for (Map<URI, Long> zombieResource : zombieMap.values()) {
+      try {
+        zombieEntry = zombieResource.entrySet().stream().filter(filter).collect(Collectors.toList()).get(0);
```
pablokraan: Use findFirst and check if there is a value instead of relying on an exception being thrown

modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/DefaultArchiveDeployer.java
lucianoRM: Casts are still necessary because they are for the ArtifactDescriptor not the Artifact. Should i still change the method argument?

modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/DefaultArchiveDeployer.java
```diff
@@ -300,10 +302,16 @@ public void deployArtifact(T artifact) throws DeploymentException {
     }
   }
 
-  private void addZombieApp(Artifact artifact) {
+  private void addZombieApp(DeployableArtifact artifact) {
     if (allResourcesExist(artifact.getResourceFiles())) {
       try {
-        artifactZombieMap.put(artifact.getArtifactName(), new ZombieArtifact(artifact.getResourceFiles()));
+        List<File> resourceFiles = new ArrayList<>();
+        resourceFiles.addAll(Arrays.asList(artifact.getResourceFiles()));
```
pablokraan: static import

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/DeploymentServiceTestCase.java
```diff
@@ -4205,12 +4121,40 @@ private void executeApplicationFlow(String flowName) throws MuleException {
         .build());
   }
 
-  private void assertZombieApplication(String appId) {
-    final Map<URI, Long> zombieMap = deploymentService.getZombieApplications();
-    assertEquals("Wrong number of zombie apps registered.", 1, zombieMap.size());
-    final Map.Entry<URI, Long> zombie = zombieMap.entrySet().iterator().next();
-    assertEquals("Wrong URL tagged as zombie.", appId, new File(zombie.getKey()).getParentFile().getName());
-    assertTrue("Invalid lastModified value for file URL.", zombie.getValue() != -1);
+  private void assertZombieApplication(String appId, int expectedZombieCount, boolean compareAgainstParentFileName) {
```
pablokraan: Is very hard to understand how to invoke this method. 
ExpectdZombieCount always receives a 1 or 0, so is basically a boolean.
compareAgainstParentFileName is another boolean, but when must be true/false?
Make the four possible combination of parameter clear in a method name (ie: add method for all of them). 
You can move all these utility functions to a separated class, so you can make visible only the new methods and keep using the old method internally

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/DeploymentServiceTestCase.java
```diff
@@ -4205,12 +4121,40 @@ private void executeApplicationFlow(String flowName) throws MuleException {
...
+    final Map<String, Map<URI, Long>> zombieMap = deploymentService.getZombieApplications();
+    assertZombie(zombieMap, appId, expectedZombieCount, compareAgainstParentFileName);
+  }
+
+  private void assertZombieDomain(String domainId, int expectedZombieCount, boolean compareAgainstParentFileName) {
+    final Map<String, Map<URI, Long>> zombieMap = deploymentService.getZombieDomains();
+    assertZombie(zombieMap, domainId, expectedZombieCount, compareAgainstParentFileName);
+  }
+
+  private void assertZombie(Map<String, Map<URI, Long>> zombieMap, String appId, int expectedZombieCount,
```
pablokraan: appId -> artifactId

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/DeploymentServiceTestCase.java
```diff
@@ -4205,12 +4121,40 @@ private void executeApplicationFlow(String flowName) throws MuleException {
...
+                   getZombieFromMap((entry) -> new File((URI) entry.getKey()).getName().equals(appId), zombieMap),
+                   is(notNullValue()));
+      }
+      assertThat("Invalid lastModified value for file URL.", zombieMap.get(appId), not(hasValue(-1)));
+    }
+  }
+
+  private Map.Entry<URI, Long> getZombieFromMap(Predicate<Map.Entry> filter, Map<String, Map<URI, Long>> zombieMap) {
+    Map.Entry<URI, Long> zombieEntry = null;
+    //For every app in zombieMap
```
pablokraan: app -> artifact

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/DeploymentServiceTestCase.java
lucianoRM: The idea was to be able to check whether an artifact was a zombie, even if it was not the only one. Maybe in the future, we added another test where the number of zombies was bigger than 1 and we still can use this method. As for now, you are right. I will leave it as it was, just checking for 1 zombie present.

pablokraan: If you want to check for multiple zombies, you will need to pass multiple artifactId and multiple compareAgainstParentFileName values

modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/api/DeployableArtifactDescriptor.java
```diff
@@ -83,6 +83,14 @@ public void setPlugins(Set<ArtifactPluginDescriptor> plugins) {
     this.plugins = plugins;
   }
 
+  /**
+   * Returns a {@link File} representing the descriptor file
+   * @return the descriptor file
```
pablokraan: Add extra line between method and parameter descirptions

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/DeploymentServiceTestCase.java
```diff
@@ -4205,12 +4115,26 @@ private void executeApplicationFlow(String flowName) throws MuleException {
...
+    if (!zombieMap.containsKey(artifactName)) {
+      Map.Entry<URI, Long> zombieEntry =
+          getZombieFromMap((entry) -> new File((URI) entry.getKey()).getName().equals(artifactName), zombieMap);
+      assertThat("Wrong URL tagged as zombie.", zombieEntry, is(notNullValue()));
+    }
+  }
+
+  private Map.Entry<URI, Long> getZombieFromMap(Predicate<Map.Entry> filter, Map<String, Map<URI, Long>> zombieMap) {
+    Map.Entry<URI, Long> zombieEntry = null;
+    //For every artifact in zombieMap
```
pablokraan: Remove comment

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/DeploymentServiceTestCase.java
```diff
@@ -4205,12 +4115,26 @@ private void executeApplicationFlow(String flowName) throws MuleException {
...
+    for (Map<URI, Long> zombieResource : zombieMap.values()) {
+      zombieEntry = zombieResource.entrySet().stream().filter(filter).findFirst().orElse(null);
```
pablokraan: Should not exit once zombieEntry is not null?

PR #5626, opened 2017-09-01
MULE-13394: Reconnect forever retry policy doesn't reconnect at all

marianogonzalez:


core/src/main/java/org/mule/runtime/core/internal/retry/reactor/AbstractRetry.java
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
```
dfeist: We need to conserve original license header AFAU, can't just copy in and change license.

marianogonzalez: i can if I acknowledge where it came from

dfeist: I'm not a layer, but it's fairly clear: 

> You must retain, in the Source form of any Derivative Works that You distribute, all copyright, patent, trademark, and attribution notices

PR #5636, opened 2017-09-03
MULE-13188: ValueProviderService creates a connection although this m…

estebanwasinger:
…ay not be required - Fix broken scenarios

----
The changes were made to be able to resolve the connection in a lazy way. Since until now, the connection handler was solved although this was not necessary.
----
Fixes Tooling Client test cases

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/resolving/DefaultExtensionResolvingContext.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/resolving/DefaultExtensionResolvingContext.java
@@ -6,16 +6,17 @@
  */
 package org.mule.runtime.core.internal.resolving;
 
-import static java.util.Optional.empty;
-import static java.util.Optional.of;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/mule/runtime/core/internal/resolving/DefaultExtensionResolvingContext.java
```diff
@@ -95,8 +101,16 @@ public BaseTypeBuilder getTypeBuilder() {
    */
   @Override
   public void dispose() {
-    if (connectionHandler != null) {
-      connectionHandler.release();
+    try {
+      connectionHandler.get().ifPresent(ConnectionHandler::release);
```
marianogonzalez: if you don't require a connection, you here craete it just in order to release it. INstead of a supplier you should use a LAzyValue and use the ifComputed() function

PR #5638, opened 2017-09-03
MULE-13291: Implement "raise-error"

afelisatti:


core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -273,6 +274,28 @@ private void declareLogger(ExtensionDeclarer extensionDeclarer, ClassTypeLoader
...
+    OperationDeclarer raiseError = extensionDeclarer.withOperation("raiseError")
+        .describedAs("Throws an error with the specified type and description.");
+
+    raiseError.withOutput().ofType(typeLoader.load(void.class));
+    raiseError.withOutputAttributes().ofType(typeLoader.load(void.class));
+
+    raiseError.onDefaultParameterGroup()
+        .withRequiredParameter("type")
+        .ofType(BaseTypeBuilder.create(JAVA).stringType()
+            .enumOf("ANY", "REDELIVERY_EXHAUSTED", "TRANSFORMATION", "EXPRESSION", "SECURITY", "CLIENT_SECURITY",
```
marianogonzalez: this is lying. This will tell the UI to render a closed selector with these values , but the processor itself allows to throw any kind of errors. Considering the creation of the APP namespace, this doesn't seem intended to me.

PR #5643, opened 2017-09-04
MULE-13400: All schedulers view allows changing their state

elrodro83:


tests/unit/src/main/java/org/mule/tck/SimpleUnitTestSupportSchedulerService.java
```diff
@@ -197,4 +204,38 @@ void stoppedScheduler(Scheduler scheduler) {
   public int getScheduledTasks() {
     return scheduler.getScheduledTasks();
   }
+
+  private class DefaultSchedulerView implements SchedulerView {
```
afelisatti: TestSchedulerView?

PR #5659, opened 2017-09-06
MULE-12768 Message sources need to be able to define how they want th…

dfeist:
…e Flow to handle back pressure.

### DIFF
--- a/core-tests/src/test/java/org/mule/runtime/core/internal/processor/strategy/AbstractProcessingStrategyTestCase.java
+++ b/core-tests/src/test/java/org/mule/runtime/core/internal/processor/strategy/AbstractProcessingStrategyTestCase.java
@@ -17,7 +17,6 @@ import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.Matchers.greaterThanOrEqualTo;
 import static org.hamcrest.Matchers.lessThanOrEqualTo;
 import static org.junit.Assert.assertThat;
-import static org.junit.Assert.fail;
 import static org.mule.runtime.core.DefaultEventContext.create;
 import static org.mule.runtime.core.api.construct.Flow.builder;
 import static org.mule.runtime.core.api.context.notification.MessageProcessorNotification.MESSAGE_PROCESSOR_POST_INVOKE;
@@ -27,6 +26,7 @@ import static org.mule.runtime.core.api.lifecycle.LifecycleUtils.setMuleContextI
 import static org.mule.runtime.core.api.processor.ReactiveProcessor.ProcessingType.BLOCKING;
 import static org.mule.runtime.core.api.processor.ReactiveProcessor.ProcessingType.CPU_LITE;
 import static org.mule.runtime.core.api.processor.ReactiveProcessor.ProcessingType.CPU_LITE_ASYNC;
+import static org.mule.runtime.core.api.rx.Exceptions.rxExceptionToMuleException;
 import static org.mule.runtime.core.internal.processor.strategy.AbstractProcessingStrategyTestCase.Mode.SOURCE;
 import static org.mule.runtime.core.internal.util.rx.Operators.requestUnbounded;
 import static org.mule.tck.probe.PollingProber.DEFAULT_POLLING_INTERVAL;
@@ -83,7 +83,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
-import java.util.function.Predicate;
 import java.util.function.Supplier;
 
 import org.hamcrest.Matcher;
@@ -98,6 +97,7 @@ import org.junit.runners.Parameterized;
 import org.reactivestreams.Publisher;
 import org.slf4j.Logger;
 import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
 
 @RunWith(Parameterized.class)
 public abstract class AbstractProcessingStrategyTestCase extends AbstractMuleContextTestCase {
@@ -177,8 +177,6 @@ public abstract class AbstractProcessingStrategyTestCase extends AbstractMuleCon
   protected Scheduler asyncExecutor;
   protected ExecutorService cachedThreadPool = newFixedThreadPool(4);
 
-
-
   @Rule
   public ExpectedException expectedException = ExpectedException.none();
 
@@ -202,6 +200,7 @@ public abstract class AbstractProcessingStrategyTestCase extends AbstractMuleCon
 
     flowBuilder = () -> builder("test", muleContext)
         .processingStrategyFactory((muleContext, prefix) -> createProcessingStrategy(muleContext, prefix))
+        .source(triggerableMessageSource)
         // Avoid logging of errors by using a null exception handler.
         .messagingExceptionHandler((exception, event) -> event);
   }
@@ -230,11 +229,11 @@ public abstract class AbstractProcessingStrategyTestCase extends AbstractMuleCon
 
   @Test
   public void singleCpuLight() throws Exception {
-    flow = flowBuilder.get().source(triggerableMessageSource).processors(cpuLightProcessor).build();
+    flow = flowBuilder.get().processors(cpuLightProcessor).build();
 
     flow.initialise();
     flow.start();
-    process(flow, testEvent());
+    processFlow(testEvent());
   }
 
   @Test
@@ -254,17 +253,17 @@ public abstract class AbstractProcessingStrategyTestCase extends AbstractMuleCon
 
     List<Processor> processors = new ArrayList<>(asList(processorsBeforeLatch));
     processors.add(latchedProcessor);
-    flow = flowBuilder.source(triggerableMessageSource).processors(processors).build();
+    flow = flowBuilder.processors(processors).build();
     flow.initialise();
     flow.start();
 
     for (int i = 0; i < invocations; i++) {
-      asyncExecutor.submit(() -> process(flow, newEvent()));
+      asyncExecutor.submit(() -> processFlow(newEvent()));
     }
 
     latchedProcessor.getAllLatchedLatch().await();
 
-    asyncExecutor.submit(() -> process(flow, newEvent()));
+    asyncExecutor.submit(() -> processFlow(newEvent()));
 
     assertThat(latchedProcessor.getUnlatchedInvocationLatch().await(BLOCK_TIMEOUT, MILLISECONDS), is(!blocks));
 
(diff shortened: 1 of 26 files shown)


core-tests/src/test/java/org/mule/runtime/core/internal/processor/strategy/AbstractProcessingStrategyTestCase.java
```diff
@@ -456,6 +505,51 @@ protected void testAsyncCpuLightNotificationThreads(AtomicReference<Thread> befo
     process(flow, testEvent());
   }
 
+  protected void testBackPressure(BackPressureStrategy backPressureStrategy, Predicate<Integer> processedAssertion,
```
elrodro83: why not use `Matcher`s instead pf predicates?

core-tests/src/test/java/org/mule/runtime/core/internal/processor/strategy/WorkQueueProcessingStrategyTestCase.java
```diff
@@ -46,6 +52,14 @@ public WorkQueueProcessingStrategyTestCase(Mode mode) {
     super(mode);
   }
 
+  @Before
+  public void before() throws RegistrationException {
+    super.before();
+    // This processing strategy depends on blocking scheduler not rejecting work from callee thread in order to apply
```
elrodro83: callee?

core/src/main/java/org/mule/runtime/core/api/source/MessageSource.java
```diff
@@ -17,7 +19,39 @@
...
+   * The {@link BackPressureStrategy} that {@link org.mule.runtime.core.api.construct.FlowConstruct} implementations that support
+   * the concept of back-pressure should use.
+   *
+   * @return the {@link BackPressureStrategy} to use.
+   */
+  default BackPressureStrategy getBackPressureStrategy() {
+    return WAIT;
+  }
+
+  enum BackPressureStrategy {
```
elrodro83: add javadoc, include a definition of backpressure

core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
```diff
@@ -221,6 +222,35 @@ public InternalEvent process(InternalEvent event) throws MuleException {
...
+        } else {
+          if (source.getBackPressureStrategy() == DROP) {
+            return empty();
+          } else {
+            RejectedExecutionException rejectedExecutionException = new RejectedExecutionException(format(OVERLOAD_ERROR_MESSAGE,
+                                                                                                          AbstractPipeline.this
+                                                                                                              .getName()));
+            return error(exceptionResolver.resolve(new MessagingException(builder(event)
+                .error(ErrorBuilder.builder().errorType(overloadErrorType)
+                    .description(format("Flow '%s' Busy.", getName()))
```
elrodro83: make error messages consistent with `OVERLOAD_ERROR_MESSAGE`

core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
```diff
@@ -221,6 +222,35 @@ public InternalEvent process(InternalEvent event) throws MuleException {
...
+          .flatMap(event -> Mono.from(event.getContext().getResponsePublisher()));
+    } else {
+      return publisher -> from(publisher).flatMap(event -> {
+        if (sink.emit(event)) {
+          return event.getContext().getResponsePublisher();
+        } else {
+          if (source.getBackPressureStrategy() == DROP) {
+            return empty();
+          } else {
+            RejectedExecutionException rejectedExecutionException = new RejectedExecutionException(format(OVERLOAD_ERROR_MESSAGE,
```
elrodro83: maybe subclass `RejectedExecutionException` for this use?

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/AbstractStreamProcessingStrategyFactory.java
```diff
@@ -51,7 +52,8 @@
   private static final String SYSTEM_PROPERTY_PREFIX = AbstractStreamProcessingStrategyFactory.class.getName() + ".";
 
   public static final int DEFAULT_BUFFER_SIZE = getInteger(SYSTEM_PROPERTY_PREFIX + "DEFAULT_BUFFER_SIZE", SMALL_BUFFER_SIZE);
-  public static final int DEFAULT_SUBSCRIBER_COUNT = getInteger(SYSTEM_PROPERTY_PREFIX + "DEFAULT_SUBSCRIBER_COUNT", 1);
+  public static final int DEFAULT_SUBSCRIBER_COUNT =
```
elrodro83: what does 12 mean here?

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/TransactionAwareWorkQueueStreamProcessingStrategyFactory.java
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Creates {@link TransactionAwareWorkQueueStreamProcessingStrategy} instances that de-multiplexes incoming messages using a
+ * ring-buffer but instead of processing events using a constrained {@link SchedulerService#cpuLightScheduler()}, or by using the
+ * proactor pattern, instead simply performs all processing on a larger work queue pool using a fixed number of threads from the
+ * {@link SchedulerService#ioScheduler()}.
+ *
+ * @since 4.0
+ */
+public class TransactionAwareWorkQueueStreamProcessingStrategyFactory extends WorkQueueStreamProcessingStrategyFactory {
```
elrodro83: many happy returns

core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
```diff
@@ -221,6 +222,35 @@ public InternalEvent process(InternalEvent event) throws MuleException {
     initialiseIfInitialisable(pipeline);
   }
 
+  private ReactiveProcessor dispatchToFlow(Sink sink) {
```
elrodro83: add comments to explain what this is doing in a western language

core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
dfeist: What value do you see in doing this?

core-tests/src/test/java/org/mule/runtime/core/internal/processor/strategy/AbstractProcessingStrategyTestCase.java
```diff
@@ -502,33 +469,37 @@ protected void testAsyncCpuLightNotificationThreads(AtomicReference<Thread> befo
...
-            }
-          })).maxConcurrency(2).build();
+          flowBuilder.get()
+              .source(triggerableMessageSource)
+              .processors(asList(cpuLightProcessor, new ThreadTrackingProcessor() {
+
+                @Override
+                public InternalEvent process(InternalEvent event) throws MuleException {
+                  try {
+                    Thread.sleep(3);
```
elrodro83: the 3 is still here

core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
```diff
@@ -222,26 +222,40 @@ public InternalEvent process(InternalEvent event) throws MuleException {
...
   }
 
+  /*
+   * Processor that dispatches incoming source Events to the internal pipeline the Sink. The way in which the Event is dispatched
+   * and how overload is handled depends on the Source back-pressure strategy.
+   */
   private ReactiveProcessor dispatchToFlow(Sink sink) {
     if (source.getBackPressureStrategy() == WAIT) {
-      return publisher -> from(publisher).doOnNext(event -> sink.accept(event))
+      // If back-pressure strategy is WAIT then using blocking `accept(Event event)` to dispatch Event
```
elrodro83: using -> use

core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
```diff
@@ -222,26 +222,40 @@ public InternalEvent process(InternalEvent event) throws MuleException {
...
           .flatMap(event -> Mono.from(event.getContext().getResponsePublisher()));
     } else {
+      // If back-pressure strategy is FAIL/DROP then using back-pressure aware `accept(Event event)` to dispatch Event
       return publisher -> from(publisher).flatMap(event -> {
         if (sink.emit(event)) {
           return event.getContext().getResponsePublisher();
         } else {
           if (source.getBackPressureStrategy() == DROP) {
+            // If Event is not accepted and the back-pressure strategy is DROP then drop Event.
             return empty();
```
elrodro83: should we at leas log a warning when an event is dropped? Having the id of the dropped event may help troubleshooting

core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
```diff
@@ -222,26 +222,40 @@ public InternalEvent process(InternalEvent event) throws MuleException {
...
           } else {
-            RejectedExecutionException rejectedExecutionException = new RejectedExecutionException(format(OVERLOAD_ERROR_MESSAGE,
-                                                                                                          AbstractPipeline.this
-                                                                                                              .getName()));
+            // If Event is not accepted and the back-pressure strategy is FAIL then respond to Source with an OVERLOAD error.
+            RejectedExecutionException rejectedExecutionException =
```
elrodro83: no subclass then?

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/AbstractStreamProcessingStrategyFactory.java
```diff
@@ -52,6 +52,8 @@
   private static final String SYSTEM_PROPERTY_PREFIX = AbstractStreamProcessingStrategyFactory.class.getName() + ".";
 
   public static final int DEFAULT_BUFFER_SIZE = getInteger(SYSTEM_PROPERTY_PREFIX + "DEFAULT_BUFFER_SIZE", SMALL_BUFFER_SIZE);
+
+  // Use more than one subscriber by default on machines with large number of cores. 3 on 24-core, 2 on 8-core, 1 on < 8 cores
```
elrodro83: reword to `1 on < 8 cores, 2 on < 8-core, 3 on < 24-core, and so on`

core-tests/src/test/java/org/mule/runtime/core/internal/processor/strategy/AbstractProcessingStrategyTestCase.java
dfeist: I've ensured this is as reliable as possible to avoid flaky tests.  I attempted to do this with a latch, but not only is it very hard, it's not a realistic back-pressure test that simulates a slow-consumer rather than latched processors.  If we see this is flaky then can revisit but i'm convinced that given loose assertions it won't be an issue.

core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
dfeist: If a source chooses to use DROP, then it's because it's processing a continuous stream of data such as temperatures and it is valid for the use case to sample data rather than process all data points. As such I don't think it makes sense to add any logging here.

core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
dfeist: I asked, in a comment, why you thought this was valuable, maybe comment got lost. The reason i don't see much value is because Error is the primary way of declaring and propagating errors not, not the exception so I don't see value in sub-class personally.

PR #5671, opened 2017-09-08
MULE-13485: improved macro expansion to from O(n^2) to O(V+E) if the …

fernandezlautaro:
…library we rely on is not lying to us. Also, each ExtensionModel is visited just once when macro expanding.


fernandezlautaro:
screen shot 2017-09-07 at 11 13 42 pm

How interesting is that `O(N^2)` is just about 4 lines of code, and all of the graph with topological sort `O(V+E)` about 400+.

That's for you @pablokraan. 😗

PR #5672, opened 2017-09-08
MULE-13403 ProcessorChains are not being synchronised properly

dfeist:


core/src/main/java/org/mule/runtime/core/privileged/processor/CompositeProcessorChainRouter.java
```diff
@@ -7,79 +7,94 @@
...
 
   @Override
   protected ReactiveProcessor getExecutableFunction() {
-    return publisher -> from(publisher).transform(messageProcessorChain);
+    return publisher -> from(publisher).flatMapMany(initial -> fromIterable(processorChains).reduce(initial, processChain()));
+  }
+
+  private BiFunction<InternalEvent, MessageProcessorChain, InternalEvent> processChain() {
+    return (event, processorChain) -> {
+      InternalEventContext childContext = child(event.getContext(), ofNullable(getLocation()));
```
elrodro83: shouldn't the location of the scatter gather be the location of the component this router?

PR #5688, opened 2017-09-11
MULE-13489: Add a test to verify that http request timeout is honored…

facundovs:
… in redirection scenarios.

### DIFF
--- a/modules/http/src/test/java/org/mule/module/http/functional/requester/AbstractHttpRedirectTimeoutTestCase.java
+++ b/modules/http/src/test/java/org/mule/module/http/functional/requester/AbstractHttpRedirectTimeoutTestCase.java
@@ -23,16 +23,17 @@ import org.junit.Rule;
 public class AbstractHttpRedirectTimeoutTestCase extends AbstractHttpRequestTestCase
 {
 
-    private static long DELAY ;
-    @Rule
-    public SystemProperty timeoutProperty ;
-
     private String REDIRECT_URL = format("http://localhost:%s/%s", httpPort.getNumber(), "secondPath");
+
+    @Rule
+    public SystemProperty timeoutProperty;
+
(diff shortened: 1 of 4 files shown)


modules/http/src/test/java/org/mule/module/http/functional/requester/HttpRedirectExceededTimeoutTestCase.java
```diff
@@ -0,0 +1,42 @@
...
+    public void testRedirectTimeout() throws Exception
+    {
+        try
+        {
+            runFlow("requestFlow");
+            fail("Timeout exception must be triggered");
+        }
+        catch(Exception timeoutException)
+        {
+            assertThat(timeoutException.getCause(), instanceOf(TimeoutException.class));
```
afelisatti: Use `ExpectedException` instead.

modules/http/src/test/java/org/mule/module/http/functional/requester/AbstractHttpRedirectTimeoutTestCase.java
```diff
@@ -0,0 +1,65 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.eclipse.jetty.server.Request;
+import org.junit.Rule;
+
+public class AbstractHttpRedirectTimeoutTestCase extends AbstractHttpRequestTestCase
+{
+
+    private static long DELAY ;
```
afelisatti: This should be an instance variable.

afelisatti: Also, check the format here and in the `Rule` below.

PR #5690, opened 2017-09-11
MULE-13496: Adding uri and query params to HttpMessage test class

santiagopurucker:


tests/functional/src/main/java/org/mule/functional/util/http/HttpMessage.java
```diff
@@ -27,6 +30,8 @@ public HttpMessage(HttpExchange httpExchange) {
     headerNames.stream()
         .forEach(headerName -> headersBuilder.putAll(headerName, httpExchange.getRequestHeaders().get(headerName)));
     this.headers = headersBuilder.build();
+    uriParams = httpExchange.getRequestURI().getPath();
```
afelisatti: This would be just `uri` at this point.

tests/functional/src/main/java/org/mule/functional/util/http/HttpMessage.java
```diff
@@ -36,4 +41,29 @@ public HttpMessage(HttpExchange httpExchange) {
...
+
+  public Map<String, String> getQueryParams() {
+    return queryParams;
+  }
+
+  public String getUriParams() {
+    return uriParams;
+  }
+
+  private Map<String, String> queryToMap(String query) {
```
afelisatti: This will not support repeated query keys.

tests/functional/src/main/java/org/mule/functional/util/http/HttpMessage.java
```diff
@@ -6,18 +6,22 @@
...
 package org.mule.functional.util.http;
 
+import org.mule.runtime.api.util.MultiMap;
 import org.mule.runtime.core.api.util.IOUtils;
 
 import com.google.common.collect.ImmutableMultimap;
 import com.google.common.collect.Multimap;
 import com.sun.net.httpserver.HttpExchange;
 
+import java.util.HashMap;
```
afelisatti: Unused imports.

PR #5693, opened 2017-09-11
MULE-13363: Allow access to session properties from privileged API

elrodro83:
* Move MuleSession to privileged. Refactor event builders to account for
this


core/src/main/java/org/mule/runtime/core/api/policy/OperationPolicyProcessor.java
```diff
@@ -65,27 +68,33 @@ public BaseEvent process(BaseEvent operationEvent) throws MuleException {
...
-      BaseEvent variablesProviderEvent =
-          latestPolicyState.orElseGet(() -> BaseEvent.builder(operationEvent.getContext()).message(of(null)).build());
-      policyStateHandler.updateState(policyStateId, variablesProviderEvent);
-      BaseEvent policyEvent = policyEventConverter.createEvent(operationEvent, variablesProviderEvent);
-      Processor operationCall = buildOperationExecutionWithPolicyFunction(nextProcessor, operationEvent);
-      policyStateHandler.updateNextOperation(policyStateId.getExecutionIdentifier(), operationCall);
-      return executePolicyChain(operationEvent, policyStateId, policyEvent);
-    });
+    return from(publisher)
+        .cast(PrivilegedEvent.class)
```
pablolagreca: Are we completely sure there's no way a BaseEvent can be send as parameter?. Consider all other casts in the code

elrodro83: BaseEvent is just an interface, as well as PrivilegedEvent. The implementation is the same in either case.
That being said, there is https://www.mulesoft.org/jira/browse/MULE-13494 pending to review these casts.

PR #5705, opened 2017-09-13
MULE-13510: InputResolvers for ArrayType parameters, doesn't show the…

estebanwasinger:
…m as array types

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataInputDelegate.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataInputDelegate.java
@@ -8,15 +8,20 @@ package org.mule.runtime.module.extension.internal.metadata;
 
 import static java.lang.String.format;
 import static java.util.stream.Collectors.toList;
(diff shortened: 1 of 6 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataInputDelegate.java
```diff
@@ -122,4 +127,18 @@ NamedTypeResolver getParameterResolver(String parameterName) {
       return failure(parameter.getType(), newFailure(e).onParameter(parameter.getName()));
     }
   }
+
+  private MetadataType adaptToListIfNecessary(MetadataType resolvedType, ParameterModel parameter,
```
alegmarra: can't use `org.mule.runtime.api.meta.Typed` and unify this with the output resolving logic?

estebanwasinger: Yes, but the reusable insignificant because `Output` case has more things to consider like: `Pagination`, `List<Result>`, Attributes Handling, etc

PR #5727, opened 2017-09-15
MULE-13204: Remove dom4j and jaxen exported packages from core libs

elrodro83:


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/schema/DefaultExtensionSchemaGenerator.java
```diff
@@ -82,7 +85,9 @@ public void onSource(SourceModel model) {
 
     schemaBuilder.registerEnums();
 
-    return renderSchema(schemaBuilder.build());
+    // Make sure the xml libs use the container classloader internally
```
pablokraan: xml -> XML

PR #5731, opened 2017-09-18
MULE-13559: Warnings with password disclosed in Spring Registry.

fsgonz:


modules/spring-config/src/main/java/org/mule/config/spring/SpringRegistry.java
```diff
@@ -448,7 +450,7 @@ private void doRegisterObject(String key, Object value) throws RegistrationExcep
                 {
                     if (logger.isWarnEnabled())
                     {
-                        logger.warn(String.format("Spring registry already contains an object named '%s'. The previous object will be overwritten.", key));
+                        logger.warn(String.format("Spring registry already contains an object named '%s'. The previous object will be overwritten.", maskUrlPassword(key, BARE_URL_PATTERN)));
```
elrodro83: i would mask the key of the object being reistered. Otherwise, this log entry that was intended for troubleshooting looses it's meaning.

fsgonz: The object is registered in EndpointAliasesCoreExtension (gateway), so I close this PR.

PR #5744, opened 2017-09-19
MULE-12399: Validate Pojos used as Subtypes

pablomusumeci:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/InputParametersTypeModelValidator.java
```diff
@@ -0,0 +1,141 @@
...
+      }
+
+      private void validateCallback(SourceModel model, Optional<SourceCallbackModel> callback) {
+        callback.ifPresent(cb -> cb.getAllParameterModels()
+            .forEach(parameterModel -> validateJavaType(model, parameterModel.getType(), problems, validatedTypes)));
+      }
+    }.walk(extensionModel);
+
+    extensionModel.getSubTypes().forEach(subTypesModel -> getClassForValidation(subTypesModel.getBaseType())
+        .filter(validatedTypes::contains).ifPresent(type -> {
```
marianogonzalez: this will ONLY validate the oneas already validated. You seem to be missing a not and to not have good enough tests

PR #5756, opened 2017-09-20
MULE-11600: Lifecycle is incorrectly applied after application deployment fails.

fsgonz:


fsgonz:
Sorry. I have to verify some tests.

PR #5762, opened 2017-09-20
MULE-12716: Remove registries from API

elrodro83:
* Second pass, support for privileged api clients

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/privileged/registry/LegacyRegistryUtils.java
+++ b/core/src/main/java/org/mule/runtime/core/privileged/registry/LegacyRegistryUtils.java
@@ -22,7 +22,11 @@ import java.util.Collection;
  * @deprecated inject {@link Registry} where possible instead of using this utility.
  */
 @Deprecated
-public class LegacyRegistryUtils {
+public final class LegacyRegistryUtils {
+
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/mule/runtime/core/privileged/registry/LegacyRegistryUtils.java
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Provides a way for privileged API clients to access certain functionality of the internal {@link MuleRegistry}.
+ *
+ * @since 4.0
+ *
+ * @deprecated inject {@link Registry} where possible instead of using this utility.
+ */
+@Deprecated
+public class LegacyRegistryUtils {
```
pablokraan: add private constructor

tests/unit/src/main/java/org/mule/tck/util/MuleContextUtils.java
```diff
@@ -85,6 +92,23 @@ public static void mockNotificationsHandling(final MuleRegistry registry) {
     }
   }
 
+  public static void mockRegistry(MuleContext context) {
```
pablokraan: Add javadocs

PR #5783, opened 2017-09-21
MULE-11870: Adding test case to verify custom ocsp responder behavior

facundovs:


modules/http/src/test/java/org/mule/module/http/functional/requester/HttpStandardRevocationConfig.java
```diff
@@ -0,0 +1,26 @@
...
+    @Override
+    protected String getConfigFile()
+    {
+        return "http-requester-standard-revocation-config.xml";
+    }
+
+    @Test
+    public void acceptsEmptyTrustStore() throws Exception
+    {
+
```
afelisatti: Why is this test empty?

modules/http/src/test/java/org/mule/module/http/functional/requester/ocsp/AbstractHttpOcspRevocationTestCase.java
```diff
@@ -0,0 +1,105 @@
...
+        if (certAlias != null)
+        {
+            certAliasSystemProperty = new SystemProperty("certAlias", certAlias);
+            dynamicOcspPort = new DynamicPort("dynamicOcspPort");
+            ocspPort = dynamicOcspPort.getNumber();
+            ocspResponder = certAliasSystemProperty.getValue();
+        }
+        else
+        {
+            ocspPort = FIXED_OCSP_PORT ;
```
afelisatti: format

modules/http/src/test/java/org/mule/module/http/functional/requester/ocsp/AbstractHttpOcspRevocationTestCase.java
```diff
@@ -0,0 +1,105 @@
...
+     * The entity4 certificate has a hardcoded ocsp url: http://localhost:1111
+     */
+    static final int FIXED_OCSP_PORT = 1111;
+
+
+    static final String DEFAULT_OCSP_RESPONDER= "server";
+
+    private Process process ;
+
+    private final String ocspList ;
```
afelisatti: format

modules/http/src/test/java/org/mule/module/http/functional/requester/ocsp/AbstractHttpOcspRevocationTestCase.java
```diff
@@ -0,0 +1,105 @@
...
+
+    private int ocspPort;
+
+    private String ocspResponder ;
```
afelisatti: format

modules/http/src/test/java/org/mule/module/http/functional/requester/ocsp/AbstractHttpOcspRevocationTestCase.java
```diff
@@ -0,0 +1,105 @@
...
+import org.junit.Rule;
+
+public abstract class AbstractHttpOcspRevocationTestCase extends AbstractHttpTlsRevocationTestCase
+{
+
+    @Rule
+    public DynamicPort dynamicOcspPort ;
+
+    @Rule
+    public SystemProperty certAliasSystemProperty ;
```
afelisatti: format

modules/http/src/test/java/org/mule/module/http/functional/requester/ocsp/AbstractHttpOcspRevocationTestCase.java
```diff
@@ -0,0 +1,105 @@
...
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+
+public abstract class AbstractHttpOcspRevocationTestCase extends AbstractHttpTlsRevocationTestCase
+{
+
+    @Rule
+    public DynamicPort dynamicOcspPort ;
```
afelisatti: format

modules/http/src/test/java/org/mule/module/http/functional/requester/ocsp/AbstractHttpOcspRevocationTestCase.java
```diff
@@ -0,0 +1,105 @@
...
+            ocspResponder = DEFAULT_OCSP_RESPONDER;
+        }
+    }
+
+
+    @Before
+    public void setUp() throws Exception
+    {
+        // Running the ocsp server in a fixed port causes that sometimes the server fails at initialisation.
+        assumeFalse("Since openssl ocsp command has a flaky behavior the test will be ignored when the port number is not dynamic.", FIXED_OCSP_PORT == ocspPort);
```
afelisatti: Does this mean this won't run in parallel with other tests that might get 1111 assigned through `DynamicPort`?

afelisatti: Looks like since you're trying to validate that the dynamic port is the one used if 1111 is assigned the test won't work, however, the reason mentions openssl flakiness. Could you clarify this?

core/src/main/java/org/mule/api/security/tls/TlsConfiguration.java
```diff
@@ -376,15 +387,52 @@ private KeyStore createTrustStore() throws CreateException
...
+            if (keyStore.isCertificateEntry(alias))
+            {
+                trustAnchors.add(new TrustAnchor((X509Certificate) keyStore.getCertificate(alias), null));
+            }
+        }
+
+        return trustAnchors;
+    }
+
+    public static void assertNotNull(Object value, String message)
```
afelisatti: Shouldn't you use Mule's `Preconditions`?

transports/ssl/src/main/java/org/mule/transport/ssl/revocation/CrlFile.java
```diff
@@ -49,22 +50,14 @@ public void setPath(String path)
     }
 
     @Override
-    public ManagerFactoryParameters configFor(KeyStore trustStore)
+    public ManagerFactoryParameters configFor(KeyStore trustStore, Set<TrustAnchor> defaultTrustAnchors)
     {
+        assertNotNull(trustStore, "tls:crl-file requires the 'path' attribute");
```
afelisatti: Why 2 asserts regarding the same thing (trustStore) and different messages?

transports/ssl/src/main/java/org/mule/transport/ssl/revocation/CrlFile.java
```diff
@@ -49,22 +50,14 @@ public void setPath(String path)
...
-            while (aliases.hasMoreElements())
-            {
-                String alias = aliases.nextElement();
-                if (trustStore.isCertificateEntry(alias))
-                {
-                    trustAnchors.add(new TrustAnchor((X509Certificate) trustStore.getCertificate(alias), null));
-                }
-            }
-
+            Set<TrustAnchor> trustAnchors = getTrustAnchorsFromKeyStore(trustStore);
```
afelisatti: Why is this defined in `TlsConfiguration`? I've only seen it used here.

core/src/main/java/org/mule/api/security/tls/TlsConfiguration.java
```diff
@@ -311,38 +320,40 @@ protected void checkKeyStoreContainsAlias(KeyStore keyStore) throws KeyStoreExce
...
-                    ManagerFactoryParameters tmfParams = revocationCheck.configFor(trustStore);
-                    trustManagerFactory.init(tmfParams);
-                }
-                else
-                {
-                    trustManagerFactory.init(trustStore);
-                }
+        try
+        {
+            KeyStore trustStore = trustStoreName != null ? createTrustStore() : null;
```
afelisatti: How could the trustStore be null? Wasn't there a check above?

modules/http/src/test/java/org/mule/module/http/functional/requester/HttpStandardRevocationConfig.java
alepulver: I used it to test with google, but it would require accessing an external server. If that's acceptable I'll just assert that it fetches a query.

modules/http/src/test/java/org/mule/module/http/functional/requester/ocsp/AbstractHttpOcspRevocationTestCase.java
alepulver: The openssl command has an issue reopening the same port one closed. If you run the command, send an OCSP query, close the process (even wait a few seconds) and then start it again in the same port it fails.
It seems to be trying to reuse it...

alepulver: We can't change the port in the certificate because it's generated manually.

core/src/main/java/org/mule/api/security/tls/TlsConfiguration.java
alepulver: Yes. The check verifies that both trust store and revocation are null. If revocation is not null but trust store is, we must construct a custom trust manager for revocation with default java trust store.

transports/ssl/src/main/java/org/mule/transport/ssl/revocation/CrlFile.java
alepulver: It could be useful for future revocation code. But if you want I can move it there.

core/src/main/java/org/mule/api/security/tls/TlsConfiguration.java
```diff
@@ -311,38 +321,40 @@ protected void checkKeyStoreContainsAlias(KeyStore keyStore) throws KeyStoreExce
...
             {
-                throw new CreateException(
-                        CoreMessages.failedToLoad("Trust Manager (" + trustManagerAlgorithm + ")"), e, this);
+                trustManagerFactory.init(trustStore);
             }
         }
+        catch (Exception e)
+        {
+            throw new CreateException(
+                    CoreMessages.failedToLoad("Trust Manager (" + trustManagerAlgorithm + ")"), e, this);
```
afelisatti: static import

core/src/main/java/org/mule/api/security/tls/TlsConfiguration.java
```diff
@@ -311,38 +321,40 @@ protected void checkKeyStoreContainsAlias(KeyStore keyStore) throws KeyStoreExce
...
 
-            try
-            {
-                KeyStore trustStore = createTrustStore();
-                trustManagerFactory = TrustManagerFactory.getInstance(trustManagerAlgorithm);
+        // Revocation checking is only supported for PKIX algorithm
+        if (revocationEnabled && !REVOCATION_KEYSTORE_ALGORITHM.equalsIgnoreCase(trustManagerAlgorithm))
+        {
+            String errorText = formatInvalidCrlAlgorithm(getTrustManagerAlgorithm());
+            throw new CreateException(CoreMessages.createStaticMessage(errorText), this);
```
afelisatti: static import

transports/ssl/src/main/java/org/mule/transport/ssl/revocation/CrlFile.java
afelisatti: Yes, move it to where it's used

PR #5799, opened 2017-09-23
MULE-12716: Remove registries from API

elrodro83:
* 3rd pass

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/registry/RegistryProcessor.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/registry/RegistryProcessor.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.internal.registry;
+
+import static java.lang.String.format;
+import static org.reflections.ReflectionUtils.getAllFields;
+import static org.reflections.ReflectionUtils.getAllMethods;
+import static org.reflections.ReflectionUtils.withAnnotation;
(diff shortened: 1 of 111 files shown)


core/src/main/java/org/mule/runtime/core/internal/serialization/AbstractSerializationProtocol.java
```diff
@@ -19,12 +21,15 @@
...
 
+import javax.inject.Inject;
+
 /**
- * Base class for implementations of {@link SerializationProtocol} This class implements all the base behavioral
- * contract allowing its extensions to only care about the actual serialization/deserialization part.
+ * Base class for implementations of {@link SerializationProtocol} This class implements all the base behavioral contract allowing
+ * its extensions to only care about the actual serialization/deserialization part.
  */
 public abstract class AbstractSerializationProtocol implements SerializationProtocol, MuleContextAware {
```
pablokraan: Do we still need MuleContextAware class?

elrodro83: no, but the setter is still needed becuause of how the lifecycle of these objects is handled

core/src/main/java/org/mule/runtime/core/internal/registry/RegistryProcessor.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+
+import javax.inject.Inject;
+
+/**
+ * Injects the {@link Registry} object for objects stored in the {@link TransientRegistry}.
+ *
+ * @deprecated as of 3.7.0 since these are only used by {@link TransientRegistry} which is also deprecated. Use post processors
```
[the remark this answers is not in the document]
elrodro83: the support this class relies on was deprecated on 3.7.0, but is hasn't been removed yet, Referenced the issue for cleaning that up

PR #5821, opened 2017-09-26
MULE-12982: Cleanup DefaultTransformationService so it does not expose deprecated methods

lucianoRM:


lucianoRM:
Jenkins verify

PR #5833, opened 2017-09-27
MULE-13546: Support deployment time configuration in Mule 4.

fsgonz:


modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/descriptor/ArtifactDescriptorFactory.java
```diff
@@ -24,4 +26,14 @@
...
+
+  /**
+   * Creates an artifact descriptor from a folder.
+   *
+   * @param artifactFolder an existing folder containing artifact files
+   * @param deploymentProperties properties provided for the deployment process.
+   * @return a non null descriptor
+   * @throws ArtifactDescriptorCreateException if the factory is not able to create a descriptor from the folder.
+   */
+  T create(File artifactFolder, Optional<Properties> deploymentProperties) throws ArtifactDescriptorCreateException;
```
pablolagreca: Why do we have both create(File artifactFolder) and create(File artifactFolder, Optional<Properties> deploymentProperties) if deploymentProperties argument is an Optional?

PR #5843, opened 2017-09-28
MULE-9658: Empty timeZone in poll doesn't fallback to server time zone 

gabrielegreco16:


modules/schedulers/src/main/java/org/mule/modules/schedulers/cron/CronSchedulerFactory.java
```diff
@@ -47,7 +50,18 @@ protected Scheduler doCreate(String name, PollingReceiverWorker job)
 
     protected TimeZone resolveTimeZone(String name)
     {
-        TimeZone resolvedTimeZone = timeZone == null ? getDefault() : getTimeZone(timeZone);
+        TimeZone resolvedTimeZone = timeZone == null || timeZone.equals("") ? Calendar.getInstance().getTimeZone()
+                : getTimeZone(timeZone);
+        if(timeZone!=null) {
```
fsgonz: Formatter. Braces according to 3.x formatter should be in the next line.

modules/schedulers/src/main/java/org/mule/modules/schedulers/cron/CronSchedulerFactory.java
```diff
@@ -47,7 +50,18 @@ protected Scheduler doCreate(String name, PollingReceiverWorker job)
...
+            if ((!timeZone.equals(TZ_GMT_ID) && resolvedTimeZone.getID().equals(TZ_GMT_ID))) {
+                throw new SchedulerCreationException("Invalid Timezone");
+            }
+        }
```
fsgonz: Same as above. Apply formatter in the changes you make.

modules/schedulers/src/test/java/org/mule/modules/schedulers/cron/CronSchedulerFactoryTest.java
```diff
@@ -32,4 +32,6 @@ public void testSchedulerCreation()
         assertEquals("my expression", ((CronScheduler) scheduler).getCronExpression());
     }
 
+
```
fsgonz: Extra lines.

modules/schedulers/src/main/java/org/mule/modules/schedulers/cron/CronSchedulerFactory.java
fsgonz: Nested if. Merge this into one condition.

PR #5856, opened 2017-09-28
 MULE-8207: Fix ConcurrentModificationException in FileMessageReceiver.

facundovs:
When no files are found in FileMessageReceiver polling, an empty list is returned.
However this empty list is mutable (an instance of ArrayList).
We should return an immutable list instead for avoiding Concurrent Exceptions in heavy load.

### DIFF
--- a/transports/file/src/test/java/org/mule/transport/file/FilePollEmptyResultTestCase.java
+++ b/transports/file/src/test/java/org/mule/transport/file/FilePollEmptyResultTestCase.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.transport.file;
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.nullValue;
+import static org.junit.Assert.assertThat;
+import static org.mule.util.FileUtils.openDirectory;
+
+import org.mule.api.exception.RollbackSourceCallback;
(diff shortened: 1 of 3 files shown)


transports/file/src/test/java/org/mule/transport/file/FilePollEmptyResultTestCase.java
```diff
@@ -0,0 +1,72 @@
...
+import org.mule.util.concurrent.Latch;
+
+import org.junit.Before;
+import org.junit.Test;
+
+
+public class FilePollEmptyResultTestCase extends FunctionalTestCase
+{
+
+    private static long TEST_TIMEOUT = 5000;
```
pablokraan: final

transports/file/src/test/java/org/mule/transport/file/FilePollEmptyResultTestCase.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private static Throwable concurrentException;
+
+    private Latch latch = new Latch();
```
pablokraan: final

transports/file/src/test/java/org/mule/transport/file/FilePollEmptyResultTestCase.java
```diff
@@ -0,0 +1,72 @@
...
+    {
+        openDirectory(getFileInsideWorkingDirectory("test").getAbsolutePath());
+        openDirectory(getFileInsideWorkingDirectory("test2").getAbsolutePath());
+        muleContext.setExceptionListener(new TestExceptionHandler());
+    }
+
+    @Test
+    public void testNoConcurrentExceptionWhenEmptyPollResult() throws InterruptedException
+    {
+        assertThat(latch.await(TEST_TIMEOUT, MILLISECONDS), is(false));
```
pablokraan: Is the test broken when the fix is not there?

facundovs: Yes. The test breaks consistently without the changes.

PR #5868, opened 2017-09-30
MULE-13538: remove spring.schemas

alepulver:


modules/extensions-support/src/test/resources/heisenberg-test-manifest.xml
```diff
@@ -13,9 +13,6 @@
     <exportedResources>
```
pablolagreca: removed this file

PR #5875, opened 2017-10-02
MULE-10519: Move all exception classes away from MessagingException

elrodro83:


core/src/main/java/org/mule/runtime/core/privileged/exception/EventProcessingException.java
```diff
@@ -43,6 +44,13 @@ public CoreEvent getEvent() {
     return event;
   }
 
+  /**
+   * @return the Mule component that causes the failure if known, {@code null} otherwise.
+   */
+  public Component getFailingComponent() {
+    return null;
```
pablolagreca: Something missing here?

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/transaction/TransactionalOperationTestCase.java
```diff
@@ -55,11 +56,12 @@ public void executeTransactionless() throws Exception {
...
   @Test
   public void localTxDoesntSupportMultipleResources() throws Exception {
-    EventProcessingException exception = flowRunner("localTxDoesntSupportMultipleResources").runExpectingException();
-    Error error = exception.getEvent().getError().get();
-    assertThat(error.getDescription(),
-               is(containsString("the current transaction doesn't support it and could not be bound")));
-    assertThat(error.getCause(), is(instanceOf(TransactionException.class)));
+    flowRunner("localTxDoesntSupportMultipleResources").runExpectingException(allOf(instanceOf(TransactionException.class),
+                                                                                    hasMessage(containsString("the current transaction doesn't support it and could not be bound"))));
+    // Error error = exception.getEvent().getError().get();
```
pablolagreca: ????

core/src/main/java/org/mule/runtime/core/privileged/exception/EventProcessingException.java
elrodro83: no, this is overridden by MessagingException that does have this info. I needed this method accessible through a Privileged API to use it in `ExpectedError`

PR #5886, opened 2017-10-03
MULE-10519: Move all exception classes away from MessagingException

elrodro83:
* Remove plugin reactor hooks for error handling


core/src/main/java/org/mule/runtime/core/privileged/processor/chain/AbstractMessageProcessorChain.java
```diff
@@ -122,8 +122,11 @@ private ReactiveProcessor applyInterceptors(List<BiFunction<Processor, ReactiveP
 
     // #1 Update MessagingException with failing processor if required, create Error and set error context.
     interceptors.add((processor, next) -> stream -> from(stream)
-        .transform(next)
-        .onErrorMap(MessagingException.class, resolveMessagingException(processor)));
+        .concatMap(event -> just(event)
+            .transform(next)
+            .onErrorMap(MessagingException.class, resolveMessagingException(processor))
```
dfeist: These need to go the other way around don't they?

core/src/main/resources/META-INF/mule-module.properties
```diff
@@ -59,6 +59,7 @@ artifact.export.resources=/META-INF/org/mule/runtime/core/models,\
                           /META-INF/mime.types
 
 artifact.privileged.classPackages=org.mule.runtime.core.internal.message,\
+                                  org.mule.runtime.core.privileged,\
```
dfeist: Is this needed?

core/src/main/java/org/mule/runtime/core/privileged/processor/chain/AbstractMessageProcessorChain.java
```diff
@@ -179,21 +182,25 @@ public ProcessingType getProcessingType() {
...
+    // scheduling such as RejectedExecutionException's can be handled cleanly.
+    interceptors.add((processor, next) -> stream -> from(stream)
+        .concatMap(event -> just(event)
+            .transform(next)
+            .doOnEach(signal -> currentMuleContext.set(null))
+            .onErrorResume(RejectedExecutionException.class,
+                           throwable -> Mono.from(((BaseEventContext) event.getContext())
+                               .error(resolveException((Component) processor, event, throwable)))
+                               .then(Mono.empty()))
+            // If we already got a MessagingException, avoid wrapping it again
```
dfeist: Remove comment?

core/src/main/java/org/mule/runtime/core/privileged/processor/chain/AbstractMessageProcessorChain.java
```diff
@@ -179,21 +182,25 @@ public ProcessingType getProcessingType() {
...
+            .onErrorResume(MessagingException.class,
+                           throwable -> {
+                             throwable = resolveMessagingException(processor).apply(throwable);
+                             return Mono.from(((BaseEventContext) event.getContext()).error(throwable)).then(Mono.empty());
+                           })
+            .onErrorResume(throwable -> Mono
```
dfeist: Is this still needed? Having it here might help avoid hanging on unexpected exceptions thrown within chain.

core/src/main/resources/META-INF/mule-module.properties
elrodro83: yep, mutin needs that

core/src/main/java/org/mule/runtime/core/privileged/processor/chain/AbstractMessageProcessorChain.java
elrodro83: just a last check, just in case.

core/src/main/java/org/mule/runtime/core/privileged/processor/chain/AbstractMessageProcessorChain.java
elrodro83: why?

core/src/main/java/org/mule/runtime/core/privileged/processor/chain/AbstractMessageProcessorChain.java
elrodro83: i put then in the other order, won;t the messaging exception be built twice?

core/src/main/java/org/mule/runtime/core/privileged/processor/chain/AbstractMessageProcessorChain.java
dfeist: makes sense taking into account the new onErrorResume.

core/src/main/java/org/mule/runtime/core/privileged/processor/chain/AbstractMessageProcessorChain.java
dfeist: k

PR #5892, opened 2017-10-03
MULE-13648 - Deploying a bug fix version of an application is not causing correct redeploy

pablolagreca:


core/src/main/java/org/mule/runtime/core/api/config/DefaultMuleConfiguration.java
```diff
@@ -600,6 +604,21 @@ public void setDefaultObjectSerializer(ObjectSerializer defaultObjectSerializer)
     this.defaultObjectSerializer = defaultObjectSerializer;
   }
 
+  /**
+   * @return the name of the folder associated to this artifact to use for storing artifact related data
+   */
+  public String getDataFolderName() {
+    return dataFolderName != null ? dataFolderName : getId();
```
pablokraan: Is not dataFolder the same as workingDirectory?

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/descriptor/AbstractArtifactDescriptorFactory.java
```diff
@@ -124,6 +124,13 @@ private void validate(T descriptor) {
                                                          descriptor.getName(), descriptor.getRequiredProduct().name(),
                                                          runtimeProduct.name()));
     }
+    String bundleDescriptorVersion = descriptor.getBundleDescriptor().getVersion();
```
pablokraan: Can you extract these new lines into a "validateVersion" method?

modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/DefaultArchiveDeployer.java
```diff
@@ -152,7 +149,7 @@ private void undeployArtifactIfItsAPatch(T artifact) {
           }
           return false;
         }).findAny();
-    foundMatchingArtifact.ifPresent(deployer::undeploy);
+    foundMatchingArtifact.ifPresent(this::undeploy);
```
pablokraan: I think that using this method instead of the original will cause the artifact to be fully undeployed (causing the data folder to be deleted), but that is not the expected behavior if when the artifact has to be undeployed because there is a new artifact patch

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/rules/MuleDeployment.java
```diff
@@ -456,4 +470,11 @@ public String attachProperties() {
     }
     return null;
   }
+
+  /**
+   * @return the mule instance controller.
```
pablokraan: mule -> Mule

core/src/main/java/org/mule/runtime/core/api/config/DefaultMuleConfiguration.java
pablolagreca: No, it's just the name of the folder to use for temporary storage. The working directory is the whole file path (including the MULE_BASE/.mule )

modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/DefaultArchiveDeployer.java
pablolagreca: No, it doesn't, that happens on uninstall I think. I tested  exactly that in the system tests. I had to use this method because if not the notifications were not send.

pablokraan: That is incorrect. There are two methods to undeploy an artifact, undeploy and undeployArtifactWithoutUninstall. Undeploy method will uninstall the artifact on this line https://github.com/mulesoft/mule/blob/a29de774dff7a76800d99a6252863464642f6252/modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/DefaultArchiveDeployer.java#L336

PR #5893, opened 2017-10-03
MULE-13704: JMS Listener never begins transactions

estebanwasinger:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/DefaultSourceCallbackContext.java
```diff
@@ -145,4 +150,10 @@ public void addVariable(String variableName, Object value) {
   public <T, A> SourceCallback<T, A> getSourceCallback() {
     return sourceCallback;
   }
+
+  private I18nMessage createWrongConnectionMessage(Object connection) {
+    return createStaticMessage(format("Internal Error. The transacted source [%s] from the Extension [%s] tried to bind an " +
+        "connection of type [%s] which is not a current connection. ", sourceCallback.getOwningSourceName(),
```
marianogonzalez: i don't understand what this error means. the user probably won't either

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/DefaultSourceCallback.java
```diff
@@ -238,4 +238,20 @@ public TransactionConfig getTransactionConfig() {
   public SourceConnectionManager getSourceConnectionManager() {
     return messageSource.getSourceConnectionManager();
   }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public String getOwningSourceName() {
```
marianogonzalez: why don't you just have one method which returns the SourceModel?

estebanwasinger: just to not expose everything

PR #5896, opened 2017-10-04
MULE-13664: adding tests showing the datatype is correctly set after …

fernandezlautaro:
…calling operations


modules/extensions-xml-support/src/test/java/org/mule/test/functional/ModulesConsumingMimeTypeTestCase.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    try {
+      return org.mule.runtime.core.api.util.IOUtils.toString(inputStream);
+    } finally {
+      closeQuietly(inputStream);
+    }
+  }
+
+  //  private void compareXML(String expected, String actual) throws Exception {
```
alegmarra: remove

PR #5913, opened 2017-10-06
MULE-13729: Collection Aggregator does not honor the arrival order.

facundovs:
This issue occurs since arrival order is being saved as an invocation property which causes that the arrivalOrder is always overrode by the last value in all the messages to aggregate (because all they have the same reference to the flow var list).
A good solution is to replace the current logic, saving the arrivalOrder in an object that wraps the muleEvent.

### DIFF
--- a/core/src/main/java/org/mule/routing/EventGroup.java
+++ b/core/src/main/java/org/mule/routing/EventGroup.java
@@ -25,9 +25,12 @@ import org.mule.util.store.DeserializationPostInitialisable;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
+import java.util.TreeMap;
+import java.util.TreeSet;
 
 import org.apache.commons.collections.IteratorUtils;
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/mule/routing/EventGroup.java
```diff
@@ -229,14 +228,32 @@ public Object getGroupId()
                 return EMPTY_EVENTS_ARRAY;
             }
             List<Serializable> keys = eventsObjectStore.allKeys(eventsPartitionKey);
-            MuleEvent[] eventArray = new MuleEvent[keys.size()];
-            for (int i = 0; i < keys.size(); i++)
+
+            MuleEvent[] eventArray = new MuleEvent[keys.size()];;
```
elrodro83: remove extra semicolon

core/src/main/java/org/mule/routing/EventGroup.java
```diff
@@ -229,14 +228,32 @@ public Object getGroupId()
...
+
+            if (sortByArrival)
             {
-                eventArray[i] = eventsObjectStore.retrieve(keys.get(i), eventsPartitionKey);
+                WrapperOrderEvent[] wrapperEventArray = new WrapperOrderEvent[keys.size()];
```
elrodro83: why not use a TreeSet and then return its toArray()?

core/src/main/java/org/mule/routing/EventGroup.java
```diff
@@ -404,21 +420,40 @@ public MuleMessageCollection toMessageCollection() throws ObjectStoreException
...
+                for (int i = 0; i < keys.size(); i++)
+                {
+                    wrapperEventArray[i] = eventsObjectStore.retrieve(keys.get(i), eventsPartitionKey);
+                }
+
+                Arrays.sort(wrapperEventArray, new ArrivalOrderEventComparator());
+
+                for (int i = 0; i < wrapperEventArray.length; i++)
+                {
+                    messages.add(wrapperEventArray[i].event.getMessage());
```
elrodro83: why not use a TreeSet and then messages.addAll()?

core/src/main/java/org/mule/routing/EventGroup.java
```diff
@@ -404,21 +420,40 @@ public MuleMessageCollection toMessageCollection() throws ObjectStoreException
...
+            else
+            {
 
-        if (sortByArrival)
-        {
-            Collections.sort(messages, new ArrivalOrderMessageComparator());
+                for (Serializable id : keys)
+                {
+                    MuleMessage message = eventsObjectStore.retrieve(id, eventsPartitionKey).event.getMessage();
+                    messages.add(message);
```
elrodro83: inline `message`

core/src/main/java/org/mule/routing/EventGroup.java
```diff
@@ -478,7 +513,7 @@ protected MuleSession getMergedSession() throws ObjectStoreException
         {
             if (!key.equals(lastStoredEventKey))
             {
-                MuleEvent event = eventsObjectStore.retrieve(key, eventsPartitionKey);
+                MuleEvent event = eventsObjectStore.retrieve(key, eventsPartitionKey).event;
```
elrodro83: inline `event`

core/src/main/java/org/mule/util/store/PersistentObjectStorePartition.java
```diff
@@ -438,6 +439,10 @@ protected void serialize(File outputFile, StoreValue<T> storeValue) throws Objec
             {
                 DeserializationPostInitialisable.Implementation.init(storedValue.getValue(), muleContext);
             }
+            if (storedValue.getValue() instanceof WrapperOrderEvent)
```
elrodro83: this is smelly code. In any case, `WrapperOrderEvent` should implement `DeserializationPostInitialisable`

core/src/main/java/org/mule/routing/WrapperOrderEvent.java
```diff
@@ -0,0 +1,35 @@
...
+ * LICENSE.txt file.
+ */
+
+package org.mule.routing;
+
+import org.mule.api.MuleEvent;
+
+import java.io.Serializable;
+
+public class WrapperOrderEvent implements Serializable
```
elrodro83: make package visibility or move as an inner class of `EventGroup`

elrodro83: add javadocs

core/src/main/java/org/mule/routing/EventGroup.java
```diff
@@ -228,18 +230,78 @@ public Object getGroupId()
...
-            return eventArray;
+
+            return convertWrapperOrderEventCollectionToMuleEventArray(wrapperOrderEvents);
+        }
+    }
+
+    private Collection<WrapperOrderEvent> getWrapperEventsFromStore() throws ObjectStoreException
+    {
+        Collection<WrapperOrderEvent> wrapperOrderEvents = new ArrayList<>();
+        List<Serializable> keys = eventsObjectStore.allKeys(eventsPartitionKey);
```
elrodro83: i still don't understand why you need this temporary List

core/src/main/java/org/mule/routing/EventGroup.java
```diff
@@ -228,18 +230,78 @@ public Object getGroupId()
...
+
+    /**
+     * Orders a collection of wrapperOrderEvents using the {@link ArrivalOrderEventComparator}
+     *
+     * @param wrapperOrderEvents the collection of wrapperOrderEvents to order.
+     * @return a ordered collection of wrapperOrderEvents
+     */
+    private Collection <WrapperOrderEvent> orderWrapperOrderEvents (Collection<WrapperOrderEvent> wrapperOrderEvents)
+    {
+        TreeSet<WrapperOrderEvent> wrapperOrderEventsSet = new TreeSet<>(new ArrivalOrderEventComparator());
```
elrodro83: declare as Set or SortedSet

PR #5921, opened 2017-10-09
MULE-12159: add loaded extensions to the error message when an XSD is…

alepulver:
… not found

### DIFF
--- a/modules/spring-config/src/main/java/org/mule/runtime/config/internal/ModuleDelegatingEntityResolver.java
+++ b/modules/spring-config/src/main/java/org/mule/runtime/config/internal/ModuleDelegatingEntityResolver.java
@@ -26,7 +26,6 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


modules/spring-config/src/main/java/org/mule/runtime/config/internal/ModuleDelegatingEntityResolver.java
```diff
@@ -107,7 +108,12 @@ public InputSource resolveEntity(String publicId, String systemId) throws SAXExc
     }
     if (inputSource == null) {
       if (checkedEntities.get(systemId) != null) {
-        throw new MuleRuntimeException(createStaticMessage("Can't resolve %s %s", publicId == null ? "" : publicId, systemId));
+        String loadedExtensions = String.join(", ", extensions.stream().map(e -> e.getName()).collect(Collectors.toList()));
+        String namespaceNotFound =
+            publicId == null ? format("Can't resolve %s", systemId) : format("Can't resolve %s (%s)", publicId, systemId);
+        String message = format("%s, An extension might be missing from the dependencies (currently loaded: %s)",
```
marianogonzalez: an extension -> `a dependency` or `a plugin`.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/ModuleDelegatingEntityResolver.java
```diff
@@ -107,7 +108,12 @@ public InputSource resolveEntity(String publicId, String systemId) throws SAXExc
     }
     if (inputSource == null) {
       if (checkedEntities.get(systemId) != null) {
-        throw new MuleRuntimeException(createStaticMessage("Can't resolve %s %s", publicId == null ? "" : publicId, systemId));
+        String loadedExtensions = String.join(", ", extensions.stream().map(e -> e.getName()).collect(Collectors.toList()));
```
marianogonzalez: this is not enough. Things like ApiKit or other plugins will not be registered here

PR #5927, opened 2017-10-10
MULE-13746: Race condition when verifying group expiration just before resequencer finishes.

fsgonz:


core/src/main/java/org/mule/routing/correlation/ResequenceMessagesCorrelatorCallback.java
```diff
@@ -56,13 +57,20 @@ public MuleEvent aggregateEvents(EventGroup events) throws AggregationException
         MuleEvent[] results;
         try
         {
-            results = (events == null) ? new MuleEvent[0] : events.toArray(false);
+            results = (events == null || events.toArray(false).length == 0) ? null : events.toArray(false);
```
elrodro83: use ´events.size()´

elrodro83: move the early return here

PR #5930, opened 2017-10-10
Mule-13751 Issue with #[function.datestamp('yyyyMMdd')] extra ) added

gabrielegreco16:


fsgonz:
This format shouldn't be supported. function.datestamp.<format> must be used.

PR #5931, opened 2017-10-10
MULE-13752: Re enable validation:all scope

marianogonzalez:


core/src/main/java/org/mule/runtime/core/privileged/processor/chain/HasMessageProcessors.java
```diff
@@ -0,0 +1,16 @@
...
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.privileged.processor.chain;
+
+import org.mule.runtime.core.api.processor.Processor;
+
+import java.util.List;
+
+public interface HasMessageProcessors {
```
alegmarra: jdoc since

PR #5950, opened 2017-10-11
MULE-13711: JMS Publisher always joins the current TX, if one exists,…

estebanwasinger:
… regardless the chosen transactional Action

### DIFF
--- a/modules/extensions-spring-support/src/test/resources/models/tx-ext.json
+++ b/modules/extensions-spring-support/src/test/resources/models/tx-ext.json
@@ -131,15 +131,15 @@
               "name": "expirationPolicy",
               "description": "Configures the minimum amount of time that a dynamic configuration instance can remain idle before the runtime considers it eligible for expiration. This does not mean that the platform will expire the instance at the exact moment that it becomes eligible. The runtime will actually purge the instances when it sees it fit.",
               "modelProperties": {
-                "org.mule.runtime.extension.api.property.InfrastructureParameterModelProperty": {
-                  "sequence": 4
-                },
                 "org.mule.runtime.extension.api.property.QNameModelProperty": {
                   "value": {
                     "namespaceURI": "http://www.mulesoft.org/schema/mule/core",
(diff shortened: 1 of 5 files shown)


modules/extensions-spring-support/src/test/resources/models/tx-ext.json
```diff
@@ -427,15 +427,15 @@
...
-                      "sequence": 3
-                    },
                     "org.mule.runtime.extension.api.property.QNameModelProperty": {
                       "value": {
                         "namespaceURI": "http://www.mulesoft.org/schema/mule/core",
                         "localPart": "abstract-reconnection-strategy",
                         "prefix": "mule"
                       }
+                    },
+                    "org.mule.runtime.extension.api.property.InfrastructureParameterModelProperty": {
```
marianogonzalez: is this correct or just pablon's bug?

modules/extensions-spring-support/src/test/resources/transactional-parameters-injection-config.xml
```diff
@@ -24,7 +24,7 @@
     </flow>
 
     <flow name="injectInOperationJoinNotSupported">
-        <tx:inject-transactional-action config-ref="localTx1" action="NOT_SUPPORTED"/>
+        <tx:inject-transactional-action config-ref="localTx1" transactionalAction="NOT_SUPPORTED"/>
```
marianogonzalez: why does this changes name? Wasn't transactionalAction it's actual name?

estebanwasinger: the name must be always `transactionalAction` the bug I fixed now fixes the name to be always `transactionalAction` doesn't matter how is defined in the Java Code

modules/extensions-spring-support/src/test/resources/models/tx-ext.json
estebanwasinger: pablon's bug

PR #5951, opened 2017-10-11
MULE-13538: move back reconnect from deprecated to core

alepulver:


marianogonzalez:
mule-core-deprecated also contains a `flowType`. What about that?

PR #5955, opened 2017-10-11
MULE-13647: Use Deployment Properties to disable logging on tooling applications

mbuchwald:


core/src/main/java/org/mule/runtime/core/api/config/MuleProperties.java
```diff
@@ -201,4 +201,9 @@
   public static final String MULE_LOG_VERBOSE_CLASSLOADING = SYSTEM_PROPERTY_PREFIX + "classloading.verbose";
   public static final String MULE_MEL_AS_DEFAULT = SYSTEM_PROPERTY_PREFIX + "test.mel.default";
   public static final String MULE_DISABLE_RESPONSE_TIMEOUT = SYSTEM_PROPERTY_PREFIX + "timeout.disable";
+
+  /**
+   * Deployment properties used
+   */
+  public static final String MULE_MUTE_APP_LOGS_DEPLOYMENT_PROPERTY = "_mutedApp";
```
gsfernandes: perhaps I would suggest to use a different name: SYSTEM_PROPERTY_PREFIX + "application.deployment.muteLog"

modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/LoggerContextConfigurer.java
```diff
@@ -82,8 +92,11 @@ protected void configure(MuleLoggerContext context) {
   }
 
   protected void update(MuleLoggerContext context) {
-    boolean forceConsoleLog = System.getProperty(MULE_FORCE_CONSOLE_LOG) != null;
+    if (mute(context)) {
```
gsfernandes: Change name to something more meaningful, `shouldConfigureContext`?

modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/LoggerContextConfigurer.java
```diff
@@ -99,6 +112,19 @@ protected void update(MuleLoggerContext context) {
...
+    if (!context.isApplicationClassloader()) {
+      return false;
+    }
+
+    ArtifactDescriptor descriptor = context.getArtifactDescriptor();
+    if (descriptor == null || !descriptor.getDeploymentProperties().isPresent()) {
+      return false;
+    }
+    Properties properties = descriptor.getDeploymentProperties().get();
+    return Boolean.parseBoolean(properties.getProperty(MULE_MUTE_APP_LOGS_DEPLOYMENT_PROPERTY, "false"));
```
gsfernandes: static import

modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/MuleLoggerContext.java
```diff
@@ -132,7 +139,17 @@ protected boolean isApplicationClassloader() {
...
+    return artifactDescriptor;
+  }
+
   protected String getArtifactName() {
     return artifactName;
   }
+
+  @Override
+  public void close() {
+    this.artifactDescriptor = null;
```
gsfernandes: First call super.close() then do this close logic for the class.

gsfernandes: Put a comment // Clean up reference to avoid class loader leaks

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/api/ToolingService.java
```diff
@@ -35,6 +35,9 @@
    * Only requested components will be executed. All sources for flows will be stop unless
    * they are requested to be started by the client.
    *
+   * The application will be deployed using {@link org.mule.runtime.core.api.config.MuleProperties#MULE_MUTE_APP_LOGS_DEPLOYMENT_PROPERTY}
+   * so
```
gsfernandes: complete javadoc, "so..."

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/DefaultToolingService.java
```diff
@@ -44,11 +48,16 @@ public ConnectivityTestingServiceBuilder newConnectivityTestingServiceBuilder()
...
     application.install();
     application.lazyInit();
     application.start();
     return application;
   }
 
+  private Properties createApplicationProperties() {
+    Properties properties = new Properties();
+    properties.setProperty(MULE_MUTE_APP_LOGS_DEPLOYMENT_PROPERTY, "true");
+    return properties;
```
gsfernandes: Confirm with Fabian that properties are not going to be modified so we can create this only once and reuse the same instance for every createApplication invocation.

mbuchwald: They are not meant to be modified right now, but could be on the future, so I wouldn't do this

PR #5959, opened 2017-10-12
MULE-13779: Avoid lookup in registry per flow execution.

afelisatti:


core/src/main/java/org/mule/runtime/core/internal/execution/ModuleFlowProcessingPhase.java
```diff
@@ -105,6 +105,7 @@ public ModuleFlowProcessingPhase(PolicyManager policyManager) {
   @Override
   public void initialise() throws InitialisationException {
     final ErrorTypeRepository errorTypeRepository = muleContext.getErrorTypeRepository();
+    componentLocator = muleContext.getConfigurationComponentLocator();
```
dfeist: Have you looked for all other uses of `getConfigurationComponentLocator` to ensure they are not done as part of message processing?

core/src/main/java/org/mule/runtime/core/internal/execution/ModuleFlowProcessingPhase.java
```diff
@@ -124,7 +125,7 @@ public void runPhase(final ModuleFlowProcessingPhaseTemplate template, final Mes
                        final PhaseResultNotifier phaseResultNotifier) {
     try {
       final MessageSource messageSource = messageProcessContext.getMessageSource();
-      final FlowConstruct flowConstruct = (FlowConstruct) muleContext.getConfigurationComponentLocator().find(Location.builder()
+      final FlowConstruct flowConstruct = (FlowConstruct) componentLocator.find(Location.builder()
```
dfeist: The fact that `muleContext.getConfigurationComponentLocator()` foes not cache the value internally is inconsistent with other muleContext getters such as `getExpressionManager` ``getLockFactory` etc.

PR #5961, opened 2017-10-13
MULE-13739: Wrong HTTP and WSC stereotypes

marianogonzalez:


modules/extensions-spring-support/src/test/resources/models/test-oauth.json
```diff
@@ -21,7 +21,7 @@
     {
       "stereotype": {
         "type": "CONFIG",
-        "namespace": "TEST-OAUTH",
+        "namespace": "TEST--OAUTH",
```
alegmarra: 🤔

PR #5964, opened 2017-10-13
MULE 13721:MULE_REPLYTO_STOP not working when JMS is acting as an outbound connector

gabrielegreco16:


gabrielegreco16:
Waiting for Tests verification.

PR #5973, opened 2017-10-17
MULE-13705: Db Config not found through Lazy Initialization when flow-ref to subflow present

mbuchwald:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/MinimalApplicationModelGenerator.java
```diff
@@ -84,7 +88,20 @@ private void enableComponentDependencies(ComponentModel requestedComponentModel,
...
+    // While there are new dependencies resolved, calculate their dependencies
+    // This fixes bugs related to not resolving dependencies of dependencies, such as a config for a config
+    // e.g. tlsContext for http request, or a flow-ref inside a flow that is being referenced in another flow.
+    while (difference.size() > 0) {
+      Set<String> newDependencies = dependencyResolver.findComponentModelsDependencies(difference);
+      newDependencies.removeAll(allRequiredComponentModels);
+      allRequiredComponentModels.addAll(newDependencies);
+      difference = newDependencies;
+    }
+
```
gsfernandes: Maybe we can move this to a new private method

PR #5981, opened 2017-10-17
Mule 13782

alepulver:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/ModuleDelegatingEntityResolver.java
```diff
@@ -118,6 +117,29 @@ public InputSource resolveEntity(String publicId, String systemId) throws SAXExc
...
   }
 
+  private String overrideSystemIdForCompatibility(String publicId, String systemId) throws SAXException, IOException {
+    if (systemId.equals(CORE_XSD)) {
+      Boolean useDeprecated = muleEntityResolver.resolveEntity(publicId, CORE_DEPRECATED_XSD) != null;
+      Boolean usingCompatibility = muleEntityResolver.resolveEntity(publicId, COMPATIBILITY_XSD) != null;
+      Boolean runningTests = false;
+
+      try {
+        Class.forName("org.mule.tck.junit4.AbstractMuleTestCase", true, currentThread().getContextClassLoader());
```
pablokraan: What happens is the application contains the mule-unit-test jar?

alepulver: In that case it would work.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/ModuleDelegatingEntityResolver.java
```diff
@@ -118,6 +117,32 @@ public InputSource resolveEntity(String publicId, String systemId) throws SAXExc
...
+        return CORE_CURRENT_XSD;
+      }
+    }
+
+    return systemId;
+  }
+
+  private Boolean isRunningTests(StackTraceElement[] stackTrace) {
+    for (StackTraceElement element : stackTrace) {
+      System.out.println(element.getClassName());
```
pablokraan: Remove this

PR #5996, opened 2017-10-18
MULE-13666: lazy init does not handle spring:config

alepulver:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/ModuleDelegatingEntityResolver.java
```diff
@@ -118,6 +111,34 @@ public InputSource resolveEntity(String publicId, String systemId) throws SAXExc
     return inputSource;
   }
 
+  private String overrideSystemIdForCompatibility(String publicId, String systemId) throws SAXException, IOException {
```
gsfernandes: Are this changes related to this task?

PR #6009, opened 2017-10-19
MULE-13769: Transactions are rolledback and commited twice

estebanwasinger:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/transaction/DefaultTransactionHandle.java
```diff
@@ -6,22 +6,47 @@
...
+    return true;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public void commit() throws TransactionException {
+    Transaction transaction = TransactionCoordination.getInstance().getTransaction();
+    if (transaction != null) {
```
alegmarra: optional if present

estebanwasinger: is unnecessary and also can't be done because `commit()` and `rollback()` throws checked exceptions

PR #6015, opened 2017-10-19
MULE-13721: MULE_REPLYTO_STOP not working when JMS is acting as an outbound connector

gabrielegreco16:
MULE 13721:MULE_REPLYTO_STOP not working when JMS is acting as an outbound connector


fsgonz:
Waiting to see where the jms outbound-endpoint destination is ignored in case of replyTo set.

PR #6018, opened 2017-10-19
MULE-13804: extension-model.json still ship in extension binary

juandesi:


modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/resources/ExtensionModelResourceFactoryTestCase.java
```diff
@@ -6,24 +6,29 @@
...
+import org.mule.runtime.module.extension.internal.resources.documentation.ExtensionDocumentationResourceGenerator;
 import org.mule.tck.size.SmallTest;
 import org.mule.test.heisenberg.extension.HeisenbergExtension;
 
+import org.custommonkey.xmlunit.XMLUnit;
 import org.junit.Before;
 import org.junit.Test;
 
 @SmallTest
 public class ExtensionModelResourceFactoryTestCase extends AbstractGeneratedResourceFactoryTestCase {
```
marianogonzalez: class name should change too

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/resources/ExtensionModelResourceFactoryTestCase.java
```diff
@@ -33,15 +38,15 @@ public void before() {
...
 
   @Test
   public void generate() throws Exception {
     GeneratedResource resource = resourceFactory.generateResource(extensionModel).get();
-    assertThat(resource.getPath(), equalTo(EXTENSION_MODEL_JSON_FILE_NAME));
-
-    ExtensionModel deserialized = new ExtensionModelJsonSerializer().deserialize(new String(resource.getContent()));
-    assertThat(extensionModel.getName(), equalTo(deserialized.getName()));
+    assertThat(resource.getPath(), equalTo(RESOURCE_NAME));
+    XMLUnit.setIgnoreWhitespace(true);
```
marianogonzalez: there was no test for this before?

juandesi: nope :D, there are for the generation but not for the resource generator using spi

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/resources/ExtensionModelResourceFactoryTestCase.java
juandesi: I don't agree, all extension resource factories should be tested here, we currently have only one, but this could change

PR #6020, opened 2017-10-19
MULE-13779: Avoid call to deprecated method that causes NoSuchMethodE…

afelisatti:
…xceptions


afelisatti:
Currently running tests but this can be reviewed.

PR #6025, opened 2017-10-19
MULE-13782: disable mule-test.xsd in real applications

alepulver:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/ModuleDelegatingEntityResolver.java
```diff
@@ -128,6 +129,12 @@ private String overrideSystemIdForCompatibility(String publicId, String systemId
       } else {
         return CORE_CURRENT_XSD;
       }
+    } else if (systemId.equals(TEST_XSD)) {
+      Boolean runningTests = isRunningTests(new Throwable().getStackTrace());
+      if (!runningTests) {
+        String message = "mule-test.xsd can't be used in real applications";
+        throw new MuleRuntimeException(createStaticMessage(message));
```
fernandezlautaro: I'm seeing an issue when developing a plugin (thru SDK or smart connector) which name is `test`.
Unless it's forbidden to call your extension `test`.

PR #6033, opened 2017-10-20
MULE-13833: Change default sslType to use the JVM's one.

afelisatti:


afelisatti:
Under most VM's this will cover all three current TLS versions and ciphers for those (other implementations can be customised to match that behaviour but will cover TLSv1 as we currently do by default).

PR #6036, opened 2017-10-20
MULE-13814: Remove org.apache.logging.log4j:log4j-slf4j-impl from

elrodro83:
classpath


tests/runner/src/main/java/org/mule/test/runner/api/AetherClassPathClassifier.java
```diff
@@ -121,6 +123,14 @@ public AetherClassPathClassifier(DependencyResolver dependencyResolver,
 
     this.dependencyResolver = dependencyResolver;
     this.artifactClassificationTypeResolver = artifactClassificationTypeResolver;
+
+    try {
```
pablokraan: Extract to an utility method

PR #6055, opened 2017-10-24
MULE-13863: Notifications fail in test:processor when flow executions

elrodro83:
wait


core/src/main/java/org/mule/runtime/core/api/context/notification/ServerNotificationManager.java
```diff
@@ -195,7 +197,7 @@ public void fireNotification(Notification notification) {
         });
       }
     } finally {
-      disposeLock.readLock().unlock();
+      activeFires.decrementAndGet();
```
alepulver: Wouldn't a `if (x == 0 && disposed.get()) { waitFinish.unlock(); }` make the `sleep(20)` unnecessary later? So it could wait on the lock.

PR #6060, opened 2017-10-25
MULE-11242: Lifecycle not applied to prototype objects when lookup by type is used

lucianoRM:


tests/test-policies/src/main/java/org/mule/functional/policy/api/TestPolicyBuildingDefinitionProvider.java
```diff
@@ -45,6 +45,7 @@ public void init() {
                                        fromChildConfiguration(PolicyChain.class).withWrapperIdentifier("source").build())
         .withSetterParameterDefinition("operationPolicyChain",
                                        fromChildConfiguration(PolicyChain.class).withWrapperIdentifier("operation").build())
+        .asPrototype()
```
pablolagreca: Remove this. It doesn't make sense to make ti prototype.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/SpringRegistry.java
```diff
@@ -184,16 +184,18 @@ public Object lookupObject(String key, boolean applyLifecycle) {
         }
         return null;
       }
+      applyLifecycleIfPrototype(object, key, applyLifecycle);
```
pablolagreca: Can you validate that a prototype object created with the SpringConfig does not gets applied lifecycle?. We don't support Initializable, Startable, Stoppable or Disposable in spring module.

PR #6080, opened 2017-10-26
Mule 13868

alepulver:


modules/dsl-api/src/main/java/org/mule/runtime/dsl/api/component/ComponentBuildingDefinition.java
```diff
@@ -58,6 +59,7 @@
   private Optional<TypeConverter> typeConverter = empty();
   private Optional<TypeConverter> keyTypeConverter = empty();
   private boolean alwaysEnabled = false;
+  private String gloablName;
```
pablolagreca: I would call it registration name instead. And I think that we should comment in the javadoc that is expected to allow a single element created from this definition in the whole artifact.

PR #6082, opened 2017-10-26
MULE-13878: Lazy init is not initializing batch properly

mbuchwald:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/LazyMuleArtifactContext.java
```diff
@@ -36,15 +36,13 @@
...
 import org.mule.runtime.core.api.config.bootstrap.ArtifactType;
 import org.mule.runtime.core.internal.connectivity.DefaultConnectivityTestingService;
 import org.mule.runtime.core.internal.metadata.MuleMetadataService;
 import org.mule.runtime.core.internal.value.MuleValueProviderService;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
marianogonzalez: not

PR #6087, opened 2017-10-27
Vm

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/OperationModelLoaderDelegate.java
```diff
@@ -144,6 +150,10 @@ private void declareOperation(OperationDeclarer operation, boolean supportsConfi
     addExecutionType(operation, operationMethod);
     ParameterDeclarationContext declarationContext = new ParameterDeclarationContext(OPERATION, operation.getDeclaration());
     loader.getMethodParametersLoader().declare(operation, operationMethod.getParameters(), declarationContext);
+    loader.getFieldParametersLoader().declare(operation, fieldParameters, declarationContext).forEach(p -> {
```
estebanwasinger: this information should not be provided by `OperationContainerWrapper`/`OperationContainerElement` abstractions?

marianogonzalez: this is exactly how it's done for sources. The field parameter loader has custom contributors to handle that. 

If sources do it like this, I guess this should too

PR #6091, opened 2017-10-30
MULE-13856: Bring back behaviour exclusively for compatibility plugin

estebanwasinger:


core/src/main/java/org/mule/runtime/core/internal/execution/ResolvePreviousTransactionInterceptor.java
```diff
@@ -0,0 +1,47 @@
...
+import org.mule.runtime.core.api.execution.ExecutionCallback;
+import org.mule.runtime.core.api.transaction.Transaction;
+import org.mule.runtime.core.api.transaction.TransactionConfig;
+import org.mule.runtime.api.tx.TransactionException;
+import org.mule.runtime.core.api.transaction.TransactionCoordination;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ResolvePreviousTransactionInterceptor<T> implements ExecutionInterceptor<T> {
```
marianogonzalez: add some kind of comment saying that this is for compatibility only. Moving to a compatibility package would also rock

PR #6100, opened 2017-10-31
MULE-13918: Add an option for sending policy's logs to app's logs

balbifm:


modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/MuleLoggerContext.java
```diff
@@ -56,22 +63,44 @@
...
-    ownerClassLoaderHash =
+    this.ownerClassLoaderHash =
         ownerClassLoader != null ? ownerClassLoader.hashCode() : getClass().getClassLoader().getSystemClassLoader().hashCode();
 
-    if (ownerClassLoader instanceof ArtifactClassLoader) {
-      artifactClassloader = true;
-      artifactName = getArtifactName((ArtifactClassLoader) ownerClassLoader);
-      artifactDescriptor = getArtifactDescriptor((ArtifactClassLoader) ownerClassLoader);
-      applicationClassloader = ownerClassLoader instanceof RegionClassLoader;
+    Optional<ArtifactClassLoader> artifactClassloader = getArtifactClassLoader(ownerClassLoader);
```
marianogonzalez: how exactly does this achieve the goal? As far as I can see, this only properly obtains artifact metadata, but the actual logging separation is done in the `DispatchingLogger` class through the `ownerClassLoaderHash` field , which is not changed here.

Also, I see you added tests about the artifact metadata being properly obtained, and that's fine. But I see no test verifying that the policy log is actually going where you want.

balbifm: We achieve this by changing the artifactName that the context uses.
Let me check how can I add those tests.

modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/ArtifactAwareContextSelector.java
```diff
@@ -100,22 +83,60 @@ static ClassLoader resolveLoggerContextClassLoader(ClassLoader classLoader) {
       loggerClassLoader = loggerClassLoader.getParent();
     }
 
-    if (!(loggerClassLoader instanceof ArtifactClassLoader)) {
+    if (loggerClassLoader == null) {
       return SYSTEM_CLASSLOADER;
     } else if (isRegionClassLoaderMember(loggerClassLoader)) {
       loggerClassLoader = loggerClassLoader.getParent();
+
+      if (isPolicyClassLoader(loggerClassLoader) && policyLogsToAppEnabled(loggerClassLoader)) {
```
marianogonzalez: this is expensive to evaluate each time. if this could be moved to the cache it'd be great

modules/launcher/src/test/java/org/mule/runtime/module/launcher/log4j2/ArtifactAwareContextSelectorTestCase.java
```diff
@@ -192,6 +213,17 @@ private void assertConfigurationLocation(LoggerContext ctx) {
     assertThat(ctx.getConfigLocation(), equalTo(CONFIG_LOCATION.toURI()));
   }
 
+  private RegionClassLoader getPolicyRegionClassLoader(PolicyTemplateDescriptor policyTemplateDescriptor) {
+    return new RegionClassLoader(POLICY_TEMPLATE_ARTIFACT_ID, policyTemplateDescriptor, regionClassLoader,
```
marianogonzalez: use a mock. Don't create a full classloader if you don't have to

modules/launcher/src/test/java/org/mule/runtime/module/launcher/log4j2/ArtifactAwareContextSelectorTestCase.java
```diff
@@ -164,6 +171,20 @@ public void returnsMuleLoggerContextForInternalArtifactClassLoader() {
     assertThat(serviceCtx, sameInstance(systemContext));
   }
 
+  @Test
+  public void returnsParentContextForPolicyClassloader() throws MalformedURLException {
+    ClassLoader childClassLoader = new URLClassLoader(new URL[0], regionClassLoader);
```
marianogonzalez: use a mock. Don't create a full classloader if you don't have to

modules/launcher/src/test/java/org/mule/runtime/module/launcher/log4j2/ArtifactAwareContextSelectorTestCase.java
```diff
@@ -164,6 +171,20 @@ public void returnsMuleLoggerContextForInternalArtifactClassLoader() {
...
+    PolicyTemplateDescriptor policyTemplateDescriptor = new PolicyTemplateDescriptor(POLICY_TEMPLATE_NAME, empty());
+    RegionClassLoader policyRegionClassLoader = spy(getPolicyRegionClassLoader(policyTemplateDescriptor));
+    MuleArtifactClassLoader policyClassLoader = getPolicyArtifactClassLoader(policyTemplateDescriptor, policyRegionClassLoader);
+
+    LoggerContext appCtx = selector.getContext("", childClassLoader, true);
+    LoggerContext policyCtx = selector.getContext("", policyClassLoader, true);
+
+    assertThat(policyCtx, instanceOf(MuleLoggerContext.class));
```
marianogonzalez: this is testing implementation, not behavior. the only assertion with value is the one below

PR #6101, opened 2017-10-31
MULE-13237: Validate that tlsContext is present either as a ref or a child

juandesi:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/exception/RequiredParameterNotSetException.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package org.mule.runtime.module.extension.internal.runtime.exception;
+
+import static java.lang.String.format;
+import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
+import static org.mule.runtime.api.util.NameUtils.hyphenize;
+import org.mule.runtime.api.exception.MuleRuntimeException;
+import org.mule.runtime.api.meta.model.parameter.ParameterModel;
+
+public class RequiredParameterNotSetException extends MuleRuntimeException {
```
marianogonzalez: jdoc

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ParametersResolver.java
```diff
@@ -66,19 +71,27 @@
...
+  private ConfigurationProperties properties;
+
+  private final Boolean lazyInitEnabled;
   private final MuleContext muleContext;
   private final Map<String, ?> parameters;
 
   private ParametersResolver(MuleContext muleContext, Map<String, ?> parameters) {
     this.muleContext = muleContext;
     this.parameters = parameters;
+    inject();
```
marianogonzalez: to inject in self on each new instance is going to hurt application start time considerably. I would aim to resolve Lazyness mode at `AbstractObjectFactory` and just pass the boolean to this constructor

PR #6103, opened 2017-10-31
MULE-13632: Exception thrown when doing a hot redeploy of an app using the derby driver

estebanwasinger:


modules/artifact/src/main/java/org/mule/module/artifact/classloader/DefaultResourceReleaser.java
```diff
@@ -176,7 +177,16 @@ private void leakPreventionForDerbyEmbeddedDriver(Object driverObject) {
         m.invoke(driverObject, "jdbc:derby:;shutdown=true", null);
       }
     } catch (Throwable e) {
-      logger.warn("Unable to unregister Derby's embedded driver", e);
+      Throwable cause = e.getCause();
+      if(cause instanceof SQLException){
```
marianogonzalez: format

modules/artifact/src/main/java/org/mule/module/artifact/classloader/DefaultResourceReleaser.java
```diff
@@ -176,7 +177,16 @@ private void leakPreventionForDerbyEmbeddedDriver(Object driverObject) {
...
+        // A successful shutdown always results in an SQLException to indicate that Derby has shut down and that
+        // there is no other exception.
+        if(logger.isDebugEnabled()) {
```
marianogonzalez: no test case?

PR #6108, opened 2017-11-01
MULE-13920: @Stereotypes annotation is not being exported

estebanwasinger:


modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/scopes/RoutersExecutionTestCase.java
```diff
@@ -141,6 +143,12 @@ public void twoRoutesRouterNone() throws Exception {
   }
 
   @Test
+  public void stereotypedRoutes() throws Exception {
```
marianogonzalez: this tests the happy path. There should be a test which verifies that the thing fails if you put an MP which doesn't fit the stereotype

estebanwasinger: that is why I created this test `ScopeStereotypeValidationTestCase`

PR #6115, opened 2017-11-02
MULE-13891: Non repeatable streams are not managed to guarantee closi…

marianogonzalez:
…ng them

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/streaming/DefaultStreamingManager.java
+++ b/core/src/main/java/org/mule/runtime/core/api/streaming/DefaultStreamingManager.java
@@ -8,13 +8,15 @@ package org.mule.runtime.core.api.streaming;
 
 import static org.mule.runtime.core.api.lifecycle.LifecycleUtils.disposeIfNeeded;
 import static org.mule.runtime.core.api.lifecycle.LifecycleUtils.initialiseIfNeeded;
+import static org.mule.runtime.core.api.util.IOUtils.closeQuietly;
 import static org.slf4j.LoggerFactory.getLogger;
-
+import static reactor.core.publisher.Mono.from;
 import org.mule.runtime.api.lifecycle.Disposable;
 import org.mule.runtime.api.lifecycle.Initialisable;
 import org.mule.runtime.api.lifecycle.InitialisationException;
 import org.mule.runtime.api.scheduler.Scheduler;
 import org.mule.runtime.api.scheduler.SchedulerService;
+import org.mule.runtime.api.streaming.Cursor;
 import org.mule.runtime.api.streaming.CursorProvider;
 import org.mule.runtime.core.api.MuleContext;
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/mule/runtime/core/api/streaming/DefaultStreamingManager.java
```diff
@@ -124,6 +129,19 @@ public CursorProvider manage(CursorProvider provider, CoreEvent creatorEvent) {
    * {@inheritDoc}
    */
   @Override
+  public void manage(InputStream stream, CoreEvent creatorEvent) {
```
dfeist: Why do you need to pass event, doesn't EventContext suffice?

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/streaming/BytesStreamingExtensionTestCase.java
```diff
@@ -199,6 +201,26 @@ public void streamProviderSerialization() throws Exception {
...
+  public void nonRepeatableStreamClosesAsync() throws Exception {
+    InputStream stream = (InputStream) flowRunner("toNonRepeatableStream").keepStreamsOpen()
+        .withPayload(data).run().getMessage().getPayload().getValue();
+
+    byte[] bytes = new byte[data.length()];
+    assertThat(stream.read(bytes), is(data.length()));
+    assertThat(stream.read(), is(-1));
+    assertThat(new String(bytes), equalTo(data));
+  }
+
```
dfeist: Need to test the error case too.

marianogonzalez: which error case?

core/src/main/java/org/mule/runtime/core/api/streaming/DefaultStreamingManager.java
marianogonzalez: i pass the event mainly for consistency with the other manage method which is already there and receives the full event. A very secondary `future proofing` point could also be made. By passing the entire event the API is still good if we ever need something else than the context. But that argument pales compared to the consistency one

dfeist: ok. thought it was that, just checking,

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/streaming/BytesStreamingExtensionTestCase.java
dfeist: The case where an error is thrown in a flow and error is handled by an error-handler. This is a different code path and its critical that streams are also closed in this case, so it should really be covered with tests

PR #6119, opened 2017-11-02
MULE-13941: Dynamic flow reference from a For-Each causes ConcurrentModificationException.

ndinu:
Even thought getChildren() returned an unmodifiableList, this type of list only wraps the given list with one that only support get operations. If the underlying list is changed while iterating the unmodifiableList, the ConcurrentModificationException will be thrown.

This is why the ImmutableList was chosen :

A high-performance, immutable, random-access List implementation.
Does not permit null elements.

Unlike Collections.unmodifiableList, which is a view of a separate collection that can still change, an instance of ImmutableList contains its own private data and will never change.
ImmutableList is convenient for public static final lists ("constant lists") and also lets you easily make a "defensive copy" of a list provided to your class by a caller.

### DIFF
--- a/core/src/main/java/org/mule/api/processor/DefaultMessageProcessorPathElement.java
+++ b/core/src/main/java/org/mule/api/processor/DefaultMessageProcessorPathElement.java
@@ -10,6 +10,8 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
+import com.google.common.collect.ImmutableList;
+
 public class DefaultMessageProcessorPathElement implements MessageProcessorPathElement
 {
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mule/api/processor/DefaultMessageProcessorPathElement.java
```diff
@@ -41,7 +43,7 @@ public void setParent(MessageProcessorPathElement parent)
     @Override
     public List<MessageProcessorPathElement> getChildren()
     {
-        return Collections.unmodifiableList(children);
+        return ImmutableList.copyOf(children);
```
fsgonz: Import the static method (though it was not used before with Collection.copy it should be imported like this according to the style rules).

fsgonz: Though on second thought, it is appropriate to leave the ImmutableList so that it is clear that it is an immutable list. Sorry.

PR #6139, opened 2017-11-06
MULE-13969: SFTP receiver reconnection is invoking exceptionListener …

facundovs:
…twice.

SftpMessageReceiver is calling the exception listener to handle connection exceptions and after propagating the exception to the superclass.
However, the PollingReceiverWorker invokes the exception listener too, causing lifecycles errors because Connector#stop and Connector#start are called twice per reconnection.

### DIFF
--- a/transports/sftp/src/test/java/org/mule/transport/sftp/SftpReconnectionTestCase.java
+++ b/transports/sftp/src/test/java/org/mule/transport/sftp/SftpReconnectionTestCase.java
@@ -6,18 +6,28 @@
  */
 package org.mule.transport.sftp;
 
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collection;
 
 import com.jcraft.jsch.SftpException;
+
+import org.junit.Before;
 import org.junit.Test;
 import org.junit.runners.Parameterized;
+
+import org.mule.api.MuleContext;
 import org.mule.api.MuleException;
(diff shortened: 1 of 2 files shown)


transports/sftp/src/test/java/org/mule/transport/sftp/SftpReconnectionTestCase.java
```diff
@@ -94,4 +113,20 @@ protected boolean startServerOnStartUp()
     {
         return false;
     }
+
+    private class TestExceptionStrategy extends AbstractSystemExceptionStrategy {
```
pablokraan: static

transports/sftp/src/test/java/org/mule/transport/sftp/SftpReconnectionTestCase.java
```diff
@@ -94,4 +113,20 @@ protected boolean startServerOnStartUp()
...
+
+        private int numberOfExceptionsHandled = 0;
+        public TestExceptionStrategy(MuleContext muleContext)
```
pablokraan: Add extra line between fields and methods

transports/sftp/src/test/java/org/mule/transport/sftp/SftpReconnectionTestCase.java
```diff
@@ -94,4 +113,20 @@ protected boolean startServerOnStartUp()
...
+        {
+            super(muleContext);
+        }
+
+        @Override
+        protected void handleReconnection(ConnectException ex)
+        {
+            numberOfExceptionsHandled ++;
```
pablokraan: fix formatting

PR #6148, opened 2017-11-07
MULE-13309: Same error is reported multiple times when compiling an i…

estebanwasinger:
…nvalid extension


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/resources/BaseExtensionResourcesGeneratorAnnotationProcessor.java
```diff
@@ -88,9 +91,17 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
         });
       });
 
+
       return false;
+    } catch (MuleRuntimeException e) {
+      if (e.getCause() instanceof IllegalModelDefinitionException) {
```
marianogonzalez: why don't you use ExceptionUtils.findCause() or similar to have a more in depth search of causes?

PR #6171, opened 2017-11-10
MULE-13670: add UUID to Serializable classes

alepulver:


modules/spring-config/src/main/java/org/springframework/beans/factory/support/DefaultListableBeanFactory.java
```diff
@@ -98,6 +98,8 @@
 public class DefaultListableBeanFactory extends AbstractAutowireCapableBeanFactory
     implements ConfigurableListableBeanFactory, BeanDefinitionRegistry, Serializable {
 
+  private static final long serialVersionUID = 7110590944074180546L;
```
elrodro83: why is this Spring stuff serializable?

PR #6176, opened 2017-11-10
MULE-10724: custom-transformer attributes get mixed up

lucianoRM:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/spring/ObjectFactoryClassRepository.java
```diff
@@ -54,21 +55,26 @@
...
                                                     Supplier<Boolean> isLazyInitFunction,
-                                                    Consumer<Object> instancePostCreationFunction) {
+                                                    Optional<Consumer<Object>> instancePostCreationFunctionOptional) {
     try {
-      return objectFactoryClassCache.get(componentBuildingDefinition,
-                                         () -> getObjectFactoryDynamicClass(componentBuildingDefinition,
-                                                                            objectFactoryType,
-                                                                            createdObjectType, isLazyInitFunction,
-                                                                            instancePostCreationFunction));
+      if (instancePostCreationFunctionOptional.isPresent()) {
```
elrodro83: use `map`/`orElse` instead of `isPresent`

lucianoRM: Tried that but since objectFactoryClassCache.get() declares that it throws an ExecutionException, the code needed to make it work with map/orElse would be worst than the one we have now.

PR #6186, opened 2017-11-14
MULE-10832: org.mule.runtime.module.repository.internal.RepositorySystemTestCase fails when executed offline

elrodro83:


tests/unit/src/main/java/org/mule/tck/junit4/rule/RequiresConnectivity.java
```diff
@@ -0,0 +1,44 @@
...
+
+import static org.junit.Assume.assumeTrue;
+
+import org.junit.rules.ExternalResource;
+
+import java.io.IOException;
+import java.net.URL;
+
+/**
+ * Checks that a connection can be established to a provided Url. If not, the test case is ignored.
```
pablokraan: Url > URL

tests/unit/src/main/java/org/mule/tck/junit4/rule/RequiresConnectivity.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @since 4.1
+ */
+public class RequiresConnectivity extends ExternalResource {
```
pablokraan: Why you use a rule instead of an assumption?

tests/unit/src/main/java/org/mule/tck/junit4/rule/RequiresConnectivity.java
```diff
@@ -0,0 +1,44 @@
...
+
+  private String connectivityUrl;
+
+  /**
+   * Constructs a JUnit Rule to assume connectivity to a provided url.
```
pablokraan: url > URL

tests/unit/src/main/java/org/mule/tck/junit4/rule/RequiresConnectivity.java
elrodro83: just for the code to be cleaner if this is to be reused.

PR #6189, opened 2017-11-14
MULE-10304: Application redeployment fails to undeploy old application

pablokraan:


modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/ArtifactArchiveInstaller.java
```diff
@@ -74,6 +75,12 @@ public File installArtifact(final URI artifactUri) throws IOException {
 
       artifactName = getBaseName(fullPath);
       artifactDir = new File(artifactParentDir, artifactName);
+
+      // Removes previous deployed artifact
+      if (artifactDir.exists() && !deleteTree(artifactDir)) {
+        throw new IOException("Cannot delete existing folder " + artifactDir);
```
pablolagreca: is artifactDir enough for logging? Or should it be artifactDir.getAbsolutePath()?

PR #6204, opened 2017-11-17
MULE-13997: Use IO pool to execute blocking policies

elrodro83:


core/src/main/java/org/mule/runtime/core/api/policy/PolicyChain.java
```diff
@@ -79,20 +137,100 @@ public void stop() throws MuleException {
     if (processorChain != null) {
       processorChain.stop();
     }
+    stopIfNeeded(processingStrategy);
   }
 
   @Override
   public CoreEvent process(CoreEvent event) throws MuleException {
-    return processorChain.process(event);
+    return MessageProcessors.processToApply(event, chainWithPs);
```
dfeist: static import

core/src/main/java/org/mule/runtime/core/api/policy/PolicyChain.java
```diff
@@ -79,20 +137,100 @@ public void stop() throws MuleException {
...
+      this.beforeNext = beforeNext;
+      this.afterNext = afterNext;
+      this.followingNext = followingNext;
+    }
+
+    @Override
+    public void start() throws MuleException {
+      SchedulerConfig beforeConfig = schedulerBaseConfig.withName("policy-source-before");
+      if (beforeNext.stream().allMatch(p -> CPU_LITE_ASYNC.equals(p))) {
+        beforeScheduler = schedulerService.cpuLightScheduler(beforeConfig);
```
dfeist: No need to schedule this processing type.

core/src/main/java/org/mule/runtime/core/api/policy/PolicyChain.java
```diff
@@ -79,20 +137,100 @@ public void stop() throws MuleException {
...
+      } else if (beforeNext.stream().anyMatch(p -> BLOCKING.equals(p) || IO_RW.equals(p))) {
+        beforeScheduler = schedulerService.ioScheduler(beforeConfig);
+      } else if (beforeNext.stream().anyMatch(p -> CPU_INTENSIVE.equals(p))) {
+        beforeScheduler = schedulerService.cpuIntensiveScheduler(beforeConfig);
+      }
+
+      SchedulerConfig afterConfig = schedulerBaseConfig.withName("policy-source-after");
+      if (afterNext.stream().allMatch(p -> CPU_LITE_ASYNC.equals(p))) {
+        afterScheduler = schedulerService.cpuLightScheduler(afterConfig);
```
dfeist: Ditto.

core/src/main/java/org/mule/runtime/core/api/policy/PolicyChain.java
```diff
@@ -79,20 +137,100 @@ public void stop() throws MuleException {
...
+
+
+    @Override
+    public ReactiveProcessor onProcessor(ReactiveProcessor processor) {
+      if (processor.equals(followingNext) || (processor instanceof InterceptedReactiveProcessor
+          && ((InterceptedReactiveProcessor) processor).getProcessor().equals(followingNext))) {
+        return publisher -> Flux.from(publisher)
+            .publishOn(afterScheduler != null ? fromExecutor(afterScheduler) : immediate())
+            .transform(processor);
+      } else {
```
dfeist: Need to handle the "after async_cpu_light" in onProcessor too, to ensure we don't do any policy processing in requester selector threads, if requesters are used in policies.

core/src/main/java/org/mule/runtime/core/api/policy/PolicyChain.java
```diff
@@ -79,20 +134,137 @@ public void stop() throws MuleException {
...
+
+    private SchedulerService schedulerService;
+    private SchedulerConfig schedulerBaseConfig;
+    private String schedulersNamePrefix;
+    private List<Processor> beforeNext;
+    private List<Processor> afterNext;
+
+    private Scheduler beforeScheduler;
+    private Scheduler afterScheduler;
+    private Scheduler cpuLiteBeforeScheduler;
```
dfeist: do we really need two different cpu_light schedulers?

core/src/main/java/org/mule/runtime/core/api/policy/PolicyChain.java
```diff
@@ -79,20 +134,137 @@ public void stop() throws MuleException {
...
+      this.schedulerService = schedulerService;
+      this.schedulerBaseConfig = schedulerBaseConfig;
+      this.schedulersNamePrefix = schedulersNamePrefix;
+      this.beforeNext = beforeNext;
+      this.afterNext = afterNext;
+    }
+
+    @Override
+    public void start() throws MuleException {
+      beforeScheduler = schedulerBasedOnProcessingTypes(beforeNext, schedulersNamePrefix + "-before");
```
dfeist: If before does IO and so does after, will this mean two different schedulers? Is there any way to avoid/simplify this?

core/src/main/java/org/mule/runtime/core/api/policy/PolicyChain.java
```diff
@@ -53,7 +83,31 @@ public void setProcessors(List<Processor> processors) {
...
+        beforeNext.add(processor);
+      }
+    }
+
+    processingStrategy = new SourcePolicyProcessingStrategy(schedulerService, muleContext.getSchedulerBaseConfig(),
+                                                            getLocation().getLocation(), beforeNext, afterNext);
+
+    chainBuilder.setProcessingStrategy(processingStrategy);
+    processorChain = chainBuilder.build();
+    chainWithPs = processingStrategy.onPipeline(processorChain);
```
dfeist: I'm wondering if it might be cleaner for processing strategy to have `onChain` that can optionally be implemented and would be used one for each chain.  I'm not that convinced though because in general chains should be transparent and it shouldn't really matter, or change behaviour, if 8 processors are in one chain or two.  So maybe this is simply a special case, that while not that elegant is best we handle like this.  Just thoughts..

core/src/main/java/org/mule/runtime/core/api/policy/PolicyChain.java
```diff
@@ -79,20 +134,137 @@ public void stop() throws MuleException {
...
+          .publishOn(beforeScheduler != null ? fromExecutor(beforeScheduler) : immediate())
+          .transform(pipeline);
+    }
+
+
+    @Override
+    public ReactiveProcessor onProcessor(ReactiveProcessor processor) {
+      ReactiveProcessor reactiveProcessor = handleCpuLiteAsync(processor);
+
+      if (!afterNext.isEmpty()
```
dfeist: It would be a lot easier to understand this if this logic was extracted into a private method, maybe as a predicate.

core/src/main/java/org/mule/runtime/core/api/policy/PolicyChain.java
```diff
@@ -53,7 +83,31 @@ public void setProcessors(List<Processor> processors) {
...
+      }
+
+      if (seenNext) {
+        afterNext.add(processor);
+      } else {
+        beforeNext.add(processor);
+      }
+    }
+
+    processingStrategy = new SourcePolicyProcessingStrategy(schedulerService, muleContext.getSchedulerBaseConfig(),
```
dfeist: Does this mean that each and every policy will have its own processing strategy and set of pools? Isn't there a way which can define a single processingStrategy and then reuse it for all policies?

elrodro83: the pool is the same in the end. Only the scheduler is created per policy.

core/src/main/java/org/mule/runtime/core/api/policy/PolicyChain.java
elrodro83: this is so just the name is reflected in the logs, may help troubleshooting, and doesn't have any real cost.

core/src/main/java/org/mule/runtime/core/api/policy/PolicyChain.java
elrodro83: there is no cost ins having 2 schedulers, the actual thread pool is the same.

PR #6208, opened 2017-11-21
MULE-14115: In Db Module, CLOB implementation is JDBC version dependant.

facundovs:


modules/db/src/test/java/org/mule/module/db/integration/insert/InsertClobInputParamTestCase.java
```diff
@@ -51,9 +52,11 @@ public void usesStringOnImplicitParam() throws Exception
     @Test
     public void usesStringOnExplicitParam() throws Exception
     {
-        MuleEvent response = runFlow("usesStringOnExplicitParam", TEST_MESSAGE);
+        String hugeClobValue = randomAlphabetic(555555);
```
pablokraan: Do you really need a string that long? If yes, indicate why

PR #6212, opened 2017-11-22
MULE-14103: processingStrategy not applied to sub-flow processors

elrodro83:


core/src/main/java/org/mule/runtime/core/internal/processor/chain/SubflowMessageProcessorChainBuilder.java
```diff
@@ -36,7 +38,7 @@ protected MessageProcessorChain createInterceptingChain(Processor head, List<Pro
   /**
    * Generates message processor identifiers specific for subflows.
    */
-  static class SubFlowMessageProcessorChain extends DefaultMessageProcessorChain {
+  static class SubFlowMessageProcessorChain extends DefaultMessageProcessorChain implements ProcessorChainPrototype {
```
dfeist: We should really replace `SubFlowMessageProcessorChain` with a `SubFlowMessageProcessorChainBuilder` instead of leaving it as a chain and then converting into builder and back into chain again elsewhere.

PR #6214, opened 2017-11-22
MULE-14124: Operation field parameters don't work if used with a Para…

marianogonzalez:
…meterGroup

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
@@ -38,6 +38,7 @@ import org.mule.runtime.api.lifecycle.InitialisationException;
 import org.mule.runtime.api.lifecycle.Lifecycle;
 import org.mule.runtime.api.meta.model.ComponentModel;
 import org.mule.runtime.api.meta.model.ExtensionModel;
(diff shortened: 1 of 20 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -230,19 +238,57 @@ protected void doInitialise() throws InitialisationException {
...
+              if (resolver != null) {
+                try {
+                  params.put(getMemberName(p), resolveValue(resolver, resolvingContext));
+                } catch (MuleException e) {
+                  throw new MuleRuntimeException(e);
+                }
+              }
+            });
+      } else {
+        ParameterGroupDescriptor groupDescriptor = group.getModelProperty(ParameterGroupModelProperty.class)
```
estebanwasinger: This whole else could be refactored into several methods, it's a little tricky understand what is doing.

tests/test-extensions/vegan-extension/src/main/java/org/mule/test/vegan/extension/VeganArguments.java
```diff
@@ -0,0 +1,53 @@
...
+
+  public String getArgument3() {
+    return argument3;
+  }
+
+  @Override
+  public String toString() {
+    return Stream.of(argument1, argument2, argument3)
+        .filter(a -> a != null)
+        .collect(Collectors.joining(","));
```
estebanwasinger: Static import 🌝

PR #6217, opened 2017-11-23
MULE-12551: Fix commons-beanutils vulnerability

aiannucci:


core/src/test/java/org/mule/transformer/simple/GetBeanPropertyTestCase.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.junit.Test;
+
+@SmallTest
+public class GetBeanPropertyTestCase extends AbstractMuleTestCase
+{
+    private CustomBean bean = new CustomBean();
+    private GetBeanProperty transformer = new GetBeanProperty();
+
+    @Test(expected = TransformerException.class)
```
afelisatti: Use expected exception instead.

PR #6219, opened 2017-11-23
MULE-14035: Improving http retry policy test cases.

facundovs:


modules/http/src/main/java/org/mule/module/http/internal/request/DefaultHttpRequester.java
```diff
@@ -89,7 +89,7 @@
     static String RETRY_ATTEMPTS_PROPERTY = SYSTEM_PROPERTY_PREFIX + "http.client.maxRetries";
     static final String REMOTELY_CLOSED = "Remotely closed";
     static final int DEFAULT_RETRY_ATTEMPTS = 3;
-    private static final int RETRY_ATTEMPTS  = getInteger(RETRY_ATTEMPTS_PROPERTY, DEFAULT_RETRY_ATTEMPTS);
+    private static int RETRY_ATTEMPTS  = getInteger(RETRY_ATTEMPTS_PROPERTY, DEFAULT_RETRY_ATTEMPTS);
```
dfeist: Can't this be avoided, not ideal to make something not final for a test case.

PR #6224, opened 2017-11-25
MULE-14133: Change DISCARD and DISCARD_OLDEST behavior so that the http connector does not timeout.

fsgonz:


modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/GrizzlyRequestDispatcherFilter.java
```diff
@@ -122,6 +124,19 @@ public void responseReady(HttpResponse httpResponse, ResponseStatusCallback resp
         return ctx.getSuspendAction();
     }
 
+    private void prepareStopAction(final FilterChainContext ctx, final HttpRequestPacket request, HttpStatus status)
```
dfeist: This is more sendEmptyResponse() isn't it?

core/src/main/java/org/mule/config/pool/MuleDiscardOldestPolicy.java
```diff
@@ -0,0 +1,46 @@
...
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.config.pool;
+
+import java.util.concurrent.RejectedExecutionHandler;
+import java.util.concurrent.ThreadPoolExecutor;
+
+public class MuleDiscardOldestPolicy implements RejectedExecutionHandler
```
dfeist: Class level javadoc. How does this differ to DiscardOldestPolicy? Can it replace it? If we don't replace it then can we at least extend it?

core/src/main/java/org/mule/config/pool/MuleDiscardPolicy.java
```diff
@@ -0,0 +1,41 @@
...
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.config.pool;
+
+import java.util.concurrent.RejectedExecutionHandler;
+import java.util.concurrent.ThreadPoolExecutor;
+
+public class MuleDiscardPolicy implements RejectedExecutionHandler
```
dfeist: Class level javadoc. How does this differ to DiscardOldestPolicy? Can it replace it? If we don't replace it then can we at least extend it?

core/src/main/java/org/mule/config/pool/StoppableRunnable.java
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.config.pool;
+
+
+public abstract class StoppableRunnable implements Runnable
```
dfeist: Class-level javadoc.  Also it's more `cancel` than `stop` isn't it? If possible compose Runnable and another interface if one exists.

modules/http/src/test/java/org/mule/module/http/functional/listener/HttpListenerWorkerThreadingProfileTestCase.java
```diff
@@ -76,28 +89,94 @@ public void setup()
...
+
+    private void assertMaxThreadsActive(String flowName, String url, int maxThreadsActive, int expectedHttpStatus) throws Exception
     {
         maxActiveNumberOfRequestExecutedLatch = new CountDownLatch(maxThreadsActive);
 
         sendRequestUntilNoMoreWorkers(flowName, url, maxThreadsActive);
 
         try
         {
+
```
dfeist: Remove?

core/src/main/java/org/mule/config/pool/MuleDiscardOldestPolicy.java
fsgonz: I've extended DiscardOldestPolicy though in this case no logic can be reused (though it is correct extend that class).

core/src/main/java/org/mule/config/pool/MuleDiscardPolicy.java
fsgonz: I've extended DiscardPolicy though in this case no logic can be reused (though it is correct extend that class).

core/src/main/java/org/mule/config/pool/CancellableRunnable.java
```diff
@@ -0,0 +1,17 @@
...
+package org.mule.config.pool;
+
+/**
+ * <code>CancellableRunnable</code> defines an interface where
+ * some logic can be executed in case the work may be cancelled
+ * by a rejection policy.
+ */
+public abstract class CancellableRunnable implements Runnable
+{
+    public abstract void cancelExecution();
```
dfeist: minor: -> `cancel()`

modules/http/src/main/java/org/mule/module/http/api/HttpConstants.java
```diff
@@ -78,6 +78,11 @@ public int getDefaultPort()
     {
         public static String GRIZZLY_MEMORY_MANAGER_SYSTEM_PROPERTY = "org.glassfish.grizzly.DEFAULT_MEMORY_MANAGER";
     }
+    
+    public static final class HttpResponseReasonPhrases
+    {
+        public static String DROPPED = "DROPPED";
```
dfeist: Why not add an enum?  `DROPPED(202, "DROPPED")`?

modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/GrizzlyRequestDispatcherFilter.java
```diff
@@ -122,6 +125,19 @@ public void responseReady(HttpResponse httpResponse, ResponseStatusCallback resp
...
+    {  
+        final HttpResponsePacket.Builder responsePacketBuilder = HttpResponsePacket.builder(request);
+        responsePacketBuilder.status(status.getStatusCode());
+        HttpResponsePacket packet = responsePacketBuilder.build();
+        // this is necessary to avoid chunk-encoding
+        packet.setContentLength(0);
+        packet.setReasonPhrase(reason);
+
+        ctx.write(packet);
+        ctx.notifyDownstream(HttpServerFilter.RESPONSE_COMPLETE_EVENT);
```
dfeist: static import?

PR #6226, opened 2017-11-27
MULE-14135: getFullStackTrace has to be added to ExceptionUtils to guarantee retrocompatibility. #6225

fsgonz:


fsgonz:
This shouldn't be in Mule 4.

PR #6228, opened 2017-11-28
MULE-13210: Allow Core Extensions to Inject container services

pablokraan:
_ Enabling injection of ServiceRepository and service instances on core extensions and server plugins by defining fields or setters annotated with @Inject


pablokraan:
jenkins validate

PR #6229, opened 2017-11-28
MULE-14082: add CRL feature to Mule 4

alepulver:


core/src/main/java/org/mule/runtime/core/privileged/security/RevocationCheck.java
```diff
@@ -0,0 +1,31 @@
...
+import java.security.KeyStore;
+import java.security.cert.TrustAnchor;
+import java.util.Set;
+
+import javax.net.ssl.ManagerFactoryParameters;
+
+/**
+ * Interface for certificate revocation checkers that prepare a trust manager factory configuration.
+ *
+ * @since 3.9
```
afelisatti: I think we need to update this since unfortunately it's not in 4.0

modules/tls/src/main/java/org/mule/runtime/module/tls/internal/revocation/CrlFile.java
```diff
@@ -0,0 +1,125 @@
...
+import javax.net.ssl.ManagerFactoryParameters;
+
+/**
+ * Local file based certificate revocation checker, which requires a CRL file to be accessible and ignores
+ * extension points (additional tags for CRLDP and OCSP) in the certificate.
+ *
+ * Note that the signer of the CRL must be the client certificate's CA, in case that's not possible consider
+ * using {@link CustomOcspResponder}.
+ *
+ * @since 3.9
```
afelisatti: Same

modules/tls/src/main/java/org/mule/runtime/module/tls/internal/revocation/CustomOcspResponder.java
```diff
@@ -0,0 +1,102 @@
...
+
+import javax.net.ssl.CertPathTrustManagerParameters;
+import javax.net.ssl.ManagerFactoryParameters;
+
+/**
+ * Uses a custom OCSP responder for certificate revocation checks, with a specific trusted certificate for
+ * revocating other keys. This ignores extension points (additional tags for CRLDP or OCSP) present in the
+ * certificate, if any.
+ *
+ * @since 3.9
```
afelisatti: Same

modules/tls/src/main/java/org/mule/runtime/module/tls/internal/revocation/StandardRevocationCheck.java
```diff
@@ -0,0 +1,121 @@
...
+
+import javax.net.ssl.CertPathTrustManagerParameters;
+import javax.net.ssl.ManagerFactoryParameters;
+
+/**
+ * Uses the standard JVM certificate revocation checks, which depend on the certificate having the
+ * corresponding extension points (additional tags for CRLDP or OCSP), and the availability
+ * of revocation servers.
+ *
+ * @since 3.9
```
afelisatti: Same

modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/declaration/DefaultXmlArtifactDeclarationLoader.java
```diff
@@ -728,7 +730,26 @@ private void handleInfrastructure(final ParameterModel paramModel,
...
+
+              copyExplicitAttributes(config.getConfigAttributes(), tls);
+
+              config.getChildren().forEach(child -> {
+                ParameterObjectValue.Builder childBuilder = newObjectValue();
+
+
+                if (child.getIdentifier().equals(TLS_REVOCATION_CHECK_ELEMENT_IDENTIFIER)) {
+
+                  cloneAsDeclaration(child.getChildren().get(0),
```
alegmarra: partial apps that are deployed for DS may not have this child element, you have to check if an element exists before accessing it. If none is present just continue, no failure.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/InfrastructureElementModelDelegate.java
```diff
@@ -155,21 +159,48 @@ public void visitSimpleValue(ParameterSimpleValue text) {
...
-                cloneParameters(objectValue, nested);
-                tlsConfig.withNestedComponent(nested.build());
+                if (TLS_REVOCATION_CHECK_ELEMENT_IDENTIFIER.equals(name)) {
+                  InternalComponentConfiguration.Builder revocationCheck = InternalComponentConfiguration.builder()
+                      .withIdentifier(builder()
+                          .namespace(TLS_PREFIX)
+                          .name(TLS_REVOCATION_CHECK_ELEMENT_IDENTIFIER)
+                          .build());
+
+                  InfrastructureTypeMapping.getQName(objectValue.getTypeId())
```
alegmarra: this can be refactored even more, the conditional should cover only this block and the `cloneParameters` in the else. The first and last parts of both if and else blocks are the same. This code is already a mess, so everything that we can simplify is valuable

modules/tls/src/main/java/org/mule/runtime/module/tls/internal/config/TlsComponentBuildingDefinitionProvider.java
```diff
@@ -56,14 +60,37 @@ public void init() {
         .add(baseStore.withIdentifier(TRUSTSTORE).withTypeDefinition(fromType(TrustStoreConfig.class))
             .withSetterParameterDefinition("insecure", fromSimpleParameter("insecure").build()).build());
 
+    componentBuildingDefinitions
+        .add(baseDefinition.withIdentifier("standard-revocation-check")
```
alegmarra: you already have DslConstants for this elements

PR #6237, opened 2017-12-01
MULE-14078: Review graceful shutdown with <async/> scope

lucianoRM:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/SpringRegistry.java
```diff
@@ -407,30 +409,36 @@ public Object unregisterObject(String key) throws RegistrationException {
       return object;
     }
 
+    private void doRegisterObjectSynchronized(String key, Object value) throws RegistrationException {
```
elrodro83: inline this method

PR #6246, opened 2017-12-04
MULE-14152: Poll Doesn't wait for the execution to finish until it sends

elrodro83:
the next event

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/source/scheduler/DefaultSchedulerMessageSource.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/source/scheduler/DefaultSchedulerMessageSource.java
@@ -55,6 +55,7 @@ public class DefaultSchedulerMessageSource extends AbstractComponent
 
   private final PeriodicScheduler scheduler;
   private final NotificationHelper notificationHelper;
+  private final boolean synchronous;
 
   private Scheduler pollingExecutor;
   private ScheduledFuture<?> schedulingJob;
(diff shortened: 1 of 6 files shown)


modules/spring-config/src/main/resources/META-INF/mule-core-common.xsd
```diff
@@ -1493,6 +1493,13 @@
                 <xsd:sequence>
                     <xsd:element ref="scheduling-strategy" minOccurs="1" maxOccurs="1"/>
                 </xsd:sequence>
+                <xsd:attribute name="synchronous" default="false" type="xsd:string">
```
dfeist: Can we use something clearer? `synchronous` makes it seem that it has to do with threading.  Are we sure best default behaviour is false?

modules/spring-config/src/main/resources/META-INF/mule-core-common.xsd
```diff
@@ -1493,6 +1493,13 @@
...
+                    <xsd:annotation>
+                        <xsd:documentation>
+                            Whether additional polling triggers of the scheduler will be executed if there is another exceution currently running.
```
dfeist: If `false` then will executions back-up? Should make this clear here and maybe reconsider the default..

modules/spring-config/src/main/resources/META-INF/mule-core-common.xsd
elrodro83: `false` as behavior is the current behaviour in 4.0.0, so we don't break compatibility. Let me think about the name.

modules/spring-config/src/main/resources/META-INF/mule-core-common.xsd
elrodro83: see previos response regarding the default

core/src/main/java/org/mule/runtime/core/internal/source/scheduler/DefaultSchedulerMessageSource.java
```diff
@@ -125,21 +134,32 @@ private final void run() {
...
+    synchronized (this) {
+      if (disallowConcurrentExecution && executing) {
+        execute = false;
+      } else {
+        execute = true;
+        executing = true;
+      }
+    }
+
+    if (execute) {
```
dfeist: Isn't it clearer to collapse this?

core/src/main/java/org/mule/runtime/core/internal/source/scheduler/DefaultSchedulerMessageSource.java
```diff
@@ -125,21 +134,32 @@ private final void run() {
...
-
   /**
    * Triggers the forced execution of the polling message processor ignoring the configured scheduler.
    */
   private void poll() {
-    Message request = of(null);
-    pollWith(request);
+    boolean execute = false;
+    synchronized (this) {
+      if (disallowConcurrentExecution && executing) {
```
dfeist: Maybe just style but i'd use:
 ```
if(executing.compareAndSet(false, true) || !disallowConcurrentExecution) {
  pollWith(of(null));
}
```

PR #6247, opened 2017-12-04
MULE-14160: HTTP Connector does not route OAuth tokens request throug…

facundovs:
…h proxy.

### DIFF
--- a/modules/oauth/src/main/java/org/mule/module/oauth2/internal/AbstractTokenRequestHandler.java
+++ b/modules/oauth/src/main/java/org/mule/module/oauth2/internal/AbstractTokenRequestHandler.java
@@ -15,9 +15,11 @@ import org.mule.api.MuleContext;
 import org.mule.api.MuleEvent;
 import org.mule.api.MuleException;
 import org.mule.api.context.MuleContextAware;
-import org.mule.module.http.api.HttpConstants;
 import org.mule.module.http.api.client.HttpRequestOptions;
 import org.mule.module.http.api.client.HttpRequestOptionsBuilder;
+import org.mule.module.http.api.requester.HttpRequesterConfig;
+import org.mule.module.http.api.requester.HttpRequesterConfigBuilder;
+import org.mule.module.http.api.requester.proxy.ProxyConfig;
 import org.mule.transport.ssl.api.TlsContextFactory;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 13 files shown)


modules/oauth/src/main/java/org/mule/module/oauth2/internal/AbstractTokenRequestHandler.java
```diff
@@ -62,9 +64,22 @@ public void setTokenUrl(String tokenUrl)
         this.tokenUrl = tokenUrl;
     }
 
-    public void setTlsContextFactory(final TlsContextFactory tlsContextFactory)
+    public void buildHttpRequestOptions (final TlsContextFactory tlsContextFactory, ProxyConfig proxyConfig) throws MuleException
     {
-        httpRequestOptions = HttpRequestOptionsBuilder.newOptions().method(POST.name()).disableStatusCodeValidation().tlsContextFactory(tlsContextFactory).build();
+        if (tlsContextFactory != null && proxyConfig != null)
```
afelisatti: This amount of ifs is unnecessary, use the builders.

modules/oauth/src/main/java/org/mule/module/oauth2/internal/authorizationcode/DefaultAuthorizationCodeGrantType.java
```diff
@@ -256,8 +259,18 @@ public void start() throws MuleException
         }
         if (tokenRequestHandler != null)
         {
-            tokenRequestHandler.setOauthConfig(this);
+            tokenRequestHandler.setOauthConfig(this, proxyConfig);
             tokenRequestHandler.init();
         }
     }
+
+    public ProxyConfig getProxyConfig()
```
afelisatti: Shouldn't this be higher up in the hierarchy considering both types use it? Seems like it should be part of the config element.

modules/oauth/src/main/java/org/mule/module/oauth2/internal/clientcredentials/ClientCredentialsGrantType.java
```diff
@@ -147,4 +153,14 @@ public void setTokenManager(TokenManagerConfig tokenManager)
     {
         this.tokenManager = tokenManager;
     }
+
+    public ProxyConfig getProxyConfig()
```
afelisatti: This is what I mean.

modules/oauth/src/main/resources/META-INF/mule-oauth2.xsd
```diff
@@ -65,6 +66,13 @@
                             </xsd:documentation>
                         </xsd:annotation>
                     </xsd:element>
+                    <xsd:element ref="spring:property" minOccurs="0" maxOccurs="unbounded">
```
elrodro83: This Is fine for 3.x, but for sustaining branches spring properties must be used, so the schema is not modified.

facundovs: Sure. I have created other PR for sustaining branches. Please see: https://github.com/mulesoft/mule/pull/6247

elrodro83: there shouldn be any schema changes in this pr then.

modules/oauth/src/main/java/org/mule/module/oauth2/internal/authorizationcode/DefaultAuthorizationCodeGrantType.java
facundovs: Well, I can't see a configuration element where putting the reference to the proxy configuration. For example the tlsContext is referenced at this same level. 
Regarding the object model, I could put the proxyConfig fields and its respective getters and setters in the abstract class, but it looks like an interface. (There are others shared attributes like the tlsContext that should also be there).
At last, the tlsContext can be configured as a nested element in the gran type configuration, I am not sure if the same approach applies to the proxy configuration (I mean in the minor/major release and not here where we are only defining spring properties).

modules/oauth/src/main/resources/META-INF/mule-oauth2.xsd
facundovs: It is not possible to put spring properties here, without adding this element in the schema.
At least in grant type element,, the spring properties are not inherited.

modules/oauth/src/main/java/org/mule/module/oauth2/internal/authorizationcode/DefaultAuthorizationCodeGrantType.java
afelisatti: I don't understand why it cannot go into `AbstractGrantType` instead of being repeated in both types.

PR #6248, opened 2017-12-04
MULE-8923: maxRows on db:select element is not really substitutable

pablokraan:


modules/db/src/main/java/org/mule/module/db/internal/config/processor/AbstractDbProcessorDefinitionParser.java
```diff
@@ -78,29 +78,29 @@ protected void parseTargetExpression(Element element, BeanDefinitionBuilder buil
 
     protected Object parseStatementFactory(Element element)
     {
-        QueryStatementFactory defaultStatementFactory = new QueryStatementFactory();
+        BeanDefinitionBuilder defaultStatementFactory = BeanDefinitionBuilder.genericBeanDefinition(QueryStatementFactory.class);
```
lucianoRM: static import

PR #6254, opened 2017-12-05
MULE-14041: Update version of grizzly according to new release due to

fsgonz:
100 continue handling bug.


modules/http/src/test/java/org/mule/module/http/functional/AbstractHttpExpectHeaderServerTestCase.java
```diff
@@ -112,8 +112,6 @@ public void run()
 
                 reader.close();
                 writer.close();
-                socket.close();
```
afelisatti: You should add another test, not change the current one.

PR #6258, opened 2017-12-05
MULE-13979: Find a workaround for not having to detect policy logger …

pablokraan:
…context in runtime

### DIFF
--- a/modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/ArtifactAwareContextSelector.java
+++ b/modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/ArtifactAwareContextSelector.java
@@ -6,8 +6,9 @@
  */
 package org.mule.runtime.module.launcher.log4j2;
 
+import static com.google.common.cache.CacheBuilder.newBuilder;
 import static java.lang.ClassLoader.getSystemClassLoader;
-
+import static java.lang.Thread.currentThread;
 import org.mule.runtime.api.lifecycle.Disposable;
 import org.mule.runtime.deployment.model.api.policy.PolicyTemplateDescriptor;
 import org.mule.runtime.deployment.model.internal.domain.MuleSharedDomainClassLoader;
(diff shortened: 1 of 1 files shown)


modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/ArtifactAwareContextSelector.java
```diff
@@ -94,23 +100,43 @@ public void removeContext(LoggerContext context) {
    * @return the {@link ClassLoader} owner of the logger context
    */
   static ClassLoader resolveLoggerContextClassLoader(ClassLoader classLoader) {
-    ClassLoader loggerClassLoader = classLoader == null ? Thread.currentThread().getContextClassLoader() : classLoader;
+    WeakReference<ClassLoader> loggerClassLoaderReference = classLoaderMap.get(classLoader);
+    ClassLoader loggerClassLoader;
+
+    if (loggerClassLoaderReference == null) {
+      loggerClassLoader = getLoggerClassLoader(classLoader);
```
elrodro83: even though the map is synchronized, this code may be executed concurrently and the loggerClassLoader calculated extra times needlessly.

modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/ArtifactAwareContextSelector.java
```diff
@@ -56,6 +60,8 @@
   private final MuleLoggerContextFactory loggerContextFactory = new MuleLoggerContextFactory();
 
   private LoggerContextCache cache = new LoggerContextCache(this, getClass().getClassLoader());
+  private static final Map<ClassLoader, WeakReference<ClassLoader>> classLoaderMap =
```
elrodro83: rename to `loggerClassLoaders` or `loggerClassLoadersMap`

modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/ArtifactAwareContextSelector.java
pablokraan: sure, but that is a minor performance overhead, I can't think of any solution that will avoid that duplicated calculation without requiring a locking, which will be worst that this solution. Do you have an alternative solution?

pablokraan: disregard my previous comment...

PR #6265, opened 2017-12-06
MULE-14054: It should be possible to enable http messages logging per…

pablokraan:
… app


modules/launcher/src/main/java/org/mule/runtime/module/launcher/log4j2/LoggerContextConfigurer.java
```diff
@@ -207,18 +210,24 @@ private void addDefaultArtifactContext(MuleLoggerContext context) {
...
+        addDefaultAppender(context, logFile.getAbsolutePath());
+      }
     }
   }
 
-  private void removeConsoleAppender(LoggerContext context) {
+  private void removeConsoleAppender(LoggerContext context, Function<String, Boolean> appenderNameMatcher) {
     for (Appender appender : getRootLogger(context).getAppenders().values()) {
-      if (appender instanceof ConsoleAppender) {
+      if (appender instanceof ConsoleAppender && appenderNameMatcher.apply(appender.getName())) {
```
elrodro83: move the `appender instanceof ConsoleAppender &&` code to each predicate, remove `Console` from this method name.

PR #6275, opened 2017-12-11
 MULE-14137: Implement Partial fetching in Multilevel DataSense

alegmarra:


modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/metadata/MetadataOperationTestCase.java
```diff
@@ -339,24 +382,40 @@ public void resolverOutputWithContextClassLoader() throws Exception {
   }
 
   @Test
-  public void shouldInheritOperationResolvers() throws Exception {
-    location = Location.builder().globalName(SHOULD_INHERIT_OPERATION_RESOLVERS).addProcessorsPart().addIndexPart(0).build();
-    assertInheritedResolvers(PERSON_METADATA_KEY);
-  }
-
-  @Test
+  @Ignore("MULE-14190: MetadataScope fundamentally broken")
```
marianogonzalez: I'm honoured

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/ValueProviderFactory.java
```diff
@@ -80,25 +77,28 @@ ValueProvider createValueProvider() throws ValueResolvingException {
...
+  private void injectValueProviderFields(ValueProvider resolver) throws ValueResolvingException {
     List<String> missingParameters = new ArrayList<>();
     for (InjectableParameterInfo injectableParam : factoryModelProperty.getInjectableParameters()) {
+      Object parameterValue = null;
       String parameterName = injectableParam.getParameterName();
       try {
-        Object parameterValue = parameterValueResolver.getParameterValue(parameterName);
+        parameterValue = parameterValueResolver.getParameterValue(parameterName);
+      } catch (org.mule.runtime.module.extension.internal.runtime.ValueResolvingException ignored) {
+        // no op
```
marianogonzalez: why? very weird

alegmarra: because `getParameterValue` can throw that `ValueResolvingException` if the resolution fails, but we don't want neither to fail there nor throw that exception, just to mark the parameter as missing and then throw the general exception bellow with all the info.

alegmarra: that said, I'll add a debug log

modules/extensions-spring-support/src/test/resources/models/vegan.json
```diff
@@ -7709,7 +7730,7 @@
             "type": "String"
           }
         },
-        "hasDynamicType": true,
+        "hasDynamicType": false,
```
estebanwasinger: why is this false now?

alegmarra: because it was never used and inhereted by metadataScope, which was removed from that class

PR #6282, opened 2017-12-12
MULE-14194: HTTP connector fails to close streaming content connection from HEAD request

ndinu:


modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/ResponseStreamingCompletionHandler.java
```diff
@@ -103,6 +104,10 @@ public void completed(WriteResult result)
                 {
                     doComplete();
                 }
+                else if (isDone && httpResponsePacket.getRequest().getMethod().equals(HEAD))
```
afelisatti: Can we analyse whether we should force an `EmptyHttpEntity` at some other point instead?

modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/GrizzlyRequestDispatcherFilter.java
```diff
@@ -106,6 +108,15 @@ public void responseReady(HttpResponse httpResponse, ResponseStatusCallback resp
             {
                 try
                 {
+                    //If the HTTP Method is HEAD we don't want to process the whole body only to be discarted later
+                    if(httpRequest.getMethod().equals("HEAD"))
```
afelisatti: Use the constants in `HttpConstants`

modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/ResponseCompletionHandler.java
```diff
@@ -116,11 +118,13 @@ public void completed(WriteResult result)
             if (!isDone)
             {
                 sendResponse();
+                if(isDone && httpResponsePacket.getRequest().getMethod().equals(HEAD)){
```
afelisatti: Why is this still required?

PR #6292, opened 2017-12-14
MULE-14258: StreamingHelper exposed for priviledge APIs

aamura:
StreamingHelper class exposed when you are using Declaration Enricher Class


marianogonzalez:
It doesn't makes sense for this factory to be API. And for this factory to be API, the ArgumentResolver should also be promoted to API which doesn't make sense either.

If the goal is to obtain a StreamingHelper inside a privileged extension, then just create a Privileged StreamingHelperFactory class

aamura:
I renamed the Factory class to StreamingHelperFactory

PR #6294, opened 2017-12-14
MULE-14259 - Classloader issue when working with VM and application p…

pablolagreca:
…atch deployment

### DIFF
--- a/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/domain/DefaultDomainFactory.java
+++ b/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/domain/DefaultDomainFactory.java
@@ -16,6 +16,7 @@ import static org.mule.runtime.deployment.model.internal.DefaultRegionPluginClas
 import static org.mule.runtime.module.reboot.api.MuleContainerBootstrapUtils.getMuleDomainsDir;
 import org.mule.runtime.api.meta.model.ExtensionModel;
 import org.mule.runtime.api.service.ServiceRepository;
+import org.mule.runtime.deployment.model.api.DeployableArtifactDescriptor;
 import org.mule.runtime.deployment.model.api.domain.Domain;
 import org.mule.runtime.deployment.model.api.domain.DomainDescriptor;
 import org.mule.runtime.deployment.model.api.plugin.ArtifactPlugin;
@@ -24,9 +25,10 @@ import org.mule.runtime.deployment.model.internal.domain.DomainClassLoaderBuilde
 import org.mule.runtime.deployment.model.internal.plugin.PluginDependenciesResolver;
 import org.mule.runtime.module.artifact.api.classloader.ClassLoaderRepository;
 import org.mule.runtime.module.artifact.api.classloader.MuleDeployableArtifactClassLoader;
-import org.mule.runtime.module.deployment.impl.internal.artifact.ArtifactFactory;
+import org.mule.runtime.module.deployment.impl.internal.artifact.AbstractDeployableArtifactFactory;
 import org.mule.runtime.module.deployment.impl.internal.plugin.DefaultArtifactPlugin;
 import org.mule.runtime.module.extension.internal.loader.ExtensionModelLoaderManager;
+import org.mule.runtime.module.license.api.LicenseValidator;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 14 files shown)


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/domain/DefaultDomainFactory.java
```diff
@@ -159,7 +141,34 @@ public Domain createArtifact(File domainLocation, Optional<Properties> deploymen
...
+    List<ArtifactPluginDescriptor> artifactPluginDescriptors =
+        domainDescriptor.getPlugins().stream().collect(Collectors.toList());
+    List<ArtifactPluginDescriptor> resolvedArtifactPluginDescriptors =
+        pluginDependenciesResolver.resolve(artifactPluginDescriptors);
+
+    DomainClassLoaderBuilder artifactClassLoaderBuilder =
+        domainClassLoaderBuilderFactory.createArtifactClassLoaderBuilder();
+    MuleDeployableArtifactClassLoader domainClassLoader =
+        artifactClassLoaderBuilder
+            .addArtifactPluginDescriptors(resolvedArtifactPluginDescriptors.toArray(new ArtifactPluginDescriptor[0]))
```
marianogonzalez: `new ArtifactPluginDescriptor[0]` ?  I would expect a `[1]` instead. Anyway, I believe that the best practice is to use `ArtiactPluginDescriptor[]::new` and the stream automatically dims the array properly

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/domain/DefaultDomainFactory.java
```diff
@@ -159,7 +141,34 @@ public Domain createArtifact(File domainLocation, Optional<Properties> deploymen
       throw new IllegalArgumentException("Mule domain name may not contain spaces: " + domainName);
     }
 
-    return createArtitact(findDomain(domainName, deploymentProperties));
+    DomainDescriptor domainDescriptor = findDomain(domainName, deploymentProperties);
 
+    List<ArtifactPluginDescriptor> artifactPluginDescriptors =
+        domainDescriptor.getPlugins().stream().collect(Collectors.toList());
```
marianogonzalez: static import

modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/internal/artifact/ArtifactUtils.java
```diff
@@ -0,0 +1,37 @@
...
+   * <p/>
+   * The artifact name gets resolved based on the version of the artifact. {@link BundleDescriptor}s that only differ on the patch
+   * version will generate the same artifact name.
+   *
+   * @param bundleDescriptor the bundle descriptor of the deployable archive
+   * @return a unique artifact name within the artifact kind (app or domain).
+   */
+  public static String generateArtifactName(BundleDescriptor bundleDescriptor) {
+    MuleVersion artifactVersion = new MuleVersion(bundleDescriptor.getVersion());
+    return format("%s-%s-%s.%s", bundleDescriptor.getGroupId(), bundleDescriptor.getArtifactId(), artifactVersion.getMajor(),
```
marianogonzalez: aren't you still having the same issue only that harder to reproduce? Because you're no longer using the patch version, the broken test should work. But I still could have an app with ongoing job instances which gets upgraded to the next minor. This new version adds stuff, it doesn't change any of the existing flows. The problem persists.

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/domain/DefaultDomainFactory.java
pablolagreca: It doesn't work. Changing it to new ArtifactPluginDescriptor[resolvedArtifactPluginDescriptors.size()]

modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/internal/artifact/ArtifactUtils.java
pablolagreca: I understand but I would cover this in a different issue. I've created  https://www.mulesoft.org/jira/browse/MULE-14260 ask the PM to prioritize it please.

PR #6297, opened 2017-12-14
MULE-14194: HTTP connector fails to close streaming content connection from HEAD request

ndinu:


modules/http-api/src/main/java/org/mule/runtime/http/api/domain/message/response/HttpResponseBuilder.java
```diff
@@ -20,6 +20,16 @@
 
   private ResponseStatus responseStatus = new ResponseStatus();
 
+  public HttpResponseBuilder(HttpResponse httpResponse) {
```
aiannucci: add javadoc to this public constructor

modules/http-api/src/main/java/org/mule/runtime/http/api/domain/message/HttpMessageBuilder.java
```diff
@@ -29,6 +29,21 @@
...
+  protected HttpMessageBuilder() {}
+
+  protected HttpMessageBuilder(HttpMessage httpMessage) {
+    entity(httpMessage.getEntity());
+    headers(httpMessage);
+  }
+
+  private void headers(HttpMessage httpMessage) {
+    for (String headerName : httpMessage.getHeaderNames()) {
+      for (String headerValue : httpMessage.getHeaderValues(headerName)) {
```
alegmarra: use `put(K key, Collection<V> values)`

PR #6299, opened 2017-12-15
MULE-14263: Plugin bean factories are being loaded with the applicati…

facundovs:
…on classloader.

### DIFF
--- a/modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/spring/ObjectFactoryClassRepository.java
+++ b/modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/spring/ObjectFactoryClassRepository.java
@@ -90,50 +90,52 @@ public class ObjectFactoryClassRepository {
      * create as FactoryBean a dynamic class that will have the same attributes and methods as the ObjectFactory that the user
      * defined. This way our API does not expose spring specific classes.
      */
-    Enhancer enhancer = new Enhancer();
-    // Use SmartFactoryBean since it's the only way to force spring to pre-instantiate FactoryBean for singletons
-    enhancer.setInterfaces(new Class[] {SmartFactoryBean.class});
-    enhancer.setSuperclass(objectFactoryType);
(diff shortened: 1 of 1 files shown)


modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/spring/ObjectFactoryClassRepository.java
```diff
@@ -125,6 +129,10 @@
           return proxy.invokeSuper(obj, args);
         }
     });
+
+    if (originalClassLoader != null) {
```
pablolagreca: If the previous code throws an exception this code is never call. It must so it should be in a finally clause

modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/spring/ObjectFactoryClassRepository.java
```diff
@@ -90,42 +90,52 @@
...
+      enhancer.setInterfaces(new Class[] {SmartFactoryBean.class});
+      enhancer.setSuperclass(objectFactoryType);
+      enhancer.setCallbackType(MethodInterceptor.class);
+      if (SmartFactoryBean.class.getClassLoader() != objectFactoryType.getClassLoader()) {
+        // CGLIB needs access to both the spring interface and the extended factory class.
+        // If the factory class is defined in a plugin, its classloader has to be passed.
+        CompositeClassLoader loader = new CompositeClassLoader(ObjectFactoryClassRepository.class.getClassLoader(),
+                                                               objectFactoryType.getClassLoader());
+        enhancer.setClassLoader(loader);
+        originalClassLoader = Thread.currentThread().getContextClassLoader();
```
pablolagreca: This must be done above in the variable declaration and not here. If not in the finally clause you may be setting a null.

facundovs: but I'm checking that originalClassLoader is not null in the finally block.

PR #6301, opened 2017-12-15
 MULE-14264: Fix HttpRequestProxyConfigTestCase

facundovs:
For ensuring the request is going through proxy, we are handling the request in a socket server, and immediately closing the socket. In some random scenarios, closing the socket immediately causes a null message exception (probably caused by the grizzly implementation). We should also consider the "Connection reset by peer" message, which is possible if a connection is closed in an unexpected way.


modules/http/src/test/java/org/mule/module/http/functional/requester/HttpRequestProxyConfigTestCase.java
```diff
@@ -175,6 +178,15 @@ public void run()
                     Thread.yield();
                 }
 
+                try
+                {
+                    sleep(100);
```
afelisatti: How are you certain this is enough?

PR #6305, opened 2017-12-15
MULE-14268: Fixed binding of variables in operations

3miliano:


marianogonzalez:
Please re send. Looks like you did something wrong and included several old commits

PR #6307, opened 2017-12-15
MULE-13831 Use shared reactor stream(s) rather than once per event fo…

dfeist:
…r performance and back-pressure


core/src/main/java/org/mule/runtime/core/internal/processor/strategy/ProactorStreamProcessingStrategyFactory.java
```diff
@@ -43,46 +43,35 @@
...
-                                                          .withName(schedulersNamePrefix + "." + BLOCKING.name())),
-                                                  () -> muleContext.getSchedulerService()
-                                                      .cpuIntensiveScheduler(muleContext.getSchedulerBaseConfig()
-                                                          .withName(schedulersNamePrefix + "." + CPU_INTENSIVE.name())),
-                                                  getMaxConcurrency());
-    }
+    return new ProactorStreamProcessingStrategy(() -> muleContext.getSchedulerService()
+        .customScheduler(muleContext.getSchedulerBaseConfig()
+            .withName(schedulersNamePrefix + RING_BUFFER_SCHEDULER_NAME_SUFFIX)
+            .withMaxConcurrentTasks(getSubscriberCount() + 1)),
```
elrodro83: what was the reason for this `+1`? I know it was in the previous code, but can you add a comment?

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/TransactionAwareProactorStreamProcessingStrategyFactory.java
```diff
@@ -22,56 +26,41 @@
...
-                                                                  () -> muleContext.getSchedulerService()
-                                                                      .cpuIntensiveScheduler(muleContext.getSchedulerBaseConfig()
-                                                                          .withName(schedulersNamePrefix + "."
-                                                                              + CPU_INTENSIVE.name())),
-                                                                  getMaxConcurrency());
-    }
+    return new TransactionAwareProactorStreamProcessingStrategy(() -> muleContext.getSchedulerService()
+        .customScheduler(muleContext.getSchedulerBaseConfig()
+            .withName(schedulersNamePrefix + RING_BUFFER_SCHEDULER_NAME_SUFFIX)
+            .withMaxConcurrentTasks(getSubscriberCount() + 1)),
```
elrodro83: same as abobe. Also, can common code be extracted?

PR #6308, opened 2017-12-18
MULE-14198: flow-ref is missing target attribute

elrodro83:


core/src/main/java/org/mule/runtime/core/internal/util/rx/Operators.java
```diff
@@ -46,6 +51,19 @@ private Operators() {}
     };
   }
 
+  public static Function<CoreEvent, CoreEvent> outputToTarget(CoreEvent originalEvent, String target,
+                                                              String targetValueExpression,
+                                                              ExtendedExpressionManager expressionManager) {
+    return result -> {
+      if (target != null) {
+        TypedValue targetValue = expressionManager.evaluate(targetValueExpression, getTargetBindingContext(result.getMessage()));
+        return builder(originalEvent).addVariable(target, targetValue.getValue(), targetValue.getDataType()).build();
```
marianogonzalez: out of this PR scope, but this api is very chota

core/src/main/java/org/mule/runtime/core/internal/util/rx/Operators.java
```diff
@@ -46,6 +51,19 @@ private Operators() {}
     };
   }
 
+  public static Function<CoreEvent, CoreEvent> outputToTarget(CoreEvent originalEvent, String target,
+                                                              String targetValueExpression,
+                                                              ExtendedExpressionManager expressionManager) {
+    return result -> {
+      if (target != null) {
```
marianogonzalez: @afelisatti you probably want to leverage this on your PR

core/src/main/java/org/mule/runtime/core/internal/util/rx/Operators.java
```diff
@@ -46,6 +51,19 @@ private Operators() {}
     };
   }
 
+  public static Function<CoreEvent, CoreEvent> outputToTarget(CoreEvent originalEvent, String target,
+                                                              String targetValueExpression,
+                                                              ExtendedExpressionManager expressionManager) {
```
afelisatti: Wouldn't `ExpressionManager` do?

elrodro83: `ExpressionLanguage` actually

PR #6321, opened 2017-12-19
 MULE-14193: Wrong ArtifactType in SpringXmlConfigurationBuilder

elrodro83:
* default for beans in registry-bootstrap is APP/DOMAIN/POLICY instead of just APP


core/src/main/java/org/mule/runtime/core/api/config/bootstrap/ArtifactType.java
```diff
@@ -6,17 +6,25 @@
...
 
 /**
  * List the possible types of deployable artifacts
  *
  * @since 3.7.0
  */
 public enum ArtifactType {
-  APP("app"), DOMAIN("domain"), PLUGIN("plugin"), POLICY("policy"), SERVICE("service"), SERVER_PLUGIN("serverPlugin"), ALL(
-      "app/domain");
+  APP("app"), DOMAIN("domain"), PLUGIN("plugin"), POLICY("policy"), SERVICE("service"), SERVER_PLUGIN("serverPlugin"),
```
pablokraan: I dont think that the enum must include plugin, service and server plugin, as those artifact do not have a mule context/registry so there is no place where the bootstrap properties can be applied.

core/src/main/java/org/mule/runtime/core/internal/config/bootstrap/AbstractRegistryBootstrap.java
```diff
@@ -264,7 +281,7 @@ private void registerObjects(List<ObjectBootstrapProperty> bootstrapProperties)
 
   private void registerObject(ObjectBootstrapProperty bootstrapProperty) throws Exception {
     try {
-      if (!bootstrapProperty.getArtifactType().equals(ALL) && !bootstrapProperty.getArtifactType().equals(artifactType)) {
+      if (!bootstrapProperty.getArtifactTypes().contains(artifactType) && !bootstrapProperty.getArtifactTypes().contains(ALL)) {
```
pablokraan: Is still required to compare against ALL? I think that ALL should not be used at all

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/PolicyMuleContextBuilder.java
```diff
@@ -0,0 +1,41 @@
...
+import static org.mule.runtime.core.api.util.StringUtils.isBlank;
+
+import org.mule.runtime.core.api.config.DefaultMuleConfiguration;
+
+import java.util.Map;
+
+/**
+ * Takes Mule policy descriptor into account when building the context.
+ */
+public class PolicyMuleContextBuilder extends SupportsPropertiesMuleContextBuilder {
```
pablokraan: `@since`

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/SupportsPropertiesMuleContextBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+ * @since 4.1
+ */
+abstract class SupportsPropertiesMuleContextBuilder extends DefaultMuleContextBuilder {
+
+  private static final String SYS_PROPERTY_PREFIX = "sys.";
+  private static final String MULE_PROPERTY_PREFIX = "mule.config.";
+
+  private final Map<String, String> artifactProperties;
+
+  protected SupportsPropertiesMuleContextBuilder(ArtifactType artifactType, Map<String, String> artifactProperties) {
```
pablokraan: Add javadoc

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultPolicyInstanceProviderFactory.java
```diff
@@ -48,7 +49,7 @@ public ApplicationPolicyInstance create(Application application, PolicyTemplate
                                           PolicyParametrization parametrization) {
     return new DefaultApplicationPolicyInstance(application, policyTemplate, parametrization, serviceRepository,
                                                 classLoaderRepository, policyTemplate.getArtifactPlugins(),
-                                                extensionModelLoaderRepository);
+                                                extensionModelLoaderRepository, null);
```
pablokraan: Why no context listener is passed here?

modules/deployment/src/test/resources/org/foo/hello/HelloExtension.java
```diff
@@ -26,9 +26,15 @@
   @Inject
   private FooService fooService;
 
+  @Inject
+  private HelloRegistryBean registryBean;
+
   public HelloExtension() {}
 
   public String getMessage() {
+    if(registryBean == null) {
```
pablokraan: Fix code style :)

core/src/main/java/org/mule/runtime/core/api/config/bootstrap/ArtifactType.java
elrodro83: These values are also used to load the artifact descriptor, not just to build the Mulecontext

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultPolicyInstanceProviderFactory.java
elrodro83: i added it just to use in the tests. It is null-checked

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/SupportsPropertiesMuleContextBuilder.java
```diff
@@ -27,6 +28,11 @@
 
   private final Map<String, String> artifactProperties;
 
+  /**
+   * Constructs a new {@link MuleContext} builder with the given {@code artifactProperties}.
+   * @param artifactType the type of artifact the target {@link MuleContext} is for.
```
pablokraan: add extra line

PR #6331, opened 2017-12-20
MULE-14204: Error count of flow statistics not tracked

elrodro83:


core/src/main/java/org/mule/runtime/core/internal/exception/ErrorHandler.java
```diff
@@ -170,4 +173,12 @@ public void setRootContainerName(String rootContainerName) {
     }
   }
 
+  public void setStatistics(FlowConstructStatistics flowStatistics) {
```
afelisatti: Are you certain that at this point you'll have the injected exception listeners as well? (e.g.: OnCriticalErrorHandler)

PR #6335, opened 2017-12-21
MULE-14044: Implement Java interop Module

alegmarra:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/function/ReflectiveExpressionFunctionExecutor.java
```diff
@@ -74,7 +79,7 @@ public Object call(Object[] parameters, BindingContext context) {
 
   @Override
   public void initialise() throws InitialisationException {
-    initialiseIfNeeded(componentInstance);
+    initialiseIfNeeded(componentInstance, muleContext);
```
marianogonzalez: no DI?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/IntrospectionUtils.java
```diff
@@ -518,7 +518,16 @@ public static MetadataType getFieldMetadataType(Field field, ClassTypeLoader typ
 
   public static Optional<Field> getField(Class<?> clazz, String name) {
     Collection<Field> candidates = getAllFields(clazz, withName(name));
-    return isEmpty(candidates) ? empty() : Optional.of(candidates.iterator().next());
+    if (isEmpty(candidates)) {
+      return empty();
+    }
+
+    Field field = candidates.stream()
+        .filter(f -> f.getDeclaringClass().equals(clazz))
```
marianogonzalez: this will probably have side effects and doesn't really comply with the intent of this method. Why do you need this? It should be a separate method then (not to mention than doing a getAllFields to then filter by declaring class is very unneficient

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/IntrospectionUtils.java
```diff
@@ -518,7 +518,16 @@ public static MetadataType getFieldMetadataType(Field field, ClassTypeLoader typ
...
+        .findFirst()
+        .orElseGet(() -> candidates.iterator().next());
+
+    return Optional.of(field);
```
marianogonzalez: redundant

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/metadata/MetadataNegativeTestCase.java
```diff
@@ -214,11 +212,4 @@ public void failToGetMetadataWithMissingMetadataKeyLevels() throws Exception {
     assertMetadataFailure(result.getFailures().get(0), "Missing levels: [city]", INVALID_METADATA_KEY, "", COMPONENT);
   }
 
-  @Test
-  public void operationCantResolverVoidAsOutputTypeFromList() throws Exception {
```
marianogonzalez: why? what replaces this?

alegmarra: nothing, we now allow it

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/function/ReflectiveExpressionFunctionExecutor.java
alegmarra: done

PR #6337, opened 2017-12-21
MULE-12768: Message sources need to be able to define how they want t…

marianogonzalez:
…he Flow to handle back

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceAdapter.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceAdapter.java
@@ -221,15 +221,15 @@ public final class SourceAdapter implements Startable, Stoppable, Initialisable 
           .map(e -> sourceOverloadErrorType.equals(e.getErrorType()))
           .orElse(false);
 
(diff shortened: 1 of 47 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceAdapter.java
```diff
@@ -199,7 +216,19 @@ public DefaultSourceCompletionHandler(SourceCallbackExecutor onSuccessExecutor,
 
     @Override
     public Publisher<Void> onFailure(MessagingException exception, Map<String, Object> parameters) {
-      return from(onErrorExecutor.execute(exception.getEvent(), parameters, context))
+      final CoreEvent event = exception.getEvent();
```
dfeist: This needs to change in a future PR right? ModuleFlowProcessingPhase shouldn't really even call onFailure on back-pressue should it, or are you ok with keeping this here?

marianogonzalez: That's actually a question for you. I did it here because this is where you are communicating the error. If you start sending it somewhere else, fine by me

PR #6344, opened 2017-12-22
MULE-14173: Add a splash information about services when started

elrodro83:


modules/service/src/main/java/org/mule/runtime/module/service/internal/manager/MuleServiceManager.java
```diff
@@ -86,6 +88,25 @@ private void startServices() throws MuleException {
...
+                header.add("Started " + service.toString());
+                header.add("");
+
+                for (String splashMessageLine : service.splashMessageLines()) {
+                  header.add(splashMessageLine);
+                }
+              }
+
+            };
+            splashScreen.setHeader(null);
```
pablokraan: Why do you have to set it to null?

modules/service/src/main/java/org/mule/runtime/module/service/internal/manager/ServiceSplashScreen.java
```diff
@@ -0,0 +1,32 @@
...
+ * LICENSE.txt file.
+ */
+
+package org.mule.runtime.module.service.internal.manager;
+
+import static java.lang.System.lineSeparator;
+
+import org.mule.runtime.api.service.Service;
+import org.mule.runtime.core.internal.util.splash.SplashScreen;
+
```
pablokraan: Add javadoc

modules/service/src/main/java/org/mule/runtime/module/service/internal/manager/ServiceSplashScreen.java
```diff
@@ -0,0 +1,32 @@
...
+final class ServiceSplashScreen extends SplashScreen {
+
+  private final Service service;
+
+  ServiceSplashScreen(Service service) {
+    this.service = service;
+    doHeader();
+  }
+
+  protected void doHeader() {
```
pablokraan: private

PR #6350, opened 2017-12-26
MULE-14319: encoded characters in SMTP username cause failure on reconnection

ndinu:


transports/email/src/test/java/org/mule/transport/email/functional/SmtpTransportConfigTestCase.java
```diff
@@ -0,0 +1,76 @@
...
+        return Arrays.asList(new Object[][]{
+                {ConfigVariant.FLOW, "smtp-transport-config-test.xml"}
+        });
+    }
+
+    @Test
+    public void testSend() throws Exception {
+        runFlow("endpointTransportConfigFlow");
+        latch.await();
+        assertThat(user, is(TEST_USER));
```
fsgonz: Shouldn't the password be verified as well? It would be better to add special characters in the password and add an assert involving that also. Can it be retrieved from the transport's uri in the same way?

transports/email/src/test/java/org/mule/transport/email/functional/SmtpTransportConfigTestCase.java
```diff
@@ -0,0 +1,76 @@
...
+    public SmtpTransportConfigTestCase(ConfigVariant variant, String configResources)
+    {
+        super(variant, STRING_MESSAGE, "smtp", configResources);
+    }
+
+    @Parameterized.Parameters
+    public static Collection<Object[]> parameters()
+    {
+        return Arrays.asList(new Object[][]{
+                {ConfigVariant.FLOW, "smtp-transport-config-test.xml"}
```
fsgonz: Static import. As this is a new test please use a static import for this static field.

transports/email/src/test/java/org/mule/transport/email/functional/SmtpTransportConfigTestCase.java
fsgonz: It is not verified as the password is not visible from the SmtpTransport.

PR #6353, opened 2017-12-27
MULE-14308: @Parameter as operation attribute not working for routers

alegmarra:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/ScopeModelLoaderDelegate.java
```diff
@@ -96,9 +97,17 @@ void declareScope(ExtensionDeclarer extensionDeclarer,
...
                            processorChain.stream().map(ExtensionParameter::getName).collect(toList())));
 
+
+    ParameterDeclarationContext declarationContext = new ParameterDeclarationContext(SCOPE, scope.getDeclaration());
     loader.getMethodParametersLoader().declare(scope,
                                                scopeMethod.getParameters(),
-                                               new ParameterDeclarationContext(SCOPE, scope.getDeclaration()));
+                                               declarationContext);
+
+    final List<ExtensionParameter> fieldParameters = scopeMethod.getEnclosingType().getParameters();
```
marianogonzalez: this is pretty much the same code as for operations. CAn this be reused somehow?

PR #6365, opened 2017-12-28
MULE-14309: Domain is not being disposed with its own classloader.

facundovs:


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/domain/DefaultMuleDomain.java
```diff
@@ -220,7 +222,17 @@ public void stop() {
...
+        Reference<MuleException> error = new Reference<>();
+        withContextClassLoader(deploymentClassLoader.getClassLoader(), () -> {
+          try {
+            artifactContext.getMuleContext().stop();
+          } catch (MuleException e) {
+            error.set(e);
+          }
+        });
+        if (error.get() != null) {
+          throw error.get();
```
elrodro83: do you need this? `withContextClassLoader` already wraps and throws any exception

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/domain/DefaultMuleDomain.java
```diff
@@ -232,9 +244,21 @@ public void dispose() {
     if (logger.isInfoEnabled()) {
       logger.info(miniSplash(format("Disposing domain '%s'", getArtifactName())));
     }
+
+    ClassLoader originalClassloader = currentThread().getContextClassLoader();
+
+    if (originalClassloader != deploymentClassLoader.getClassLoader()) {
```
elrodro83: you may use `withContextClassLoader` here too, extracting the below code to a lambda

PR #6368, opened 2017-12-29
MULE-14346: Support lazyInit for Domains

gsfernandes:


modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/descriptor/AbstractArtifactDescriptorFactory.java
```diff
@@ -60,15 +60,23 @@ public AbstractArtifactDescriptorFactory(DescriptorLoaderRepository descriptorLo
 
   @Override
   public T create(File artifactFolder, Optional<Properties> deploymentProperties) throws ArtifactDescriptorCreateException {
+    final M artifactModel = createArtifactModel(artifactFolder);
+    return createArtifact(artifactFolder, deploymentProperties, artifactModel);
+  }
+
+  public T createArtifact(File artifactFolder, Optional<Properties> deploymentProperties, M artifactModel) {
```
pablolagreca: why should this method be public and why does it takes an articact model when the artifact model depends on the implementation class?

PR #6373, opened 2018-01-02
MULE-14356: Fix Policy context is sharing SecurityManager with App's

balbifm:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/SpringMuleContextServiceConfigurator.java
```diff
@@ -262,7 +263,8 @@ void createArtifactServices() {
     loadServiceConfigurators();
 
     defaultContextServices.entrySet().stream()
-        .filter(service -> !APPLICATION_ONLY_SERVICES.contains(service.getKey()) || artifactType.equals(APP))
+        .filter(service -> !APPLICATION_ONLY_SERVICES.contains(service.getKey()) || artifactType.equals(APP)
```
pablokraan: Add a test

balbifm: Test added

PR #6376, opened 2018-01-02
 MULE-14357: ClassCastException in FileConnector restart

facundovs:


transports/file/src/test/java/org/mule/transport/file/FileAgeSetUpTestCase.java
```diff
@@ -0,0 +1,90 @@
...
+    protected void doSetUp() throws Exception
+    {
+        mockServiceDescriptor();
+        connector = createConnector();
+        muleContext.start();
+        connector.initialise();
+    }
+
+    @Test
+    public void getFileAgeInRestart() throws Exception
```
pablokraan: It is possible to simulate a restart instead of using createReceiver?

PR #6378, opened 2018-01-02
MULE-14358: APIKit needs access to MessageProcessors class

dfmacias:


dfmacias:
APIKit needs this change after all

PR #6385, opened 2018-01-03
MULE-14202: Tooling services doesn't work if operation contains a Lis…

alegmarra:
…t as parameter with reference support

### DIFF
--- a/modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/object/DefaultValueResolverParsingDelegate.java
+++ b/modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/object/DefaultValueResolverParsingDelegate.java
@@ -6,10 +6,11 @@
  */
 package org.mule.runtime.module.extension.internal.config.dsl.object;
 
+import static org.mule.runtime.core.api.util.StringUtils.isBlank;
 import org.mule.metadata.api.model.MetadataType;
-import org.mule.runtime.core.api.MuleContext;
 import org.mule.runtime.extension.api.dsl.syntax.DslElementSyntax;
 import org.mule.runtime.module.extension.internal.runtime.resolver.RegistryLookupValueResolver;
+import org.mule.runtime.module.extension.internal.runtime.resolver.StaticValueResolver;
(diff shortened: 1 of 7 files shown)


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/object/DefaultValueResolverParsingDelegate.java
```diff
@@ -32,11 +33,11 @@ public boolean accepts(MetadataType metadataType) {
...
    * @param key the parsed entity key
    * @param metadataType a {@link MetadataType}
-   * @param dslElementResolver
+   * @param elementDsl the {@link DslElementSyntax} of the parsed element
    * @return A {@link ValueResolver} that performs a registry lookup using the given {@code key}
    */
   @Override
   public ValueResolver<Object> parse(String key, MetadataType metadataType, DslElementSyntax elementDsl) {
-    return new RegistryLookupValueResolver<>(key);
+    return isBlank(key) ? new StaticValueResolver<>(key) : new RegistryLookupValueResolver(key);
```
marianogonzalez: this should only apply if in tooling mode. If this happens in runtime then it's a bug and should blow up with a meaningful error

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/object/DefaultValueResolverParsingDelegate.java
```diff
@@ -32,11 +31,11 @@ public boolean accepts(MetadataType metadataType) {
...
    * @param key the parsed entity key
    * @param metadataType a {@link MetadataType}
-   * @param dslElementResolver
+   * @param elementDsl the {@link DslElementSyntax} of the parsed element
    * @return A {@link ValueResolver} that performs a registry lookup using the given {@code key}
    */
   @Override
   public ValueResolver<Object> parse(String key, MetadataType metadataType, DslElementSyntax elementDsl) {
-    return new RegistryLookupValueResolver<>(key);
+    return new RegistryLookupValueResolver(key);
```
marianogonzalez: undo

PR #6388, opened 2018-01-03
MULE-14310: Empty configuration name causes extension to fail

ndinu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/ConfigModelLoaderDelegate.java
```diff
@@ -58,7 +59,11 @@ private void declareConfiguration(ExtensionDeclarer declarer, ExtensionElement e
     Optional<Configuration> configurationAnnotation = configType.getAnnotation(Configuration.class);
     if (configurationAnnotation.isPresent()) {
       final Configuration configuration = configurationAnnotation.get();
-      configurationDeclarer = declarer.withConfig(configuration.name());
+      if (isEmpty(configuration.name())) {
```
alegmarra: isBlank, that way you cover `"    "` too

alegmarra: you can simplify this as:
String configName = isBlank(configuration.name()) ? DEFAULT_CONFIG_NAME : configuration.name();
configurationDeclarer = declarer.withConfig(configName);

PR #6402, opened 2018-01-08
MULE-14394 Notification listener exceptions should not be propagated …

dfeist:
…to runtime

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/context/notification/ServerNotificationHandler.java
+++ b/core/src/main/java/org/mule/runtime/core/api/context/notification/ServerNotificationHandler.java
@@ -11,6 +11,12 @@ import org.mule.runtime.api.notification.NotificationListener;
 
 public interface ServerNotificationHandler {
 
+  /**
+   * Fire the {@link Notification}. Regardless of is a notification is fired synchronously or asynchronously any {@link Throwable}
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mule/runtime/core/api/context/notification/ServerNotificationHandler.java
```diff
@@ -11,6 +11,12 @@
 
 public interface ServerNotificationHandler {
 
+  /**
+   * Fire the {@link Notification}. Regardless of is a notification is fired synchronously or asynchronously any {@link Throwable}
```
pablokraan: Regardless of is a notification?

core/src/main/java/org/mule/runtime/core/internal/context/notification/Sender.java
```diff
@@ -26,8 +32,10 @@ public void dispatch(N notification, NotifierCallback notifier) {
     if (pair.getSelector().test(notification)) {
       try {
         notifier.notify(pair.getListener(), notification);
-      } catch (Exception e) {
-        // Exceptions from listeners do not affect the notification processing
+      } catch (Throwable e) {
+        // Exceptions or errorrs from listeners do not affect the notification processing
```
pablokraan: errorrs

PR #6404, opened 2018-01-08
MULE-14311: Make ExtensionResourcesGeneratorAnnotationProcessor work …

estebanwasinger:
…in compile time


modules/extensions-spring-support/src/test/resources/models/metadata.json
```diff
@@ -11978,7 +11977,8 @@
               "parent": "",
               "genericTypes": [],
               "isMap": false
-            }
+            },
+            "label": "Person"
```
estebanwasinger: I'm reviewing why the order changed

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/AbstractModelLoaderDelegate.java
```diff
@@ -43,6 +47,15 @@
     this.loader = loader;
   }
 
+  protected ClassLoader getExtensionClassLoader() {
+    Class<?> extensionType = getExtensionType();
+    if (extensionType != null) {
+      return extensionType.getClassLoader();
+    } else {
+      return ExtensionModel.class.getClassLoader();
```
alegmarra: TCCL

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/InjectedFieldsModelValidator.java
```diff
@@ -102,7 +102,8 @@ public void visitObject(ObjectType objectType) {
                       validateType(model, type, DefaultEncoding.class);
                     }
                   } catch (Exception e) {
-                    problemsReporter.addWarning(new Problem(model, "Could not validate Class: " + e.getMessage()));
+                    //TODO - This is too verbose and fails always in AST Mode
```
alegmarra: ticket? no way to flag when we are in AST mode?

estebanwasinger: I will create a ticket

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/AbstractModelLoaderDelegate.java
estebanwasinger: I Can't 😞 , If I do that I will get the Maven Plugin Classloader which doesn't have visibility to some Mule Classes.

alegmarra: Override this method on each implementation, and by default in the abstract return the TCC

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/InjectedFieldsModelValidator.java
```diff
@@ -61,113 +62,119 @@
...
+                @Override
+                public void visitObject(ObjectType objectType) {
+                  if (!objectType.getAnnotation(InfrastructureTypeAnnotation.class).isPresent()) {
+                    try {
+                      Class<?> type = getType(objectType, classLoaderModelProperty.getClassLoader());
+                      if (validatedTypes.add(type)) {
+                        validateType(model, type, DefaultEncoding.class);
+                      }
+                    } catch (Exception e) {
+                      //TODO - This is too verbose and fails always in AST Mode
```
alegmarra: 🤔  Still wrong

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/AbstractModelLoaderDelegate.java
estebanwasinger: Implementations of this class are: `OperationModelLoaderDelegate`, `ParameterModelLoaderDelegate`, etc, which doesn't make sense that each one of this have to redefine these

PR #6411, opened 2018-01-08
MULE-13025: Enable Test Connectivity on SOAP connectors

juandesi:


modules/extensions-soap-support/src/main/java/org/mule/runtime/module/extension/soap/internal/runtime/connection/ForwardingSoapClientConnectionProvider.java
```diff
@@ -90,19 +91,21 @@ public void disconnect(ForwardingSoapClient connection) {
     connection.disconnect();
   }
 
-  // TODO: MULE-12207 connectivity testing for Soap Connect extensions.
   @Override
   public ConnectionValidationResult validate(ForwardingSoapClient connection) {
-    return success();
+    try {
+      serviceProvider.validateConfiguration();
```
marianogonzalez: isn't this something the SDK can automaticaly always invoke?

PR #6422, opened 2018-01-09
MULE-9669: Added a toString to DefaultHttpListenerConfig to detect the port used if the listener fails to start

facunlk:
Added a toString to DefaultHttpListenerConfig to detect the port used if the listener fails to start

### DIFF
--- a/modules/http/src/main/java/org/mule/module/http/internal/listener/DefaultHttpListenerConfig.java
+++ b/modules/http/src/main/java/org/mule/module/http/internal/listener/DefaultHttpListenerConfig.java
@@ -338,4 +338,15 @@ public class DefaultHttpListenerConfig extends AbstractAnnotatedObject implement
     {
         this.connectionIdleTimeout = connectionIdleTimeout;
     }
+    
+    @Override
+    public String toString() {
+        StringBuilder buf = new StringBuilder();
+        buf.append("org.mule.module.http.internal.listener.DefaultHttpListenerConfig{name=")
+           .append(name)
+           .append(", url=")
+           .append(listenerUrl())
+           .append("}");
+        return buf.toString();
+    }
 }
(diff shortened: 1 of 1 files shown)


modules/http/src/main/java/org/mule/module/http/internal/listener/DefaultHttpListenerConfig.java
```diff
@@ -338,4 +338,15 @@ public void setConnectionIdleTimeout(int connectionIdleTimeout)
     {
         this.connectionIdleTimeout = connectionIdleTimeout;
     }
+    
+    @Override
+    public String toString() {
+        StringBuilder buf = new StringBuilder();
+        buf.append("org.mule.module.http.internal.listener.DefaultHttpListenerConfig{name=")
```
afelisatti: I don't think the full class name is necessary.

facunlk: I sent a change to remove the package and leave only the classname.

modules/http/src/main/java/org/mule/module/http/internal/listener/DefaultHttpListenerConfig.java
```diff
@@ -338,4 +338,15 @@ public void setConnectionIdleTimeout(int connectionIdleTimeout)
     {
         this.connectionIdleTimeout = connectionIdleTimeout;
     }
+    
+    @Override
+    public String toString() {
+        StringBuilder buf = new StringBuilder();
+        buf.append("DefaultHttpListenerConfig{name=")
```
elrodro83: use `getClass().getSimpleName()` instead of hardcoding the class name

afelisatti:
Thanks, @facunlk!

PR #6429, opened 2018-01-10
MULE-14111: Add support to indicate whether an error is handleable or not in extension model

ndinu:


estebanwasinger:
but also check what ana said

PR #6435, opened 2018-01-10
MULE-13792: Soap Service: add flag to log request/response

juandesi:


modules/soap-api/src/main/java/org/mule/runtime/soap/api/message/dispatcher/DefaultHttpMessageDispatcher.java
```diff
@@ -48,25 +57,40 @@ public DefaultHttpMessageDispatcher(HttpClient client) {
...
       throw new DispatchingException("An error occurred while sending the SOAP request");
     } catch (TimeoutException e) {
       throw new DispatchingException("The SOAP request timed out", e);
     }
   }
 
+  /**
+   * Logs the content if it's log enabled.
+   */
+  private InputStream retrieveLogging(String title, InputStream content) {
```
alegmarra: why `retrieve`? this is not returning something different. Just `log` or `doLog` is much clearer

modules/soap-api/src/main/java/org/mule/runtime/soap/api/message/dispatcher/DefaultHttpMessageDispatcher.java
```diff
@@ -48,25 +57,40 @@ public DefaultHttpMessageDispatcher(HttpClient client) {
...
+    if (LOGGER.isDebugEnabled()) {
+      String c = IOUtils.toString(content);
+      LOGGER.debug("Logging " + title);
```
marianogonzalez: Logging doesn't mean anything. Add more context to identify the request. THink of a highly concurrent scenario, you'll need to trace each message to an event or request id

modules/soap-api/src/main/java/org/mule/runtime/soap/api/message/dispatcher/DefaultHttpMessageDispatcher.java
juandesi: if we consume the Stream while logging we need to return a new one.

modules/soap-api/src/main/java/org/mule/runtime/soap/api/message/dispatcher/DefaultHttpMessageDispatcher.java
juandesi: makes sense, let me find a way to identify requests

alegmarra: does the context headers have some kind of correlationId?

modules/soap-api/src/main/java/org/mule/runtime/soap/api/message/dispatcher/DefaultHttpMessageDispatcher.java
alegmarra: is still the same content, you just read it for logging

PR #6455, opened 2018-01-15
MULE-14467: Adding trace logs for request life cycle

santiagopurucker:
Depends on PR https://github.com/mulesoft/mule-http-connector/pull/185

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/policy/CompositeOperationPolicy.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/policy/CompositeOperationPolicy.java
@@ -9,10 +9,12 @@ package org.mule.runtime.core.internal.policy;
 import static java.util.Optional.empty;
 import static org.mule.runtime.core.privileged.processor.MessageProcessors.processToApply;
 import static org.mule.runtime.core.privileged.processor.MessageProcessors.processWithChildContext;
+import static org.slf4j.LoggerFactory.getLogger;
 import static reactor.core.publisher.Mono.error;
 import static reactor.core.publisher.Mono.from;
 import static reactor.core.publisher.Mono.just;
 
+import org.mule.runtime.api.event.EventContext;
 import org.mule.runtime.api.exception.MuleException;
 import org.mule.runtime.api.message.Message;
 import org.mule.runtime.core.api.event.CoreEvent;
@@ -20,12 +22,14 @@ import org.mule.runtime.core.api.policy.OperationPolicyParametersTransformer;
 import org.mule.runtime.core.api.policy.Policy;
 import org.mule.runtime.core.api.processor.Processor;
 
+import org.reactivestreams.Publisher;
+import org.slf4j.Logger;
+
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
-import org.reactivestreams.Publisher;
 
 /**
  * {@link OperationPolicy} created from a list of {@link Policy}.
@@ -36,6 +40,8 @@ import org.reactivestreams.Publisher;
  */
 public class CompositeOperationPolicy extends
     AbstractCompositePolicy<OperationPolicyParametersTransformer, OperationParametersProcessor> implements OperationPolicy {
+
+  private static final Logger LOGGER = getLogger(CompositeOperationPolicy.class);
 
 
   private final Processor nextOperation;
@@ -97,7 +103,11 @@ public class CompositeOperationPolicy extends
    */
   @Override
   protected Publisher<CoreEvent> processNextOperation(CoreEvent event) {
-    return just(event).transform(nextOperation).doOnNext(response -> this.nextOperationResponse = response);
+    logOperationEvent(event.getContext(), event.getMessage());
+    return just(event).transform(nextOperation).doOnNext(response -> {
+      this.nextOperationResponse = response;
+      logOperationEvent(response.getContext(), response.getMessage());
+    });
   }
 
   /**
@@ -138,6 +148,13 @@ public class CompositeOperationPolicy extends
                                      empty());
     } catch (Exception e) {
       return error(e);
+    }
+  }
+
+  private void logOperationEvent(EventContext eventContext, Message message) {
+    if (LOGGER.isTraceEnabled()) {
+      LOGGER.trace("Executing operation\n" + eventContext.toString() + "\n"
+          + message.getAttributes().getValue().toString());
     }
   }
 }
--- a/core/src/main/java/org/mule/runtime/core/internal/policy/CompositeOperationPolicy.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/policy/CompositeOperationPolicy.java
@@ -9,12 +9,10 @@ package org.mule.runtime.core.internal.policy;
 import static java.util.Optional.empty;
 import static org.mule.runtime.core.privileged.processor.MessageProcessors.processToApply;
 import static org.mule.runtime.core.privileged.processor.MessageProcessors.processWithChildContext;
-import static org.slf4j.LoggerFactory.getLogger;
 import static reactor.core.publisher.Mono.error;
 import static reactor.core.publisher.Mono.from;
 import static reactor.core.publisher.Mono.just;
 
-import org.mule.runtime.api.event.EventContext;
 import org.mule.runtime.api.exception.MuleException;
 import org.mule.runtime.api.message.Message;
 import org.mule.runtime.core.api.event.CoreEvent;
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/mule/runtime/core/internal/policy/CompositeOperationPolicy.java
```diff
@@ -97,8 +102,10 @@ public CoreEvent process(CoreEvent event) throws MuleException {
    */
   @Override
   protected Publisher<CoreEvent> processNextOperation(CoreEvent event) {
+    logOperationEvent(event.getContext(), event.getMessage());
```
elrodro83: this "before" log should be done inside the publisher chain returned by this method

core/src/main/java/org/mule/runtime/core/internal/policy/CompositeSourcePolicy.java
```diff
@@ -125,7 +125,10 @@ public CompositeSourcePolicy(List<Policy> parameterizedPolicies,
    */
   @Override
   protected Publisher<CoreEvent> processPolicy(Policy policy, Processor nextProcessor, CoreEvent event) {
-    return just(event).transform(sourcePolicyProcessorFactory.createSourcePolicy(policy, nextProcessor));
+    logPolicy(getCoreEventId(event), getPolicyName(policy), getCoreEventAttributesAsString(event));
```
elrodro83: this "before" log should be done inside the publisher chain returned by this method

core/src/main/java/org/mule/runtime/core/internal/policy/SourcePolicyFailureResult.java
```diff
@@ -31,6 +37,7 @@
   public SourcePolicyFailureResult(MessagingException messagingException, Supplier<Map<String, Object>> errorResponseParameters) {
     this.messagingException = messagingException;
     this.errorResponseParameters = errorResponseParameters;
+    logSourcePolicyResult();
```
elrodro83: log this outside this constructor, when this object is used

core/src/main/java/org/mule/runtime/core/internal/policy/SourcePolicySuccessResult.java
```diff
@@ -38,6 +44,7 @@ public SourcePolicySuccessResult(CoreEvent result, Supplier<Map<String, Object>>
     this.result = result;
     this.responseParameters = responseParameters;
     this.messageSourceResponseParametersProcessor = messageSourceResponseParametersProcessor;
+    logSourcePolicyResult();
```
elrodro83: log this outside this constructor, when this object is used

core/src/main/java/org/mule/runtime/core/internal/policy/CompositeOperationPolicy.java
```diff
@@ -130,4 +137,11 @@ public CoreEvent process(CoreEvent event) throws MuleException {
       return error(e);
     }
   }
+
+  private void logOperationEvent(EventContext eventContext, Message message) {
+    if (LOGGER.isTraceEnabled()) {
+      LOGGER.trace("Executing operation\n" + eventContext.toString() + "\n"
```
elrodro83: do you need to add the context id? it should be included as part of the context information logged in every line. Same applies in other parts of this PR.

elrodro83: Change `Executing operation` for something indicating it it's about to execute or already executed, since that seems to be the use case.

core/src/main/java/org/mule/runtime/core/internal/policy/CompositeOperationPolicy.java
```diff
@@ -130,4 +137,11 @@ public CoreEvent process(CoreEvent event) throws MuleException {
...
+          + message.getAttributes().getValue().toString());
```
elrodro83: is the attributes `toString` worth logging? it is not implemented to show any meaningful data. Maybe the component path of the operation being executed provided more meaningful info.

core/src/main/java/org/mule/runtime/core/internal/policy/CompositeSourcePolicy.java
```diff
@@ -125,7 +125,11 @@ public CompositeSourcePolicy(List<Policy> parameterizedPolicies,
    */
   @Override
   protected Publisher<CoreEvent> processPolicy(Policy policy, Processor nextProcessor, CoreEvent event) {
-    return just(event).transform(sourcePolicyProcessorFactory.createSourcePolicy(policy, nextProcessor));
+    return just(event)
+        .doOnNext(s -> logPolicy(getCoreEventId(event), getPolicyName(policy), getCoreEventAttributesAsString(event)))
```
balbifm: Should have different messages for the start and the end of the policy execution. Right now is always just showing `Processing`

core/src/main/java/org/mule/runtime/core/internal/policy/DefaultPolicyManager.java
```diff
@@ -70,6 +77,9 @@
   public SourcePolicy createSourcePolicyInstance(Component source, CoreEvent sourceEvent,
                                                  Processor flowExecutionProcessor,
                                                  MessageSourceResponseParametersProcessor messageSourceResponseParametersProcessor) {
+
+    logEvent(sourceEvent.getContext(), sourceEvent.getMessage());
```
balbifm: This seems that is going to log the same that the first policy when it starts. Also, this won't be just for policies. We may have to remove it

core/src/main/java/org/mule/runtime/core/internal/policy/CompositeOperationPolicy.java
```diff
@@ -140,4 +150,11 @@ public CoreEvent process(CoreEvent event) throws MuleException {
       return error(e);
     }
   }
+
+  private void logOperationEvent(String startingMessage, EventContext eventContext, Message message) {
+    if (LOGGER.isTraceEnabled()) {
```
balbifm: The policy name is not being logged anywhere for operation policies. Why don't you put the logs in the same place where it is in the CompositeSourcePolicy and with the same format?

core/src/main/java/org/mule/runtime/core/internal/policy/CompositeSourcePolicy.java
```diff
@@ -180,4 +188,37 @@ public CompositeSourcePolicy(List<Policy> parameterizedPolicies,
...
+      LOGGER.trace("Event Id: " + eventId + ".\n Processing " + policyName + "\n" + message);
+    }
+  }
+
+  private String getCoreEventId(CoreEvent event) {
+    return event.getContext().getId();
+  }
+
+  private String getCoreEventAttributesAsString(CoreEvent event) {
+    return event.getMessage().getAttributes().getValue().toString();
```
balbifm: null is a valid attributes value as far as I know, so this might explode with NullPointerException

core/src/main/java/org/mule/runtime/core/internal/policy/CompositeSourcePolicy.java
```diff
@@ -180,4 +188,37 @@ public CompositeSourcePolicy(List<Policy> parameterizedPolicies,
...
+  }
+
+  private String getPolicyName(Policy policy) {
+    return policy.getPolicyChain().getRootContainerLocation().getGlobalName();
+  }
+
+  private void logSourcePolicySuccessfullResult(SourcePolicySuccessResult result) {
+    if (LOGGER.isTraceEnabled()) {
+      LOGGER.trace("Event id: " + result.getResult().getContext().getId() + " finished processing. \n" +
+          result.getResult().getMessage().getAttributes().getValue().toString());
```
balbifm: Same, possible NPE here

core/src/main/java/org/mule/runtime/core/internal/policy/DefaultPolicyManager.java
```diff
@@ -80,13 +90,18 @@ public SourcePolicy createSourcePolicyInstance(Component source, CoreEvent sourc
                                                                                                                                             .getSuccessfulExecutionResponseParametersFunction()
                                                                                                                                             .apply(flowExecutionResult),
                                                                                                                                         messageSourceResponseParametersProcessor)))
+          .doOnNext(result -> logSourcePolicySuccessfullResult(result.getRight()))
```
balbifm: Same thing here... you are already logging at the end of the CompositeSourcePolicy, these logging here seems redundant

core/src/main/java/org/mule/runtime/core/internal/policy/CompositeOperationPolicy.java
santiagopurucker: eventContext is the only object that has the information that we need to log, and the only method that expose them is toString(). It prints it in a slightly different format than we do, but as the method was already defined I didn't dare to modify it

balbifm: But why log where you are logging and not where I am saying? It is not useful to log that if you don't know in which policy you are

PR #6462, opened 2018-01-15
MULE-12798: Allow parameter groups repeat Display Name

ndinu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/ParameterModelsLoaderDelegate.java
```diff
@@ -254,6 +256,12 @@ private void parseConfigOverride(ExtensionParameter extensionParameter, Paramete
 
     declarer.withDslInlineRepresentation(groupAnnotation.showInDsl());
 
+    DisplayName displayName = groupParameter.getAnnotation(DisplayName.class).orElse(null);
+
+    if (displayName != null) {
```
estebanwasinger: This can be replaced with `groupParameter.getAnnotation(xxxxxx.class).ifPresent(displayName -> <logic-to-build-display-model>);`

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/HeisenbergOperations.java
```diff
@@ -176,8 +176,9 @@ public String transform(String transformation) {
     return transformation;
   }
 
-  public void disguice(@ParameterGroup(name = "currentLook") BarberPreferences currentLook,
-                       @ParameterGroup(name = "disguise", showInDsl = true) BarberPreferences disguise) {
+  public void disguice(@ParameterGroup(name = "currentLook") @DisplayName("Current look") BarberPreferences currentLook,
```
estebanwasinger: The root cause of this ticket was to allow repeat param group display name, but I didn't see any repeated displayname testing if that works correctly.

PR #6519, opened 2018-01-20
MULE-14518: Skip NullsafeWrapper at runtime if parameter defaults can…

alegmarra:
…not be instantiated

### DIFF
--- a/modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/resources/BaseExtensionResourcesGeneratorAnnotationProcessor.java
+++ b/modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/resources/BaseExtensionResourcesGeneratorAnnotationProcessor.java
@@ -16,7 +16,6 @@ import static org.mule.runtime.core.api.util.ClassUtils.withContextClassLoader;
 import static org.mule.runtime.core.api.util.ExceptionUtils.extractOfType;
 import static org.mule.runtime.module.extension.api.loader.AbstractJavaExtensionModelLoader.TYPE_PROPERTY_NAME;
 import static org.mule.runtime.module.extension.api.loader.AbstractJavaExtensionModelLoader.VERSION;
-
 import org.mule.runtime.api.exception.MuleRuntimeException;
 import org.mule.runtime.api.meta.model.ExtensionModel;
 import org.mule.runtime.core.api.registry.SpiServiceRegistry;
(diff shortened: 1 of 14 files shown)


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/resources/BaseExtensionResourcesGeneratorAnnotationProcessor.java
```diff
@@ -159,6 +166,11 @@ private String getVersion(Name qualifiedName) {
 
   }
 
+  private boolean simulateRuntimeLoading() {
```
estebanwasinger: I know what It means, but add a brief JDoc explaining why we are doing this.

estebanwasinger: I know what It means, but add a brief JDoc explaining why we are doing this.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/DefaultJavaModelLoaderDelegate.java
```diff
@@ -106,6 +106,10 @@ public ExtensionDeclarer declare(ExtensionLoadingContext context) {
             .withCategory(extensionElement.getCategory())
             .withModelProperty(new ExtensionTypeDescriptorModelProperty(extensionElement));
 
+    // TODO MULE-14517: This workaround should be replaced for a better and more complete mechanism
+    context.getParameter("COMPILATION_MODE")
```
estebanwasinger: Use the constant defined in `BaseExtensionResourcesGeneratorAnnotationProcessor.java`

alegmarra: can't, BaseExtensionResourcesGeneratorAnnotationProcessor is ext-spring-support, this is ext-support. I could declare the constant here, but then I'd have to make it API and I prefer to keep it all internal since this is a hack and will be removed in 4.2

estebanwasinger: sorry I got confused. dismiss it.

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/resources/BaseExtensionResourcesGeneratorAnnotationProcessor.java
estebanwasinger: Add it later

PR #6520, opened 2018-01-21
MULE-14519: HTTP: Avoid multiple creations of the same headers map

elrodro83:
- Prepare API

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/util/ClassUtils.java
+++ b/core/src/main/java/org/mule/runtime/core/api/util/ClassUtils.java
@@ -14,7 +14,6 @@ import static org.mule.runtime.api.util.Preconditions.checkArgument;
 import static org.mule.runtime.core.api.util.ExceptionUtils.tryExpecting;
 
 import org.mule.runtime.api.exception.MuleRuntimeException;
-import org.mule.runtime.core.internal.util.FunctionalUtils;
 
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/mule/runtime/core/api/util/ClassUtils.java
```diff
@@ -860,7 +861,10 @@ public static void withContextClassLoader(ClassLoader classLoader, Runnable runn
    * @param f the function to memoize
    * @param cache the map where cached values are stored
    * @return the memoized function
+   *
+   * @deprecated use {@link FunctionalUtils#memoize(Function, Map)}
```
dfeist: ClassUtils is API, and FunctionUtils isn't, so this isn't really a valid deprecation message is it?

modules/http-api/src/main/java/org/mule/runtime/http/api/domain/message/HttpMessage.java
```diff
@@ -48,6 +49,11 @@
    */
   Collection<String> getHeaderValuesIgnoreCase(String headerName);
 
+  /**
+   * @return all headers
+   */
+  MultiMap<String, String> getHeaders();
```
dfeist: Is this mutable or immutable? Maybe good idea to be clear about this in JDoc too.

PR #6537, opened 2018-01-24
MULE-14495: Move static caches form IntrospectionUtils to an object in the app registry

elrodro83:
Fix classloader leak from cache

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/InputParametersTypeModelValidator.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/InputParametersTypeModelValidator.java
@@ -14,6 +14,7 @@ import static org.mule.metadata.api.model.MetadataFormat.JAVA;
 import static org.mule.runtime.extension.api.util.ExtensionMetadataTypeUtils.getType;
 import static org.mule.runtime.extension.api.util.NameUtils.getComponentModelTypeName;
 import static org.mule.runtime.module.extension.internal.util.IntrospectionUtils.getFieldsWithGetters;
+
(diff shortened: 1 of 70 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/InputParametersTypeModelValidator.java
```diff
@@ -53,6 +55,8 @@
  */
 public final class InputParametersTypeModelValidator implements ExtensionModelValidator {
 
+  private ReflectionCache reflectionCache = new ReflectionCache();
```
alegmarra: we avoid state in validators and loaders since they're never disposed. check this doesn't generate a new leak and if it cannot be parameterized instead

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ParameterGroupArgumentResolver.java
```diff
@@ -8,20 +8,25 @@
...
 
   private final ParameterGroupDescriptor group;
 
-  public ParameterGroupArgumentResolver(ParameterGroupDescriptor group) {
+  private final ReflectionCache reflectionCache;
+
+  public ParameterGroupArgumentResolver(ParameterGroupDescriptor group, ReflectionCache reflectionCache) {
     checkState(group.getType().isInstantiable(), "Class %s cannot be instantiated.");
     this.group = group;
+    this.reflectionCache = reflectionCache;
```
alegmarra: value resolver should support the @Inject mechanism to receive the cache, maybe that simplifies the logic instead of having the parameter

PR #6538, opened 2018-01-24
MULE-14495: Move static caches form IntrospectionUtils to an object in

elrodro83:
the app registry
* Addendum


elrodro83:
Apply feedback from https://github.com/mulesoft/mule/pull/6537

PR #6543, opened 2018-01-26
MULE-14131: No parser found for compatibility component if declared in domain

lucianoRM:


modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/internal/application/MuleApplicationClassLoader.java
```diff
@@ -58,14 +60,36 @@ protected String findLibrary(String name) {
...
+
+  /**
+   * Resolves the plugin classloader of the thread context class loader artifact and all it's ancestor
+   */
+  public static List<ClassLoader> resolveContextArtifactPluginClassLoaders() {
+    Set<ClassLoader> resolvedClassLoaders = new HashSet<>();
+    ClassLoader originalClassLoader = currentThread().getContextClassLoader();
+    ClassLoader tmpClassLoader = originalClassLoader;
+    resolvedClassLoaders.addAll(resolveContextArtifactPluginClassLoadersForCurrentClassLoader());
+    try {
```
pablokraan: When you code, try to add empty spaces to separate different parts of the code to make it easier to read.

modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/internal/application/MuleApplicationClassLoader.java
```diff
@@ -58,14 +60,36 @@ protected String findLibrary(String name) {
...
+      while (tmpClassLoader.getParent() != null
+          && MuleDeployableArtifactClassLoader.class.isAssignableFrom(tmpClassLoader.getParent().getClass())) {
+        tmpClassLoader = tmpClassLoader.getParent();
+        currentThread().setContextClassLoader(tmpClassLoader);
+        resolvedClassLoaders.addAll(resolveContextArtifactPluginClassLoadersForCurrentClassLoader());
+      }
+    } finally {
```
pablokraan: technically, the finally block must wrap the part when the change is done. (ie: on every change)

PR #6549, opened 2018-01-26
MULE-14376 Annotated every element exposed on the API that have usage restrictions

pablokraan:


pepe.txt
```diff
@@ -0,0 +1,1092 @@
+diff --git a/core/src/main/java/org/mule/runtime/core/api/security/MuleSecurityManagerConfigurator.java b/core/src/main/java/org/mule/runtime/core/api/security/MuleSecurityManagerConfigurator.java
```
pablolagreca: delete

PR #6556, opened 2018-01-30
MULE-14376: Annotated every element exposed on the API that have usag…

pablokraan:
…e restrictions (#6480)

_ Cleaning API on more modules


pablokraan:
jenkins validate

PR #6571, opened 2018-02-01
MULE-14581 Reduce level of log for RejectedExecutionExceptions that o…

dfeist:
…ccur when shutdown timeout is exceeded.

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
@@ -247,7 +247,15 @@ public abstract class AbstractPipeline extends AbstractFlowConstruct implements 
         .doOnNext(beforeProcessors())
         .transform(processingStrategy.onPipeline(pipeline))
         .doOnNext(afterProcessors())
-        .doOnError(throwable -> LOGGER.error("Unhandled exception in Flow ", throwable));
+        .doOnError(throwable -> {
+          if (throwable instanceof RejectedExecutionException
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
```diff
@@ -247,7 +247,15 @@ protected ReactiveProcessor processFlowFunction() {
         .doOnNext(beforeProcessors())
         .transform(processingStrategy.onPipeline(pipeline))
         .doOnNext(afterProcessors())
-        .doOnError(throwable -> LOGGER.error("Unhandled exception in Flow ", throwable));
+        .doOnError(throwable -> {
+          if (throwable instanceof RejectedExecutionException
+              && throwable.getStackTrace()[3].getMethodName().contains("onComplete")) {
```
elrodro83: i don't like this magic `3` here. Any update in reactor may change that.

core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
```diff
@@ -247,7 +247,25 @@ protected ReactiveProcessor processFlowFunction() {
...
+          } else {
+            LOGGER.error("Unhandled exception in Flow ", throwable);
+          }
+        });
+  }
+
+  boolean isCompleteSignalRejectedExecutionException(Throwable throwable) {
+    if (throwable instanceof RejectedExecutionException) {
+      for (StackTraceElement element : throwable.getStackTrace()) {
+        if (element.getMethodName().contains("onComplete")) {
```
elrodro83: can you also check the package/class?

PR #6578, opened 2018-02-14
MULE-14601: Json Schema Validator treats duplicate keys in an inconsistent way depending on the input type.

facundovs:


distributions/standalone/src/main/resources/MIGRATION.txt
```diff
@@ -7,6 +7,7 @@ Please check https://developer.mulesoft.com/docs/display/current/Mule+ESB for mo
 Migration changes from 3.9.x to 3.10.x
 MULE-14381: By default, the File Connector processes empty files. For disabling this behavior use the mule.transport.file.ignoreEmptyFiles=true system property.
 MULE-14507: The HTTP Requester now sends a zero content length in empty payloads scenarios. For disabling this behavior use the mule.http.client.avoidZeroContentLength=true system property.
+MULE-14601: The JSON Schema Validator now fails if it receives a payload with duplicate keys. For allowing duplicate keys use the mule.json.validator.allowDuplicateKeys=true system property.
```
marianogonzalez: Why did you choose to disallow duplicate keys as a default? Does the Json schema spec have the concept of allowing/disallowing duplicate keys? If the schema doesn't forbid it by default, then the validator shouldn't either.  It's weird to me that you had to add that on the parser level instead of the validator itself. I want to be sure this is the right move, instead of simply adding a new parameter to the validator object to enable this strict check.

facundovs: AWAIK, Json Schema Validator works with valid json objects (in our code it receives a JsonNode that was already parsed). I've investigated the json schema api that we use and it only provides the STRICT_DUPLICATE_DETECTION feature to its JsonParser, so no way to configure it for the validator. Regarding the default behavior, the Json specification doesn't prohibit the duplicate keys, though it states they SHOULD be avoided. This specification also mentions the possible duplicate-keys treatments: 1) to remain only the last value, 2) to fail, 3)  to remain all the keys. With this change, by default we take the second approach, and applying the system property, the third one. I'll discard the first approach because it entails data loss.

marianogonzalez: if the spec doesn't' disallowed it, then you cannot change this. You would be breaking backwards compatibility, since some json documents currently considered valid will suddenly start to fail. The validator cannot fail on a recommendation. You should add an optional parameter which specifies if this should validated or not. By default, the current behavior must be kept

modules/json/src/test/java/org/mule/module/json/validation/ValidateSchemaDuplicateKeysTestCase.java
```diff
@@ -0,0 +1,79 @@
...
+    @Rule
+    public SystemProperty duplicateKeys;
+
+    @Rule
+    public ExpectedException expectedException;
+
+    public ValidateSchemaDuplicateKeysTestCase(String allowDuplicateKeys, ExpectedException expectedException)
+    {
+        this.expectedException = expectedException;
+        duplicateKeys = new SystemProperty("allowDuplicateKeys", allowDuplicateKeys);
```
marianogonzalez: why is this still here? Why is the SP still relevant? If so, you should still have a constant and there should be an entry in the release notes.

modules/json/src/main/resources/META-INF/mule-json.xsd
```diff
@@ -232,6 +232,13 @@
                         </xsd:documentation>
                     </xsd:annotation>
                 </xsd:attribute>
+                <xsd:attribute name="allowDuplicateKeys" type="mule:substitutableBoolean" default="true">
+                    <xsd:annotation>
+                        <xsd:documentation>
+                            Defines if the validator must allow duplicate keys. Default value is true.
```
marianogonzalez: remember to sync with the Studio team so that they add support for this

modules/json/src/test/java/org/mule/module/json/validation/ValidateSchemaDuplicateKeysTestCase.java
facundovs: I only remain the system property to make the test parameterized.
See https://github.com/mulesoft/mule/pull/6578/files#diff-af3298d655d6d2aca2540266785c205fR9

PR #6581, opened 2018-02-15
MULE-14605: Deadlock when two JDBC connections are created concurrent…

facundovs:
…ly and before the DriverManager classloading takes place

### DIFF
--- a/modules/db/src/main/java/org/mule/module/db/internal/domain/connection/AbstractConnectionFactory.java
+++ b/modules/db/src/main/java/org/mule/module/db/internal/domain/connection/AbstractConnectionFactory.java
@@ -18,9 +18,15 @@ import javax.sql.DataSource;
 public abstract class AbstractConnectionFactory implements ConnectionFactory
 {
 
-    protected AbstractConnectionFactory()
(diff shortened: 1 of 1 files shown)


modules/db/src/main/java/org/mule/module/db/internal/domain/connection/AbstractConnectionFactory.java
```diff
@@ -17,14 +18,25 @@
...
 
+    /**
+     * Ensures DriverManager classloading takes place before any connection creation.
+     * It prevents a JDK deadlock that only occurs when two JDBC Connections of different DB vendors
+     * are created concurrently and the {@link DriverManager} hasn't been loaded yet.
+     * For more information, see MULE-14605.
+     */
+    static
+    {
+        DriverManager.getLoginTimeout();
```
pablokraan: If I'm understanding this correctly, this static block will be executed the first time the AbstractConnectionFactory class is referenced. If there are two applications, each having their own drivers, then the problem still exists as the initialization will occur during on the first app, but not in the second.

PR #6588, opened 2018-02-19
MULE-14569: Ensure that exported resources in artifact descriptor alw…

pablokraan:
…ays use linux file separator


modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/api/DeployableArtifactDescriptor.java
```diff
@@ -72,7 +73,17 @@ public File getArtifactLocation() {
...
+    this.configResources = sanitizePaths(configResources);
+  }
+
+  private Set<String> sanitizePaths(Set<String> configResources) {
+    if (configResources == null || configResources.isEmpty()) {
+      return configResources;
+    }
+
+    Set<String> result = new HashSet<>();
+    configResources.stream().forEach(s -> result.add(separatorsToUnix(s)));
```
elrodro83: using `map` and `collect` would be a cleaner code IMO

PR #6591, opened 2018-02-21
MULE-14613: Add the java.home and java.vm.name of the JDK used by mul…

anosenzo:
…e in the splash log

### DIFF
--- a/modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainerStartupSplashScreen.java
+++ b/modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainerStartupSplashScreen.java
@@ -8,6 +8,8 @@ package org.mule.runtime.module.launcher;
 
 import static java.util.Arrays.asList;
 import static java.util.Arrays.sort;
+import static java.util.Arrays.stream;
+import static java.util.stream.Collectors.toMap;
 import static org.apache.commons.lang3.StringUtils.defaultString;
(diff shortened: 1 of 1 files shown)


modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainerStartupSplashScreen.java
```diff
@@ -104,4 +108,13 @@ private void listMuleSystemProperties() {
         .forEach(property -> muleProperties.put(property, System.getProperty(property)));
     listItems(muleProperties, "Mule system properties:");
   }
+
+  private void listJavaSystemProperties() {
+    List<String> javaPropertiesNamesToList = new ArrayList<>(Arrays.asList("java.vendor", "java.vm.name", "java.home"));
```
elrodro83: Use `Arrays.stream` or `Stream.of`

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainerStartupSplashScreen.java
```diff
@@ -104,4 +108,13 @@ private void listMuleSystemProperties() {
...
+    Map<String, String> javaProperties = new HashMap<>();
+
+    javaPropertiesNamesToList.stream()
```
elrodro83: use `collect` and avoid building a map here

PR #6593, opened 2018-02-21
MULE-14623: JMSMessageToObject causes NPE in debug mode.

facundovs:


transports/jms/src/main/java/org/mule/transport/jms/transformers/JMSMessageToObject.java
```diff
@@ -58,7 +58,7 @@ public Object transformMessage(MuleMessage message, String outputEncoding) throw
         {
             if (logger.isDebugEnabled())
             {
-                logger.debug("Source object is " + ClassUtils.getSimpleName(message.getPayload().getClass()));
+                logger.debug("Source object is " + getSimpleName(message.getPayload().getClass()));
```
pablokraan: Don't you have the same NPE here?

facundovs: I have tested it and NPE doesn't occur here since a null payload is always represented by a NullPayload instance.

PR #6599, opened 2018-02-23
MULE-14632:  WSC - Invalid property name on class java.lang.String

juandesi:


modules/soap-api/src/main/java/org/mule/runtime/soap/api/transport/ReflectiveHttpConfigBasedRequester.java
```diff
@@ -74,7 +74,9 @@ public ReflectiveHttpConfigBasedRequester(String configName, ExtensionsClient cl
     DefaultOperationParametersBuilder params = builder().configName(configName)
         .addParameter("method", method)
         .addParameter("url", url)
-        .addParameter("headers", new MultiMap<>(headers));
+        .addParameter("headers", new MultiMap<>(headers))
+        // TODO(MULE-13066): REMOVE THIS LINE WHEN DONE!
+        .addParameter("targetValue", "#[payload]");
```
marianogonzalez: this doesn't seem right according to what we discussed yesterday. What if the user specified he's own targetValue?

juandesi: the user does not specify anything,  the only thing that the user can do is specify a configuration, we take the content from the payload always

alegmarra: the user can't set the targetValue to "message". The client execution always return the Result<T, A> and in this case it will be exposed as a Pair of <payload, headers>

marianogonzalez: I still believe that it'd be better to hanle this on the getDefaultResolver() method but..........................

PR #6605, opened 2018-02-23
MULE-14601: Json Schema Validator treats duplicate keys in an inconsi…

facundovs:
…stent way depending on the input type.


facundovs:
See same pr but for 3.x [here](https://github.com/mulesoft/mule/pull/6578).

PR #6612, opened 2018-03-01
MULE-14644: Add test HTTP/HTTPS proxy

afelisatti:


afelisatti:
This component is present in HTTP and SFTP connectors, I'm adding it to unit so that it can be easily reused.

PR #6614, opened 2018-03-02
MULE-14636: HTTP Connector is not including in its type catalog the declaration for Origin type

juandesi:
…


tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/HeisenbergExtension.java
```diff
@@ -64,6 +67,7 @@
 @Sources({HeisenbergSource.class, DEARadioSource.class, AsyncHeisenbergSource.class, ReconnectableHeisenbergSource.class})
 @Export(classes = {HeisenbergExtension.class, HeisenbergException.class}, resources = "methRecipe.json")
 @SubTypeMapping(baseType = Weapon.class, subTypes = {Ricin.class})
+@SubTypeMapping(baseType = Fruit.class, subTypes = {Banana.class, Apple.class})
```
alegmarra: why is this test not impacting in heisenberg.json?

juandesi: it is jeje

PR #6616, opened 2018-03-02
MULE-14665 Race condition in reactor causes premature thread interrup…

dfeist:
…tion resulting in empty payload with repeatable streaming with no error thrown/logged.

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/streaming/bytes/AbstractInputStreamBuffer.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/streaming/bytes/AbstractInputStreamBuffer.java
@@ -152,6 +152,7 @@ public abstract class AbstractInputStreamBuffer extends AbstractStreamingBuffer 
     } catch (ClosedChannelException e) {
       if (stream instanceof FileInputStream && FileInputStream.class.equals(stream.getClass())) {
         // Delegate FileChannel is being used, handle ClosedChannelException by returning -1.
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/mule/runtime/core/internal/streaming/bytes/AbstractInputStreamBuffer.java
```diff
@@ -143,7 +150,14 @@ protected int consumeStream(ByteBuffer buffer) throws IOException {
     try {
       result = streamChannel.read(buffer);
     } catch (ClosedChannelException e) {
-      result = -1;
+      if (stream instanceof FileInputStream) {
```
marianogonzalez: cannot make these assumptions

core/src/main/java/org/mule/runtime/core/internal/streaming/bytes/AbstractInputStreamBuffer.java
```diff
@@ -143,7 +150,15 @@ protected int consumeStream(ByteBuffer buffer) throws IOException {
     try {
       result = streamChannel.read(buffer);
     } catch (ClosedChannelException e) {
-      result = -1;
+      if (stream instanceof FileInputStream && FileInputStream.class.equals(stream.getClass())) {
```
elrodro83: why these 2 conditions and not just the instanceof?

PR #6617, opened 2018-03-02
MULE-14668: Fix SFTPAutoDeleteTestCase

facundovs:


transports/sftp/src/test/java/org/mule/transport/sftp/SftpAutoDeleteTestCase.java
```diff
@@ -94,7 +96,28 @@ public void endpointAutoDeleteFalse() throws Exception
     public void endpointAutoDeleteTrue() throws Exception
     {
         testDirectory(AUTO_DELETE_ON);
-        assertThat(Arrays.asList(sftpClient.listFiles()), is(empty()));
+        Prober prober = new PollingProber(10000, 1000);
```
afelisatti: Have you verified whether such a large timeout is necessary?

PR #6620, opened 2018-03-05
MULE-14293: CXF ProxyService does not work correctly with proactor

elrodro83:
processing strategy

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/processor/InterceptingMessageProcessor.java
+++ b/core/src/main/java/org/mule/runtime/core/api/processor/InterceptingMessageProcessor.java
@@ -6,7 +6,9 @@
  */
 package org.mule.runtime.core.api.processor;
 
+import org.mule.runtime.api.interception.ProcessorInterceptorFactory;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mule/runtime/core/api/processor/InterceptingMessageProcessor.java
```diff
@@ -15,14 +17,26 @@
...
    */
   void setListener(Processor listener);
 
+  /**
+   * Indicates if the processing strategy of the parent flow is applied to the intercepted processor's chain.
+   *
+   * @return {@code true} if the {@code listener} will run wholly in the same thread and blocking if needed. {@code false} if the
+   *         lfow's {@link ProcessingStrategy} will be used.
+   */
+  default boolean isBlocking() {
```
dfeist: Consider:
- Using BlockingProcessorStrategy in child chain for CXF instead of this flag.
- If we think this may be a common scenario think about if it makes sense to represent this as a processing type.

PR #6621, opened 2018-03-05
MULE-14669: Fix Policy using a XML module that uses another plugin th…

balbifm:
…at is also present in the app is not deployed


gsfernandes:
You would need to cherry-pick this to 4.1.x branch too.

PR #6629, opened 2018-03-08
MULE-14683: Mime Type Disappears Dissapears in VM inbound when under

elrodro83:
load

### DIFF
--- a/transports/vm/src/test/java/org/mule/transport/vm/VMQueueTestCase.java
+++ b/transports/vm/src/test/java/org/mule/transport/vm/VMQueueTestCase.java
@@ -6,6 +6,7 @@
  */
 package org.mule.transport.vm;
 
+import static java.util.Arrays.asList;
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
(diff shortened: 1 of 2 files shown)


transports/vm/src/test/java/org/mule/transport/vm/VMQueueTestCase.java
```diff
@@ -95,6 +104,28 @@ public void testPassThrough() throws Exception
         }
     }
 
+    @Test
+    public void testPassThroughKeepsMimeType() throws Exception
+    {
+        MuleClient client = muleContext.getClient();
+        Set<String> polos = new HashSet<String>(Arrays.asList(new String[]{"Marco", "Niccolo", "Maffeo"}));
+        Iterator<String> people = polos.iterator();
```
fsgonz: I think a foreach loop through the array would be clearer.

transports/vm/src/test/java/org/mule/transport/vm/VMQueueTestCase.java
```diff
@@ -95,6 +104,28 @@ public void testPassThrough() throws Exception
...
+        SimpleDataType<Object> customDataType = new SimpleDataType<>(String.class, "application/custom");
+        while (people.hasNext())
+        {
+            DefaultMuleMessage message = new DefaultMuleMessage(people.next(), null, (Map<String, Object>) null, null, muleContext, customDataType);
+            client.dispatch("vm://entry", message);
+        }
+        
+        for (int i = 0; i < 3; ++i)
```
fsgonz: Maybe a loop through the collection (or at least a for with the collection size/array length as limit).

PR #6631, opened 2018-03-08
MULE-14616: StreamingHelper does not support TypedValue properties

ndinu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/streaming/DefaultStreamingHelper.java
```diff
@@ -106,6 +110,14 @@ private Object resolveCursorIteratorProvider(Iterator value) {
     return factory.of(event, value);
   }
 
+  private Object resolveCursorTypedValueProvider(TypedValue value) {
```
estebanwasinger: This will always return a TypedValue

estebanwasinger: Also, this method is not present in the `StreamingHelper` interface

estebanwasinger: dsadsa

PR #6633, opened 2018-03-09
MULE-14588: Deployment failed due to NoSuchMethodError: Error creatin…

gsfernandes:
…g bean with name 'http_request' when running parallel deployments

The cache added to this factory holds also references to proxies created for Mule (container) componentBuildingDefinitions but this cache is being tied to the application lifecycle, therefore at some point the proxies has unregistered the callbacks but still the Enhancer has enabled the useCache (in order to avoid OOM and creating of classes for the container that are not needed) and the NoSuchMethod is thrown.
As cglib already has the cache managed internally and always returns the same class (when create class is invoked) and only for those cases where classes are created for the application context the cache is set to false) we can just remove this cache that introduced this issues.

### DIFF
--- a/modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/spring/BeanDefinitionFactory.java
+++ b/modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/spring/BeanDefinitionFactory.java
@@ -459,8 +459,4 @@ public class BeanDefinitionFactory {
     return wrapperIdentifierAndTypeMap;
   }
 
-  /**
-   * Release resources from the bean factory.
-   */
-  public void destroy() {}
 }
--- a/modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/spring/BeanDefinitionFactory.java
+++ b/modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/spring/BeanDefinitionFactory.java
@@ -462,7 +462,5 @@ public class BeanDefinitionFactory {
   /**
    * Release resources from the bean factory.
    */
-  public void destroy() {
-    objectFactoryClassRepository.destroy();
-  }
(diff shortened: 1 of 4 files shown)


modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/spring/BeanDefinitionFactory.java
```diff
@@ -462,7 +462,5 @@ public void onMultipleValues(KeyAttributeDefinitionPair[] definitions) {
   /**
    * Release resources from the bean factory.
    */
-  public void destroy() {
-    objectFactoryClassRepository.destroy();
-  }
+  public void destroy() {}
```
elrodro83: can this method be removed altogether?

modules/spring-config/src/test/java/org/mule/runtime/config/dsl/spring/ObjectFactoryClassRepositoryTestCase.java
```diff
@@ -0,0 +1,69 @@
...
+    // First we create the objectFactory that will create an Enhancer with the callback registered and using internal cache
+    ObjectFactory objectFactory = objectFactoryClass.newInstance();
+    assertThat(((SmartFactoryBean) objectFactory).isSingleton(), is(true));
+    Class firstProxyGenerated = objectFactory.getClass();
+
+    objectFactoryClass = getObjectFactoryClass();
+    objectFactory = objectFactoryClass.newInstance();
+    assertThat(((SmartFactoryBean) objectFactory).isSingleton(), is(true));
+
+    assertThat(firstProxyGenerated, equalTo(objectFactory.getClass()));
```
elrodro83: sameInstance instead of equalsTo

PR #6634, opened 2018-03-09
MULE-13354 Introduce split-aggregate router consistent with scatter-g…

dfeist:
…ather.

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
+++ b/core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
@@ -135,6 +135,7 @@ class MuleExtensionModelDeclarer {
     declareErrorHandler(extensionDeclarer, typeLoader);
     declareTry(extensionDeclarer, typeLoader);
     declareScatterGather(extensionDeclarer, typeLoader);
+    declareSplitAggregate(extensionDeclarer, typeLoader);
     declareFirstSuccessful(extensionDeclarer);
     declareRoundRobin(extensionDeclarer);
     declareConfiguration(extensionDeclarer, typeLoader);
@@ -549,7 +550,7 @@ class MuleExtensionModelDeclarer {
         .ofType(typeLoader.load(Integer.class))
         .defaultingTo(Integer.MAX_VALUE)
         .withExpressionSupport(NOT_SUPPORTED)
-        .describedAs("This value determines the maximum level of parallelism that will be used by this router. .");
+        .describedAs("This value determines the maximum level of parallelism that will be used by this router.");
     scatterGather.onDefaultParameterGroup()
         .withOptionalParameter(TARGET_PARAMETER_NAME)
         .ofType(typeLoader.load(String.class))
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/mule/runtime/core/internal/routing/ScatterGatherRouter.java
```diff
@@ -36,7 +37,7 @@
  *
  * @since 3.5.0
  */
-public class ScatterGatherRouter extends AbstractForkJoinRouter {
+public class ScatterGatherRouter extends AbstractForkJoinRouter implements Router {
```
elrodro83: Is it ok for this to be both a Scope and a Router? Does this have any impact on the extension model?

core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -572,6 +573,47 @@ private void declareScatterGather(ExtensionDeclarer extensionDeclarer, ClassType
     // ConstructModel doesn't support it.)
   }
 
+  private void declareSplitAggregate(ExtensionDeclarer extensionDeclarer, ClassTypeLoader typeLoader) {
```
elrodro83: are you missing the `collection` parameter here?

core/src/main/java/org/mule/runtime/core/internal/routing/ScatterGatherRouter.java
dfeist: I don't see why it would be an issue. Router is sub-type of Scope.

core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -572,6 +573,47 @@ private void declareScatterGather(ExtensionDeclarer extensionDeclarer, ClassType
...
+        .ofType(typeLoader.load(String.class))
+        .defaultingTo(PAYLOAD)
+        .withExpressionSupport(REQUIRED)
+        .describedAs(TARGET_VALUE_PARAMETER_DESCRIPTION)
+        .withRole(BEHAVIOUR)
+        .withDisplayModel(DisplayModel.builder().displayName(TARGET_VALUE_PARAMETER_DISPLAY_NAME).build())
+        .withLayout(LayoutModel.builder().tabName(ADVANCED_TAB).build())
+        .withModelProperty(new TargetModelProperty());
+
+    // TODO MULE-13316 Define error model (Routers should be able to define error type(s) thrown in ModelDeclarer but
```
elrodro83: what do you need to do differently of what `flow-ref` and `choice` are doing?

core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
dfeist: yes

elrodro83: why? add a comment

dfeist: I was missing it. Added it now :-)

PR #6655, opened 2018-03-14
MULE-14722: No repeatable streams functionality in policies

elrodro83:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceAdapter.java
```diff
@@ -307,7 +308,8 @@ private void rollback() {
     @Override
     public Map<String, Object> createResponseParameters(CoreEvent event) throws MessagingException {
       try {
-        ResolverSetResult parameters = SourceAdapter.this.successCallbackParameters.resolve(from(event, configurationInstance));
+        ResolverSetResult parameters =
+            SourceAdapter.this.successCallbackParameters.resolve(from(event, configurationInstance, false));
```
marianogonzalez: aren't you changing behavior here? The previous logic was equivalent to using `true`. The connectors expect the resolverset to have cursors not providers. 

Did you run all the SDK and http tests?

core/src/main/java/org/mule/runtime/core/internal/policy/CompositeOperationPolicy.java
```diff
@@ -131,8 +138,21 @@ public CoreEvent process(CoreEvent event) throws MuleException {
   @Override
   public Publisher<CoreEvent> process(CoreEvent operationEvent) {
     try {
+      Map<String, Object> operationParameters = new HashMap<>(getParametersProcessor().getOperationParameters());
+      Map<String, Object> providers = new HashMap<>();
+
+      // TODO Avoid iterating this map every time
```
pablolagreca: Create issue?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceAdapter.java
elrodro83: Actually, this change in behavior is needed. Tests run ok, so it doesn't break anything and it fixes the source policies issue.

elrodro83: Actually, this change in behavior is needed. Tests run ok, so it doesn't break anything and it fixes the source policies issue.

PR #6665, opened 2018-03-19
MULE-14012: Create HttpRequestOptions concept in HTTP service

afelisatti:


modules/http-api/src/main/java/org/mule/runtime/http/api/client/HttpRequestOptions.java
```diff
@@ -0,0 +1,51 @@
...
+import org.mule.api.annotation.NoImplement;
+import org.mule.runtime.http.api.client.auth.HttpAuthentication;
+import org.mule.runtime.http.api.domain.message.request.HttpRequest;
+
+/**
+ * Options for setting up an {@link HttpRequest}. Instances can only be obtained through a {@link HttpRequestOptionsBuilder}.
+ *
+ * @since 4.2
+ */
+@NoImplement
```
elrodro83: isn't this to be implemented by the service implelemtation?

modules/http-api/src/main/java/org/mule/runtime/http/api/client/HttpClient.java
```diff
@@ -44,9 +44,36 @@
...
+  /**
+   * Sends a HttpRequest blocking the current thread until a response is available or the request times out.
+   *
+   * @param request the {@link HttpRequest} to send
+   * @param options the {@link HttpRequestOptions} to use
+   * @return the received {@link HttpResponse}
+   * @throws IOException if an error occurs while executing
+   * @throws TimeoutException if {@code responseTimeout} is exceeded
+   */
+  HttpResponse send(HttpRequest request, HttpRequestOptions options) throws IOException, TimeoutException;
```
elrodro83: add a default implementation that calls the previous `send` method, to keep apis compatible.

modules/http-api/src/main/java/org/mule/runtime/http/api/client/HttpClient.java
```diff
@@ -62,8 +89,37 @@ HttpResponse send(HttpRequest request, int responseTimeout, boolean followRedire
...
+   * {@link CompletableFuture#whenCompleteAsync(BiConsumer, Executor)}, to handle the response is those scenarios since they guarantee
+   * executing on a different thread.
+   *
+   * @param request the {@link HttpRequest} to send
+   * @param options the {@link HttpRequestOptions} to use
+   * @return a {@link CompletableFuture} that will complete once the {@link HttpResponse} is available
    */
-  CompletableFuture<HttpResponse> sendAsync(HttpRequest request, int responseTimeout, boolean followRedirects,
-                                            HttpAuthentication authentication);
+  CompletableFuture<HttpResponse> sendAsync(HttpRequest request, HttpRequestOptions options);
```
elrodro83: add a default implementation that calls the previous `sendAsync` method, to keep apis compatible.

modules/http-api/src/main/java/org/mule/runtime/http/api/client/HttpRequestOptions.java
afelisatti: If it was any additions to it would require default implementations but that can be handled by the builder which is in place to make usage easier.

PR #6671, opened 2018-03-20
MULE-14764: HttpRequester lifecycle is not managed properly in OAuth scenarios.

facundovs:


modules/http/src/main/java/org/mule/module/http/internal/request/client/HttpConnectorMessageProcessorProvider.java
```diff
@@ -113,4 +114,11 @@ public void dispose()
     {
         cachedMessageProcessors.invalidateAll();
     }
+
+    @Override
+    public void stop() throws MuleException
+    {
+        cachedMessageProcessors.invalidateAll();
```
afelisatti: If this is the same as `dispose` why have both?

modules/http/src/main/java/org/mule/module/http/internal/request/client/HttpRequesterBuilder.java
```diff
@@ -162,6 +162,10 @@ else if (httpRequester.getConfig() == null)
                     }
                 }
             }
+            else
+            {
+                requestConfig.start();
```
afelisatti: Is there a chance you'll start it several times?

modules/http/src/main/java/org/mule/module/http/internal/request/client/HttpConnectorMessageProcessorProvider.java
facundovs: The issue ocurrs when the application is stopped from ARM.
In this case the dispose phase is not executed.

modules/http/src/main/java/org/mule/module/http/internal/request/client/HttpRequesterBuilder.java
facundovs: This scenario is covered by the start method code:

https://github.com/mulesoft/mule/blob/mule-3.x/modules/http/src/main/java/org/mule/module/http/internal/request/DefaultHttpRequesterConfig.java#L340

modules/http/src/main/java/org/mule/module/http/internal/request/client/HttpConnectorMessageProcessorProvider.java
afelisatti: Yes, but you have some other scenarios to consider:
1) The component is stopped and an undeploy occurs, you'd invalidate the cache again.
2) The component is up and running and an undeploy occurs, you'd invalidate the cache twice.

PR #6689, opened 2018-03-27
MULE-14811: ProcessingStrategy not propagated to `async` children

elrodro83:


core/src/main/java/org/mule/runtime/core/internal/processor/AsyncDelegateMessageProcessor.java
```diff
@@ -60,25 +69,32 @@
...
+  @Inject
+  private MuleContext muleContext;
   @Inject
   private SchedulerService schedulerService;
   @Inject
   private ConfigurationComponentLocator componentLocator;
 
   protected Logger logger = LoggerFactory.getLogger(getClass());
+
+  private boolean ownProcessingStrategy = false;
```
dfeist: Is there a use case for <async> having it's own processing strategy, rather than using flow processingStrategy?

core/src/main/java/org/mule/runtime/core/internal/processor/AsyncDelegateMessageProcessor.java
```diff
@@ -90,31 +106,62 @@ public void initialise() throws InitialisationException {
...
   @Override
   public void start() throws MuleException {
-    scheduler = schedulerService
-        .ioScheduler(muleContext.getSchedulerBaseConfig().withName(name != null ? name : getLocation().getLocation()));
-    reactorScheduler = fromExecutorService(scheduler);
+    if (ownProcessingStrategy) {
+      startIfNeeded(processingStrategy);
+    }
+
+    sink = processingStrategy
```
dfeist: This ownProcessingStrategy isn't invalid, but is it necessary? Or is this only needed to support custom processing strategy necessary?

core/src/main/java/org/mule/runtime/core/internal/processor/AsyncDelegateMessageProcessor.java
elrodro83: yes, for allow setting a `maxConcurrency` in the async (next PR to come)

core/src/main/java/org/mule/runtime/core/internal/processor/AsyncDelegateMessageProcessor.java
elrodro83: yes, for allow setting a `maxConcurrency` in the async (next PR to come)

PR #6692, opened 2018-03-28
MULE-14743: Processor interceptor after() runs in connector thread and

elrodro83:
not flow thread when processor is ASYNC_CPU_LITE
* Proactor processing strategy infinitely retrying on any error


dfeist:
No way to unit test this? Also, there was already an independent JIRA issue for this that can now be closed. If you can't find it I'll pass you link later.

dfeist:
https://www.mulesoft.org/jira/browse/MULE-14687

PR #6695, opened 2018-03-29
MULE-14828: Fix only one service being stopped when container stops

balbifm:


afelisatti:
@balbifm `Format doesn't match!`

PR #6702, opened 2018-04-03
MULE-14835: In until-successful scope, if an exception ocurrs only th…

facundovs:
…e last retry payload is closed.

### DIFF
--- a/core/src/main/java/org/mule/routing/AbstractUntilSuccessfulProcessingStrategy.java
+++ b/core/src/main/java/org/mule/routing/AbstractUntilSuccessfulProcessingStrategy.java
@@ -10,11 +10,14 @@ import static org.mule.util.ClassUtils.isConsumable;
 import org.mule.DefaultMuleMessage;
 import org.mule.VoidMuleEvent;
 import org.mule.api.MessagingException;
+import org.mule.api.MuleContext;
 import org.mule.api.MuleEvent;
 import org.mule.api.MuleException;
 import org.mule.api.MuleMessage;
 import org.mule.api.MuleRuntimeException;
+import org.mule.api.util.StreamCloserService;
 import org.mule.config.i18n.MessageFactory;
+import org.mule.exception.AbstractExceptionListener;
 
 import java.io.NotSerializableException;
 import java.io.Serializable;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/mule/routing/AsynchronousUntilSuccessfulProcessingStrategy.java
```diff
@@ -361,7 +361,15 @@ private void retrieveAndProcessEvent(final Serializable eventStoreKey) throws Ob
     {
         final MuleEvent persistedEvent = getUntilSuccessfulConfiguration().getObjectStore().retrieve(eventStoreKey);
         final MuleEvent mutableEvent = threadSafeCopy(persistedEvent);
-        processEvent(mutableEvent);
+        try
+        {
+            processEvent(mutableEvent);
```
elrodro83: the caller of this maethod already handles exceptions. Why not handle this tehre?

tests/integration/src/test/resources/org/mule/test/integration/routing/until-successful-stream-closer-config.xml
```diff
@@ -0,0 +1,29 @@
...
+<mule xmlns:http="http://www.mulesoft.org/schema/mule/http" xmlns="http://www.mulesoft.org/schema/mule/core"
+      xmlns:spring="http://www.springframework.org/schema/beans"
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-current.xsd
+http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd
+http://www.mulesoft.org/schema/mule/http http://www.mulesoft.org/schema/mule/http/current/mule-http.xsd">
+
+    <spring:beans>
+        <spring:bean id="objectStore" class="org.mule.util.store.SimpleMemoryObjectStore"/>
+        <spring:bean id="failureProcessor"
```
elrodro83: use the test processor instead of a spring bean for this.

core/src/main/java/org/mule/routing/AbstractUntilSuccessfulProcessingStrategy.java
```diff
@@ -150,4 +153,19 @@ protected void ensureSerializable(MuleMessage message) throws NotSerializableExc
         }
     }
 
+    /**
+     * Any processor executed during the until-successful scope can put a closeable payload in the message,
+     * that won't be closed if an unexpected exception occurs.
```
elrodro83: this is the previous behavior, right? It is confusing.

core/src/main/java/org/mule/routing/AsynchronousUntilSuccessfulProcessingStrategy.java
facundovs: The caller method doesn't have access to the mutable event that is created here.
I could move the event creation code to the caller and change the method signature, but I am not sure if it's worth it

PR #6710, opened 2018-04-04
MULE-14849: DuplicateExportedPackageException: There are multiple art…

gsfernandes:
…ifacts exporting the same package

### DIFF
--- a/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/ApplicationStartedSplashScreen.java
+++ b/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/ApplicationStartedSplashScreen.java
@@ -24,6 +24,8 @@ import java.util.Set;
  */
 public class ApplicationStartedSplashScreen extends ArtifactStartedSplashScreen<ApplicationDescriptor> {
 
+  private static final String PLUGIN_VALUE_FORMAT = " - %s (%s)";
+
   @Override
   public void createMessage(ApplicationDescriptor descriptor) {
     doBody(format("Started app '%s'", descriptor.getName()));
@@ -38,7 +40,7 @@ public class ApplicationStartedSplashScreen extends ArtifactStartedSplashScreen<
     if (!plugins.isEmpty()) {
       doBody("Application plugins:");
       for (ArtifactPluginDescriptor plugin : plugins) {
-        doBody(format(VALUE_FORMAT, plugin.getName()));
+        doBody(format(PLUGIN_VALUE_FORMAT, plugin.getName(), plugin.getBundleDescriptor().getArtifactFileName()));
       }
     }
   }
--- a/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/DefaultApplicationFactory.java
+++ b/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/DefaultApplicationFactory.java
@@ -41,6 +41,7 @@ import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.Properties;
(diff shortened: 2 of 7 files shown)


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/ApplicationStartedSplashScreen.java
```diff
@@ -24,6 +24,8 @@
  */
 public class ApplicationStartedSplashScreen extends ArtifactStartedSplashScreen<ApplicationDescriptor> {
 
+  private static final String PLUGIN_VALUE_FORMAT = " - %s (%s)";
+
   @Override
   public void createMessage(ApplicationDescriptor descriptor) {
     doBody(format("Started app '%s'", descriptor.getName()));
```
gsfernandes: Should we also log the domain reference?

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/DefaultApplicationFactory.java
```diff
@@ -126,6 +127,11 @@ public Application createArtifact(ApplicationDescriptor descriptor) throws IOExc
         pluginDependenciesResolver.resolve(domain.getDescriptor().getPlugins(),
                                            new ArrayList<>(getArtifactPluginDescriptors(descriptor)));
 
+    // Refreshes the list of plugins on the descriptor with the resolved from domain and transitive plugin dependencies
+    Set resolvedArtifactPlugins = new LinkedHashSet<>();
+    resolvedArtifactPlugins.addAll(resolvedArtifactPluginDescriptors);
+    descriptor.setPlugins(resolvedArtifactPlugins);
```
gsfernandes: I don't like this as it is modifying the descriptor, other approach would be to have a resolvedPlugins field and set the list of plugins descriptor as they were resolved (using semver and also depending on what are the plugins already defined at the domain level)

elrodro83: Should at least log a warning if the transitive dependencies versions are inconsistent, IMO

modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/internal/plugin/BundlePluginDependenciesResolver.java
```diff
@@ -233,6 +252,23 @@ private ClassLoaderModelBuilder createBuilderWithoutExportedPackages(ClassLoader
     return classLoaderModelBuilder;
   }
 
+  private ClassLoaderModelBuilder createBuilderWithoutDependency(ClassLoaderModel originalClassLoaderModel,
+                                                                 BundleDependency dependencyToBeExcluded) {
+    ClassLoaderModelBuilder classLoaderModelBuilder = new ClassLoaderModelBuilder()
+        .dependingOn(originalClassLoaderModel.getDependencies().stream()
+            .filter(dependency -> !dependency.equals(dependencyToBeExcluded)).collect(
+                                                                                      Collectors
+                                                                                          .toCollection(() -> new LinkedHashSet<>())))
```
elrodro83: why not just `toSet()` collector?

gsfernandes: Order matters here, as later these dependencies will end up creating the classloaders for each plugin and later the extension models are generated for each plugin (classloader) and it is expected that the plugins list is sorted by dependencies as extension models require context to resolve things (for instance: http cannot be loaded if extension model for sockets was not loaded before).

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/DefaultApplicationFactory.java
gsfernandes: Something like:

PR #6721, opened 2018-04-06
MULE-14770: Content parameter is recognized as a route when using the aggregators module

lucianoRM:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/ExtensionModelHelper.java
```diff
@@ -156,6 +170,41 @@ public void visit(ConstructModel model) {
     }
   }
 
+  private Optional<? extends NestableElementModel> findNestableElementModel(ComponentIdentifier componentIdentifier) {
+    try {
+      return extensionNestableElementModelByComponentIdentifier.get(componentIdentifier, () -> {
+        String componentName = toCamelCase(componentIdentifier.getName(), COMPONENT_NAME_SEPARATOR);
+        for (ExtensionModel extensionModel : extensionsModels) {
+          if (extensionModel.getXmlDslModel().getPrefix().equals(componentIdentifier.getNamespace())) {
+            List<HasNestableElementModels> nestableElementModelsProviders = singletonList(extensionModel);
```
alegmarra: instead of doing this you need to use an IdempotentExtensionWalker and visit the constructs

alegmarra: you can use the stop() method on the walker once the model is present

PR #6724, opened 2018-04-09
MULE-14752: The wording in the Info Dialog doesn't make sense and is …

afelisatti:
…not consistent

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
+++ b/core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
@@ -208,16 +208,16 @@ class MuleExtensionModelDeclarer {
     validator.onDefaultParameterGroup()
         .withOptionalParameter("idExpression")
         .ofType(typeLoader.load(String.class))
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -240,18 +240,18 @@ private void declareIdempotentValidator(ExtensionDeclarer extensionDeclarer, Cla
...
-        .describedAs("Processes the nested list of message processors asynchronously using a thread pool");
+        .describedAs("Processes the nested list of message processors asynchronously.");
 
     async.withChain();
     async.onDefaultParameterGroup()
         .withOptionalParameter("name")
         .withExpressionSupport(NOT_SUPPORTED)
         .ofType(typeLoader.load(String.class))
-        .describedAs("Name that will be used to name the async scheduler");
+        .describedAs("Name that will be used to identify the async threads.");
```
elrodro83: let's discuss this one tomorrow, it doesn't sound completely right to me yet.

PR #6747, opened 2018-04-13
MULE-14878: Mule 3 and 4 HTTP connectors' correlation ids should interop

afelisatti:


modules/http/src/main/java/org/mule/module/http/internal/listener/HttpRequestToMuleEvent.java
```diff
@@ -134,6 +145,27 @@ else if (entity instanceof InputStreamHttpEntity)
...
+        String xCorrelationId = defaultMuleMessage.getInboundProperty(X_CORRELATION_ID);
+        String muleCorrelationId = defaultMuleMessage.getInboundProperty(MULE_CORRELATION_ID_PROPERTY);
+        if (muleCorrelationId != null)
+        {
+            if (xCorrelationId != null)
+            {
+                LOGGER.warn("'X-Correlation-ID' and 'MULE_CORRELATION_ID' headers found.");
+                if (!xCorrelationId.equals(muleCorrelationId))
+                {
+                    LOGGER.warn("Correlation values do not match. 'MULE_CORRELATION_ID' will be used.");
```
marianogonzalez: will be hard for the user reading the log to correlate this message to the one above. Ideally each situation should have one complete message, which should also include the correlation values

PR #6775, opened 2018-04-23
MULE-14923: Artemix Full Qualified Queue Name is truncated due to the use of :: separator.

fsgonz:


core/src/main/java/org/mule/endpoint/ResourceNameEndpointURIBuilder.java
```diff
@@ -76,7 +76,8 @@ else if (authority != null && !authority.equals(address))
         }
         
         int x = address.indexOf(":", y);
-        if (x > -1)
+        int repeatedColon = address.indexOf("::", y);
```
elrodro83: use a regex to contemplate both cases

PR #6778, opened 2018-04-23
DO NOT MERGE: MULE-14798: Create a RepositoryResolver to obtain the extension plugin

juandesi:


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/MTFArtifactUnderTestFinder.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * This is responsible for finding an artifact and create a {@link BundleDescriptor} to be included in an application
+ * that is going to be used to run MTF test cases.
+ * <p>
+ * It expects the classLoaderModel of the application provides the GAV of the module and an absolute path to were is
+ * located.
+ *
+ * @since 4.2
+ */
+public class MTFArtifactUnderTestFinder {
```
alegmarra: We need to take this for another spin, making this more generic (maybe as simple as having an array of 'localDependencies', each providing the uri and gav)

Also, @gsfernandes do you know if adding a local `repository` to the app's pom file would do the trick without having to modify any runtime code? Something like: 

<repositories>
    ......
    <repository>
       <id>my-local-repo</id>
       <url>file://chupito/plugins/chupame-extension/target/munit/repository</url>
    </repository>
    ......
</repositories>

If that works, it will work on any runtime from 4.1.1 and beyond, and is not a big deal since we have complete control over the pom and the target. That repository would have only the plugin under test

gsfernandes: Sure, that could work too.

PR #6779, opened 2018-04-24
MULE-14929: HTTP with one config and two listeners fails to find the correct resolution path.

fsgonz:


modules/http/src/test/resources/http-listener-multiple-listeners.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<mule xmlns:http="http://www.mulesoft.org/schema/mule/http" xmlns="http://www.mulesoft.org/schema/mule/core"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd
+http://www.mulesoft.org/schema/mule/http http://www.mulesoft.org/schema/mule/http/current/mule-http.xsd">
+    
+    <http:listener-config name="HTTP_Listener_Configuration" host="0.0.0.0" port="${port}" />
+    
+    <flow name="00187067-supportFlow">
```
afelisatti: Rename to avoid support references.

modules/http/src/test/resources/http-listener-multiple-listeners.xml
```diff
@@ -0,0 +1,18 @@
...
+        <http:listener config-ref="HTTP_Listener_Configuration" path="/api/*" />
+        <set-payload value="#['Response via /api/*']" />
+    </flow>
+    <flow name="00187067-supportFlow1">
```
afelisatti: Same

modules/http/src/test/java/org/mule/module/http/internal/listener/MultipleListenerTestCase.java
```diff
@@ -0,0 +1,68 @@
...
+    @Override
+    protected String getConfigFile()
+    {
+        return "http-listener-multiple-listeners.xml";
+    }
+
+    @Parameters
+    public static Collection<Object[]> data()
+    {
+        return asList(new Object[][] {{PATH_1, RESPONSE_1}, {PATH_2, RESPONSE_1}});
```
afelisatti: Seems like there should be a test for the second response too, so as to validate that part is still working as expected.

PR #6795, opened 2018-04-26
MULE-14927: fixing encoding error on windows

fernandezlautaro:


pmantellini:
It was tested with the original app and smart connector from the ticket and it seems to be fixed.

PR #6810, opened 2018-04-30
MULE-14956: Make minMuleVersion configurable when using ArtifactFileBuilder

mtqp:


mtqp:
API Gateway contain tests that cover the added feature.

mtqp:
Done

PR #6828, opened 2018-05-08
MULE-14976: adding connection properties to optimize when doing macro…

fernandezlautaro:
… expansions, similar to what MULE-12260 did before, but better.


fernandezlautaro:
this PR makes the org.mule.test.functional.ModuleStereotypesTestCase test go green

PR #6830, opened 2018-05-08
MULE-8817: Upgrade Spring to 4.3.16

aiannucci:


modules/cxf/src/test/java/org/mule/module/cxf/CxfInboundMessageProcessorTestCase.java
```diff
@@ -77,11 +77,9 @@ public MuleEvent process(MuleEvent event) throws MuleException
         MuleEvent event = getTestEvent(msg, getTestInboundEndpoint(REQUEST_RESPONSE));
         
         MuleEvent response = processor.process(event);
-        
         Object payload = response.getMessage().getPayload();
-        assertTrue(payload instanceof OutputHandler);
         
-        ((OutputHandler) payload).write(response, System.out);
```
elrodro83: why was this removed?

modules/spring-config/pom.xml
```diff
@@ -69,12 +27,6 @@
             <artifactId>spring-context</artifactId>
         </dependency>
 
-        <dependency>
-            <groupId>cglib</groupId>
-            <artifactId>cglib</artifactId>
-            <version>${springCglibVersion}</version>
```
elrodro83: remove this property from the root pom if it is not used elsewhere

modules/spring-extras/src/main/java/org/mule/module/spring/events/MuleEventMulticaster.java
```diff
@@ -412,6 +413,17 @@ else if (!(listener instanceof MuleEventListener))
...
     }
 
+    /***
+     * Implemented only for compliance with Spring 4.2+
+     * Please use the {@link #multicastEvent(ApplicationEvent) multicastEvent} method.
+     * @param event
+     * @param eventType
+     */
+    @Override
+    public void multicastEvent(ApplicationEvent event, ResolvableType eventType) {
```
elrodro83: is this related to the spring upgrade?

modules/spring-config/pom.xml
aiannucci: It was declared in this POM on line 17, removed as part of this commit.

modules/spring-extras/src/main/java/org/mule/module/spring/events/MuleEventMulticaster.java
aiannucci: Yes, that's why I mention in the Javadoc that we must implement this for compliance with the Spring 4.2+ ApplicationEventMulticaster interface that this class implements.

modules/cxf/src/test/java/org/mule/module/cxf/CxfInboundMessageProcessorTestCase.java
aiannucci: We shouldn't write to sysout directly and this had no purpose. I added an assertion on the response message instead.

PR #6835, opened 2018-05-08
MULE-14984: Allow FunctionalTestProcessor to fire synchronous

elrodro83:
notifications


tests/component-plugin/src/main/java/org/mule/functional/api/component/FunctionalTestProcessor.java
```diff
@@ -275,6 +274,11 @@ protected CoreEvent doProcess(CoreEvent event) throws Exception {
     return replyMessage;
   }
 
+  protected FunctionalTestNotification builtEventReceivedNotification(final Message message, CoreEvent replyMessage)
```
aiannucci: built->build

PR #6838, opened 2018-05-09
MULE-14948: SFTP Connector is not sending header=false as the output …

ndinu:
…mimetype


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/DefaultSourceCallback.java
```diff
@@ -216,6 +227,36 @@ private void executeFlow(SourceCallbackContext context, MessageProcessContext me
...
+          existingEncoding = mediaType.getCharset().get();
+        }
+      }
+    }
+
+    if (mediaType == null) {
+      mediaType = ANY;
+    }
+
+    if (initialisationParameters.get(MIME_TYPE_PARAMETER_NAME) != null) {
```
alegmarra: don't `get` two times

PR #6845, opened 2018-05-11
MULE-14729: Changes to support JDK9

aiannucci:


modules/container/src/main/java/org/mule/runtime/container/internal/JreExplorer.java
```diff
@@ -60,7 +60,9 @@ static void explorePaths(final List<String> jdkPaths, final Set<String> packages
     checkArgument(jdkPaths != null && !jdkPaths.isEmpty(), "jdkPaths cannot be empty");
 
     for (String jdkPath : jdkPaths) {
-      explorePath(packages, resources, services, jdkPath);
+      if (jdkPath != null) {
```
afelisatti: Why this now?

PR #6862, opened 2018-05-15
MULE-15034: enabling SC to *just* export resources

fernandezlautaro:


modules/spring-config/src/main/resources/META-INF/mule-module.xsd
```diff
@@ -216,7 +216,7 @@
         <xsd:complexContent>
             <xsd:extension base="mule:abstractExtensionType">
                 <xsd:sequence maxOccurs="unbounded" minOccurs="1">
-                    <xsd:choice minOccurs="1" maxOccurs="unbounded">
+                    <xsd:choice minOccurs="0" maxOccurs="unbounded">
```
alegmarra: this is enabling the very hackish

modules/extensions-xml-support/src/test/resources/modules/module-resources.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module name="module-resources"
+        xmlns="http://www.mulesoft.org/schema/mule/module"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="
+           http://www.mulesoft.org/schema/mule/module http://www.mulesoft.org/schema/mule/module/current/mule-module.xsd">
+    <!-- =================================================================================================================
+    ============ This smart connector is empty as it will be used just to export resources (such as DWL files) ===========
```
alegmarra: this is very hackish

modules/spring-config/src/main/resources/META-INF/mule-module.xsd
fernandezlautaro: the same amount of hackish as the Java SDK, so it's a bug on the XML SDK for not behaving the same:
```
package org.mule.extension.internal;

import org.mule.runtime.extension.api.annotation.Export;
import org.mule.runtime.extension.api.annotation.Extension;

/**
 * This is the main class of an extension, is the entry point from which configurations, connection providers, operations
 * and sources are going to be declared.
 */
@Xml(prefix = "lauti")
@Extension(name = "Lauti")
@Export(resources = "holis.txt")
public class LautiExtension {
}
```

The above builds, it generates an `ExtensionModel` that can be consumed from Studio/Mozart (and it was generated from the maven archetype).
No seas más papista que el papa.

modules/extensions-xml-support/src/test/resources/modules/module-resources.xml
fernandezlautaro: see above.

modules/spring-config/src/main/resources/META-INF/mule-module.xsd
pablokraan: >> No seas más papista que el papa.
@fernandezlautaro Are you talking about me?

alegmarra: IMO that's wrong too, I'd fail now but backwards compatibility is a bitch

fernandezlautaro: m
e
t
e

e
l

p
r

PR #6887, opened 2018-05-24
MULE-15041: When doing testConnectivity reconnection should be forced to set failsDeployment on false

lucianoRM:


modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/config/LifecycleAwareConfigurationInstanceTestCase.java
```diff
@@ -123,6 +130,11 @@ protected void doSetUp() throws Exception {
     ((MuleContextWithRegistries) muleContext).getRegistry().registerObject(OBJECT_CONNECTION_MANAGER, connectionManager);
     ((MuleContextWithRegistries) muleContext).getRegistry().registerObject(OBJECT_TIME_SUPPLIER, timeSupplier);
 
+    when(configurationProperties.resolveBooleanProperty(anyString())).thenReturn(of(false));
```
gsfernandes: I would change this to only return false for the lazy deployment property.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/LazyConnectivityTestingService.java
```diff
@@ -52,13 +52,7 @@ public ConnectionValidationResult testConnection(Location location) {
...
       }
       List<Throwable> causalChain = getCausalChain(e);
-      return causalChain.stream()
-          .filter(exception -> exception.getClass().equals(ConnectionException.class)
-              && ((ConnectionException) exception).getErrorType().isPresent())
-          .map(exception -> failure(exception.getMessage(), ((ConnectionException) exception).getErrorType().get(),
-                                    (Exception) exception))
-          .findFirst()
-          .orElse(unknownFailureResponse(lastMessage(causalChain), e));
+      return unknownFailureResponse(lastMessage(causalChain), e);
```
gsfernandes: 👍

modules/spring-config/src/main/java/org/mule/runtime/config/internal/LazyMuleArtifactContext.java
```diff
@@ -121,6 +121,7 @@ public LazyMuleArtifactContext(MuleContext muleContext, ConfigResource[] artifac
 
 
     muleContext.getCustomizationService().overrideDefaultServiceImpl(LAZY_COMPONENT_INITIALIZER_SERVICE_KEY, this);
+
```
gsfernandes: No need for this change

modules/spring-config/src/main/java/org/mule/runtime/config/internal/LazyMuleArtifactContext.java
```diff
@@ -119,7 +119,6 @@ public LazyMuleArtifactContext(MuleContext muleContext, ConfigResource[] artifac
     muleContext.getCustomizationService().registerCustomServiceClass(NON_LAZY_VALUE_PROVIDER_SERVICE,
                                                                      MuleValueProviderService.class);
 
-
```
gsfernandes: Just remove this change as it is not modifying the code

PR #6894, opened 2018-05-28
MULE-15086: TransactedPollingJmsMessageReceiver not working for jms spec 1.1

alvaro29d:


transports/jms/src/main/java/org/mule/transport/jms/Jms102bSupport.java
```diff
@@ -316,4 +316,11 @@ else if (producer instanceof QueueSender)
             throw new IllegalArgumentException("Producer and domain type do not match");
         }
     }
+
+    @Override
+    public void closeConsumerIfNeeded(MessageConsumer consumer)
+    {
+        // In this case the consumer it doesn't need to be closed
```
fsgonz: In this case the consumer doesn't need to be closed.

PR #6896, opened 2018-05-30
MULE-14901: 50% performance degradation with empty

elrodro83:
MessageProcessorNotificationListener
* Make notification class resolution lazy

### DIFF
--- a/core/src/main/java/org/mule/api/context/notification/ServerNotification.java
+++ b/core/src/main/java/org/mule/api/context/notification/ServerNotification.java
@@ -12,9 +12,11 @@ import org.mule.api.MuleContext;
 import org.mule.api.MuleMessage;
 import org.mule.api.config.MuleConfiguration;
 import org.mule.api.context.MuleContextAware;
+import org.mule.context.notification.ServerNotification;
 import org.mule.endpoint.MuleEndpointURI;
 import org.mule.util.ClassUtils;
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mule/api/context/notification/ServerNotification.java
```diff
@@ -264,4 +266,27 @@ private static Object verifyNullResource(Object resource)
...
+        @Override
+        public String toString()
+        {
+            if(eventClassName == null)
+            {
+                synchronized (this)
+                {
+                    if(eventClassName == null)
+                    {
+                        eventClassName = ServerNotification.this.getClass().getSimpleName();
```
gsfernandes: Would be the same value to do this instead of `ClassUtils.getClassName(getClass())`?

core/src/main/java/org/mule/api/context/notification/ServerNotification.java
```diff
@@ -62,7 +64,7 @@
     public static final int NULL_ACTION = 0;
     public static final Object NULL_RESOURCE = "";
 
-    public final String EVENT_NAME = ClassUtils.getClassName(getClass());
+    public final Object EVENT_NAME = new LazyEventName();
```
gsfernandes: Be aware that this will break API, mule-agent has to be changed.

PR #6905, opened 2018-06-06
MULE-15032: Dispose method not invoked for ConfigurationPropertiesProvider

lucianoRM:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/config/CompositeConfigurationPropertiesProvider.java
```diff
@@ -7,17 +7,40 @@
...
 
+public class CompositeConfigurationPropertiesProvider implements ConfigurationPropertiesProvider, Initialisable, Disposable {
+
+  private static final Logger LOGGER = getLogger(CompositeConfigurationPropertiesProvider.class);
   private List<ConfigurationPropertiesProvider> configurationPropertiesProviders;
 
+  @Override
+  public void dispose() {
+    for (ConfigurationPropertiesProvider configurationPropertiesProvider : configurationPropertiesProviders) {
+      disposeIfNeeded(configurationPropertiesProvider, LOGGER);
```
elrodro83: use a try catch to ensure all are disposed. then throw an aggregate exception if any exception happened.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/model/ApplicationModel.java
```diff
@@ -518,16 +529,15 @@ public String getDescription() {
...
-                } catch (InitialisationException e) {
-                  throw new MuleRuntimeException(e);
-                }
               }
             }));
     return configConfigurationPropertiesProviders;
   }
 
+  public void close() {
+    disposeIfNeeded(configurationProperties.getConfigurationPropertiesResolver(), LOGGER);
```
elrodro83: make sure an exception in dispose doesn't mess up the resto of the stop process

tests/test-extensions/test-config-properties-extension/src/main/java/org/mule/test/crafted/config/properties/extension/SecureConfigurationPropertiesProvider.java
```diff
@@ -34,11 +44,44 @@ public SecureConfigurationPropertiesProvider(ResourceProvider resourceProvider,
...
 
+  @Override
+  public void initialise() throws InitialisationException {
+    super.initialise();
+    initializationCount++;
+  }
+
+  @Override
+  public void dispose() {
+    disposeCount++;
```
elrodro83: missing `super.dispose()`?

tests/test-extensions/test-config-properties-extension/src/main/java/org/mule/test/crafted/config/properties/extension/SecureConfigurationPropertiesProvider.java
```diff
@@ -34,11 +44,44 @@ public SecureConfigurationPropertiesProvider(ResourceProvider resourceProvider,
...
+  }
+
   @Override
   public Optional<ConfigurationProperty> getConfigurationProperty(String configurationAttributeKey) {
     if (configurationAttributeKey.startsWith(SECURE_PREFIX)) {
       String effectiveKey = configurationAttributeKey.substring(SECURE_PREFIX.length());
       return Optional.ofNullable(configurationAttributes.get(effectiveKey));
+    } else if (configurationAttributeKey.startsWith(LIFECYCLE_PREFIX)) {
```
elrodro83: what's the purpose of this?

lucianoRM: It't the only solution i could come up with to test that the lifecycle phases were applied. It was either that or declaring a public static field  or getter accesible from the test class.
Keep in mind that the configuration properties extension is just used for testing purposes.

PR #6918, opened 2018-06-08
MULE-15146: Mule 4 locks on IntrospectionUtils.lambda$getAnnotatedField

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/ReflectiveOperationExecutorFactory.java
```diff
@@ -46,7 +47,7 @@ public ReflectiveOperationExecutorFactory(Class<T> implementationClass, Method o
 
   @Override
   public ComponentExecutor<M> createExecutor(M operationModel, Map<String, Object> parameters) {
-    DefaultObjectBuilder objectBuilder = new DefaultObjectBuilder(implementationClass);
+    DefaultObjectBuilder objectBuilder = new DefaultObjectBuilder(implementationClass, new ReflectionCache());
```
elrodro83: why create a new cache here instead of getting the one from the registry?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ParameterGroupArgumentResolver.java
```diff
@@ -36,7 +33,7 @@ public ParameterGroupArgumentResolver(ParameterGroupDescriptor group, Reflection
   public LazyValue<T> resolve(ExecutionContext executionContext) {
     return new LazyValue<>(() -> {
       try {
-        return new ParameterGroupObjectBuilder<T>(group, reflectionCache).build((EventedExecutionContext) executionContext);
+        return builder.build((EventedExecutionContext) executionContext);
```
elrodro83: the builder is not stateless, so reusing it may break something

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/ReflectiveOperationExecutorFactory.java
marianogonzalez: Registry doesn't exist at this point. This is just to create an internal object that will never be seen again.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ParameterGroupArgumentResolver.java
marianogonzalez: fixed

PR #6924, opened 2018-06-11
MULE-13198: ResultToMessageIterator consumes 2 messages per iteration

estebanwasinger:


core/src/main/java/org/mule/runtime/core/internal/util/message/ResultToMessageIterator.java
```diff
@@ -51,7 +51,7 @@ public Message next() {
       return (Message) value;
     }
 
-    return toMessage((Result) delegate.next(), cursorProviderFactory, event);
+    return toMessage((Result) value, cursorProviderFactory, event);
```
alegmarra: test

alegmarra: test

PR #6929, opened 2018-06-14
MULE-14938

pablolagreca:


modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/AbstractDeploymentTestCase.java
```diff
@@ -972,9 +974,13 @@ protected Application findApp(final String appName, int totalAppsExpected) {
   }
 
   protected DefaultMuleDomain createDefaultDomain() {
-    return new DefaultMuleDomain(new DomainDescriptor(DEFAULT_DOMAIN_NAME),
+    DomainDescriptor descriptor = new DomainDescriptor(DEFAULT_DOMAIN_NAME);
+    return new DefaultMuleDomain(descriptor,
                                  new DomainClassLoaderFactory(getClass().getClassLoader())
-                                     .create("domain/" + DEFAULT_DOMAIN_NAME, containerClassLoader,
+                                     .create("domain/" + DEFAULT_DOMAIN_NAME,
+                                             new RegionClassLoader("domainRegion", descriptor,
```
gsfernandes: wow, good catch! Could this break something or actually stop working something that was working before due to this bug?

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/classloader/DefaultClassLoaderNode.java
```diff
@@ -0,0 +1,229 @@
...
+    return delegateNodes;
+  }
+
+  public void setParent(ClassLoaderNode parent) {
+    this.parent = parent;
+  }
+
+  @Override
+  public Optional<ClassLoaderNode> getParent() {
+    return Optional.ofNullable(parent);
```
gsfernandes: static import

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/classloader/DefaultClassLoaderNode.java
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+  public String toString() {
+    Stack<ClassLoaderNode> classLoaderNodes = new Stack<>();
+    ClassLoaderNode currentNode = this;
+
+    int maximumIdLength = 0;
+    do {
+      classLoaderNodes.push(currentNode);
+      maximumIdLength = Math.max(maximumIdLength, currentNode.getId().length());
```
gsfernandes: static import

PR #6930, opened 2018-06-14
MULE-15152 Simultaneous reading and writing causes NPE in grizzly

fsgonz:


modules/http/src/test/java/org/mule/module/http/functional/requester/HttpRequesterQuickConnectionClose.java
```diff
@@ -0,0 +1,135 @@
...
+import java.net.ServerSocket;
+import java.net.Socket;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertThat;
+
+public class HttpRequesterQuickConnectionClose extends FunctionalTestCase
```
afelisatti: Isn't this the same as in AHC?

PR #6934, opened 2018-06-15
MULE-15170: security-manager is not initialized on lazy Mule context

gsfernandes:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/LazyMuleArtifactContext.java
```diff
@@ -229,6 +237,13 @@ public void initializeComponents(Predicate<org.mule.runtime.config.internal.mode
 
       // First unregister any already initialized/started component
       unregisterBeans(alreadyCreatedApplicationComponents);
+      if (alreadyCreatedApplicationComponents.contains(OBJECT_SECURITY_MANAGER)) {
```
pablolagreca: This should not be necessary.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/DependencyNode.java
```diff
@@ -14,10 +21,12 @@
 public class DependencyNode {
 
   private String componentName;
+  private ComponentIdentifier componentIdentifier;
   private final Type type;
 
-  public DependencyNode(String componentName, Type type) {
+  public DependencyNode(String componentName, ComponentIdentifier componentIdentifier, Type type) {
```
pablolagreca: componentIdentifier -> Set it as Optional so users know the parameter may be omitted and also put it last in the list of parameters

PR #6955, opened 2018-06-22
MULE-15037: Add support for file encoding for configuration properties and secure configuration properties

mbuchwald:


modules/spring-config/src/main/java/org/mule/runtime/config/api/dsl/model/properties/DefaultConfigurationPropertiesProvider.java
```diff
@@ -19,10 +19,7 @@
 import org.mule.runtime.config.internal.dsl.model.config.ConfigurationPropertiesException;
 import org.mule.runtime.config.internal.dsl.model.config.DefaultConfigurationProperty;
 
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.IOException;
-import java.io.InputStream;
+import java.io.*;
```
pablolagreca: fix

PR #6956, opened 2018-06-22
MULE-15195: Abstract transformer uses new error handler per operation

afelisatti:


core/src/main/java/org/mule/runtime/core/privileged/exception/AbstractExceptionListener.java
```diff
@@ -102,7 +101,7 @@ public final synchronized void initialise() throws InitialisationException {
 
   protected void doInitialise(MuleContext context) throws InitialisationException {
     requireNonNull(muleContext);
-    logger.info("Initialising exception listener: " + toString());
+    logger.debug("Initialising exception listener: " + toString());
```
elrodro83: wrap in `logger.isDebugEnabled()`

PR #6957, opened 2018-06-22
MULE-15197: PoolingConnectionHandler returns invalid connections to t…

ndinu:
…he pool


core/src/main/java/org/mule/runtime/core/internal/connection/PoolingConnectionHandler.java
```diff
@@ -65,6 +65,9 @@ public void release() {
 
     boolean returnAttempted = false;
     try {
+      if (!connectionProvider.validate(connection).isValid()) {
+        return;
```
alegmarra: add LOGGER.debug declaring that the connection was invalid and it will be destroyed instead.

PR #6958, opened 2018-06-25
MULE-14889: Compilation should fail if OutputResolver is used on Void operations

alegmarra:


modules/extensions-spring-support/src/test/java/org/mule/runtime/module/extension/internal/capability/xml/DefaultExtensionSchemaGeneratorTestCase.java
```diff
@@ -201,6 +201,9 @@ public static ExtensionModel loadExtension(Class<?> clazz, ExtensionModelLoader
     Map<String, Object> params = new HashMap<>();
     params.put(TYPE_PROPERTY_NAME, clazz.getName());
     params.put(VERSION, getProductVersion());
+    // TODO MULE-14517: This workaround should be replaced for a better and more complete mechanism
+    params.put("COMPILATION_MODE", true);
```
estebanwasinger: there is not a place where define a global constant to not repeat the string?

PR #6961, opened 2018-06-25
EE-6047: Fix distribution tests in Windows

GustavoARod:
- Improving the way how a successful un-deployment is detected because in some cases it's not enough to rely on the anchor file.

### DIFF
--- a/tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/Controller.java
+++ b/tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/Controller.java
@@ -237,6 +237,10 @@ public class Controller {
     return new File(appsDir, appName + ANCHOR_SUFFIX).exists();
   }
 
+  protected boolean isUndeployed(String appName) {
+    return !(new File(appsDir, appName + ANCHOR_SUFFIX).exists() || new File(appsDir, appName).exists());
+  }
+
   protected boolean isDomainDeployed(String domainName) {
     return new File(domainsDir, domainName + ANCHOR_SUFFIX).exists();
   }
--- a/tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/MuleProcessController.java
+++ b/tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/MuleProcessController.java
@@ -81,6 +81,10 @@ public class MuleProcessController {
     return getController().isDeployed(appName);
   }
 
+  public boolean isUndeployed(String appName) {
+    return getController().isUndeployed(appName);
+  }
+
   public File getArtifactInternalRepository(String artifactName) {
     return getController().getArtifactInternalRepository(artifactName);
   }


tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/Controller.java
```diff
@@ -237,6 +237,10 @@ protected boolean isDeployed(String appName) {
     return new File(appsDir, appName + ANCHOR_SUFFIX).exists();
   }
 
+  protected boolean isUndeployed(String appName) {
+    return !(new File(appsDir, appName + ANCHOR_SUFFIX).exists() || new File(appsDir, appName).exists());
```
aiannucci: If an application folder exists, but the anchor file doesn't, this would return false, which is not true

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/MuleProcessController.java
```diff
@@ -81,6 +81,10 @@ public boolean isDeployed(String appName) {
     return getController().isDeployed(appName);
   }
 
+  public boolean isUndeployed(String appName) {
+    return getController().isUndeployed(appName);
```
aiannucci: An application has only two deployment states, deployed or undeployed. So what would be the difference between this and !isDeployed?

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/Controller.java
GustavoARod: 🤔 That's precisely the intention of the method. 
My idea, from the lesson, learned of the classloading test case, is that if there's a folder with the same name as the app, that means that the mechanism of un-deployment didn't work well and the app wasn't undeployed successfully. So I want the user to know that the app isn't fully un-deployed.

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/MuleProcessController.java
GustavoARod: I completely agree that an app could have only two states; deployed and undeployed. Nevertheless, by the way that isDeployed() API is implemented, I think (also from first-hand experience), that a `true` output of this method totally means that the app got deployed. Nevertheless, a `false` value can't ensure 100% the opposite. I'd say something like that saying of "Correlation does not imply causation"

GustavoARod: My supporting argument, at least for the un-deployment case, is that relying just on the anchor file it's not enough. There's no a transactional-like behavior on the un-deployment process, and the deletion of the anchor file is just the beginning of that process. I think, there's a high chance that in many cases, as it's happening on Windows, something could happen in the middle of the process, that could leave it halfway. 

Maybe for the deployment scenario, the anchor file gets created at the end of the process, and therefore it's a clear sign that the app got deployed correctly.

PR #6964, opened 2018-06-26
MULE-15187: Fix core tests in Windows

GustavoARod:
- Adjusting linebreaks that make some string comparisons fail on Windows

### DIFF
--- a/core-tests/src/test/java/org/mule/runtime/core/api/message/ErrorBuilderTestCase.java
+++ b/core-tests/src/test/java/org/mule/runtime/core/api/message/ErrorBuilderTestCase.java
@@ -16,14 +16,16 @@ import static org.junit.Assert.assertThat;
 import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
 import static org.mule.runtime.api.message.Message.of;
 import static org.mule.runtime.core.internal.message.ErrorBuilder.builder;
+import static org.mule.tck.junit4.matcher.IsEqualIgnoringLineBreaks.equalIgnoringLineBreaks;
+
 import org.mule.runtime.api.exception.ComposedErrorException;
+import org.mule.runtime.api.exception.DefaultMuleException;
 import org.mule.runtime.api.exception.ErrorMessageAwareException;
 import org.mule.runtime.api.exception.MuleException;
 import org.mule.runtime.api.i18n.I18nMessage;
 import org.mule.runtime.api.message.Error;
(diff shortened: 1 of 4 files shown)


core-tests/src/test/java/org/mule/runtime/core/api/message/ErrorBuilderTestCase.java
```diff
@@ -109,40 +110,41 @@ public void givesStringRepresentation() {
...
+                   + "org.mule.runtime.core.internal.message.ErrorBuilder$ErrorImplementation\n"
+                   + "{\n"
+                   + "  description=unknown description\n"
+                   + "  detailedDescription=unknown description\n"
+                   + "  errorType=MULE:TEST\n"
+                   + "  cause=java.io.IOException\n"
+                   + "  errorMessage=-\n"
+                   + "  childErrors=[]\n"
+                   + "}]\n"
+                   + "}"));
```
aiannucci: We should use a better JSON assertion for this kind of problems.

core-tests/src/test/java/org/mule/runtime/core/internal/el/DefaultExpressionManagerTestCase.java
```diff
@@ -254,12 +255,15 @@ public void parseLogMessage() throws MuleException {
...
-                                                  NULL_BINDING_CONTEXT),
-               is(expectedOutput));
+    String melParsedTemplate = expressionManager
+        .parseLogTemplate("current message is #[mel:message]", testEvent(), TEST_CONNECTOR_LOCATION, NULL_BINDING_CONTEXT)
+        .replace("\r\n", "\n");
+    assertThat(melParsedTemplate, is(expectedOutput));
+
+    String dwParsedTemplate = expressionManager
+        .parseLogTemplate("current message is #[message]", testEvent(), TEST_CONNECTOR_LOCATION, NULL_BINDING_CONTEXT)
+        .replace("\r\n", "\n");
```
aiannucci: Can we extract this to a function or, way better, a matcher?

GustavoARod: I think a matcher is a great idea. I'll go for it. It could also help us with the other "JSON" assertions on this PR, since those "JSON" are not strictly speaking compliant with the JSON format standard, therefore they'll demand more boilerplate code (besides the overhead) to compare them using tools like Google's Gson lib, for example.

PR #6974, opened 2018-06-30
MULE-15246: Cannot replace a catch all listener with another

afelisatti:


modules/http/src/main/java/org/mule/module/http/internal/listener/HttpListenerRegistry.java
```diff
@@ -361,58 +391,97 @@ private int getPathPartsSize(String path)
...
-    public class PathMap
+    private boolean isCatchAllPath(String path)
     {
+        return WILDCARD_CHARACTER.equals(path);
+    }
 
+    /**
+     * Represents a URI path, which can be a parent to regular sub paths, a catch all sub path (/*) and a URI param path (/{param}).
+     * Request handler and matcher pairs (handlers for different methods, for example) can be added and removed from it. If all
+     * its handlers and children handlers' are removed, it will notify its parent the path itself can be removed. This means that
```
fsgonz: Possessive in children.

modules/http/src/main/java/org/mule/module/http/internal/listener/HttpListenerRegistry.java
```diff
@@ -361,58 +391,97 @@ private int getPathPartsSize(String path)
...
+        private Path catchAllUriParam;
 
-        public PathMap getCatchAllPathMap()
+        /**
+         * Makes this path self-aware within the path tree so that it can interact with its parent.
+         *
+         * @param name the name for this path
+         * @param parent the parent path
+         */
+        public void addIdentity(String name, Path parent) {
```
fsgonz: I don't know if this has to be done in the constructor (there would be Paths without name in this case).

modules/http/src/main/java/org/mule/module/http/internal/listener/HttpListenerRegistry.java
```diff
@@ -422,76 +491,150 @@ private boolean matchesNextSubPaths(String subPath, String nextSubPath)
...
         }
 
-        public void addChildPathMap(final String path, final PathMap pathMap)
+        /**
+         * Adds a new sub or uri param path, assigning itself as its parent and naming it.
+         *
+         * @param path the path name
+         * @param pathNode the node for this path
+         */
+        public void addChildPathMap(final String path, final Path pathNode)
```
fsgonz: Maybe this shouldn't be called addChildPath**Map** anymore.

PR #6991, opened 2018-07-05
MULE-15296: Missing Interceptor that sets the BindingOperationInfo if operation is one-way.

fsgonz:


modules/cxf/src/test/resources/proxy-service-one-way-flow.xml
```diff
@@ -0,0 +1,29 @@
...
+               http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd
+               http://www.mulesoft.org/schema/mule/http http://www.mulesoft.org/schema/mule/http/current/mule-http.xsd
+               http://www.mulesoft.org/schema/mule/cxf http://www.mulesoft.org/schema/mule/cxf/current/mule-cxf.xsd">
+
+    <configuration>
+        <http:config useTransportForUris="true"/>
+    </configuration>
+
+    <flow name="proxy-body">
+        <http:inbound-endpoint host="localhost" port="${port1}" path="body" />
```
elrodro83: use the connector instead of the transport

PR #7006, opened 2018-07-11
MULE-15326: Grizzly AHC race condition when using New Relic.

fsgonz:


afelisatti:
Remember to change this to be a service test in Mule 4.

PR #7014, opened 2018-07-11
MULE-15270: Mime Type Attributes are not added to a source when MetadataScope annotation is used

ndinu:


modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/validation/MediaTypeModelValidatorTestCase.java
```diff
@@ -0,0 +1,312 @@
...
+    Method operationMethod = TestMethods.class.getMethod(operationMethodName);
+    Type operationReturnType = new TypeWrapper(ResolvableType.forMethodReturnType(operationMethod), typeLoader);
+    ExtensionOperationDescriptorModelProperty extensionOperationDescriptorModelProperty =
+        mock(ExtensionOperationDescriptorModelProperty.class, new ReturnsDeepStubs());
+    when(operationModel.getModelProperty(ExtensionOperationDescriptorModelProperty.class))
+        .thenReturn(of(extensionOperationDescriptorModelProperty));
+    when(extensionOperationDescriptorModelProperty.getOperationReturnType()).thenReturn(operationReturnType);
+  }
+
+  @Test(expected = IllegalModelDefinitionException.class)
```
alegmarra: I know it's a pain, but use ExpectedException Rule and match by exception type _and_ error message substring. Otherwise it may fail for a different reason than the expected for this test. Take a look at `org.mule.runtime.module.extension.internal.loader.validation.MetadataComponentModelValidatorTestCase`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/SourceTypeWrapper.java
```diff
@@ -80,6 +80,10 @@ public SourceTypeWrapper(Class<T> aClass, ClassTypeLoader typeLoader) {
     return getMethodAnnotatedWith(OnBackPressure.class);
   }
 
+  public Type getOutputType() {
+    return getSuperClassGenerics().get(0);
```
alegmarra: same issue of empty superClassGenerics, check size and return Object otherwise

PR #7025, opened 2018-07-13
MULE-15348: Improve BindingContext construction with references to other

elrodro83:
binding contexts


core/src/main/java/org/mule/runtime/core/internal/el/DefaultBindingContextBuilder.java
```diff
@@ -125,14 +129,14 @@ public DefaultBindingContextBuilder(BindingContext bindingContext) {
 
   @Override
   public BindingContext build() {
-    return new BindingContextImplementation(parent, unmodifiableMap(bindings),
+    return new BindingContextImplementation(new ArrayList<>(delegates), unmodifiableMap(bindings),
```
afelisatti: Why the new array list?

elrodro83: because the list in the builder would still be mutable, and being a `LinkedList`, it is cheaper to build an `ArrayList` than to use a `Collections.unmodifiableList`

PR #7026, opened 2018-07-14
MULE-15358: Race condition in creation of correlation groups

fsgonz:


core/src/main/java/org/mule/routing/AbstractAggregator.java
```diff
@@ -103,6 +106,14 @@ public void initialise() throws InitialisationException
         eventCorrelator.setFailOnTimeout(isFailOnTimeout());
     }
 
+    private Lock createLock() throws InitialisationException
+    {
+        MuleLockFactory muleLockFactory = new MuleLockFactory();
```
elrodro83: `LockFactory` must be obtained from the muleContext or injected

core/src/test/java/org/mule/routing/correlation/EventCorrelatorTestCase.java
```diff
@@ -70,6 +72,8 @@
     @Mock(answer = Answers.RETURNS_DEEP_STUBS)
     private ListableObjectStore mockProcessedGroups;
     @Mock(answer = Answers.RETURNS_DEEP_STUBS)
+    private Lock mockLock;
+    @Mock(answer = Answers.RETURNS_DEEP_STUBS)
```
elrodro83: use an actual lock implementation rather than a mock. Otherwise some race conditions may occur that cause flakiness in this test

core/src/main/java/org/mule/routing/correlation/EventCorrelator.java
```diff
@@ -296,10 +296,41 @@ public MuleEvent process(MuleEvent event) throws RoutingException
...
+        if (logger.isDebugEnabled())
+        {
+            logger.debug("An event was received for an event group that has already been processed, "
+                         + "this is probably because the async-reply timed out. Correlation Id is: "
+                         + groupId + ". Dropping event");
+        }
+        // Fire a notification to say we received this message
+        muleContext.fireNotification(new RoutingNotification(event.getMessage(),
+                                                             event.getMessageSourceURI().toString(),
+                                                             RoutingNotification.MISSED_AGGREGATION_GROUP_EVENT));
```
elrodro83: static import

core/src/test/java/org/mule/routing/correlation/EventCorrelatorTestCase.java
fsgonz: I've changed the logic of the creation of the lock. So now it gets the lock internally in the EventProcessor from the factory in the muleContext. No need to pass it as a parameter in the constructor.

core/src/main/java/org/mule/routing/correlation/EventCorrelator.java
```diff
@@ -686,4 +686,14 @@ private void disposeIfDisposable(Object o)
             ((Disposable) o).dispose();
         }
     }
+    
+    private Lock getLock()
```
elrodro83: make this thread safe

PR #7033, opened 2018-07-17
Fix class cast

alegmarra:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/loader/java/type/SourceElement.java
```diff
@@ -52,4 +54,8 @@ default MetadataType getAttributesMetadataType() {
     return getSourceReturnType(getSuperClassGenerics().get(1));
   }
 
+  default Optional<Type> getOutputType() {
```
estebanwasinger: which is the difference between this and `getReturnType()` ?

PR #7043, opened 2018-07-19
 MULE-15270: Mime Type Attributes are not added to a source when MetadataScope annotation is used

ndinu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/MimeTypeParametersDeclarationEnricher.java
```diff
@@ -211,24 +174,33 @@ public void visitBinaryType(BinaryType binaryType) {
...
 
-      final OutputDeclaration output = declaration.getOutput();
-      output.setType(type.apply(toMetadataFormat(mediaType)), output.hasDynamicType());
+      property.getMediaType().ifPresent(mediaType -> {
+        final OutputDeclaration output = declaration.getOutput();
+        output.setType(type.apply(toMetadataFormat(mediaType)), output.hasDynamicType());
+      });
+    }
+
+    private boolean shouldOverrideMetadataFormat(ExecutableComponentDeclaration declaration) {
```
estebanwasinger: explain the inner logic, seems to be random rules, but surely makes sense for something

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/MediaTypeModelValidator.java
```diff
@@ -0,0 +1,177 @@
...
+ *
+ * <p>
+ * This validator goes through all operations and sources and verifies that:
+ * <ul>
+ * <li>If no OutputStaticResolver is given and the outputType is a String Type or Binary Type, it must have the MediaType
+ * annotation with a valid value</li>
+ * <li>If it has an OutputStaticResolver it must NOT have the MediaType annotation with a value</li>
+ * </ul>
+ *
+ * @since 4.2
```
estebanwasinger: full version

PR #7047, opened 2018-07-20
MULE-15412: Transactions doesn't work on lazy deployment context

estebanwasinger:


core/src/main/java/org/mule/runtime/core/internal/connection/DelegateConnectionManagerAdapter.java
```diff
@@ -250,7 +250,8 @@ public void unbind(Object config) {
     @Override
     public <C> ConnectionHandler<C> getConnection(Object config) throws ConnectionException {
       Object proxyInstance =
-          newProxyInstance(config.getClass().getClassLoader(), new Class[] {ConnectionHandler.class},
+          newProxyInstance(config.getClass().getClassLoader(),
+                           new Class[] {ConnectionHandler.class, ConnectionHandlerAdapter.class},
```
marianogonzalez: isn't this redundant? Isn't connectionHandlerAdapter enough?

PR #7048, opened 2018-07-20
MULE-14866: Only asynchronous notifications should receive a copy of the MuleEvent.

fsgonz:


core/src/test/java/org/mule/construct/processor/PipelineMessageNotificationTestCase.java
```diff
@@ -383,7 +383,7 @@ public MuleEvent process(MuleEvent event) throws MuleException
             public boolean test()
             {
                 verify(notificationManager, times(1)).fireNotification(
-                        argThat(new PipelineMessageNotificiationArgumentMatcher(PROCESS_START, false, event)));
+                        argThat(new PipelineMessageNotificiationArgumentMatcher(PROCESS_START, false, null)));
```
elrodro83: why?

core/src/test/java/org/mule/mule/enricher/MessageEnricherTestCase.java
```diff
@@ -183,6 +186,44 @@ public MuleEvent process(MuleEvent event) throws MuleException
...
+                        }
+                    }).start();
+                }
+            }
+        });
+        MessageEnricher enricher = createEnricher(new SensingNullMessageProcessor());
+        final MuleEvent in  = createBlockingEvent();
+        try
+        {
+            processEnricherInChain(enricher, in);
```
elrodro83: don't need the try-catch here

core/src/main/java/org/mule/api/context/notification/ServerNotification.java
```diff
@@ -264,4 +266,18 @@ private static Object verifyNullResource(Object resource)
...
     }
 
+    public static boolean resolveEventReference(MuleEvent event)
+    {
+        if (event instanceof VoidMuleEvent)
+        {
+            return true;
+        }
+        
+        if (event instanceof DefaultMuleEvent)
```
elrodro83: only have this check and return true otherwise

fsgonz:
I have to reconsider the general solution.

PR #7053, opened 2018-07-23
MULE-15418: SDK Perf: Avoid field lookup when field is already available

elrodro83:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/MethodArgumentResolverDelegate.java
```diff
@@ -230,9 +229,9 @@ private void initArgumentResolvers() {
   }
 
   @Override
-  public LazyValue<Object>[] resolve(ExecutionContext executionContext, Class<?>[] parameterTypes) {
+  public Supplier<Object>[] resolve(ExecutionContext executionContext, Class<?>[] parameterTypes) {
```
marianogonzalez: these were originally Suppliers but @pablolagreca changed them in order to implement some feature for MUnit. Sync up with him but I believe that unless you completely refactor that feature this is a no-go

marianogonzalez: Additionally, LazyValue is just an an enhanced type of supplier. What are you to gain from this change? Unless invoked twice, there's no difference between a LazyValue and a Supplier, specially in this case in which there's no concurrent access

elrodro83: Instantiation/validation and resolution of the LazyValue accounts for little over 1% of the cpu usage for all operations calls of a proxy scenario application, and it's an easy fix. Will check with @pablolagreca thou.

PR #7056, opened 2018-07-24
 MULE-15201: Services take to long to discover and load - fix embedde…

marianogonzalez:
…d distro

### DIFF
--- a/tests/unit/src/main/java/org/mule/tck/ZipUtils.java
+++ b/tests/unit/src/main/java/org/mule/tck/ZipUtils.java
@@ -18,7 +18,6 @@ import java.io.UncheckedIOException;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.zip.ZipEntry;
-import java.util.zip.ZipInputStream;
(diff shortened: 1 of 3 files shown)


tests/unit/src/main/java/org/mule/tck/ZipUtils.java
```diff
@@ -74,33 +73,4 @@ public static void compress(File targetFile, ZipResource[] resources) {
...
     }
   }
-
-  /**
-   * Uncompresses the given {@code zipFile} into the {@code targetFile}
-   *
-   * @param zipFile    the source file
-   * @param targetFile the target directory
-   */
-  public static void uncompress(File zipFile, File targetFile) throws IOException {
```
pablolagreca: Are we sure this method is not being used by anybody else?. Perhaps the gateway team?

marianogonzalez: yes, I added that in the original PR and now I realized I could reused FileUtils.unzip

PR #7082, opened 2018-08-02
MULE-15491: Dynamic Db datasource factory does not dispose cached

fsgonz:
datasources.

### DIFF
--- a/modules/db/src/main/java/org/mule/module/db/internal/resolver/database/DynamicDbConfigResolver.java
+++ b/modules/db/src/main/java/org/mule/module/db/internal/resolver/database/DynamicDbConfigResolver.java
@@ -10,8 +10,7 @@ package org.mule.module.db.internal.resolver.database;
 import static com.mchange.v2.c3p0.DataSources.destroy;
 import static org.mule.common.Result.Status.FAILURE;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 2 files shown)


modules/db/src/main/java/org/mule/module/db/internal/resolver/database/DynamicDbConfigResolver.java
```diff
@@ -36,8 +43,9 @@
  * configured is resolved to the same static {@link DataSourceConfig}
  * </p>
  */
-public class DynamicDbConfigResolver extends AbstractAnnotatedObject implements DbConfigResolver
+public class DynamicDbConfigResolver extends AbstractAnnotatedObject implements DbConfigResolver, Disposable
 {
+    protected transient Log logger = LogFactory.getLog(getClass());
```
elrodro83: use slf4j api

modules/db/src/main/java/org/mule/module/db/internal/resolver/database/DynamicDbConfigResolver.java
```diff
@@ -113,4 +121,28 @@ public TestResult test()
...
     }
+    
+    @Override
+    public void dispose()
+    {
+        Collection<DbConfig> configs = cache.values();
+        for (DbConfig config : configs)
+        {
+            DataSource dataSource = config.getDataSource();
+            destroyResolvedDataSource(dataSource);
```
elrodro83: inline

PR #7090, opened 2018-08-03
MULE-15333: parse template is not setting mimeType to result

lucianoRM:


core/src/main/java/org/mule/runtime/core/internal/processor/simple/ParseTemplateProcessor.java
```diff
@@ -41,8 +56,14 @@ private void loadContentFromLocation() throws InitialisationException {
       if (location == null) {
         throw new IllegalArgumentException("Location cannot be null");
       }
-      content = IOUtils.getResourceAsString(location, this.getClass());
-
+      content = getResourceAsString(location, this.getClass());
+      if (outputMimeType == null) {
+        MediaType fromLocationMediaType = parse(mimetypesFileTypeMap.getContentType(location));
+        //This is because BINARY is the default value returned if nothing can be resolved. We should not force the value.
+        if (!BINARY.equals(fromLocationMediaType)) {
```
afelisatti: Is this really so wrong? You’ll be using ANY if not BINARY

core/src/main/java/org/mule/runtime/core/internal/processor/simple/ParseTemplateProcessor.java
```diff
@@ -58,11 +79,23 @@ private void evaluateCorrectArguments() {
 
   }
 
+  private TypedValue getPayload(Object value) {
```
afelisatti: Can’t you build this in the Mesage builder?

core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -384,6 +384,18 @@ private void declareParseTemplate(ExtensionDeclarer extensionDeclarer, ClassType
         .ofType(typeLoader.load(String.class))
         .withExpressionSupport(NOT_SUPPORTED)
         .describedAs("The location of the template. The order in which the processor will attempt to load the file is: from the file system, from a URL, then from the classpath.");
+
```
estebanwasinger: Now if the operation can return parsed templates in their correct mimeType, those parsed templates won't be just simple Strings, these could be Objects, Arrays, Anythings, so the new Output of the operation should be `Any` instead of `String` -> https://github.com/mulesoft/mule/blob/bb8e65fac6bdafc9f27a76a51affcea66b4d95ad/core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java#L372

core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -384,6 +384,18 @@ private void declareParseTemplate(ExtensionDeclarer extensionDeclarer, ClassType
...
+    parseTemplate.onDefaultParameterGroup()
+        .withOptionalParameter("outputEncoding")
+        .ofType(typeLoader.load(String.class))
+        .withExpressionSupport(NOT_SUPPORTED)
```
estebanwasinger: These parameters should support expressions

core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -384,6 +384,18 @@ private void declareParseTemplate(ExtensionDeclarer extensionDeclarer, ClassType
...
+        .describedAs("The encoding to be assigned to the result generated when parsing the template.");
+
+    parseTemplate.onDefaultParameterGroup()
+        .withOptionalParameter("outputMimeType")
+        .ofType(typeLoader.load(String.class))
+        .withExpressionSupport(NOT_SUPPORTED)
```
estebanwasinger: same

core/src/main/java/org/mule/runtime/core/internal/processor/simple/ParseTemplateProcessor.java
```diff
@@ -41,8 +56,14 @@ private void loadContentFromLocation() throws InitialisationException {
       if (location == null) {
         throw new IllegalArgumentException("Location cannot be null");
       }
-      content = IOUtils.getResourceAsString(location, this.getClass());
-
+      content = getResourceAsString(location, this.getClass());
```
estebanwasinger: You should use the configured encoding to read the file because the file will be already encoded and the given encoding will be useless.

core/src/main/java/org/mule/runtime/core/internal/processor/simple/ParseTemplateProcessor.java
lucianoRM: Isn't that better? I don't know what could happen if BINARY is the mimeType. Maybe that would force some logic elsewhere when in fact we don't know if it should

core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
lucianoRM: If that is the case then not only this definition should be changed but the logic that parses the config value too. 
Check that the set-payload and set-variable processors have the same attributes and neither of those accept expressions.

core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
lucianoRM: same

core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
afelisatti: Wouldn't they be Strings in different formats?

lucianoRM: I would need to check again but as i understood it, the parseLogTemplate() method always returns a String.

core-tests/src/test/java/org/mule/runtime/core/internal/processor/simple/ParseTemplateProcessorTestCase.java
```diff
@@ -81,7 +92,7 @@ public void testParseTemplateWithBothLocationAndContentDefined() throws Initiali
     parseTemplateProcessor.initialise();
   }
 
-  @Test(expected = IllegalArgumentException.class)
+  @Test(expected = InitialisationException.class)
```
afelisatti: Use the rule instead. It will allow you to match the error message as well to make sure the correct failure is occurring.

core-tests/src/test/java/org/mule/runtime/core/internal/processor/simple/ParseTemplateProcessorTestCase.java
```diff
@@ -109,13 +120,83 @@ public void testParseTemplateFromLocation() throws InitialisationException, IOEx
...
+
+  @Test
+  public void unsupportedEncodingThrowsException() throws Exception {
+    expectedException.expect(IllegalArgumentException.class);
+    parseTemplateProcessor.setOutputEncoding("invalidEncoding");
+  }
+
+  @Test
+  public void invalidMimeTypeStringThrowsException() throws Exception {
+    expectedException.expect(IllegalArgumentException.class);
```
afelisatti: Match the error message too

core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -384,6 +384,18 @@ private void declareParseTemplate(ExtensionDeclarer extensionDeclarer, ClassType
         .ofType(typeLoader.load(String.class))
         .withExpressionSupport(NOT_SUPPORTED)
         .describedAs("The location of the template. The order in which the processor will attempt to load the file is: from the file system, from a URL, then from the classpath.");
+
+    parseTemplate.onDefaultParameterGroup()
```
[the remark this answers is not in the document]
lucianoRM: Yes

core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
lucianoRM: It does not make sense to add expression support for loading the template where the 'content' or 'location' that define the template are static. 
Besides that would cause the  need to reload the template from a file every time a new encoding is resolved from an expression

PR #7093, opened 2018-08-07
MULE-15511: Fix Services throwing UndeclaredThrowableException with c…

balbifm:
…hecked Exceptions

### DIFF
--- a/core-tests/src/test/java/org/mule/runtime/core/internal/util/DefaultMethodInvokerTestCase.java
+++ b/core-tests/src/test/java/org/mule/runtime/core/internal/util/DefaultMethodInvokerTestCase.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.internal.util;
+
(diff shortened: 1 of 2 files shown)


modules/service/src/main/java/org/mule/runtime/module/service/internal/manager/LazyServiceProxy.java
```diff
@@ -104,19 +105,23 @@ public Service forApplication(MethodInvoker methodInvoker) {
...
+      } else if (methodClass == Startable.class) {
+        return handleStart();
+      } else if (methodClass == Stoppable.class) {
+        return handleStop();
+      } else if (methodClass == TypeSupplier.class) {
+        return assembly.getServiceContract();
+      } else {
+        return methodInvoker.invoke(service.get(), method, args);
+      }
+    } catch (InvocationTargetException e) {
```
marianogonzalez: move this to DefaultMethodInvoker so that it becomes reusable

modules/service/src/test/java/org/mule/runtime/module/service/internal/manager/LazyServiceTestCase.java
```diff
@@ -70,6 +78,14 @@ public void invokeRealMethod() throws Exception {
     verify(actualService).doSomething();
   }
 
+  @Test
+  public void throwRealException() throws Exception {
```
marianogonzalez: move to DefaultMethodInvokerTestCase

PR #7095, opened 2018-08-07
MULE-15421: lookup transformer on registry of registry helper

Bardammu:
If merged, this PR makes the registry helper to lookup transformers on
the registry if any transformer is found on its cache.

### DIFF
--- a/core-tests/src/test/java/org/mule/runtime/core/internal/registry/MuleRegistryHelperTransformerLookupTestCase.java
+++ b/core-tests/src/test/java/org/mule/runtime/core/internal/registry/MuleRegistryHelperTransformerLookupTestCase.java
@@ -79,7 +79,7 @@ public class MuleRegistryHelperTransformerLookupTestCase extends AbstractMuleTes
 
   @Test
   public void lookupTransformersOnRegistries() {
-    List<Transformer> transformers = muleRegistryHelper.lookupTransformers(APPLE_DATA_TYPE ,DataType.STRING);
+    List<Transformer> transformers = muleRegistryHelper.lookupTransformers(APPLE_DATA_TYPE, DataType.STRING);
 
     Mockito.verify(registry, times(1)).lookupObjects(Transformer.class);
     assertEquals(1, transformers.size());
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mule/runtime/core/internal/registry/MuleRegistryHelper.java
```diff
@@ -217,6 +217,10 @@ protected Transformer resolveTransformer(DataType source, DataType result) throw
       readLock.unlock();
     }
 
+    if (results.isEmpty()) {
```
marianogonzalez: what is this for? If the cache is remembering the fact that there's no such transformer, why would a subsequent lookup succeed when the previous logic has failed? If a new transformer is in fact found, why isn't it cached?

core-tests/src/test/java/org/mule/runtime/core/internal/registry/MuleRegistryHelperTransformerLookupTestCase.java
```diff
@@ -31,15 +34,20 @@
...
   private final MuleContext muleContext = mock(MuleContext.class);
   private final MuleRegistryHelper muleRegistryHelper = new MuleRegistryHelper(registry, muleContext);
   private final Converter stringToOrange = new MockConverterBuilder().from(DataType.STRING).to(ORANGE_DATA_TYPE).build();
   private final Converter orangeToString = new MockConverterBuilder().from(ORANGE_DATA_TYPE).to(DataType.STRING).build();
+  private final Converter appleToString = new MockConverterBuilder().from(APPLE_DATA_TYPE).to(DataType.STRING).build();
 
   @Before
   public void setUp() throws Exception {
+    Collection transformers = Collections.singletonList(appleToString);
+    when(registry.lookupObjects(Transformer.class)).thenReturn(transformers);
```
marianogonzalez: this is not how transformers are resolved. They are registered by a post processor on initialisation time

marianogonzalez:
The commit title has to reference the issue's title, not what the solution was

core-tests/src/test/java/org/mule/runtime/core/internal/registry/MuleRegistryHelperTransformerLookupTestCase.java
marianogonzalez: this is not a good solution for the root problem. The root problem is that TransformerResolvers from the domain are not being considered by the app. This is a hack that will a) severely hurt performance when running in a domain b) not take advantage of any fix/improvements done in the Transformer resolution framework

PR #7097, opened 2018-08-07
MULE-15202: fix more tests

marianogonzalez:


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/artifact/ConnectionManagerConfigurationBuilder.java
```diff
@@ -36,13 +37,15 @@
...
-      } else {
+      if (isConfigLess(parentArtifact)) {
         registerDefaultConnectionManager(muleContext);
+      } else {
+        ConnectionManagerAdapter parentConnectionManager = lookupObject(parentContext, OBJECT_CONNECTION_MANAGER);
+        if (parentConnectionManager != null) {
+          ConnectionManager connectionManager =
+              new CompositeConnectionManager(new DelegateConnectionManagerAdapter(muleContext), parentConnectionManager);
+          registerConnectionManager(muleContext, connectionManager);
+        }
```
elrodro83: missing the `else` here?

PR #7107, opened 2018-08-10
MULE-15421: Domain project cannot submit HTTP Request with only the URL

Bardammu:


core/src/main/java/org/mule/runtime/core/internal/registry/CompositeMuleRegistryHelper.java
```diff
@@ -0,0 +1,42 @@
...
+package org.mule.runtime.core.internal.registry;
+
+import org.mule.runtime.api.metadata.DataType;
+import org.mule.runtime.core.api.MuleContext;
+import org.mule.runtime.core.api.transformer.Transformer;
+import org.mule.runtime.core.api.transformer.TransformerException;
+
+/**
+ * Composes {@link MuleRegistryHelper} of application that includes the {@link MuleRegistryHelper} of a domain.
+ */
```
alegmarra: `@since 4.2.0`

PR #7110, opened 2018-08-10
MULE-15378: Be able to generate a Metadata cache ID for a component m…

alegmarra:
…odel based on model parameters

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
+++ b/core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
@@ -747,6 +747,7 @@ class MuleExtensionModelDeclarer {
         .withOptionalParameter("defaultErrorHandler-ref")
         .ofType(typeLoader.load(String.class))
         .withExpressionSupport(NOT_SUPPORTED)
+        .withAllowedStereotypes(singletonList(ERROR_HANDLER))
         .describedAs("The default error handler for every flow. This must be a reference to a global error handler.");
 
     configuration.onDefaultParameterGroup()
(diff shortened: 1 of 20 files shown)


core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -747,6 +747,7 @@ private void declareConfiguration(ExtensionDeclarer extensionDeclarer, ClassType
         .withOptionalParameter("defaultErrorHandler-ref")
         .ofType(typeLoader.load(String.class))
         .withExpressionSupport(NOT_SUPPORTED)
+        .withAllowedStereotypes(singletonList(ERROR_HANDLER))
```
gsfernandes: Why this change? Is this related to another issue?

modules/extensions-soap-support/src/main/java/org/mule/runtime/module/extension/soap/internal/loader/SoapInvokeOperationDeclarer.java
```diff
@@ -117,7 +117,7 @@ private void declareMetadata(OperationDeclarer operation, ClassTypeLoader loader
                                                                                 InvokeOutputTypeResolver::new,
                                                                                 NullMetadataResolver::new);
     operation.withModelProperty(new MetadataResolverFactoryModelProperty(() -> factory));
-    operation.withModelProperty(new MetadataKeyIdModelProperty(loader.load(WebServiceTypeKey.class), KEYS_GROUP));
+    operation.withModelProperty(new MetadataKeyIdModelProperty(loader.load(WebServiceTypeKey.class), KEYS_GROUP, "SoapInvoke"));
```
gsfernandes: Is this hardcoded value needed?

alegmarra: it should be a constant (will change), but is indeed needed. That's the name of the Metadata category for the resolver, and in this case there's no need for it to be anonymous.

core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
alegmarra: the allowedStereotypes of a String parameter identify it as a reference to a global element. In this case, the parameter was declared as an errorHandler-ref, but the model said nothing about it.

alegmarra: This 'reference' logic is used here: https://github.com/mulesoft/mule/pull/7110/files#diff-a9272277dbb0602334e99271758985f7R182

PR #7118, opened 2018-08-13
Lifecycle

marianogonzalez:


core-tests/src/test/java/org/mule/runtime/core/internal/construct/DefaultFlowTestCase.java
```diff
@@ -319,4 +319,55 @@ public void lifecycleOrder() throws MuleException {
...
+    doThrow(stopException).when((Stoppable) source).stop();
+
+    muleContext = spy(muleContext);
+    flow = (DefaultFlow) Flow.builder(FLOW_NAME, muleContext)
+        .source(source)
+        .processors(singletonList(mock(Processor.class)))
+        .build();
+
+    flow.initialise();
+    //when(muleContext.isStarted()).thenReturn(true);
```
elrodro83: remove

core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
```diff
@@ -321,14 +324,11 @@ protected void doStartProcessingStrategy() throws MuleException {
...
     try {
       startIfStartable(pipeline);
-    } catch (MuleException e) {
-      // If the pipeline couldn't be started we would need to stop the processingStrategy (if possible) in order to avoid leaks
-      doStop();
-      doStopProcessingStrategy();
-      throw e;
+    } catch (Exception e) {
+      stopOnFailure(e);
+      return;
```
elrodro83: shouldn't this rethrow the exception?

marianogonzalez: no. it is thrown by stopOnFailure()

PR #7120, opened 2018-08-14
MULE-15267: test fix

mbuchwald:


core/src/main/java/org/mule/runtime/core/internal/context/thread/notification/ThreadNotificationLogger.java
```diff
@@ -27,16 +27,12 @@ public ThreadNotificationLogger(ThreadNotificationService threadNotificationServ
...
-    return THREAD_LOGGING;
-  }
-
   public void setStartingThread(String eventId) {
     setStartingThread(eventId, false);
   }
 
   public void setStartingThread(String eventId, boolean avoidIfSet) {
-    if (!isEnabled()) {
+    if (!isThreadLoggingEnabled()) {
```
elrodro83: this will read the system property every time, which is performance heavy

core/src/main/java/org/mule/runtime/core/internal/context/thread/notification/ThreadNotificationLogger.java
```diff
@@ -49,7 +45,7 @@ public void setStartingThread(String eventId, boolean avoidIfSet) {
   }
 
   public void setFinishThread(String eventId) {
-    if (!isEnabled()) {
+    if (!isThreadLoggingEnabled()) {
```
elrodro83: same

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/ProactorStreamProcessingStrategyFactory.java
```diff
@@ -198,7 +198,7 @@ private boolean scheduleIoRwEvent(CoreEvent event) {
     }
 
     private Flux<CoreEvent> scheduleWithLogging(ReactiveProcessor processor, Scheduler processorScheduler, CoreEvent event) {
-      if (THREAD_LOGGING) {
+      if (isThreadLoggingEnabled()) {
```
elrodro83: this will read the system property every time, which is performance heavy

core/src/main/java/org/mule/runtime/core/api/config/MuleConfiguration.java
```diff
@@ -129,4 +129,9 @@
    */
   DynamicConfigExpiration getDynamicConfigExpiration();
 
+  /**
+   * @return whether the Thread Logging Notifications are enabled
+   */
+  boolean isThreadLoggingEnabled();
```
elrodro83: add `@since`

PR #7123, opened 2018-08-14
MULE-15501: Initial support for artifact specific resource requests within a region.

afelisatti:


modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/RegionClassLoader.java
```diff
@@ -216,19 +232,73 @@ public boolean removeClassLoader(ArtifactClassLoader artifactClassLoader) {
...
+
+        // Check whether it's an exported resource from a matching artifact
+        List<ArtifactClassLoader> exportingArtifactClassLoaders = resourceMapping.get(normalizedResource);
+        if (exportingArtifactClassLoaders != null) {
+          for (ArtifactClassLoader artifactClassLoader : exportingArtifactClassLoaders) {
+            BundleDescriptor descriptor = artifactClassLoader.getArtifactDescriptor().getBundleDescriptor();
+            if (isRequestedArtifact(descriptor, groupId, artifactId, version, () -> {
+              LOGGER.warn("Required version '{}' for artifact '{}:{}' not found. Searching in available version '{}'...",
+                          version, descriptor.getGroupId(), descriptor.getArtifactId(), descriptor.getVersion());
+              return true;
```
gsfernandes: Shouldn't return false on this case?

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/RegionClassLoader.java
```diff
@@ -216,19 +232,73 @@ public boolean removeClassLoader(ArtifactClassLoader artifactClassLoader) {
...
+        // Check whether it's a resource from an API dependency, since all those should be considered exported
+        ClassLoaderModel classLoaderModel = this.getArtifactDescriptor().getClassLoaderModel();
+        for (BundleDependency dependency : classLoaderModel.getDependencies()) {
+          Optional<String> classifier = dependency.getDescriptor().getClassifier();
+          if (classifier.isPresent() && API_CLASSIFIERS.contains(classifier.get())) {
+            BundleDescriptor descriptor = dependency.getDescriptor();
+            if (isRequestedArtifact(descriptor, groupId, artifactId, version, () -> false)) {
+              return descriptorMapping.computeIfAbsent(descriptor, (CheckedFunction<BundleDescriptor, URLClassLoader>) d -> {
+                try {
+                  return new URLClassLoader(new URL[] {dependency.getBundleUri().toURL()});
```
gsfernandes: I would do a close of those classloaders when the region is disposed.

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/RegionClassLoader.java
afelisatti: My thinking here is that when requesting a plugin resource, the requested version might not be what ends up in the app and there will only be one plugin version, so if there's a partial match we should try and search in it. Since the resource is part of the API (being exported) chances are you'll find what you need.

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/RegionClassLoader.java
afelisatti: You're right, done.

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/RegionClassLoader.java
gsfernandes: You are right should try if it matches the groupId:artifactId as the resolved plugin (in case of plugins) could be affected by semver so you don't know exactly which version to reference.

PR #7129, opened 2018-08-15
MULE-15559: `lookup` function locks its calling thread

elrodro83:


core/src/main/java/org/mule/runtime/core/internal/el/mvel/function/LookupFunction.java
```diff
@@ -49,6 +55,10 @@
  */
 public class LookupFunction implements ExpressionFunction {
 
+  private static final Logger LOGGER = getLogger(LookupFunction.class);
+  private static final RunOnce DISCOURAGE_NOTICE =
+      Once.of(() -> LOGGER.warn("Usage of `lookup` is discouraged, and should be replaced with `flow-ref` components ideally."));
```
afelisatti: `featuring a target variable`?

PR #7131, opened 2018-08-16
MULE-15567: DocumentBuilderFactory class must be set as sysproperty to avoid degradation.

fsgonz:


fsgonz:
This will not be set by default.

fsgonz:
This will not be set by default.

PR #7135, opened 2018-08-17
MULE-15497: Modify class loader model to include all dependency versi…

afelisatti:
…on for raml artifacts

### DIFF
--- a/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/AbstractMavenClassLoaderModelLoader.java
+++ b/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/AbstractMavenClassLoaderModelLoader.java
@@ -253,6 +253,7 @@ public abstract class AbstractMavenClassLoaderModelLoader implements ClassLoader
     }
   }
 
+  // TODO: MULE-15577 - Review plugin and API definition resolution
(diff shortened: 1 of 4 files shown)


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/AbstractMavenClassLoaderModelLoader.java
```diff
@@ -233,19 +238,53 @@ private ClassLoaderModel createLightPackageClassLoaderModel(File artifactFile, M
...
+        .filter(dependency -> {
+          Optional<String> classifier = dependency.getDescriptor().getClassifier();
+          return classifier.isPresent() && API_CLASSIFIERS.contains(classifier.get());
+        })
+        .collect(toSet());
+
+    Deque<org.mule.maven.client.api.model.BundleDependency> dependenciesToCheck = new ArrayDeque<>(apiDependencies);
+    Set<org.mule.maven.client.api.model.BundleDependency> missingApiDependencies = new HashSet<>();
+
+    while (!dependenciesToCheck.isEmpty()) {
```
gsfernandes: Just realize we have almost the algorithm at org.mule.runtime.deployment.model.internal.plugin.BundlePluginDependenciesResolver#resolvePluginsDependencies
We should refactor all of this and have it in one resolver.

PR #7139, opened 2018-08-21
MULE-15573: MULE-15573: ServiceConfigurationError: org.mule.runtime.d…

gsfernandes:
…eployment.model.api.artifact.ArtifactConfigurationProcessor: Error reading configuration file - Fixed FunctionalTestCase

### DIFF
--- a/modules/container/src/main/resources/META-INF/mule-test-module.properties
+++ b/modules/container/src/main/resources/META-INF/mule-test-module.properties
@@ -1,0 +1,3 @@
+module.name=container
+
+artifact.export.classPackages=org.mule.runtime.container.internal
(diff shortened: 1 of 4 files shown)


modules/container/src/main/resources/META-INF/mule-test-module.properties
```diff
@@ -0,0 +1,3 @@
+module.name=container
+
+artifact.export.classPackages=org.mule.runtime.container.internal
```
elrodro83: ?

gsfernandes: MuleArtifactFunctionalTestCase extends FunctionalTestCase which has to access these clases in order to reuse the code for ModuleRepository and just change the ClasspathModuleDiscoverer temporary folder logic. This is a mechanism that has the runner to export internal classes for test only.

PR #7140, opened 2018-08-22
MULE-15505 - Spring component scan does not work

pablolagreca:


modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/RegionClassLoader.java
```diff
@@ -64,11 +65,14 @@
...
 
   protected static final String REGION_OWNER_CANNOT_BE_REMOVED_ERROR = "Region owner cannot be removed";
-
   static {
     registerAsParallelCapable();
   }
 
+
+  private static final Pattern DOT_REPLACEMENT_PATTERN = Pattern.compile("\\.");
+  private static final String PACKAGE_PATH_SEPARATOR = "/";
```
afelisatti: Will this work in windows?

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/RegionClassLoader.java
```diff
@@ -285,6 +301,20 @@ public final URL findResource(final String name) {
           }
         }
       }
+    } else if (name.endsWith(CLASS_EXTENSION)) {
+      // This is when a class is requested as a resource like with spring classpath scanning.
+      int lastIndexOfPackageSeparator = name.lastIndexOf(PACKAGE_PATH_SEPARATOR);
+      String resourceFolder = name.substring(0, lastIndexOfPackageSeparator != -1 ? lastIndexOfPackageSeparator : 0);
```
afelisatti: Does this mean `""` is a valid exported package?

modules/artifact/src/test/java/org/mule/runtime/module/artifact/api/classloader/RegionClassLoaderTestCase.java
```diff
@@ -497,10 +503,83 @@ public void doesNotFindResourceFromApiWithWrongVersion() throws Exception {
...
+                               ImmutableList.of(new URL("http://com.mycompany/SomeClass.class"),
+                                                new URL("http://com.mycompany/SomeOtherClass.class")));
+  }
+
+  private void findExportedPackageAsResource(String resource, URL resourceExpectedUrl, String resourcePackage) {
+    ClassLoader parentClassLoader = mock(ClassLoader.class);
+    createClassLoaders(parentClassLoader);
+
+    appClassLoader.addResource(resource, resourceExpectedUrl);
+    when(lookupPolicy.getPackageLookupStrategy(resourcePackage)).thenReturn(ChildFirstLookupStrategy.CHILD_FIRST);
```
afelisatti: static import

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/RegionClassLoader.java
```diff
@@ -285,6 +301,20 @@ public final URL findResource(final String name) {
           }
         }
       }
+    } else if (name.endsWith(CLASS_EXTENSION)) {
```
afelisatti: I think this also affects `getResources`

pablolagreca: It does, getResources delegates to findResources. There's a test cases validating that scenario.

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/RegionClassLoader.java
pablolagreca: It should since we are using the normalize function wrapping the generated string for the path. I can also validated in windows afterwards.

PR #7148, opened 2018-08-27
Mule 15379

alegmarra:


core/src/main/java/org/mule/runtime/core/internal/metadata/MuleMetadataService.java
```diff
@@ -168,21 +123,19 @@ public void initialise() throws InitialisationException {
...
   }
 
   public MetadataCache getMetadataCache(String id) {
     try {
-      return caches.get(id);
-    } catch (ExecutionException e) {
-      throw new MuleRuntimeException(createStaticMessage("Could not get the cache with id:" + id), e);
+      return cacheManager.getOrCreateCache(id);
+    } catch (Exception e) {
+      e.printStackTrace();
```
estebanwasinger: is this on purpose?

core/src/main/java/org/mule/runtime/core/internal/metadata/cache/DefaultPersistentMetadataCacheManager.java
```diff
@@ -0,0 +1,148 @@
...
+    });
+  }
+
+  @Override
+  public void dispose(String keyHash) {
+    withKeyLock(keyHash, key -> {
+      try {
+        metadataStore.get().remove(key);
+      } catch (ObjectDoesNotExistException e) {
+        LOGGER.debug("No exact match found: " + key);
```
estebanwasinger: I would change it to something more explicit, key `More than one match for key: $key. Disposing all matching values`

core/src/main/java/org/mule/runtime/core/internal/metadata/MuleMetadataService.java
alegmarra: of course not 😥

alegmarra:
TODO: review connection order

PR #7150, opened 2018-08-27
MULE-15604: MuleArtifactContext should allow Tooling Runtime Client t…

gsfernandes:
…o reuse logic for loading config resource files

### DIFF
--- a/modules/spring-config/src/main/java/org/mule/runtime/config/internal/ArtifactConfigResolver.java
+++ b/modules/spring-config/src/main/java/org/mule/runtime/config/internal/ArtifactConfigResolver.java
@@ -11,7 +11,7 @@ import org.mule.runtime.config.api.dsl.processor.ArtifactConfig;
 /**
  * Resolves the {@link ArtifactConfig} for the given {@link ArtifactConfigResolverContext context}.
  *
- * @since 4.1.4
+ * @since 4.1.4, 4.2.0
  */
 public interface ArtifactConfigResolver {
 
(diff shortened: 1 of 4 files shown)


modules/spring-config/src/main/java/org/mule/runtime/config/internal/ArtifactConfigResolverContext.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Provides a context for the {@link ArtifactConfigResolver} to hook up the input for the resolver.
+ *
+ * @since 4.1.4, 4.2.0
+ */
+public interface ArtifactConfigResolverContext {
```
pablolagreca: Can you annotate this with the API annotation that means DO NOT USE?

modules/spring-config/src/main/java/org/mule/runtime/config/internal/ArtifactConfigResolver.java
```diff
@@ -0,0 +1,26 @@
...
+package org.mule.runtime.config.internal;
+
+import org.mule.runtime.config.api.dsl.processor.ArtifactConfig;
+
+/**
+ * Resolves the {@link ArtifactConfig} for the given {@link ArtifactConfigResolverContext context}.
+ *
+ * @since 4.1.4, 4.2.0
+ */
+public interface ArtifactConfigResolver {
```
pablolagreca: Same annotation as in the other comment.

PR #7153, opened 2018-08-27
MULE-15499: Create resource loading service

afelisatti:


modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/MuleArtifactClassLoader.java
```diff
@@ -101,6 +149,7 @@ public void addShutdownListener(ShutdownListener listener) {
 
   @Override
   public void dispose() {
+    pathMapping.clear();
```
gsfernandes: Shouldn't you do a urlClassLoader.close() on each value of the Map?

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/RegionClassLoader.java
```diff
@@ -273,7 +267,7 @@ public final URL findResource(final String name) {
         if (exportingArtifactClassLoaders != null) {
           for (ArtifactClassLoader artifactClassLoader : exportingArtifactClassLoaders) {
             BundleDescriptor descriptor = artifactClassLoader.getArtifactDescriptor().getBundleDescriptor();
-            if (isRequestedArtifact(descriptor, groupId, artifactId, version, () -> {
+            if (descriptor != null && isRequestedArtifact(descriptor, groupId, artifactId, version, () -> {
```
gsfernandes: This was added due to test runner scenarios, right? I would add a comment to this line

PR #7155, opened 2018-08-29
MULE-15624: dbCreateStruct function should be able to resolve BLOB and CLOB types

fsgonz:
CLOB

### DIFF
--- a/modules/db/src/main/java/org/mule/module/db/internal/domain/connection/DefaultDbConnection.java
+++ b/modules/db/src/main/java/org/mule/module/db/internal/domain/connection/DefaultDbConnection.java
@@ -343,7 +343,7 @@ public class DefaultDbConnection extends AbstractDbConnection
         }
         catch (SQLException e)
         {
-            logger.warn("Unable to resolve lobs. Proceeding with original attributes.", e.getMessage());
+            logger.warn("Unable to resolve lobs: {}. Proceeding with original attributes.", e.getMessage());
         }
         return delegate.createStruct(typeName, attributes);
     }
@@ -424,15 +424,14 @@ public class DefaultDbConnection extends AbstractDbConnection
 
     private boolean shouldResolveAttributeWithJdbcType(int dbDataType, String dbDataTypeName, DbType jdbcType)
     {
-        if (dbDataType == UNKNOWN_DATA_TYPE || dbDataType == jdbcType.getId())
+        if (dbDataType == UNKNOWN_DATA_TYPE)
         {
-            if (dbDataTypeName.equals(jdbcType.getName()))
-            {
-                return true;
-            }
+            return dbDataTypeName.equals(jdbcType.getName());
         }
-        
-        return false;
+        else
+        {
+            return dbDataType == jdbcType.getId();
+        }
     }
 
     protected void doResolveLobIn(Object[] attributes, int index, String dataTypeName) throws SQLException
--- a/modules/db/src/main/java/org/mule/module/db/internal/domain/connection/DefaultDbConnection.java
+++ b/modules/db/src/main/java/org/mule/module/db/internal/domain/connection/DefaultDbConnection.java
@@ -7,14 +7,9 @@
 
 package org.mule.module.db.internal.domain.connection;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
-import org.mule.module.db.internal.domain.transaction.TransactionalAction;
-import org.mule.module.db.internal.resolver.param.ParamTypeResolverFactory;
-import org.mule.util.IOUtils;
-
 import static java.lang.String.format;
 import static org.mule.module.db.internal.domain.type.JdbcTypes.BLOB_DB_TYPE;
+import static org.mule.module.db.internal.domain.type.JdbcTypes.CLOB_DB_TYPE;
 import static org.mule.util.IOUtils.toByteArray;
 
 import java.io.InputStream;
@@ -38,12 +33,22 @@ import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.Executor;
 
+import org.mule.module.db.internal.domain.transaction.TransactionalAction;
+import org.mule.module.db.internal.resolver.param.ParamTypeResolverFactory;
+import org.mule.util.IOUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 /**
  * Delegates {@link Connection} behaviour to a delegate
  */
 public class DefaultDbConnection extends AbstractDbConnection
 {
-    protected transient Log logger = LogFactory.getLog(getClass());
+    private final Logger logger = LoggerFactory.getLogger(this.getClass());
+
+    private static final int UNKNOWN_DATA_TYPE = -1;
+
+    public static final int DATA_TYPE_INDEX = 5;
 
     public static final int ATTR_TYPE_NAME_INDEX = 6;
 
@@ -337,7 +342,7 @@ public class DefaultDbConnection extends AbstractDbConnection
         }
         catch (SQLException e)
         {
-            logger.warn("Unable to resolve lobs. Proceeding with original attributes.");
+            logger.warn("Unable to resolve lobs. Proceeding with original attributes.", e);
         }
         return delegate.createStruct(typeName, attributes);
     }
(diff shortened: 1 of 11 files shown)


modules/db/src/main/java/org/mule/module/db/internal/domain/connection/DefaultDbConnection.java
```diff
@@ -34,6 +43,9 @@
  */
 public class DefaultDbConnection extends AbstractDbConnection
 {
+    protected transient Log logger = LogFactory.getLog(getClass());
```
elrodro83: use slf4j

modules/db/src/main/java/org/mule/module/db/internal/domain/connection/DefaultDbConnection.java
```diff
@@ -319,6 +331,14 @@ public Array createArrayOf(String typeName, Object[] elements) throws SQLExcepti
     @Override
     public Struct createStruct(String typeName, Object[] attributes) throws SQLException
     {
+        try
+        {
+            resolveLobs(typeName, attributes);
+        }
+        catch (SQLException e)
+        {
+            logger.warn("Unable to resolve lobs. Proceeding with original attributes.");
```
elrodro83: should at least the message of the exception be logged?

modules/db/src/main/java/org/mule/module/db/internal/domain/connection/DefaultDbConnection.java
```diff
@@ -34,6 +43,9 @@
  */
 public class DefaultDbConnection extends AbstractDbConnection
 {
+    protected transient Log logger = LogFactory.getLog(getClass());
+
+    public static final int ATTR_TYPE_NAME_INDEX = 6;
```
elrodro83: is this constant among db engines? if so, is there no constant or an alternative way to get it in the JDBC apis?

modules/db/src/main/java/org/mule/module/db/internal/domain/connection/DefaultDbConnection.java
```diff
@@ -363,5 +383,85 @@ public int getNetworkTimeout() throws SQLException
...
+    }
+    
+
+    protected void doResolveLobIn(Object[] attributes, int index, String dataTypeName) throws SQLException
+    {
+        if (dataTypeName.equals(BLOB_DB_TYPE.getName()))
+        {
+            attributes[index] = createBlob(attributes[index]);
+        }
+        else
```
elrodro83: should check the data type for this too?

modules/db/src/main/java/org/mule/module/db/internal/domain/connection/DefaultDbConnection.java
```diff
@@ -363,5 +383,85 @@ public int getNetworkTimeout() throws SQLException
...
+        {
+            blob.setBytes(1, toByteArray((InputStream) attribute));
+        }
+        else if (attribute instanceof String)
+        {
+            blob.setBytes(1, ((String) attribute).getBytes());
+        }
+        else
+        {
+            throw new IllegalArgumentException(format("Cannot create a %s from a value of type %s", Struct.class.getName(), attribute.getClass()));
```
elrodro83: put quotes around the variable values

modules/db/src/main/java/org/mule/module/db/internal/domain/connection/DefaultDbConnection.java
```diff
@@ -363,5 +383,85 @@ public int getNetworkTimeout() throws SQLException
...
+        {
+            clob.setString(1, (String) attribute);
+        }
+        else if (attribute instanceof InputStream)
+        {
+            clob.setString(1, IOUtils.toString((InputStream) attribute));
+        }
+        else
+        {
+            throw new IllegalArgumentException(format("Cannot create a %s from a value of type %s", Struct.class.getName(), attribute.getClass()));
```
elrodro83: put quotes around the variable values

modules/db/src/main/java/org/mule/module/db/internal/domain/connection/OracleDbConnection.java
```diff
@@ -77,4 +87,50 @@ private Method getCreateArrayOfMethod(Connection delegate)
...
+    {
+        Map<Integer, String> dataTypes = new HashMap<Integer, String>();
+
+        PreparedStatement ps = this.prepareStatement(QUERY_TYPE_ATTRS);
+        ps.setString(1, typeName);
+        ResultSet resultSet = null;
+
+        try
+        {
+            resultSet = ps.executeQuery();
```
elrodro83: doesnt ResultSet implement AutoClosable/Closable?

modules/db/src/main/java/org/mule/module/db/internal/domain/connection/DefaultDbConnection.java
fsgonz: It is the jdbc api

https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getAttributes-java.lang.String-java.lang.String-java.lang.String-java.lang.String-

https://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html

But there is no constant in the jdbc api for that index, that I find. Every example I found uses the index without a constant (in a sense I find it reasonable as it is something related exclusively to the method).

modules/db/src/main/java/org/mule/module/db/internal/domain/connection/OracleDbConnection.java
fsgonz: It is not needed from java 7.

modules/db/src/main/java/org/mule/module/db/internal/domain/connection/DefaultDbConnection.java
```diff
@@ -337,7 +342,7 @@ public Struct createStruct(String typeName, Object[] attributes) throws SQLExcep
         }
         catch (SQLException e)
         {
-            logger.warn("Unable to resolve lobs. Proceeding with original attributes.");
+            logger.warn("Unable to resolve lobs. Proceeding with original attributes.", e);
```
elrodro83: this will log the full stack trace, only the message of the exception is of interest

modules/db/src/main/java/org/mule/module/db/internal/domain/connection/DefaultDbConnection.java
```diff
@@ -383,45 +388,44 @@ public int getNetworkTimeout() throws SQLException
...
-                String dataType = resultSet.getString(ATTR_TYPE_NAME_INDEX);
-                
-                doResolveLobIn(attributes, index, dataType);
-                
-                index++;
-            }
-        }
-        finally 
+
+        resultSet = this.getMetaData().getAttributes(this.getCatalog(), null, typeName, null);
```
elrodro83: why did you remove the try? this won't close the ResultSet like this

modules/db/src/main/java/org/mule/module/db/internal/domain/connection/DefaultDbConnection.java
```diff
@@ -383,45 +388,44 @@ public int getNetworkTimeout() throws SQLException
...
+
+    protected void doResolveLobIn(Object[] attributes, int index, int dataType, String dataTypeName) throws SQLException
     {
-        if (dataTypeName.equals(BLOB_DB_TYPE.getName()))
+        if (dataTypeName.equals(BLOB_DB_TYPE.getName()) && (dataType == UNKNOWN_DATA_TYPE || dataType == BLOB_DB_TYPE.getId()))
         {
             attributes[index] = createBlob(attributes[index]);
         }
-        else
+        else if (dataTypeName.equals(CLOB_DB_TYPE.getName()) && (dataType == UNKNOWN_DATA_TYPE || dataType == CLOB_DB_TYPE.getId()))
```
elrodro83: extract the common condition between this and the previuos ti an outer if

modules/db/src/main/java/org/mule/module/db/internal/domain/connection/OracleDbConnection.java
```diff
@@ -114,22 +120,13 @@ protected void resolveLobs(String typeName, Object[] attributes) throws SQLExcep
         ps.setString(1, typeName);
         ResultSet resultSet = null;
 
-        try
-        {
-            resultSet = ps.executeQuery();
+        resultSet = ps.executeQuery();
```
elrodro83: try-with-resources

modules/db/src/main/java/org/mule/module/db/internal/domain/connection/DefaultDbConnection.java
```diff
@@ -342,7 +343,7 @@ public Struct createStruct(String typeName, Object[] attributes) throws SQLExcep
         }
         catch (SQLException e)
         {
-            logger.warn("Unable to resolve lobs. Proceeding with original attributes.", e);
+            logger.warn("Unable to resolve lobs. Proceeding with original attributes.", e.getMessage());
```
elrodro83: missing the placeholder iun the message

modules/db/src/main/java/org/mule/module/db/internal/domain/connection/DefaultDbConnection.java
```diff
@@ -319,6 +337,14 @@ public Array createArrayOf(String typeName, Object[] elements) throws SQLExcepti
     @Override
     public Struct createStruct(String typeName, Object[] attributes) throws SQLException
     {
+        try
+        {
+            resolveLobs(typeName, attributes);
+        }
+        catch (SQLException e)
+        {
+            logger.warn("Unable to resolve lobs. Proceeding with original attributes.", e.getMessage());
```
elrodro83: missing placeholder in the message

modules/db/src/main/java/org/mule/module/db/internal/domain/connection/OracleDbConnection.java
```diff
@@ -77,4 +87,49 @@ private Method getCreateArrayOfMethod(Connection delegate)
...
+    private Map<Integer, String> getDataTypes(String typeName) throws SQLException
+    {
+        Map<Integer, String> dataTypes = new HashMap<Integer, String>();
+
+        try (PreparedStatement ps = this.prepareStatement(QUERY_TYPE_ATTRS))
+        {
+            ps.setString(1, typeName);
+            ResultSet resultSet = null;
+
+            resultSet = ps.executeQuery();
```
elrodro83: inline

modules/db/src/main/java/org/mule/module/db/internal/domain/connection/DefaultDbConnection.java
```diff
@@ -364,4 +390,96 @@ public int getNetworkTimeout() throws SQLException
...
+        }
+        else if (shouldResolveAttributeWithJdbcType(dataType, dataTypeName, CLOB_DB_TYPE))
+        {
+            attributes[index] = createClob(attributes[index]);
+        }
+    }
+
+    private boolean shouldResolveAttributeWithJdbcType(int dbDataType, String dbDataTypeName, DbType jdbcType)
+    {
+        if (dbDataType == UNKNOWN_DATA_TYPE || dbDataType == jdbcType.getId())
```
elrodro83: join both ifs with an &&

modules/db/src/main/java/org/mule/module/db/internal/domain/connection/DefaultDbConnection.java
fsgonz: I've change this so that in case the dbDataType is known from the jdbc implementation, it is used to determine if the attribute has a type that has to be resolved. Otherwise the jdbc type name is used.

PR #7161, opened 2018-08-30
MULE-15403 - Forbidden flow names must tell me which one is triggerin…

pablolagreca:
…g the exception


modules/spring-config/src/main/java/org/mule/runtime/config/internal/model/ApplicationModel.java
```diff
@@ -845,7 +848,14 @@ private void validateNameHasValidCharacters() {
     executeOnEveryRootElement(componentModel -> {
       String nameAttributeValue = componentModel.getNameAttribute();
       if (nameAttributeValue != null) {
-        verifyStringDoesNotContainsReservedCharacters(nameAttributeValue);
+        try {
+          verifyStringDoesNotContainsReservedCharacters(nameAttributeValue);
+        } catch (IllegalArgumentException e) {
+          throw new MuleRuntimeException(I18nMessageFactory.createStaticMessage(String
```
elrodro83: static imports

PR #7187, opened 2018-09-10
MULE-15665: Http null host when port is invalid

lucianoRM:


modules/http/src/test/java/org/mule/module/http/internal/request/DefaultHttpRequesterTestCase.java
```diff
@@ -7,29 +7,72 @@
...
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Spy;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.runners.MockitoJUnitRunner;
+import org.mockito.stubbing.Answer;
 
-
+@RunWith(MockitoJUnitRunner.class)
```
afelisatti: Can't you just declare the spy inline?

lucianoRM: I think so but, isn't this the preferred way?

PR #7202, opened 2018-09-12
MULE-15154: Until successfull should support 0 as a maxRetries value

mbuchwald:


core/src/main/java/org/mule/runtime/core/internal/routing/UntilSuccessful.java
```diff
@@ -70,7 +72,8 @@ public void initialise() throws InitialisationException {
     this.nestedChain = newChain(getProcessingStrategy(locator, getRootContainerLocation()), processors);
     super.initialise();
     timer = schedulerService.cpuLightScheduler();
-    policyTemplate = new SimpleRetryPolicyTemplate(millisBetweenRetries, maxRetries);
+    policyTemplate =
+        maxRetries > 0 ? new SimpleRetryPolicyTemplate(millisBetweenRetries, maxRetries) : new NoRetryPolicyTemplate();
```
elrodro83: fail if lower than 0

core/src/main/java/org/mule/runtime/core/api/retry/policy/RetryPolicy.java
```diff
@@ -74,7 +76,11 @@
                                        Consumer<Throwable> onExhausted,
                                        Function<Throwable, Throwable> errorFunction,
                                        Scheduler retryScheduler) {
-    return publisher;
+    return from(publisher).onErrorMap(e -> {
```
elrodro83: isn't this same (or very similar code) repeated in `SimpleRetryPolicy`?

PR #7209, opened 2018-09-13
MULE-15579: When returning or receiving Object or InputStream the Met…

Bardammu:
…adataType should be Any

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/IntrospectionUtils.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/IntrospectionUtils.java
@@ -294,7 +294,9 @@ public final class IntrospectionUtils {
     }
 
     if ((returnType.isSameType(Object.class) ||
-        returnType.isSameType(InputStream.class)) && type != null) {
+        returnType.isAssignableTo(InputStream.class) ||
+        returnType.isAssignableTo(Byte[].class) ||
+        returnType.isAssignableTo(byte[].class)) && type != null) {
 
       MetadataType metadataType = typeBuilder().anyType().build();
       MetadataTypeEnricher enricher = new MetadataTypeEnricher();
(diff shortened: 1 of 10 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/IntrospectionUtils.java
```diff
@@ -290,6 +293,15 @@ public static MetadataType getReturnType(Type returnType) {
       }
     }
 
+    if ((returnType.isSameType(Object.class) ||
+        returnType.isSameType(InputStream.class)) && type != null) {
```
estebanwasinger: I would change it to `returnType.getMetadataType().instanceOf(BinaryType.class)` so you cover all binary cases

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/HeisenbergScopes.java
```diff
@@ -47,6 +47,7 @@ public void executeAnything(Chain chain, CompletionCallback<Void, Void> cb) {
     chain.process(cb::success, (t, e) -> cb.error(t));
   }
 
+  @MediaType(value = TEXT_PLAIN, strict = false)
```
estebanwasinger: You add this because a validation started to fail, right? 
So you need to detect this case and use -> `org.mule.runtime.module.extension.internal.loader.validation.ModelValidationUtils#isCompiletime` to ignore if the error occurs in a non compiletime scenario.

estebanwasinger: Review all modified tests and check if they run correctly if you ignore the validation on `compiletime`

Bardammu: I'll comment here for tracking purpose

Ignoring the validation if not in _compiletime_,  all tests of the test case `MediaTypeModelValidatorTestCase` that don't expect an exception to be thrown are green.

On test cases `JavaDeclarationDelegateTestCase` and `DynamicMetadataDeclarationEnricherTestCase` the metadataType is obtain from the instrocpectionUtils itself insteed of harcoded it.

Without the added annotation on `HeisenbergScopes` the only test case that fails is `ExtensionActivatorTestCase, which is not relevant on not compile time.

Bardammu: I compiled the java modules with 4.2.0-SNAPSHOT and all test passed as well.

PR #7212, opened 2018-09-13
MULE-15578: Extensions are unable to contribute with an ObjectStoreMa…

marianogonzalez:
…nager to the Object Store Connector

### DIFF
--- a/modules/extensions-spring-support/pom.xml
+++ b/modules/extensions-spring-support/pom.xml
@@ -298,6 +298,20 @@
         </dependency>
         <dependency>
             <groupId>org.mule.tests</groupId>
+            <artifactId>custom-os-extension</artifactId>
+            <version>${project.version}</version>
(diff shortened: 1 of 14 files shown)


modules/extensions-spring-support/pom.xml
```diff
@@ -296,6 +296,20 @@
...
+            <groupId>org.mule.tests</groupId>
+            <artifactId>custom-os-extension</artifactId>
+            <version>${project.version}</version>
+            <classifier>mule-plugin</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mule.connectors</groupId>
+            <artifactId>mule-objectstore-connector</artifactId>
+            <version>1.1.2-SNAPSHOT</version>
```
estebanwasinger: why a snapshot version?  try to use a property

PR #7213, opened 2018-09-14
MULE-15379: Support resources MetadataCache at Mule Runtime level - A…

gsfernandes:
…llow support for sharing persistent object store data between applications

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/store/PartitionedPersistentObjectStore.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/store/PartitionedPersistentObjectStore.java
@@ -35,7 +35,7 @@ public class PartitionedPersistentObjectStore<T extends Serializable> extends Ab
   private static final Logger LOGGER = getLogger(PartitionedPersistentObjectStore.class);
   public static final String OBJECT_STORE_DIR = "objectstore";
 
-  protected MuleContext muleContext;
+  private MuleContext muleContext;
   private File storeDirectory;
   private Map<String, PersistentObjectStorePartition> partitionsByName = new HashMap<>();
   private boolean initialized = false;
@@ -50,7 +50,7 @@ public class PartitionedPersistentObjectStore<T extends Serializable> extends Ab
   }
 
   @Override
-  public void open() throws ObjectStoreException {
+  public synchronized void open() throws ObjectStoreException {
     if (!initialized) {
       initObjectStoreDirectory();
       loadPreviousStoredPartitions();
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/mule/runtime/core/internal/store/PartitionedPersistentObjectStore.java
```diff
@@ -50,7 +50,7 @@ public PartitionedPersistentObjectStore(MuleContext context) {
   }
 
   @Override
-  public synchronized void open() throws ObjectStoreException {
+  public void open() throws ObjectStoreException {
```
alegmarra: are you sure of this?

core/src/main/java/org/mule/runtime/core/internal/store/SharedPartitionedPersistentObjectStore.java
```diff
@@ -0,0 +1,25 @@
...
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.internal.store;
+
+import java.io.File;
+import java.io.Serializable;
+
+public class SharedPartitionedPersistentObjectStore<T extends Serializable> extends PartitionedPersistentObjectStore<T> {
```
alegmarra: docs

core/src/main/java/org/mule/runtime/core/internal/store/PartitionedPersistentObjectStore.java
```diff
@@ -35,7 +35,7 @@
   private static final Logger LOGGER = getLogger(PartitionedPersistentObjectStore.class);
   public static final String OBJECT_STORE_DIR = "objectstore";
 
-  private MuleContext muleContext;
+  protected MuleContext muleContext;
```
alegmarra: why?

core/src/main/java/org/mule/runtime/core/internal/store/PartitionedPersistentObjectStore.java
gsfernandes: Not, my bad. That is not supposed to be changed.

core/src/main/java/org/mule/runtime/core/internal/store/PartitionedPersistentObjectStore.java
gsfernandes: Part of PoC, not needed this change. Good catch!

PR #7221, opened 2018-09-16
MULE-15601: If you are not using compatibility mode MEL expression is a

elrodro83:
* Ignore benchmark concurrency


afelisatti:
@elrodro83 check the validation failures before merging

PR #7224, opened 2018-09-17
MULE-15708: async in sub-flow must use its own processingStrategy

elrodro83:
instance

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/processor/AsyncDelegateMessageProcessor.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/processor/AsyncDelegateMessageProcessor.java
@@ -19,12 +19,13 @@ import static org.mule.runtime.core.api.lifecycle.LifecycleUtils.startIfNeeded;
 import static org.mule.runtime.core.api.lifecycle.LifecycleUtils.stopIfNeeded;
 import static org.mule.runtime.core.internal.component.ComponentUtils.getFromAnnotatedObject;
 import static org.mule.runtime.core.internal.event.DefaultEventContext.child;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mule/runtime/core/internal/processor/AsyncDelegateMessageProcessor.java
```diff
@@ -133,6 +143,27 @@ public void initialise() throws InitialisationException {
     super.initialise();
   }
 
+  /**
+   * A fallback method for creating a {@link ProcessingStrategyFactory} to be used in case the user hasn't specified one through
```
afelisatti: extra `through` I think

core/src/main/java/org/mule/runtime/core/internal/processor/AsyncDelegateMessageProcessor.java
```diff
@@ -133,6 +143,27 @@ public void initialise() throws InitialisationException {
...
+   * class name system property
+   *
+   * @return a {@link DirectProcessingStrategyFactory}
+   */
+  protected ProcessingStrategyFactory createDefaultProcessingStrategyFactory() {
+    return new DirectProcessingStrategyFactory();
+  }
+
+  private ProcessingStrategyFactory defaultProcessingStrategy() {
+    final ProcessingStrategyFactory defaultProcessingStrategyFactory =
```
afelisatti: Will this change? Might be better to store it.

elrodro83: shouldn't change, that default can only be set via system property

PR #7232, opened 2018-09-19
 MULE-15653: NPE while resolving CacheId for Metadata on Web Service C…

gsfernandes:
…onsumer - ExtensionComponent not initialized when used on sub-flow without a flow referencing it and using LazyInit


modules/spring-config/src/main/java/org/mule/runtime/config/internal/LazyMuleArtifactContext.java
```diff
@@ -179,6 +184,19 @@ private void applyLifecycle(List<String> createdComponentModels, boolean applySt
     });
   }
 
+  private void applyLifecycleMessageProcessorChainBuilder(Object object, String phase) {
+    if (object instanceof MessageProcessorChainBuilder) {
+      MessageProcessorChain messageProcessorChain = ((MessageProcessorChainBuilder) object).build();
+      messageProcessorChain.getMessageProcessors().stream().forEach(processor -> {
```
elrodro83: can't you apply lifecycle directly on the chain instead of iteratin and applying on each element?

PR #7235, opened 2018-09-20
 MULE-15343: When sftp transport is configured with append mode for du…

gsfernandes:
…plicate handling strategy tempDirOutbound should not be used

### DIFF
--- a/transports/sftp/src/main/java/org/mule/transport/sftp/SftpMessageDispatcher.java
+++ b/transports/sftp/src/main/java/org/mule/transport/sftp/SftpMessageDispatcher.java
@@ -16,6 +16,9 @@
 
 package org.mule.transport.sftp;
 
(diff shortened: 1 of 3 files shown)


transports/sftp/src/main/java/org/mule/transport/sftp/SftpMessageDispatcher.java
```diff
@@ -90,25 +89,32 @@ protected void doDispatch(MuleEvent event) throws Exception
             filename = client.duplicateHandling(destDir, filename, sftpUtil.getDuplicateHandling());
             transferFilename = filename;
 
+            boolean appendMode = sftpUtil.getDuplicateHandling().equals(SftpConnector.PROPERTY_DUPLICATE_HANDLING_APPEND);
```
fsgonz: static import

PR #7236, opened 2018-09-20
MULE-15722: HTTP Requester is sending content-type even if no payload…

afelisatti:
… is sent


modules/http/src/test/java/org/mule/module/http/functional/requester/HttpRequestContentTypeTestCase.java
```diff
@@ -96,10 +116,21 @@ protected void writeResponse(HttpServletResponse response) throws IOException
...
+    private void validateContentTypeFromFlow(String flowName, String method) throws Exception
     {
-        runFlow(flowName, TEST_MESSAGE);
+        runFlow(flowName, method);
         assertThat(getFirstReceivedHeader(CONTENT_TYPE_PROPERTY), equalTo(EXPECTED_CONTENT_TYPE));
     }
 
+    private void validateNoContentTypeFromFlow(String payload, String method) throws Exception {
+        runFlow(payload, method);
+        assertThat(headers.get(CONTENT_TYPE_PROPERTY), is(Matchers.<String>empty()));
```
lucianoRM: can't you use emptyString() or is that not available?

PR #7247, opened 2018-09-24
MULE-15176: Upgrade Guava to 26.0

aiannucci:


distributions/standalone/assembly-whitelist.txt
```diff
@@ -189,6 +189,7 @@
 /mule-standalone-${productVersion}/lib/opt/abdera-parser-1.1.3.jar
 /mule-standalone-${productVersion}/lib/opt/abdera-server-1.1.3.jar
 /mule-standalone-${productVersion}/lib/opt/abdera-spring-1.1.3.jar
+/mule-standalone-${productVersion}/lib/opt/animal-sniffer-annotations-1.14.jar
```
elrodro83: what about these extra libs?

aiannucci: Those are transitive dependencies in the new versions of Guava

PR #7249, opened 2018-09-25
MULE-15739: Simplify Parameter Group resolution for operators

elrodro83:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ConnectionArgumentResolver.java
```diff
@@ -47,8 +47,8 @@
     return new LazyValue<>(() -> {
       ConnectionHandler connectionHandler =
           ((ExecutionContextAdapter<ComponentModel>) executionContext).getVariable(CONNECTION_PARAM);
-      checkArgument(connectionHandler != null,
-                    "No connection was provided for the component [" + executionContext.getComponentModel().getName() + "]");
+      checkArgument(connectionHandler != null, () -> "No connection was provided for the component ["
```
marianogonzalez: if you're optimizing for performance, then do an if and avoid the lambda

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/objectbuilder/ObjectBuilderUtils.java
```diff
@@ -26,7 +26,7 @@
    */
   public static <T> T createInstance(Class<T> prototypeClass) {
     try {
-      return withContextClassLoader(prototypeClass.getClassLoader(), () -> prototypeClass.newInstance());
```
marianogonzalez: this method is static and public. Not safe to predict where it's been called from

PR #7253, opened 2018-09-25
EE-6307: Repeatable streams are closed in streaming batch:aggregator

marianogonzalez:


core/src/main/java/org/mule/runtime/core/api/util/StreamingUtils.java
```diff
@@ -207,14 +206,48 @@ public static CursorStreamProvider asCursorProvider(byte[] bytes) {
    */
   public static CoreEvent consumeRepeatablePayload(CoreEvent event) {
     TypedValue payload = event.getMessage().getPayload();
-    final Object originalPayload = payload.getValue();
 
-    if (originalPayload == null) {
+    if (payload.getValue() == null) {
```
elrodro83: couldn't `originalPayload` still be null in some cases?

PR #7261, opened 2018-09-26
MULE-15648: Handle additionalPluginDependencies when creating classlo…

gsfernandes:
…aders

### DIFF
--- a/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/ArtifactClassLoaderModelBuilder.java
+++ b/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/ArtifactClassLoaderModelBuilder.java
@@ -132,12 +132,12 @@ public abstract class ArtifactClassLoaderModelBuilder extends ClassLoaderModel.C
                   String dependencyArtifactId = getAttribute(dependency, ARTIFACT_ID);
                   String dependencyVersion = getAttribute(dependency, VERSION);
                   additionalDependencies.add(new BundleDependency.Builder()
-                                                 .setDescriptor(new BundleDescriptor.Builder()
-                                                                    .setGroupId(dependencyGroupId)
-                                                                    .setArtifactId(dependencyArtifactId)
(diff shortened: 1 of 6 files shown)


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/ArtifactClassLoaderModelBuilder.java
```diff
@@ -63,79 +67,148 @@ public ArtifactClassLoaderModelBuilder(File applicationFolder) {
...
+  }
+
+  private void processAdditionalPluginLibraries(Plugin packagingPlugin) {
+    doProcessAdditionalPluginLibraries(packagingPlugin);
+  }
+
+  protected void doProcessAdditionalPluginLibraries(Plugin packagingPlugin) {
+    Object configuration = packagingPlugin.getConfiguration();
+    if (configuration != null) {
+      Xpp3Dom additionalPluginDependenciesDom = ((Xpp3Dom) configuration).getChild("additionalPluginDependencies");
```
lucianoRM: We should move these strings to static variables. Maybe the best solution will be to add them in the mule-maven-plugin and import them from here.

gsfernandes: Sure, let's do that.

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/descriptor/BundleDependency.java
```diff
@@ -21,6 +22,7 @@
   private BundleDescriptor descriptor;
   private BundleScope scope;
   private URI bundleUri;
+  private Set<BundleDependency> additionalDependencies;
```
lucianoRM: It doesn't seem right for me to have a BundleDependency with additionalDependencies. 
It feels like it's not a proper Object model. Could't that information be in the ClassLoaderModel instead?

PR #7275, opened 2018-10-01
MULE-15763: Add jdoc to underline the importance of setting the correct context in lifecycle methods.

fsgonz:


core/src/main/java/org/mule/api/lifecycle/Stoppable.java
```diff
@@ -18,6 +18,17 @@
...
+    /**
+     * Stops a lifecycle component.
+     *
+     * <p>
+     * Note: in most cases the correct context has to be set before stopping the component. 
+     * Special attention has to be paid to the classloader from which the component is 
+     * stopped.
+     * </p>
+     * 
+     * @throws MuleException exception during the start process.
```
elrodro83: stop process

core/src/main/java/org/mule/api/lifecycle/Stoppable.java
```diff
@@ -18,6 +18,17 @@
...
-    
+
+    /**
+     * Stops a lifecycle component.
+     *
+     * <p>
+     * Note: in most cases the correct context has to be set before stopping the component. 
+     * Special attention has to be paid to the classloader from which the component is 
+     * stopped.
+     * </p>
```
elrodro83: this isn't needed

PR #7282, opened 2018-10-02
MULE-15735: Error resolving value provider from studio when there is …

gsfernandes:
…a dependency with provided scope

### DIFF
--- a/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/DeployableMavenClassLoaderModelLoader.java
+++ b/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/DeployableMavenClassLoaderModelLoader.java
@@ -63,7 +63,7 @@ public class DeployableMavenClassLoaderModelLoader extends AbstractMavenClassLoa
 
   @Override
   protected boolean includeProvidedDependencies(ArtifactType artifactType) {
-    return artifactType.equals(APP);
+    return artifactType.equals(APP) || artifactType.equals(DOMAIN);
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/DeployableMavenClassLoaderModelLoader.java
```diff
@@ -63,7 +63,7 @@ protected boolean includeTestDependencies(Map<String, Object> attributes) {
 
   @Override
   protected boolean includeProvidedDependencies(ArtifactType artifactType) {
-    return artifactType.equals(APP);
+    return artifactType.equals(APP) || artifactType.equals(DOMAIN);
```
elrodro83: not policies?

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/artifact/AbstractDeployableDescriptorFactory.java
```diff
@@ -65,8 +70,14 @@ protected void doDescriptorConfig(M artifactModel, T descriptor, File artifactLo
     Set<ArtifactPluginDescriptor> pluginDescriptors = new HashSet<>();
     for (BundleDependency bundleDependency : descriptor.getClassLoaderModel().getDependencies()) {
       if (bundleDependency.getDescriptor().isPlugin()) {
-        File pluginFile = new File(bundleDependency.getBundleUri());
-        pluginDescriptors.add(artifactPluginDescriptorLoader.load(pluginFile, descriptor));
+        if (bundleDependency.getBundleUri() == null) {
+          LOGGER.warn(String.format(
```
elrodro83: static import

PR #7285, opened 2018-10-02
MULE-15769: IMAP Reconnection Policy not working

plbalbi:


core/src/main/java/org/mule/transport/AbstractConnector.java
```diff
@@ -1303,10 +1350,11 @@ public void registerListener(InboundEndpoint endpoint,
             {
                 receiver.start();
             }
-        }
-        finally
+
+        } finally
```
fsgonz: formatter.

PR #7287, opened 2018-10-02
MULE-15749: API dependency resolution doesn't take into account pom r…

afelisatti:
…epositories


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/AbstractMavenClassLoaderModelLoader.java
```diff
@@ -289,6 +299,13 @@ protected abstract LightweightClassLoaderModelBuilder newLightweightClassLoaderM
     return missingApiDependencies.stream().map(this::convertBundleDependency).collect(toSet());
   }
 
+  private File getPom(org.mule.maven.client.api.model.BundleDependency dependency) {
+    String zipPath = dependency.getBundleUri().getPath();
+    String dependencyPath = zipPath.substring(0, zipPath.lastIndexOf("/"));
```
gsfernandes: This will work on Windows, right? I mean, URI getPath would not be OS dependent, no? Check that just to make sure that is fine.

PR #7290, opened 2018-10-04
MULE-15777: Relative log config file path should be resolved consistently regardless of the OS

GustavoARod:
- Relative paths are now being resolved based on mule.home folder in a
consistent way for both, Unix and Windows

### DIFF
--- a/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/ApplicationDescriptorFactory.java
+++ b/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/ApplicationDescriptorFactory.java
@@ -6,6 +6,7 @@
  */
 package org.mule.runtime.module.deployment.impl.internal.application;
 
+import static org.mule.runtime.container.api.MuleFoldersUtil.getMuleHomeFolder;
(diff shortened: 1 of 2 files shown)


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/ApplicationDescriptorFactory.java
```diff
@@ -46,7 +48,9 @@ protected void doDescriptorConfig(MuleApplicationModel artifactModel, Applicatio
                                     File artifactLocation) {
     super.doDescriptorConfig(artifactModel, descriptor, artifactLocation);
     if (artifactModel.getLogConfigFile() != null) {
-      descriptor.setLogConfigFile(new File(artifactModel.getLogConfigFile()));
+      Path logConfigFilePath = new File(artifactModel.getLogConfigFile()).toPath();
+      Path muleHomeFolderPath = getMuleHomeFolder().toPath();
+      descriptor.setLogConfigFile(muleHomeFolderPath.resolve(logConfigFilePath).toFile());
```
elrodro83: can you add a unit test to ensure this works fine with absolute paths?

PR #7296, opened 2018-10-05
MULE-15788 More details when DefaultExpressionManager.validateExpression fails

rscadrde:
Add more details.

https://support.mulesoft.com/s/case/50034000015fyNqAAI/orgmuleexpressiondefaultexpressionmanagerisvalidexpressionstring-hides-details

### DIFF
--- a/core/src/main/java/org/mule/expression/DefaultExpressionManager.java
+++ b/core/src/main/java/org/mule/expression/DefaultExpressionManager.java
@@ -103,7 +103,7 @@ public class DefaultExpressionManager implements ExpressionManager, MuleContextA
 
     /**
      * Checks whether an evaluator is registered with the manager
-     * 
+     *
      * @param name the name of the expression evaluator
      * @return true if the evaluator is registered with the manager, false otherwise
      */
@@ -114,7 +114,7 @@ public class DefaultExpressionManager implements ExpressionManager, MuleContextA
 
     /**
      * Checks whether an enricher is registered with the manager
-     * 
+     *
      * @param name the name of the expression enricher
      * @return true if the enricher is registered with the manager, false otherwise
      */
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mule/expression/DefaultExpressionManager.java
```diff
@@ -598,6 +598,9 @@ public boolean isValidExpression(String expression)
         catch (InvalidExpressionException e)
         {
             logger.warn(e.getMessage());
+            if (logger.isDebugEnabled()) {
+            	logger.debug("Exception when checking '"+expression+"'", e);
```
afelisatti: This should be something like `"Expression '" + expression + "' is invalid."`

core/src/main/java/org/mule/expression/DefaultExpressionManager.java
```diff
@@ -656,7 +659,7 @@ public Object match(String token)
         else if (!match.get())
         {
             throw new InvalidExpressionException(expression,
-                "Expression string is not an expression.  Use isExpression(String) to validate first");
+                "Expression '"+expression+"' is not an expression.  Use isExpression(String) to validate first");
```
afelisatti: I don't think the "string" part should be removed, because otherwise it sounds like the expression is not an expression which is confusing.

afelisatti:
@rscadrde, thank you for this contribution. Can you send the PR against the `mule-3.x` branch? We will cherry-pick it to other versions. Please also check the code format throughout to match ours (braces, spaces and so on). Also, we will need you to sign our contributors agreement before merging this (https://github.com/mulesoft/contributor-agreements).

core/src/main/java/org/mule/expression/DefaultExpressionManager.java
rscadrde: (y)

rscadrde:
I limited the change to the isValidExpression method.

See https://github.com/mulesoft/mule/pull/7302

PR #7302, opened 2018-10-05
MULE-15788 More details when DefaultExpressionManager.validateExpression fails

rscadrde:
MULE-15788 More details when DefaultExpressionManager.validateExpression fails

added expression string in case InvalidExpressionException is thrown
and allow logging complete stacktrace on debug level.

https://support.mulesoft.com/s/case/50034000015fyNqAAI/orgmuleexpressiondefaultexpressionmanagerisvalidexpressionstring-hides-details

### DIFF
--- a/core/src/main/java/org/mule/expression/DefaultExpressionManager.java
+++ b/core/src/main/java/org/mule/expression/DefaultExpressionManager.java
@@ -597,7 +597,7 @@ public class DefaultExpressionManager implements ExpressionManager, MuleContextA
         }
         catch (InvalidExpressionException e)
         {
-            logger.warn("Expression '" + expression + "' is invalid: " + e.getMessage());
+            logger.warn(e.getMessage());
             if (logger.isDebugEnabled())
             {
             	logger.debug("Expression '" + expression + "' is invalid.", e);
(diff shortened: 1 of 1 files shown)


rscadrde:
Follow up for https://github.com/mulesoft/mule/pull/7296

core/src/main/java/org/mule/expression/DefaultExpressionManager.java
```diff
@@ -597,7 +597,11 @@ public boolean isValidExpression(String expression)
         }
         catch (InvalidExpressionException e)
         {
-            logger.warn(e.getMessage());
+            logger.warn("Expression '" + expression + "' is invalid: " + e.getMessage());
```
afelisatti: This will result in 2 almost identical log lines in some scenarios. I think it's best to just keep the debug one.

rscadrde: Fine for me.

afelisatti:
Thanks for the contribution, @rscadrde. I'll be cherry-picking this to 3.9.x and 3.8.x soon.

PR #7307, opened 2018-10-09
EE-6323: Introduce WebSockets connector

marianogonzalez:


modules/http-api/src/main/java/org/mule/runtime/http/api/server/ws/WebSocketHandlerManager.java
```diff
@@ -0,0 +1,37 @@
...
+ * @since 4.1.5 as experimental
+ */
+@NoImplement
+@Experimental
+public interface WebSocketHandlerManager {
+
+  /**
+   * Allows access to the handler.
+   */
+
```
afelisatti: bad java doc

modules/http-api/src/main/java/org/mule/runtime/http/api/ws/WebSocket.java
```diff
@@ -0,0 +1,108 @@
...
+
+  /**
+   * @return The {@link URI} of the endpoint that received the originating HTTP request.
+   */
+  URI getUri();
+
+  /**
+   * @return an immutable list with the groups to which {@code this} socket belongs to. Maybe empty but never null.
+   */
+  List<String> getGroups();
```
afelisatti: Why does the socket needs this? Sounds like some other entity should be handling the grouping.

modules/http-api/src/main/java/org/mule/runtime/http/api/ws/WebSocketCloseCode.java
```diff
@@ -0,0 +1,79 @@
...
+   * Indicates that an endpoint is terminating the connection due to a protocol error.
+   */
+  PROTOCOL_ERROR(1002),
+  /**
+   * Indicates that an endpoint is terminating the connection because it has received a type of data it cannot accept
+   * (e.g. an endpoint that understands only text data may send this if it receives a binary message.)
+   */
+  INVALID_DATA(1003),
+  /**
+   * indicates that an endpoint is terminating the connection because it has received a message that is too large.
```
afelisatti: `Indicates`

modules/http-api/src/main/java/org/mule/runtime/http/api/ws/WebSocketProtocol.java
```diff
@@ -0,0 +1,53 @@
...
+import org.mule.api.annotation.Experimental;
+
+/**
+ * The supported protocols
+ */
+@Experimental
+public enum WebSocketProtocol {
+
+  /**
+   * {@code ws} protocol. Used then the Upgrade request was sent over standard HTTP
```
afelisatti: `Used when`?

modules/http-api/src/main/java/org/mule/runtime/http/api/ws/WebSocketProtocol.java
```diff
@@ -0,0 +1,53 @@
...
+   */
+  WS("ws"),
+
+  /**
+   * {@code wss} protocol. Used then the Upgrade request was sent over standard HTTPS
```
afelisatti: Same

modules/http-api/src/main/java/org/mule/runtime/http/api/ws/FragmentHandler.java
```diff
@@ -0,0 +1,51 @@
...
+ * Allows to channel the information of a multi fragment Web Socket message into an {@link InputStream}.
+ * <p>
+ * Clients <b>MUST</b> always invoke either {@link #complete()} or {@link #abort()} once the multi fragment
+ * message is over.
+ *
+ * @since 4.1.5 as experimental
+ */
+@Experimental
+@NoImplement
+public interface FragmentHandler {
```
afelisatti: Why is this part of the API? I don't see any references so I can't understand why you would need it at this level.

modules/http-api/src/main/java/org/mule/runtime/http/api/client/ws/WebSocketCallback.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * A Callback to receive outbound {@link WebSocket} events.
+ * <p>
+ * Implementations <b>MUST</b> be thread safe
+ *
+ * @since 4.1.5 as experimental
+ */
+@Experimental
+public interface WebSocketCallback {
```
afelisatti: How would the callback relate to the future? Particularly in regards to `onConnect`.

modules/http-api/src/main/java/org/mule/runtime/http/api/exception/InvalidStatusCodeException.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.mule.api.annotation.Experimental;
+
+/**
+ * Exceptions to be thrown when a received status code was not between the expected range
+ *
+ * @since 4.1.5 as experimental
+ */
+@Experimental
+public class InvalidStatusCodeException extends Exception {
```
afelisatti: Where is this used? Why isn't it reference in any java docs?

modules/http-api/src/main/java/org/mule/runtime/http/api/server/ws/WebSocketConnectionHandler.java
```diff
@@ -0,0 +1,48 @@
...
+  String getSocketId(WebSocketRequest request);
+
+  /**
+   * Invoked when the {@code socket} is established
+   *
+   * @param socket  the established {@link WebSocket}
+   * @param request the inbound {@link WebSocketRequest request}
+   * @throws WebSocketConnectionRejectedException if the connection is refused by the owning server
+   */
+  void onConnect(WebSocket socket, WebSocketRequest request) throws WebSocketConnectionRejectedException;
```
afelisatti: This will be triggered each time a new web socket is created as a result of an incoming request? How does this relate to `getSocketId`? Seems like first, a socket id will be established since that is needed for a `WebSocket` instance so that would actually execute before the `onConnect`?

afelisatti:
@marianogonzalez You will need to change the associated issue to a MULE one.

modules/http-api/src/main/java/org/mule/runtime/http/api/client/DefaultHttpRequestOptions.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.util.Optional;
+
+/**
+ * Basic implementation of {@link HttpRequestOptions}. Instances can only be obtained through {@link HttpRequestOptionsBuilder}.
+ *
+ * @since 4.1.5 as experimental. GA in 4.2
+ */
+@Experimental
+class DefaultHttpRequestOptions implements HttpRequestOptions {
```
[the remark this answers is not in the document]
marianogonzalez: ?

modules/http-api/src/main/java/org/mule/runtime/http/api/domain/message/MessageWithHeaders.java
```diff
@@ -0,0 +1,62 @@
...
+  String getHeaderValue(String headerName);
+
+  /**
+   * @param headerName name of the header
+   * @return first value of the header, regardless of the case
+   *
+   * @deprecated The underlying implementation is already case-insensitive. Use {@link #getHeaderValue(String)}
+   */
+  @Deprecated
+  String getHeaderValueIgnoreCase(String headerName);
```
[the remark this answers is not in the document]
marianogonzalez: This is just extracting the contracts we currently have. If these methods are api, we have to keep them. We could disaggregate this further to have a `clean` api defined, but I think that would only add additional complexity

modules/http-api/src/main/java/org/mule/runtime/http/api/ws/FragmentHandler.java
marianogonzalez: because I suck. moved to EE

modules/http-api/src/main/java/org/mule/runtime/http/api/client/ws/WebSocketCallback.java
marianogonzalez: it doesn't. Grizzly completes the future after these methods are executed. It's a good thing that these callbacks are decoupled from the threading model

modules/http-api/src/main/java/org/mule/runtime/http/api/exception/InvalidStatusCodeException.java
marianogonzalez: it's used in the EE impl and in the connector as well

modules/http-api/src/main/java/org/mule/runtime/http/api/server/ws/WebSocketConnectionHandler.java
marianogonzalez: getSocketId is invoked first. This is invoked after.

modules/http-api/src/main/java/org/mule/runtime/http/api/client/DefaultHttpRequestOptions.java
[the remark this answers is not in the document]
afelisatti: I didn't know you moved stuff from 4.2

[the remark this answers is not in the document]
marianogonzalez: there's no other way of ensuring compatibility between the 4.1.5 and 4.2 versions of it

PR #7308, opened 2018-10-09
MULE-15434: Add metadata category and resolver name to components

alegmarra:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/DynamicMetadataDeclarationEnricher.java
```diff
@@ -152,9 +175,10 @@ private void enrichWithDsql(OperationDeclaration declaration, MethodElement meth
                                                                                                                        .nativeOutputResolver(),
                                                                                                                    query
                                                                                                                        .entityResolver())));
+
       addQueryModelProperties(declaration, query);
       declareDynamicType(declaration.getOutput());
-      declareMetadataKeyId(declaration, new NullMetadataScopeAdapter());
+      declareMetadataKeyId(declaration, null);
```
juandesi: is this ok?

juandesi: I mean the “null”

alegmarra: sadly, it is, we use null everywhere to mark a non-existing categoryName

juandesi: is this allowed in the model? when is the category name absent?

PR #7318, opened 2018-10-12
MULE-15731 - Policy can not contain duplicated plugin with greater ve…

pablolagreca:
…rsion than the apps one


modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/internal/plugin/BundlePluginDependenciesResolver.java
```diff
@@ -82,8 +84,10 @@ public BundlePluginDependenciesResolver(ArtifactDescriptorFactory<ArtifactPlugin
 
       if (!pluginDescriptor.isPresent()) {
         filteredPluginDescriptors.add(appPluginDescriptor);
-      } else if (!isCompatibleVersion(pluginDescriptor.get().getBundleDescriptor().getVersion(),
-                                      appPluginDescriptor.getBundleDescriptor().getVersion())) {
+      } else if (pluginDescriptor.get().getBundleDescriptor().getArtifactId().equals(MULE_HTTP_CONNECTOR_ARTIFACT_ID) &&
```
elrodro83: extract `pluginDescriptor.get().getBundleDescriptor()` to a variable for readability

PR #7322, opened 2018-10-12
MULE-15674: Http Service should preserve POST request body on dancing request.

fsgonz:


modules/http/src/main/java/org/mule/module/http/internal/request/RequestResourcesUtils.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.mule.module.http.internal.domain.InputStreamHttpEntity;
+import org.mule.module.http.internal.domain.request.HttpRequest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Utils to manage request resources
+ * 
+ * @since 3.9.10
```
afelisatti: Check the version, also, we will likely need to cherry pick this

fsgonz: I've updated the date. I'll take this into account to use the correct version on cherry-picking.

PR #7339, opened 2018-10-19
MULE-15870: MuleUniversalConduit not respecting response encoding when reading response stream

fsgonz:


modules/cxf/src/main/java/org/mule/module/cxf/support/CxfUtils.java
```diff
@@ -119,4 +128,17 @@ public static String mapUnsupportedSchemas(String url)
         return url;
     }
 
+    public static String resolveEncoding(MuleMessage result, String contentType)
+    {
+        final MediaType mediaType = MediaType.parse(contentType);
+        String encoding = mediaType.charset().isPresent() ? mediaType.charset().get().name() : Charset.defaultCharset().name();
+
+        if (encoding == null)
```
afelisatti: When will this be null? Sounds like this should be used instead of the default charset before.

PR #7341, opened 2018-10-23
MULE-15728: Resource is not found when declaring extension in domain

elrodro83:
project.
* Fix tests


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultConnectionProviderObjectBuilder.java
```diff
@@ -111,7 +112,9 @@ public DefaultConnectionProviderObjectBuilder(ConnectionProviderModel providerMo
     // Use the classloader from the app instead of the extension for calling the provider methods.
     // The provider may need to access resources from the application defined in its config. To do so, it needs the app
```
juandesi: Change this comment to be more apecific on the changes introduced, also dont forget to change the method name

PR #7344, opened 2018-10-23
MULE-15905: Support DW to be absent

marianogonzalez:


core/src/main/java/org/mule/runtime/core/internal/el/ExpressionLanguageAdaptorHandler.java
```diff
@@ -74,7 +71,7 @@ public ExpressionLanguageAdaptorHandler(ExtendedExpressionLanguageAdaptor defaul
 
     exprPrefixPattern = compile(EXPR_PREFIX_PATTERN_TEMPLATE.replaceAll("LANGS", join(expressionLanguages.keySet(), '|')));
 
-    melDefault = valueOf(getProperty(MULE_MEL_AS_DEFAULT, "false"));
+    melDefault = MuleProperties.isMelDefault();
```
elrodro83: remove `isMelDefault` method and `melDefault` attribute from this class and use the utils directly.

PR #7345, opened 2018-10-23
MULE-15908: Persistent ObjectStore doesn't regenerate the partition descriptor file after a clear

estebanwasinger:


core/src/main/java/org/mule/runtime/core/internal/util/store/PersistentObjectStorePartition.java
```diff
@@ -139,6 +139,7 @@ public void clear() throws ObjectStoreException {
     synchronized (realKeyToUUIDIndex) {
       try {
         cleanDirectory(this.partitionDirectory);
+        open();
```
alegmarra: I think you only need the `createOrRetrievePartitionDescriptorFile();` part of the open

alegmarra: also, does it make sense to delete the descriptor in the first place? or instead we need to whitelist the descriptor file?

core-tests/src/test/java/org/mule/runtime/core/internal/util/store/PersistentObjectStorePartitionTestCase.java
```diff
@@ -100,11 +101,13 @@ public void clear() throws Exception {
...
 
     partition.store(KEY, VALUE);
-    assertTrue(partition.contains(KEY));
-    assertEquals(VALUE, partition.retrieve(KEY));
+    assertThat(partition.contains(KEY), is(true));
+    assertThat(VALUE, is(partition.retrieve(KEY)));
 
     partition.clear();
-    assertFalse(partition.contains(KEY));
+    assertThat(partition.contains(KEY), is(false));
```
marianogonzalez: This is testing the implementation not the behavior. What you should try to reproduce is that without your fix, writing to a store after clear fails and that afterwards you can successfully write and read from a cleared object store

core/src/main/java/org/mule/runtime/core/internal/util/store/PersistentObjectStorePartition.java
marianogonzalez: Check if mule 3 doesn't have the same bug. My guess is it does

PR #7356, opened 2018-10-25
MULE-15931: Mule fails to load all contracts in a service bundle

marianogonzalez:


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/artifact/ContainerServicesMuleContextConfigurator.java
```diff
@@ -34,7 +35,13 @@ public ContainerServicesMuleContextConfigurator(ServiceRepository serviceReposit
 
   @Override
   public void configure(CustomizationService customizationService) {
-    serviceRepository.getServices()
-        .forEach(service -> customizationService.registerCustomServiceImpl(service.getName(), service));
+    serviceRepository.getServices().forEach(service -> {
+      String name = service.getName();
+      String contract = service.getContractName();
```
afelisatti: When is the contract name anything but the default?

marianogonzalez: look at LazyServiceProxy and the Service Javadoc

PR #7359, opened 2018-10-26
MULE-15493: OAuth2 Request config always fetches token at startup

elrodro83:
Addendum: Only continue deployment for connectivity/networking errors, not for errors generated by the token server

### DIFF
--- a/modules/oauth/src/main/java/org/mule/module/oauth2/internal/OAuthTokenMuleExeption.java
+++ b/modules/oauth/src/main/java/org/mule/module/oauth2/internal/OAuthTokenMuleExeption.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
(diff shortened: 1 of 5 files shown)


modules/oauth/src/main/java/org/mule/module/oauth2/internal/OAuthTokenMuleExeption.java
```diff
@@ -0,0 +1,26 @@
...
+package org.mule.module.oauth2.internal;
+
+import org.mule.api.DefaultMuleException;
+
+/**
+ * Standard exception wrapper for token response processing problems. 
+ * 
+ * @since 3.10
+ */
+public class OAuthTokenMuleExeption extends DefaultMuleException
```
afelisatti: `Exeption` -> `Exception`

modules/oauth/src/main/java/org/mule/module/oauth2/internal/OAuthTokenMuleExeption.java
```diff
@@ -0,0 +1,26 @@
...
+{
+
+    private static final long serialVersionUID = 1L;
```
afelisatti: wat

PR #7364, opened 2018-10-28
MULE-15691 - Define and implement how we can convert a mule-applicati…

pablolagreca:
…on into a reusable plugin for apps

### DIFF
--- a/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/LightweightClassLoaderModelBuilder.java
+++ b/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/LightweightClassLoaderModelBuilder.java
@@ -11,10 +11,10 @@ import static java.util.Optional.of;
 import static java.util.stream.Collectors.toList;
 import static org.apache.commons.io.FileUtils.toFile;
 import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
+import static org.mule.runtime.container.api.MuleFoldersUtil.getExecutionFolder;
 import org.mule.maven.client.api.MavenClient;
 import org.mule.runtime.api.exception.MuleRuntimeException;
 import org.mule.runtime.api.meta.MuleVersion;
-import org.mule.runtime.container.api.MuleFoldersUtil;
 import org.mule.runtime.module.artifact.api.descriptor.BundleDependency;
 import org.mule.runtime.module.artifact.api.descriptor.BundleDescriptor;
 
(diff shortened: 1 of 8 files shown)


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/LightweightClassLoaderModelBuilder.java
```diff
@@ -50,4 +69,84 @@ public LightweightClassLoaderModelBuilder(File artifactFolder,
         .map(bundleDescriptor -> mavenClient.resolveBundleDescriptor(bundleDescriptor))
         .collect(toList());
   }
+
+  @Override
+  protected Map<BundleDescriptor, Set<BundleDescriptor>> doProcessAdditionalPluginLibraries(Plugin packagingPlugin) {
+    File temporaryPomFolder = new File(MuleFoldersUtil.getExecutionFolder(), "mavenTemp");
```
lucianoRM: static import

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/LightweightClassLoaderModelBuilder.java
```diff
@@ -50,4 +69,84 @@ public LightweightClassLoaderModelBuilder(File artifactFolder,
...
+            pluginAdditionalLibrariesMap.forEach((pluginDescriptor, additionalLibraries) -> {
+              Set<BundleDescriptor> effectivePluginAdditionalLibraries = new HashSet<>();
+              if (deployableArtifactAdditionalLibrariesMap.containsKey(pluginDescriptor)) {
+                effectivePluginAdditionalLibraries.addAll(additionalLibraries.stream()
+                    .filter(additionalLibrary -> {
+                      boolean additionalLibraryDefinedAtDeployableArtifact =
+                          existsInLibrariesMap(deployableArtifactAdditionalLibrariesMap, pluginDescriptor, additionalLibrary);
+                      if (!additionalLibraryDefinedAtDeployableArtifact) {
+                        Optional<BundleDescriptor> additionalLibraryDefinedByAnotherPlugin =
+                            findLibraryInAdditionalLibrariesMap(effectivePluginsAdditionalLibrariesMap, pluginDescriptor,
```
lucianoRM: Check my comment on mule-maven-plugin tests.
For what i can understand, an app that imports other reusable apps will behave as one. Is that correct? 
Consider App "A" and reusable app "R".
If R depends on a plugin X, then A will be able to "see" that plugin X as if it were declared in it's own pom, right? 
What happens if both A and R declare the same plugin but with different versions? 

I think that what happens in those scenarios will impact on whether or not this change is correct.

pablolagreca: What's going to happen is that the newest version of the plugin will be used. Same as if the app and another plugin have the same dependency with different versions.

lucianoRM: But i think that is not what happens. 
If an app and a plugin declare the same dependency with different versions, both will live in their own classLoader. However, if that dependency is a mule-plugin, then, it will happen as you said and the newest one will prevail. 
@gsfernandes care to comment on this?

pablolagreca: That's exactly what happens. If you declare an AD Z_v1 for Plugin X and another AD Z_v2 for another Plugin Y then there's no problem at all. Z_v1 will be added to Plugin X CL and Z_v2 will be added to Plugin Y.

PR #7378, opened 2018-10-31
MULE-15888:Malformed URL returns 500 instead of 400.

emariotti3:


modules/http-api/src/main/java/org/mule/runtime/http/api/utils/RequestMatcherRegistry.java
```diff
@@ -101,6 +101,15 @@ default T find(String method, String path) {
      */
     RequestMatcherRegistryBuilder<T> onNotFound(Supplier<T> itemSupplier);
 
+
+    /**
+     * Determines which item should be returned if an invalid method is provided. If not set, {@code null} will be returned by default.
```
afelisatti: invalid request?

modules/http-api/src/main/java/org/mule/runtime/http/api/server/HttpRequestParsingException.java
```diff
@@ -0,0 +1,19 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.http.api.server;
+
+/**
+ * Exception thrown when there is a problem processing an incoming HttpRequest to generate a MuleEvent.
+ *
+ * @since 1.2
+ */
+public class HttpRequestParsingException extends Exception {
```
afelisatti: If this is not used then the hierarchy is too much. We can have just `DecodingException` for now and create the hierarchy later if it makes sense.

modules/http-api/src/main/java/org/mule/runtime/http/api/server/DecodingException.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+package org.mule.runtime.http.api.server;
+
+/**
+ * {@code DecodingException} Is an exception thrown when there is attempt to decode a malformed
+ * or invalid text, url or url parameter.
+ *
+ * @since 1.2
+ */
+public class DecodingException extends HttpRequestParsingException {
```
afelisatti: Move to the service implementation, everything we put in the API we'll have to maintain backwards compatibility with for years to come. It's best too only move here (the API) necessary bits that require being exposed and keep everything else internal since it allows us to change it as we see fit whenever we want to.

PR #7382, opened 2018-10-31
MULE-15586: Add a timeout to lookup function

elrodro83:
* Fir for classcastexception on some scenarios


afelisatti:
no test for the scenarios?

PR #7400, opened 2018-11-05
 MULE-15933: File connector fails to recreate the input stream to a file when reconnection occurs in following components

ndinu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ResolverUtils.java
```diff
@@ -139,6 +140,92 @@ public static Object resolveCursor(Object value) {
...
+  /**
+   * Obtains a {@link CursorStream} that cannot be closed based on the {@code value}, if one is available.
+   *
+   * @return the given {@code value}, but wrapped in a way that it is not posible to close it.
+   */
+  public static Object resolveToUnclosableCursor(CursorStream value) {
+    return new UnclosableCursorStream(value);
+  }
+
+  private static class UnclosableCursorStream extends CursorStream {
```
juandesi: I would extract this from
this class,  and remove the static method thst creates instances of it.

This could probably live in other place

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ResolverUtils.java
```diff
@@ -139,6 +140,92 @@ public static Object resolveCursor(Object value) {
...
+      return delegate.isReleased();
+    }
+
+    @Override
+    public CursorProvider getProvider() {
+      return delegate.getProvider();
+    }
+
+    @Override
+    public void close() throws IOException {}
```
juandesi: add detailed comment tomthis method, so following devs will understand why this is no implemented

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/MethodArgumentResolverDelegate.java
```diff
@@ -246,6 +249,8 @@ private void initArgumentResolvers() {
       Object parameterValue = valueSupplier.get();
       if (parameterValue == null) {
         return resolvePrimitiveTypeDefaultValue(parameterType);
+      } else if (parameterType.equals(InputStream.class) && parameterValue instanceof CursorStream) {
```
juandesi: should this wrapping be done for other InputStream impls that are not ours?

ndinu: I would assume that we are the only ones implementing CursorStream, do you think of a valid case where someone else is implementing it?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ResolverUtils.java
```diff
@@ -139,6 +140,92 @@ public static Object resolveCursor(Object value) {
     return value;
   }
 
+  /**
+   * Obtains a {@link CursorStream} that cannot be closed based on the {@code value}, if one is available.
+   *
+   * @return the given {@code value}, but wrapped in a way that it is not posible to close it.
+   */
+  public static Object resolveToUnclosableCursor(CursorStream value) {
```
estebanwasinger: `value` doesn't seems to be a good name

tests/test-extensions/petstore-extension/src/main/java/org/mule/test/petstore/extension/PetStoreOperations.java
```diff
@@ -54,6 +59,27 @@
     return client.getPets(ownerName, config);
   }
 
+  @MediaType(TEXT_PLAIN)
+  public InputStream getStreamedSignature(String signature) {
+    return new ByteArrayInputStream(signature.getBytes());
+  }
+
+  public List<String> getPetsWithIntermitentConnectionProblemAndClosingStream(@Connection PetStoreClient client,
```
estebanwasinger: Just to be sure that the reconnection is done, could you return some value, like a flag telling that the reconnection has been executed, just to be sure that if in some moment the test is changed and no reconnection is done the test will be still 🍏

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/MethodArgumentResolverDelegate.java
estebanwasinger: Makes any sense to ask if the parameter type is a InputStream?

ndinu: solved

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ResolverUtils.java
ndinu: added

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ResolverUtils.java
ndinu: solved

tests/test-extensions/petstore-extension/src/main/java/org/mule/test/petstore/extension/PetStoreOperations.java
ndinu: adding

PR #7406, opened 2018-11-05
MULE-15990: ExpressionExecutionException should be mapped as Expression error type

mbuchwald:


core/src/main/java/org/mule/runtime/core/internal/exception/ErrorTypeLocatorFactory.java
```diff
@@ -70,6 +71,7 @@ public static ErrorTypeLocator createDefaultErrorTypeLocator(ErrorTypeRepository
             .addExceptionMapping(MessageTransformerException.class, errorTypeRepository.lookupErrorType(TRANSFORMATION).get())
             .addExceptionMapping(TransformerException.class, errorTypeRepository.lookupErrorType(TRANSFORMATION).get())
             .addExceptionMapping(ExpressionRuntimeException.class, errorTypeRepository.lookupErrorType(EXPRESSION).get())
+            .addExceptionMapping(ExpressionExecutionException.class, errorTypeRepository.lookupErrorType(EXPRESSION).get())
```
afelisatti: This expression should not be listed here. DW as a language uses it since that's the API definition but we should catch it and wrap it in ours (the one above).

PR #7410, opened 2018-11-07
MULE-16014: CXF Proxy terminates requests silently when soap service payload response is empty

plbalbi:


modules/cxf/src/main/java/org/mule/module/cxf/transport/MuleUniversalConduit.java
```diff
@@ -12,6 +12,7 @@
 import static org.mule.module.cxf.support.CxfUtils.resolveEncoding;
 import static org.mule.transformer.types.DataTypeFactory.XML_STRING;
 import static org.mule.transport.http.HttpConnector.HTTP_DISABLE_STATUS_CODE_EXCEPTION_CHECK;
+import static org.mule.transport.http.HttpConstants.*;
```
fsgonz: Import only the needed constants even if they are all of them.

modules/cxf/src/main/java/org/mule/module/cxf/transport/MuleUniversalConduit.java
```diff
@@ -345,6 +350,18 @@ private void sendResultBackToCxf(Message m, MuleEvent resEvent) throws Transform
                 clearClientContextIfNeeded(getMessageObserver());
                 return;
             }
+            // If this branch is reached, it means that the message payload is empty.
+            // By asserting that the inboundProperty 'http.status' exists, I'm verifying that
+            // the last inbound endpoint used was HTTP, and that it set the property correctly.
+            else if (resEvent.getMessage().getInboundPropertyNames().contains("http.status"))
+            {
+                int httpResponseStatusCode = Integer.valueOf(resEvent.getMessage().getInboundProperty("http.status").toString());
```
fsgonz: Is there a constant for http.status? Use that in case it exists. static method if possible.

plbalbi: I couldn't found a static property in a class worth importing. I moved the field name to static field.

PR #7417, opened 2018-11-08
MULE-16025: App directory is deleted in undeployment

lucianoRM:


modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/DefaultArchiveDeployer.java
```diff
@@ -336,8 +336,9 @@ private void undeployArtifact(T artifact, boolean removeData) {
       artifactArchiveInstaller.uninstallArtifact(artifact.getArtifactName());
       if (removeData) {
```
gsfernandes: Should you just change the logic when this kind of redeploy happens to call this method with removeData=false?

modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/DefaultArchiveDeployer.java
```diff
@@ -336,8 +336,9 @@ private void undeployArtifact(T artifact, boolean removeData) {
...
         final File dataFolder = getAppDataFolder(artifact.getDescriptor().getDataFolderName());
+        final File tmpDir = new File(dataFolder, APP_TMP_DIR);
         try {
-          deleteDirectory(dataFolder);
+          deleteDirectory(tmpDir);
```
pablolagreca: Maybe that's what is happening here but just to check: If you are redeploying you should not remove any temp data (tmp or .mule/). If you are undeploying you should remove all the temp data (tmp and .mule/)

PR #7420, opened 2018-11-08
MULE-15918: Test and complete additionalPluginLibraries feature

lucianoRM:


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/LightweightClassLoaderModelBuilder.java
```diff
@@ -57,17 +65,82 @@ public LightweightClassLoaderModelBuilder(File artifactFolder,
...
+        .findFirst()
+        .map(additionalBundleDependency -> {
+          String additionalBundleDependencyVersion = additionalBundleDependency.getDescriptor().getVersion();
+          String bundleDependencyVersion = bundleDependency.getDescriptor().getVersion();
+          if (areSameMajor(bundleDependencyVersion, additionalBundleDependencyVersion)) {
+            if (isNewerVersion(bundleDependencyVersion, additionalBundleDependencyVersion)) {
+              replace.set(additionalBundleDependency);
+            }
+          } else {
+            throw new MuleRuntimeException(createStaticMessage("Attempting to add different major versions of the same dependency as additional plugin dependency. If this is not explicitly defined, check transitive dependencies."
```
gsfernandes: There is no tests for this neither here or the related PR https://github.com/mulesoft/mule-ee-distributions/pull/547. Would it be possible to cover this scenario in an integration test?

lucianoRM: It's covered in the mule-maven-plugin PR, but i will add another tests in mule-ee-distributions

PR #7421, opened 2018-11-09
MULE-16014: CXF Proxy handles SOAP service response with empty payload

plbalbi:


modules/cxf/src/main/java/org/mule/module/cxf/CxfCannotProcessEmptyPayloadException.java
```diff
@@ -0,0 +1,18 @@
...
+ */
+package org.mule.module.cxf;
+
+public class CxfCannotProcessEmptyPayloadException extends Exception
+{
+    private int responseCode;
+
+    public CxfCannotProcessEmptyPayloadException(Integer responseStatusCode)
+    {
+        super("CXF cannot process and empty payload. Invoked service responded with status code: " + responseStatusCode.toString());
```
elrodro83: and -> an

elrodro83: I would rephrase this to something like `Invoked service responded with status code 'code' and an empty body`.

PR #7428, opened 2018-11-12
MULE-16037: JreExplorer no properly determines JRE libs in Java 11

elrodro83:


modules/container/src/main/java/org/mule/runtime/container/internal/JreExplorer.java
```diff
@@ -45,15 +52,30 @@ private JreExplorer() {}
...
+    // These are present in JDK 8
+    addJdkPath(jdkPaths, "sun.boot.class.path");
+    addJdkPath(jdkPaths, "java.ext.dirs");
+
+    // These are present in JDK 9, 10, 11
+    addJdkPath(jdkPaths, "sun.boot.library.path");
+    addJdkPath(jdkPaths, "java.library.path");
+
+    if (jdkPaths.isEmpty()) {
+      LOGGER.warn("No JDK path/dir system propety found. Defaulting to the whole classpath."
```
aiannucci:
```suggestion
      LOGGER.warn("No JDK path/dir system property found. Defaulting to the whole classpath."
```

PR #7434, opened 2018-11-13
MULE-14844: SDK should validate that the given default value of an En…

pangelani:
…um parameter is a valid value.


modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/validation/ParameterModelValidatorTestCase.java
```diff
@@ -140,6 +141,21 @@ public void invalidConfigOverrideWithDefault() {
...
+        .expectMessage("Parameter 'url' in the operation 'dummyOperation' has a default value which is not listed as an available option");
+
+    when(invalidParameterModel.isRequired()).thenReturn(false);
+    when(invalidParameterModel.getDefaultValue()).thenReturn("default");
+    when(invalidParameterModel.getType()).thenReturn(toMetadataType(HealthStatus.class));
+    mockParameters(operationModel, invalidParameterModel);
+
+    validate(extensionModel, validator);
+  }
+
```
ndinu: Add a test where the default value is an expression and a test where the default value is a valid enum value.

PR #7450, opened 2018-11-23
MULE-16101: Miscellaneous performance improvements

elrodro83:
* Remove annotated support from streaming objects
* Use `Caffeine` for caching the cursors. Caffeine uses the ForkJoin pool for doing expiration stuff, so no need to pass a CpuIntensive scheduler for that now.

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/streaming/CursorManager.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/streaming/CursorManager.java
@@ -10,12 +10,11 @@ import static java.util.Collections.newSetFromMap;
 import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
 
 import org.mule.runtime.api.exception.MuleRuntimeException;
-import org.mule.runtime.api.scheduler.Scheduler;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/mule/runtime/core/internal/streaming/CursorManager.java
```diff
@@ -141,35 +128,31 @@ private void registerEventContext(BaseEventContext eventContext) {
     private AtomicBoolean disposed = new AtomicBoolean(false);
     private AtomicInteger cursorCount = new AtomicInteger(0);
 
-    private final LoadingCache<CursorProvider, Set<Cursor>> cursors = CacheBuilder.newBuilder()
-        .removalListener((RemovalListener<CursorProvider, Set<Cursor>>) notification -> {
+    private final LoadingCache<CursorProvider, Set<Cursor>> cursors = Caffeine.newBuilder()
+        .removalListener((key, value, cause) -> {
```
ben-manes: FYI, you can add types to a lambda, e.g. 
`(CursorProvider key, Collection<Cursor> value, RemovalCause cause) -> ...`
which then propagates the types into later build stages.

PR #7470, opened 2018-12-03
MULE-16143: Add Javadoc in PartionableObjectStore

lucianoRM:


core/src/main/java/org/mule/api/store/PartitionableExpirableObjectStore.java
```diff
@@ -8,8 +8,25 @@
 
 import java.io.Serializable;
 
+/**
+ * A {@link PartitionableObjectStore} which is also an {@link ExpirableObjectStore}
```
marianogonzalez: add that open(String) must be called before using a non default partition (and also verify if the default partition requires calling open())

afelisatti:
@lucianoRM create a MULE issue for this.

PR #7477, opened 2018-12-05
MULE-16101: Miscellaneous performance improvements

elrodro83:
* Small improvements on usage of reactor in policies support


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/MuleApplicationPolicyProvider.java
```diff
@@ -38,6 +38,8 @@
   private final PolicyInstanceProviderFactory policyInstanceProviderFactory;
   private final List<RegisteredPolicyTemplate> registeredPolicyTemplates = new LinkedList<>();
   private final List<RegisteredPolicyInstanceProvider> registeredPolicyInstanceProviders = new LinkedList<>();
+  private Runnable policiesDeploymentChangeCallback = () -> {
```
afelisatti: Why?

PR #7491, opened 2018-12-12
MULE-15574: Mule Runtime fails to run batch jobs after stopping app w…

marianogonzalez:
…hile debugging

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/util/journal/TransactionJournalFile.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/util/journal/TransactionJournalFile.java
@@ -6,6 +6,7 @@
  */
 package org.mule.runtime.core.internal.util.journal;
 
+import static org.slf4j.LoggerFactory.getLogger;
 import org.mule.runtime.api.exception.MuleRuntimeException;
 
 import com.google.common.collect.LinkedHashMultimap;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/mule/runtime/core/internal/util/journal/TransactionJournalFile.java
```diff
@@ -42,7 +42,7 @@
    */
   private static final int MINIMUM_ENTRIES_TO_CLEAR_FILE = 10000;
 
-  protected transient Logger logger = LoggerFactory.getLogger(getClass());
+  private static Logger LOGGER = getLogger(TransactionJournalFile.class);
```
elrodro83: `final`

core/src/main/java/org/mule/runtime/core/internal/util/journal/queue/LocalTxQueueTransactionRecoverer.java
```diff
@@ -27,7 +26,7 @@
  */
 public class LocalTxQueueTransactionRecoverer {
 
-  private Logger logger = LoggerFactory.getLogger(getClass());
+  private static Logger LOGGER = getLogger(LocalTxQueueTransactionRecoverer.class);
```
elrodro83: `final`

PR #7509, opened 2018-12-14
MULE-16212: Avoid copying the whole event when just setting internal

elrodro83:
variables


afelisatti:
--validate

afelisatti:
--validate

PR #7518, opened 2018-12-18
MULE-14532: fixed thread jump

mbuchwald:


core-tests/src/test/java/org/mule/runtime/core/internal/processor/strategy/ProactorStreamEmitterProcessingStrategyTestCase.java
```diff
@@ -315,7 +315,6 @@ public void singleCpuLightConcurrentMaxConcurrency1() throws Exception {
                                                                                                     1, true)),
                        true, CPU_LITE, 1);
     assertThat(threads, hasSize(1));
-    assertThat(threads, not(hasItem(startsWith(CPU_LIGHT))));
```
elrodro83: if this is expected now, just remove the `not`

elrodro83: or better yet, count the occurrences as `ProactorStreamWorkQueueProcessingStrategyTestCase` does with the ring buffer threads

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/ProactorStreamEmitterProcessingStrategyFactory.java
```diff
@@ -133,12 +133,12 @@ public ProactorStreamEmitterProcessingStrategy(Supplier<Scheduler> ringBufferSch
     public Sink createSink(FlowConstruct flowConstruct, ReactiveProcessor function) {
       final long shutdownTimeout = flowConstruct.getMuleContext().getConfiguration().getShutdownTimeout();
       List<ReactorSink<CoreEvent>> sinks = new ArrayList<>();
-
       int concurrency = maxConcurrency < subscribers ? maxConcurrency : subscribers;
```
elrodro83: doesn't this give the same result as `getParallelism()`?

mbuchwald: No. Subscribers, maxConcurrency and Parallelism are 3 different values. In the factory, the value given for parallelism is max(CORES, maxConcurrency), so it's at least as maxConcurrency. 
After trying different configurations this was the one that got best results in perf tests (and the one that takes into account maxConcurrency... if not, even if the user sets maxConcurrency = 1, the value won't be taken into account)

PR #7527, opened 2018-12-20
MULE-16230: VersionFormatArtifactDescriptorValidator fails due to Mav…

gsfernandes:
…enBundleDescriptorLoader is not considering the effective pom model to read the version property


gsfernandes:
--validate

PR #7539, opened 2018-12-26
Mule 16252

gsfernandes:


modules/service/src/main/java/org/mule/runtime/module/service/internal/manager/ServiceRegistry.java
```diff
@@ -42,23 +47,46 @@
...
       } catch (Exception e) {
         throw new ServiceResolutionError(format("Could not inject dependency on field %s of type %s", field.getName(),
-                                                dependency.getClass().getName()),
+                                                dependencyType.getName()),
                                          e);
       }
     }
   }
 
+  private Object resolveObjectToInject(Class<?> dependencyType, boolean nullToOptional) {
```
marianogonzalez: rename `nullToOptional` to `asOptional`

modules/service/src/main/java/org/mule/runtime/module/service/internal/manager/ServiceRegistry.java
```diff
@@ -42,23 +47,46 @@
    */
   public void inject(ServiceProvider serviceProvider) throws ServiceResolutionError {
     for (Field field : getAllFields(serviceProvider.getClass(), withAnnotation(Inject.class))) {
-      final Object dependency = lookup(field.getType());
-      if (dependency == null) {
-        throw new ServiceResolutionError(format("Cannot find a service to inject into field '%s' of service provider '%s'",
-                                                field.getName(), serviceProvider.getClass().getName()));
+      Class<?> dependencyType = field.getType();
+
+      boolean nullToOptional = false;
```
marianogonzalez: rename to `asOptional`

modules/service/src/main/java/org/mule/runtime/module/service/internal/manager/ServiceRegistry.java
```diff
@@ -42,23 +47,46 @@
...
+      if (dependencyType.equals(Optional.class)) {
+        Type type = ((ParameterizedType) (field.getGenericType())).getActualTypeArguments()[0];
+        if (type instanceof ParameterizedType) {
```
marianogonzalez: doesn't `GenericUtils` already have methods for doing this?

gsfernandes: No, I didn't find a helper method to avoid doing this here.

PR #7555, opened 2019-01-02
MDM-80: Metadatatype serialization should not be so heavy

juandesi:
Format for type references is redundant


juandesi:
closing this, a new version of the runtime can talk with an old tooling version, and this will clash

PR #7556, opened 2019-01-02
MULE-16218: Interception API: Smart connectors inside subflow are not…

gsfernandes:
… skipped properly


gsfernandes:
Seems to be a different issue, locations are correctly generated when SC operations are inside an scope but the interception is not even registered. I've created https://www.mulesoft.org/jira/browse/MULE-16285 to track this down.

gsfernandes:
--validate

PR #7557, opened 2019-01-03
EE-6517: Add jmh.ignore to microbenchmarks in performance tests

fsgonz:


tests/performance/pom.xml
```diff
@@ -155,6 +155,7 @@
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-surefire-plugin</artifactId>
                         <configuration>
+                            <argLine>-Djmh.ignoreLock=true</argLine>
```
elrodro83: We want this config on this profile. This shoul only be set for the default profile.

PR #7569, opened 2019-01-08
MULE-16299: Race condition on refreshing Oauth Token

fsgonz:


modules/oauth/src/main/java/org/mule/module/oauth2/internal/authorizationcode/state/ConfigOAuthContext.java
```diff
@@ -41,7 +41,9 @@ public ResourceOwnerOAuthContext getContextForResourceOwner(final String resourc
         if (!oauthContextStore.containsKey(resourceOwnerId))
         {
             final Lock lock = lockFactory.createLock(configName + "-config-oauth-context");
+            final Lock resourceLock = createLockForResourceOwner(resourceOwnerId);
             lock.lock();
```
lucianoRM: Shouldn't it be better to use 2 try-finally blocks? 
If the first lock was acquired and the second fails for some reason, the finally block will never be executed and the first lock will remain acquired.

modules/oauth/src/main/java/org/mule/module/oauth2/internal/clientcredentials/ClientCredentialsGrantType.java
```diff
@@ -103,7 +103,8 @@ public void authenticate(MuleEvent muleEvent, HttpRequestBuilder builder) throws
             refreshAccessToken();
         }
 
-        final String accessToken = tokenManagerConfig.getConfigOAuthContext().getContextForResourceOwner(ResourceOwnerOAuthContext.DEFAULT_RESOURCE_OWNER_ID).getAccessToken();
+        final String accessToken = getAccessToken(tokenManagerConfig.getConfigOAuthContext(), ResourceOwnerOAuthContext.DEFAULT_RESOURCE_OWNER_ID);
```
lucianoRM: static import

modules/oauth/src/test/java/org/mule/module/oauth2/internal/clientcredentials/functional/ClientCredentialsTokenRequestConcurrentTestCase.java
```diff
@@ -0,0 +1,106 @@
...
+    private CountDownLatch latch = new CountDownLatch(NUM_REQUESTS);
+
+    private static final int NUM_REQUESTS = 1000;
+
+    private static final String STATUS = "http.status";
+
+    private int responseCount = 0;
+
+    @Rule
+    public SystemProperty tokenUrl = new SystemProperty("token.url", String.format("http://localhost:%d" + TOKEN_PATH, oauthServerPort.getNumber()));
```
lucianoRM: static import

PR #7572, opened 2019-01-09
MULE-16274: HTTP service client cannot do gzip response decompression

afelisatti:


modules/http-api/src/main/java/org/mule/runtime/http/api/client/HttpClientConfiguration.java
```diff
@@ -79,6 +85,10 @@ public String getName() {
     return name;
   }
 
+  public Optional<Boolean> isDecompress() {
```
elrodro83: Why an optional? Wgy not just `boolean`?

afelisatti:
--validate

afelisatti:
--validate

afelisatti:
--validate

afelisatti:
--validate

PR #7589, opened 2019-01-11
MULE-16319: Container classloader in embedded mode must have the

elrodro83:
embedded container classloader as parent

### DIFF
--- a/modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/FilteringArtifactClassLoader.java
+++ b/modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/FilteringArtifactClassLoader.java
@@ -17,14 +17,6 @@ import static java.util.Collections.emptyList;
 import static org.mule.runtime.api.util.Preconditions.checkArgument;
 import static org.mule.runtime.core.api.config.MuleProperties.MULE_LOG_VERBOSE_CLASSLOADING;
 
-import java.io.IOException;
-import java.net.URL;
-import java.util.Enumeration;
-import java.util.List;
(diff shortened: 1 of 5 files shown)


modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/FilteringArtifactClassLoader.java
```diff
@@ -72,6 +75,18 @@ public FilteringArtifactClassLoader(ArtifactClassLoader artifactClassLoader, Cla
...
+  /**
+   * Creates a new filtering classLoader
+   *
+   * @param artifactClassLoader artifact classLoader to filter. Non null
+   * @param filter filters access to classes and resources from the artifact classLoader. Non null
+   * @param exportedServices service providers that will be available from the filtered class loader. Non null.
+   */
+  public FilteringArtifactClassLoader(ArtifactClassLoader artifactClassLoader, ClassLoaderFilter filter,
+                                      List<ExportedService> exportedServices) {
+    this(getSystemClassLoader(), artifactClassLoader, filter, exportedServices);
```
gsfernandes: Should just be null here? Parent can be null and it assigns the correct class leader. I'm wondering if this will work on standalone

elrodro83: No, the java ClassLoader doesn't seem to handle a null parent from what i see in the code. This what i am doing here would be the same as it is doing:

```
    protected ClassLoader() {
        this(checkCreateClassLoader(), getSystemClassLoader());
    }
```

PR #7604, opened 2019-01-16
MULE-16281: Testcase for emitter with REE scenario

mbuchwald:


mbuchwald:
related to https://github.com/reactor/reactor-core/issues/1488

PR #7619, opened 2019-01-21
MULE-15549: Improve PlatformClassloader injection for JDK11

lucianoRM:


modules/artifact/src/main/java/org/mule/runtime/module/artifact/internal/util/ClassLoaderUtil.java
```diff
@@ -32,6 +32,8 @@
    * @return the platform classloader or empty.
    */
   public static Optional<ClassLoader> getPlatformClassLoader() {
```
aiannucci: Remove this method

PR #7627, opened 2019-01-24
Mule 15872: Using splitter aggregator implicitly inside a request-reply scope wrapped with for each does not return the results back intermittently.

emariotti3:


core/src/main/java/org/mule/routing/outbound/AbstractMessageSequenceSplitter.java
```diff
@@ -50,6 +50,7 @@
     protected MessageInfoMapping messageInfoMapping;
     protected int batchSize;
     protected String counterVariableName;
+    private static int proccessedSequences = 0;
```
mbuchwald: Not sure why this should be `static`, but probably need to be `AtomicInteger`

pom.xml
```diff
@@ -144,7 +144,7 @@
         <antlrVersion>3.5.2</antlrVersion>
         <asmVersion>6.2</asmVersion>
         <aspectjVersion>1.9.1</aspectjVersion>
-        <asyncHttpClientVersion>1.14-MULE-007</asyncHttpClientVersion>
+        <asyncHttpClientVersion>1.14-MULE-008-SNAPSHOT</asyncHttpClientVersion>
```
mbuchwald: Why is this change here?

tests/integration/src/test/java/org/mule/test/usecases/routing/response/VMRequestReplyInForEachTestCase.java
```diff
@@ -6,13 +6,70 @@
...
+
+import java.util.List;
+import java.util.Map;
+import java.util.ArrayList;
+
+import org.junit.Test;
+
 public class VMRequestReplyInForEachTestCase extends RequestReplyInForEachTestCase
 {
+    protected static final int TIMEOUT = 5000;
```
mbuchwald: If the other property is now `protected`, this shouldn't be necessary (or the other one could remain `private`)

core/src/main/java/org/mule/routing/outbound/AbstractMessageSequenceSplitter.java
```diff
@@ -146,7 +150,7 @@ public Object transform(Object input)
                 if ((!correlationSet && (enableCorrelation == CorrelationMode.IF_NOT_SET))
                     || (enableCorrelation == CorrelationMode.ALWAYS))
                 {
-                    message.setCorrelationId(correlationId);
+                    message.setCorrelationId(correlationId + processedSequences.get());
```
elrodro83: put an `_` in the middle to separate the original id from the suffix

core/src/main/java/org/mule/routing/outbound/AbstractMessageSequenceSplitter.java
```diff
@@ -50,6 +51,7 @@
     protected MessageInfoMapping messageInfoMapping;
     protected int batchSize;
     protected String counterVariableName;
+    private static AtomicInteger processedSequences = new AtomicInteger(0);
```
elrodro83: does this need to be at the class level and atomic? Can't it be scoped to the `process` method.

tests/integration/src/test/java/org/mule/test/usecases/routing/response/VMRequestReplyInForEachTestCase.java
```diff
@@ -6,13 +6,68 @@
...
         return "org/mule/test/usecases/routing/response/vm-request-reply-in-for-each.xml";
     }
 
+    @Override
+    public void setUp() throws Exception
+    {
+        super.setUp();
+        collectionValues.add(new ArrayList<String>());
+        collectionValues.add(new ArrayList<String>());
+        collectionValues.get(0).add("value1");
```
elrodro83: Use an immutable list builder from guava.

core/src/main/java/org/mule/routing/outbound/AbstractMessageSequenceSplitter.java
```diff
@@ -51,7 +50,7 @@
     protected MessageInfoMapping messageInfoMapping;
     protected int batchSize;
     protected String counterVariableName;
-    private static AtomicInteger processedSequences = new AtomicInteger(0);
+    protected boolean sequential = false;
```
elrodro83: make final

core/src/main/java/org/mule/routing/outbound/AbstractMessageSequenceSplitter.java
```diff
@@ -150,7 +146,8 @@ public Object transform(Object input)
                 if ((!correlationSet && (enableCorrelation == CorrelationMode.IF_NOT_SET))
                     || (enableCorrelation == CorrelationMode.ALWAYS))
                 {
-                    message.setCorrelationId(correlationId + processedSequences.get());
+                    String suffix = sequential? ("-" + correlationSequence) : "";
```
elrodro83: format, and inline

core/src/main/java/org/mule/routing/outbound/AbstractMessageSequenceSplitter.java
```diff
@@ -87,6 +86,8 @@ protected boolean isSplitRequired(MuleEvent event)
         return true;
     }
 
+    protected boolean isSequential() { return false; }
```
elrodro83: format

PR #7633, opened 2019-01-28
MULE-14912: fix mule-integration-tests

Bardammu:
If merged, this PR fix the tests cases ConfigurationBasedElementModelFactoryTestCases
of mule-integration-tests. A proper solution should be in place after implemented
MULE-16385


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/JavaOAuthDeclarationEnricher.java
```diff
@@ -17,6 +17,8 @@
 import static org.mule.runtime.api.meta.model.parameter.ParameterRole.BEHAVIOUR;
 import static org.mule.runtime.extension.api.connectivity.oauth.ExtensionOAuthConstants.RESOURCE_OWNER_ID_PARAMETER_NAME;
 import static org.mule.runtime.extension.api.connectivity.oauth.ExtensionOAuthConstants.UNAUTHORIZE_OPERATION_NAME;
+import static org.mule.runtime.extension.api.loader.DeclarationEnricherPhase.POST_STRUCTURE;
```
ndinu: Remove unnecessary import

PR #7646, opened 2019-02-04
MULE-14601: Json Schema Validator treats duplicate keys in an inconsistent way depending on the input type.

eze210:
Adding a test to reproduce the issue reported at SE-10767.
The issue was fixed in a commit associated to MULE-14601.

### DIFF
--- a/modules/json/src/test/java/org/mule/module/json/JsonDataTestCase.java
+++ b/modules/json/src/test/java/org/mule/module/json/JsonDataTestCase.java
@@ -6,6 +6,7 @@
  */
 package org.mule.module.json;
 
+import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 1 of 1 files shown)


modules/json/src/test/java/org/mule/module/json/JsonDataTestCase.java
```diff
@@ -103,6 +104,16 @@ public void testReadingArray() throws Exception
...
     }
 
+    @Test
+    public void testReadingANumber() throws Exception
+    {
+        JsonParser parser = new DefaultJsonParser(muleContext);
+        String jsonAsString = "{\"value\":210.0}";
+
+        JsonNode json = parser.asJsonNode(jsonAsString);
+        assertEquals(jsonAsString, json.toString());
```
elrodro83: use hamcrest

PR #7647, opened 2019-02-04
MULE-16422: MuleArtifactClassLoader does not find resources with spaces when using resource:: syntax.

martincousido:
- Changing regex in order to capture the resource path with spaces.
- Adding test with an artifact that contains folders and files with spaces.


modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/MuleArtifactClassLoader.java
```diff
@@ -50,9 +50,14 @@
...
-  static final String WILDCARD = "*";
-
-  private static final String NO_WILDCARD_NO_SPACES = "([^\\" + WILDCARD + "|\\s]+)";
+  static final String WILDCARD = "\\*";
+
+  private static final String SPACES = "\\s";
+  private static final String RESOURCE_CAPTURE_START = "([^";
+  private static final String PIPE = "|";
+  private static final String RESOURCE_CAPTURE_END = "]+)";
+  private static final String RESOURCE = RESOURCE_CAPTURE_START + WILDCARD + PIPE + RESOURCE_CAPTURE_END;
```
afelisatti: Why is the pipe needed in this case? Sounds like you just need a NO_WILDCARDS regex.

modules/artifact/src/test/java/org/mule/runtime/module/artifact/api/classloader/RegionClassLoaderTestCase.java
```diff
@@ -72,9 +72,17 @@
...
   private static final String SPECIFIC_ARTIFACT_ID = "test-artifact";
+  private static final String SPECIFIC_ARTIFACT_ID_WITH_SPACES = "test-artifact-with-spaces";
   private static final String ARTIFACT_VERSION = "1.0.0";
   private static final String SPECIFIC_RESOURCE_FORMAT = "resource::" + GROUP_ID + ":" + SPECIFIC_ARTIFACT_ID + ":%s:%s:%s:%s";
+  private static final String SPECIFIC_RESOURCE_FORMAT_WITH_SPACES =
+      "resource::" + GROUP_ID + ":" + SPECIFIC_ARTIFACT_ID_WITH_SPACES + ":%s:%s:%s:%s";
   private static final String API_RESOURCE_NAME = "test-api.raml";
+  private static final String API_RESOURE_NAME_WITH_SPACES = "raml with spaces.raml";
+  private static final String API_FOLDER_NAME_WITH_SPACES = "folder with spaces";
+  private static final String API_RESOURCE_NAME_WITH_SPACES_ENCODED = API_RESOURE_NAME_WITH_SPACES.replace(" ", "%20");
```
afelisatti: Where are these used?

afelisatti:
--validate

modules/artifact/src/test/java/org/mule/runtime/module/artifact/api/classloader/RegionClassLoaderTestCase.java
martincousido: I added two tests, one that looks for a resource with spaces in the root of the artifact and another one that looks inside a folder which also has spaces just in case.
The ones with spaces are used to retrieve the resource using the resource:: syntax in which the spaces should be present. The ones with the %20 are used to build the URL which is expected to be returned by the classloader.
See: https://github.com/mulesoft/mule/pull/7647/files#diff-c002f1fe63a6bc2918bba26dc41b9f16R512
And: 
https://github.com/mulesoft/mule/pull/7647/files#diff-c002f1fe63a6bc2918bba26dc41b9f16R505

PR #7654, opened 2019-02-05
SE-10276: Close file streams left open

JavierGuar:
File stream left open caused an exception on Windows when trying to
delete the file.
This fix was verified and there was no exception.


JavierGuar:
A test won't be needed since `org.mule.util.store.QueuePersistenceObjectStoreTestCase` fails on windows in the `clear` and `queueFilesAreRemovedWhenSerializationFails` tests, and the tests all pass with the fix.

afelisatti:
This has been addressed in #7716.

PR #7675, opened 2019-02-13
MULE-16370 Provide a way to use childContexts without relying on a flatMap

elrodro83:
* Includes improvements in policies that use the new fluxed child contexts

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/policy/DefaultPolicyInstance.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/policy/DefaultPolicyInstance.java
@@ -239,7 +239,6 @@ public class DefaultPolicyInstance extends AbstractComponent
       cpuLiteScheduler = schedulerService.cpuLightScheduler(schedulerBaseConfig.withName(schedulersNamePrefix + "-cpuLite"));
     }
 
-    // @Override
     public void stopSchedulers() {
       if (cpuLiteScheduler != null) {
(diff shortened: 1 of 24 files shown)


core/src/main/java/org/mule/runtime/core/internal/policy/DefaultPolicyInstance.java
```diff
@@ -241,8 +239,8 @@ public void start() throws MuleException {
       cpuLiteScheduler = schedulerService.cpuLightScheduler(schedulerBaseConfig.withName(schedulersNamePrefix + "-cpuLite"));
     }
 
-    @Override
-    public void stop() throws MuleException {
+    // @Override
```
mbuchwald: Remove commented annotation

core/src/main/java/org/mule/runtime/core/internal/processor/TryScope.java
```diff
@@ -101,6 +101,7 @@ public CoreEvent process(final CoreEvent event) throws MuleException {
     } else {
       return from(publisher)
           .flatMap(event -> processWithChildContext(event, nestedChain, ofNullable(getLocation()), messagingExceptionHandler));
+      // return applyWithChildContext(from(publisher), nestedChain, ofNullable(getLocation()), messagingExceptionHandler);
```
mbuchwald: remove commented code

PR #7680, opened 2019-02-14
MULE-16483: Mule Domain is not disposing its RegionClassLoader

gsfernandes:


gsfernandes:
--validate

gsfernandes:
--validate

gsfernandes:
--validate

gsfernandes:
@elrodro83 changes as requested, as only deployable artifacts would have be the owner of region classloaders I have moved this code to a base class from which Domain and Application implementation now extend to reuse this logic and code.
Hawk: one test that has failed but seems to be flaky ``` org.mule.runtime.core.privileged.util.TemplateParserTestCase.muleParserWithScapedQuote```
Distribution tests: OK, same that are currently failing on build.

PR #7691, opened 2019-02-18
MULE-16498: Add option to not recycle receiver in case of Connection Exceptopm

fsgonz:


transports/jms/src/main/java/org/mule/transport/jms/JmsConnector.java
```diff
@@ -1635,4 +1641,9 @@ public void setStillConnectingReceivers(boolean aBoolean)
     {
         this.stillConnectingReceivers.set(aBoolean);
     }
+
+    public boolean mustRecycleReceivers()
+    {
+        return false;
```
elrodro83: isn't this changing the default behavior?

fsgonz: Yes. It should be true.

PR #7693, opened 2019-02-19
MULE-16500: Janitor must be dereferenced so that taken byte arrays are freed

fsgonz:
when ghostbuster applied.


fsgonz:
--validate

PR #7718, opened 2019-02-26
MULE-16417: Fix issues affecting mule-aggregators-module tests on Java 11

garodriguezlp:
Improvements and missing parts to javaee module for fixing issues with tests on Java 11


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/stereotypes/StereotypeResolver.java
```diff
@@ -59,7 +60,8 @@ public static StereotypeModel createCustomStereotype(Class<? extends StereotypeD
   protected static StereotypeModel getStereotype(StereotypeDefinition stereotypeDefinition,
                                                  String namespace,
                                                  Map<StereotypeDefinition, StereotypeModel> stereotypesCache) {
-    return computeIfAbsent(stereotypesCache, stereotypeDefinition, definition -> {
+
+    return FunctionalUtils.computeIfAbsent(stereotypesCache, stereotypeDefinition, definition -> {
```
aiannucci: Why not static import?

modules/javaee/pom.xml
```diff
@@ -73,6 +84,16 @@
             <artifactId>jaxws-api</artifactId>
             <version>${javax.jaxws-api.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.sun.mail</groupId>
+            <artifactId>javax.mail</artifactId>
+            <version>${javax.mail.version}</version>
+        </dependency>
```
aiannucci: Java mail was not part of JDK8, so why should we include it here?

garodriguezlp: As far as I know; you might be right. I agree with you on that there was no change regarding JavaMail and the modules shipped with the JRE on Java 8 vs Java 11. 

Nevertheless, this particular addition was part of @pablolagreca 's Strategy for fixing all of the classloading issues that came out with Java 11 on the extension's tests, which all were somehow of the same kind. 

I know that the original scope of this module was just for all of those things that are no longer going to be shipped with the JRE for Java 11, but I guess @pablolagreca  must have a strong reason for having suggested this, that why I'd like to hear his opinion on this first.

aiannucci: @pablolagreca any thoughts on this?

garodriguezlp:
--validate

garodriguezlp:
--validate

PR #7724, opened 2019-03-11
MULE-16547: IMAP not receiving emails working on 3.9.2

eze210:
Using the retry policy in the doStart method too

### DIFF
--- a/transports/email/src/main/java/org/mule/transport/email/RetrieveMessageReceiver.java
+++ b/transports/email/src/main/java/org/mule/transport/email/RetrieveMessageReceiver.java
@@ -35,6 +35,7 @@ import java.io.IOException;
 import java.nio.channels.NotYetConnectedException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import javax.mail.Address;
 import javax.mail.Flags;
@@ -68,7 +69,7 @@ public class RetrieveMessageReceiver extends AbstractPollingMessageReceiver impl
     private final Object connectionLock = new Object();
 
 
-    private boolean connectionEstablished = false;
+    private AtomicBoolean connectionEstablished = new AtomicBoolean(false);
 
     public RetrieveMessageReceiver(Connector connector,
                                    FlowConstruct flowConstruct,
@@ -123,7 +124,7 @@ public class RetrieveMessageReceiver extends AbstractPollingMessageReceiver impl
                         RetrieveMessageReceiver.this.backupFolder += File.separator;
                     }
 
-                    RetrieveMessageReceiver.this.connectionEstablished = true;
+                    RetrieveMessageReceiver.this.connectionEstablished.set(true);
                 }
             }
 
@@ -175,7 +176,7 @@ public class RetrieveMessageReceiver extends AbstractPollingMessageReceiver impl
                 {
                     synchronized (connectionLock)
                     {
-                        if (RetrieveMessageReceiver.this.connectionEstablished)
+                        if (RetrieveMessageReceiver.this.connectionEstablished.get())
                         {
                             RetrieveMessageReceiver.super.doStart();
                             synchronized (folderLock)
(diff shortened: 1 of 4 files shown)


plbalbi:
The change seems alright to me. I'll add some comment on top of the second retry policy usage explaining the slow-connect situation. 
I'll maybe add a test for this race condition.

transports/email/src/main/java/org/mule/transport/email/RetrieveMessageReceiver.java
```diff
@@ -101,32 +95,34 @@ protected void doConnect() throws Exception
...
             public void doWork(RetryContext context) throws Exception
             {
-                logger.info("Attemping to establish connection with message retrieval server");
-                SessionDetails session = RetrieveMessageReceiver.this.castConnector().getSessionDetails(endpoint);
-
-                Store store = session.newStore();
-                store.connect();
-                RetrieveMessageReceiver.this.folder = store.getFolder(castConnector().getMailboxFolder());
-                if (castConnector().getMoveToFolder() != null)
+                synchronized (connectionLock)
```
afelisatti: Add comments explaining the need for the lock

transports/email/src/main/java/org/mule/transport/email/RetrieveMessageReceiver.java
```diff
@@ -166,11 +162,50 @@ protected void doStop() throws MuleException
     @Override
     protected void doStart() throws MuleException
     {
-        super.doStart();
-        synchronized (folderLock)
+        try
         {
-            folder.addMessageCountListener(this);
+            getConnector().getRetryPolicyTemplate().execute(new RetryCallback()
```
afelisatti: Add a comment explaining why this deferring is needed.

transports/email/src/main/java/org/mule/transport/email/RetrieveMessageReceiver.java
```diff
@@ -166,11 +162,50 @@ protected void doStop() throws MuleException
...
+                }
+            }, getWorkManager());
+        }
+        catch (MuleException me)
+        {
+            throw me;
         }
+        catch (Exception e)
+        {
+            // Ignore other exceptions
```
afelisatti: Why?

transports/email/src/main/java/org/mule/transport/email/RetrieveMessageReceiver.java
eze210: Comment added

transports/email/src/main/java/org/mule/transport/email/RetrieveMessageReceiver.java
eze210: Comment added

transports/email/src/main/java/org/mule/transport/email/RetrieveMessageReceiver.java
eze210: Now all caught exceptions are wrapped in a LifecycleException

transports/email/src/main/java/org/mule/transport/email/RetrieveMessageReceiver.java
```diff
@@ -62,18 +64,12 @@
     private String backupFolder = null;
     // A lock to protect concurrent access to the folder.
     private final Object folderLock = new Object();
+    // Lock to prevent connect and start from being executed simultaneously.
+    private final Object connectionLock = new Object();
 
 
     private boolean connectionEstablished = false;
```
plbalbi: Change connectionEstablished datatype to AtomicBoolean since to scheduled separate threads will be reading / writing its value.

eze210: I made the change for possible future concurrent usages of the flag, but now it is not needed since it is only used inside synchronized sections.

plbalbi: Didn't noticed the read/writes where inside a synchronized section. Agree with leaving it as AtomicBoolean regardless.

PR #7731, opened 2019-03-15
MULE-16554: Add new methods to check for key and mutability

afelisatti:


afelisatti:
--validate

afelisatti:
—validate

afelisatti:
--validate

modules/http-api/src/main/java/org/mule/runtime/http/api/domain/CaseInsensitiveMultiMap.java
```diff
@@ -69,6 +69,11 @@ public ImmutableCaseInsensitiveMultiMap(CaseInsensitiveMultiMap caseInsensitiveM
       super(caseInsensitiveMultiMap, caseInsensitiveMultiMap.optimized);
       this.paramsMap = unmodifiableMap(paramsMap);
     }
+
+    @Override
+    public boolean isMutable() {
```
marianogonzalez: i don't get the purpose of this. How does this help? Sounds like a hack made API

afelisatti: Yes, you are correct. I added this so that we can check whether instances have been recreated or are being reused (which can often be immutable and need to be copied to operate on them).

PR #7735, opened 2019-03-19
MULE-15656: Add new attribute to @ExternalLib annotation

lucianoRM:


modules/extensions-spring-support/src/test/resources/models/heisenberg.json
```diff
@@ -28430,7 +28435,8 @@
           "regexMatcher": "heisenberg.so",
           "requiredClassName": "org.heisenberg.HeisenbergJNI",
           "type": "NATIVE",
-          "optional": false
+          "optional": false,
+          "allowDeclarationAsAdditionalDependency": true
```
marianogonzalez: what is this? Weren't you going to write a spec for it?

lucianoRM: The motivation to write an spec was not actually this case, but how to be prepared for different changes to the extension model we could encounter in the future.

gsfernandes: Let's put @marianogonzalez in context here. There is an spec already wrote a long time ago about this capability for Runtime which defined this change on SDK. We could review the section related to this part only. 
On the other hand, as @lucianoRM said, the main discussion about how to handle this kind of "capabilities" provided since a particular version of Runtime, not SDK, is what we were not clear about. This particular flag defines that libraries can shared only within a particular plugin and not for the whole set of plugins defined in the application. It has nothing to do with SDK annotations as this is not something that plugin developers should care about but it should be provided to Tooling as it is now supported since 4.2 and will have impact on how the sharedRuntimeLibraries are defined in the pom.xml for the application.
Do you think that this should only be considered by Tooling API extension model only?

marianogonzalez: my point is that this if this is not a concern of the extension, then why should it be the extension's responsibility to know about this? Even if this doesn't affect the SDK annotations, this is still affecting the extensions API. A key design principle is that plugins are injected with their classpath, they don't handle it. If the extension developer doesn't know or care about this concept and has no control over it, then why is this here? It seems to me like this is a runtime concern, it is the runtime the one that should be providing this information

lucianoRM: Makes sense. A motivation for adding this in the extension's api was to open the door in case the necessity for configuring this from the extension showed up.  But we couldn't actually think of an scenario where this would be needed right now. In any case, if that scenario existed, we would need to wait until 4.3 to make any changes, so it should not matter if the information is not added to the model right now. 
I am ok with enriching the tooling's extension model with this information according to the runtime without having to have it in the api's model. If everyone else agrees, i will make the changes

lucianoRM:
Will handle this on tooling side

PR #7736, opened 2019-03-19
MULE-16571: Externalize log4j2.xml file in Mule 4 domain descriptor ca…

emariotti3:
…nnot be cast to application descriptor.


emariotti3:
--validate

PR #7738, opened 2019-03-19
MULE-16537: Error handling not working properly with foreach - Regression caused by MULE-16374

mbuchwald:


core/src/main/java/org/mule/runtime/core/privileged/processor/MessageProcessors.java
```diff
@@ -377,32 +373,26 @@ private static CoreEvent internalProcessWithChildContextBlocking(CoreEvent event
         })
         .transform(processor)
         .doOnNext(completeSuccessIfNeeded())
-        .map(event -> right(MessagingException.class, event))
-        // This Either here is used to propagate errors. If the error is sent directly through the merged with Flux, it will be
```
afelisatti: How did you workaround this?

PR #7765, opened 2019-03-31
MULE-16633 - Cool Mulesoft logo in splash screen looking bad in windows

pablolagreca:


modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainerStartupSplashScreen.java
```diff
@@ -42,7 +44,10 @@
   public void doBody() {
 
     try (InputStream resourceAsStream = this.getClass().getClassLoader().getResourceAsStream("banner.txt")) {
-      doBody(IOUtils.toString(resourceAsStream));
+      LineIterator lineIterator = lineIterator(resourceAsStream, Charset.defaultCharset());
```
lucianoRM: static import

PR #7766, opened 2019-04-01
MULE-16634: Source completion callback not called on failure

afelisatti:


modules/http-api/src/main/java/org/mule/runtime/http/api/server/async/ResponseStatusCallback.java
```diff
@@ -25,4 +25,12 @@
...
   void responseSendSuccessfully();
 
+  /**
+   * Method to process an error without sending a response, since the error that occurred cannot be recovered from. Implementors
+   * should override this particularly when dealing with SDK asynchronous sources that require a completion callback notification.
+   *
+   * @param throwable exception thrown while sending the response
+   * @since 4.1.6
+   */
+  default void responseSendError(Throwable throwable) {}
```
marianogonzalez: responseSentWIthError? onErrorSendingResponse()?

PR #7767, opened 2019-04-01
MULE-16636: CursorProvider is retained too long when cursors are opened but not consumed

marianogonzalez:


core/src/main/java/org/mule/runtime/core/internal/streaming/CursorProviderJanitor.java
```diff
@@ -89,6 +89,11 @@ public final void releaseResources() {
   public void releaseCursor(Cursor cursor) {
     try {
       cursor.release();
+      if (cursors.remove(cursor)) {
+        if (provider.isClosed() && cursors.isEmpty()) {
```
elrodro83: why not put this with an `&&` in the previous `if`?

PR #7780, opened 2019-04-04
MULE-16644: PollingSources ObjectStore names need to be part of the API

alegmarra:


modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/source/PollingSourceWrapperTestCase.java
```diff
@@ -0,0 +1,111 @@
...
+  private SchedulerService schedulerServiceMock;
+
+  @Mock
+  private ComponentLocation componentLocationMock;
+
+  @Mock
+  private SourceCallback callbackMock;
+
+  @InjectMocks
+  private PollingSourceWrapper<Object, Object> pollingSourceWrapper = new PollingSourceWrapper<Object, Object>(
```
marianogonzalez: static imports and format

PR #7783, opened 2019-04-05
MULE-16581: Runtime does not shutdown after OptionalLong[15000] milliseconds

plbalbi:


modules/service/src/main/java/org/mule/runtime/module/service/internal/manager/MuleServiceManager.java
```diff
@@ -67,15 +70,35 @@ private void startServices() throws MuleException {
...
 
+  /**
+   * Stop all discovered services. In the case of 'HTTP' and 'Scheduler' service, if present, they should be stopped last and in
+   * that order.
+   *
+   * @throws MuleException
+   */
   @Override
   public void stop() throws MuleException {
+    LinkedList<Service> servicesWithSpecialStoppingOrder = new LinkedList<>();
```
marianogonzalez: use `new ArrayList<>(2)` or simply two variables.

modules/service/src/main/java/org/mule/runtime/module/service/internal/manager/MuleServiceManager.java
```diff
@@ -67,15 +70,35 @@ private void startServices() throws MuleException {
...
-        LOGGER.warn(format("Failed to stop service '%s': %s", service.getName(), e.getMessage()), e);
+      if (service.getName().equals(SCHEDULER_SERVICE_ARTIFACT_ID)) {
+        servicesWithSpecialStoppingOrder.addLast(service);
+      } else if (service.getName().contains(HTTP_SERVICE_ARTIFACT_PREFIX)) {
+        servicesWithSpecialStoppingOrder.addFirst(service);
+      } else {
+        doStopService(service);
       }
     }
+    for (Service service : servicesWithSpecialStoppingOrder) {
```
marianogonzalez: you can leave it as it is if you want.. Keep in mind that you could have used `servicesWithSpecialStoppingOrder.forEach(this::doStopService)` instead

modules/service/src/main/java/org/mule/runtime/module/service/internal/manager/MuleServiceManager.java
plbalbi: Changed to Optionals. Solution seems cleaner.

plbalbi: Back to null.

PR #7785, opened 2019-04-05
MULE-16669: Mule throws error Illegal group reference Exception when …

JuLuzzi:
…trying to pass $ into URI Parameter in HTTP request

### DIFF
--- a/modules/http/src/main/java/org/mule/module/http/internal/request/HttpRequesterRequestBuilder.java
+++ b/modules/http/src/main/java/org/mule/module/http/internal/request/HttpRequesterRequestBuilder.java
@@ -8,14 +8,12 @@ package org.mule.module.http.internal.request;
 
 import static java.lang.String.format;
 import static java.util.regex.Matcher.quoteReplacement;
-import static java.util.regex.Pattern.compile;
 import org.mule.api.MuleEvent;
 import org.mule.module.http.internal.HttpMessageBuilder;
 import org.mule.module.http.internal.HttpParamType;
 import org.mule.module.http.internal.ParameterMap;
 
 import java.util.List;
-import java.util.regex.Pattern;
 
 public class HttpRequesterRequestBuilder extends HttpMessageBuilder
 {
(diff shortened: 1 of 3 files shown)


modules/http/src/main/java/org/mule/module/http/internal/request/HttpRequesterRequestBuilder.java
```diff
@@ -6,17 +6,21 @@
...
 import org.mule.module.http.internal.ParameterMap;
 
 import java.util.List;
+import java.util.regex.Pattern;
 
 public class HttpRequesterRequestBuilder extends HttpMessageBuilder
 {
 
-    public ParameterMap getQueryParams(MuleEvent event)
+    ParameterMap getQueryParams(MuleEvent event)
```
afelisatti: This breaks compatibility.

modules/http/src/main/java/org/mule/module/http/internal/request/HttpRequesterRequestBuilder.java
```diff
@@ -26,7 +30,7 @@ public ParameterMap getHeaders(MuleEvent event)
         return resolveParams(event, HttpParamType.HEADER);
     }
 
-    public String replaceUriParams(String path, MuleEvent event)
+    String replaceUriParams(String path, MuleEvent event)
```
afelisatti: same

modules/http/src/main/java/org/mule/module/http/internal/request/HttpRequesterRequestBuilder.java
```diff
@@ -39,12 +43,16 @@ public String replaceUriParams(String path, MuleEvent event)
 
             if (uriParamValue == null)
             {
-                throw new NullPointerException(String.format("Expression {%s} evaluated to null.", uriParamName));
+                throw new NullPointerException(format("Expression {%s} evaluated to null.", uriParamName));
             }
 
-            path = path.replaceAll(String.format("\\{%s\\}", uriParamName), uriParamValue);
+            String value = quoteReplacement(uriParamValue);
```
afelisatti: Why was this necessary?

JuLuzzi: It's necessary to solve the dollar sign problem, replaceAll method documentation: "Note that backslashes and dollar signs in the replacement string may cause the results to be different than if it were being treated as a literal replacement string. Dollar signs may be treated as references to captured subsequences as described above, and backslashes are used to escape literal characters in the replacement string."

afelisatti: inline it

PR #7788, opened 2019-04-05
MULE-16672: org.mule.runtime.api.util.ResourceLocator is caching reso…

gsfernandes:
…urces and it nos working with SNAPSHOT dependencies


gsfernandes:
--validate

PR #7801, opened 2019-04-09
MULE-16537: Error handling not working properly with foreach

elrodro83:
* Revert and workaround to fix performance regression in policies


elrodro83:
--validate

PR #7805, opened 2019-04-10
Proxy leak

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/ClassLoaderConnectionProviderWrapper.java
```diff
@@ -0,0 +1,142 @@
...
+import org.mule.runtime.core.internal.retry.ReconnectionConfig;
+
+import java.util.Optional;
+import java.util.concurrent.Callable;
+
+/**
+ * A {@link ConnectionProviderWrapper} which makes sure that all delegate methods are executed with a given {@link #classLoader}
+ *
+ * @param <C> the generic type of the connections the delegate produces
+ * @since 4.2.0
```
afelisatti: You shall not pass -> 4.2.1?

marianogonzalez: si pasaba pasaba

PR #7811, opened 2019-04-15
Something like this... Still missing unit test for logic on lightweig…

gsfernandes:
…ht to recreate the json when the jar modified timestamp is changed in order to support SNAPSHOTS


lucianoRM:
Recommended changes:
- We should differentiate cases where the app is actually packaged as heavyweight and when it was converted to that type after the first time the ClassLoaderModel is created. Maybe use an extra file with the lastModified time of the jar file to check.
- if a transitive dependency was removed by hand and the app is converted heavyweight, we should catch the error being thrown and treat it as lightweight again so that all missin dependencies are downloaded again.

PR #7814, opened 2019-04-16
MULE-16724: Lookup of converters after initialization generates leaks in MuleRegistryHelper

fsgonz:


core/src/main/java/org/mule/registry/MuleRegistryHelper.java
```diff
@@ -500,6 +504,11 @@ public void notifyTransformerResolvers(Transformer t, TransformerResolver.Regist
         }
     }
 
+    private String getConverterKey(Transformer t)
+    {
+        return t.getName()!=null?t.getName():generateTransformerName(t.getClass(), t.getReturnDataType());
```
afelisatti: Format

core/src/test/java/org/mule/agent/registry/MuleRegistryHelperTransformerLookupTestCase.java
```diff
@@ -75,4 +77,23 @@ public void cachesTransformers() throws Exception
         assertEquals(1, transformers.size());
         assertEquals(stringToOrange, transformers.get(0));
     }
+    
+    @Test
+    public void reregisterDoesNotLeakTransformer() throws Exception
+    {
+        Converter stringToBanana = new MockConverterBuilder().from(DataTypeFactory.STRING).to(BANANA_DATA_TYPE).build();
```
afelisatti: Static import

core/src/test/java/org/mule/agent/registry/MuleRegistryHelperTransformerLookupTestCase.java
```diff
@@ -75,4 +77,23 @@ public void cachesTransformers() throws Exception
...
+        Converter copyStringToBanana = new MockConverterBuilder().from(DataTypeFactory.STRING).to(BANANA_DATA_TYPE).build();
+        copyStringToBanana.setName("StringToBanana");
+        
+        muleRegistryHelper.registerTransformer(stringToBanana);
+        muleRegistryHelper.registerTransformer(copyStringToBanana);
+        muleRegistryHelper.lookupTransformers(DataTypeFactory.STRING, BANANA_DATA_TYPE);
+
+        List<Transformer> transformers = muleRegistryHelper.lookupTransformers(DataTypeFactory.STRING, BANANA_DATA_TYPE);
+        
+        assertEquals(1, transformers.size());
```
afelisatti: Hamcrest

core/src/main/java/org/mule/registry/MuleRegistryHelper.java
```diff
@@ -98,7 +102,7 @@
     /**
      * Transformers are registered on context start, then they are usually not unregistered
      */
-    private Collection<Transformer> transformers = new CopyOnWriteArrayList<Transformer>();
+    private Map<String, Transformer> transformers = new ConcurrentHashMap<String, Transformer>();
```
afelisatti: Why not a set? From my understanding you just want to avoid duplicates.

fsgonz: The Converter and Transformer hierarchy is not consistent. The abstract classes from which main converters and transformers inherit do not have the getName method (which is declared in the interface NameableObject). As a consequence of this, it not possible to create an equals method. I think refactoring that hierarchy violates retro-compatibility. The strings used for the key is similar to the one used for the registry (without the hyphen "_").

PR #7815, opened 2019-04-16
MULE-16725: Fix race condition in flow lifecycle handling

elrodro83:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/SpringXmlConfigurationBuilder.java
```diff
@@ -58,8 +58,8 @@
 
   private ApplicationContext parentContext;
   private MuleArtifactContext muleArtifactContext;
-  private ArtifactType artifactType;
-  private ComponentBuildingDefinitionProvider runtimeComponentBuildingDefinitionsProvider;
+  private final ArtifactType artifactType;
+  private final ComponentBuildingDefinitionProvider runtimeComponentBuildingDefinitionsProvider;
```
plbalbi: For this two fields that are changed to a **final** modifier, isn't ir required to set them in the static constructor?

core/src/main/java/org/mule/runtime/core/internal/registry/MuleRegistryHelper.java
```diff
@@ -52,31 +53,31 @@
   /**
    * A reference to Mule's internal registry
    */
-  private Registry registry;
+  private final Registry registry;
```
plbalbi: Also this one?

core/src/main/java/org/mule/runtime/core/internal/registry/MuleRegistryHelper.java
```diff
@@ -52,31 +53,31 @@
...
 
   /**
    * We cache transformer searches so that we only search once
    */
   protected Map<String, Transformer> exactTransformerCache = new ConcurrentHashMap<>(8);
   protected Map<String, List<Transformer>> transformerListCache = new ConcurrentHashMap<>(8);
 
-  private MuleContext muleContext;
+  private final MuleContext muleContext;
```
plbalbi: And this one

modules/spring-config/src/main/java/org/mule/runtime/config/internal/SpringXmlConfigurationBuilder.java
elrodro83: it is being set in the constructor. The `final` is added automatically by the IDE when it detects that  it is not being written anywhere else.

core/src/main/java/org/mule/runtime/core/internal/registry/MuleRegistryHelper.java
elrodro83: it is being set in the constructor. The `final` is added automatically by the IDE when it detects that  it is not being written anywhere else.

core/src/main/java/org/mule/runtime/core/internal/registry/MuleRegistryHelper.java
elrodro83: it is being set in the constructor. The `final` is added automatically by the IDE when it detects that  it is not being written anywhere else.

PR #7817, opened 2019-04-17
MULE-10525: User credentials are shown in toString() of DefaultMuleMe…

JuLuzzi:
…ssage

### DIFF
--- a/core/src/main/java/org/mule/util/StringMessageUtils.java
+++ b/core/src/main/java/org/mule/util/StringMessageUtils.java
@@ -255,7 +255,7 @@ public final class StringMessageUtils
                     {
                         value = "<<<MuleMessage>>>";
                     }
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mule/util/StringMessageUtils.java
```diff
@@ -255,7 +255,11 @@ public static String headersToString(MuleMessage m)
                     {
                         value = "<<<MuleMessage>>>";
                     }
-                    if (name.equals("password") || name.toString().contains("secret") || name.equals("pass") || name.toString().contains("authorization"))
+                    if (name.equals("password") || name.equals("pass"))
+                    {
+                        value = "****";
+                    }
+                    else if (name.toString().contains("authorization") || name.toString().contains("secret"))
```
afelisatti: You can do the toString in the else and move the if inside for a better approach

PR #7818, opened 2019-04-18
MULE-16299: Race condition on refreshing Oauth Token.

fsgonz:


modules/oauth/src/main/java/org/mule/module/oauth2/internal/authorizationcode/state/ConfigOAuthContext.java
```diff
@@ -30,45 +30,45 @@ public ConfigOAuthContext(final LockFactory lockFactory, ListableObjectStore<Res
     }
 
     /**
-     * Retrieves the oauth context for a particular user. If there's no state for that user a new state is retrieve so never returns null.
+     * Retrieves the oauth context for a particular user. If there's no state for that user a new state is retrieve so
```
lucianoRM: retrieve -> retrieved

modules/oauth/src/main/java/org/mule/module/oauth2/internal/authorizationcode/state/ConfigOAuthContext.java
```diff
@@ -30,45 +30,45 @@ public ConfigOAuthContext(final LockFactory lockFactory, ListableObjectStore<Res
...
-        if (!oauthContextStore.containsKey(resourceOwnerId))
+
+        final Lock lock = lockFactory.createLock(configName + "-config-oauth-context");
+        lock.lock();
+        try
         {
-            final Lock lock = lockFactory.createLock(configName + "-config-oauth-context");
-            final Lock resourceLock = createLockForResourceOwner(resourceOwnerId);
-            lock.lock();
+            Lock resourceLock = createLockForResourceOwner(resourceOwnerId);
```
lucianoRM: why lock on the resource owner? Isn't it always related to the same config? Or can a resource owner be in different configs?

fsgonz: This is to limit the lock to the user so that if another user needs the oauth context, this is not locked.

fsgonz: The context can be cleared or updated while the context is being retrieved (with two concurrent request for the same user). That's why it has to be locked.

PR #7820, opened 2019-04-22

NachoRaik:


transports/http/src/main/java/org/mule/transport/http/filters/HttpBasicAuthenticationFilter.java
```diff
@@ -135,6 +135,7 @@ public void authenticateInbound(MuleEvent event)
                 if (logger.isDebugEnabled())
                 {
                     logger.debug("Authentication request for user: " + username + " failed: " + e.toString());
+                    logger.debug("Cause: " + e.getCause().getMessage());
```
afelisatti: Doesn't this work by logging the exception directly? To get the entire stack logged.

transports/http/src/main/java/org/mule/transport/http/filters/HttpBasicAuthenticationFilter.java
```diff
@@ -134,7 +134,8 @@ public void authenticateInbound(MuleEvent event)
                 // Authentication failed
                 if (logger.isDebugEnabled())
                 {
-                    logger.debug("Authentication request for user: " + username + " failed: " + e.toString());
+                    logger.debug(String.format("Authentication request for user: %s failed: %s\nCause: %s",
```
afelisatti: static import

afelisatti: use `'%s'` to distinguish messages

PR #7827, opened 2019-04-24
MULE-16753: A transaction is not available error related to execution on a cpuLight thread.

emariotti3:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
```diff
@@ -28,8 +30,9 @@
  */
 public class SourceConnectionManager {
 
+  private static final ReentrantLock lock = new ReentrantLock();
   private final ConnectionManager connectionManager;
-  private final Map<Reference<Object>, ConnectionHandler<Object>> connections = new ConcurrentHashMap<>();
+  private final Map<Reference<Object>, Pair<Integer, ConnectionHandler<Object>>> connections = new ConcurrentHashMap<>();
```
marianogonzalez: Don't use an Integer. Use an `AtomicInteger` instead. In this way you don't need to be replacing map entries

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
```diff
@@ -49,11 +52,17 @@
    * @throws ConnectionException if a connection could not be obtained
    */
   <T> T getConnection(ConfigurationInstance config) throws ConnectionException {
-    ConnectionHandler<Object> connectionHandler = connectionManager.getConnection(config.getValue());
-    Object connection = connectionHandler.getConnection();
-    connections.put(new Reference<>(connection), connectionHandler);
-
-    return (T) connection;
+    lock.lock();
```
marianogonzalez: using AtomicInteger makes the need for this lock unnecessary. Also for future reference, notice that the only critical resource here is the map. So if you were to actually synchronize, you should only synchronoize that part. It doesnt make sense to syncrhonize the connect part

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
```diff
@@ -105,6 +144,15 @@ ConnectionValidationResult testConnectivity(Object connection) {
...
-    return ofNullable((ConnectionHandler<T>) connections.get(new Reference<>(connection)));
+    lock.lock();
+    try {
+      return ofNullable(connections.get(new Reference<>(connection))).map(c -> (ConnectionHandler<T>) (((Pair) c).getSecond()));
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  private int getConnectionRefCount(Reference<Object> connection) {
```
marianogonzalez: this method should go away in fvor of AtomicInteger

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
```diff
@@ -62,9 +71,21 @@
    * @param connection the connection to be released
    */
   void release(Object connection) {
-    ConnectionHandler<Object> connectionHandler = connections.remove(new Reference<>(connection));
-    if (connectionHandler != null) {
-      connectionHandler.release();
+    Reference<Object> connReference = new Reference<>(connection);
```
marianogonzalez: do we really need to synchronize here? I tend to think no, but yet again, here is where the race condition was.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
```diff
@@ -74,9 +95,21 @@ void release(Object connection) {
    * @param connection the connection to be invalidated
    */
   void invalidate(Object connection) {
-    ConnectionHandler<Object> connectionHandler = connections.remove(new Reference<>(connection));
-    if (connectionHandler != null) {
-      connectionHandler.invalidate();
+    Reference<Object> connReference = new Reference<>(connection);
+    lock.lock();
```
marianogonzalez: same comment as release

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
```diff
@@ -87,7 +120,13 @@ void invalidate(Object connection) {
    * @return a {@link ConnectionValidationResult}
    */
   ConnectionValidationResult testConnectivity(Object connection) {
-    ConnectionHandler<Object> connectionHandler = connections.get(new Reference<>(connection));
+    ConnectionHandler<Object> connectionHandler;
+    lock.lock();
```
marianogonzalez: should not lock everybody just because one thread is doing test connectivity

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
elrodro83: is there any access outside of the `lock` to justify keeping this `ConcurrentHashMap`?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
elrodro83: use `ConcurrentUtils.withLock()`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
```diff
@@ -62,9 +71,21 @@
    * @param connection the connection to be released
    */
   void release(Object connection) {
-    ConnectionHandler<Object> connectionHandler = connections.remove(new Reference<>(connection));
-    if (connectionHandler != null) {
-      connectionHandler.release();
+    Reference<Object> connReference = new Reference<>(connection);
+    lock.lock();
```
elrodro83: use `ConcurrentUtils.withLock()`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
elrodro83: use `ConcurrentUtils.withLock()`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
elrodro83: use `ConcurrentUtils.withLock()`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
```diff
@@ -105,6 +144,15 @@ ConnectionValidationResult testConnectivity(Object connection) {
    * through this component
    */
   <T> Optional<ConnectionHandler<T>> getConnectionHandler(T connection) {
-    return ofNullable((ConnectionHandler<T>) connections.get(new Reference<>(connection)));
+    lock.lock();
```
elrodro83: use `ConcurrentUtils.withLock()`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
fsgonz: If an AtomicInteger is used, these synchronized blocks can be avoided.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
```diff
@@ -28,8 +32,9 @@
  */
 public class SourceConnectionManager {
 
+  private static final ReentrantLock lock = new ReentrantLock();
```
elrodro83: why static?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
```diff
@@ -62,10 +70,17 @@
    * @param connection the connection to be released
    */
   void release(Object connection) {
-    ConnectionHandler<Object> connectionHandler = connections.remove(new Reference<>(connection));
-    if (connectionHandler != null) {
-      connectionHandler.release();
-    }
+    Reference<Object> connReference = new Reference<>(connection);
+    withLock(lock, () -> {
+      if (connections.get(new Reference<>(connection)).getFirst().decrementAndGet() > 0) {
```
elrodro83: no need to instantiate a Reference again

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
```diff
@@ -74,10 +89,17 @@ void release(Object connection) {
    * @param connection the connection to be invalidated
    */
   void invalidate(Object connection) {
-    ConnectionHandler<Object> connectionHandler = connections.remove(new Reference<>(connection));
-    if (connectionHandler != null) {
-      connectionHandler.invalidate();
-    }
+    Reference<Object> connReference = new Reference<>(connection);
```
elrodro83: too much in common with `release`. Factor out common code

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConnectionManager.java
```diff
@@ -51,8 +57,11 @@
   <T> T getConnection(ConfigurationInstance config) throws ConnectionException {
     ConnectionHandler<Object> connectionHandler = connectionManager.getConnection(config.getValue());
     Object connection = connectionHandler.getConnection();
-    connections.put(new Reference<>(connection), connectionHandler);
-
+    Reference<Object> connReference = new Reference<>(connection);
+    withLock(lock, () -> {
+      ((AtomicInteger) connections.computeIfAbsent(connReference, k -> new Pair(new AtomicInteger(0), connectionHandler))
```
marianogonzalez: computeIfAbsent method has performance issues. Replace this map with a caffeine cache.

PR #7833, opened 2019-04-24

NachoRaik:


afelisatti:
Same as 3.8.x, push the cherry-pick directly.

PR #7837, opened 2019-04-25
MULE-16756: Integrate all changes for HTTP request streaming

eze210:


core/src/main/java/org/mule/runtime/core/api/util/StreamingUtils.java
```diff
@@ -6,6 +6,7 @@
  */
 package org.mule.runtime.core.api.util;
 
+import static org.mule.runtime.api.util.MuleSystemProperties.MULE_DISABLE_REQUEST_STREAMING;
```
elrodro83: refactor this so this doesn't require changes in mule-api an can be used in runtime `4.2.0`

eze210:
These changes will be done in the http-service

PR #7839, opened 2019-04-25
MULE-16763: Choice (ExpressionFilter) does not act like Boolean.valueOf

eze210:


core/src/main/java/org/mule/api/config/MuleProperties.java
```diff
@@ -203,4 +203,5 @@
     public static final String MULE_FAIL_IF_DELETE_OPEN_FILE = SYSTEM_PROPERTY_PREFIX + "failIfDeleteOpenFile";
     public static final String CONTENT_TYPE_PROPERTY = "Content-Type";
     public static final String DISABLE_ERROR_COUNT_ON_ERROR_NOTIFICATION_DISABLED = SYSTEM_PROPERTY_PREFIX + "disable.error.count.on.error.notifications.disabled";
+    public static final String MULE_DEFAULT_BOOLEAN_VALUE = SYSTEM_PROPERTY_PREFIX + "default.boolean.value";
```
afelisatti: For mule-3.x (3.10.0) add the actual parameter, you can then send a PR to 3.9.x adding the system property instead, though I think we need to make its name more specific to where it is actually used.

core/src/test/java/org/mule/expression/ExpressionManagerTestCase.java
```diff
@@ -141,6 +148,74 @@ public void testBooleanEvaluation()
         assertFalse(muleContext.getExpressionManager().evaluateBoolean("string:false", msg));
     }
 
+    @Test
+    public void testDefaultExpressionFilterActsLikeBooleanValueOfInvalidString()
+    {
+        // Set default value
+        setProperty(MULE_DEFAULT_BOOLEAN_VALUE, "false");
```
afelisatti: If the assertion fails, the property will remain set throughout the modules tests. I'd recommend either extracting the tests to different classes and using the `SystemProperty` rule or using a try finally block very carefully.

core/src/test/java/org/mule/expression/ExpressionManagerTestCase.java
```diff
@@ -141,6 +148,124 @@ public void testBooleanEvaluation()
         assertFalse(muleContext.getExpressionManager().evaluateBoolean("string:false", msg));
     }
 
+    @Test
+    public void testDefaultExpressionFilterActsLikeBooleanValueOfInvalidStringUsingSystemProperty() throws Exception
+    {
+        testWithSystemProperty(MULE_EXPRESSION_FILTER_DEFAULT_BOOLEAN_VALUE, "false", new TestCallback()
```
afelisatti: I had no idea this existed...

core/src/test/java/org/mule/expression/ExpressionManagerTestCase.java
```diff
@@ -141,6 +148,124 @@ public void testBooleanEvaluation()
         assertFalse(muleContext.getExpressionManager().evaluateBoolean("string:false", msg));
     }
 
+    @Test
+    public void testDefaultExpressionFilterActsLikeBooleanValueOfInvalidStringUsingSystemProperty() throws Exception
```
afelisatti: Remove `test` prefixes

core/src/test/java/org/mule/expression/ExpressionManagerTestCase.java
```diff
@@ -141,6 +148,124 @@ public void testBooleanEvaluation()
...
+        {
+            @Override
+            public void run()
+            {
+                ExpressionFilter filter = new ExpressionFilter("payload");
+                assertThat(filter.isNonBooleanReturnsTrue(), is(false));
+
+                filter.setMuleContext(muleContext);
+
+                assertThat(filter.accept(new DefaultMuleMessage("on", muleContext)), is(valueOf("on")));
```
afelisatti: All these assertions are always repeated. Extract the logic to a reusable function.

core/src/main/java/org/mule/routing/filters/ExpressionFilter.java
```diff
@@ -53,6 +56,7 @@
     private ExpressionConfig config;
     private String fullExpression;
     private boolean nullReturnsTrue = false;
+    private boolean nonBooleanReturnsTrue = parseBoolean(getProperty(MULE_EXPRESSION_FILTER_DEFAULT_BOOLEAN_VALUE, "true"));
```
afelisatti: Remember to remove this from the 3.x solution, you can directly modify the `expression-filter` XSD to allow the new attribute there (you'll need a functional test there to validate it works).

tests/integration/src/test/java/org/mule/test/config/ExpressionFilterConfigTestCase.java
```diff
@@ -77,4 +81,44 @@ public void testConfig3() throws Exception
         assertEquals("something", filter.getCustomEvaluator());
         assertFalse(filter.isNullReturnsTrue());
     }
+
+    @Test
+    public void testConfigNonBooleanReturnsFalse() throws Exception
```
afelisatti: test prefixes

tests/integration/src/test/java/org/mule/test/config/ExpressionFilterConfigTestCase.java
```diff
@@ -77,4 +81,44 @@ public void testConfig3() throws Exception
...
+    {
+        EndpointBuilder eb = muleContext.getRegistry().lookupEndpointBuilder("endpoint4");
```
afelisatti: extract repeated logic

core/src/main/java/org/mule/routing/filters/ExpressionFilter.java
afelisatti: You can remove all System property related code and leave it for 3.9.x.

PR #7840, opened 2019-04-25
MULE-16730: Interception API: Validation all throws NullPointerException

elrodro83:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -255,7 +253,7 @@ private CoreEvent addContextToEvent(CoreEvent event, Context ctx) {
   }
 
   private Optional<ConfigurationInstance> resolveConfiguration(CoreEvent event) {
-    if (getLocation() != null && getPrecalculatedContext(event) != null) {
+    if (usePrecalculatedContext(event)) {
```
marianogonzalez: I should give you a comment like `rename to shouldUsePrecalculatedContext`, but unlike you, I'm a nice guy

elrodro83: just because you asked nicely :)

PR #7842, opened 2019-04-26
MULE-16771: Add shutdown hooks for deployment service

fsgonz:


modules/launcher/src/main/java/org/mule/module/launcher/DeploymentService.java
```diff
@@ -58,6 +58,19 @@
 
     void removeStartupListener(StartupListener listener);
 
+    /**
+     * Adds a shutdown listener to be triggered on deployment service shutdown
+     * 
+     * @param shutdown listener to be added
```
elrodro83: add `@since`

modules/launcher/src/main/java/org/mule/module/launcher/DeploymentService.java
```diff
@@ -58,6 +58,19 @@
...
+     */
+    void addShutdownListener(ShutdownListener listener);
+    
+    /**
+     * Removes a shutdown listener to be triggered on deployment service shutdown
+     * 
+     * @param shutdown listener to be removed
+     */
+    void removeShutdownListener(ShutdownListener listener);
```
elrodro83: add `@since`

PR #7849, opened 2019-05-02
MULE-16756: Integrate all changes for HTTP request streaming

eze210:


pom.xml
```diff
@@ -174,7 +174,7 @@
         <geronimoSpecsJaxRpcVersion>2.1</geronimoSpecsJaxRpcVersion>
         <geronimoSpecsServletApiVersion>1.0</geronimoSpecsServletApiVersion>
         <servletApiVersion>3.1.0</servletApiVersion>
-        <grizzlyVersion>2.3.36-MULE-006</grizzlyVersion>
+        <grizzlyVersion>2.3.36-MULE-008-SNAPSHOT</grizzlyVersion>
```
afelisatti: Do we need to move this? And if so, can we use a released version?

eze210: That is the grizzly version used by the AHC 009-SNAPSHOT. Shouldn't we use the same?

PR #7853, opened 2019-05-02
MULE-11900: Different default timeouts for connectionIdle on server a…

JuLuzzi:
…nd client


modules/http/src/main/java/org/mule/module/http/internal/listener/DefaultHttpListenerConfig.java
```diff
@@ -49,7 +50,9 @@
     public static final int DEFAULT_MAX_THREADS = 128;
     private final Logger logger = LoggerFactory.getLogger(getClass());
 
-    public static final int DEFAULT_CONNECTION_IDLE_TIMEOUT = 30 * 1000;
+    // The listener default timeout is bigger than our requester default timeout to avoid 'Remotely closed' exception
+    // when you start sending a request on an existing connection just before the timeout occurs.
+    public static final int DEFAULT_CONNECTION_IDLE_TIMEOUT = getDefaultConnectionIdleTimeout() + 10000;
```
afelisatti: Avoid the method call and simply import the constant (you won’t be able to do a static import but that is fine)

PR #7862, opened 2019-05-09
MULE-10702: Remove double check locking: StaticBulkQueryResolver

JuLuzzi:


modules/container/src/main/java/org/mule/runtime/container/internal/MuleClassLoaderLookupPolicy.java
```diff
@@ -98,21 +99,15 @@ public LookupStrategy getClassLookupStrategy(String className) {
 
   @Override
   public LookupStrategy getPackageLookupStrategy(String packageName) {
-    LookupStrategy lookupStrategy = lookupStrategies.get(packageName);
```
elrodro83: beware, this is not initializing the map, but rather populating it. Don't think we need to do any change here for tis class.

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultApplicationPolicyInstance.java
```diff
@@ -172,14 +176,7 @@ public PolicyTemplate getPolicyTemplate() {
...
   public void initialise() throws InitialisationException {
-    if (policyInstance == null) {
-      synchronized (this) {
-        if (policyContext == null) {
-          initPolicyContext();
-        }
-        initPolicyInstance();
-      }
-    }
+    policyInstance = new LazyValue<>(this.initPolicyInstance());
```
elrodro83: `new LazyValue<>(() -> this.initPolicyInstance())`, otherwise this is not lazy

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultApplicationPolicyInstance.java
```diff
@@ -151,8 +152,11 @@ private void enableNotificationListeners(List<NotificationListener> notification
     notificationListeners.forEach(listenerRegistry::registerListener);
   }
 
-  private void initPolicyInstance() throws InitialisationException {
-    policyInstance = policyContext.getRegistry().lookupByType(PolicyInstance.class).get();
+  private PolicyInstance initPolicyInstance() throws InitialisationException {
+    if (policyContext == null) {
+      initPolicyContext();
```
elrodro83: this should be set already, why do you need to add this?

PR #7865, opened 2019-05-09
MULE-16824: LazyInit should initialize the Mule configuration at part…

gsfernandes:
… of the component initializations phase

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/config/DefaultMuleConfiguration.java
+++ b/core/src/main/java/org/mule/runtime/core/api/config/DefaultMuleConfiguration.java
@@ -54,7 +54,7 @@ public class DefaultMuleConfiguration implements MuleConfiguration, MuleContextA
 
   protected static final Logger logger = getLogger(DefaultMuleConfiguration.class);
 
-  private Boolean lazyInit = null;
+  private boolean lazyInit;
 
   /**
    * When true, each event will keep trace information of the flows and components it traverses to be shown as part of an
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/mule/runtime/core/api/config/DefaultMuleConfiguration.java
```diff
@@ -476,6 +485,14 @@ public void setEnableStreaming(boolean enableStreaming) {
     }
   }
 
+  private boolean isLazyInit() {
```
lucianoRM: Will this method be called multiple times? Can the config value change?
Wouldn't it be better to use a LazyValue to avoid resolving the configuration property everytime?

modules/spring-config/src/main/java/org/mule/runtime/config/internal/LazyMuleArtifactContext.java
```diff
@@ -446,6 +447,21 @@ private void resetMuleSecurityManager() {
...
   }
 
+  private void resetMuleConfiguration(MinimalApplicationModelGenerator minimalApplicationModelGenerator) {
+    // Always unregister first the default configuration from Mule.
+    try {
+      muleContext.getRegistry().unregisterObject(OBJECT_MULE_CONFIGURATION);
+    } catch (Exception e) {
+      // NoSuchBeanDefinitionException can be ignored
+      if (!hasCause(e, NoSuchBeanDefinitionException.class)) {
+        throw new MuleRuntimeException(I18nMessageFactory.createStaticMessage("Error while unregistering Mule configuration"),
```
lucianoRM: static import

core/src/main/java/org/mule/runtime/core/api/config/DefaultMuleConfiguration.java
gsfernandes: ConfigurationProperties already has a cache to avoid resolving through the whole hierarchy the same key over and over.

gsfernandes: I couldn't find where the cache is so I ended caching the lazyInit flag here.

PR #7891, opened 2019-05-20
MTS-130: Improve maven client's dependency resolution.

lucianoRM:


gsfernandes:
Change the Jira and PR name (and commit)

PR #7901, opened 2019-05-21
MULE-16835: DOM to XML: Maximum attribute size limit (524288) exceeded with a large payload

JuLuzzi:


core/src/main/java/org/mule/util/xmlsecurity/DefaultXMLSecureFactories.java
```diff
@@ -12,24 +12,25 @@
 import static javax.xml.XMLConstants.ACCESS_EXTERNAL_STYLESHEET;
 import static javax.xml.stream.XMLInputFactory.IS_SUPPORTING_EXTERNAL_ENTITIES;
 import static javax.xml.stream.XMLInputFactory.SUPPORT_DTD;
+import static javax.xml.stream.XMLInputFactory.newInstance;
+import static org.apache.commons.logging.LogFactory.getLog;
```
elrodro83: we should be using slf4j for this

PR #7903, opened 2019-05-22
MULE-16878: StringToEnum Transformer should have different names for each extensionModel

mbuchwald:


core/src/main/java/org/mule/runtime/core/internal/transformer/simple/StringToEnum.java
```diff
@@ -33,13 +33,13 @@
...
    */
-  public StringToEnum(Class<? extends Enum> enumClass) {
+  public StringToEnum(Class<? extends Enum> enumClass, String namespace) {
     checkArgument(enumClass != null, "enumClass cannot be null");
     this.enumClass = enumClass;
 
     registerSourceType(DataType.fromType(String.class));
     setReturnDataType(DataType.fromType(enumClass));
-    setName(format("StringTo%sTransformer", enumClass.getSimpleName()));
+    setName(format("StringTo%s%sTransformer", namespace, enumClass.getSimpleName()));
```
marianogonzalez: don't add this namespace thing. Just simply use the enumClass.getName()

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/ExtensionActivator.java
```diff
@@ -80,7 +80,7 @@ private void registerEnumTransformers(ExtensionModel extensionModel) {
           final Class<Enum> enumClass = (Class<Enum>) type;
           if (enumTypes.add(enumClass)) {
             try {
-              StringToEnum stringToEnum = new StringToEnum(enumClass);
+              StringToEnum stringToEnum = new StringToEnum(enumClass, extensionModel.getName());
               registerObject(muleContext, getName(stringToEnum), stringToEnum, Transformer.class);
```
marianogonzalez: can you add an integration test which verifies the case in which two different extensions register enums with the exact same FQN ?

PR #7907, opened 2019-05-23
MULE-16889: DefaultDomainManager#getDomain should retrieve a compatible version of the required domain

eze210:


eze210:
--validate

eze210:
--validate

eze210:
--validate

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/domain/BundleDescriptorWrapper.java
```diff
@@ -0,0 +1,84 @@
...
+import org.mule.runtime.deployment.model.api.domain.DomainDescriptor;
+import org.mule.runtime.module.artifact.api.descriptor.BundleDescriptor;
+
+import com.vdurmont.semver4j.Semver;
+
+/**
+ * Two instances of this class wrapping instances of {@link BundleDescriptor} that matches in all parameters but the minor and
+ * the patch versions are equals and have the same hash.
+ */
+class BundleDescriptorWrapper {
```
gsfernandes: I will call this SemVerBundleDescriptorWrapper. I will add @since too.

PR #7919, opened 2019-05-27
MULE-16904: Differentitate between ldap's server wrong password and invalid credentials

NachoRaik:


core/src/main/java/org/mule/runtime/core/internal/security/DefaultMuleSecurityManager.java
```diff
@@ -80,6 +80,14 @@ public Authentication authenticate(Authentication authentication) throws Securit
         Authentication result = null;
         try {
           result = provider.authenticate(authentication);
+        } catch (UnauthorisedException e) {
```
elrodro83: this `catch` can be removed since it does the same as the `catch Exception` at the bottom

core-tests/src/test/java/org/mule/runtime/core/internal/security/DefaultMuleSecurityManagerTestCase.java
```diff
@@ -0,0 +1,112 @@
...
+    manager = new DefaultMuleSecurityManager();
+  }
+
+  @Test
+  public void authenticateCorrectly()
+          throws SecurityException, SecurityProviderNotFoundException {
+    when(provider.authenticate(authentication)).thenReturn(authentication);
+    manager.addProvider(provider);
+
+    Optional<Authentication> optional = Optional.ofNullable(manager.authenticate(authentication));
```
elrodro83: static import

core-tests/src/test/java/org/mule/runtime/core/internal/security/DefaultMuleSecurityManagerTestCase.java
```diff
@@ -0,0 +1,112 @@
...
+public class DefaultMuleSecurityManagerTestCase extends AbstractMuleTestCase {
+
+  @Mock
+  private SecurityProvider provider;
+
+  @Mock
+  private Authentication authentication;
+
+  @Rule
+  public ExpectedException expected = ExpectedException.none();
```
elrodro83: static import

core-tests/src/test/java/org/mule/runtime/core/internal/security/DefaultMuleSecurityManagerTestCase.java
```diff
@@ -0,0 +1,112 @@
...
+  public void authenticateCorrectly()
+          throws SecurityException, SecurityProviderNotFoundException {
+    when(provider.authenticate(authentication)).thenReturn(authentication);
+    manager.addProvider(provider);
+
+    Optional<Authentication> optional = Optional.ofNullable(manager.authenticate(authentication));
+
+    verify(provider).authenticate(authentication);
+
+    assertThat(optional.isPresent(), is(true));
```
elrodro83: if the value is not present, the next assertion would fail with `NoSuchElementException`, so this one here is superfluous

core-tests/src/test/java/org/mule/runtime/core/internal/security/DefaultMuleSecurityManagerTestCase.java
```diff
@@ -0,0 +1,112 @@
...
+
+  @Test
+  public void authenticateWithWrongAuthentication()
+          throws SecurityException, SecurityProviderNotFoundException {
+    when(provider.authenticate(authentication)).thenThrow(UnauthorisedException.class);
+    manager.addProvider(provider);
+
+    expected.expect(UnauthorisedException.class);
+
+    manager.authenticate(authentication);
```
elrodro83: wrap this in a `try`...

core-tests/src/test/java/org/mule/runtime/core/internal/security/DefaultMuleSecurityManagerTestCase.java
```diff
@@ -0,0 +1,112 @@
...
+
+    verify(provider).authenticate(authentication);
```
elrodro83: ... and this in a `finally`, otherwise, the exception will cause the test to end and this line will never be executed

core-tests/src/test/java/org/mule/runtime/core/internal/security/DefaultMuleSecurityManagerTestCase.java
```diff
@@ -0,0 +1,112 @@
...
+
+  @Test
+  public void authenticateWithFailedProvider()
+          throws SecurityException, SecurityProviderNotFoundException {
+    when(provider.authenticate(authentication)).thenThrow(SecurityException.class);
+    manager.addProvider(provider);
+
+    expected.expect(SecurityException.class);
+
+    manager.authenticate(authentication);
```
elrodro83: same as the previous method

core-tests/src/test/java/org/mule/runtime/core/internal/security/DefaultMuleSecurityManagerTestCase.java
```diff
@@ -0,0 +1,112 @@
...
+
+  @Test
+  public void authenticateWithoutProvider()
+          throws SecurityException, SecurityProviderNotFoundException {
+    expected.expect(SecurityProviderNotFoundException.class);
+
+    manager.authenticate(authentication);
+
+    verify(provider).authenticate(authentication);
+    verifyZeroInteractions(provider);
```
elrodro83: isn't this contradictory with the line right above?

core/src/main/java/org/mule/runtime/core/internal/security/DefaultMuleSecurityManager.java
NachoRaik: The problem is that `UnauthorisedException` is a subtype of `SecurityException`, so in that case it will get into that `catch` instead of the `catch (Exception e)` one

NachoRaik: An alternative could be leaving just the SecurityException catch, but adding there:
``` 
catch (SecurityException e) {
    if (!iter.hasNext()) {
        if (e instanceof UnauthorisedException) {
            throw new UnauthorisedException(authorizationAttemptFailed(), e);
        } else {
            throw new SecurityException(authorizationAttemptFailed(), e);
        }
    }
} ...
```

PR #7921, opened 2019-05-27

JuLuzzi:


core/src/main/java/org/mule/util/JdkVersionUtils.java
```diff
@@ -223,20 +228,20 @@ public boolean isUnder(JdkVersion jdkVersion)
         }
     }
     
-    private static final Log logger = LogFactory.getLog(JdkVersionUtils.class);
+    private static final Log logger = getLog(JdkVersionUtils.class);
```
elrodro83: replace with slf4j

PR #7922, opened 2019-05-27
MULE-16896: Bound connections not released on TransactionException

elrodro83:
* addendum

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/DefaultSourceCallbackContext.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/DefaultSourceCallbackContext.java
@@ -75,18 +75,20 @@ class DefaultSourceCallbackContext implements SourceCallbackContextAdapter {
 
     this.connection = connection;
 
-    if (sourceCallback.getTransactionConfig().isTransacted() && connection instanceof TransactionalConnection) {
-      ConnectionHandler<Object> connectionHandler = sourceCallback.getSourceConnectionManager().getConnectionHandler(connection)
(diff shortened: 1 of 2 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/DefaultSourceCallbackContext.java
```diff
@@ -75,18 +75,20 @@ public TransactionHandle bindConnection(Object connection) throws ConnectionExce
...
+        ConnectionHandler<Object> connectionHandler = sourceCallback.getSourceConnectionManager().getConnectionHandler(connection)
+            .orElseThrow(() -> new TransactionException(createWrongConnectionMessage(connection)));
+
+        sourceCallback.getTransactionSourceBinder().bindToTransaction(sourceCallback.getTransactionConfig(),
+                                                                      sourceCallback.getConfigurationInstance(),
+                                                                      sourceCallback.getSourceLocation(),
+                                                                      connectionHandler);
+        transactionHandle = DEFAULT_TRANSACTION_HANDLE;
+      }
+    } catch (TransactionException txE) {
```
marianogonzalez: why this one and not any generic one? Shouldn't the same happen for any unexpected exception?

elrodro83: this one specifically is the one that is handled in https://github.com/mulesoft/mule-jms-client/blob/master/src/main/java/org/mule/jms/commons/internal/source/polling/JmsXaMessageConsumer.java#L184 and because of the way it is handled causes the exception

PR #7923, opened 2019-05-27
MULE-16908 & MULE-16909: Support Client Credentials gran type in the SDK - Support custom parameters in OAuth Dance

marianogonzalez:


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/connection/ConnectionProviderObjectFactory.java
```diff
@@ -90,6 +90,45 @@ public ConnectionProviderResolver doGetObject() {
     return new ConnectionProviderResolver<>(builder, resolverSet, muleContext);
   }
 
+  private ConnectionProviderObjectBuilder resolveOAuthBuilder(ResolverSet resolverSet) {
+    OAuthGrantType grantType = providerModel.getModelProperty(OAuthModelProperty.class)
+        .map(OAuthModelProperty::getGrantTypes)
+        .get().get(0);
```
juandesi: get().get(0) seems super harsh, use ifPresent + forEach or fail if more thab 1

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/JavaOAuthDeclarationEnricher.java
```diff
@@ -170,16 +187,20 @@ private OperationDeclaration buildUnauthorizeOperation() {
...
-      parameter.setDescription("The id of the resource owner which access should be invalidated");
-      parameter.setExpressionSupport(SUPPORTED);
-      parameter.setLayoutModel(LayoutModel.builder().build());
-      parameter.setRequired(false);
-      parameter.setParameterRole(BEHAVIOUR);
-      parameter.setType(stringType, false);
-      group.addParameter(parameter);
+      if (supportsAuthCode) {
+        ParameterGroupDeclaration group = operation.getParameterGroup(DEFAULT_GROUP_NAME);
+        group.showInDsl(false);
```
juandesi: ? why is it doing this to the default group?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/OAuthConnectionProviderModelValidator.java
```diff
@@ -36,24 +38,48 @@ public void validate(ExtensionModel model, ProblemsReporter problemsReporter) {
...
-                                                                            format("Connection Provider of class '%s' uses OAuth2 authorization code grant type and thus should contain "
-                                                                                + "one (and only one) field of type %s. %d were found",
-                                                                                   type,
-                                                                                   AuthorizationCodeState.class.getName(),
-                                                                                   stateFields.size()));
-              }
-            });
+            .orElse(null);
+
+        if (implementingType == null) {
```
juandesi: you could use ifPresent

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OAuthOperationMessageProcessor.java
```diff
@@ -96,23 +99,45 @@ public void error(Throwable e) {
...
+          }
+        });
+
+        Optional<String> resourceOwnerId = resourceOwnerIdReference.get();
 
-        String ownerConfigName = operationContext.getConfiguration().get().getName();
         try {
-          oauthManager.refreshToken(ownerConfigName, expiredException.getResourceOwnerId(),
-                                    getOAuthConnectionProvider(operationContext));
+          connectionProvider.refreshToken(resourceOwnerId.orElse(""));
```
juandesi: should,t you fail in this case when notPresent?

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/connection/ConnectionProviderObjectFactory.java
marianogonzalez: the get(0) is because of a design error in 4.1 giving the property a List of grant types instead of just one. That list will never be empty and will never have more than one item

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/JavaOAuthDeclarationEnricher.java
marianogonzalez: client credentials doesn't have the concept of resourceOwnerId. AUth code is the only grant type that supports the concept of multitenancy

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OAuthOperationMessageProcessor.java
marianogonzalez: nope. The concept doesn't apply to all grant types

PR #7928, opened 2019-05-29
MULE-16786: Dynamic configurations eviction should consider sources as in-flight operations

ndinu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -198,7 +202,12 @@ private RetryPolicyTemplate createRetryPolicyTemplate(RetryPolicyTemplate custom
   private void stopSource() throws MuleException {
     if (sourceAdapter != null) {
       try {
+        CoreEvent initialiserEvent = getInitialiserEvent(muleContext);
+        stopUsingConfiguration(initialiserEvent);
         sourceAdapter.stop();
+        if (usesDynamicConfiguration()) {
+          disposeSource();
```
marianogonzalez: there's a bug here. If you dispose the source here, then it cannot be started again without invoking initialise() first. Add a test case in which :

1) the feature is tested
2) the source is stopped
3) the source is started
4) the feature is tested again and still works

ndinu: I will add this test, but I think this will work. When starting if the source adapter is null, it will recreate it. In the restart method we already do this :  stop() disposeSource() start().

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/DynamicConfigExpirationTestCase.java
```diff
@@ -42,12 +51,31 @@ public void expireDynamicConfigWithCustomExpiration() throws Exception {
       assertExpired(config, 1500, 100);
       fail("Config should not have been expired");
     } catch (AssertionError e) {
-      //all good
+      assertThat(e.getMessage(), is(not("Config should not have been expired")));
```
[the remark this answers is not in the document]
ndinu: I changed this because fail() will throw an asertion error. So this part of the test will always pass.

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/DynamicConfigExpirationTestCase.java
```diff
@@ -42,12 +51,31 @@ public void expireDynamicConfigWithCustomExpiration() throws Exception {
...
+  @Test
+  public void doNotExpireDynamicConfigWithCustomExpirationUsedBySource() throws Exception {
+    HeisenbergExtension config =
+        invokeDynamicConfig("dynamicWithCustomExpirationForSource", "heisenbergWithCustomExpirationForSource", "Walter Blanco");
+
+    try {
+      assertExpired(config, 10000, 1000);
+      fail("Config should not have been expired");
+    } catch (AssertionError e) {
+      assertThat(e.getMessage(), is(not("Config should not have been expired")));
```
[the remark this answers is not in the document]
ndinu: Agree

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/DynamicConfigExpirationTestCase.java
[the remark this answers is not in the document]
marianogonzalez: That is not the point of the comment. The point is to leave this test as it was, asserting the case in which there's no source involved. Then add a separate test in which do add the source into the mix.

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/source/HeisenbergMessageSourceTestCase.java
```diff
@@ -107,6 +109,29 @@ public void source() throws Exception {
     assertSourceCompleted();
   }
 
+  @Test
+  public void sourceRestartedWithDynamicConfig() throws Exception {
+    Long gatheredMoney = HeisenbergSource.gatheredMoney;
+    startFlow("source");
+
+    Thread.sleep(TIME_WAIT_MILLIS);
```
marianogonzalez: don't do this with a sleep. Do this with a PollingProber.

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/source/HeisenbergMessageSourceTestCase.java
```diff
@@ -107,6 +109,29 @@ public void source() throws Exception {
...
+    // Check that money is gathered when flow started
+    assertThat(HeisenbergSource.gatheredMoney, greaterThan(gatheredMoney));
+
+    stopFlow("source");
+
+    gatheredMoney = HeisenbergSource.gatheredMoney;
+    Thread.sleep(TIME_WAIT_MILLIS);
```
marianogonzalez: no sleeps

estebanwasinger:
--validate

ndinu:
Jenkins job to validate this PR has been started manually.

PR #7932, opened 2019-05-30
MULE-16924: Support lightweight deployment with class loader model an…

gsfernandes:
…d local repository


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/AbstractMavenClassLoaderModelLoader.java
```diff
@@ -112,46 +109,65 @@ public final ClassLoaderModel load(File artifactFile, Map<String, Object> attrib
...
-        packagerClassLoaderModel.getDependencies().stream().map(artifact -> {
-          Optional<BundleDependency> patchedBundledDependency =
-              patchBundleDependencies.stream().filter(bundleDependency -> bundleDependency.getDescriptor().getGroupId()
-                  .equals(artifact.getArtifactCoordinates().getGroupId()) &&
-                  bundleDependency.getDescriptor().getArtifactId().equals(artifact.getArtifactCoordinates().getArtifactId()))
-                  .findAny();
-          return patchedBundledDependency
-              .orElse(createBundleDependencyFromPackagerDependency(deployableArtifactRepositoryFolder).apply(artifact));
-        }).collect(toSet());
+        .exportingPackages(new HashSet<>(
```
lucianoRM: was this formatted?

gsfernandes: yes, I can remove that new line but formatter was "ok"

PR #7945, opened 2019-06-03
MULE-16934: Remove unused external transactions support

elrodro83:


tests/test-extensions/transactional-extension/src/main/java/org/mule/test/transactional/TransactionalSource.java
```diff
@@ -65,12 +69,20 @@ public void onStart(SourceCallback<TestTransactionalConnection, Object> sourceCa
...
         sourceCallback.onConnectionException(e);
       } catch (Exception e) {
         throw new RuntimeException(e);
+      } finally {
+        try {
+          if (txHandle != null) {
+            txHandle.commit();
+          }
+        } catch (TransactionException e) {
+          throw new RuntimeException(e);
```
marianogonzalez: shouldn't you rollback here?

tests/test-extensions/transactional-extension/src/main/java/org/mule/test/transactional/TransactionalSource.java
```diff
@@ -81,6 +81,11 @@ public void onStart(SourceCallback<TestTransactionalConnection, Object> sourceCa
             txHandle.commit();
           }
         } catch (TransactionException e) {
+          try {
+            txHandle.rollback();
+          } catch (TransactionException e1) {
+            throw new RuntimeException(e1);
```
marianogonzalez: I don't think this is correct. If the rollback fails you should log that but the exception raised should still be the original one. Otherwise you're hiding the root cause

PR #7940, opened 2019-05-31
MULE-16936: sinks must be used by only one thread when running a transaction

mbuchwald:
The test case is the Perf test running correctly :)

PR #7948, opened 2019-06-03
MULE-16787: Reuse route chains in error handlers

afelisatti:


core/src/main/java/org/mule/runtime/core/privileged/exception/TemplateOnErrorHandler.java
```diff
@@ -76,6 +91,29 @@
   protected String errorType = null;
   protected ErrorTypeMatcher errorTypeMatcher = null;
 
+  private final Supplier<FluxSink<CoreEvent>> routingSink = new RoundRobinFluxSinkSupplier<>(getRuntime().availableProcessors(),
```
elrodro83: Use an approach similar to https://github.com/mulesoft/mule/pull/7940#pullrequestreview-245036622 for when there are transactions

core/src/main/java/org/mule/runtime/core/privileged/exception/TemplateOnErrorHandler.java
```diff
@@ -76,6 +91,29 @@
...
+
+  private final class OnErrorHandlerFluxObjectFactory implements Supplier<FluxSink<CoreEvent>> {
+
+    @Override
+    public FluxSink<CoreEvent> get() {
+      final FluxSinkRecorder<CoreEvent> sinkRef = new FluxSinkRecorder<>();
+      Flux<CoreEvent> onErrorFlux =
+          Flux.create(sinkRef)
+              .map(beforeRouting())
+              .concatMap(route())
```
elrodro83: this is as bad as `flatMap`, we need to look further into this in a next iteration

afelisatti:
--validate

PR #7963, opened 2019-06-10
MULE-16983: Reconnection should not retry when resource has joined ac…

marianogonzalez:
…tive transaction

### DIFF
--- a/core-tests/src/test/java/org/mule/runtime/core/internal/connection/CachedConnectionHandlerTestCase.java
+++ b/core-tests/src/test/java/org/mule/runtime/core/internal/connection/CachedConnectionHandlerTestCase.java
@@ -6,6 +6,7 @@
  */
 package org.mule.runtime.core.internal.connection;
 
+import static java.lang.Thread.currentThread;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.sameInstance;
 import static org.junit.Assert.assertThat;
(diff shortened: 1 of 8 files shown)


core-tests/src/test/java/org/mule/runtime/core/internal/connection/CachedConnectionHandlerTestCase.java
```diff
@@ -121,6 +123,29 @@ public void invalidate() throws Exception {
...
   }
 
+  @Test
+  public void concurrentInvalidate() throws Exception {
+    for (int i = 0; i < 50; i++) {
+      reset(connectionProvider);
+      getConnection();
+      Latch latch = new Latch();
+
+      new Thread(() -> {
```
elrodro83: use and executor and shutdown it, or join all threads created at the end

core-tests/src/test/java/org/mule/runtime/core/internal/connection/CachedConnectionHandlerTestCase.java
```diff
@@ -121,6 +123,29 @@ public void invalidate() throws Exception {
...
+        try {
+          latch.await();
+          managedConnection.invalidate();
+        } catch (InterruptedException e) {
+          throw new RuntimeException(e);
```
elrodro83: reset the interrupted flag

PR #7970, opened 2019-06-12
MULE-16994: Use of cache instead of WeakHashMap to avoid infinite loop in DefaultLifecycleInterceptor.

fsgonz:
in DefaultLifecycleInterceptor.


fsgonz:
--validate

PR #7971, opened 2019-06-12

JuLuzzi:


core/src/main/java/org/mule/runtime/core/api/util/SystemUtils.java
```diff
@@ -62,16 +63,16 @@ public static String getenv(String name) {
   }
 
   public static boolean isSunJDK() {
-    return JAVA_VM_VENDOR.toUpperCase().indexOf("SUN") != -1
-        || JAVA_VM_VENDOR.toUpperCase().indexOf("ORACLE") != -1;
+    return JAVA_VM_VENDOR.toUpperCase().contains("sun")
```
elrodro83: `toLowerCase`

core/src/main/java/org/mule/runtime/core/api/util/SystemUtils.java
```diff
@@ -62,16 +63,16 @@ public static String getenv(String name) {
...
+        || JAVA_VM_VENDOR.toUpperCase().contains("oracle");
```
elrodro83: `toLowerCase`

core/src/main/java/org/mule/runtime/core/api/util/SystemUtils.java
```diff
@@ -62,16 +63,16 @@ public static String getenv(String name) {
...
   }
 
   public static boolean isAppleJDK() {
-    return JAVA_VM_VENDOR.toUpperCase().indexOf("APPLE") != -1;
+    return JAVA_VM_VENDOR.toUpperCase().contains("apple");
```
elrodro83: `toLowerCase`

PR #7972, opened 2019-06-12
MULE-16995: Shared runtime deps in app mess up classloading

elrodro83:
* Also, parameterize `ApplicationDeploymentTestCase` to cover both versions of the classloaderModel


modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/builder/DeployableFileBuilder.java
```diff
@@ -38,9 +39,10 @@
 public abstract class DeployableFileBuilder<T extends DeployableFileBuilder<T>> extends AbstractArtifactFileBuilder<T> {
 
   protected Properties deployProperties = new Properties();
-  private Map<PluginAdditionalDependenciesKey, List<JarFileBuilder>> additionalPluginDependencies = new HashMap<>();
+  private final Map<PluginAdditionalDependenciesKey, List<JarFileBuilder>> additionalPluginDependencies = new HashMap<>();
 
   private boolean useHeavyPackage = true;
+  private String classloaderModelVersion = "1.0";
```
gsfernandes: Change this to static field

elrodro83: nope, this is set for each instance

PR #7974, opened 2019-06-13
EE-6697: Allow to override the server id in AbstractMuleContextTestCase

eze210:


tests/unit/src/main/java/org/mule/tck/junit4/AbstractMuleContextTestCase.java
```diff
@@ -253,7 +257,7 @@ protected MuleContext createMuleContext() throws Exception {
         String workingDirectory = this.workingDirectory.getRoot().getAbsolutePath();
         LOGGER.info("Using working directory for test: " + workingDirectory);
         muleConfiguration.setWorkingDirectory(workingDirectory);
-        muleConfiguration.setId(this.getClass().getSimpleName() + "#" + name.getMethodName());
+        muleConfiguration.setId(serverId);
```
fsgonz: I would change this for contextConfigurationId as this is a general test feature.

PR #7978, opened 2019-06-14
MULE-17006: add equals/hashcode to proxy configs

elrodro83:


modules/http-api/src/main/java/org/mule/runtime/http/api/client/proxy/DefaultNtlmProxyConfig.java
```diff
@@ -22,4 +24,36 @@
   public String getNtlmDomain() {
     return ntlmDomain;
   }
+
+  @Override
+  public int hashCode() {
+    final int prime = 31;
```
marianogonzalez: use Objects.hash()

modules/http-api/src/main/java/org/mule/runtime/http/api/client/proxy/DefaultNtlmProxyConfig.java
```diff
@@ -27,10 +29,7 @@ public String getNtlmDomain() {
 
   @Override
   public int hashCode() {
-    final int prime = 31;
-    int result = super.hashCode();
-    result = prime * result + ((ntlmDomain == null) ? 0 : ntlmDomain.hashCode());
-    return result;
+    return hash(getHost(), getPort(), getPassword(), getUsername(), getNonProxyHosts(), ntlmDomain);
```
marianogonzalez: shouldn't this be `hash(super.hashCode, ntlmDomain)`... if you do it like this, any changes in the super class will break here.

PR #7979, opened 2019-06-14

JuLuzzi:


core/src/main/java/org/mule/runtime/core/internal/util/JdkVersionUtils.java
```diff
@@ -6,8 +6,9 @@
  */
 package org.mule.runtime.core.internal.util;
 
+import static java.lang.Boolean.*;
```
elrodro83: explicitly import, no *

core/src/main/java/org/mule/runtime/core/internal/util/JdkVersionUtils.java
```diff
@@ -295,6 +296,8 @@ public static void validateJdk() throws RuntimeException {
     if (!isSupportedJdkVendor()) {
       logger.info("You're executing with a JDK made by a vendor that is not on the recommended list of vendors. Vendor: "
           + JAVA_VENDOR + " Please consider changing to a recommended JDK vendor.");
+    } else if (getBoolean(getProperty("mule.jdkDebug"))) {
```
elrodro83: just `getBoolean`

elrodro83: make the property into a constant, use the MULE prefix from the `MuleProperties`

PR #7989, opened 2019-06-19
MULE-16984: Interception API: Smart connectors inside a choice router…

JuLuzzi:
… are not skipped properly


tests/allure/src/main/java/org/mule/test/allure/AllureConstants.java
```diff
@@ -42,6 +42,12 @@
     interface ConfigurationComponentLocationStory {
 
       String COMPONENT_LOCATION = "Component location";
+
+      interface LazyConnectionsStory {
+
+        String COMPONENT_LOCATION_STORY = "Component location Story";
```
elrodro83: why do you need this nested story? why not use `COMPONENT_LOCATION`?

PR #7991, opened 2019-06-22
MULE-17010: OAuth uses wrong access token when multiple concurrent refresh requests are sent

elrodro83:


modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/state/DefaultResourceOwnerOAuthContext.java
```diff
@@ -90,7 +106,33 @@ public String getResourceOwnerId() {
...
+   * @deprecated use {@link #getRefreshUserOAuthContextLock(String, LockFactory)} instead of the lock provided here
+   */
+  @Deprecated
   public void setRefreshUserOAuthContextLock(Lock refreshUserOAuthContextLock) {
     this.refreshUserOAuthContextLock = refreshUserOAuthContextLock;
   }
+
+  @Override
+  public DancerState getDancerState() {
+    return dancerState != null ? dancerState : HAS_TOKEN;
```
marianogonzalez: shouldn't NO_TOKEN be the default?

modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/state/ResourceOwnerOAuthContext.java
```diff
@@ -47,4 +52,26 @@
...
+   */
+  DancerState getDancerState();
+
+  /**
+   * @param lockNamePrefix a prefix to uniquely identify the locks in the provided {@code lockFactory}.
+   * @param lockFactory the object to get the locks from. This will ensure that the same lock instance is used after
+   *        serialization/deserialization of this context.
+   * @return a lock that can be used to avoid concurrency problems trying to update oauth context.
+   */
+  Lock getRefreshUserOAuthContextLock(String lockNamePrefix, LockFactory lockFactory);
```
marianogonzalez: add a description to the Javadoc. Also, why is the word `User` in this method name?

modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/state/DefaultResourceOwnerOAuthContext.java
elrodro83: By default, new instances are built with NO_TOKEN. This condition here accounts for the case where the object was generated by a previous version where this attribute wasn't there. In that case, the equivalent state for that is HAS_TOKEN.

modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/builder/OAuthDancerBuilder.java
```diff
@@ -21,10 +21,18 @@
  */
 public interface OAuthDancerBuilder<D> {
 
+  /**
+   * @param name the name for this dancer to use for ids generation
+   * @return this builder
```
marianogonzalez: add `@since`

modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/builder/OAuthDancerBuilder.java
```diff
@@ -21,10 +21,18 @@
...
+   */
+  default OAuthDancerBuilder<D> name(String name) {
+    return this;
```
marianogonzalez: https://i.imgur.com/KHB2PRx.jpg

modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/state/DefaultResourceOwnerOAuthContext.java
```diff
@@ -93,4 +100,19 @@ public String getResourceOwnerId() {
   public void setRefreshUserOAuthContextLock(Lock refreshUserOAuthContextLock) {
     this.refreshUserOAuthContextLock = refreshUserOAuthContextLock;
   }
+
+  @Override
+  public DancerState getDancerState() {
+    return null;
```
marianogonzalez: do not return null. Use a default value or turn this into an Optional

modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/state/DefaultResourceOwnerOAuthContext.java
```diff
@@ -93,4 +100,19 @@ public String getResourceOwnerId() {
...
+  }
+
+  @Override
+  public void setDancerState(DancerState dancerState) {
+    // Nothing to do
```
marianogonzalez: why? why can't this base class have a default handling on this? This API is lying to the people

PR #7992, opened 2019-06-24
 MULE-17045: Connections are not closed when streaming operation fails

ndinu:


ndinu:
--validate

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/ConnectionInterceptor.java
```diff
@@ -65,10 +65,11 @@ public void before(ExecutionContext<ComponentModel> executionContext) throws Exc
...
   @Override
   public Throwable onError(ExecutionContext<ComponentModel> executionContext, Throwable exception) {
-    extractConnectionException(exception).ifPresent(
-                                                    e -> setCloseCommand(executionContext,
-                                                                         () -> withConnection(executionContext,
-                                                                                              ConnectionHandler::invalidate)));
+    if (extractConnectionException(exception).isPresent()) {
+      setCloseCommand(executionContext, () -> withConnection(executionContext, ConnectionHandler::invalidate));
+    } else {
+      setCloseCommand(executionContext, () -> release(executionContext));
```
marianogonzalez: are you doing this because of an actual bug? Isn't this exactly what happens in the after() method? Actually, doesn't this make the release method to be called twice?

ndinu: In the before method, if it does support streaming, we do not set the close command, since the ConnectedInputStreamWrapper will be in charge of closing the connector. But if the operation fails, the ConnectedInputStreamWrapper will never be created and the connection will be leaked.(This is the bug that is being solved)

This method (onError) used to only set a close command when the error was originated from a ConnectionException because in the after method we also want to invalidate the connection there.

Now we also set the close command there to release the connection because if the operation supported streaming we did not add it before.

Here I added that if the error was not originated from a ConnectionException, set the close comand to release the connection, if the operation supports streaming it will be added.

When you set a command it overrides the command that was set before, so it will not be called twice.

PR #7998, opened 2019-06-25
MULE-17051: Default Error Handler should be looked up after initialization

mbuchwald:


core/src/main/java/org/mule/runtime/core/api/config/DefaultMuleConfiguration.java
```diff
@@ -771,4 +784,24 @@ public boolean equals(Object obj) {
...
+    String defaultErrorHandler = getDefaultErrorHandlerName();
+    if (defaultErrorHandler != null) {
+      FlowExceptionHandler messagingExceptionHandler = registry.<FlowExceptionHandler>lookupByName(defaultErrorHandler)
+          .orElseThrow(() -> new MuleRuntimeException(createStaticMessage(format("No global error handler defined with name '%s'.",
+                                                                                 defaultErrorHandler))));
+      if (messagingExceptionHandler instanceof MessagingExceptionHandlerAcceptor) {
+        MessagingExceptionHandlerAcceptor messagingExceptionHandlerAcceptor =
+            (MessagingExceptionHandlerAcceptor) messagingExceptionHandler;
+        if (!messagingExceptionHandlerAcceptor.acceptsAll()) {
+          throw new MuleRuntimeException(createStaticMessage("Default exception strategy must not have expression attribute. It must accept any message."));
```
marianogonzalez: this should throw an InitialisationException

PR #8001, opened 2019-06-26
MULE-17058: CXF Proxy service losing envelope data with user-defined interceptors

plbalbi:


modules/cxf/src/test/resources/log4j2-test.xml
```diff
@@ -20,6 +20,7 @@
         <!-- Mule classes -->
         <AsyncLogger name="org.mule" level="WARN"/>
         <AsyncLogger name="com.mulesoft" level="WARN"/>
```
fsgonz: Is this needed in the test?

modules/cxf/src/main/java/org/mule/module/cxf/builder/ProxyServiceMessageProcessorBuilder.java
```diff
@@ -156,4 +160,12 @@ protected String getMimeType()
         return MimeTypes.XML;
     }
 
+    @Override
+    protected void addInInterceptorCorrectingInterceptors(ServerFactoryBean sfb)
+    {
+        if (isProxy() && isProxyEnvelope() && CxfUtils.interceptorOfClassIsPresent(sfb.getInInterceptors(), SAAJInInterceptor.class))
```
fsgonz: static import

fsgonz: Is this only needed for the case of SAAJInInterceptor.

modules/cxf/src/test/resources/cxf-proxy-service-tests-config.xml
```diff
@@ -0,0 +1,34 @@
...
+        <cxf:proxy-service port="HelloServicePort"
+                           namespace="http://test.Pablo.name/"
+                           service="HelloServiceService"
+                           payload="envelope" enableMuleSoapHeaders="false"
+                           wsdlLocation="classpath:${wsdl.uri}">
+            <cxf:inInterceptors>
+                <spring:bean class="org.apache.cxf.binding.soap.saaj.SAAJInInterceptor"/>
+            </cxf:inInterceptors>
+        </cxf:proxy-service>
+        <logger message="#['Payload AFTER proxy-service: ' + message.payloadAs(java.lang.String)]"
```
fsgonz: remove loggers in test.

modules/cxf/src/test/resources/log4j2-test.xml
fsgonz: Remove loggers in test.

modules/cxf/src/main/java/org/mule/module/cxf/builder/ProxyServiceMessageProcessorBuilder.java
plbalbi: As far I looked into it, yes. Either way, the interceptor check if in the CXF's message contents there is a SOAPMessage, from which it regenerates the xml stream with the envelope

PR #8006, opened 2019-06-27
MULE-17070: Scatter-gather router inside exception handler is failing due to original exception

plbalbi:


tests/integration/src/test/resources/org/mule/test/integration/exceptions/exception-strategy-with-scopes-config.xml
```diff
@@ -0,0 +1,38 @@
...
+    </flow>
+
+    <flow name="scatter-gather-in-default-es">
+        <scripting:transformer>
+            <scripting:script engine="Groovy"><![CDATA[throw new Exception("BOOM!")]]></scripting:script>
+        </scripting:transformer>
+        <default-exception-strategy logException="false">
+            <scatter-gather>
+                <processor-chain>
+                    <logger message="I'm just a good-old logger" level="INFO"/>
```
fsgonz: We usually don't use loggers in test apps. Use a test component.

PR #8018, opened 2019-07-02
MULE-17091: Object streams not automatically closed if <non-repeatabl…

marianogonzalez:
…e-iterable> option is selected

### DIFF
--- a/core-tests/src/test/java/org/mule/runtime/core/internal/util/queue/LocalTxQueueTransactionTestCase.java
+++ b/core-tests/src/test/java/org/mule/runtime/core/internal/util/queue/LocalTxQueueTransactionTestCase.java
@@ -86,7 +86,7 @@ public class LocalTxQueueTransactionTestCase extends AbstractMuleContextTestCase
     });
     return new LocalTxQueueTransactionContext(txLog, provider, lock);
   }
-  
+
   @Override
   protected void configureMuleContext(MuleContextBuilder contextBuilder) {
     DefaultMuleConfiguration muleConfiguration = new DefaultMuleConfiguration();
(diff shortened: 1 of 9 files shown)


core/src/main/java/org/mule/runtime/core/internal/streaming/bytes/factory/NullCursorStreamProviderFactory.java
```diff
@@ -38,7 +39,7 @@ public NullCursorStreamProviderFactory(ByteBufferManager bufferManager, Streamin
    */
   @Override
   protected Object resolve(InputStream inputStream, EventContext creatorRootEventContext) {
-    streamingManager.manage(inputStream, creatorRootEventContext);
+    streamingManager.manage((Closeable) inputStream, creatorRootEventContext);
```
elrodro83: could `inputStream` not be a closeable here?

core/src/main/java/org/mule/runtime/core/api/streaming/DefaultStreamingManager.java
```diff
@@ -149,11 +150,16 @@ public CursorProvider manage(CursorProvider provider, EventContext creatorEventC
    */
   @Override
   public void manage(InputStream stream, EventContext creatorEventContext) {
-    if (stream instanceof Cursor) {
+    manage((Closeable) stream, creatorEventContext);
```
elrodro83: could `stream` not be a closeable here?

core/src/main/java/org/mule/runtime/core/internal/streaming/bytes/factory/NullCursorStreamProviderFactory.java
marianogonzalez: nope. InputStream class implements Closeable

core/src/main/java/org/mule/runtime/core/api/streaming/DefaultStreamingManager.java
marianogonzalez: same answer

core-tests/src/test/java/org/mule/runtime/core/internal/util/queue/LocalTxQueueTransactionTestCase.java
```diff
@@ -86,7 +86,7 @@ private LocalTxQueueTransactionContext createLocalTxContext(LocalTxQueueTransact
     });
     return new LocalTxQueueTransactionContext(txLog, provider, lock);
   }
-  
+
```
elrodro83: missing rebase?

PR #8023, opened 2019-07-03
MULE-17102: Composite Converters which contain the same transformer chain not filtered out.

emariotti3:


core/src/main/java/org/mule/runtime/core/privileged/transformer/CompositeConverter.java
```diff
@@ -182,4 +182,29 @@ public void setPriorityWeighting(int weighting) {}
...
+    if (this == o) {
+      return true;
+    }
+    try {
+      CompositeConverter compositeConverter = (CompositeConverter) o;
+      if (!this.getName().equals(compositeConverter.getName())) {
+        return false;
+      }
+      List<Converter> converterList = compositeConverter.getConverters();
+      if (this.chain.size() != converterList.size()) {
```
elrodro83: why use `chains` instead of `getConverters()` for `this`?

core/src/main/java/org/mule/runtime/core/privileged/transformer/CompositeConverter.java
```diff
@@ -182,4 +182,29 @@ public void setPriorityWeighting(int weighting) {}
...
+        return false;
+      }
+      for (int i = 0; i < this.chain.size(); i++) {
+        if (this.chain.get(i) != converterList.get(i) && !this.chain.get(i).equals(converterList.get(i))) {
```
elrodro83: List#equals already does this, also the size check before, no need to replicate this logic here.

core/src/main/java/org/mule/runtime/core/privileged/transformer/CompositeConverter.java
```diff
@@ -182,4 +182,29 @@ public void setPriorityWeighting(int weighting) {}
   public String toString() {
     return this.getClass().getSimpleName() + "[name: " + getName() + "; chain: " + getConverters().toString() + "]";
   }
+
+  @Override
+  public boolean equals(Object o) {
```
elrodro83: implement `hashCode` also in a consistent way.

core-tests/src/test/java/org/mule/runtime/core/privileged/transformer/CompositeConverterTestCase.java
```diff
@@ -167,4 +174,50 @@ public void appliesTransformerChainOnMessage() throws Exception {
...
+
+    int hashCodeConverterAFakeClone =
+        new CompositeConverter(byteArrayToObjectConverter2, byteArrayToInputStreamConverter2).hashCode();
+    int hashCodeConverterReverseClone =
+        new CompositeConverter(byteArrayToInputStreamConverter, byteArrayToObjectConverter).hashCode();
+    int hashCodeConverterNotAClone =
+        new CompositeConverter(byteArrayToInputStreamConverter, byteArrayToSerializableConverter).hashCode();
+
+    assertThat(Arrays.asList(hashCodeConverterAClone), everyItem(equalTo(hashCodeConverterA)));
+    assertThat(Arrays.asList(hashCodeConverterReverseClone, hashCodeConverterNotAClone, hashCodeConverterAFakeClone),
```
elrodro83: beware, hashCodes MUST be the same for equals object, but the opposite is not necessarily true.

core/src/main/java/org/mule/runtime/core/privileged/transformer/CompositeConverter.java
```diff
@@ -182,4 +183,25 @@ public void setPriorityWeighting(int weighting) {}
...
+      }
+      return this.getConverters().equals(compositeConverter.getConverters());
+    } catch (Exception e) {
+      return false;
+    }
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(this.getConverters());
```
elrodro83: hashCode must use the name, since that is also used in equals

core-tests/src/test/java/org/mule/runtime/core/privileged/transformer/CompositeConverterTestCase.java
```diff
@@ -208,16 +208,16 @@ public void hashCodeForCompositeConvertersChangesWithDifferentTransformationChai
...
+
     int hashCodeConverterReverseClone =
         new CompositeConverter(byteArrayToInputStreamConverter, byteArrayToObjectConverter).hashCode();
     int hashCodeConverterNotAClone =
         new CompositeConverter(byteArrayToInputStreamConverter, byteArrayToSerializableConverter).hashCode();
 
-    assertThat(Arrays.asList(hashCodeConverterAClone), everyItem(equalTo(hashCodeConverterA)));
-    assertThat(Arrays.asList(hashCodeConverterReverseClone, hashCodeConverterNotAClone, hashCodeConverterAFakeClone),
+    assertThat(Arrays.asList(hashCodeConverterAClone, hashCodeConverterAnotherClone), everyItem(equalTo(hashCodeConverterA)));
+    assertThat(Arrays.asList(hashCodeConverterReverseClone, hashCodeConverterNotAClone),
```
elrodro83: remove this assertion

PR #8029, opened 2019-07-05
MULE-17112: Internal libraries of a plugin are overridden by a sharedLib fron an app

elrodro83:
Also includes:

* Unify duplicated logic to explore jar files
* Fix artifact uri determination for plugin additional dependencies

### DIFF
--- a/modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/descriptor/ClassLoaderModel.java
+++ b/modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/descriptor/ClassLoaderModel.java
@@ -9,6 +9,8 @@ package org.mule.runtime.module.artifact.api.descriptor;
 
 import static java.lang.Boolean.FALSE;
 import static java.util.Arrays.asList;
+import static java.util.Collections.emptySet;
+import static java.util.Collections.unmodifiableSet;
 import static org.apache.commons.io.FilenameUtils.separatorsToUnix;
 import static org.mule.runtime.api.util.Preconditions.checkArgument;
 
@@ -27,23 +29,27 @@ public final class ClassLoaderModel {
    * Defines a {@link ClassLoaderModel} with empty configuration
    */
   public static final ClassLoaderModel NULL_CLASSLOADER_MODEL =
-      new ClassLoaderModel(new URL[0], new HashSet<>(), new HashSet<>(), new HashSet<>(), new HashSet<>(), new HashSet<>(),
-                           false);
+      new ClassLoaderModel(new URL[0], emptySet(), emptySet(), emptySet(), emptySet(), emptySet(), emptySet(), emptySet(), false);
 
   private final URL[] urls;
   private final Set<String> exportedPackages;
   private final Set<String> exportedResources;
+  private final Set<String> notImportedPackages;
+  private final Set<String> notImportedResources;
   private final Set<BundleDependency> dependencies;
   private final Set<String> privilegedExportedPackages;
   private final Set<String> privilegedArtifacts;
   private final boolean includeTestDependencies;
 
   private ClassLoaderModel(URL[] urls, Set<String> exportedPackages, Set<String> exportedResources,
+                           Set<String> notImportedPackages, Set<String> notImportedResources,
                            Set<BundleDependency> dependencies, Set<String> privilegedExportedPackages,
                            Set<String> privilegedArtifacts, boolean includeTestDependencies) {
     this.urls = urls;
     this.exportedPackages = exportedPackages;
     this.exportedResources = exportedResources;
+    this.notImportedPackages = notImportedPackages;
+    this.notImportedResources = notImportedResources;
     this.dependencies = dependencies;
     this.privilegedExportedPackages = privilegedExportedPackages;
     this.privilegedArtifacts = privilegedArtifacts;
@@ -65,10 +71,25 @@ public final class ClassLoaderModel {
   }
 
   /**
-   * @return the resource packages to be exported on the {@link ClassLoader}. Non null
+   * @return the resources to be exported on the {@link ClassLoader}. Non null
    */
   public Set<String> getExportedResources() {
     return exportedResources;
+  }
+
+  /**
+   * @return the Java packages to be loaded from the artifact itself, even if some other artifact in the region exports it. Non
+   *         null
+   */
+  public Set<String> getNotImportedPackages() {
+    return notImportedPackages;
+  }
+
+  /**
+   * @return the resources to be loaded from the artifact itself, even if some other artifact in the region exports it. Non null
+   */
+  public Set<String> getNotImportedResources() {
+    return notImportedResources;
   }
 
   /**
@@ -104,12 +125,14 @@ public final class ClassLoaderModel {
    */
   public static class ClassLoaderModelBuilder {
 
-    private Set<String> packages = new HashSet<>();
-    private Set<String> resources = new HashSet<>();
+    private final Set<String> packages = new HashSet<>();
+    private final Set<String> resources = new HashSet<>();
+    private final Set<String> notImportedPackages = new HashSet<>();
+    private final Set<String> notImportedResources = new HashSet<>();
     private List<URL> urls = new ArrayList<>();
     protected Set<BundleDependency> dependencies = new HashSet<>();
-    private Set<String> privilegedExportedPackages = new HashSet<>();
-    private Set<String> privilegedArtifacts = new HashSet<>();
+    private final Set<String> privilegedExportedPackages = new HashSet<>();
+    private final Set<String> privilegedArtifacts = new HashSet<>();
     private Boolean includeTestDependencies = FALSE;
 
     /**
@@ -127,6 +150,8 @@ public final class ClassLoaderModel {
 
       this.packages.addAll(source.exportedPackages);
       this.resources.addAll(source.exportedResources);
+      this.notImportedPackages.addAll(source.notImportedPackages);
+      this.notImportedResources.addAll(source.notImportedResources);
       this.urls = new ArrayList<>(asList(source.urls));
       this.dependencies.addAll(source.dependencies);
       this.privilegedExportedPackages.addAll(source.privilegedExportedPackages);
@@ -136,7 +161,7 @@ public final class ClassLoaderModel {
     /**
      * Indicates which package are exported on the model.
      *
-     * @param packages packages to export. No null.
+     * @param packages packages to export. Non null.
      * @return same builder instance.
      */
     public ClassLoaderModelBuilder exportingPackages(Set<String> packages) {
@@ -148,7 +173,7 @@ public final class ClassLoaderModel {
     /**
      * Indicates which resource are exported on the model.
      *
-     * @param resources resources to export. No null.
+     * @param resources resources to export. Non null.
      * @return same builder instance.
      */
     public ClassLoaderModelBuilder exportingResources(Set<String> resources) {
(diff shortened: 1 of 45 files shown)


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/AbstractMavenClassLoaderModelLoader.java
```diff
@@ -161,14 +166,42 @@ protected ClassLoaderModel createHeavyPackageClassLoaderModel(File artifactFile,
           .collect(toSet());
     }
 
-    loadUrls(artifactFile, classLoaderModelBuilder, bundleDependencies);
+    // TODO MULE-17114 retriev this data from the json if present, if not then call this
```
gsfernandes: retriev -> retrieve

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/descriptor/ClassLoaderModel.java
```diff
@@ -157,10 +182,34 @@ public ClassLoaderModelBuilder exportingResources(Set<String> resources) {
       return this;
     }
 
+    /**
+     * Indicates which Java packages are loaded from the artifact itself, even if some other artifact in the region exports it.
+     *
+     * @param packages packages to not import. Non null.
+     * @return same builder instance.
+     */
+    public ClassLoaderModelBuilder notImportingPackages(Set<String> packages) {
```
pablokraan: I would rename this to "localPackages" or similar to avoid the negative name.

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/descriptor/ClassLoaderModel.java
```diff
@@ -157,10 +182,34 @@ public ClassLoaderModelBuilder exportingResources(Set<String> resources) {
...
+    }
+
+    /**
+     * Indicates which resources are loaded from the artifact itself, even if some other artifact in the region exports it.
+     *
+     * @param resources resources to not import. Non null.
+     * @return same builder instance.
+     */
+    public ClassLoaderModelBuilder notImportingResources(Set<String> resources) {
+      checkArgument(resources != null, "resources cannot be null");
```
pablokraan: On the Java package side, you can check if the package is exported in the artifact's API and transitively determine the closure, but to review which resources can or not be consumed locally requires more code analysis.

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/descriptor/ClassLoaderModel.java
```diff
@@ -157,10 +182,34 @@ public ClassLoaderModelBuilder exportingResources(Set<String> resources) {
       return this;
     }
 
+    /**
+     * Indicates which Java packages are loaded from the artifact itself, even if some other artifact in the region exports it.
```
pablokraan: I would clarify why you can ignore the packages already provided on the region. As discussed the other day, the reason is because the API exported on the artifact does not contain theses packages, but they are contained on the artifact's jars.

modules/container/src/test/java/org/mule/runtime/container/internal/MuleClassLoaderLookupPolicyProviderTestCase.java
```diff
@@ -109,6 +110,34 @@ public void maintainsOriginalLookupStrategy() {
                hasItem(getClass().getClassLoader()));
   }
 
+  @Test
+  public void maintainsOriginalLookupStrategy2() {
```
pablokraan: maintainsOriginalLookupStrategy2 -> find a name that explain what this test does without using the 2 suffix

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/AbstractMavenClassLoaderModelLoader.java
```diff
@@ -161,14 +166,42 @@ protected ClassLoaderModel createHeavyPackageClassLoaderModel(File artifactFile,
...
+      final URI dependencyArtifactUri;
+      try {
+        dependencyArtifactUri = dependencyArtifactUrl.toURI();
+      } catch (URISyntaxException e) {
+        throw new MuleRuntimeException(e);
+      }
+
+      final JarInfo exploredJar = fileJarExplorer.explore(dependencyArtifactUri);
+
+      classLoaderModelBuilder.notImportingPackages(exploredJar.getPackages());
```
pablokraan: This is not right to me. Packages contained on the artifact cannot be simply used without more analysis.
You need to calculate the artifact's API closure (ie: all the packages that must be exported from this artifact).
Packages in the closure that are already provided by the region cannot be overridden by the artifact. Otherwise you can get ClassCastExceptions.
You can only ignore packages from the region (the parent first ones) when they are not part of your exported API.

Besides that, there are other checks to do: no artifact can override packages from java and the SYSTEM_PACKAGES (ie: org.mule.runtime, com.mulesoft.mule.runtime)

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/AbstractMavenClassLoaderModelLoader.java
```diff
@@ -161,14 +166,42 @@ protected ClassLoaderModel createHeavyPackageClassLoaderModel(File artifactFile,
...
+      classLoaderModelBuilder.notImportingResources(exploredJar.getResources());
+    }
+  }
+
+  protected JarExplorer buildJarExplorer() {
```
pablokraan: Looks like you want to mock using extension of this class. Cannot you include a JarExplorerFactory as a param instead?

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/ApplicationDeploymentTestCase.java
```diff
@@ -1284,24 +1285,50 @@ public void deploysAppWithPluginExportingAlreadyProvidedAppPackage() throws Exce
...
+    startDeployment();
+
+    assertDeploymentSuccess(applicationDeploymentListener, differentLibPluginAppFileBuilder.getId());
+
+    executeApplicationFlow("main");
+  }
+
+  @Test
+  @Description("Ensure that when a plugin2 depends on plugin1, when using something exported from plugin1, that is used")
+  public void pluginDependingAndExportingFromOtherPlugin() throws Exception {
```
pablokraan: Add a test to ensure that, when an artifact contains package P1, which is also exported on the region, and that P1 is part on the artifact exported API, then the artifact uses P1 form the region.

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/DomainDeploymentTestCase.java
```diff
@@ -205,27 +208,34 @@ public void deploysDomainWithSharedLibPrecedenceOverApplicationLib() throws Exce
...
-        new ArtifactPluginFileBuilder("echoPlugin1").configuredWith(EXPORTED_CLASS_PACKAGES_PROPERTY, "org.foo")
-            .containingClass(pluginEcho1TestClassFile, "org/foo/Plugin1Echo.class")
-            .dependingOn(new JarFileBuilder("barUtils2_0", barUtils2_0JarFile));
-
-    final DomainFileBuilder domainFileBuilder =
-        new DomainFileBuilder(domainId).dependingOnSharedLibrary(new JarFileBuilder("barUtils1.0", barUtils1_0JarFile))
-            .definedBy("empty-domain-config.xml");
+  @Issue("MULE-17112")
+  @Description("If a plugin uses a library and the domain sets another version of that library as a sharedLib, the plugin internally uses its own version of the lib and not the domain's.")
+  public void pluginWithDependencyAndConflictingVersionSharedByApp() throws Exception {
```
pablokraan: Add a another test similar to the one requested in the app tests

modules/deployment/src/test/resources/log4j2-test.xml
```diff
@@ -20,6 +20,7 @@
         <!-- Mule classes -->
         <AsyncLogger name="org.mule" level="WARN"/>
         <AsyncLogger name="com.mulesoft" level="WARN"/>
+<!--         <AsyncLogger name="org.mule.runtime.module.artifact" level="DEBUG"/> -->
```
pablokraan: remove/uncomment this

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/AbstractMavenClassLoaderModelLoader.java
elrodro83: The wording of this comment reminded me of 
image

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/AbstractMavenClassLoaderModelLoader.java
```diff
@@ -161,14 +176,43 @@ protected ClassLoaderModel createHeavyPackageClassLoaderModel(File artifactFile,
...
+      try {
+        dependencyArtifactUri = dependencyArtifactUrl.toURI();
+      } catch (URISyntaxException e) {
+        throw new MuleRuntimeException(e);
+      }
+
+      final JarInfo exploredJar = jarExplorerFactory.get().explore(dependencyArtifactUri);
+
+      final Set<String> localPackages = new HashSet<>(exploredJar.getPackages());
+      localPackages.removeAll(exportedPackages);
```
gsfernandes: I think that this will only cover one part of what @pablokraan has suggested to change here (original comment was https://github.com/mulesoft/mule/pull/8029#discussion_r301590317):
**You can only ignore packages from the region (the parent first ones) when they are not part of your exported API.** 
I think that system packages have to be ignored here too: **SYSTEM_PACKAGES (ie: org.mule.runtime, com.mulesoft.mule.runtime)** but the part that I'm having hard time to understand is:
**Packages in the closure that are already provided by the region cannot be overridden by the artifact. Otherwise you can get ClassCastExceptions.**. 
I understand this but at the same time this is exactly what we need to do here, in order to avoid loading spring packages from the sharedLibraries (API from owner classloader in region) and use the spring packages from the internal library bundled within the plugin. So, how does this work with the general rule from Kraan's comment?

gsfernandes: I keep thinking about this and perhaps this is already covered, let say that you have artifact A that exports packageA and packageB as it depends on artifact B, so artifact B exports only packageB. At the same time what you are doing is mark as local package a packageC which comes from introspecting the libraries in artifact A (it depends on a library C).
Now, how about this case, you also have an artifact C in the region (another plugin) which exports a package C. In that case, as artifact A from its API is not requiring to "import" the package C, it will be using its local classes from its own classloader assuming that you will never have to access to those classes from artifact C. right?
If that's the case, I think these changes are fine and you could just dismiss my previous comment.

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/AbstractMavenClassLoaderModelLoader.java
```diff
@@ -161,14 +176,43 @@ protected ClassLoaderModel createHeavyPackageClassLoaderModel(File artifactFile,
           .collect(toSet());
     }
 
-    loadUrls(artifactFile, classLoaderModelBuilder, bundleDependencies);
+    // TODO MULE-17114 retrieve this data from the json if present, if not then call this
+    populateLocalPackages(artifactFile, classLoaderModelBuilder, bundleDependencies, exportedPackages, exportedResources);
```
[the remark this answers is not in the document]
elrodro83: yes, missed that one, will add it.

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/maven/AbstractMavenClassLoaderModelLoader.java
elrodro83: yes, from what we talked with @pablokraan there are some caes where a plugin may be handling objects created by another plugin, and of both plugins have the same local class a `ClassCastException` may occur. I think these cases would be very rare, since there would be a common interface in the region for handling those, but that is not being enforced. We should discuss how those kind of static validations may be enforced in the packager.

PR #8046, opened 2019-07-15
MULE-17110: NTLM authentication with dynamic credentials is successful despite credentials changes

plbalbi:


modules/http-api/src/main/java/org/mule/runtime/http/api/client/auth/HttpAuthentication.java
```diff
@@ -91,6 +91,18 @@ static HttpNtlmAuthenticationBuilder ntlm(String username, String password) {
    */
   boolean isPreemptive();
 
+  /**
+   * Returns true if any of the provided credentials for authentications is an expression, meaning it could vary between different
```
afelisatti: Since this is the API and there's no expression logic here, I'd just say it will force the connection to close. You can add something like "useful with dynamic credentials to guarantee a clean slate on each request" but not expressions which don't exist on this layer.

modules/http-api/src/main/java/org/mule/runtime/http/api/client/auth/DefaultHttpAuthentication.java
```diff
@@ -21,19 +21,21 @@
...
   private final boolean preemptive;
+  private boolean forceConnectionClose;
 
   private final String domain;
   private final String workstation;
 
-  DefaultHttpAuthentication(HttpAuthenticationType type, String username, String password, boolean preemptive, String domain,
-                            String workstation) {
+  DefaultHttpAuthentication(HttpAuthenticationType type, String username, String password, boolean preemptive,
+                            boolean doCredentialsVary, String domain, String workstation) {
```
afelisatti: `forceConnectionClose`

modules/http-api/src/main/java/org/mule/runtime/http/api/client/auth/HttpAuthentication.java
```diff
@@ -91,6 +91,18 @@ static HttpNtlmAuthenticationBuilder ntlm(String username, String password) {
...
+   * evaluations.
+   *
+   * @return whether the authentication credentials may vary.
+   *
+   * @since 4.3
+   */
+  default boolean forceConnectionClose() {
```
afelisatti: Why is this required at the general auth level instead of just NTLM?

plbalbi: Maybe it should go in the NTLMAuth. The http-connector implementation of this method doesn't go directly into the NtlmAuthentication object, but in one called UsernamePasswordAuthentication. But I think it also works if I move to NTLM. Will try.

plbalbi: Just moved it.

modules/http-api/src/main/java/org/mule/runtime/http/api/client/auth/HttpAuthentication.java
```diff
@@ -121,5 +121,16 @@ static HttpNtlmAuthenticationBuilder builder() {
      */
     String getWorkstation();
 
+    /**
+     * Whether or not the Connection close header should be enforced after the authentication. Useful with the auth credentials
```
afelisatti: `Useful when`

PR #8053, opened 2019-07-17
MULE-17168: Remove softreferences for app classloaders to avoid OOM on redeploy.

fsgonz:


modules/artifact/src/main/java/org/mule/module/artifact/classloader/ThreadGroupContextClassLoaderSoftReferenceBuster.java
```diff
@@ -4,28 +4,32 @@
...
+
 /**
  * Reference utilities to bust Thread Group Context soft references.
  * 
  * @since 4.3.0
  */
 public class ThreadGroupContextClassLoaderSoftReferenceBuster {
 
-  private static final Logger LOGGER = LoggerFactory.getLogger(ThreadGroupContextClassLoaderSoftReferenceBuster.class);
+  private final transient static Logger logger = LoggerFactory.getLogger(ThreadGroupContextClassLoaderSoftReferenceBuster.class);
```
elrodro83: why?

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/MuleArtifactClassLoader.java
```diff
@@ -322,6 +326,8 @@ private void shutdownListeners() {
    */
   protected ResourceReleaser createResourceReleaserInstance() {
     if (resourceReleaserInstance == null) {
+      createClass(threadGroupContextClassLoaderSoftReferenceBusterLocation);
```
elrodro83: why is this here? nobody uses the generated class afterwards

fsgonz: Because it is used internally by the resourceReleaser.

modules/artifact/src/main/java/org/mule/module/artifact/classloader/ThreadGroupContextClassLoaderSoftReferenceBuster.java
```diff
@@ -4,28 +4,32 @@
...
+
 /**
  * Reference utilities to bust Thread Group Context soft references.
  * 
  * @since 4.3.0
  */
 public class ThreadGroupContextClassLoaderSoftReferenceBuster {
 
-  private static final Logger LOGGER = LoggerFactory.getLogger(ThreadGroupContextClassLoaderSoftReferenceBuster.class);
+  private final static Logger logger = LoggerFactory.getLogger(ThreadGroupContextClassLoaderSoftReferenceBuster.class);
```
elrodro83: uppercase LOGGER

PR #8054, opened 2019-07-17
MULE-17217: ExpressionBasedParameterResolvers should use the resolved value as caching key

plbalbi:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ExpressionBasedParameterResolver.java
```diff
@@ -104,4 +104,18 @@ public void setMelDefault(Boolean melDefault) {
   public void setMelAvailable(Boolean melAvailable) {
     this.melAvailable = melAvailable;
   }
+
+  @Override
+  public int hashCode() {
+    return this.resolve().hashCode();
```
marianogonzalez: only the extension can call resolve(). SDK should never call that method on its own

PR #8057, opened 2019-07-18
MULE-17218: Mule SDK is leaking extensions configurations over the time

ndinu:


ndinu:
--validate

ndinu:
> refactor the test as discussed

I have refactored the way to test this in a different way as discussed, let me know if you find it suitable.

PR #8062, opened 2019-07-19
MULE-17210: Too many ephemeral events are created just to set a variable

marianogonzalez:


core/src/main/java/org/mule/runtime/core/api/util/StreamingUtils.java
```diff
@@ -294,33 +294,40 @@ private static void handlePossibleException(Reference<Throwable> exception) thro
...
    *
    * @param value            the typed value to update
-   * @param event            the current event
+   * @param eventContext     the current event context
    * @param streamingManager the streaming manager
    * @return updated {@link TypedValue instance}
    */
-  public static TypedValue updateTypedValueForStreaming(final TypedValue value, final CoreEvent event,
+  public static TypedValue updateTypedValueForStreaming(final TypedValue value,
+                                                        final EventContext eventContext,
```
elrodro83: rename to `rootEventContext`

PR #8068, opened 2019-07-23
MULE-17062: Avoid thread switches from cpuL to cpuL in ProcessingStrategy

JuLuzzi:


core/src/main/java/org/mule/runtime/core/internal/processor/strategy/AbstractProcessingStrategy.java
```diff
@@ -55,7 +53,7 @@ public Sink createSink(FlowConstruct flowConstruct, ReactiveProcessor pipeline)
     };
   }
 
-  protected ExecutorService decorateScheduler(Scheduler scheduler) {
+  protected ScheduledExecutorService decorateScheduler(ScheduledExecutorService scheduler) {
```
elrodro83: can the param and return type be `ExecutorService` (not `Scheduled*`)? Or does that change break something?

core/src/main/java/org/mule/runtime/core/internal/util/rx/ConditionalExecutorServiceDecorator.java
```diff
@@ -34,8 +32,8 @@
  */
 public class ConditionalExecutorServiceDecorator implements ScheduledExecutorService {
```
elrodro83: can this be a `ExecutorService` ?

tests/unit/pom.xml
```diff
@@ -55,6 +55,12 @@
             <artifactId>mule-tests-model</artifactId>
             <version>${project.version}</version>
         </dependency>
+
+        <dependency>
+            <groupId>org.mule.services</groupId>
+            <artifactId>mule-service-scheduler</artifactId>
```
elrodro83: remove this dependency. Using this would make the test be less unit and more integration

PR #8102, opened 2019-08-08
MULE-17293: Stereotypes enrichment is done before all types are (#8099)

elrodro83:


ndinu:
Change name from CONSISTENCY phase according to how it will change in mule-extensions-api

PR #8112, opened 2019-08-12
MULE-17311: Services started by an app plugin retain app classloader after reploy/undeploy

fsgonz:


fsgonz:
Sorry. I'll try to add a test.

PR #8116, opened 2019-08-13
 MULE-17025: perfomance improvement

mbuchwald:


core/src/main/java/org/mule/runtime/core/privileged/exception/TemplateOnErrorHandler.java
```diff
@@ -205,6 +208,14 @@ protected void doInitialise(MuleContext muleContext) throws InitialisationExcept
     }
 
     errorTypeMatcher = createErrorType(muleContext.getErrorTypeRepository(), errorType, configurationProperties);
+    if (this.getLocation() != null) {
```
marianogonzalez: this.getLocation() is expensive. execute only once and store in var

PR #8117, opened 2019-08-13
MULE-17307: MetadataTypes from the extension model have different instances representing the same type

elrodro83:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/ParameterDslDeclarationEnricher.java
```diff
@@ -65,7 +75,32 @@ protected void onParameter(ParameterGroupDeclaration parameterGroup, ParameterDe
...
+                      type.getAnnotations().stream()
+                          .collect(toMap(ann -> ann.getClass(), Functions.identity()));
+
+                  declaration.getType().getAnnotation(ClassInformationAnnotation.class)
+                      .ifPresent(paramClassInfo -> normalizedAnnotationsByClass.put(ClassInformationAnnotation.class,
+                                                                                    paramClassInfo));
+                  declaration.getType().getAnnotation(ParameterDslAnnotation.class)
+                      .ifPresent(paramClassInfo -> normalizedAnnotationsByClass.put(ParameterDslAnnotation.class,
+                                                                                    paramClassInfo));
+                  return new DefaultObjectType(((ObjectType) type).getFields(), ((ObjectType) type).isOrdered(),
```
marianogonzalez: what happens if the type is a dynamic one and the ClassInformation annotation simply points to `java.lang.Object`

PR #8125, opened 2019-08-15
MULE-17344: Connections associated with InputStreams should be released when stream is fully consumed

ndinu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/AbstractReturnDelegate.java
```diff
@@ -235,63 +237,23 @@ protected MediaType resolveMediaType(Object value, Optional<MediaType> contextMi
     return contextMimeType.orElse(mediaType).withCharset(contextEncoding.orElse(existingEncoding));
   }
 
-  protected class ConnectedInputStreamWrapper extends InputStream {
+  protected class ConnectedInputStreamWrapper extends AutoCloseInputStream {
```
marianogonzalez: add a unit test that verifies that the connection handler is closed when this stream is fully consumed

PR #8129, opened 2019-08-16
MULE-17346: Move ParameterDslDeclarationEnricher to extensions-api

elrodro83:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/loader/AbstractJavaExtensionModelLoader.java
```diff
@@ -140,11 +139,10 @@
                                                                                                new ParameterLayoutOrderDeclarationEnricher(),
                                                                                                new BackPressureDeclarationEnricher(),
                                                                                                new ObjectStoreParameterDeclarationEnricher(),
-                                                                                               new PollingSourceDeclarationEnricher(),
```
marianogonzalez: why did you remove the polling source one as well?

PR #8142, opened 2019-08-22
MULE-17374. Add a way to track who release the streams

machaval:


core/src/main/java/org/mule/runtime/core/internal/streaming/bytes/AbstractCursorStream.java
```diff
@@ -20,10 +22,16 @@
  */
 abstract class AbstractCursorStream extends CursorStream {
 
+  public static final String TRACK_CURSOR_RELEASED = SYSTEM_PROPERTY_PREFIX + "track.releaser";
```
marianogonzalez: private

core/src/main/java/org/mule/runtime/core/internal/streaming/bytes/AbstractCursorStream.java
```diff
@@ -20,10 +22,16 @@
...
+
+  public static Boolean TRACK_RELEASER = Boolean.getBoolean(TRACK_CURSOR_RELEASED);
```
marianogonzalez: private and final

PR #8153, opened 2019-08-28
Fix Object Store corrupted file move

r-murphy:
Hi There.
Sorry, there is no JIRA ticket. But at the moment it seems impossible to create new JIRA accounts, as the link to create an account always redirects to Anypoint, and my Anypoint login does not work for JIRA.

If someone can create a JIRA ticket for this, I'd be happy to recreate the branch/commit/PR.

The summary of this issue is that the code to move corrupted files from the object store folder was not behaving correctly. It was not creating the required folder tree first before moving.

The existing test was passing but was not correct. The corrupted file was being moved in the test but was being renamed to have the object store name, rather than being nested under a folder having the object store name. The test has been fixed to expect the correct nesting.

Expected (and new result):
`<working_dir>/corrupted-files/junit4867812718450648275/temp1676195331268480483.obj`, where `junit4867812718450648275` is the object store name and `temp1676195331268480483.obj` is the corrupted file.

Previous result (incorrect, but was passing):
`<working_dir>/corrupted-files/junit4867812718450648275`, where `junit4867812718450648275` is the corrupted file.

### DIFF
--- a/core-tests/src/test/java/org/mule/runtime/core/internal/util/store/PersistentObjectStorePartitionTestCase.java
+++ b/core-tests/src/test/java/org/mule/runtime/core/internal/util/store/PersistentObjectStorePartitionTestCase.java
@@ -13,7 +13,6 @@ import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 import static org.mockito.Mockito.when;
-import static org.mule.runtime.core.api.util.FileUtils.openDirectory;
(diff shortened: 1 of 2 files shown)


r-murphy:
Hi @aiannucci can you let me know the reason this was closed? 
Thanks

r-murphy:
Hi @aiannucci can you let me know the reason this was closed?
Thanks

aiannucci:
Hi, this was closed automatically when the base branch was deleted. If it still makes sense please reopen it against the master branch

PR #8165, opened 2019-09-02
MULE-17356: Domain deployment scenarios

eze210:


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/domain/AmbiguousDomainReferenceException.java
```diff
@@ -9,14 +9,11 @@
...
- * Exception type for representing a failure due to domain not found.
- */
-public final class IncompatibleDomainVersionException extends MuleException {
+public class AmbiguousDomainReferenceException extends MuleException {
 
-  public IncompatibleDomainVersionException(String expectedDomainName, String availableVersion) {
-    super(createStaticMessage(format("Expected domain '%s' couldn't be retrieved. It is available the '%s' version",
-                                     expectedDomainName, availableVersion)));
+  public AmbiguousDomainReferenceException(BundleDescriptor bundleDescriptor) {
+    super(createStaticMessage(format("More than one compatible domain were found for bundle descriptor %s", bundleDescriptor)));
```
gsfernandes: Would it be possible to list those ambiguous domains found?

gsfernandes: As part of the error message.

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/domain/DefaultDomainManager.java
```diff
@@ -21,65 +20,78 @@
...
-    DomainDescriptor domainDescriptor = domain.getDescriptor();
-    BundleDescriptor bundleDescriptor = domainDescriptor.getBundleDescriptor();
-    if (bundleDescriptor == null) {
-      return domainDescriptor.getName();
-    } else {
-      return bundleDescriptor.getArtifactFileName();
+  @Override
+  public boolean containsCompatible(BundleDescriptor descriptor) {
+    for (Domain domain : domainsByName.values()) {
+      if (isCompatibleBundle(domain.getDescriptor().getBundleDescriptor(), descriptor)) {
```
gsfernandes: I don't remember if the current application's logs are showing to which domain the application has been "linked" after it is deployed. Maybe we could show that if it is not present. Right?

eze210: There is a "table" in the logs with application name, domain name, and status (deployed, failed...)

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/domain/AmbiguousDomainReferenceException.java
eze210: I added the domain names to the error message

PR #8168, opened 2019-09-02
MULE-16943: SDK operations optimizations: Avoid non-blocking overhead for known blocking calls

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
```diff
@@ -483,14 +507,18 @@ protected boolean usesDynamicConfiguration() {
    * returns an empty value.
    */
   protected Optional<ConfigurationInstance> getStaticConfiguration() {
-    if (!requiresConfig.get() || (isConfigurationSpecified() && configurationProvider.get().isDynamic())) {
+    if (!requiresConfig()) {
```
elrodro83: isn't this changing the behavior?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultExecutionMediator.java
```diff
@@ -167,15 +167,15 @@ public void error(Throwable e) {
...
+    executeCommand.accept(callbackDecorator);
   }
 
   private void executeWithRetry(ExecutionContextAdapter<M> context,
                                 RetryPolicyTemplate retryPolicy,
                                 List<Interceptor> executedInterceptors,
                                 Consumer<ExecutorCallback> executeCommand,
-                                BiConsumer<ExecutorCallback, Object> onSuccess,
-                                BiConsumer<ExecutorCallback, Throwable> onError,
+                                //BiConsumer<ExecutorCallback, Object> onSuccess,
```
elrodro83: remove

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OAuthOperationMessageProcessor.java
```diff
@@ -72,34 +70,33 @@ public OAuthOperationMessageProcessor(ExtensionModel extensionModel,
...
 
         OAuthConnectionProviderWrapper connectionProvider = getOAuthConnectionProvider(operationContext);
         if (connectionProvider == null) {
-          delegate.error(e);
+          callback.error(e);
           return;
         }
 
-        Reference<Optional<String>> resourceOwnerIdReference = new Reference<>(empty());
+        Reference<Optional<String>> resourceOwnerIdReference = new Reference<>(Optional.empty());
```
elrodro83: static import

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
marianogonzalez: nope, not at all. On the contrary. this is not the first version of this optimization. This is the one that fixed all the tests that I broke :p

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OAuthOperationMessageProcessor.java
marianogonzalez: can't. it conflicts with other imports

PR #8170, opened 2019-09-02
MULE-17401: Scheduler source should route emitted message through FlowProcessMediator

plbalbi:


core/src/main/java/org/mule/runtime/core/internal/source/scheduler/DefaultSchedulerMessageSource.java
```diff
@@ -157,37 +162,30 @@ private void poll() {
...
-          .doOnError(MessagingException.class,
-                     me -> ((BaseEventContext) me.getEvent().getContext()).error(me))
-          .doOnSuccess(result -> ((BaseEventContext) result.getContext()).success())
-          .doFinally(s -> {
-            synchronized (DefaultSchedulerMessageSource.this) {
-              executing = false;
-            }
-          })
-          .subscribe(requestUnbounded());
+      setCurrentEvent(InternalEvent.builder(EventContextFactory.create(flowConstruct, getLocation())).message(Message.of(null))
```
elrodro83: review if this is needed or already done by the ModuleFlowProcessingPhase

core/src/main/java/org/mule/runtime/core/internal/source/scheduler/SchedulerFlowProcessingTemplate.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @Override
+  public Publisher<CoreEvent> routeEventAsync(Publisher<CoreEvent> eventPub) {
+    return from(eventPub).transform(messageProcessor);
+  }
+
+  @Override
+  public Publisher<Void> sendResponseToClient(CoreEvent response, Map<String, Object> parameters) {
+    return Mono.empty();
```
elrodro83: static import

core/src/main/java/org/mule/runtime/core/internal/source/scheduler/SchedulerFlowProcessingTemplate.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @Override
+  public Publisher<Void> sendFailureResponseToClient(MessagingException exception, Map<String, Object> parameters) {
+    return Mono.empty();
```
elrodro83: static import

core/src/main/java/org/mule/runtime/core/internal/source/scheduler/DefaultSchedulerMessageSource.java
plbalbi: It's not done in ModuleFlowProcessingPhase. The current event is set in the interceptors applied to each processor. In my opinion is needed here.

PR #8184, opened 2019-09-04
MULE-17407: Async is defining null ReplyToHandler instead of void/empty one

mbuchwald:


core/src/main/java/org/mule/execution/MessageProcessorNotificationExecutionInterceptor.java
```diff
@@ -59,7 +60,7 @@ public MuleEvent execute(final MessageProcessor messageProcessor, final MuleEven
         MuleEvent result = null;
         MessagingException exceptionThrown = null;
 
-        boolean nonBlocking = event.isAllowNonBlocking() && event.getReplyToHandler() != null;
+        boolean nonBlocking = event.isAllowNonBlocking() && (event.getReplyToHandler() != null && !(event.getReplyToHandler() instanceof AsyncReplyToHandler));
```
fsgonz: I think you can remove the outer parenthesis.

core/src/main/java/org/mule/processor/AsyncReplyToHandler.java
```diff
@@ -0,0 +1,26 @@
...
+import org.mule.api.MessagingException;
+import org.mule.api.MuleEvent;
+import org.mule.api.MuleException;
+import org.mule.api.MuleMessage;
+import org.mule.api.transport.ReplyToHandler;
+
+public class AsyncReplyToHandler implements ReplyToHandler {
+
+    @Override
+    public void processReplyTo(MuleEvent event, MuleMessage returnMessage, Object replyTo) throws MuleException {
```
fsgonz: formatter

core/src/main/java/org/mule/processor/AsyncReplyToHandler.java
```diff
@@ -0,0 +1,26 @@
...
+
+    }
+
+    @Override
```
fsgonz: formatter

PR #8189, opened 2019-09-05
MULE-17421: RoundRobin implementation is blocking

elrodro83:
Also fix race condition in the counter handling

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/routing/AbstractSinkRouter.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/routing/AbstractSinkRouter.java
@@ -11,7 +11,6 @@ import static org.mule.runtime.core.api.rx.Exceptions.checkedConsumer;
 import static org.mule.runtime.core.internal.util.rx.RxUtils.subscribeFluxOnPublisherSubscription;
 import static reactor.core.publisher.Flux.from;
 
-import org.mule.runtime.core.api.el.ExpressionManagerSession;
 import org.mule.runtime.core.api.event.CoreEvent;
 import org.mule.runtime.core.privileged.routing.RoutingException;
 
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/mule/runtime/core/internal/routing/RoundRobin.java
```diff
@@ -6,39 +6,156 @@
...
-   * Process the event using the next target route in sequence
+   * Router that generates executable routes for each one configured and handles the logic of deciding which one will be executed
+   * when an event arrives. It also provides access to the routes publishers so they can be merged together.
    */
-  @Override
-  public CoreEvent route(CoreEvent event) throws MuleException {
-    int modulo = getAndIncrementModuloN(routes.size());
-    if (modulo < 0) {
-      throw new CouldNotRouteOutboundMessageException(this);
+  private class SinkRouter {
```
afelisatti: Why not crate an abstract one shared with choice?

afelisatti: I see nothing different except the route method.

core/src/main/java/org/mule/runtime/core/internal/routing/AbstractSinkRouter.java
```diff
@@ -0,0 +1,74 @@
...
+        // Since there's always at least one route, the default one, one route will always be decorated.
+        routes.add(subscribeFluxOnPublisherSubscription(nextRoute.getPublisher(), router));
+      }
+    }
+    return routes;
+  }
+
+  /**
+   * Decides which route should execute for an incoming event and executes it, purposely separating those actions so that a single
+   * {@link ExpressionManagerSession} can be used for all routes decision process.
```
afelisatti: Update

PR #8219, opened 2019-09-13
MULE-17431 Make FlowConstruct in ExtensionMessageSource a LazyValue

diegostrubolini:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -139,7 +138,8 @@
   private SourceAdapter sourceAdapter;
   private RetryPolicyTemplate retryPolicyTemplate;
   private Scheduler retryScheduler;
-  private FlowConstruct flowConstruct;
+  // FlowConstruct is obtained when needed because during MUnit's tooling tests this should never be evaluated.
```
marianogonzalez: Munit and Lazy init mode

PR #8229, opened 2019-09-17
MULE-17420: XML SDK: link with the inner models on the xml sdk component

elrodro83:
definition


modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/loader/XmlExtensionLoaderDelegate.java
```diff
@@ -714,47 +681,29 @@ private void addConnectionProvider(ConfigurationDeclarer configurationDeclarer,
...
-                }
-              }
-            }
-          }
-        }
-      }
-    }
-    if (testConnectionComponentModels.size() > 1) {
+                                                                       ExtensionModelHelper extensionModelHelper) {
+    final List<ComponentModel> testConnectionComponentModels_new = globalElementsComponentModel
```
fernandezlautaro: I'd drop the `_new` suffix

PR #8230, opened 2019-09-17
MULE-17466: Maps should not be required to have getters

marianogonzalez:


modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/validation/InputParameterTypesValidatorTestCase.java
```diff
@@ -93,6 +98,14 @@ public void invalidModelDueToOperationWithArgumentParameterWithoutGetter() {
   public void validModelDueToOperationWithArgumentParameterWithGetter() {
     validateOperationParameterOfType(toMetadataType(PojoWithParameterWithGetter.class));
   }
+  @Test
```
elrodro83: empty line before

PR #8234, opened 2019-09-18
MULE-16491: Pattern matching for error types in error handlers' type attribute

eze210:


core/src/main/java/org/mule/runtime/core/api/exception/WildcardErrorTypeMatcher.java
```diff
@@ -0,0 +1,62 @@
...
+  @Override
+  public boolean match(ErrorType errorType) {
+    if (matchIdentifier(errorType) && matchNamespace(errorType) && matchParent(errorType)) {
+      return true;
+    }
+
+    return isChild(errorType);
+  }
+
+  private boolean matchParent(ErrorType errorType) {
```
afelisatti: How do you know who the parent is from something like "HTTP:*" or "*:CONNECTIVITY"?

core-tests/src/test/java/org/mule/runtime/core/privileged/exception/WildcardErrorTypeMatcherTestCase.java
```diff
@@ -0,0 +1,180 @@
...
+  @Test
+  public void doesNotMatchSibling() {
+    ErrorTypeMatcher transformationMatcher = new WildcardErrorTypeMatcher(transformationErrorType);
+
+    assertThat(transformationMatcher.match(expressionErrorType), is(false));
+  }
+
+  @Test
+  public void matchWildcardIdentifier() {
+    ErrorType wildcardErrorType = ErrorTypeBuilder.builder().namespace(CORE_NAMESPACE_NAME).identifier(WILDCARD_TOKEN)
```
afelisatti: This mixes the matching and the typing together. Error types represent actual errors that can occur, they are collected and restricted, guaranteeing you cannot throw an error you haven't declared or catch a non existent error.

core-tests/src/test/java/org/mule/runtime/core/privileged/exception/WildcardErrorTypeMatcherTestCase.java
```diff
@@ -0,0 +1,168 @@
...
+    ErrorType test2ErrorType = ErrorTypeBuilder.builder().namespace(CORE_NAMESPACE_NAME).identifier("test2")
+        .parentErrorType(transformationErrorType).build();
+
+    ErrorTypeMatcher wildcardMatcher = new WildcardErrorTypeMatcher(wildcardErrorType);
+    assertThat(wildcardMatcher.match(test1ErrorType), is(true));
+    assertThat(wildcardMatcher.match(test2ErrorType), is(true));
+  }
+
+  @Test
+  public void wildcardErrorTypeMatchesWithParent() {
```
afelisatti: This is the one we said should go.

afelisatti:
Make sure to track the pending documentation of this on another ticket and make it a blocker for 4.3.0.

PR #8246, opened 2019-09-23
MULE-17415: Polling Sources fail to pick the same item multiple times

marianogonzalez:


core-tests/src/test/java/org/mule/runtime/core/ComposeTest.java
```diff
@@ -0,0 +1,30 @@
...
+package org.mule.runtime.core;
+
+import org.junit.Test;
+import reactor.core.publisher.Mono;
+
+public class ComposeTest {
+
+
+  @Test
+  public void test() {
```
elrodro83: remove sysouts, add assertions

PR #8253, opened 2019-09-27
MULE-17198: Replace lazyInit with the AST dependency graph

elrodro83:


core/src/main/java/org/mule/runtime/core/internal/source/scheduler/DefaultSchedulerMessageSource.java
```diff
@@ -197,7 +197,8 @@ protected void setIsExecuting(boolean value) {
    */
   @Override
   public void initialise() throws InitialisationException {
-    this.flowConstruct = getFromAnnotatedObjectOrFail(muleContext.getConfigurationComponentLocator(), this);
+    getFromAnnotatedObject(componentLocator, this)
```
gsfernandes: What about the "orFail" that was before there? Shouldn't this fail if not present?

elrodro83: no, because that is a valid scenario now with the lazy Init changes. Moreover, this case was not possible on previous versions, so the check was superfluous

PR #8261, opened 2019-10-01
MULE-17521: Avoid using an `AutoCloseInputStream` for streams

elrodro83:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/AbstractReturnDelegate.java
```diff
@@ -246,10 +248,25 @@ private ConnectedInputStreamWrapper(InputStream delegate, ConnectionHandler<?> c
       this.connectionHandler = connectionHandler;
     }
 
+    /**
+     * Automatically closes the stream if the end of stream was reached.
+     *
+     * @param n number of bytes read, or -1 if no more bytes are available
+     * @throws IOException if the stream could not be closed
+     * @since 2.0
```
ndinu: remove since anotation

PR #8274, opened 2019-10-05
MULE-17114: Improve tests and adding more coverage for localPackages …

gsfernandes:
…and localResources loaded from class loader model


modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/artifact/DeployableArtifactDescriptorFactoryTestCase.java
```diff
@@ -372,6 +373,13 @@ private void assertClassLoaderModelWithPluginDependencyAndAdditionalDependencies
     // urls
     assertThat(testEmptyPluginDescriptor.getClassLoaderModel().getDependencies(), hasSize(0));
 
+    assertThat(testEmptyPluginDescriptor.getClassLoaderModel().getLocalPackages(), IsCollectionWithSize.hasSize(19));
```
elrodro83: static import

PR #8276, opened 2019-10-07
MULE-17536: FunctionalReadWriteLock: Provide withReadLock methods

elrodro83:
without the releaser


tests/performance/src/main/java/org/mule/runtime/core/concurrent/FunctionalReadWriteLockBenchmark.java
```diff
@@ -0,0 +1,46 @@
...
+    return rwLock.withReadLock(releaser -> {
+      return Math.random();
+    });
+  }
+
+  @Benchmark
+  @Threads(Threads.MAX)
+  public double withReadLock() {
+    return rwLock.withReadLock(() -> {
+      return Math.random();
```
eze210: static import

PR #8277, opened 2019-10-07
MULE-17535: Performance improvements for MultiMap immutability

elrodro83:
* Optimize cloning for immutability
* Add support for unmodifiable wrapping


core/src/main/java/org/mule/runtime/core/api/util/CaseInsensitiveHashMap.java
```diff
@@ -172,6 +172,9 @@ public String toString() {
     if (this instanceof ImmutableCaseInsensitiveHashMap) {
```
eze210: Can this check be avoided using polymorphism?

PR #8278, opened 2019-10-08
MULE-17398: Operations that return a PagingProvider do not populate the type catalog properly

SebaElizalde:


SebaElizalde:
--validate

PR #8288, opened 2019-10-10
MULE-17593: Add a plugins blacklist for MULE-17112 fix

eze210:


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/plugin/PluginLocalDependenciesBlacklist.java
```diff
@@ -0,0 +1,104 @@
...
+
+    if (!doArtifactIdsMatch(blacklistedDescriptor, pluginBundleDescriptor)) {
+      return false;
+    }
+
+    return isBlacklistedVersionGreaterOrEqual(blacklistedDescriptor.getVersion(), pluginBundleDescriptor.getVersion());
+  }
+
+  private static boolean isBlacklistedVersionGreaterOrEqual(String blacklistedVersion, String pluginVersion) {
+    Semver blacklistedSemver = new Semver(blacklistedVersion);
```
elrodro83: can you use the methods from semver here to compare versions?

eze210: yep

PR #8303, opened 2019-10-15
MULE-17560: DefaultPersistentMetadataCacheManager is created per appl…

gsfernandes:
…ication contexto therefore the shared OS partitioned fails with concurrency

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/config/RuntimeLockFactoryUtil.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/config/RuntimeLockFactoryUtil.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.internal.config;
+
+import org.mule.runtime.api.lock.LockFactory;
+import org.mule.runtime.core.internal.lock.ServerLockFactory;
+
+/**
+ * Utility method for getting the runtime {@link LockFactory}.
+ */
+public class RuntimeLockFactoryUtil {
+
+  private static LockFactory lockFactory;
+
+  static {
+    lockFactory = new ServerLockFactory();
+  }
+
+  public static LockFactory getRuntimeLockFactory() {
+    return lockFactory;
+  }
+
+}
--- a/core/src/main/java/org/mule/runtime/core/internal/config/RuntimeLockFactoryUtil.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/config/RuntimeLockFactoryUtil.java
@@ -11,6 +11,8 @@ import org.mule.runtime.core.internal.lock.ServerLockFactory;
 
 /**
  * Utility method for getting the runtime {@link LockFactory}.
+ *
+ * @since 4.3, 4.2.2
  */
 public class RuntimeLockFactoryUtil {
 
--- a/core/src/main/java/org/mule/runtime/core/internal/lock/ServerLockFactory.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/lock/ServerLockFactory.java
@@ -17,7 +17,7 @@ import java.util.concurrent.locks.Lock;
  * An implementation of {@link LockFactory} that is defined at the container level and can be shared
  * between deployable artifacts to synchronize access to shared resources.
  *
- * @since 4.2.2
+ * @since 4.3.0, 4.2.2
  */
 public class ServerLockFactory implements LockFactory, Disposable {
 
--- a/core/src/main/java/org/mule/runtime/core/internal/lock/ServerLockFactory.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/lock/ServerLockFactory.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.internal.lock;
+
+import org.mule.runtime.api.lifecycle.Disposable;
+import org.mule.runtime.api.lock.LockFactory;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+
+/**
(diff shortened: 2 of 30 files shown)


core/src/main/java/org/mule/runtime/core/internal/config/RuntimeLockFactoryUtil.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package org.mule.runtime.core.internal.config;
+
+import org.mule.runtime.api.lock.LockFactory;
+import org.mule.runtime.core.internal.lock.ServerLockFactory;
+
+/**
+ * Utility method for getting the runtime {@link LockFactory}.
+ */
+public class RuntimeLockFactoryUtil {
```
elrodro83: `@since`

core/src/main/java/org/mule/runtime/core/internal/metadata/cache/DefaultPersistentMetadataCacheManager.java
```diff
@@ -44,23 +48,52 @@
...
 
+  /**
+   * Default implementation should use an {@link ObjectStoreManager} that is tied to the deployable artifact lifecyle.
+   */
   @Inject
   @Named(OBJECT_STORE_MANAGER)
   private ObjectStoreManager objectStoreManager;
 
+  /**
+   * Default implementation should a {@link LockFactory} that comes from the deployavle artifact context.
```
elrodro83: `deployable`

core/src/main/java/org/mule/runtime/core/internal/store/SharedPartitionedPersistentObjectStore.java
```diff
@@ -21,16 +25,41 @@
...
-  public SharedPartitionedPersistentObjectStore(File workingDirectory) {
+  public SharedPartitionedPersistentObjectStore(File workingDirectory, LockFactory lockFactory) {
     checkArgument(workingDirectory != null, "workingDirectory cannot be null");
     this.workingDirectory = workingDirectory;
+    this.lock = lockFactory.createLock(getWorkingDirectory());
+  }
+
+  @Override
+  public void open() throws ObjectStoreException {
+    lock.lock();
```
elrodro83: use `ConcurrencyUtils#withLock`

core/src/main/java/org/mule/runtime/core/internal/store/SharedPartitionedPersistentObjectStore.java
```diff
@@ -21,16 +25,41 @@
...
+    try {
+      super.open();
+    } finally {
+      lock.unlock();
+    }
+  }
+
+  @Override
+  public synchronized void open(String partitionName) throws ObjectStoreException {
+    lock.lock();
```
elrodro83: use `ConcurrencyUtils#withLock`

core/src/main/java/org/mule/runtime/core/internal/lock/ServerLockFactory.java
```diff
@@ -0,0 +1,79 @@
...
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+
+/**
+ * An implementation of {@link LockFactory} that is defined at the container level and can be shared
+ * between deployable artifacts to synchronize access to shared resources.
+ *
+ * @since 4.2.2
```
lucianoRM: add since 4.3.0 too

core/src/main/java/org/mule/runtime/core/internal/metadata/cache/DefaultPersistentMetadataCacheManager.java
```diff
@@ -44,23 +48,52 @@
...
+   */
   @Inject
```
lucianoRM: shouldn't we name all injected parameters to avoid errors if more than one is in the registry?

core/src/main/java/org/mule/runtime/core/internal/metadata/cache/DefaultPersistentMetadataCacheManager.java
```diff
@@ -44,23 +48,52 @@
...
   private LockFactory lockFactory;
 
   private LazyValue<ObjectStore<MetadataCache>> metadataStore;
 
+  private Supplier<ObjectStoreManager> objectStoreManagerSupplier;
+
+  public DefaultPersistentMetadataCacheManager() {}
+
+  /**
+   * Allows to create an instance of this manager with the given parameters instead of letting the registry to create it.
```
lucianoRM: Allows to create an instance of this manager with the given parameters instead of letting the registry REMOVE[to]REMOVE create it.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/SpringMuleContextServiceConfigurator.java
```diff
@@ -241,6 +243,7 @@
       .put(OBJECT_SCHEDULER_BASE_CONFIG, getBeanDefinition(SchedulerBaseConfigFactory.class))
       .put(OBJECT_CLUSTER_SERVICE, getBeanDefinition(DefaultClusterService.class))
       .put(LAZY_COMPONENT_INITIALIZER_SERVICE_KEY, getBeanDefinition(NoOpLazyComponentInitializer.class))
+      .put(METADATA_CACHE_MANAGER_KEY, getBeanDefinition(DefaultPersistentMetadataCacheManager.class))
```
lucianoRM: is this only to be able to inject things into instances? cant you use the muleContext's injector instead?

core/src/main/java/org/mule/runtime/core/internal/store/SharedPartitionedPersistentObjectStore.java
gsfernandes: Doesn't work for me as I have to keep the current interface and throw a checked exception ObjectStoreException and the utility withLock will wrap it with a ReactorException.

core/src/main/java/org/mule/runtime/core/internal/metadata/cache/DefaultPersistentMetadataCacheManager.java
gsfernandes: Yes, but it is not named AFAIK the lock factory in registry right now.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/SpringMuleContextServiceConfigurator.java
gsfernandes: I had to add this to the registry instead of registry-bootstrap mechanism as I need to change the implementation of the service if a shared persistence folder for OS is set on lazy mule artifact contexts.

PR #8307, opened 2019-10-15
MULE-17647: Allow domains to be redeployed without restarting its stopped applications

eze210:


modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/ArtifactDeployer.java
```diff
@@ -20,10 +20,21 @@
    *
    * The deployer executes the artifact installation phases until the artifact is deployed After this method call the Artifact
    * will be installed in the container and started.
+   *  @param artifact artifact to be deployed
+   *  @param startArtifact Whether the artifact should be started after initialisation
+   */
+  void deploy(final T artifact, boolean startArtifact);
```
fsgonz: This will break backwards compatibility. Add at least a default. There are some classes in the maven plugin which implements this interface.

modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/ArtifactDeployer.java
```diff
@@ -20,10 +20,21 @@
    *
    * The deployer executes the artifact installation phases until the artifact is deployed After this method call the Artifact
    * will be installed in the container and started.
+   *  @param artifact artifact to be deployed
+   *  @param startArtifact Whether the artifact should be started after initialisation
```
fsgonz: Whether -> whether

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/ApplicationDependingOnDomainDeploymentTestCase.java
```diff
@@ -216,4 +224,70 @@ public void compatibleDomainNameIsSetAfterDeployment() throws Exception {
     ApplicationDescriptor applicationDescriptor = application.getDescriptor();
     assertThat(applicationDescriptor.getDomainName(), is(emptyDomain101FileBuilder.getId()));
   }
+
+  @Test
+  public void stoppedApplicationsAreNotStartedWhenDomainIsRedeployed() throws Exception {
+    DeploymentListener mockDeploymentListener = spy(new DeploymentStatusTracker());
```
fsgonz: I think these two tests have duplicate code. Maybe they can be refactored taking into account this.

modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/ArtifactDeployer.java
fsgonz: As you noticed, the other cases depend on another interface. My mistake. This is internal.

PR #8316, opened 2019-10-17
MULE-17627: fix scheduler stopping logic

marianogonzalez:


core/src/main/java/org/mule/runtime/core/internal/processor/strategy/AbstractReactorStreamProcessingStrategy.java
```diff
@@ -119,10 +119,17 @@ public void stop() throws MuleException {
...
+   * Stops all schedulers that should be stopped by this ProcessingStrategy
+   *
+   * @return whether or not any scheduler were stopped
+   */
+  protected boolean stopSchedulers() {
     if (cpuLightScheduler != null) {
       cpuLightScheduler.stop();
     }
+
+    return true;
```
elrodro83: shouldn't this be `false`, and `true` inside the if above?

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/StreamEmitterProcessingStrategyFactory.java
```diff
@@ -152,6 +159,8 @@ protected void stopSchedulers() {
           }
         }
       }
+
+      return shouldStop;
```
elrodro83: shouldn't this be `true` only if `super.stopSchedulers()` returned `true` and `flowDispatchScheduler != null`?

PR #8317, opened 2019-10-17
MULE-17667: Automatically determine if a parameter is the ComponentId

elrodro83:
* Add missing name parameters to core extension declarer elements
* Account for synthetic name parameter added for top level pojos

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
+++ b/core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
@@ -137,7 +137,7 @@ class MuleExtensionModelDeclarer {
     // constructs
     declareObject(extensionDeclarer, typeLoader);
     declareFlow(extensionDeclarer, typeLoader);
-    declareSubflow(extensionDeclarer);
+    declareSubflow(extensionDeclarer, typeLoader);
     declareChoice(extensionDeclarer, typeLoader);
     declareErrorHandler(extensionDeclarer, typeLoader);
     declareTry(extensionDeclarer, typeLoader);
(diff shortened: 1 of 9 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/ParameterModelsLoaderDelegate.java
```diff
@@ -98,7 +99,7 @@ public ParameterModelsLoaderDelegate(List<ParameterDeclarerContributor> contribu
                                          ParameterDeclarationContext declarationContext,
                                          ParameterGroupDeclarer parameterGroupDeclarer) {
     List<ParameterDeclarer> declarerList = new ArrayList<>();
-    checkAnnotationsNotUsedMoreThanOnce(parameters, Connection.class, Config.class, MetadataKeyId.class);
+    checkAnnotationsNotUsedMoreThanOnce(parameters, ComponentId.class, Connection.class, Config.class, MetadataKeyId.class);
```
ndinu: This seems to be checked in ParameterModelValidator, I think it is better if it keeps failing there.

core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -571,6 +573,8 @@ private void declareSubflow(ExtensionDeclarer extensionDeclarer) {
         .allowingTopLevelDefinition()
         .withStereotype(SUB_FLOW);
 
+    subFlow.onDefaultParameterGroup().withRequiredParameter("name").asComponentId()
+        .ofType(BaseTypeBuilder.create(JAVA).stringType().build());
```
ndinu: use typeLoader

ndinu: use typeLoader

PR #8318, opened 2019-10-17
MULE-17541: Lazy init: <object class=... '/> not lazy initialized - A…

gsfernandes:
…void use of Stereotype and set as 'alwaysEnabled'


core/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -172,7 +171,6 @@ ExtensionDeclarer createExtensionModel() {
 
   private void declareObject(ExtensionDeclarer extensionDeclarer, ClassTypeLoader typeLoader) {
     ConstructDeclarer object = extensionDeclarer.withConstruct("object")
```
elrodro83: shouldn't this be deprecated?

gsfernandes: Yes, I will add the deprecation model.

PR #8322, opened 2019-10-21
MULE-17674: Only allow MuleSession instances when deserializing corresponding header

eze210:


core/src/main/java/org/mule/session/AbstractSessionHandler.java
```diff
@@ -45,6 +47,18 @@
...
 
+    private ObjectSerializer setSerializerClass(ObjectSerializer serializer, Class clazz)
+    {
+        if (serializer instanceof ClassSpecificObjectSerializer)
+        {
+            return serializer;
+        }
+        else
+        {
+            return new ClassSpecificObjectSerializer(serializer, clazz);
```
aiannucci: Will this create a new instance every time? Is this really needed?

core/src/test/java/org/mule/MuleSessionHandlerTestCase.java
```diff
@@ -119,6 +124,20 @@ public void testSessionProperties() throws Exception
         assertEquals(list, obj);
     }
 
+    @Test
+    public void testDontDeserializeOtherThanMuleSession() throws MuleException
+    {
+        DefaultMuleMessage message = new DefaultMuleMessage("Test Message", muleContext);
+        SessionHandler handler = new SerializeAndEncodeSessionHandler();
+
+        String encodedSet = new String(Base64.encode(SerializationUtils.serialize(new HashSet<String>())));
```
fsgonz: static import Base64.encode and SerializationUtils.serialize

core/src/main/java/org/mule/serialization/internal/ClassSpecificObjectSerializer.java
```diff
@@ -0,0 +1,93 @@
...
+        this.classToAccept = classToAccept;
+    }
+
+    @Override
+    protected <T> T doDeserialize(InputStream inputStream, ClassLoader classLoader) throws Exception
+    {
+        checkArgument(inputStream != null, "Cannot deserialize a null stream");
+        checkArgument(classLoader != null, "Cannot deserialize with a null classloader");
+
+        return (T) SerializationUtils.deserialize(inputStream, classLoader, muleContext, new ObjectInputStreamProvider()
```
fsgonz: static import.

core/src/main/java/org/mule/serialization/internal/ClassSpecificObjectSerializer.java
```diff
@@ -0,0 +1,93 @@
...
+         * @param classLoader the ClassLoader from which classes should be loaded
+         * @param inputStream the InputStream to work on
+         * @throws IOException              in case of an I/O error
+         * @throws StreamCorruptedException if the stream is corrupted
+         */
+        public ClassSpecificObjectInputStream(ClassLoader classLoader, InputStream inputStream, Class classToAccept) throws IOException, StreamCorruptedException
+        {
+            super(classLoader, inputStream);
+            this.classToAccept = classToAccept;
+            this.firstClass = true;
```
fsgonz: set the true value in the field declaration.

core/src/main/java/org/mule/serialization/internal/ClassSpecificObjectSerializer.java
eze210: The base class (AbstractObjectSerializer) already declares a deserialize() method

fsgonz: You're right.

PR #8328, opened 2019-10-23
MULE-17398: Operations that return a PagingProvider do not populate the type catalog properly

SebaElizalde:


modules/extensions-spring-support/src/test/resources/models/heisenberg.json
```diff
@@ -4421,6 +4421,573 @@
...
+          },
+          "parameterGroupModels": [
+            {
+              "parameters": [
+                {
+                  "type": {
+                    "format": "java",
+                    "type": "Object",
+                    "annotations": {
+                      "typeId": "org.mule.runtime.extension.api.runtime.config.ConfigurationProvider"
```
marianogonzalez: this should never appear here

modules/extensions-spring-support/src/test/resources/models/heisenberg.json
```diff
@@ -4421,6 +4421,573 @@
...
+        {
+          "blocking": true,
+          "executionType": "BLOCKING",
+          "output": {
+            "type": {
+              "format": "java",
+              "type": "Array",
+              "annotations": {
+                "classInformation": {
+                  "classname": "org.mule.runtime.extension.api.runtime.streaming.PagingProvider",
```
marianogonzalez: this shouldn't appear here

PR #8332, opened 2019-10-24
MULE-17687: NPE when stream is remanaged after previous managing decorator is collected

marianogonzalez:


core/src/main/java/org/mule/runtime/core/internal/streaming/CursorManager.java
```diff
@@ -88,7 +88,27 @@ private void hookEventTermination(BaseEventContext ownerContext) {
     private final Cache<Integer, WeakReference<ManagedCursorProvider>> providers = Caffeine.newBuilder().build();
 
     private ManagedCursorProvider addProvider(ManagedCursorProvider provider) {
-      return providers.get(identityHashCode(provider.getDelegate()), hash -> ghostBuster.track(provider)).get();
+      final int hash = identityHashCode(provider.getDelegate());
+      ManagedCursorProvider managedProvider = getOrAddManagedProvider(provider, hash);
+
+      // this can happen in cases like when foreach splits a text document using a stream. N+1 iterations might
```
fsgonz: I would use "This can happen when a foreach component splits a text document using a stream"

PR #8333, opened 2019-10-25
MULE-17689: DefaultMetadataCacheManager is not reusing the ObjectStor…

gsfernandes:
…e partition

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/store/SharedPartitionedPersistentObjectStore.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/store/SharedPartitionedPersistentObjectStore.java
@@ -10,14 +10,19 @@ import static org.mule.runtime.api.util.Preconditions.checkArgument;
 
 import org.mule.runtime.api.lock.LockFactory;
 import org.mule.runtime.api.store.ObjectStoreException;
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/mule/runtime/core/internal/store/SharedPartitionedPersistentObjectStore.java
```diff
@@ -27,6 +32,29 @@
...
+
+        @Override
+        public PersistentObjectStorePartition put(String key, PersistentObjectStorePartition value) {
+          // Creates an instance of the information to avoid referencing to the muleContext as the same persistentObjectStorePartition
+          // is used by different muleContexts
+          return super.put(key, new PersistentObjectStorePartitionData(value.getPartitionName(), value.getPartitionDirectory()));
+        }
+
+        @Override
+        public PersistentObjectStorePartition putIfAbsent(String key, PersistentObjectStorePartition value) {
```
lucianoRM: is it really necessary? Doesn't the ConcurrentHashMap implementation call put() when calling putIfAbsent()?

gsfernandes: No, it does call an internal method that I can't override but it is not delegating to the put method.

PR #8339, opened 2019-10-29
MULE-17689: Clean up resources on stop should also remove references …

gsfernandes:
…to beans and locations requested


modules/spring-config/src/main/java/org/mule/runtime/config/internal/LazyMuleArtifactContext.java
```diff
@@ -360,6 +360,7 @@ public void initializeComponents(Predicate<ComponentAst> componentModelPredicate
               .map(comp -> comp.getLocation().getLocation())
               .collect(toSet()));
 
+
```
lucianoRM: remove empty line

PR #8342, opened 2019-10-29
MULE-17386: CommonSourcePolicy callback invoked twice when event is failed

JuLuzzi:


core/src/main/java/org/mule/runtime/core/internal/execution/FlowProcessMediator.java
```diff
@@ -193,21 +193,15 @@ private void dispatch(PhaseContext ctx) throws Exception {
...
-                                     dispatchResponse(ctx);
-                                   }
+                                   ctx.result = value;
+                                   dispatchResponse(ctx);
                                  }
 
                                  @Override
                                  public void error(Throwable e) {
-                                   //TODO: Remove this if when MULE-17386 is fixed
-                                   if (ctx.result == null) {
```
marianogonzalez: when I added this workaround I was trying the proxy performance test, which doesn't use any policies. Are you sure that the fix in `CommonSourcePolicy` is enough to be able to resolve this?

PR #8349, opened 2019-10-31
MULE-17505: Runtime is not starting ConnectionProviders during metada…

gsfernandes:
…ta resolution phase

### DIFF
--- a/modules/spring-config/src/main/java/org/mule/runtime/config/internal/LazyMuleArtifactContext.java
+++ b/modules/spring-config/src/main/java/org/mule/runtime/config/internal/LazyMuleArtifactContext.java
@@ -251,7 +251,7 @@ public class LazyMuleArtifactContext extends MuleArtifactContext
    */
   private void startConfigurationProviders(List<Object> components) {
     components.stream()
-        .filter(component -> ConfigurationProvider.class.isAssignableFrom(component.getClass()))
+        .filter(component -> component instanceof ConfigurationProvider)
         .forEach(configurationProviders -> {
           try {
             getMuleRegistry().applyLifecycle(configurationProviders, Initialisable.PHASE_NAME, Startable.PHASE_NAME);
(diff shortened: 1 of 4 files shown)


modules/spring-config/src/main/java/org/mule/runtime/config/internal/LazyMuleArtifactContext.java
```diff
@@ -232,12 +233,34 @@ private void applyLifecycle(List<Object> components, boolean applyStartPhase) {
...
 
+  /**
+   * Starts {@link ConfigurationProvider} components as they should be started no matter if the request has set to not apply start
+   * phase in the rest of the components.
+   *
+   * @param components list of components created
+   */
+  private void startConfigurationProviders(List<Object> components) {
+    components.stream()
+        .filter(component -> ConfigurationProvider.class.isAssignableFrom(component.getClass()))
```
marianogonzalez: why not just a plain old `instanceof`?

modules/spring-config/src/main/java/org/mule/runtime/config/internal/LazyMuleArtifactContext.java
```diff
@@ -232,12 +233,34 @@ private void applyLifecycle(List<Object> components, boolean applyStartPhase) {
...
+        .forEach(configurationProviders -> {
+          try {
+            getMuleRegistry().applyLifecycle(configurationProviders, Initialisable.PHASE_NAME, Startable.PHASE_NAME);
```
marianogonzalez: static imports

tests/test-extensions/metadata-extension/src/main/java/org/mule/test/metadata/extension/MetadataConnectionProvider.java
```diff
@@ -38,4 +42,9 @@ public void disconnect(MetadataConnection metadataConnection) {
   public ConnectionValidationResult validate(MetadataConnection metadataConnection) {
     return ConnectionValidationResult.success();
   }
+
+  @Override
+  public void start() throws MuleException {
+    STARTED = true;
```
marianogonzalez: shouldn't this be set to false on the test setup to prevent false positives from previous that also use this class?

PR #8356, opened 2019-11-01
MULE-17704: OAuth doesn't refresh token when client maxRetries property set to zero

eze210:


modules/oauth/src/test/java/org/mule/module/oauth2/internal/clientcredentials/functional/ClientCredentialsFullConfigTestCase.java
```diff
@@ -110,6 +110,18 @@ public void customTokenResponseParametersAreCaptured() throws Exception
...
     @Test
     public void authenticationFailedTriggersRefreshAccessToken() throws Exception
+    {
+        assertAuthenticationFailedTriggersRefreshAccessToken();
+    }
+
+    @Test
+    public void authenticationFailedTriggersRefreshAccessTokenWithMaxRetriesSetToZero() throws Exception
+    {
+        System.setProperty("mule.http.client.maxRetries", "0");
```
elrodro83: use a junit rule

elrodro83: or make sure it is cleaned up to its previous value after the test

PR #8358, opened 2019-11-04
MULE-17386: CommonSourcePolicy callback invoked twice when event is f…

JuLuzzi:
…ailed

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
@@ -239,7 +239,9 @@ public abstract class AbstractPipeline extends AbstractFlowConstruct implements 
   private ReactiveProcessor dispatchToFlow() {
     return publisher -> from(publisher)
         .doOnNext(assertStarted())
-        .flatMap(routeThroughProcessingStrategy()).compose(pub -> pub.subscriberContext(context -> {
+        .flatMap(routeThroughProcessingStrategy())
+        // This replaces the onErrorContinue key if it exists, to prevent it from being propagated within the flow
+        .compose(pub -> pub.subscriberContext(context -> {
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
```diff
@@ -235,7 +239,15 @@ public CoreEvent process(CoreEvent event) throws MuleException {
...
         .doOnNext(assertStarted())
-        .flatMap(routeThroughProcessingStrategy());
+        .flatMap(routeThroughProcessingStrategy()).compose(pub -> pub.subscriberContext(context -> {
+          Optional<Object> onErrorStrategy = context.getOrEmpty(KEY_ON_NEXT_ERROR_STRATEGY);
+          if (onErrorStrategy.isPresent()
+              && onErrorStrategy.get().toString().contains(ON_NEXT_FAILURE_STRATEGY)) {
+            BiFunction<Throwable, Object, Throwable> onErrorContinue = (e, o) -> null;
+            return context.put(KEY_ON_NEXT_ERROR_STRATEGY, onErrorContinue);
+          }
+          return context;
```
elrodro83: add a comment explaining why this has to be done this way

core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
```diff
@@ -235,7 +239,15 @@ public CoreEvent process(CoreEvent event) throws MuleException {
   private ReactiveProcessor dispatchToFlow() {
     return publisher -> from(publisher)
         .doOnNext(assertStarted())
-        .flatMap(routeThroughProcessingStrategy());
+        .flatMap(routeThroughProcessingStrategy()).compose(pub -> pub.subscriberContext(context -> {
```
elrodro83: line break before compose (each reactor operator in its line)

PR #8361, opened 2019-11-05
MULE-16527: Handle exception when using UntilSuccessful#process

plbalbi:


core/src/main/java/org/mule/runtime/core/internal/routing/UntilSuccessful.java
```diff
@@ -74,7 +76,19 @@ public void dispose() {
 
   @Override
   public CoreEvent process(CoreEvent event) throws MuleException {
-    return processToApply(event, this);
+    try {
+      return processToApply(event, this);
+    } catch (Exception error) {
+      Throwable cause = error.getCause();
+      if (cause instanceof RetryContextInitializationException &&
```
elrodro83: check for `cause != null`

plbalbi: Nice catch

PR #8362, opened 2019-11-06
MULE-17725: DomainClassLoaderFactory doesn't remove cached class loaders on dispose

eze210:


modules/deployment-model/src/test/java/org/mule/runtime/deployment/model/internal/domain/DomainClassLoaderFactoryTestCase.java
```diff
@@ -95,6 +96,33 @@ public void createClassLoaderFromDomainDescriptor() {
     assertThat(domainClassLoader.getArtifactId(), equalTo(artifactId));
   }
 
+  @Test
+  public void secondClassLoaderCreationReturnsCachedInstance() {
```
fsgonz: merge these two tests in one verify that before disposing the same instance is returned from the factory and after disposing you get a new instance.

modules/deployment-model/src/test/java/org/mule/runtime/deployment/model/internal/domain/DomainClassLoaderFactoryTestCase.java
```diff
@@ -95,6 +96,33 @@ public void createClassLoaderFromDomainDescriptor() {
...
+    final String domainName = "custom-domain";
```
fsgonz: refactor the creating of the factory in one method to avoid duplicating code.

PR #8363, opened 2019-11-06
MULE-17371: Add Parameter models to ComponentAst

elrodro83:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/SpringComponentModel.java
```diff
@@ -80,6 +95,80 @@ public ComponentLocation getLocation() {
...
+            return new NoSuchElementException(mxgPrefix + " Available: " + availableParams);
+          }
+        }));
+  }
+
+  private Optional<ParameterModel> findParameterModel(String paramName) {
+    // For sources, we need to account for the case where parameters in the callbacks may have colliding names.
+    // This logic ensures that the parameter fetching logic is consistent with the logic that handles this scenario in previous
+    // implementations.
+    Optional<ParameterModel> parameterModel = getModel(SourceModel.class)
```
gsfernandes: @ndinu we should review this as it is related to the decision made about allowing to have repeated names on source callbacks

PR #8373, opened 2019-11-07
MULE-17732: JMS transport is not validating timeout

mbuchwald:


transports/jms/src/main/java/org/mule/transport/jms/JmsMessageDispatcher.java
```diff
@@ -292,7 +296,13 @@ private MuleMessage internalBlockingSendAndReceive(Session session, MessageProdu
         final MessageConsumer consumer = createReplyToConsumer(jmsMessage, event, session, replyTo, topic);
         try
         {
+            StopWatch timeoutValidator = new StopWatch();
+            timeoutValidator.start();
             Message result = consumer.receive(timeout);
+            timeoutValidator.stop();
+            if (result == null && timeoutValidator.getTime() >= timeout) {
```
fsgonz: should the validation that the result is null be added. Shouldn't the timeout be respected even if a result was received?

mbuchwald: If there's a result, we should return it without error. It might happen that the result had just came when the timeout was finishing (or even have to consider that there might be a slight delay between the receive and `timeoutValidator.stop()`). Also, being consiste with mule 4

PR #8383, opened 2019-11-13
MTS-301: Add resolved value providers cache generator key

lucianoRM:


modules/spring-config/src/main/java/org/mule/runtime/config/api/dsl/model/metadata/DslElementBasedValueProviderCacheIdGenerator.java
```diff
@@ -0,0 +1,223 @@
...
+
+
+  private Optional<ValueProviderCacheId> resolveIdForInjectedElement(DslElementModel<?> injectedElement) {
+    if (!(injectedElement.getModel() instanceof EnrichableModel)) {
+      return empty();
+    }
+    List<ValueProviderCacheId> parts = new LinkedList<>();
+    List<String> parametersRequiredForMetadata =
+        ((EnrichableModel) injectedElement.getModel())
+            .getModelProperty(RequiredForMetadataModelProperty.class)
```
gsfernandes: RequiredForMetadataModelProperty.class, would this be fine?
Are we going to assume that also parameters required from Metadata are required for ValueProviders. Is this the case of connection/configuration elements only?

modules/spring-config/src/test/resources/log4j2-test.xml
```diff
@@ -23,6 +23,8 @@
 
         <AsyncLogger name="com.mulesoft" level="WARN"/>
 
+        <AsyncLogger name="org.mule.runtime.config.dsl.model" level="DEBUG"/>
```
gsfernandes: Remove the DEBUG configuration.

core/src/main/java/org/mule/runtime/core/internal/value/cache/ValueProviderCacheId.java
```diff
@@ -0,0 +1,142 @@
...
+      this.attributes = new HashMap<>();
+    }
+
+    public static ValueProviderCacheIdBuilder fromElementWithName(String sourceElementName) {
+      ValueProviderCacheIdBuilder builder = new ValueProviderCacheIdBuilder();
+      builder.sourceElementName = sourceElementName;
+      return builder;
+    }
+
+    public ValueProviderCacheIdBuilder withHashValueFrom(Object hashedObject) {
```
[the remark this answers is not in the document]
lucianoRM: The fact that we are using the string representation for every hashedValue is a coincidence. It may not be the same in the future.  Besides, Objects.hash() already works with Object instances, so there is no need to downcast or anything weird.

modules/spring-config/src/main/java/org/mule/runtime/config/api/dsl/model/metadata/DslElementBasedValueProviderCacheIdGenerator.java
```diff
@@ -0,0 +1,223 @@
...
+
+  /**
+   * {@inheritDoc}
+   * <p/>
+   * The returned {@link ValueProviderCacheId} will contain all acting parameters required by the {@link org.mule.runtime.extension.api.values.ValueProvider} as parts.
+   * In case the {@link DslElementModel} corresponds to a Source or Operation, if the {@link org.mule.runtime.extension.api.values.ValueProvider} requires a connection or
+   * a configuration, their id will be added as part.
+   * The resolution of a config or connection id as part is different from the one done when their are the one's holding the resolving parameter.
+   * In the case they are parts needed by another {@link org.mule.runtime.extension.api.values.ValueProvider}, acting parameters will not exist. Therefore, only
+   * parameters required for metadata are used as input to calculate the {@link ValueProviderCacheId}.
```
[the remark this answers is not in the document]
lucianoRM: I don't see the issue here.

modules/spring-config/src/main/java/org/mule/runtime/config/api/dsl/model/metadata/DslElementBasedValueProviderCacheIdGenerator.java
lucianoRM: I also thought about this and i think it's the best approach. 
First, we need to consider that this is only for when connections or configurations are being referenced from an operation or source valueProvider. Otherwise, when we are retrieving values for a Config, we don't use this annotation but the acting parameters provided by the ValueProviderModel.

So, when an operation's value provider makes use of a configuration for example, the parameters that define the values being retrieved should be the ones that define how to access the external system. We already know that for any MetadataResolver (Keys, Output, Input or Attributes) this is done by parameters annotated with @RequiredForMetadata (Or all of them by default). In this case, it makes sense to do the same. Otherwise, in which cases does a ValueProvider from an Operation or Source need a Config or Connection?

PR #8387, opened 2019-11-14
MULE-17741: FirstSuccessful is blocking -- Implementation for 4.1.x

mbuchwald:


core/src/main/java/org/mule/runtime/core/internal/routing/FirstSuccessful.java
```diff
@@ -6,42 +6,131 @@
...
+  public Publisher<CoreEvent> apply(Publisher<CoreEvent> publisher) {
+    return from(publisher)
+        .flatMap(event -> Mono.from(processWithChildContext(event, new FirstSuccessfulRoutingProcessor(), of(getLocation()))));
+  }
+
+  private class FirstSuccessfulRoutingProcessor implements Processor {
+
+    @Override
+    public CoreEvent process(CoreEvent event) throws MuleException {
+      CoreEvent returnEvent = null;
```
mbuchwald: This is the same logic that was in the FSRoutingStrategy, I simply transformed it in a Processor

PR #8392, opened 2019-11-14
Scheduler thing

elrodro83:


core-tests/src/test/java/org/mule/runtime/core/internal/el/function/LookupFunctionTestCase.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeoutException;
+
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+
+public class LookupFunctionTestCase extends AbstractMuleTestCase {
```
fsgonz: Add allure components.

core-tests/src/test/java/org/mule/runtime/core/internal/transformer/simple/AbstractMessageTransformerTestCase.java
```diff
@@ -58,18 +72,37 @@ public DataType getReturnDataType() {
...
-    MuleContext muleContextSpy = spy(muleContext);
-    transformer.setMuleContext(muleContextSpy);
     transform();
     transform();
     transform();
     verify(muleContextSpy, times(1)).getDefaultErrorHandler(any());
   }
 
+  @Test
+  public void doesntCreateTransformationEventIfEventProvided() throws MuleException {
```
fsgonz: I would not use contractions in code names, but I've found other places where this is done.

PR #8395, opened 2019-11-15
MULE-17082: Race condition on test connectivity while deploying

fsgonz:
applications.

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/LifecycleAwareConfigurationInstance.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/LifecycleAwareConfigurationInstance.java
@@ -77,7 +77,7 @@ public final class LifecycleAwareConfigurationInstance extends AbstractIntercept
 
 
 
-  private static final Integer DEFAULT_ASYNC_TEST_CONNECTIVITY_TIMEOUT = 10000;
+  private static final Integer DEFAULT_ASYNC_TEST_CONNECTIVITY_TIMEOUT = 30000;
 
   private static final int ASYNC_TEST_CONNECTIVITY_TIMEOUT =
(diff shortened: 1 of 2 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/LifecycleAwareConfigurationInstance.java
```diff
@@ -72,6 +73,7 @@
  */
 public final class LifecycleAwareConfigurationInstance extends AbstractInterceptable implements ConfigurationInstance {
 
+  private static final int ASYNC_TEST_CONNECTIVITY_TIMEOUT = 5;
```
marianogonzalez: this should be configurable, as a sysprop at the very least. 10 seconds is actually too few in some cases.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/LifecycleAwareConfigurationInstance.java
```diff
@@ -72,6 +75,14 @@
  */
 public final class LifecycleAwareConfigurationInstance extends AbstractInterceptable implements ConfigurationInstance {
 
+  private static final String ASYNC_TEST_CONNECTIVITY_TIMEOUT_PROPERTY =
```
marianogonzalez: This should be defined in MuleSystemProperties class

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/LifecycleAwareConfigurationInstance.java
```diff
@@ -72,6 +75,14 @@
...
+      MuleSystemProperties.SYSTEM_PROPERTY_PREFIX + "async.test.connectivity.timeout";
+
+  private static final Integer DEFAULT_ASYNC_TEST_CONNECTIVITY_TIMEOUT = 10000;
```
marianogonzalez: still too small. My previous comment said that this was not big enough for SaaS services. 30s would be a good default

PR #8402, opened 2019-11-19
MULE-15645: Enable flowStack by default

elrodro83:


core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
```diff
@@ -328,6 +332,11 @@ boolean isCompleteSignalRejectedExecutionException(Throwable throwable) {
       if (getStatistics().isEnabled()) {
         getStatistics().incReceivedEvents();
       }
+
+      FlowCallStack flowCallStack = event.getFlowCallStack();
+      if (flowCallStack instanceof DefaultFlowCallStack) {
```
fsgonz: Shouldn't the push and pop methods be part of the interface? This will always be a stack.

core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
```diff
@@ -353,6 +362,10 @@ boolean isCompleteSignalRejectedExecutionException(Throwable throwable) {
   }
 
   private void fireCompleteNotification(CoreEvent event, MessagingException messagingException) {
+    FlowCallStack flowCallStack = event.getFlowCallStack();
```
fsgonz: Shouldn't the push and pop methods be part of the interface? This will always be a stack.

core/src/main/java/org/mule/runtime/core/internal/context/notification/DefaultFlowCallStack.java
```diff
@@ -41,6 +50,16 @@ private DefaultFlowCallStack(final Deque<FlowStackElement> innerStack) {
    * @param flowStackElement the element to add
    */
   public void push(FlowStackElement flowStackElement) {
+    if (innerStack.size() >= MAX_DEPTH) {
+      StringBuilder messageBuilder = new StringBuilder();
+
+      messageBuilder.append("Too many child contexts nested. ")
```
fsgonz: "Too many nested child context."

core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
elrodro83: Only the runtime internals must push/pop, the exposed interface ir for read-only

core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
elrodro83: same as before

core/src/main/java/org/mule/runtime/core/internal/context/notification/DefaultFlowCallStack.java
elrodro83: image

jbarrington: @elrodro83 please kindly stop using that image. I find it incredibly insensitive and not something I'd expect to see in a Mulesoft repo.

PR #8403, opened 2019-11-19
MULE-17762: MuleContext.getSecurityManager() accesses the registry on…

marianogonzalez:
… every invocation


core/src/main/java/org/mule/runtime/core/internal/context/DefaultMuleContext.java
```diff
@@ -588,16 +590,21 @@ public void setSecurityManager(SecurityManager securityManager) {
    */
   @Override
   public SecurityManager getSecurityManager() {
-    SecurityManager securityManager = muleRegistryHelper.lookupObject(OBJECT_SECURITY_MANAGER);
     if (securityManager == null) {
```
afelisatti: `this.securityManager`?

marianogonzalez: that would be redundant

PR #8415, opened 2019-11-20
MULE-17386: CommonSourcePolicy callback invoked twice when event is failed

JuLuzzi:


core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
```diff
@@ -260,8 +261,11 @@ private ReactiveProcessor dispatchToFlow() {
       Optional<Object> onErrorStrategy = context.getOrEmpty(KEY_ON_NEXT_ERROR_STRATEGY);
       if (onErrorStrategy.isPresent()
           && onErrorStrategy.get().toString().contains(ON_NEXT_FAILURE_STRATEGY)) {
```
plbalbi: I'd check if the strategy corresponds to an error continue by using `errorStrategy.get().getClass().getName()` instead of the `toString()`

PR #8420, opened 2019-11-21
MULE-17770: Allow UTF-8 in disposition content headers for http request

fsgonz:


modules/http/src/test/java/org/mule/module/http/functional/listener/HttpRequesterPartsTestCase.java
```diff
@@ -0,0 +1,59 @@
...
+{
+
+  private static final String AHC_UTF_PROPERTY = "ahc.request.part.headers.allowUtf8";
+  private static final String MIME_UTF_PROPERTY = "mail.mime.allowutf8";
+
+  @Rule
+  public DynamicPort listenPort = new DynamicPort("port");
+  
+  @Rule
+  public SystemProperty encodingRequester = new SystemProperty(AHC_UTF_PROPERTY, "true");
```
afelisatti: I think this will have the same problem as before.

PR #8428, opened 2019-11-25
MULE-17780: Avoid creating an extra intercepting chain when there are no

elrodro83:
interceptors

* Also address some flaky tests in `core-tests`

### DIFF
--- a/core-tests/src/test/java/org/mule/runtime/core/privileged/processor/chain/DefaultMessageProcessorChainTestCase.java
+++ b/core-tests/src/test/java/org/mule/runtime/core/privileged/processor/chain/DefaultMessageProcessorChainTestCase.java
@@ -106,6 +106,7 @@ import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import org.reactivestreams.Publisher;
+
 import reactor.core.publisher.Flux;
 
 @RunWith(Parameterized.class)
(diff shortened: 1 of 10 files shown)


core-tests/src/test/java/org/mule/runtime/core/privileged/processor/chain/DefaultMessageProcessorChainTestCase.java
```diff
@@ -185,9 +186,9 @@ public void before() throws MuleException {
 
   @After
   public void after() throws MuleException {
-    stopIfNeeded(muleContext.getSchedulerService());
     flow.stop();
     flow.dispose();
+    stopIfNeeded(muleContext.getSchedulerService());
```
marianogonzalez: shouldn't this always happen on the test infrastructure ?

core/src/main/java/org/mule/runtime/core/internal/processor/chain/SubflowMessageProcessorChainBuilder.java
```diff
@@ -79,21 +81,20 @@ public Location getRootContainerLocation() {
   }
 
   @Override
-  protected MessageProcessorChain createInterceptingChain(Processor head, List<Processor> processors,
-                                                          List<Processor> processorForLifecycle) {
-    return new SubFlowMessageProcessorChain(name, head, processors, processorForLifecycle);
+  protected MessageProcessorChain createSimpleChain(List<Processor> processors,
```
marianogonzalez: I would rename this to `createDefaultChain` or simply `createChain`

elrodro83: privileged api...

PR #8429, opened 2019-11-25
MULE-17765: Validation module does not map MULE:EXPRESSION errors to …

JuLuzzi:
…custom application errors and error logging differs when error is raised in sub-flow


elrodro83:
--validate

PR #8436, opened 2019-11-28
MULE-17544: Fix AsyncHeisenbergMessageSourceTestCase flaky test

JuLuzzi:


tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/HeisenbergExtension.java
```diff
@@ -100,7 +100,7 @@
   private int start = 0;
   private int stop = 0;
   private int dispose = 0;
-  public static int sourceTimesStarted = 0;
+  public static volatile Integer sourceTimesStarted = 0;
```
elrodro83: this can be an `int` with the `volatile`, or an `AtomicInteger` directly if not.

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/HeisenbergSource.java
```diff
@@ -173,15 +173,15 @@ public void onStart(SourceCallback<String, Object> sourceCallback) throws MuleEx
         context.fireOnHandle(NEXT_BATCH, TypedValue.of(frequency));
         sourceCallback.handle(result, context);
       }
-    }, 0, frequency, MILLISECONDS);
+    }, 0, 4000, MILLISECONDS);
```
elrodro83: the frequency should still be the one from the config. It you need to change the frequency for a single test, change it in that test xml. If you need it to be generic, update the default value directly

PR #8438, opened 2019-11-28
MULE-16461: Remove `concatMap` in XML SDK chain execution

elrodro83:
Also:

* Make the `xml-sdk` operation *have* a chain instead of *being* a chain.
* Move the `xml-sdk` operation from `core` to `extensions-xml-support`, where it belongs.
* Some minor changes regarding error handling in `xml-sdk`


modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/processor/ModuleOperationMessageProcessor.java
```diff
@@ -0,0 +1,362 @@
...
+
+          final ComponentIdentifier identifier = (ComponentIdentifier) getAnnotation(ANNOTATION_NAME);
+          if (identifier.getNamespace() == null || "tns".equals(identifier.getNamespace())) {
+            final String[] peekedWithNamespace = flowCallStack.peek().getFlowName().split("\\:");
+            String peekedNamespace = peekedWithNamespace[0];
+
+            flowCallStack.push(new FlowStackElement(peekedNamespace + ":" + identifier.getName(), null));
+          } else {
+            flowCallStack.push(new FlowStackElement(identifier.getNamespace() + ":" + identifier.getName(), null));
+          }
```
plbalbi: Extract this, or just the `FlowStackElement` creation part to make the chain more readable

plbalbi: Maybe just the `FlowStackElement`element creations, so its consistent with the pop below

PR #8441, opened 2019-11-29
MULE-17800: ParseTemplateProcessor Type Inconsistencies when setting target and target variable

mbuchwald:


core/src/main/java/org/mule/runtime/core/internal/processor/simple/ParseTemplateProcessor.java
```diff
@@ -33,6 +36,8 @@
 public class ParseTemplateProcessor extends SimpleMessageProcessor {
 
   private static final MimetypesFileTypeMap mimetypesFileTypeMap = new MimetypesFileTypeMap();
+  private static final Boolean KEEP_TYPE_TARGET_AND_TARGET_VAR =
+      new Boolean(getProperty("mule.parse.template.keep.target.var.type", "true"));
```
fsgonz: define this in Mule Properties. Use prefix.

fsgonz: use prefix.

PR #8456, opened 2019-12-04
MULE-17820: Use SmallMap to store event variables

marianogonzalez:


core/src/main/java/org/mule/runtime/core/api/util/CaseInsensitiveHashMap.java
```diff
@@ -63,6 +63,16 @@
     return EMPTY_MAP;
   }
 
+  /**
+   * Creates a new instance which is backed by the given {@code map}. Said map is required to be empty
```
elrodro83: non empty

marianogonzalez: not.. It must be empty

PR #8458, opened 2019-12-04
MULE-17825: Group error handling internal parameters into unified pojos

elrodro83:


core/src/main/java/org/mule/runtime/core/privileged/exception/TemplateOnErrorHandler.java
```diff
@@ -86,11 +87,10 @@
   private static final Logger LOGGER = getLogger(TemplateOnErrorHandler.class);
 
   private static final Pattern ERROR_HANDLER_LOCATION_PATTERN = compile(".*/.*/.*");
-  private static final String ERROR_EXCEPTION = "error.exception.";
-  private static final String ERROR_SINK = "error.sink.";
-  private static final String ERROR_SUCCESS_CALLBACK = "error.success.";
-  private static final String ERROR_RETHROWN_CALLBACK = "error.error.";
-  private static final String ERROR_EVENT = "error.event.";
+  // private static final String ERROR_EXCEPTION = "error.exception.";
```
marianogonzalez: remove

core/src/main/java/org/mule/runtime/core/internal/exception/ErrorHandlerContext.java
```diff
@@ -0,0 +1,77 @@
...
+   * @param event
+   * @return an {@link ErrorHandlerContext} or {@code null} if none was set on the event
+   */
+  public static ErrorHandlerContext from(CoreEvent event) {
+    return ((InternalEvent) event).getInternalParameter(ERROR_HANDLER_CONTEXT);
+  }
+
+  private final Map<String, ErrorHandlerContextItem> items = new SmallMap<>();
+
+  public void configure(String parameterId, Exception exception, CoreEvent originalEvent,
```
marianogonzalez: shouldn't this method be called something else? it looks more like a setParamInfo or something like that

PR #8472, opened 2019-12-10
MULE-17757: Migrate uses of `MessagingExceptionResolver#resolve(MessagingException, MuleContext)`

elrodro83:
* Remove policyManager bean from registry for non-app deployments
* Partial revert of tests impact from MULE-16371


core-tests/src/test/java/org/mule/runtime/core/api/DefaultMuleEventTestCase.java
```diff
@@ -52,7 +48,7 @@
   public static final String PROPERTY_NAME = "test";
   public static final String PROPERTY_VALUE = "foo";
 
-  private final Message muleMessage = of("test-data");
+  private Message muleMessage = of("test-data");
```
fsgonz: Why removing final?

core-tests/src/test/java/org/mule/runtime/core/internal/el/ExpressionLanguageExtensionTestCase.java
```diff
@@ -26,16 +27,15 @@
...
 import java.util.Calendar;
-import java.util.HashMap;
 import java.util.Map;
 
-import org.junit.Test;
-
 public class ExpressionLanguageExtensionTestCase extends AbstractELTestCase {
 
-  private final String a = "hi";
+  private String a = "hi";
```
fsgonz: why remove final? If it is not final as b, maybe it shouldn't be declared as an instance field but as a local variable in the test. Actually if it is something to compare to, this should be a constant.

core-tests/src/test/java/org/mule/runtime/core/internal/routing/UntilSuccessfulTestCase.java
```diff
@@ -109,15 +107,14 @@ public UntilSuccessfulTestCase(boolean tx) {
...
-              OBJECT_ERROR_TYPE_REPO_REGISTRY_KEY, createDefaultErrorTypeRepository());
+    return singletonMap(REGISTRY_KEY, componentLocator);
   }
 
   @Override
   protected void doSetUp() throws Exception {
     super.doSetUp();
     flow = createAndRegisterFlow(muleContext, APPLE_FLOW, componentLocator);
-    untilSuccessful = buildUntilSuccessful("1000");
+    untilSuccessful = buildUntilSuccessful("100");
```
fsgonz: use constant declarative field.

core/src/main/java/org/mule/runtime/core/internal/registry/SimpleRegistry.java
```diff
@@ -140,21 +141,35 @@ private void injectFieldDependencies() throws InitialisationException {
       Class<?> dependencyType = field.getType();
 
       boolean nullToOptional = false;
-      if (dependencyType.equals(Optional.class)) {
+      boolean collection = false;
+      if (dependencyType.equals(Optional.class) || Collection.class.isAssignableFrom(dependencyType)) {
```
fsgonz: Is this for the case in which it is an Optional and it is also assignableFrom(dependencyType) to manage precedence? Try to make it clearer if this is the case and refactor if this is not a case with two conditions to reduce cyclomatic complexity.

core/src/main/java/org/mule/runtime/core/internal/registry/SimpleRegistry.java
```diff
@@ -169,20 +184,32 @@ private void injectFieldDependencies() throws InitialisationException {
         Class<?> dependencyType = method.getParameterTypes()[0];
 
         boolean nullToOptional = false;
-        if (dependencyType.equals(Optional.class)) {
+        boolean collection = false;
+        if (dependencyType.equals(Optional.class) || Collection.class.isAssignableFrom(dependencyType)) {
```
fsgonz: same commnet as before. I think this is duplicated.

elrodro83:
--validate

PR #8491, opened 2019-12-14
MULE-17828: fix tests

marianogonzalez:


core-tests/src/test/java/org/mule/runtime/core/internal/el/dataweave/DataWeaveExpressionLanguageAdaptorTestCase.java
```diff
@@ -649,7 +649,8 @@ public void compilationExceptionPropagates() {
   @Test
   public void evaluateInvalidCompiledExpression() throws MuleException {
     ExpressionCompilationException e = new ExpressionCompilationException(createStaticMessage("oopsy"));
-    expectedEx.expect(is(sameInstance(e)));
+    expectedEx.expect(ExpressionRuntimeException.class);
```
elrodro83: we should assert thet the error message is meaningful here

core/src/main/java/org/mule/runtime/core/internal/processor/simple/ParseTemplateProcessor.java
```diff
@@ -112,6 +116,8 @@ private MediaType buildMediaType() {
 
   @Override
   public CoreEvent process(CoreEvent event) {
+    evaluateCorrectArguments();
```
elrodro83: can't this be checked on initialize? or does it cause problems

marianogonzalez: it breaks backwards compatibility. THat's why the test broke

core-tests/src/test/java/org/mule/runtime/core/internal/el/dataweave/DataWeaveExpressionLanguageAdaptorTestCase.java
marianogonzalez: no point in doing that. I'm asserting that the thrown assertion is the same as in the mock

marianogonzalez:
--validate

PR #8493, opened 2019-12-16
MULE-17861: Multiple LifecycleExceptions causes OOM errors when running tests

lucianoRM:


tests/unit/src/main/java/org/mule/tck/junit4/AbstractMuleContextTestCase.java
```diff
@@ -179,6 +198,11 @@ public final void setUpMuleContext() throws Exception {
       }
 
       doSetUp();
+    } catch (LifecycleException e) {
+      //Set to null in order to clean references to the muleContext and avoid OOM errors.
```
marianogonzalez: format

tests/unit/src/main/java/org/mule/tck/junit4/AbstractMuleContextTestCase.java
```diff
@@ -179,6 +198,11 @@ public final void setUpMuleContext() throws Exception {
...
+      if (LIFECYCLE_EXCEPTION_COMPONENT_FIELD != null) {
+        LIFECYCLE_EXCEPTION_COMPONENT_FIELD.set(e, null);
```
marianogonzalez: I don't understand the change or the PR title is wrong. This just makes cleanup in case the tests fails, but does nothing to prevent it from failing. Why would it fail repetetadly in the first place?

lucianoRM: Yes, the title is wrong. I found this while looking for why that test was flaky but it has nothing to do with that particular test case. I will change the title.

tests/unit/src/main/java/org/mule/tck/junit4/AbstractMuleContextTestCase.java
lucianoRM: what's wrong with this?

afelisatti:
--validate

lucianoRM:
--validate

lucianoRM:
--validate

PR #8495, opened 2019-12-16
MULE-17877: Fix issue that caused files to be processed multiple times

giordanolucas:
RESTC-13804: Fix issue that caused files to be processed multiple times when the flow took long to execute (More than 1 minute).


ndinu:
Be careful with the format changes, if those are not approved by the formatter, you will break the build. Also, a MULE issue must be generated to reference in this PR.

giordanolucas:
--validate

PR #8500, opened 2019-12-17
MULE-17863: Avoid file leak in TlsConfiguration

eze210:


core/src/main/java/org/mule/runtime/core/privileged/security/tls/TlsConfiguration.java
```diff
@@ -346,18 +348,21 @@ private KeyStore createTrustStore() throws CreateException {
     trustStorePassword = null == trustStorePassword ? "" : trustStorePassword;
 
     KeyStore trustStore;
+    InputStream is = null;
 
     try {
       trustStore = KeyStore.getInstance(trustStoreType);
-      InputStream is = IOUtils.getResourceAsStream(trustStoreName, getClass());
+      is = IOUtils.getResourceAsStream(trustStoreName, getClass());
```
elrodro83: why not try-with-resources for this one?

eze210: Because if an IOException is thrown in the close, it would be caught in the line 337.
This way, in the finally I'm closing the InputStream quietly.
Do you think it would be better to throw that exception and catch it in the mentioned line?

PR #8505, opened 2019-12-18
MULE-17844: Remove `append-string-transformer`

elrodro83:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/CoreComponentBuildingDefinitionProvider.java
```diff
@@ -803,11 +802,6 @@ public void init() {
     transformerComponentBuildingDefinitions.add(getCoreTransformerBaseBuilder(AutoTransformer.class)
         .withIdentifier("auto-transformer")
         .build());
-    transformerComponentBuildingDefinitions.add(getCoreMuleMessageTransformerBaseBuilder()
-        .withIdentifier("append-string-transformer")
-        .withTypeDefinition(fromType(StringAppendTransformer.class))
```
marianogonzalez: shouldn't you delete the `StringAppendTransformer` class as well?

elrodro83: yes indeed

PR #8506, opened 2019-12-18
MULE-16129: Cannot deactivate streaming in operations that return lists/maps of streams

ndinu:


ndinu:
--validate

ndinu:
--validate

core/src/main/java/org/mule/runtime/core/api/streaming/DefaultStreamingManager.java
```diff
@@ -198,6 +204,32 @@ public void manage(InputStream stream, CoreEvent creatorEvent) {
...
+          && muleContext.getConfiguration().isInheritIterableRepeatability()) {
+        cursorStreamProviderFactory = forBytes().getNullCursorProviderFactory();
+      } else {
+        cursorStreamProviderFactory = forBytes().getDefaultCursorProviderFactory();
+      }
+    } else if (provider instanceof CursorStreamProviderFactory) {
+      cursorStreamProviderFactory = (CursorStreamProviderFactory) provider;
+      cursorIteratorProviderFactory = forObjects().getDefaultCursorProviderFactory();
+    } else {
+      cursorStreamProviderFactory = forBytes().getDefaultCursorProviderFactory();
```
marianogonzalez: IllegalArgumentException should be thrown here

core/src/main/java/org/mule/runtime/core/api/streaming/StreamingManager.java
```diff
@@ -103,4 +106,13 @@
    */
   @Deprecated
   void manage(InputStream inputStream, CoreEvent creatorEvent);
+
+  /**
+   * @param provider the provider for which you want the correspondent byte and object providers.
```
marianogonzalez: this jdoc is too vague. Here you should explain that this is for the case in which items in an iterable stream contain nested byte streams.

ndinu:
--validate

PR #8508, opened 2019-12-19
Mule 17880

facunlk:
negative maxEntries shouldn't throw an exception during expiration


facunlk:
--validate

facunlk:
--validate

PR #8513, opened 2019-12-19
MULE-17812: Refresh Dance not executed after TokenExpiredException

SebaElizalde:


SebaElizalde:
--validate

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OAuthOperationMessageProcessor.java
```diff
@@ -77,7 +76,7 @@ public OAuthOperationMessageProcessor(ExtensionModel extensionModel,
   @Override
   protected Mono<CoreEvent> doProcess(CoreEvent event, ExecutionContextAdapter<OperationModel> operationContext) {
     return super.doProcess(event, operationContext)
-        .onErrorResume(AccessTokenExpiredException.class, e -> {
+        .onErrorResume(Exception.class, e -> {
```
ndinu: I would add a test for this fix, where the AccessTokenExpiredException was the cause of another exception

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -503,6 +504,10 @@ public void doDispose() {
   }
 
   protected ExecutionMediator createExecutionMediator() {
+    if (executionMediatorFactory != null) {
+      return executionMediatorFactory.get();
+    }
+
     return new DefaultExecutionMediator(extensionModel, componentModel, connectionManager, muleContext.getErrorTypeRepository());
```
ndinu: What if you always use a supplier, but when creating the instance, you create the default supplier?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -594,4 +599,8 @@ private boolean isInterceptedComponent(ComponentLocation location, InternalEvent
     }
     return false;
   }
+
+  public void setExecutionMediatorFactory(Supplier<ExecutionMediator> executionMediatorFactory) {
+    this.executionMediatorFactory = executionMediatorFactory;
```
ndinu: Maybe move this to the constructor? this way you can make it final.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -152,6 +152,7 @@
   protected ComponentExecutor componentExecutor;
   protected ReturnDelegate returnDelegate;
   protected PolicyManager policyManager;
+  private Supplier<ExecutionMediator> executionMediatorFactory;
```
ndinu: I would call it executionMediatorSupplier

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
```diff
@@ -542,6 +542,10 @@ private void setCacheIdGenerator() {
 
   protected abstract ParameterValueResolver getParameterValueResolver();
 
+  public ConnectionManagerAdapter getConnectionManager() {
```
ndinu: Maybe we  should find a way not to expose this?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -503,6 +504,10 @@ public void doDispose() {
   }
 
   protected ExecutionMediator createExecutionMediator() {
+    if (executionMediatorFactory != null) {
```
marianogonzalez: remove this if. The factory should default to a supplier with the default behavior.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OAuthOperationMessageProcessor.java
SebaElizalde: Added in the integration tests.

PR #8524, opened 2019-12-26
MULE-17743: Remove the use of the eventContext ResponsePublisher

elrodro83:
Also, encapsulate the error handler from the flow, cleanup flow callers.


core/src/main/java/org/mule/runtime/core/internal/construct/DefaultFlowBuilder.java
```diff
@@ -221,10 +216,11 @@ public CoreEvent process(final CoreEvent event) throws MuleException {
     @Override
     public ReactiveProcessor referenced() {
       return pub -> from(pub)
+          .doOnNext(assertStarted())
           // Insert the incoming event into the flow, routing it through the processing strategy
-          .flatMap(routeThroughProcessingStrategy()
+          .flatMap(routeThroughProcessingStrategyMapper()
```
marianogonzalez: didn't you just deprecated `routeThroughProcessingStrategyMapper()`?

core/src/main/java/org/mule/runtime/core/internal/construct/DefaultFlowBuilder.java
```diff
@@ -234,7 +230,8 @@ public ReactiveProcessor referenced() {
     public Publisher<CoreEvent> apply(Publisher<CoreEvent> publisher) {
       return from(publisher)
           .doOnNext(assertStarted())
-          .flatMap(routeThroughProcessingStrategy())
+          // Insert the incoming event into the flow, routing it through the processing strategy
+          .flatMap(routeThroughProcessingStrategyMapper())
```
marianogonzalez: same

core/src/main/java/org/mule/runtime/core/internal/event/DefaultEventContext.java
```diff
@@ -167,7 +167,7 @@ public BaseEventContext getRootContext() {
    */
   public DefaultEventContext(FlowConstruct flow, ComponentLocation location, String correlationId,
                              Optional<CompletableFuture<Void>> externalCompletion) {
-    this(flow, flow.getExceptionListener(), location, correlationId, externalCompletion);
+    this(flow, NullExceptionHandler.getInstance(), location, correlationId, externalCompletion);
```
marianogonzalez: is there any place in which a non null exception handler is set? If the answer is no, then simply remove the concept from here.

core/src/main/java/org/mule/runtime/core/internal/construct/DefaultFlowBuilder.java
elrodro83: yep, but this usage is more tricky to migrate since it depends on how it is called. Will add a TODO for it

core/src/main/java/org/mule/runtime/core/internal/construct/DefaultFlowBuilder.java
elrodro83: same as previous comment

core/src/main/java/org/mule/runtime/core/internal/event/DefaultEventContext.java
elrodro83: good catch, more cleanup on the way

PR #8533, opened 2020-01-02
MULE-17874: SDK Should not reject dynamic metadata resolution when an…

Bardammu:
… optional key is null

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataKeyIdObjectResolver.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataKeyIdObjectResolver.java
@@ -262,6 +262,10 @@ final class MetadataKeyIdObjectResolver {
     return keyParts.isEmpty();
   }
 
+  public boolean isKeyRequired() {
+    return keyParts.stream().noneMatch(k -> k.isRequired() || k.getDefaultValue() != null);
+  }
+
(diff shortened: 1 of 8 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataKeyIdObjectResolver.java
```diff
@@ -262,6 +262,10 @@ public boolean isKeyLess() {
     return keyParts.isEmpty();
   }
 
+  public boolean isKeyRequired() {
+    return keyParts.stream().noneMatch(k -> k.isRequired() || k.getDefaultValue() != null);
```
ndinu: What if keyParts is empty? Will this say that the key is required?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataMediator.java
```diff
@@ -171,7 +171,7 @@ public MetadataMediator(T componentModel) {
       }
 
       keyValue = keyValueResult.get();
-      if (keyValue == null && !keyIdObjectResolver.isKeyLess()) {
+      if (keyValue == null && !keyIdObjectResolver.isKeyRequired()) {
```
ndinu: I think that this should fail if key is required. It looks like this will fail when key is not required.

PR #8538, opened 2020-01-03
MULE-17873: do not fail on runtime when metadata keys has no type res…

Bardammu:
…olver

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/MetadataComponentModelValidator.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/MetadataComponentModelValidator.java
@@ -106,7 +106,7 @@ public class MetadataComponentModelValidator implements ExtensionModelValidator 
         validateMetadataReturnType(extensionModel, model, problemsReporter);
         MetadataResolverFactory resolverFactory = getMetadataResolverFactory(model);
         validateMetadataOutputAttributes(model, resolverFactory, problemsReporter);
-        validateMetadataKeyId(model, resolverFactory, problemsReporter);
+        validateMetadataKeyId(extensionModel, model, resolverFactory, problemsReporter);
         validateCategoriesInScope(model, resolverFactory, problemsReporter);
         validateResolversName(model, resolverFactory, names, problemsReporter);
       }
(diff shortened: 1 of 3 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/MetadataComponentModelValidator.java
```diff
@@ -144,14 +144,15 @@ private void validateResolversName(ComponentModel model, MetadataResolverFactory
...
+  private void validateMetadataKeyId(ExtensionModel extensionModel, ComponentModel model, MetadataResolverFactory resolverFactory,
                                      ProblemsReporter problemsReporter) {
     final String modelTypeName = capitalize(getComponentModelTypeName(model));
     Optional<MetadataKeyIdModelProperty> keyId = model.getModelProperty(MetadataKeyIdModelProperty.class);
     if (keyId.isPresent()) {
 
       if (resolverFactory.getOutputResolver() instanceof NullMetadataResolver
-          && !thereIsAnInputTypeResolverDefined(getAllInputResolvers(model, resolverFactory))) {
+          && !thereIsAnInputTypeResolverDefined(getAllInputResolvers(model, resolverFactory))
+          && isCompileTime(extensionModel)) {
```
marianogonzalez: make this the very first condition so that mule startup doesn't go through unnecessary processing.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/MetadataComponentModelValidator.java
```diff
@@ -350,4 +351,8 @@ private boolean shouldValidateComponentOutputMetadata(ConnectableComponentModel
         .orElse(true);
   }
 
+  private boolean isCompileTime(ExtensionModel extensionModel) {
```
marianogonzalez: this looks like something we would do (or we potentially are doing) in several places. Move this to `org.mule.runtime.module.extension.api.util.MuleExtensionUtils` and check for other places where this code is repeated.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -474,4 +475,14 @@ public static boolean isExpression(Object value) {
     }
   }
 
+  /**
+   * Indicates if the extension model is being built on compile time.
+   *
+   * @param extensionModel the extension model to check
+   * @return a boolean indicating if the the extension model is being build on runtime or not
```
marianogonzalez: add `@since 4.3.0`

PR #8542, opened 2020-01-06
MULE-17917: Missing LegacyConnector class in the compatibility module

elrodro83:


core/src/main/java/org/mule/runtime/core/privileged/transport/LegacyConnector.java
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * @deprecated Transport infrastructure is deprecated.
+ */
+@Deprecated
+public interface LegacyConnector extends Lifecycle, NameableObject, Connectable, LifecycleStateEnabled {
+
+  /**
+   * Only use this method to use the Connector's MuleContext. Otherwise you can be used the wrong MuleContext because a Connector
+   * can be defined at the domain level or de app level.
```
eze210: the

PR #8546, opened 2020-01-07
MULE-17894: RETRY_EXHAUSTED ErrorType gets replaced by the ErrorType that generated the exhaustion (if present)

IvanAndresFritzler:
Removing the cause from the MessagingException prevents the ErrorType replacement during the MessagingException resolution

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/routing/UntilSuccessfulRouter.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/routing/UntilSuccessfulRouter.java
@@ -274,10 +274,12 @@ class UntilSuccessfulRouter {
       if (throwable instanceof MessagingException) {
         exceptionEvent = ((MessagingException) throwable).getEvent();
       }
+      // MessagingException cause is set to null in order to avoid RETRY:EXHAUSTED ErrorType replacement
+      // during ErrorType resolution (see ChainErrorHandlingUtils and MessagingExceptionResolver)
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/mule/runtime/core/internal/routing/UntilSuccessfulRouter.java
```diff
@@ -274,10 +274,12 @@ private CoreEvent eventWithCurrentContextDeleted(CoreEvent event) {
       if (throwable instanceof MessagingException) {
         exceptionEvent = ((MessagingException) throwable).getEvent();
       }
+      // MessagingException cause is set to null in order to avoid RETRY:EXHAUSTED ErrorType replacement
+      // during ErrorType resolution (see ChainErrorHandlingUtils and MessagingExceptionResolver)
       return new MessagingException(exceptionEvent,
                                     new RetryPolicyExhaustedException(createStaticMessage(UNTIL_SUCCESSFUL_MSG_PREFIX,
                                                                                           cause.getMessage()),
-                                                                      cause, owner),
+                                                                      null, owner),
```
elrodro83: should this be done internally by `RetryPolicyExhaustedException`? What other uses are there for it?

IvanAndresFritzler: Good point. I will check for the uses and make the change if possible.

PR #8553, opened 2020-01-08
MULE-17921: Support Platform Managed OAuth tokens

marianogonzalez:


modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/builder/OAuthPlatformManagedDancerBuilder.java
```diff
@@ -41,114 +31,15 @@
    */
   OAuthPlatformManagedDancerBuilder connectionId(String connectionId);
 
+  OAuthPlatformManagedDancerBuilder organizationId(String organizationId);
```
elrodro83: add javadoc

modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/builder/OAuthPlatformManagedDancerBuilder.java
```diff
@@ -41,114 +31,15 @@
...
+
+  OAuthPlatformManagedDancerBuilder platformUrl(String platformUrl);
```
elrodro83: add javadoc

PR #8554, opened 2020-01-08
MULE-17894: Reverting the changes due to backward compatibility issue

IvanAndresFritzler:
Mule allows to make assertions about exception causes and even access its fields. thus setting the RetryPolicyException cause to null cannot be used as a fix for MULE-17894


IvanAndresFritzler:
--validate

PR #8557, opened 2020-01-08
Mule 17527

JuLuzzi:


modules/http-api/src/main/java/org/mule/runtime/http/api/domain/CaseInsensitiveMultiMap.java
```diff
@@ -30,6 +32,12 @@
 
   private static final CaseInsensitiveMultiMap EMPTY_MAP = new CaseInsensitiveMultiMap().toImmutableMultiMap();
 
+  private final LazyValue<DataType> dataType = new LazyValue<>(() -> DataType.builder()
```
elrodro83: cannot this field be `static`? Also, why does this need to be lazy?

modules/http-api/src/main/java/org/mule/runtime/http/api/domain/message/HttpMessageBuilder.java
```diff
@@ -27,21 +28,18 @@
  */
 public abstract class HttpMessageBuilder<B extends HttpMessageBuilder, M extends HttpMessage> {
 
-  protected MultiMap<String, String> headers;
+  protected LazyValue<CaseInsensitiveMultiMap> headers = new LazyValue<>(this::initHeaders);
```
elrodro83: why does this need to be lazy?

modules/http-api/src/main/java/org/mule/runtime/http/api/domain/message/request/HttpRequestBuilder.java
```diff
@@ -36,7 +38,7 @@
   private HttpProtocol protocol = HTTP_1_1;
 
   HttpRequestBuilder(boolean preserveHeadersCase) {
-    headers = new CaseInsensitiveMultiMap(!preserveHeadersCase);
+    headers = new LazyValue<>(() -> new CaseInsensitiveMultiMap(!preserveHeadersCase));
```
elrodro83: why the lazy? this will eventually always be called when `build` is called

PR #8558, opened 2020-01-09
MULE-16764: Don't override item sequence info in parent event

szaffarano:
Fix a regression introduced in #MULE-16764


szaffarano:
--validation

aiannucci:
--validate

PR #8566, opened 2020-01-13
MULE-17939: Operation considers applied policies regarding thread switches

elrodro83:


core/src/main/java/org/mule/runtime/core/api/processor/strategy/ProcessingStrategy.java
```diff
@@ -48,6 +48,10 @@ default void registerInternalSink(Publisher<CoreEvent> flux, String sinkRepresen
     Flux.from(flux).subscribe();
   }
 
+  default Publisher<CoreEvent> registerInternalFlux(Publisher<CoreEvent> flux) {
```
marianogonzalez: jdoc

core/src/main/java/org/mule/runtime/core/internal/policy/DefaultPolicyManager.java
```diff
@@ -61,7 +61,7 @@
 
   private static final Logger LOGGER = LoggerFactory.getLogger(DefaultPolicyManager.class);
 
-  private static final OperationPolicy NO_POLICY_OPERATION =
+  public static final OperationPolicy NO_POLICY_OPERATION =
```
marianogonzalez: I would expose this through a getter

core/src/main/java/org/mule/runtime/core/internal/processor/AsyncDelegateMessageProcessor.java
```diff
@@ -192,15 +198,14 @@ public void start() throws MuleException {
   public void stop() throws MuleException {
     super.stop();
     stopIfNeeded(delegate);
-
+    stopIfNeeded(backpressureHandler);
```
marianogonzalez: this should be wrapped in FunctionalUtils.safely() so that if this fails, the rest is disposed correctly

core/src/main/java/org/mule/runtime/core/internal/processor/AsyncDelegateMessageProcessor.java
```diff
@@ -293,4 +309,57 @@ public void setMaxConcurrency(Integer maxConcurrency) {
...
+        try {
+          final CoreEvent event = asyncQueue.peek();
+          if (event != null) {
+            eventDispatcher.accept(event);
+            asyncQueue.remove(event);
+          }
+        } catch (FromFlowRejectedExecutionException free) {
+          // nothing to do, a retry will come next
+        }
+      }, 0, 2, MILLISECONDS);
```
marianogonzalez: isn't this too much frequency? 

Also:
a) isn't a busy wait preferable in cases of high throughput
b) isn't a wait()/notify() approach preferable in most cases?

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/AbstractProcessingStrategy.java
```diff
@@ -144,7 +144,7 @@ public final BackPressureReason emit(CoreEvent event) {
       long remainingCapacity = fluxSink.requestedFromDownstream();
       if (remainingCapacity == 0) {
         return EVENTS_ACCUMULATED;
-      } else if (remainingCapacity > (bufferSize > CORES * 4 ? CORES : 0)) {
+      } else if (remainingCapacity > (bufferSize > CORES * 2 ? CORES : 0)) {
```
marianogonzalez: weren't you going to actually increase the size of this buffer to reduce the error rate?

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/ProactorStreamEmitterProcessingStrategyFactory.java
```diff
@@ -161,6 +150,7 @@ public ReactiveProcessor onProcessor(ReactiveProcessor processor) {
       }
     }
 
+
```
marianogonzalez: remove

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/ProactorStreamEmitterProcessingStrategyFactory.java
```diff
@@ -174,10 +164,7 @@ protected ReactiveProcessor proactor(ReactiveProcessor processor, ScheduledExecu
         return publisher -> scheduleProcessor(processor, retryScheduler, from(publisher))
             .subscriberContext(ctx -> ctx.put(PROCESSOR_SCHEDULER_CONTEXT_KEY, scheduler));
       } else if (maxConcurrency == MAX_VALUE) {
-        if (processor.getProcessingType() == CPU_INTENSIVE
-            && (processor instanceof InterceptedReactiveProcessor)
-            && SDK_OPERATION_CLASS != null
-            && SDK_OPERATION_CLASS.isAssignableFrom(((InterceptedReactiveProcessor) processor).getProcessor().getClass())) {
+        if (processor instanceof OperationInnerProcessor) {
```
marianogonzalez: this is shady.. this is assuming SDK behavior by testing an interface that any component can implement, not just the SDK. Is this wrong or the comment below should be updated?

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/StreamEmitterProcessingStrategyFactory.java
```diff
@@ -206,6 +207,13 @@ public void registerInternalSink(Publisher<CoreEvent> flux, String sinkRepresent
       activeSinksCount.incrementAndGet();
     }
 
+    @Override
+    public Publisher<CoreEvent> registerInternalFlux(Publisher<CoreEvent> flux) {
```
marianogonzalez: This method name is not really appropriate.. It seems to be more something like configureFlux or something like this.... Actually, if this were about configuring something internal, then it shouldn't return anything

modules/extensions-spring-support/src/test/resources/heisenberg-stateful-operation-config.xml
```diff
@@ -6,7 +6,7 @@
                http://www.mulesoft.org/schema/mule/heisenberg http://www.mulesoft.org/schema/mule/heisenberg/current/mule-heisenberg.xsd">
 
     <heisenberg:config name="heisenberg"
-                       myName="#[mel:myName]"
+                       myName="#[vars.myName]"
```
marianogonzalez: 🐷 🔙

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/SdkInternalContext.java
```diff
@@ -0,0 +1,115 @@
...
+import org.mule.runtime.extension.api.runtime.config.ConfigurationInstance;
+import org.mule.runtime.extension.api.runtime.operation.CompletableComponentExecutor.ExecutorCallback;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Function;
+
+import reactor.util.context.Context;
+
+public class SdkInternalContext implements EventInternalContext<SdkInternalContext> {
```
marianogonzalez: jdoc

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/SdkInternalContext.java
```diff
@@ -0,0 +1,115 @@
...
+    return policyToApply;
+  }
+
+  public void setPolicyToApply(OperationPolicy policyToApply) {
+    this.policyToApply = policyToApply;
+  }
+
+  @Override
+  public SdkInternalContext copy() {
+    return this;
```
marianogonzalez: don't you need to copy the execution params? Is it ok to make a copy in which the callbacks are all the same?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -179,6 +197,15 @@
   private String resolvedProcessorRepresentation;
   private boolean initialised = false;
 
+  private ProcessingStrategy processingStrategy;
+  private boolean ownedProcessingStrategy = false;
+  private FluxSinkSupplier<CoreEvent> fluxSupplier;
+  private final Object fluxSupplierDisposeLock = new Object();
```
marianogonzalez: group all final fields together

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -179,6 +197,15 @@
...
+
+  private Scheduler outerFluxCompletionScheduler;
+
+  private final AtomicInteger activeOuterPublishersCount = new AtomicInteger(0);
```
marianogonzalez: group all final fields together

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -161,6 +178,7 @@
   private final RetryPolicyTemplate fallbackRetryPolicyTemplate = new NoRetryPolicyTemplate();
 
   protected final ExtensionModel extensionModel;
+  private final boolean hasChainNested;
```
marianogonzalez: hasNestedChain

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -212,73 +246,89 @@ public CoreEvent process(CoreEvent event) throws MuleException {
...
+    final BiFunction<Throwable, Object, Throwable> localOperatorErrorHook =
+        getLocalOperatorErrorHook(this, errorTypeLocator, exceptionContextProviders);
+
+    return subscriberContext()
+        .flatMapMany(ctx -> {
+          final FluxSinkRecorder<Either<Throwable, CoreEvent>> errorSwitchSinkSinkRef = new FluxSinkRecorder<>();
+
+          Flux<CoreEvent> transformed = from(propagateCompletion(from(publisher), create(errorSwitchSinkSinkRef)
+              .map(result -> {
+                return result.reduce(me -> {
```
marianogonzalez: add comments clarifying or extract to method with meaningful name, or both, or make a netflix documentary explaining this

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -212,73 +246,89 @@ public CoreEvent process(CoreEvent event) throws MuleException {
...
+                      }
+
+                      @Override
+                      public void complete(Object value) {
+                        errorSwitchSinkSinkRef.next(right(Throwable.class, (CoreEvent) value));
+                      }
+                    };
+
+                    if (!isAsync() && NO_POLICY_OPERATION
+                        .equals(((SdkInternalContext) ((InternalEvent) event).<SdkInternalContext>getSdkInternalContext())
```
marianogonzalez: turn this condition into a method in the SDK internal context

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -212,73 +246,89 @@ public CoreEvent process(CoreEvent event) throws MuleException {
...
+            // first event comes through
+            return transformed;
+          }
+        });
   }
 
   private void onEvent(CoreEvent event, ExecutorCallback executorCallback) {
     try {
+      SdkInternalContext sdkInternalContext =
+          ((SdkInternalContext) ((InternalEvent) event).<SdkInternalContext>getSdkInternalContext());
```
marianogonzalez: extract to static method in `org.mule.runtime.module.extension.api.util.MuleExtensionUtils`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -369,15 +416,120 @@ protected void doInitialise() throws InitialisationException {
...
+                                            ctx.get(POLICY_IS_PROPAGATE_MESSAGE_TRANSFORMATIONS));
+        }
+        return innerChainCtx;
+      });
+    }
+
+    sdkInternalContext.setConfiguration(resolveConfiguration(event));
+    final Map<String, Object> resolutionResult = getResolutionResult(event, sdkInternalContext.getConfiguration());
+    sdkInternalContext.setResolutionResult(resolutionResult);
+    sdkInternalContext.setPolicyToApply(getLocation() != null
```
marianogonzalez: where is this context cleared/replaced?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -666,8 +861,9 @@ protected ParameterValueResolver getParameterValueResolver() {
     }
   }
 
-  @Override
-  public abstract ProcessingType getProcessingType();
+  public ProcessingType getInnerProcessingType() {
+    return CPU_LITE;
```
marianogonzalez: add a clarifying comment about the inner chain processing type being the one that actually reflects the operation

pom.xml
```diff
@@ -2004,6 +2007,13 @@
             <properties>
                 <vmtype>org.eclipse.jdt.internal.launching.macosx.MacOSXType</vmtype>
             </properties>
+        </profile>
+        <profile>
+            <id>yourkit-agent</id>
+            <properties>
+<!--                 <yourkit.agent.path>/Applications/YourKit-Java-Profiler-2019.8.app/Contents/Resources/bin/mac/libyjpagent.dylib</yourkit.agent.path> -->
```
marianogonzalez: remove comment or make it multiline and explain when/how to enable

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/AbstractProcessingStrategy.java
elrodro83: no need to do that, this remained of a test i made, will revert it

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/ProactorStreamEmitterProcessingStrategyFactory.java
elrodro83: `OperationInnerProcessor` is not API

modules/extensions-spring-support/src/test/resources/heisenberg-stateful-operation-config.xml
elrodro83: it has to happen sometime...

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/SdkInternalContext.java
elrodro83: this is the most consistent with the current behavior, where all internal parameters are just copied over.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
elrodro83: this is create above in this method. Will add a clear for it

core/src/main/java/org/mule/runtime/core/api/processor/strategy/ProcessingStrategy.java
```diff
@@ -48,7 +48,14 @@ default void registerInternalSink(Publisher<CoreEvent> flux, String sinkRepresen
     Flux.from(flux).subscribe();
   }
 
-  default Publisher<CoreEvent> registerInternalFlux(Publisher<CoreEvent> flux) {
+  /**
+   * For fluxes created outside of the main flux for a flow, have them accounted for in the processing strategy for a graceful
```
marianogonzalez: is graceful shutdown a particular use case of this method or is the only and sole purpose of it? Depending on that, either change the jdoc or rename to something more specific

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/ProactorStreamEmitterProcessingStrategyFactory.java
marianogonzalez: but still, some other dev which doesn't have the full context could implement that interface and a problem could appear here

core/src/main/java/org/mule/runtime/core/internal/processor/AsyncDelegateMessageProcessor.java
```diff
@@ -335,22 +341,36 @@ public void handleBackpressure(CoreEvent event) {
...
+                  }
+                } catch (FromFlowRejectedExecutionException free) {
+                  synchronized (this) {
+                    scheduledDrain = null;
+                  }
+                  // nothing to do, a retry will come next
+                  return;
+                }
+              }
+            }, 200, MILLISECONDS);
```
marianogonzalez: why do we still need to keep this polling every 200 ms? If this tasks gets scheduled because of a brief high demand lapse, then this task will continue to run every 200ms without need. Can't we use a traditional single threaded wait/notify()?

elrodro83: the thing is that the wait/notify mechanism will still keep a thread busy waiting for elements to arrive at the queue.
With this approach, this task is active only while it makes sense to poll the queue.

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/ProactorStreamEmitterProcessingStrategyFactory.java
elrodro83: will add a note on the jdoc for `OperationInnerProcessor `

core/src/main/java/org/mule/runtime/core/internal/processor/AsyncDelegateMessageProcessor.java
```diff
@@ -318,67 +317,61 @@ public void setMaxConcurrency(Integer maxConcurrency) {
...
+            currentThread().interrupt();
+            return;
           }
         }
-      }
+      });
     }
 
     @Override
     public void start() {
```
marianogonzalez: delete method then. Doesn't look like you're actually forced to implement Startable

PR #8573, opened 2020-01-14
MULE-17948: Add the concepts of connections URI and Descriptor on OCS

marianogonzalez:


modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/PlatformManagedConnectionDescriptor.java
```diff
@@ -0,0 +1,49 @@
...
+  String getUri();
+
+  /**
+   * @return The name that the user has given this connection in the Anypoint Platform
+   */
+  String getName();
+
+  /**
+   * Returns parameters additional to the standard OAuth ones that the user has configured for this connection. These usually
+   * correspond to custom parameters tha the service provider accepts or requires on its requests, but may contain other
```
elrodro83: `tha the` -> `that the`

modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/PlatformManagedConnectionDescriptor.java
```diff
@@ -0,0 +1,49 @@
...
+   * parameters as well.
+   *
+   * @return A {@link Map} which keys are the parameter names
+   */
+  Map<String, Object> getParameters();
```
elrodro83: parameters is to generic. Can this be renamed to `getCustomParameters` or `getUserParameters`?

marianogonzalez: it's actually intentional. I didn't want it to be too specific as this might have unknown uses in the future. It's the same logic I asked you to follow yesterday on your PR's. If I change it as you ask, I not only risk having an API naming problem in the future, I also risk you asking me the same things I ask of you yesterday.

PR #8578, opened 2020-01-14
MULE-17435: Reproduce flaky error 'Cannot load Extension class...' in jenkins jobs

lucianoRM:


tests/runner/src/main/java/org/mule/test/runner/api/DependencyResolver.java
```diff
@@ -140,7 +136,7 @@ public ArtifactDescriptorResult readArtifactDescriptor(Artifact artifact, List<R
   public ArtifactResult resolveArtifact(Artifact artifact) throws ArtifactResolutionException {
     checkNotNull(artifact, "artifact cannot be null");
 
-    final ArtifactRequest request = new ArtifactRequest(artifact, resolutionContext.getRemoteRepositories(), null);
+    final ArtifactRequest request = new ArtifactRequest(artifact, resolveRepositories(), null);
```
gsfernandes: Should this be called with resolveRepositories(resolutionContext.getRemoteRepositories())

PR #8582, opened 2020-01-14
Don't override BOP info

juanchib:


fsgonz:
@juanchib, I'll close this one and reopen one with changes in the test so that this is merged to mule-3.x and then cherry-picked to the other dev branches.

PR #8583, opened 2020-01-15
MULE-17952: Race condition when cloning a DefaultFlowCallStack

elrodro83:


core/src/main/java/org/mule/runtime/core/internal/context/notification/DefaultFlowCallStack.java
```diff
@@ -118,10 +132,12 @@ private String doToString(Function<FlowStackElement, String> toString) {
     StringBuilder stackString = new StringBuilder();
 
     int i = 0;
-    for (FlowStackElement flowStackElement : innerStack) {
-      stackString.append("at ").append(toString.apply(flowStackElement));
-      if (++i != innerStack.size()) {
-        stackString.append(lineSeparator());
+    synchronized (innerStack) {
+      for (FlowStackElement flowStackElement : innerStack) {
```
fsgonz: can we create a Utils class for this as ExceptionUtils but for the flowStack?

fsgonz: Ok, we keep this in the doToString.

PR #8584, opened 2020-01-15
MULE-17894: RETRY_EXHAUSTED ErrorType gets replaced by the ErrorType that caused the exhaustion

IvanAndresFritzler:
Error handling resolves the MessagingExceptions by looking for the root cause (first Mule exception in the cause tree) and getting its ErrorType, unless the MessagingException event has an ErrorType defined. This fix involves the creation of a new CoreEvent that represents the retry exhausted error and has its Error set to MULE:RETRY_EXHAUSTED.


IvanAndresFritzler:
--validate

PR #8592, opened 2020-01-16
MULE-17874: SDK Should not reject dynamic metadata resolution when an optional key is null

Bardammu:


Bardammu:
--validate

Bardammu:
For the records, SonarQube fails because of an uncovered line on the connector test, so we can dismiss the warning

PR #8594, opened 2020-01-17
MULE-17975: Paging Provider is closed twice when it returns an empty page on the first call to getPage.

SebaElizalde:


Bardammu:
Where is the test?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/streaming/PagingProviderWrapper.java
```diff
@@ -47,8 +48,10 @@ public PagingProviderWrapper(PagingProvider<C, T> delegate, ExtensionModel exten
    */
   @Override
   public void close(C connection) throws MuleException {
-    closed = true;
-    delegate.close(connection);
+    if (!closed.get()) {
+      closed.set(true);
```
Bardammu: You have yo use compareAndSet

SebaElizalde: nice catch

PR #8600, opened 2020-01-22
MULE-17935: add additional property

ndinu:


modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/builder/OAuthPlatformManagedDancerBuilder.java
```diff
@@ -47,6 +47,14 @@
    */
   OAuthPlatformManagedDancerBuilder platformUrl(String platformUrl);
 
+  /**
+   * Sets the environment identifier that needs to be connected to.
+   *
+   * @param environmentId and environment Id
```
Bardammu: environment id and environment id and environment id.

ndinu:
--validate

PR #8619, opened 2020-01-31
MULE-18019: Fix reconnection for Paged Operations on the first call to getPage.

SebaElizalde:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/streaming/PagingProviderProducer.java
```diff
@@ -77,16 +87,19 @@ public int getSize() {
...
-        connectionSupplier.close();
+      R result = function.apply(connection);
+      connectionSupplier.close();
+      return result;
+    } catch (Exception e) {
+      if (isFirstPage) {
+        closeDelegate(connection);
       }
+      extractConnectionException(e).ifPresent(ex -> connectionSupplier.invalidateConnection());
+      connectionSupplier.invalidateConnection();
```
gabrieldalborgo: Remove this

SebaElizalde: thanks, fixed

PR #8626, opened 2020-02-04
MULE-18030: HttpPolicyPointcutParameters, only use the headers if a

elrodro83:
pointcut actually needs them


core/src/main/java/org/mule/runtime/core/api/policy/PolicyProvider.java
```diff
@@ -29,6 +29,22 @@
    */
   List<Policy> findSourceParameterizedPolicies(PolicyPointcutParameters policyPointcutParameters);
 
+  /**
+   * @return {@code true} if there is at least one source policy applied that relies on the headers to check for the pointcut
+   *         match.
+   *
+   * @since 4.3
+   */
+  boolean isAnySourcePolicyHeadersAware();
```
balbifm: This is coupling HTTP to the policy engine. It should probably be in http-policy-api

core/src/main/java/org/mule/runtime/core/internal/policy/DefaultPolicyManager.java
```diff
@@ -200,6 +200,11 @@ public SourcePolicy createSourcePolicyInstance(Component source, CoreEvent sourc
     return sourcePolicy;
   }
 
+  @Override
+  public boolean isAnySourcePolicyHeadersAware() {
```
balbifm: I got lost, where is this used?

elrodro83: removed this

core/src/main/java/org/mule/runtime/core/api/policy/PolicyProvider.java
elrodro83: yep, fixed

PR #8633, opened 2020-02-06
MULE-18040: Scatter-Gather router does not honor route order in

elrodro83:
aggregated response hashmap when there are 12 or more routes defined


elrodro83:
--validate

PR #8649, opened 2020-02-13
MULE-18059: Flow in initial state stopped is started when app is stopped

elrodro83:
and started


core/src/main/java/org/mule/runtime/core/internal/construct/AbstractFlowConstruct.java
```diff
@@ -109,7 +109,7 @@ public final void initialise() throws InitialisationException {
   @Override
   public final void start() throws MuleException {
     // Check if Initial State is Stopped
-    if (!isStopped() && initialState.equals(INITIAL_STATE_STOPPED)) {
+    if (muleContext.isStarting() && initialState.equals(INITIAL_STATE_STOPPED)) {
```
eze210: previous comment doesn't match with code

eze210: it's a joke

PR #8653, opened 2020-02-17
MULE-18072: Improve performance of repeatable streaming

marianogonzalez:


core/src/main/java/org/mule/runtime/core/internal/streaming/AtomicStreamingStatistics.java
```diff
@@ -38,6 +38,11 @@ public int decrementOpenCursors() {
     return openCursors.decrementAndGet();
   }
 
+  @Override
+  public int decrementOpenCursors(int howMany) {
+    return openCursors.addAndGet(-howMany);
```
elrodro83: isn't there a decrementAndGet thar recives the param?

marianogonzalez: no. Decrement and increment is on a 1:1 basis

PR #8655, opened 2020-02-17
MULE-18063: Need a way to stop built-in schedule for CH apps

szaffarano:


core-tests/src/test/java/org/mule/runtime/core/internal/source/polling/DefaultSchedulerMessageSourceTestCase.java
```diff
@@ -75,48 +89,51 @@ public void setUp() throws Exception {
...
   }
 
   @Test
-  public void disposeScheduler() throws Exception {
-    SchedulerService schedulerService = muleContext.getSchedulerService();
-    reset(schedulerService);
+  @Description("Verifies that no events are processed when the scheduler is disabled")
+  @Issue("MULE-18063")
+  public void simplePollDisabledWithScheduler() throws Exception {
+    String wasDisabled = setProperty(MULE_DISABLE_SCHEDULERS, "true");
```
elrodro83: make this test parameterized and use a rule for the system property

core/src/main/java/org/mule/runtime/core/internal/source/scheduler/DefaultSchedulerMessageSource.java
```diff
@@ -63,6 +66,7 @@
 
   private final PeriodicScheduler scheduler;
   private final boolean disallowConcurrentExecution;
+  private final boolean disabled;
```
elrodro83: make this static and read the value on initialization

core-tests/src/test/java/org/mule/runtime/core/internal/source/polling/DefaultSchedulerMessageSourceTestCase.java
szaffarano: Done.  I split this TestCase in two to make one of them parameterized.

PR #8666, opened 2020-02-20
MULE-18089: Flow source stopped event if it was not started because of a

elrodro83:
ComponentInitialStateManager


anosenzo:
--validate

PR #8669, opened 2020-02-21
MULE-18098: Support backward version of ObjectStore Stereotype to avo…

gsfernandes:
…id compatibility issues


gsfernandes:
--validate

PR #8673, opened 2020-02-25
MULE-16933: Increase coverage for processing strategy dispose in unit

elrodro83:
* Addendum for tx


core/src/main/java/org/mule/runtime/core/internal/processor/strategy/StreamPerThreadSink.java
```diff
@@ -92,18 +97,29 @@ public void dispose() {
     long startMillis = currentTimeMillis();
 
     while (!sinks.asMap().isEmpty()
-        && currentTimeMillis() - startMillis > shutdownTimeout
```
elrodro83: yeah, yeah, i know...

elrodro83:
--validate

PR #8679, opened 2020-02-25
MULE-18047: Add the component that generated a cursor provider

szaffarano:


core/src/main/java/org/mule/runtime/core/internal/streaming/CursorProviderAlreadyClosedException.java
```diff
@@ -0,0 +1,63 @@
...
+  private static String resolveMessage(String message, Optional<ComponentLocation> originatingLocation,
+                                       Optional<Exception> closerResponsible) {
+    String openedByDescription = format("The cursor provider was open by %s", originatingLocation.map(cl -> cl.getLocation())
+        .orElse("unknown"));
+
+    String responsibleDescription = closerResponsible.map(r -> {
+      StringWriter stringWriter = new StringWriter();
+      r.printStackTrace(new PrintWriter(stringWriter));
+      return format("The cursor provider was closed by: \n %s", stringWriter.toString());
+    }).orElse(format("Run with -M-D%s=true for more details", TRACK_CURSOR_PROVIDER_CLOSE));
```
elrodro83: mention just setting a property. This is valid for command line, but in CH or runtime fabric these are set differently

szaffarano: Do you mean that we should provide an alternative method to enable this feature?

elrodro83: no, just saying that there are other ways of sessing system properties other than command line

core/src/main/java/org/mule/runtime/core/api/streaming/bytes/AbstractCursorStreamProvider.java
```diff
@@ -51,6 +80,9 @@ public final CursorStream openCursor() {
    */
   @Override
   public void close() {
+    if (TRACK_CURSOR_PROVIDER_CLOSE) {
```
marianogonzalez: do this AFTER setting the closed flag

core/src/main/java/org/mule/runtime/core/api/streaming/bytes/InMemoryCursorStreamProvider.java
```diff
@@ -28,14 +29,34 @@
...
   }
 
+  /**
+   * Creates a new instance
+   *
+   * @param wrappedStream the stream to buffer from
+   * @param config        the config of the generated buffer
+   * @param bufferManager the {@link ByteBufferManager} that will be used to allocate all buffers
+   * 
+   * @deprecated Please use
```
marianogonzalez: add the version in which it was deprecated

core/src/main/java/org/mule/runtime/core/api/streaming/bytes/AbstractCursorStreamProvider.java
```diff
@@ -28,21 +33,45 @@
 
   private final AtomicBoolean closed = new AtomicBoolean(false);
 
+  private final ComponentLocation originatingLocation;
+
+  private Exception closerResponsible;
+
+  private static final boolean TRACK_CURSOR_PROVIDER_CLOSE = getBoolean(TRACK_CURSOR_PROVIDER_CLOSE_PROPERTY);
```
marianogonzalez: the SP should not be consumed here. This should be just a final instance attribute provided by the constructor. Otherwise, we don't only loose the ability to spot enable it, it also requires ugly reflection on the test cases. The SystemProperty should be consumed on the AbstractCursorProviderFactory

core/src/main/java/org/mule/runtime/core/internal/streaming/CursorProviderAlreadyClosedException.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.Optional;
+
+import org.mule.runtime.api.component.location.ComponentLocation;
+
+/**
+ * This {@link RuntimeException} is thrown when it is trying to open a closed
+ * {@link org.mule.runtime.api.streaming.CursorProvider}. It can contain optionally the
```
marianogonzalez: add imports

core/src/main/java/org/mule/runtime/core/internal/streaming/object/AbstractCursorIteratorProvider.java
```diff
@@ -23,22 +28,45 @@
 
   protected final Iterator stream;
   private final AtomicBoolean closed = new AtomicBoolean(false);
+  private final ComponentLocation originatingLocation;
+  private Exception closerResponsible;
+
+  private static final boolean TRACK_CURSOR_PROVIDER_CLOSE = getBoolean(TRACK_CURSOR_PROVIDER_CLOSE_PROPERTY);
```
marianogonzalez: same as before. The SP should be consumed in one unique spot and then propagated through constructors

core/src/main/java/org/mule/runtime/core/internal/streaming/object/AbstractCursorIteratorProvider.java
```diff
@@ -47,6 +75,9 @@ public final CursorIterator openCursor() {
    */
   @Override
   public void close() {
+    if (TRACK_CURSOR_PROVIDER_CLOSE) {
```
marianogonzalez: invert

core/src/main/java/org/mule/runtime/core/internal/streaming/object/InMemoryCursorIteratorProvider.java
```diff
@@ -26,13 +27,27 @@
    *
    * @param stream the stream to buffer from
    * @param config        the config of the generated buffer
+   * @param originatingLocation indicates where the cursor was created
    */
-  public InMemoryCursorIteratorProvider(Iterator stream, InMemoryCursorIteratorConfig config) {
-    super(stream);
+  public InMemoryCursorIteratorProvider(Iterator stream, InMemoryCursorIteratorConfig config,
```
marianogonzalez: `@since`

core/src/main/java/org/mule/runtime/core/internal/streaming/CursorProviderAlreadyClosedException.java
szaffarano: I changed the message to a more generic one.

core/src/main/java/org/mule/runtime/core/api/streaming/bytes/factory/AbstractCursorStreamProviderFactory.java
```diff
@@ -34,6 +36,7 @@
 
   protected final ByteBufferManager bufferManager;
   protected final StreamingManager streamingManager;
+  protected final boolean trackCursorProviderClose;
```
marianogonzalez: static

core-tests/src/test/java/org/mule/runtime/core/internal/streaming/AbstractTroubleshootCursorProviderTestCase.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    cursorProvider.close();
+
+    cursorProvider.openCursor();
+  }
+
+  protected abstract CursorProvider createCursorProvider();
+
+  private void setStaticField(Class cls, String fieldName, Object value) throws NoSuchFieldException, IllegalAccessException {
```
marianogonzalez: unused

core/src/main/java/org/mule/runtime/core/api/streaming/bytes/factory/AbstractCursorStreamProviderFactory.java
```diff
@@ -36,7 +36,7 @@
 
   protected final ByteBufferManager bufferManager;
   protected final StreamingManager streamingManager;
-  protected final boolean trackCursorProviderClose;
+  protected static boolean trackCursorProviderClose;
```
marianogonzalez: don't just make it static... resolve the value statically.. the original comment aims to not resolving the SP each time the factory is created

core/src/main/java/org/mule/runtime/core/internal/streaming/object/factory/AbstractCursorIteratorProviderFactory.java
```diff
@@ -32,11 +32,10 @@
 public abstract class AbstractCursorIteratorProviderFactory implements CursorIteratorProviderFactory {
 
   protected final StreamingManager streamingManager;
-  protected final boolean trackCursorProviderClose;
+  protected final static boolean trackCursorProviderClose = getBoolean(TRACK_CURSOR_PROVIDER_CLOSE_PROPERTY);
```
marianogonzalez: I don't understand how this test can pass. First parametrization runs fine, but for the second one, this value is not updated.

To avoid reflection hackerism, you can add a `setCursorProviderCloseTracking(boolean enabled)` method on the abstract factories (not in the API interface)

szaffarano:
--validate

PR #8680, opened 2020-02-26
MULE-18102: Concurrent flow reference instantiation issue

IvanAndresFritzler:
FlowRefFactoryBean relies on other beans in order to instantiate a flow. In the case of a subflow, SpringConfigurationComponentLocator is unable to handle a concurrent instantiation.
I tried to minimize as much as possible the amount of code that has to be synchronized in order to avoid the location overlaps.

### DIFF
--- a/modules/spring-config/src/main/java/org/mule/runtime/config/internal/factories/FlowRefFactoryBean.java
+++ b/modules/spring-config/src/main/java/org/mule/runtime/config/internal/factories/FlowRefFactoryBean.java
@@ -100,6 +100,8 @@ public class FlowRefFactoryBean extends AbstractComponentFactory<Processor> impl
   @Inject
   private ConfigurationComponentLocator locator;
 
+  private static final Object referencedFlowLock = new Object();
+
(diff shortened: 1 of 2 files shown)


modules/spring-config/src/main/java/org/mule/runtime/config/internal/factories/FlowRefFactoryBean.java
```diff
@@ -100,6 +107,8 @@
   @Inject
   private ConfigurationComponentLocator locator;
 
+  private static final Object referencedProcessorLock = new Object();
```
elrodro83: this should not be final, because in that case multiple applications in the same runtime may be synchronized with each other needlessly. Better get the lock from the MuleContext (there already are some app level locks there.)

IvanAndresFritzler: I decided to lock on the ApplicationContext instance and it's working, sounds ok for you?

PR #8699, opened 2020-03-03
MULE-18138: Every time a connection is created/autowired, its class is

elrodro83:
looked up in the app classloader

Piggyback:

* Make loggers static fields so the logger is not created on each instance created
* Improve synchronization of queues infrastructure


core/src/main/java/org/mule/runtime/core/internal/util/queue/TransactionalQueueManager.java
```diff
@@ -31,23 +31,24 @@
...
-  private Map<String, RecoverableQueueStore> queuesAccessedForRecovery = new HashMap<>();
+  private final Map<String, RecoverableQueueStore> queuesAccessedForRecovery = new HashMap<>();
 
   /**
    * {@inheritDoc}
    *
    * @return an instance of {@link TransactionalQueueSession}
    */
   @Override
-  public synchronized QueueSession getQueueSession() {
```
marianogonzalez: nice

PR #8710, opened 2020-03-05
MULE-18148: Memory leak on transactional try with error handlers

elrodro83:


core/src/main/java/org/mule/runtime/core/internal/exception/ExceptionRouter.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * @since 4.3
+ */
+public interface ExceptionRouter extends Consumer<Exception>, Disposable {
+
+  @Override
+  void accept(Exception t);
+
+  @Override
+  void dispose();
```
fsgonz: Is this method been overridden because of clarity?

elrodro83: yeah, i will remove both methods, since they are inherited

PR #8718, opened 2020-03-06
MULE-18157: Event modifications lost when setting a securityContext

elrodro83:


elrodro83:
--validate

elrodro83:
--validate

PR #8721, opened 2020-03-09
MULE-18164: Fix policy fails to deploy when async scope is being used

balbifm:


balbifm:
This is already (and unintentionally) fixed in 4.3

modules/deployment/src/test/resources/policy-using-async-scope.xml
```diff
@@ -0,0 +1,20 @@
...
+<mule xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xmlns:test-policy="http://www.mulesoft.org/schema/mule/test-policy"
+      xmlns="http://www.mulesoft.org/schema/mule/core"
+      xsi:schemaLocation="http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd
+               http://www.mulesoft.org/schema/mule/test-policy http://www.mulesoft.org/schema/mule/test-policy/current/mule-test-policy.xsd">
+
+    <test-policy:proxy name="async-policy">
+        <test-policy:operation>
+            <async>
+                <logger level="INFO" message="#[mel:new org.foo.EchoTest()]"/>
```
elrodro83: do not use mel, use DW

PR #8722, opened 2020-03-10
EE-7188: Mule license checker is not writing to STDERR

szaffarano:


modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -96,6 +96,9 @@
   private static MuleLog4jContextFactory log4jContextFactory;
 
   static {
+    // Before doing any further action, we need to check for a valid license
+    validateLicense();
```
marianogonzalez: this is effectively the same as reverting MULE-15508

szaffarano: The reason because I changed this behaviour was that if we call `validateLicense()` from `MuleContainer::start`, when `MuleContainer` is instanced, `artifactResourcesRegistry` instance variable [will be built](https://github.com/mulesoft/mule/blob/mule-4.x/modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java#L95), which indirectly calls `MuleEmbeddedLicenseCheck.checkLicense();` because of the aspect defined in [LicenseCheckAspect.aj](https://github.com/mulesoft/mule-ee/blob/mule-4.x/core/src/main/java/com/mulesoft/mule/runtime/core/internal/config/license/LicenseCheckAspect.aj).  
In short, the validation will happen earlier in another place, and the error message will be a `RuntimeException`.
We could defer the `validateLicense()` invocation until the MuleContainer constructor, but it still will block the start of the runtime until the validation finishes.  What do you think?

PR #8727, opened 2020-03-11
OCS add system property for setting platfotm authentication url suffix

gabrieldalborgo:


modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/connectivity/oauth/ocs/PlatformManagedOAuthConfigTestCase.java
```diff
@@ -0,0 +1,61 @@
...
+
+@SmallTest
+@RunWith(MockitoJUnitRunner.class)
+public class PlatformManagedOAuthConfigTestCase extends AbstractMuleTestCase {
+
+  private static final String CLIENT_ID = "a";
+  private static final String SECRET_ID = "b";
+  private static final String ORG_ID = "b";
+  private static final String SERVICE_URL = "c";
+  private static final String PLATFORM_URL = "http://localhost/accounts";
```
marianogonzalez: PLATFORM_URL -> PLATFORM_AUTH_URL

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/oauth/ocs/PlatformManagedOAuthConfig.java
```diff
@@ -33,6 +34,8 @@
  */
 public class PlatformManagedOAuthConfig extends OAuthConfig<PlatformManagedOAuthGrantType> {
 
+  private static final String DEFAULT_PLATFORM_AUTH_URL_SUFFIX = "/oauth2/token";
```
marianogonzalez: should be defined in the same place as `OCS_PLATFORM_AUTH_URL` is

core/src/main/java/org/mule/runtime/core/api/util/StringUtils.java
```diff
@@ -169,5 +169,17 @@ public static String trim(String str) {
     return org.apache.commons.lang3.StringUtils.trim(str);
   }
 
+  /**
+   * Remove trailing slash from {@code url}
+   *
+   * @param url    an url
```
marianogonzalez: `add @since 4.3.0`

PR #8738, opened 2020-03-16
MULE-18191: mule4 failed to stream rest of payload to client

szaffarano:


modules/http-api/src/main/java/org/mule/runtime/http/api/exception/ConnectionAlreadyClosedException.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.io.IOException;
+
+/**
+ * This Exception is thrown when there is an error that implies losing the connection with the client.
+ * 
+ * @since 4.4.0
+ */
+public class ConnectionAlreadyClosedException extends IOException {
+
```
elrodro83: serialversionuuid

modules/http-api/src/main/java/org/mule/runtime/http/api/exception/ConnectionAlreadyClosedException.java
```diff
@@ -0,0 +1,26 @@
...
+  public ConnectionAlreadyClosedException() {
```
elrodro83: is this usage without a cause valid? based on what we chatted on slack i would assume not.

elrodro83:
and use a MULE ticket for this :)

modules/http-api/src/main/java/org/mule/runtime/http/api/exception/ConnectionAlreadyClosedException.java
szaffarano: done

modules/http-api/src/main/java/org/mule/runtime/http/api/exception/ConnectionAlreadyClosedException.java
szaffarano: It can be used without a cause, because the connector only asks for the exception type.  But besides that, I'm setting the cause in the service, so you are right, I've already removed this constructor.

modules/http-api/src/main/java/org/mule/runtime/http/api/exception/ConnectionAlreadyClosedException.java
```diff
@@ -0,0 +1,24 @@
...
+package org.mule.runtime.http.api.exception;
+
+import java.io.IOException;
+
+/**
+ * This Exception is thrown when there is an error that implies losing the connection with the client.
+ * 
+ * @since 4.4.0
+ */
+public class ConnectionAlreadyClosedException extends IOException {
```
elrodro83: per the comment on https://github.com/mulesoft/mule-http-connector/pull/432#discussion_r394339065, this should be general and not just part of the http api

szaffarano: Do you mean that this class should be in mule-api?

szaffarano:
--validate

PR #8739, opened 2020-03-16
MULE-18189: Race condition on sdk operations in scatter-gather routes

elrodro83:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -378,20 +378,20 @@ private void onEvent(CoreEvent event, ExecutorCallback executorCallback) {
...
 
         fluxSupplier.get().next(operationEvent);
       };
 
       if (getLocation() != null) {
         ((DefaultFlowCallStack) event.getFlowCallStack())
             .setCurrentProcessorPath(resolvedProcessorRepresentation);
-        sdkInternalContext.getPolicyToApply().process(event, operationExecutionFunction, () -> resolutionResult, getLocation(),
-                                                      executorCallback);
+        sdkInternalContext.getPolicyToApply(getLocation())
```
marianogonzalez: IIRC, getLocation() was kind of expensive. Since already invoked in line 390, I would extract to a variable

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -378,20 +378,20 @@ private void onEvent(CoreEvent event, ExecutorCallback executorCallback) {
...
+            .process(event, operationExecutionFunction, () -> resolutionResult, getLocation(), executorCallback);
```
marianogonzalez: what if not present? Is NPE completely impossible here?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/SdkInternalContext.java
```diff
@@ -46,62 +46,63 @@ public static SdkInternalContext from(CoreEvent event) {
    * SDK components may be nested within each other, so some of the context must be kept separately for the component it belongs
    * to.
    */
-  private final Deque<LocationSpecificSdkInternalContext> locationSpecificContext = new ArrayDeque<>();
+  private final Map<ComponentLocation, LocationSpecificSdkInternalContext> locationSpecificContext = new SmallMap<>();
```
marianogonzalez: are you sure that `ComponentLocatin#equals()` is fast enough for `SmallMap` ?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -405,11 +405,11 @@ private void onEventSynchronous(CoreEvent event, ExecutorCallback executorCallba
...
       SdkInternalContext sdkInternalContext = from(event);
 
-      final Optional<ConfigurationInstance> configuration = sdkInternalContext.getConfiguration();
-      final Map<String, Object> resolutionResult = sdkInternalContext.getResolutionResult();
+      final Optional<ConfigurationInstance> configuration = sdkInternalContext.getConfiguration(getLocation());
+      final Map<String, Object> resolutionResult = sdkInternalContext.getResolutionResult(getLocation());
 
       OperationExecutionFunction operationExecutionFunction = (parameters, operationEvent, callback) -> {
-        sdkInternalContext.setOperationExecutionParams(configuration, parameters, operationEvent, callback);
+        sdkInternalContext.setOperationExecutionParams(getLocation(), configuration, parameters, operationEvent, callback);
```
marianogonzalez: same thing about getLocation(), here and in all the other occurrences below

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/execution/SdkInternalContextTestCase.java
```diff
@@ -0,0 +1,70 @@
...
+
+    final ComponentLocation comp1 = fromSingleComponent("comp1");
+    final ComponentLocation comp2 = fromSingleComponent("comp2");
+
+    final List<ComponentLocation> completedForComponents = new ArrayList<>();
+
+    pushContext(ctx, comp1, completedForComponents);
+    pushContext(ctx, comp2, completedForComponents);
+
+    ctx.getOperationExecutionParams(comp1).getCallback().complete(comp1);
```
marianogonzalez: does this really reproduces the case you want to test? You are after all calling this is a pretty synchronous manner...

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/execution/SdkInternalContextTestCase.java
```diff
@@ -0,0 +1,70 @@
...
+        completedForComponents.add(location);
+      }
+
+      @Override
+      public void error(Throwable e) {
+        throw new MuleRuntimeException(e);
+      }
+    });
+  }
+
```
marianogonzalez: empty line

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
elrodro83: was, it os now caches from some weeks ago

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/SdkInternalContext.java
elrodro83: it does equals on String, which is cheaper that calculating all of its hashes

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/execution/SdkInternalContextTestCase.java
elrodro83: with the api change in the `SdkInternalContext` i am forcing the caller to avoid the issue. If i weren't passing the location in this test, then yes, it would reproduce the issue.

PR #8741, opened 2020-03-18
MULE-18200: 4.3.0: expression errors hang MUnit test execution

elrodro83:
present

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
@@ -311,8 +311,25 @@ public abstract class ComponentMessageProcessor<T extends ComponentModel> extend
             .map(event -> {
               try {
                 return addContextToEvent(event, ctx);
-              } catch (MuleException t) {
-                throw propagateWrappingFatal(localOperatorErrorHook.apply(t, event));
+              } catch (Exception t) {
+                // Force the error mapper from the chain to be used.
+                // When using Mono.create with sink.error, the error mapper from the
+                // context is ignored, so it has to be explicitly used here.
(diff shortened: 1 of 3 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -311,8 +311,25 @@ public CoreEvent process(CoreEvent event) throws MuleException {
...
+                    // impose a timeout in the operation, which we don't want to.
+                    // In this case, the flux will be complete when there are no more inflight operations.
+                    || ctx.getOrDefault(WITHIN_PROCESS_TO_APPLY, false)) {
+                  // if `sink.error` is called here, it will cancel the flux altogether.
+                  // That's why an `Either` is used here,
+                  // so the error can be propagated afterwards in a way consistent with our expected error handling.
+                  errorSwitchSinkSinkRef.next(left(mapped, CoreEvent.class));
+                }
+
+                throw propagateWrappingFatal(mapped);
```
marianogonzalez: is it correct to throw the exception even if the if resolved to `true` and the errorSink received the `next()` call?

elrodro83: yes, it will depend on what pipeline is using this component to determine if the exception is needed (Monos) or the result in the callback (Fluxes). In either case, it is a map, so it needs to return a mapped event, or in this case propagate the exception.

marianogonzalez: ok.. i'll trust you... but I'm looking at you with suspicious eyes...

https://i.pinimg.com/originals/48/4f/42/484f42995f748050bd50574d42aa0610.gif

PR #8756, opened 2020-03-25
MULE-18219: Fix flaky DynamicConfigurationProviderTestCase

elrodro83:
Also fix:
* Shutting down of app with dynamics configs restate already expired ones

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
@@ -94,6 +94,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.BiFunction;
 import java.util.function.Function;
 
 import javax.inject.Inject;
@@ -158,7 +159,7 @@ public abstract class ExtensionComponent<T extends ComponentModel> extends Abstr
 
   protected MetadataCacheIdGenerator<ComponentAst> cacheIdGenerator;
 
-  private Function<CoreEvent, Optional<ConfigurationInstance>> configurationResolver;
+  private BiFunction<CoreEvent, Boolean, Optional<ConfigurationInstance>> configurationResolver;
 
   private List<ErrorMapping> errorMappings = emptyList();
 
@@ -206,7 +207,7 @@ public abstract class ExtensionComponent<T extends ComponentModel> extends Abstr
 
   private void initConfigurationResolver() {
     if (!requiresConfig.get()) {
-      configurationResolver = event -> empty();
+      configurationResolver = (event, stopping) -> empty();
       return;
     }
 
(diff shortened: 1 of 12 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
```diff
@@ -158,7 +159,7 @@
 
   protected MetadataCacheIdGenerator<ComponentAst> cacheIdGenerator;
 
-  private Function<CoreEvent, Optional<ConfigurationInstance>> configurationResolver;
+  private BiFunction<CoreEvent, Boolean, Optional<ConfigurationInstance>> configurationResolver;
```
marianogonzalez: I don't think this added parameter is a good idea. Why not just injecting the MuleContext and using the isStopping() method?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
```diff
@@ -215,26 +216,31 @@ private void initConfigurationResolver() {
...
-        if (instance == null) {
-          throw new IllegalModelDefinitionException(format(
-                                                           "Root component '%s' contains a reference to config '%s' but it doesn't exists",
-                                                           getLocation().getRootContainerName(),
-                                                           configurationProvider));
-        }
+      configurationResolver = (event, stopping) -> {
+        if (stopping) {
+          // Avoid populating the cache of dynamic connections if stopping
+          return configurationProvider.get().getIfPresent(event);
```
marianogonzalez: This isn't a matter of not populating the cache. If the context is stopping, the dynamic config shouldn't be spawned at all.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DynamicConfigurationProvider.java
```diff
@@ -140,15 +141,25 @@ public ConfigurationInstance get(Event event) {
         if (connectionProviderResolver.getResolverSet().isPresent()) {
           providerResult = ((ResolverSet) connectionProviderResolver.getResolverSet().get()).resolve(resolvingContext);
         }
-        return getConfiguration(new Pair<>(result, providerResult), (CoreEvent) event);
+
+        if (muleContext.isStopping() || muleContext.isStopped()) {
```
marianogonzalez: this if seems to be inverted, but also conceptually wrong.. 

if the context is not stopping, then you need to do getConfiguration() as usual so that the cache is used. If the context IS stopping, then no new config should be created. Also, the compute method you added should never be explicitly called. If a dynamic config is not added to the cache, then it will never be stopped and will generate a leak.

elrodro83: the cache is a loading cache, that's the issue, doing a get on it populates it, that's what i'm avoiding with this change, and generating a different compute method

marianogonzalez: I know.. my point is that a get() to a dynamic config should simply fail when the context is stopping.

PR #8759, opened 2020-03-27
MULE-18227: NPE with scatter-gather within parallel-foreach

elrodro83:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/SdkInternalContext.java
```diff
@@ -46,16 +48,17 @@ public static SdkInternalContext from(CoreEvent event) {
    * SDK components may be nested within each other, so some of the context must be kept separately for the component it belongs
    * to.
    */
-  private final Map<ComponentLocation, LocationSpecificSdkInternalContext> locationSpecificContext = new SmallMap<>();
+  private final Map<Pair<ComponentLocation, String>, LocationSpecificSdkInternalContext> locationSpecificContext =
```
marianogonzalez: Mmmmm... shouldn't a ConcurrentHashMap be faster than a SmallMap synchronized like this?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/SdkInternalContext.java
```diff
@@ -46,16 +48,17 @@ public static SdkInternalContext from(CoreEvent event) {
...
+      synchronizedMap(new SmallMap<>());
 
   private Function<Context, Context> innerChainSubscriberContextMapping = identity();
 
-  public void removeContext(ComponentLocation location) {
-    locationSpecificContext.remove(location);
+  public void removeContext(Pair<ComponentLocation, String> locationAndEventId) {
```
marianogonzalez: I think that a specific class should be created for this rather than a Pair

elrodro83:
--validate

PR #8761, opened 2020-03-28

gsfernandes:
When a plugin is declared in a domain it will belong to the domain region class loader and it will have as its parent the domain region class loader.
So, whenever an operation/source of the plugin is invoked by Runtime/SDK the TCCL will set the plugin class loader. Therefore if the plugin wants to load a class or resource from the application it will not be able to do that as it will ended up delegating the domain region class loader and this one has not visibility over the application. Be aware that the operation/source will be present in the application and not in the domain.
In order to solve this we have discussed different alternatives:
* Add an API to load resources and classes from the application, this will not work for cases like external libraries used by the plugin that use SPI or other mechanism to load resources/classes
* Set as not supported, in particular cases like Java Module which doesn't have a configuration and should be declared in an application's scope only
* Or do what we have done here, basically SDK when invoking a method for a plugin it will change the TCCL to be a composite class loader that delegates first to the plugin class loader and if not present try it with the TCCL set by the Runtime which would be defined by where the component is defined (domain or application).

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/util/CompositeClassLoader.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/util/CompositeClassLoader.java
@@ -45,16 +45,25 @@ public class CompositeClassLoader extends ClassLoader {
   }
 
   @Override
-  public Class<?> loadClass(String name) throws ClassNotFoundException {
+  protected Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundException {
+    Class<?> result = null;
     ClassNotFoundException firstException = null;
     for (ClassLoader classLoader : delegates) {
       try {
-        return classLoader.loadClass(name);
+        result = classLoader.loadClass(name);
+        break;
       } catch (ClassNotFoundException e) {
         firstException = e;
       }
     }
-    throw firstException;
+    if (result == null) {
+      throw firstException;
+    }
+
+    if (resolve) {
+      resolveClass(result);
+    }
+    return result;
   }
 
   @Override
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/GeneratedMethodComponentExecutor.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/GeneratedMethodComponentExecutor.java
@@ -24,6 +24,7 @@ import org.mule.runtime.api.meta.model.ComponentModel;
 import org.mule.runtime.api.meta.model.parameter.ParameterGroupModel;
 import org.mule.runtime.core.api.MuleContext;
 import org.mule.runtime.core.api.context.MuleContextAware;
+import org.mule.runtime.core.internal.util.CompositeClassLoader;
 import org.mule.runtime.extension.api.runtime.operation.ExecutionContext;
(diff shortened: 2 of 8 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/GeneratedMethodComponentExecutor.java
```diff
@@ -164,7 +166,7 @@ public void setMuleContext(MuleContext context) {
         }
         return resolvedParams;
       } finally {
-        setContextClassLoader(thread, extensionClassLoader, currentClassLoader);
+        setContextClassLoader(thread, compositeClassLoader, currentClassLoader);
```
gsfernandes: We would need to review this with the improvement made by Mariano Gonzalez for performance as switching class loaders in TCCL was a "problem". On this case we cannot 'reuse' the composite as it will be created with the current TCCL which could be a domain or application.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultExecutionMediator.java
```diff
@@ -201,11 +202,12 @@ private void executeCommand(CompletableComponentExecutor<M> executor,
       }
       final Thread currentThread = Thread.currentThread();
       final ClassLoader currentClassLoader = currentThread.getContextClassLoader();
-      setContextClassLoader(currentThread, currentClassLoader, extensionClassLoader);
+      final CompositeClassLoader compositeClassLoader = new CompositeClassLoader(extensionClassLoader, currentClassLoader);
+      setContextClassLoader(currentThread, currentClassLoader, compositeClassLoader);
```
gsfernandes: Same here.

core/src/main/java/org/mule/runtime/core/internal/util/CompositeClassLoader.java
```diff
@@ -45,16 +45,25 @@ public CompositeClassLoader(ClassLoader first, ClassLoader... others) {
...
         firstException = e;
       }
     }
-    throw firstException;
+    if (result == null) {
+      throw firstException;
+    }
+
+    if (resolve) {
+      resolveClass(result);
```
elrodro83: why this? shouldn't this be called on the actual classloader instead of the composite?

gsfernandes: You are right, but we cannot call neither loadClass(s,b) or resolveClass(c) to the delegates as both methods are protected. So, we would need to just ignore the resolve parameter assuming that once the class is "used" it will trigger to load the nested classes (imports, etc) but it will go to the class loader that loaded this class instead of using the CompositeClassLoader (that could ended up finding the class in other class loader).

gsfernandes: The whole idea of the PR was not change the CompositeClassLoader, instead using it as the TCCL for execution of plugin operations.
This change was only made in order to fix the issue with this class loader when it was introduced in 4.1.5. I will move this change to another PR as there is already a ticket reported for this https://www.mulesoft.org/jira/browse/MULE-18244.

PR #8762, opened 2020-03-30
MULE-18234: App returns correct results, but "Test Connection" FAILS

lucianoRM:


core/src/main/java/org/mule/runtime/core/internal/lifecycle/phases/DefaultLifecycleObjectSorter.java
```diff
@@ -75,13 +80,13 @@ protected int doAddObject(String name, Object object, List<Object> bucket) {
    */
   @Override
   public List<Object> getSortedObjects() {
-    List<Object> sorted = new ArrayList<>(objectCount);
+    Set<Object> sorted = newSetFromMap(new LinkedHashMap<>(objectCount));
```
elrodro83, line 83: what's the point of doing this?

lucianoRM:
--validate

modules/extensions-xml-support/src/test/java/org/mule/runtime/extension/internal/config/dsl/XmlSdkConfigurationProviderLifecycleTestCase.java
```diff
@@ -0,0 +1,71 @@
...
+import org.mule.runtime.core.api.config.MuleConfiguration;
+import org.mule.runtime.extension.api.runtime.config.ConfigurationProvider;
+
+import com.google.common.collect.ImmutableList;
+
+import java.util.List;
+
+import org.junit.Test;
+
+public class XmlSdkConfigurationProviderLifecycleTestCase {
```
elrodro83, line 38: allure annotations

core/src/main/java/org/mule/runtime/core/internal/lifecycle/phases/DefaultLifecycleObjectSorter.java
lucianoRM: It was supposed to be an IdentityHashMap to avoid returning the same object instance in the list mor that once. 
Changing it...

PR #8768, opened 2020-04-01
DEL-31: Fix Javadoc Generation for 4.3.0

lvzh90:


pom.xml
```diff
@@ -264,7 +264,7 @@
         <antrun.plugin.version>1.8</antrun.plugin.version>
         <install.plugin.version>2.5.2</install.plugin.version>
         <checkstyle.plugin.version>2.17</checkstyle.plugin.version>
-        <javaFormatter.plugin.version>1.9.0</javaFormatter.plugin.version>
+        <javaFormatter.plugin.version>2.0.1</javaFormatter.plugin.version>
```
anosenzo, line 267: Why do you need to change the formatter plugin?

aiannucci: Why not? It matches the version of mule-api and other projects now

lvzh90: Excatly

PR #8769, opened 2020-04-01
MULE-18255: OCSClient must support configuring the api version of the service by system property

ndinu:


modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/builder/OAuthPlatformManagedDancerBuilder.java
```diff
@@ -47,6 +47,17 @@
...
   OAuthPlatformManagedDancerBuilder platformUrl(String platformUrl);
 
+  /**
+   * Sets the platform API version to a specific one.
+   *
+   * @param apiVersion the platform API version
+   * @return {@code this} builder
+   * @since 4.3.1 4.4.0
+   */
+  default OAuthPlatformManagedDancerBuilder apiVersion(String apiVersion) {
```
marianogonzalez, line 57: this method should be introduced in 4.4.0 only. 4.3.x should deal with sys prop only

marianogonzalez:
Issue key shouold be MULE

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/oauth/ocs/PlatformManagedOAuthConfig.java
```diff
@@ -62,16 +64,26 @@ public static PlatformManagedOAuthConfig from(String ownerConfigName,
...
                                           delegateGrantType);
   }
 
+  private static String getProperty(ConfigurationProperties configurationProperties, String key, boolean isRequired) {
+    if (isRequired) {
+      return configurationProperties.resolveStringProperty(key)
+          .orElseThrow(() -> new IllegalArgumentException(format("OCS property '%s' has not been set", key)));
+    } else {
+      return configurationProperties.resolveStringProperty(key)
+          .orElse(null);
```
marianogonzalez, line 81: default value should be v1

ndinu: discussed privately

modules/oauth-api/src/main/java/org/mule/runtime/oauth/api/builder/OAuthPlatformManagedDancerBuilder.java
ndinu: Done

PR #8771, opened 2020-04-01
MULE-18244: Connectors that use third-party libraries that create cus…

gsfernandes:
…tom classloader are not able to load classes if they rely on loadClass(s,b) from CompositeClassLoader

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/util/CompositeClassLoader.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/util/CompositeClassLoader.java
@@ -47,7 +47,6 @@ public class CompositeClassLoader extends ClassLoader {
   /**
    * Overrides the loadClass in order to support scenarios where a custom class loader is created in a plugin
    * and these calls to this method explicitly.
-   * We cannot resolve the class at this point as we should delegate to protected methods and it is no possible.
    *
    * @param name    The <a href="#name">binary name</a> of the class
    * @param resolve If <tt>true</tt> then resolve the class
@@ -59,7 +58,13 @@ public class CompositeClassLoader extends ClassLoader {
     ClassNotFoundException firstException = null;
     for (ClassLoader classLoader : delegates) {
       try {
-        return classLoader.loadClass(name);
+        Class<?> result = classLoader.loadClass(name);
+
+        if (resolve) {
+          resolveClass(result);
+        }
+
+        return result;
       } catch (ClassNotFoundException e) {
         firstException = e;
       }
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mule/runtime/core/internal/util/CompositeClassLoader.java
```diff
@@ -44,8 +44,18 @@ public CompositeClassLoader(ClassLoader first, ClassLoader... others) {
...
+   * We cannot resolve the class at this point as we should delegate to protected methods and it is no possible.
+   *
+   * @param name    The <a href="#name">binary name</a> of the class
+   * @param resolve If <tt>true</tt> then resolve the class
+   * @return The resulting <tt>Class</tt> object
+   * @throws ClassNotFoundException If the class could not be found
+   */
   @Override
-  public Class<?> loadClass(String name) throws ClassNotFoundException {
+  public Class<?> loadClass(String name, boolean resolve) throws ClassNotFoundException {
```
elrodro83, line 58: i think the two methods should be overriddes, and the resolve flag should be also propagated to the delegated loadClass, right?

gsfernandes: The loadClass(name) in ClassLoader delegates to the loadClass(name, resolve) so no need to override that method. Problem with the resolve as explained in the other PR is that the method is protected therefore we cannot call it on the list of delegates. So, we will just not support the resolve=true meaning that it will be a way to load classes eagerly. This is used in some frameworks but it should not affect to us.

gsfernandes: It is not clear to me yet what the resolve will do as according to javadoc:
"Links the specified class. This (misleadingly named) method may be used by a class loader to link a class". So, I ended up checking what we are doing in our ClassLoaders and found that the FineGrainControlClassLoader has kind of the same dilemma: 
It implements the loadClass(name, resolve) method but checks for the lookPolicy, this will give him a classloader to delegate the load of the class, for instance you are at P1 CL and try load a class from application so the P1 CL will get from lookupPolicy the APP CL and call the loadClass(name) method and then P1 CL will do the the call to resolveClass(clazz).

loadClass(name, resolve)-> https://github.com/mulesoft/mule/blob/mule-4.2.2/modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/FineGrainedControlClassLoader.java#L67
Then if the policy returns a different class loader it will call its loadClass(name)
https://github.com/mulesoft/mule/blob/mule-4.2.2/modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/FineGrainedControlClassLoader.java#L93
Finally the loadClass(name, resolve) method will do the resolve no matter if the class was actually loaded with this class loader or other:
https://github.com/mulesoft/mule/blob/mule-4.2.2/modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/FineGrainedControlClassLoader.java#L110

So, the conclusion is I would push another change to call the resolveClass to the class resolved and no need to implement loadClass(name) only as it delegates to loadClass(name, resolve).

PR #8772, opened 2020-04-02
MULE-18248: Illegal type name exception in AIX OS using JDK8, the same

elrodro83:
app in 4.3.0-rc2 started fine


garodriguezlp:
--validate

PR #8779, opened 2020-04-04

gsfernandes:
When a plugin is declared in a domain it will belong to the domain region class loader and it will have as its parent the domain region class loader.
So, whenever an operation/source of the plugin is invoked by Runtime/SDK the TCCL will set the plugin class loader. Therefore if the plugin wants to load a class or resource from the application it will not be able to do that as it will ended up delegating the domain region class loader and this one has not visibility over the application. Be aware that the operation/source will be present in the application and not in the domain.
In order to solve this we have discussed different alternatives:
* Add an API to load resources and classes from the application, this will not work for cases like external libraries used by the plugin that use SPI or other mechanism to load resources/classes
* Set as not supported, in particular cases like Java Module which doesn't have a configuration and should be declared in an application's scope only
* Or do what we have done here, basically SDK when invoking a method for a plugin it will change the TCCL to be a composite class loader that delegates first to the plugin class loader and if not present try it with the TCCL set by the Runtime which would be defined by where the component is defined (domain or application).


JuLuzzi (dismissed):

elrodro83 (approved):

eze210:
--validate

elrodro83 (dismissed):

elrodro83 (approved):

PR #8784, opened 2020-04-06
MULE-18205: Review performance overhead of

elrodro83:
FlowProcessMediator#policySuccessError

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/message/InternalEvent.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/message/InternalEvent.java
@@ -46,6 +46,21 @@ public interface InternalEvent extends PrivilegedEvent {
   <T> T getInternalParameter(String key);
 
   /**
+   * @return a {@link EventInternalContext} with state from source-flow dispatch
+   * @since 4.4, 4.3.1
+   */
+  public EventInternalContext getFlowProcessMediatorPhaseContext();
+
+  /**
+   * Sets context related to the source-flow dispatch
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/mule/runtime/core/internal/message/InternalEvent.java
```diff
@@ -45,6 +45,21 @@
...
+  public EventInternalContext getFlowProcessMediatorPhaseContext();
+
+  /**
+   * Sets context related to the source-flow dispatch
+   * 
+   * @return a {@link EventInternalContext} with state from source-flow dispatch
+   * @param context an {@link EventInternalContext}
+   * @since 4.4, 4.3.1
+   */
+  public void setFlowProcessMediatorPhaseContext(EventInternalContext flowProcessMediatorPhaseContext);
```
marianogonzalez, line 61: I would remove the `Phase` thing from this name. The Phase concept is genetic garbage inherited from 3.x... I wouldn't bubble that up all the way here

marianogonzalez, line 61: At a quick glance, all other implementations of `EventInternalContext` are named explicitly when used in method signatures such as this. Although I see why you want to keep this so abstract, this API is telling people that ANY implementation is valid when it's not the case. I think that we should promote PhaseContext to a first class citizen like OperationPolicyContext and rename it to something better while at it (this comment also implies the getter method above and other variable names)

PR #8785, opened 2020-04-07
MULE-18269: XML SDK primary parameters are sent as null

elrodro83:


modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/processor/ModuleOperationMessageProcessor.java
```diff
@@ -164,6 +165,8 @@ public ModuleOperationMessageProcessor(Map<String, String> properties,
       if (parameters.containsKey(parameterName)) {
         final String xmlValue = parameters.get(parameterName).trim();
         result.put(parameterName, new Pair<>(xmlValue, parameterModel.getType()));
+      } else if (PRIMARY_CONTENT.equals(parameterModel.getRole())) {
```
marianogonzalez, line 168: this actually applies to any ooptional parameter, not just the primary content ones. Also, if that default value is an expression, where is that evaluated?

elrodro83: the execution evaluates that, will check for other optional params

elrodro83: optional parameters have tests and are green: `ModuleSimpleTestCase#testSetPayloadParamDefaultUseOptionalFlow`

elrodro83:
--validate

elrodro83:
--validate

PR #8789, opened 2020-04-08
MULE-16668: Excessive validations on pooled connections.

ndinu:


core/src/main/java/org/mule/runtime/core/internal/connection/PoolingConnectionHandler.java
```diff
@@ -68,10 +68,6 @@ public void release() {
 
     boolean returnAttempted = false;
     try {
-      if (!connectionProvider.validate(connection).isValid()) {
-        LOGGER.debug("Connection is not valid, it is destroyed and not returned to the pool.");
-        return;
-      }
       poolingListener.onReturn(connection);
 
       pool.returnObject(connection);
```
marianogonzalez, line 73: this needs to happen on a finally clause. If line 71 files, this needs to be executed anyway

ndinu:
--validate

ndinu:
--validate

ndinu:
--validate

PR #8790, opened 2020-04-08
MULE-18288: Prevent memory leaks from plugins that use scala

elrodro83:


modules/artifact/src/main/java/org/mule/module/artifact/classloader/ScalaClassValueReleaser.java
```diff
@@ -0,0 +1,129 @@
...
+      LOGGER.warn("Unable to initialize ScalaClassValueReleaser", t);
+    }
+
+    this.classValueMapField = classValueMapField;
+  }
+
+  @Override
+  public void release() {
+    try {
+      WeakHashMap classValueMap = getClassValueMap();
```
fsgonz, line 52: shouldn't we verify that that the scala releaser could be succesfully intiialized so that no NPE is raised during each redeployment?

PR #8791, opened 2020-04-08
MULE-18281: ArtifactAstDependencyGraph is not taking configRef parameters into account

lucianoRM:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/LazyMuleArtifactContext.java
```diff
@@ -225,8 +223,9 @@ public LazyMuleArtifactContext(MuleContext muleContext, ConfigResource[] artifac
     }
 
     initialize();
+    this.graph = generateFor(applicationModel);
```
gsfernandes, line 226: Put a comment at least: 
// Macro Expansion will generate new component models so we will create the graph at the end.

PR #8794, opened 2020-04-13
MULE-18286: Error mappings not honored while running application on L…

gsfernandes:
…azyInit

### DIFF
--- a/modules/spring-config/src/main/java/org/mule/runtime/config/internal/MuleArtifactContext.java
+++ b/modules/spring-config/src/main/java/org/mule/runtime/config/internal/MuleArtifactContext.java
@@ -155,7 +155,7 @@ public class MuleArtifactContext extends AbstractRefreshableConfigApplicationCon
   public static final String INNER_BEAN_PREFIX = "(inner bean)";
 
   protected final ComponentBuildingDefinitionRegistry componentBuildingDefinitionRegistry =
-      new ComponentBuildingDefinitionRegistry();
+          new ComponentBuildingDefinitionRegistry();
   private final OptionalObjectsController optionalObjectsController;
   private final Map<String, String> artifactProperties;
   private final ArtifactDeclaration artifactDeclaration;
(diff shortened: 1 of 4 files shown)


gsfernandes:
@afelisatti @lucianoRM I had to made changes in order to fix tests. Nice to see that we have plenty of tests for errors.

lucianoRM:
--validate

gsfernandes:
--validate

gsfernandes:
--validate

modules/spring-config/src/main/java/org/mule/runtime/config/internal/MuleArtifactContext.java
```diff
@@ -466,9 +466,9 @@ private void registerErrorMappings(SpringComponentModel componentModel, Set<Stri
         ComponentIdentifier source = parameters.containsKey(SOURCE_TYPE)
             ? buildFromStringRepresentation(parameters.get(SOURCE_TYPE)) : ANY;
 
-        muleContext.getErrorTypeRepository()
-            .lookupErrorType(source)
-            .orElseThrow(() -> new MuleRuntimeException(createStaticMessage("Could not find error '%s'.", source)));
+        if (!muleContext.getErrorTypeRepository().lookupErrorType(source).isPresent()) {
```
lucianoRM, line 469: Why is this? I think that the functional approach is better.

gsfernandes: https://sonarqube.kbuild.msap.io/coding_rules?open=squid%3AS2201&rule_key=squid%3AS2201

gsfernandes: 'Return values from functions without side effects should not be ignored'
When the call to a function doesn't have any side effects, what is the point of making the call if the results are ignored? In such case, either the function call is useless and should be dropped or the source code doesn't behave as expected.

gsfernandes: There is another way to still get the functional approach but makes code more unreadable.

lucianoRM: ok, make sense.

gsfernandes:
--validate

gsfernandes:
--validate

PR #8796, opened 2020-04-14
MULE-18234: App returns correct results, but "Test Connection" FAILS - formatter

lucianoRM:


lucianoRM:
--validate

lucianoRM:
--validate

PR #8806, opened 2020-04-16
MULE-18310: Reduce time taken by tests - part 1

elrodro83:


modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/FilteringArtifactClassLoader.java
```diff
@@ -73,6 +75,12 @@ public FilteringArtifactClassLoader(ClassLoader parent, ArtifactClassLoader arti
     this.artifactClassLoader = artifactClassLoader;
     this.filter = filter;
     this.exportedServices = exportedServices;
+
+    verboseLogging = LOGGER.isDebugEnabled() || isVerboseLoggingEnabled();
```
marianogonzalez, line 79: this breaks the feature of changing log levels at runtime

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/MethodArgumentResolverDelegate.java
```diff
@@ -192,8 +192,9 @@ private void initArgumentResolvers() {
         argumentResolver = ERROR_ARGUMENT_RESOLVER;
       } else if (SourceCallbackContext.class.equals(parameterType)) {
         argumentResolver = SOURCE_CALLBACK_CONTEXT_ARGUMENT_RESOLVER;
-      } else if (isParameterContainer(annotations.keySet(), typeLoader.load(parameterType)) &&
-          !((ParameterGroup) annotations.get(ParameterGroup.class)).showInDsl()) {
+      } else if (annotations.keySet().contains(ParameterGroup.class)
```
marianogonzalez, line 195: I find it surprising that this moves the needle, specially since this only happens ones per MuleContext, but I choose to believe

tests/unit/src/main/java/org/mule/tck/SimpleUnitTestSupportLifecycleSchedulerDecorator.java
```diff
@@ -87,16 +88,12 @@ public void execute(Runnable command) {
 
   @Override
   public ScheduledFuture<?> scheduleWithCronExpression(Runnable command, String cronExpression) {
-    ScheduledFuture<?> recurrentTask = decorated.scheduleWithCronExpression(wrap(command), cronExpression);
-    recurrentTasks.add(recurrentTask);
```
marianogonzalez, line 91: aren't you breaking the assertions around tasks being completed and properly disposed and other shuutdown validations with this? My feat is that we reduce execution time at the expense of no longer catching potential leaks and other herbs

elrodro83: no, the only code where objects of this class are being built mocks these methods, so these methods were never actually being executed.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/MethodArgumentResolverDelegate.java
elrodro83: it avoids calling the `typeLoader.load` in some cases, which may be expensive.

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/FilteringArtifactClassLoader.java
```diff
@@ -76,11 +76,11 @@ public FilteringArtifactClassLoader(ClassLoader parent, ArtifactClassLoader arti
...
     this.exportedServices = exportedServices;
 
-    verboseLogging = LOGGER.isDebugEnabled() || isVerboseLoggingEnabled();
+    verboseLogging = valueOf(getProperty(MULE_LOG_VERBOSE_CLASSLOADING));
   }
 
-  private boolean isVerboseLoggingEnabled() {
-    return LOGGER.isInfoEnabled() && valueOf(getProperty(MULE_LOG_VERBOSE_CLASSLOADING));
+  private boolean isVerboseLogging() {
+    return LOGGER.isDebugEnabled() || verboseLogging;
```
marianogonzalez, line 83: invert this for better perf

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/FineGrainedControlClassLoader.java
```diff
@@ -56,11 +56,11 @@ public FineGrainedControlClassLoader(URL[] urls, ClassLoader parent, ClassLoader
...
     checkArgument(lookupPolicy != null, "Lookup policy cannot be null");
     this.lookupPolicy = lookupPolicy;
-    verboseLogging = LOGGER.isDebugEnabled() || isVerboseLoggingEnabled();
+    verboseLogging = valueOf(getProperty(MULE_LOG_VERBOSE_CLASSLOADING));
   }
 
-  private boolean isVerboseLoggingEnabled() {
-    return LOGGER.isInfoEnabled() && valueOf(getProperty(MULE_LOG_VERBOSE_CLASSLOADING));
+  private boolean isVerboseLogging() {
+    return LOGGER.isDebugEnabled() || verboseLogging;
```
marianogonzalez, line 63: this is not only repeated code, but in the original version INFO level was being used whereas now you degraded it to DEBUG

elrodro83: nope, this is in the same place but is not the same method. This does the same evaluation that was previously done in the constructor, to account for log level changes. The repeated code you mention i think is with `FilteringArtifactClassLoader`. There is no common superclass between the 2 to extract this logic to, and it seems overkill to extract an utils for just 2 lines (plus i would need to pass the logger around anyway)

PR #8811, opened 2020-04-17
MULE-18312: LazyMuleArtifactContext ignores custom Global Error Handler

mlischetti:


gsfernandes:
Would be better to move out from the parsers definitions if we do this in AST. Which is the source of true for the dependencies on componets.
How that would work? Would be need add a custom logic to support these types? We have to review as there were a few usages of "alwaysEnabled" at parsers definition. We need to also check if other plugins that define their parsers if they are using this or not...
I would like to think this perhaps in the extension model if not how we can handle custom parsers?
If this is the way to go we would need to remove the explicit sorting that it is being done in MuleArtifactContext for alwaysEnabled beans.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/CoreComponentBuildingDefinitionProvider.java
```diff
@@ -484,6 +484,7 @@ public void init() {
         .withSetterParameterDefinition(INHERIT_ITERABLE_REPEATABILITY,
                                        fromSimpleParameter(INHERIT_ITERABLE_REPEATABILITY).build())
         .withRegistrationName(OBJECT_MULE_CONFIGURATION)
+        .alwaysEnabled(true)
```
gsfernandes, line 487: If we do this we can get rid of the `resetMuleConfiguration();` in LazyMuleArtifactContext as it will be always disposed/unregistered before it gets registered again.

mlischetti:
--validate

PR #8816, opened 2020-04-21
MULE-18111: Add information about the last commit included in the distro build

JuLuzzi:


.gitignore
```diff
@@ -30,3 +30,6 @@ testdir/
 hostkey.ser
 *.h2.db
 *.factorypath
+
+# Manifest
+src/main/resources/META-INF/MANIFEST.MF
```
anosenzo, line 35: Why did you need to include this line in the .gitignore?

elrodro83, line 35: is this needed?

PR #8817, opened 2020-04-21
MULE-18284: ErrorType of an extension not found in a policy, when the

elrodro83:
extension is also in the app

### DIFF
--- a/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultPolicyTemplateFactory.java
+++ b/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultPolicyTemplateFactory.java
@@ -8,12 +8,14 @@
 package org.mule.runtime.module.deployment.impl.internal.policy;
 
 import static java.lang.String.format;
+import static java.util.Collections.emptySet;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
 import static org.mule.runtime.api.util.Preconditions.checkArgument;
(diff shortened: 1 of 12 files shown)


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultPolicyTemplateFactory.java
```diff
@@ -73,13 +77,38 @@ public PolicyTemplate createArtifact(Application application, PolicyTemplateDesc
...
 
     List<ArtifactPlugin> artifactPlugins = createArtifactPluginList(policyClassLoader, resolvedPolicyPluginsDescriptors);
+    return new DefaultPolicyTemplate(policyClassLoader.getArtifactId(), descriptor, policyClassLoader,
+                                     artifactPlugins,
+                                     resolveOwnArtifactPlugins(application, descriptor, policyClassLoader, artifactPlugins));
+  }
 
-    validateArtifactLicense(policyClassLoader.getClassLoader(), artifactPlugins, licenseValidator);
+  // Need all the plugins that the policy itself depends on, while keeping a relationship with the appropriate classloader.
+  private List<ArtifactPlugin> resolveOwnArtifactPlugins(Application application, PolicyTemplateDescriptor descriptor,
```
gsfernandes, line 86: I have tried to understand why this needs to be done only in the context of Policies and not with Applications. Still not clear to me the reason but seems that descriptors have different plugin registered. As this fixes the issue we should merge this and I would add a comment to https://www.mulesoft.org/jira/browse/MULE-15842 which also related to hardcoded code for HTTP on policies.

PR #8820, opened 2020-04-22
MULE-18324: Metadata cache id generation must use ComponentAst instead

elrodro83:
of DslElementModel


modules/spring-config/src/main/java/org/mule/runtime/config/api/dsl/model/metadata/ComponentBasedValueProviderCacheIdGenerator.java
```diff
@@ -6,62 +6,33 @@
...
-import org.mule.runtime.config.api.dsl.model.DslElementModelFactory;
-import org.mule.runtime.config.internal.model.ComponentModel;
 import org.mule.runtime.core.internal.locator.ComponentLocator;
 import org.mule.runtime.core.internal.value.cache.ValueProviderCacheId;
 import org.mule.runtime.core.internal.value.cache.ValueProviderCacheIdGenerator;
 
 import java.util.Optional;
 
+/**
+ * @deprecated Use {@link ComponentAstBasedValueProviderCacheIdGenerator} directly.
```
lucianoRM, line 20: Why not remove this class too?

modules/spring-config/src/main/java/org/mule/runtime/config/api/dsl/model/metadata/ComponentBasedMetadataCacheIdGenerator.java
```diff
@@ -10,30 +10,27 @@
...
 import org.mule.runtime.core.internal.metadata.cache.MetadataCacheIdGenerator;
-import org.mule.runtime.core.internal.metadata.cache.MetadataCacheIdGeneratorFactory.ComponentLocator;
 
 import java.util.Optional;
 
 /**
  * A {@link ComponentModel} based implementation of a {@link MetadataCacheIdGenerator}
  *
  * @since 4.1.4, 4.2.0
+ * @deprecated Use {@link ComponentAstBasedMetadataCacheIdGenerator} directly.
```
lucianoRM, line 24: why not remove this class too?

modules/spring-config/src/main/java/org/mule/runtime/config/api/dsl/model/metadata/ComponentBasedValueProviderCacheIdGenerator.java
elrodro83: tooling is using it

modules/spring-config/src/main/java/org/mule/runtime/config/api/dsl/model/metadata/ComponentBasedMetadataCacheIdGenerator.java
elrodro83: tooling is using it

lucianoRM: But we should change it so that it uses the new implementation. Al least, let's create a jira ticket and add a comment for it.

modules/spring-config/src/main/java/org/mule/runtime/config/api/dsl/model/metadata/ComponentBasedValueProviderCacheIdGenerator.java
lucianoRM: But we should change it so that it uses the new implementation. Al least, let's create a jira ticket and add a comment for it.

PR #8822, opened 2020-04-23
MULE-18326: Remove dsl root elements from ApplicationModel#muleComponentModels

elrodro83:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/ComponentLocationVisitor.java
```diff
@@ -94,8 +93,7 @@ public void accept(Pair<ComponentAst, List<ComponentAst>> item) {
   }
 
   public void accept(ComponentAst componentModel, List<ComponentAst> hierarchy) {
-    if (((ComponentModel) componentModel).getParent() == null) {
-      // do not process root element
+    if (((ComponentModel) componentModel).getComponentLocation() != null) {
```
gsfernandes, line 96: If I'm not wrong I see locations with value "null" for some components. Would it be a problem here?

PR #8830, opened 2020-04-28
MULE-18355: Avoid component location recalculation after xmlSdk1 macroexpansion

elrodro83:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/model/ApplicationModel.java
```diff
@@ -337,24 +337,22 @@ public ApplicationModel(ArtifactConfig artifactConfig, ArtifactDeclaration artif
...
-    recursiveStreamWithHierarchy().forEach(clv);
+    recursiveStreamWithHierarchy().forEach(new ComponentLocationVisitor());
   }
 
   public void macroExpandXmlSdkComponents(Set<ExtensionModel> extensionModels) {
     expandModules(extensionModels, () -> {
       // TODO MULE-13894 do this only on runtimeMode=true once unified extensionModel names to use camelCase (see smart connectors
       // connectors and crafted declared extension models)
+
+      // TODO MULE-17419 (AST) Remove these 2 actions
```
gsfernandes, line 348: If we don't resolveComponentTypes those componentAst will not be associated to a parameter model etc therefore it will not be able to call getParameterAst by the parsers.

modules/spring-config/src/test/java/org/mule/runtime/config/internal/factories/FlowRefFactoryBeanTestCase.java
```diff
@@ -262,7 +262,7 @@ public void dynamicFlowRefDoesNotExist() throws Exception {
   private FlowRefFactoryBean createFlowRefFactoryBean(String name) throws Exception {
     FlowRefFactoryBean flowRefFactoryBean = new FlowRefFactoryBean();
     flowRefFactoryBean.setName(name);
-    flowRefFactoryBean.setAnnotations(singletonMap(LOCATION_KEY, fromSingleComponent("flow")));
+    flowRefFactoryBean.setAnnotations(singletonMap(LOCATION_KEY, DefaultComponentLocation.from("flow")));
```
gsfernandes, line 265: static import

elrodro83: there's already Flux.from (or Mono.from, done really remember now) statically imported in this class

modules/spring-config/src/main/java/org/mule/runtime/config/internal/model/ApplicationModel.java
elrodro83: if those are already resolved for the AST of the smart connector definition, the idea is to keep and use them. In that case, there's no need to recalculate them here. That's one of the golas of the linked TODO.

gsfernandes: 👍

PR #8834, opened 2020-05-04
MULE-18375: Working fix

balbifm:


core/src/main/java/org/mule/runtime/core/internal/policy/CompositeOperationPolicy.java
```diff
@@ -248,20 +248,53 @@ public void process(CoreEvent operationEvent,
...
+      if (throwable instanceof MessagingException) {
+        CoreEvent event = ((MessagingException) throwable).getEvent();
+        OperationPolicyContext context = OperationPolicyContext.from(event);
+        // Restore original event if chain failed before or during operation
+        if (!event.equals(context.getOriginalEvent()) && context.getNextOperationResponse() == null) {
+          throwable = new MessagingException(context.getOriginalEvent(), ((MessagingException) throwable));
+        }
+      }
+
+      callback.error(throwable);
```
marianogonzalez, line 279: this should happen in a finally clause. If another bug makes this method fail (let's say, somebody breaks `OperationPolicyContext.from`) , then the callback is never invoked and the process hungs, making it a painful to trace.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -409,10 +416,55 @@ private void onEvent(CoreEvent event, ExecutorCallback executorCallback) {
       };
 
       if (location != null) {
-        ((DefaultFlowCallStack) event.getFlowCallStack())
-            .setCurrentProcessorPath(resolvedProcessorRepresentation);
+        ((DefaultFlowCallStack) event.getFlowCallStack()).setCurrentProcessorPath(resolvedProcessorRepresentation);
+
+        Function<CoreEvent, CoreEvent> resultMapper = isTargetWithPolicies(event) ? (policyResult) -> {
+          if (returnDelegate instanceof PayloadTargetReturnDelegate) {
```
marianogonzalez, line 422: This is a no-go. making an instanceof at the return delegate completely defeats its purpose. Also, this code is super fragile and will potentially break each time we add another return delegate. Also, this if is kind of repeating the return delegate business logic to begin with.

Finally, creating lambda functions which carry contextual state is a performance problem. We need to avoid that

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -409,10 +416,55 @@ private void onEvent(CoreEvent event, ExecutorCallback executorCallback) {
...
+
+        ExecutorCallback effectiveCallback = new ExecutorCallback() {
+
+          @Override
+          public void complete(Object o) {
+            try {
+              CoreEvent policyResult = resultMapper.apply((CoreEvent) o);
+              executorCallback.complete(policyResult);
+            } catch (Throwable t) {
+              // WTF should I do here??
```
marianogonzalez, line 444: figure this out before attempting to merge

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -409,10 +416,55 @@ private void onEvent(CoreEvent event, ExecutorCallback executorCallback) {
...
+              executorCallback.error(t);
+            }
+          }
+
+          @Override
+          public void error(Throwable throwable) {
+            //            if (throwable instanceof MessagingException) {
```
marianogonzalez, line 451: resolve this

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -502,6 +555,8 @@ protected void doInitialise() throws InitialisationException {
         throw new InitialisationException(createStaticMessage("Could not resolve transactional configuration"), e, this);
       }
       returnDelegate = createReturnDelegate();
+      valueReturnDelegate = getValueReturnDelegate();
+      targetValueCompiledExpression = expressionManager.compile(targetValue, getTargetBindingContext(Message.of("")));
```
marianogonzalez, line 559: no. This is a concern for the return delegate. Should not do this here

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -411,8 +412,34 @@ private void onEvent(CoreEvent event, ExecutorCallback executorCallback) {
...
+                ExecutionContextAdapter operationContext = createExecutionContext(event);
+                policyResult = returnDelegate.asReturnValue(o, operationContext);
+              } catch (MuleException e) {
+                executorCallback.error(e);
+              } catch (Throwable t) {
+                executorCallback.error(unwrap(t));
+              }
+            }
+
+            executorCallback.complete(policyResult);
```
balbifm, line 432: This is getting executed even when the Catch clausules above are executed

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -411,8 +412,34 @@ private void onEvent(CoreEvent event, ExecutorCallback executorCallback) {
...
             .pushCurrentProcessorPath(resolvedProcessorRepresentation);
-        sdkInternalContext.getPolicyToApply(location, eventId)
-            .process(event, operationExecutionFunction, () -> resolutionResult, location, executorCallback);
+
+        ExecutorCallback effectiveCallback = new ExecutorCallback() {
+
+          @Override
+          public void complete(Object o) {
+            CoreEvent policyResult = (CoreEvent) o;
+            if (isTargetWithPolicies(event)) {
```
elrodro83, line 421: shouldn't this be the only place where `target` is handled? Why do need to know if a policy is applied?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -411,8 +412,33 @@ private void onEvent(CoreEvent event, ExecutorCallback executorCallback) {
...
+              try {
+                ExecutionContextAdapter operationContext = createExecutionContext(event);
+                policyResult = returnDelegate.asReturnValue(o, operationContext);
+                executorCallback.complete(policyResult);
```
balbifm, line 425: You need this on the else as well

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
szaffarano: When the operation doesn't have a target/targetValue, the behaviour is the same we had before the change: the returnDelegate is used in [ComponentMessageProcessor#mapped](https://github.com/mulesoft/mule/blob/MULE-18375/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java#L478), otherwise, [ComponentMessageProcessor#mapped](https://github.com/mulesoft/mule/blob/MULE-18375/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java#L478) uses a ValueReturnDelegate and here, after `PolicyEventMapper#onOperationPolicyFinish` was executed, is used the appropriate returnDelegate.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -502,6 +524,7 @@ protected void doInitialise() throws InitialisationException {
         throw new InitialisationException(createStaticMessage("Could not resolve transactional configuration"), e, this);
       }
       returnDelegate = createReturnDelegate();
+      valueReturnDelegate = getValueReturnDelegate();
```
marianogonzalez, line 527: where is `getValueReturnDelegate()` defined?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -226,6 +226,7 @@
   protected ExecutionMediator executionMediator;
   protected CompletableComponentExecutor componentExecutor;
   protected ReturnDelegate returnDelegate;
+  protected ReturnDelegate valueReturnDelegate;
```
marianogonzalez, line 229: why do we need this instead of simply reusing the same returnDelegate? What's the difference?

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessorTestCase.java
```diff
@@ -413,6 +414,17 @@ public void executeWithPolicy() throws Exception {
     verify(mockOperationPolicy).process(same(event), any(OperationExecutionFunction.class), any(), any(), any());
   }
 
+  @Test
+  public void executeWithPolicyReturnsOperationResultOnTarget() throws Exception {
+    // Assert return delegate only once
```
marianogonzalez, line 419: complete or delete

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessorTestCase.java
```diff
@@ -413,6 +414,17 @@ public void executeWithPolicy() throws Exception {
...
+    // fail("implemet");
+  }
+
+  @Test
+  public void executeReturnsOperationResultButKeepsAttributes() {
```
marianogonzalez, line 424: complete or delete

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessorTestCase.java
```diff
@@ -17,6 +17,7 @@
 import static org.hamcrest.CoreMatchers.nullValue;
 import static org.hamcrest.CoreMatchers.sameInstance;
 import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
```
marianogonzalez, line 20: remove

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -437,18 +441,62 @@ private void onEventSynchronous(CoreEvent event, ExecutorCallback executorCallba
...
+    };
+  }
+
   private ExecutorCallback mapped(ExecutorCallback callback, ExecutionContextAdapter<T> operationContext) {
     return new ExecutorCallback() {
 
       @Override
       public void complete(Object value) {
-        callback.complete(returnDelegate.asReturnValue(value, operationContext));
+        callback.complete(valueReturnDelegate.asReturnValue(value, operationContext));
```
marianogonzalez, line 499: isn't this the other way around? if there are no policies, this is the callback wrapper that is used, and you're wrongfully changing the return delefate in this case

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -437,18 +441,62 @@ private void onEventSynchronous(CoreEvent event, ExecutorCallback executorCallba
...
 
-      operationExecutionFunction.execute(resolutionResult, event, executorCallback);
+      operationExecutionFunction.execute(resolutionResult, event,
+                                         computeOperationReturnExecutionCallback(event, executorCallback));
     } catch (Throwable t) {
       executorCallback.error(unwrap(t));
     }
   }
 
+  private ExecutorCallback computeOperationReturnExecutionCallback(CoreEvent event, ExecutorCallback delegateCallback) {
```
marianogonzalez, line 451: isn't this for the policy use case only? In that case, this name is misleading

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -437,18 +441,62 @@ private void onEventSynchronous(CoreEvent event, ExecutorCallback executorCallba
...
+    return new ExecutorCallback() {
+
+      @Override
+      public void complete(Object o) {
+        ExecutionContextAdapter operationContext = null;
+        CoreEvent originalEvent = null;
+        try {
+          if (returnDelegate instanceof ValueReturnDelegate) {
```
marianogonzalez, line 459: I would make a best effort to avoid this. This only makes sense in the particular policy use case you're targeting. This is wrong in any other use case. So either do this differently, or make it absolutely impossible for this method to be used in any other use case.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/SdkInternalContext.java
```diff
@@ -199,5 +201,13 @@ public CoreEvent getOperationEvent() {
     public ExecutorCallback getCallback() {
       return callback;
     }
+
+    public void setExecutionContextAdapter(ExecutionContextAdapter executionContextAdapter) {
```
marianogonzalez, line 205: should this actually be mutable? @elrodro83 what say you?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
marianogonzalez: this deserves a large javadoc explaining why

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/SdkInternalContext.java
elrodro83: no it shouldn't there are other places already in use around here where this state may be kept (sdk context, for instance)

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
elrodro83: male it private, not protected

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -226,6 +227,18 @@
...
   protected ReturnDelegate returnDelegate;
+  /**
+   * TODO: MULE-18483 When a policy is applied to an operation that has defined a target, it's necessary to wait until the policy
+   * finishes to calculate the return value with {@link #returnDelegate}. But in this case, because of in order to execute the
+   * rest of the policy we need to transform the {@link Result} returned by the operation into a {@link CoreEvent}, we use
+   * {@link #valueReturnDelegate} as a helper class to do this transformation. It's used only when there is an operation that
+   * defines a target, and at the same time, there are operation policies applied to it. Finally, when the policy finishes, the
+   * proper {@link #returnDelegate} is executed.
+   * <p>
+   * It's an horrible solution, we only need a piece of code that transforms an {@link Object} into a {@link CoreEvent} and not a
```
marianogonzalez, line 238: rephrase, and point to the fact that the piece of code we need is already in the ValueReturnDelegate, we just need to extract it

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -403,16 +416,21 @@ private void onEvent(CoreEvent event, ExecutorCallback executorCallback) {
...
       };
 
       if (location != null) {
         ((DefaultFlowCallStack) event.getFlowCallStack())
             .pushCurrentProcessorPath(resolvedProcessorRepresentation);
-        sdkInternalContext.getPolicyToApply(location, eventId)
-            .process(event, operationExecutionFunction, () -> resolutionResult, location, executorCallback);
+
+        ExecutorCallback effectiveCallback =
+            isTargetWithPolicies(event) ? getExecutionCallbackForPolicyAndOperationWithTarget(event, executorCallback)
```
marianogonzalez, line 429: `isTargetWithPolicies(event)` is it really necessary to evaluate this method on every single invocation? Won't the value be constant? Cannot this be cached?

balbifm: No, which policies get applied to a particular operation depends on the incoming event

PR #8836, opened 2020-05-04
MULE-18013: Cursor position is not reset on reconnection for operations that have an InputStream parameter as TypedValue, in a ParameterGroup or in a Map.

SebaElizalde:


core-tests/src/test/java/org/mule/runtime/core/internal/streaming/bytes/CursorStreamProviderTestCase.java
```diff
@@ -48,7 +48,7 @@
 @RunWith(Parameterized.class)
 @SmallTest
 @Feature(STREAMING)
-public class CursorStreamProviderTestCase extends AbstractByteStreamingTestCase {
+public class \CursorStreamProviderTestCase extends AbstractByteStreamingTestCase {
```
marianogonzalez, line 51: typo

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -976,11 +977,31 @@ private void addConnectionInterceptors(InterceptorChain.Builder chainBuilder) {
...
     componentModel.getAllParameterModels().forEach(
-                                                   p -> getType(p.getType(), getClassLoader(extensionModel))
-                                                       .filter(clazz -> InputStream.class.isAssignableFrom(clazz)
-                                                           || Iterator.class.isAssignableFrom(clazz))
-                                                       .ifPresent(clazz -> streamParams.add(p.getName())));
-
+                                                   p -> {
+                                                     getType(p.getType(), getClassLoader(extensionModel))
+                                                         .filter(clazz -> InputStream.class.isAssignableFrom(clazz)
+                                                             || Iterator.class.isAssignableFrom(clazz))
```
SebaElizalde, line 983: What case is the "Iterator" covering? Can an operation receive an Iterator as a parameter directly? Could not find tests for this case. What would be an example so I can add one?

PR #8837, opened 2020-05-04
MULE-18380: Add a tolerance time to wait connections to be closed on mule shutdown

eze210:


modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/GrizzlyServer.java
```diff
@@ -90,8 +125,8 @@ public RequestHandlerManager addRequestHandler(ListenerRequestMatcher listenerRe
         return this.listenerRegistry.addRequestHandler(this, requestHandler, listenerRequestMatcher);
     }
 
-    public void cleanIdleConnections()
+    public void setCleanIdleConnections()
```
fsgonz, line 128: Is this the only case where this method is used?

modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/GrizzlyServer.java
```diff
@@ -45,6 +50,22 @@ public synchronized void start() throws IOException
...
+                synchronized (openConnectionsSync) {
+                    openConnections += 1;
+                }
+                clientConnection.getMonitoringConfig().addProbes(new ConnectionProbe.Adapter()
+                {
+                    @Override
+                    public void onCloseEvent(Connection connection)
+                    {
+                        synchronized (openConnectionsSync) {
+                            openConnections -= 1;
```
fsgonz, line 62: maybe an AtomicInteger with a decrementAndGet would suffice. Is that correct?

modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/GrizzlyServer.java
```diff
@@ -45,6 +50,22 @@ public synchronized void start() throws IOException
             public void onAcceptEvent(Connection serverConnection, Connection clientConnection)
             {
                 clientConnection.getAttributes().setAttribute("Server Connection", serverConnection);
+                synchronized (openConnectionsSync) {
```
fsgonz, line 53: maybe there's the possibility to use an atomicInteger.

modules/http/src/main/java/org/mule/module/http/internal/listener/DefaultHttpListenerConfig.java
```diff
@@ -320,11 +320,11 @@ public synchronized void stop() throws MuleException
 
     public void stopServer()
     {
-        server.stop();
         if (server instanceof GrizzlyServer)
         {
-            ((GrizzlyServer) server).cleanIdleConnections();
+            ((GrizzlyServer) server).setCleanIdleConnections();
```
fsgonz, line 325: I would not use setCleanIdleConnections as the name of the function without parmeters. Maybe add a parameter.

modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/GrizzlyServer.java
fsgonz: It's ok. Maybe it's clearer this way.

modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/GrizzlyServer.java
fsgonz: Added for this.

PR #8845, opened 2020-05-07
MULE-18294: Polling source schedulers must not restart on reconnections

ndinu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -196,13 +196,14 @@ private synchronized void createSource(boolean restarting) throws Exception {
     }
   }
 
-  private void startSource(boolean restarting) throws MuleException {
+  private void startSource(boolean restarting, Map<String, Object> restartingContext) throws MuleException {
```
marianogonzalez, line 199: don't use maps. Build a explicit pojo.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -133,7 +137,13 @@ public void onStart(SourceCallback<T, A> sourceCallback) throws MuleException {
         .withName(formatKey("executor")));
 
     stopRequested.set(false);
-    scheduler.schedule(executor, () -> poll(sourceCallback));
+    if (restarting.compareAndSet(true, false)) {
+      delegateRunnable.setDelegate(() -> poll(sourceCallback));
```
marianogonzalez, line 141: I don't understand the point of this... if this guy is reconnecting is because it failed while executing. That means that you first simply `poll()` and then simply ` scheduler.schedule(executor, () -> poll(sourceCallback));`... everything else you added seems unnecessary. What am I missing?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -133,7 +137,13 @@ public void onStart(SourceCallback<T, A> sourceCallback) throws MuleException {
...
+      poll(sourceCallback);
+    } else {
+      delegateRunnable = new DelegateRunnable(() -> poll(sourceCallback));
```
marianogonzalez, line 144: what's the point of this?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -615,4 +648,22 @@ private void release() {
       }
     }
   }
+
+  private class DelegateRunnable implements Runnable {
```
marianogonzalez, line 652: this class does literally nothing

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -204,6 +217,25 @@ private int compareWatermarks(Serializable w1, Serializable w2, Comparator compa
...
+    restartingContext.put(RUNNABLE_KEY, delegateRunnable);
+    return restartingContext;
+  }
+
+  @Override
+  public void finishRestart(Map<String, Object> restartingContext) {
+    restarting.set(true);
+
+    executor = (org.mule.runtime.api.scheduler.Scheduler) restartingContext.get(POLLING_SOURCE_EXECUTOR_KEY);
+    delegateRunnable = (DelegateRunnable) restartingContext.get(RUNNABLE_KEY);
```
marianogonzalez, line 236: although I insist this is not necesary, this isn't a bad idea as the delegate you're keeping from the prior instance is quite likely to have references to the old instance...potentially producing leaks

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/Restartable.java
```diff
@@ -0,0 +1,35 @@
...
+ * @since 4.2.3 4.3.1 4.4.0
+ */
+public interface Restartable {
+
+  /**
+   * Method that needs to be called when a restart is performed.
+   *
+   * @return a context needed after a restart is performed.
+   */
+  Map<String, Object> beginRestart();
```
marianogonzalez, line 27: rename to `getRestartContext` and turn that map into an actual `SourceRestartContext` class

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/Restartable.java
```diff
@@ -0,0 +1,35 @@
...
+
+  /**
+   * Method that needs to be called to finish the process of restarting.
+   *
+   * @param restartingContext the context generated by {@link #beginRestart()}
+   */
+  void finishRestart(Map<String, Object> restartingContext);
```
marianogonzalez, line 34: rename to `restart()`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
ndinu: In the end I would need to use a POJO with at least a field that is a map, right? What is stored in the context is specific to the restarting of a polling source.

marianogonzalez: why? right now you're using a map on which you only put two entries with well known keys. Just use a pojo with two well known fields... If in the future you need a map, you add a map, but I don't see why you would need that and you certainly don't need it now

ndinu: My point is, the Restartable interface needs to be generic, not everything that can be restarted needs a Runnable and a Scheduler. Does it make sense?

marianogonzalez: but this is internal. We can change it as much as we want.. Rename it to `RestartableSource` if you want

ndinu:
--validate

ndinu:
--validate

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/source/PollingSourceTestCase.java
```diff
@@ -122,6 +135,37 @@ public void whenReconnectingAfterConnectionExceptionSchedulerRunsWithoutStartDel
     assertAllPetsAdopted();
   }
 
+  @Description("This test reflects a behavior that we must preserve, when a polling source is stopped and started the scheduler must be rescheduled.")
```
marianogonzalez, line 138: I get what you mean, but the term rescheduled is quite misleading

modules/extensions-spring-support/src/test/resources/polling-source-config.xml
```diff
@@ -68,7 +68,7 @@
                 <fixed-frequency frequency="1000"/>
             </scheduling-strategy>
         </petstore:pet-adoption-source>
-        <raise-error type="TEST:EXPECTED"/>
+        <test:processor throwException="true"/>
```
marianogonzalez, line 71: why this change?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/DelegateRunnable.java
```diff
@@ -0,0 +1,55 @@
...
+  private Runnable delegate;
+  private Lock lock = new ReentrantLock();
+
+  public DelegateRunnable(Runnable delegate) {
+    this.delegate = delegate;
+  }
+
+  @Override
+  public void run() {
+    if (lock.tryLock()) {
```
marianogonzalez, line 37: add a comment explaining why

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -204,6 +216,21 @@ private int compareWatermarks(Serializable w1, Serializable w2, Comparator compa
     return comparator.compare(w1, w2);
   }
 
+  @Override
+  public RestartContext getRestartContext() {
+    restarting.set(true);
```
marianogonzalez, line 221: a get method should never have side effects

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -204,6 +216,21 @@ private int compareWatermarks(Serializable w1, Serializable w2, Comparator compa
...
+    delegateRunnable.setDelegate(null);
+    return new RestartContext(executor, delegateRunnable);
+  }
+
+  @Override
+  public void restart(RestartContext restartContext) {
+    restarting.set(true);
+
+    executor = restartContext.getExecutor();
+    delegateRunnable = restartContext.getDelegateRunnable();
```
marianogonzalez, line 231: this is still wrong. If you do this here.. the poll() method that this delegates into is the poll from the old instance you just discarded. this should be
```
delegateRunnable = restartContext.getDelegateRunnable();
delegateRunnable.setDelegate(() -> poll());
```

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/source/PollingSourceTestCase.java
```diff
@@ -51,6 +58,12 @@ protected String getConfigFile() {
     return "polling-source-config.xml";
   }
 
+  @Before
+  public void resetCounters() throws Exception {
+    PetFailingPollingSource.STARTED_POLLS = 0;
```
marianogonzalez, line 63: This test is yielding false positives. The test pases because the number of invokations you want is correct, but you're not seeing that the instance that is actually modifying this is not the one you wanted. Change this by a Map<Source, Integer> and also assert which instances had how many invokations

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -204,6 +216,21 @@ private int compareWatermarks(Serializable w1, Serializable w2, Comparator compa
     return comparator.compare(w1, w2);
   }
 
+  @Override
+  public RestartContext getRestartContext() {
+    restarting.set(true);
+    delegateRunnable.setDelegate(null);
```
marianogonzalez, line 222: this should happen on the restart method, not here

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/RestartContext.java
```diff
@@ -0,0 +1,33 @@
...
+import org.mule.runtime.api.scheduler.Scheduler;
+
+/**
+ * Context needed to perform the restart of a source
+ *
+ * @since 4.2.3 4.3.1 4.4.0
+ */
+public class RestartContext {
+
+  private Scheduler executor;
```
marianogonzalez, line 18: both fields final

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
ndinu: At this point the delegate is null, I can set it to null here instead of in the getRestartContext method, but I cannot set the delegate here since the poll method needs the SourceCallbackContext which is not available yet, we get it in the doStart method.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
ndinu: The source need to know that a restart is coming so that it does not kill the scheduler. Maybe we should change the name of the method?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/DelegateRunnable.java
marianogonzalez: this comment is unsufficient. The main reason for this was to make sure that the delegate is not invoked if the scheduler fires while the restart is in progress

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
marianogonzalez: you can change the mehthod name, but still, is this the right place for doing this?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
marianogonzalez: then it is clear that this should happen in the doStart() method instead of here.........

PR #8847, opened 2020-05-08
MULE-18397: Error when using expression for targetValue

elrodro83:


modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/processor/ModuleOperationMessageProcessor.java
```diff
@@ -162,7 +162,9 @@ public ModuleOperationMessageProcessor(Map<String, String> properties,
 
     for (ParameterModel parameterModel : parameterModels) {
       final String parameterName = parameterModel.getName();
-      if (parameters.containsKey(parameterName)) {
+      if (parameterName.equals("target") || parameterName.equals("targetValue")) {
```
gsfernandes, line 165: Don't we have a constant somewhere for this?

elrodro83: i searched for it. The answer will surprise you.

marianogonzalez: org.mule.runtime.extension.api.ExtensionConstants#TARGET_PARAMETER_NAME

PR #8848, opened 2020-05-11
MULE-18319: Remove root/parent from ComponentModel

elrodro83:
* Fix NPE in `property` processing in LazyInit


elrodro83:
--validate

elrodro83:
--validate

PR #8852, opened 2020-05-12
MULE-17419: XML SDK: macroexpansion while keeping AST immutable

elrodro83:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/model/ApplicationModel.java
```diff
@@ -331,35 +323,67 @@ public ApplicationModel(ArtifactConfig artifactConfig, ArtifactDeclaration artif
...
-                                                                                                       runtimeMode));
+    final ArtifactAst previousAst = ast;
+    ast = new MacroExpansionModulesModel(previousAst, extensionModels).expand();
 
+    if (ast != previousAst) {
       // Have to index again the component models with macro expanded ones
-      indexComponentModels();
-    });
+      // TODO MULE-17197 (AST) uncomment this
+      // indexComponentModels();
```
gsfernandes, line 385: Meantime, would this break anything? I guess it was just a matter of performance to have them indexed, right?

elrodro83: from what i've seen, the maps populated by this method is used in the dsl processing. Now, there is another question: why does the dsl processing of an app care about its macroexpanded components? Maybe this method did something else once and now it makes no sense to call it here?

elrodro83:
--validate

PR #8864, opened 2020-05-14
MULE-18407: Remove `type` from ComponentModel

elrodro83:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/model/ComponentModel.java
```diff
@@ -485,530 +424,6 @@ public void setComponentType(TypedComponentIdentifier.ComponentType componentTyp
...
-                               model,
-                               parameterModel -> parameterModelFilter.test(parameterModel)
-                                   && !inlineGroupedParameters.contains(parameterModel));
-      }
-
-    });
-
-    // Last resort to try to find a matching metadata type for this component
-    if (!getModel(HasStereotypeModel.class).isPresent()) {
-      extensionModelHelper.findMetadataType(getType())
```
gsfernandes, line 566: How did you do now with these cases? This was relying on the Type (class) to build the metadataTypeModelAdapter for the parameterAst... how does this work from now on?

elrodro83: I moved this here: https://github.com/mulesoft/mule/pull/8864/files#diff-343feb8864aee8603a8a65aa05f549a3R162

PR #8869, opened 2020-05-19
MULE-18419: Race condition in FluxSink when events are buffered and the

fsgonz:
flux is completed.

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/rx/FluxSinkRecorder.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/rx/FluxSinkRecorder.java
@@ -33,23 +33,25 @@ public class FluxSinkRecorder<T> implements Consumer<FluxSink<T>> {
   @Override
   public void accept(FluxSink<T> fluxSink) {
     FluxSinkRecorderDelegate<T> previousDelegate = this.delegate;
-    delegate = new DirectDelegate<>(fluxSink);
-    previousDelegate.accept(fluxSink);
+    synchronized (this) {
+      delegate = new DirectDelegate<>(fluxSink);
+      previousDelegate.accept(fluxSink);
+    }
   }
 
-  public FluxSink<T> getFluxSink() {
+  public synchronized FluxSink<T> getFluxSink() {
     return delegate.getFluxSink();
   }
 
-  public void next(T response) {
+  public synchronized void next(T response) {
     delegate.next(response);
   }
 
-  public void error(Throwable error) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mule/runtime/core/internal/rx/FluxSinkRecorder.java
```diff
@@ -33,23 +33,25 @@
   @Override
   public void accept(FluxSink<T> fluxSink) {
     FluxSinkRecorderDelegate<T> previousDelegate = this.delegate;
```
marianogonzalez, line 35: shouldn't this be synchronized as well?

core/src/main/java/org/mule/runtime/core/internal/rx/FluxSinkRecorder.java
```diff
@@ -33,23 +33,25 @@
...
-    delegate = new DirectDelegate<>(fluxSink);
-    previousDelegate.accept(fluxSink);
+    synchronized (this) {
+      delegate = new DirectDelegate<>(fluxSink);
+      previousDelegate.accept(fluxSink);
```
marianogonzalez, line 38: shouldn't this be outside the critical region?

core/src/main/java/org/mule/runtime/core/internal/rx/FluxSinkRecorder.java
```diff
@@ -33,23 +33,25 @@
...
+    }
   }
 
-  public FluxSink<T> getFluxSink() {
+  public synchronized FluxSink<T> getFluxSink() {
```
marianogonzalez, line 42: have you meassured the perf impact of synchonizing all these methods?

fsgonz: I'm on it. I've run the tests in 4.2.2

core/src/main/java/org/mule/runtime/core/internal/rx/FluxSinkRecorder.java
fsgonz: This is the change that prevents the race conditon.

core/src/main/java/org/mule/runtime/core/internal/rx/FluxSinkRecorder.java
fsgonz: It is synchronized inside the method.

PR #8874, opened 2020-05-20
MULE-18425: Add types support to `ExtensionWalker`

elrodro83:
* Some additional cleanup on nearby code


modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/ExtensionModelHelper.java
```diff
@@ -382,14 +383,30 @@ protected void onNestable(ComposableModel owner, NestableElementModel model) {
               }
             }
 
+            @Override
+            protected void onType(MetadataType type) {
+              final Optional<DslElementSyntax> dslElement = resolveDslElementModel(type, currentExtension);
```
marianogonzalez, line 388: what if the same type is found several times?

elrodro83: will handle this in the walker itself

PR #8876, opened 2020-05-22
MULE-18434: Previous policy operation being propagated when requester…

Pgarrett:
… fails


Pgarrett:
Moved to branch from fix/mule-18434-mule-4.2.x

PR #8879, opened 2020-05-22
MULE-18433: Add test for validating that policies are not being appli…

JaatadiaMulesoft:
…ed to operation originated from a policy.


aiannucci:
--validate

PR #8882, opened 2020-05-26
MULE-18439: AST Loop detection should considerate error handlers cases

mbuchwald:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/model/ComponentModel.java
```diff
@@ -174,6 +175,8 @@ public ComponentLocation getLocation() {
         .orElse(getIdentifier().equals(ON_ERROR_CONTINE_IDENTIFIER)
             || getIdentifier().equals(ON_ERROR_PROPAGATE_IDENTIFIER))) {
       return ofNullable(getRawParameters().get(NAME_ATTRIBUTE_NAME));
+    } else if (getIdentifier().equals(ERROR_HANDLER_IDENTIFIER) && getRawParameters().get("ref") != null) {
```
elrodro83, line 178: `getRawParameters().get("ref") != null` -> `getRawParameterValue("ref").isPresent()`

PR #8884, opened 2020-05-26
MULE-18440: Fix race condition when looking for policies and invalida…

balbifm:
…ting cache at the same time


balbifm:
--validate

PR #8885, opened 2020-05-26
MULE-18442: Avoid creation of DefaultPolicyManager for each applied P…

balbifm:
…olicy

### DIFF
--- a/modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/ApplicationPolicyDeploymentTestCase.java
+++ b/modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/ApplicationPolicyDeploymentTestCase.java
@@ -93,6 +93,7 @@ public class ApplicationPolicyDeploymentTestCase extends AbstractDeploymentTestC
 
   @Parameterized.Parameters(name = "Parallel: {0}")
   public static List<Boolean> params() {
+    // Only run without parallel deployment since this configuration does not affect policy deployment at all
     return asList(false);
   }
 
--- a/modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/ApplicationPolicyDeploymentTestCase.java
+++ b/modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/ApplicationPolicyDeploymentTestCase.java
@@ -159,6 +159,7 @@ public class ApplicationPolicyDeploymentTestCase extends AbstractDeploymentTestC
   }
 
   @Test
+  @Issue("MULE-18442")
   public void applicationPolicyHasNoOpPolicyManagerInjected() throws Exception {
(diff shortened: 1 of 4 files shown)


modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/ApplicationPolicyDeploymentTestCase.java
```diff
@@ -83,6 +90,12 @@
 
   private static File simpleExtensionJarFile;
   private static File withErrorDeclarationExtensionJarFile;
+
+  @Parameterized.Parameters(name = "Parallel: {0}")
+  public static List<Boolean> params() {
```
elrodro83, line 95: remove this

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/ApplicationPolicyDeploymentTestCase.java
```diff
@@ -119,6 +132,30 @@ public void appliesApplicationPolicy() throws Exception {
     doApplicationPolicyExecutionTest(parameters -> true, 1, POLICY_PROPERTY_VALUE);
   }
 
+  @Test
+  public void applicationPolicyHasNoOpPolicyManagerInjected() throws Exception {
```
elrodro83, line 136: `@Issue`

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/ApplicationPolicyRedeploymentTestCase.java
```diff
@@ -35,6 +39,11 @@
   private static final String OS_POLICY_ID = "object-store-policy";
   public static final String FLOW_NAME = "main";
 
+  @Parameterized.Parameters(name = "Parallel: {0}")
+  public static List<Boolean> params() {
```
elrodro83, line 43: remove

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/ApplicationPolicyDeploymentTestCase.java
balbifm: Don't think we should remove. We are running policy tests twice for parallel and non-parallel deployment when that does not affect policies at all.

balbifm:
--validate

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/ApplicationPolicyDeploymentTestCase.java
elrodro83: can you add that as a comment in code?

PR #8929, opened 2020-06-16
MULE-18493: HTTP Connector leaking sockets in state CLOSE_WAIT when client uses HTTP/1.0

pabloperalta:


pabloperalta:
--validate

IvanAndresFritzler:
--validate

IvanAndresFritzler:
--validate

modules/http/src/main/java/org/mule/module/http/internal/listener/grizzly/ResponseDeferringCompletionHandler.java
```diff
@@ -98,9 +99,12 @@ public void completed(WriteResult result)
 
   private void doComplete()
   {
-    responseStatusCallback.responseSendSuccessfully();
-    ctx.notifyDownstream(RESPONSE_COMPLETE_EVENT);
-    resume();
+    if(!isCompleted) {
```
eze210, line 102: Use an AtomicBoolean and compareAndSet

pabloperalta: Hecho

PR #8935, opened 2020-06-18
MULE-18481: add system property to difference old feature with the new fix

jcgrethe:


core/src/main/java/org/mule/processor/IdempotentRedeliveryPolicy.java
```diff
@@ -285,10 +285,16 @@ private String getIdForEvent(MuleEvent event) throws Exception
     {
         if (useSecureHash)
         {
+            byte[] bytes;
             MuleMessage message = event.getMessage();
             Object payload = message.getPayload();
-            byte[] payloadBytes = (byte[]) objectToByteArray.transform(payload);
-            byte[] bytes = concat(payloadBytes, message.getUniqueId().getBytes());
+            if(Boolean.parseBoolean(System.getProperty("RollBackExceptionStrategy","false")))
```
mbuchwald: 2 corrections: 
1. Static imports
2. Have this as a constant, since it won't change its value during the entire execution (reading purposes, but more importantly avoid going to check system property per each time the code is executed)

core/src/main/java/org/mule/processor/IdempotentRedeliveryPolicy.java
```diff
@@ -285,10 +285,16 @@ private String getIdForEvent(MuleEvent event) throws Exception
...
+            {
+                byte[] payloadBytes = (byte[]) objectToByteArray.transform(payload);
+                bytes = concat(payloadBytes, message.getUniqueId().getBytes());
+            } else {
```
mbuchwald, line 296:
```suggestion
            } 
            else 
            {
```
Yeah, super bothersome but we should be consistent with the ugly use of braces in mule 3

tests/integration/src/test/java/org/mule/test/integration/exceptions/RollbackExceptionStrategyRedeliveryCountTestCase.java
```diff
@@ -34,6 +34,7 @@ protected String getConfigFile()
     @Test
     public void testRollbackExceptionStrategyNumberOfRetries() throws Exception
     {
+        System.setProperty("RollBackExceptionStrategy", "true");
```
mbuchwald: Make sure to set this as false in a @after, because if not other test might fail depending on the order of execution.

core/src/main/java/org/mule/processor/IdempotentRedeliveryPolicy.java
```diff
@@ -57,6 +59,7 @@
     private ObjectStore<AtomicInteger> store;
     private LockFactory lockFactory;
     private String idrId;
+    private boolean forceRedelivery;
```
mbuchwald, line 62:
```suggestion
    private static final boolean FORCE_REDELIVERY = getBoolean(MULE_FORCE_REDELIVERY);
```
(And change how it's used)

core/src/main/java/org/mule/processor/IdempotentRedeliveryPolicy.java
```diff
@@ -106,6 +109,7 @@ public void initialise() throws InitialisationException
         String flowName = flowConstruct.getName();
         idrId = String.format("%s-%s-%s",appName,flowName,"idr");
         lockFactory = muleContext.getLockFactory();
+        forceRedelivery = getBoolean(MULE_FORCE_REDELIVERY);
```
mbuchwald, line 112:
```suggestion
```

PR #8957, opened 2020-06-20
MULE-18489: CCE with @ConfigOverrides on params of different types

elrodro83:
* Avoid performance overhead when handling parameters of primitive types


elrodro83:
--validate

PR #8971, opened 2020-06-23
MULE-18514: XSD schema for metadata not found if located either in sub folders or the project itself

SebaElizalde:


SebaElizalde:
--validate

PR #8979, opened 2020-06-24
MULE-18509: Review flaky tests in branch support/4.2.x

szaffarano:


modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/ClassLoaderLeakTestCase.java
```diff
@@ -198,6 +199,7 @@ public void redeployPreviousAppEagerlyGCd() throws Exception {
 
   @Test
   @Issue("MULE-18480")
+  @Ignore("MULE-18520")
   @Description("When an artifact is redeployed through the deployment service by name, objects associated to the original deployment are released befroe deploying the new one.")
```
fsgonz, line 203: are released befroe -> are released before.

PR #8994, opened 2020-06-29
fix/MULE18534: Stored Procedure with CLOB In Out parameter fails to execute on Oracle

julodiaz:
Jenkin's Test: https://jenkins-onprem.build.msap.io/job/Mule-runtime/view/3.x.x/job/mule/view/change-requests/job/PR-8994/

### DIFF
--- a/modules/db/src/main/java/org/mule/module/db/internal/domain/type/ClobResolvedDataType.java
+++ b/modules/db/src/main/java/org/mule/module/db/internal/domain/type/ClobResolvedDataType.java
@@ -59,10 +59,10 @@ public class ClobResolvedDataType extends ResolvedDbType
                 }
 
                 super.setParameterValue(statement, index, clob);
-            } catch (Throwable t) {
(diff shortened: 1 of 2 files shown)


modules/db/src/main/java/org/mule/module/db/internal/domain/type/ClobResolvedDataType.java
```diff
@@ -29,26 +34,31 @@ public ClobResolvedDataType(int id, String name)
...
-            }
-            else
-            {
+        if (value != null && !(value instanceof Clob)) {
+            String valueString;
+
+            if (value instanceof String) {
+                valueString = (String) value;
+            } else if (value instanceof InputStream) {
+                valueString = IOUtils.toString((InputStream) value);
```
nicomz, line 43: Clob have a setCharacterStream method. In case of value is instanceOf inputStream you should use this method.

PR #8997, opened 2020-06-30
Backport to mule3.x: MULE-17863: Avoid file leak in TlsConfiguration

ca-stefan-cordes:
Avoid file leak when re-deploying:
https://www.mulesoft.org/jira/browse/MULE-17863


fsgonz:
Created a new PR with a test and naming conventions.

PR #9000, opened 2020-06-30
MULE-18542: Timer threads leak on redeploy when using Oracle Driver.

emariotti3:


modules/artifact/src/test/java/org/mule/runtime/module/artifact/classloader/ResourceReleaserTestCase.java
```diff
@@ -171,6 +176,33 @@ public void createsInstanceOnlyOnce() throws IOException {
...
+
+      @Override
+      public void run() {
+        System.out.println("Timer");
+      }
+    };
+    timerThread.schedule(task, 10, 1000);
+    killOracleThreadsMethod.invoke(resourceReleaser);
+
+    sleep(1000);
```
marianogonzalez, line 197: this test in inherently flaky.

modules/artifact/src/main/java/org/mule/module/artifact/classloader/JdbcResourceReleaser.java
```diff
@@ -45,6 +45,7 @@
 public class JdbcResourceReleaser implements ResourceReleaser {
 
   public static final String DIAGNOSABILITY_BEAN_NAME = "diagnosability";
+  public static final String ORACLE_DRIVER_TIMER_THREAD_NAME = "Timer";
```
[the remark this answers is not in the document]
emariotti3: I've looked into this, but threads do not appear to have any references to the driver once spawned. The only thing we could possibly check would by the class-loader type... other than that, the only way I found to make this less intrusive was to set that system property to disable killing of the threads.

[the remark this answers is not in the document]
fsgonz: Maybe we should check if there a way to ask the driver to remove this threads on disposal. Timer is too generic a name. Not sure if there is a way for this. Maybe verifying how the driver behaves on shutdown regarding those threads and attempt to replicate that behavior on redeployment.

modules/artifact/src/main/java/org/mule/module/artifact/classloader/JdbcResourceReleaser.java
```diff
@@ -264,4 +273,61 @@ private void shutdownMySqlConnectionCleanupThreads(Class<?> classAbandonedConnec
...
+      try {
+        enumerate(threads);
+      } catch (Throwable t) {
+        logger
+            .debug("An error occurred trying to obtain available Threads. Thread cleanup will be skipped.", t);
+        return;
+      }
+
+      for (Thread thread : threads) {
+        if (thread.getClass().getSimpleName().equals(ORACLE_DRIVER_TIMER_THREAD_CLASS_NAME)
```
fsgonz, line 292: refactor this as a method.

modules/artifact/src/main/java/org/mule/module/artifact/classloader/JdbcResourceReleaser.java
```diff
@@ -44,7 +48,11 @@
  */
 public class JdbcResourceReleaser implements ResourceReleaser {
 
+  private static final String avoidThreadDisposal = getProperty("avoid.dispose.oracle.threads");
```
fsgonz, line 51: use Boolean.getBoolean which verifies if there exists a system property, otherwise it assigns a default.
Use a constant for the string and rename this as jdb.resource.releaser.avoid.dispose.oracle.threads

modules/artifact/src/main/java/org/mule/module/artifact/classloader/JdbcResourceReleaser.java
```diff
@@ -264,4 +273,61 @@ private void shutdownMySqlConnectionCleanupThreads(Class<?> classAbandonedConnec
...
+                       e);
+          }
+        }
+      }
+    } catch (Exception e) {
+      logger.debug(e.getMessage());
+    }
+  }
+
+  private boolean isThreadLoadedByThisClassLoader(ClassLoader threadBaseClassLoader) {
```
fsgonz, line 310: What you have to verify is that th context classloader is the one from the app.

modules/artifact/src/main/java/org/mule/module/artifact/classloader/JdbcResourceReleaser.java
```diff
@@ -48,7 +49,9 @@
  */
 public class JdbcResourceReleaser implements ResourceReleaser {
 
-  private static final String avoidThreadDisposal = getProperty("avoid.dispose.oracle.threads");
+  private static final String AVOID_DISPOSE_ORACLE_THREADS_PROPERTY_NAME = "avoid.dispose.oracle.threads";
+  private static final boolean JDBC_RESOURCE_RELEASER_AVOID_DISPOSE_ORACLE_THREADS =
```
fsgonz, line 53: You should only use getBoolean without the getProperty

getBoolean(AVOID_DISPOSE_ORACLE_THREADS_PROPERTY_NAME)

Return true if and only if the system property is true.

modules/artifact/src/main/java/org/mule/module/artifact/classloader/JdbcResourceReleaser.java
```diff
@@ -264,4 +276,65 @@ private void shutdownMySqlConnectionCleanupThreads(Class<?> classAbandonedConnec
...
+      Thread[] threads = new Thread[activeCount()];
+      try {
+        enumerate(threads);
+      } catch (Throwable t) {
+        logger
+            .debug("An error occurred trying to obtain available Threads. Thread cleanup will be skipped.", t);
+        return;
+      }
+
+      for (Thread thread : threads) {
```
fsgonz, line 294: Add a comment to explain why this has to be done and that the timer cannot be cancelled by reflection (some users may use only the ojdbc jar)

modules/artifact/src/main/java/org/mule/module/artifact/classloader/JdbcResourceReleaser.java
```diff
@@ -264,4 +276,71 @@ private void shutdownMySqlConnectionCleanupThreads(Class<?> classAbandonedConnec
...
+        enumerate(threads);
+      } catch (Throwable t) {
+        logger
+            .debug("An error occurred trying to obtain available Threads. Thread cleanup will be skipped.", t);
+        return;
+      }
+
+      /* IMPORTANT: this is done to avoid metaspace OOM caused by oracle driver
+      thread leak. This is only meant to stop TimerThread threads spawned
+      by oracle driver's HAManger class. This timer cannot be fetched
```
marianogonzalez, line 296: `HAManger`-> `HAManager`

modules/artifact/src/main/java/org/mule/module/artifact/classloader/JdbcResourceReleaser.java
```diff
@@ -264,4 +276,71 @@ private void shutdownMySqlConnectionCleanupThreads(Class<?> classAbandonedConnec
...
+      would be required.
+      * */
+      for (Thread thread : threads) {
+        if (isThreadApplicationTimerThread(thread)) {
+          try {
+            thread.stop();
+            thread.interrupt();
+          } catch (Throwable e) {
+            logger
+                .debug("An error occurred trying to close the '" + thread.getName() + "' Thread. This might cause memory leaks.",
```
marianogonzalez, line 307: WARN

modules/artifact/src/main/java/org/mule/module/artifact/classloader/JdbcResourceReleaser.java
```diff
@@ -264,4 +276,71 @@ private void shutdownMySqlConnectionCleanupThreads(Class<?> classAbandonedConnec
...
+                       e);
+          }
+        }
+      }
+    } catch (Exception e) {
+      logger.debug(e.getMessage());
```
marianogonzalez, line 313: this should log not only the message, but the context in which it happened and the entire exception

modules/artifact/src/main/java/org/mule/module/artifact/classloader/JdbcResourceReleaser.java
```diff
@@ -264,4 +276,71 @@ private void shutdownMySqlConnectionCleanupThreads(Class<?> classAbandonedConnec
...
+    while (threadClassLoader != null) {
+      lastClassLoader = threadClassLoader;
+      // It has to be the same reference not equals to
+      if (threadClassLoader == resourceReleaserClassLoader) {
+        return true;
+      }
+      threadClassLoader = threadClassLoader.getParent();
+    }
+
+    while (resourceReleaserClassLoader != null) {
```
marianogonzalez, line 337: this doesn't make sense. If this is a concern, then the two while loops should be nested and this one should be the master iteration

modules/artifact/src/main/java/org/mule/module/artifact/classloader/JdbcResourceReleaser.java
```diff
@@ -264,4 +276,71 @@ private void shutdownMySqlConnectionCleanupThreads(Class<?> classAbandonedConnec
...
+    ClassLoader lastClassLoader = null;
+    ClassLoader resourceReleaserClassLoader = this.getClass().getClassLoader();
+
+    while (threadClassLoader != null) {
+      lastClassLoader = threadClassLoader;
+      // It has to be the same reference not equals to
+      if (threadClassLoader == resourceReleaserClassLoader) {
+        return true;
+      }
+      threadClassLoader = threadClassLoader.getParent();
```
marianogonzalez, line 334: this doesn't look correct. By doing this, you're completely bypassing the lookup policies. Why is this needed? In which case would the driver not be loaded by the extension classloader?

modules/artifact/src/main/java/org/mule/module/artifact/classloader/JdbcResourceReleaser.java
```diff
@@ -44,7 +49,13 @@
...
 public class JdbcResourceReleaser implements ResourceReleaser {
 
+  private static final String AVOID_DISPOSE_ORACLE_THREADS_PROPERTY_NAME = "avoid.dispose.oracle.threads";
+  private static final boolean JDBC_RESOURCE_RELEASER_AVOID_DISPOSE_ORACLE_THREADS =
+      getBoolean(AVOID_DISPOSE_ORACLE_THREADS_PROPERTY_NAME);
+
   public static final String DIAGNOSABILITY_BEAN_NAME = "diagnosability";
+  public static final String ORACLE_DRIVER_TIMER_THREAD_NAME = "Timer-";
+  public static final String ORACLE_DRIVER_TIMER_THREAD_CLASS_NAME = "TimerThread";
   private final transient Logger logger = LoggerFactory.getLogger(getClass());
```
marianogonzalez, line 59: this logger should be static

modules/artifact/src/main/java/org/mule/module/artifact/classloader/JdbcResourceReleaser.java
```diff
@@ -264,4 +276,71 @@ private void shutdownMySqlConnectionCleanupThreads(Class<?> classAbandonedConnec
...
+  private boolean isThreadApplicationTimerThread(Thread thread) {
+    return thread.getClass().getSimpleName().equals(ORACLE_DRIVER_TIMER_THREAD_CLASS_NAME)
+        && thread.getName().contains(ORACLE_DRIVER_TIMER_THREAD_NAME)
+        && isThreadLoadedByThisClassLoader(thread.getContextClassLoader());
+  }
+
+  private boolean isThreadLoadedByThisClassLoader(ClassLoader threadBaseClassLoader) {
+    ClassLoader threadClassLoader = threadBaseClassLoader;
+    ClassLoader lastClassLoader = null;
+    ClassLoader resourceReleaserClassLoader = this.getClass().getClassLoader();
```
marianogonzalez, line 326: this looks incorrect. You need to check that the thread was started by the application being disposed, not by the classloader that loaded this class. Also, have you considered domains here?

modules/artifact/src/main/java/org/mule/module/artifact/classloader/JdbcResourceReleaser.java
```diff
@@ -264,4 +276,71 @@ private void shutdownMySqlConnectionCleanupThreads(Class<?> classAbandonedConnec
...
+      lastClassLoader = threadClassLoader;
+      // It has to be the same reference not equals to
+      if (threadClassLoader == resourceReleaserClassLoader) {
+        return true;
+      }
+      threadClassLoader = threadClassLoader.getParent();
+    }
+
+    while (resourceReleaserClassLoader != null) {
+      if (lastClassLoader == resourceReleaserClassLoader) {
```
marianogonzalez, line 338: same as before. Doesn't look like you're asserting the correct thing

modules/artifact/src/test/java/org/mule/runtime/module/artifact/classloader/ResourceReleaserTestCase.java
```diff
@@ -171,6 +178,36 @@ public void createsInstanceOnlyOnce() throws IOException {
     }
   }
 
+  @Test
+  public void stopOracleDriverLeakyThreads() throws Exception {
+    JdbcResourceReleaser resourceReleaser = new JdbcResourceReleaser();
+
+    Method killOracleThreadsMethod = resourceReleaser.getClass().getDeclaredMethod("disposeOracleDriverThreads");
```
marianogonzalez, line 185: don't do this. The runtime will use this by calling the release() method. Therefore, simply call the release method

modules/artifact/src/test/java/org/mule/runtime/module/artifact/classloader/ResourceReleaserTestCase.java
```diff
@@ -171,6 +178,36 @@ public void createsInstanceOnlyOnce() throws IOException {
...
+    killOracleThreadsMethod.setAccessible(true);
+
+    Timer timerThread = new Timer();
+    TimerTask task = new TimerTask() {
+
+      @Override
+      public void run() {
+        System.out.println("Timer");
```
marianogonzalez, line 193: no System.out in tests

modules/artifact/src/test/java/org/mule/runtime/module/artifact/classloader/ResourceReleaserTestCase.java
```diff
@@ -171,6 +178,36 @@ public void createsInstanceOnlyOnce() throws IOException {
...
+    timerThread.schedule(task, 10, 100000);
+
+    Thread[] threads = new Thread[activeCount()];
+    enumerate(threads);
+
+    killOracleThreadsMethod.invoke(resourceReleaser);
+
+    for (Thread thread : threads) {
+      if (thread.getName().contains(ORACLE_DRIVER_TIMER_THREAD_NAME)) {
+        check(15000, 100, () -> TERMINATED.equals(thread.getState()));
```
marianogonzalez, line 206: this is testing nothing:
1) You're not testing the classloading navigation part. This doesn't give any indication on wether this actually works with the DB connector
2) You're starting a timer, but at no point you gave it the correct name
3) You're not considering the possibility of the thread not being found and the check() never taking place.

modules/artifact/src/main/java/org/mule/module/artifact/classloader/JdbcResourceReleaser.java
```diff
@@ -264,4 +280,98 @@ private void shutdownMySqlConnectionCleanupThreads(Class<?> classAbandonedConnec
...
+      /* IMPORTANT: this is done to avoid metaspace OOM caused by oracle driver
+      thread leak. This is only meant to stop TimerThread threads spawned
+      by oracle driver's HAManager class. This timer cannot be fetched
+      by reflection because, in order to do so, other oracle dependencies
+      would be required.
+      * */
+      for (Thread thread : threads) {
+        if (isOracleTimerThread(undeployedArtifactClassLoader, thread)) {
+          try {
+            thread.stop();
```
elrodro83, line 309: https://stackoverflow.com/questions/16504140/thread-stop-deprecated

modules/artifact/src/main/java/org/mule/module/artifact/classloader/JdbcResourceReleaser.java
```diff
@@ -264,4 +280,98 @@ private void shutdownMySqlConnectionCleanupThreads(Class<?> classAbandonedConnec
     throw new ClassNotFoundException("No MySql's AbandonedConnectionCleanupThread class was found");
   }
 
+  private void disposeOracleDriverThreads() {
```
elrodro83, line 283: this method should not finish return until all the interrupted threads are dead, because they may take a while in actually finishing

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/DbDriverThreadLeakTestCase.java
```diff
@@ -0,0 +1,150 @@
...
+  private TestDeploymentListener deploymentListener;
+
+  public DbDriverThreadLeakTestCase(boolean parallellDeployment, String appName, String xmlFile) {
+    super(parallellDeployment);
+    this.appName = appName;
+    this.xmlFile = xmlFile;
+  }
+
+  @Test
+  public void oracleDriverTimerThreadsReleasedOnUndeploy() throws Exception {
```
elrodro83, line 51: `@Issue`

PR #9001, opened 2020-06-30
MULE-16688: When using properties placeholder on import tag deployment properties are not considered

mbuchwald:


mbuchwald:
--validate

PR #9019, opened 2020-07-02
MULE-18548: In memory object store used in cache should be able to be expired in secondary node

mbuchwald:


core/src/test/java/org/mule/util/store/InMemoryStoreTestCase.java
```diff
@@ -47,6 +50,28 @@ public void testSimpleTimedExpiry() throws Exception
...
+        store.setMuleContext(context);
+
+        // store entries in quick succession
+        storeObjects("1", "2", "3");
+
+        // they should still be alive at this point
+        assertObjectsInStore("1", "2", "3");
+
+        // wait until the entry TTL has been exceeded
+        Thread.sleep(entryTTL + 1000);
```
elrodro83, line 69: poller

PR #9030, opened 2020-07-06
MULE-16924: Support lightweight deployment with class loader model and local repository

fsgonz:


fsgonz:
This is a cherry-pick to fix a broker test in 4.2.1 branch.

PR #9031, opened 2020-07-07
CMTS-17: CacheId Generator for Value Providers without a complete DSL

lucianoRM:


modules/spring-config/src/test/resources/log4j2-test.xml
```diff
@@ -19,7 +19,7 @@
 
         <!-- Mule classes -->
         <AsyncLogger name="org.mule" level="WARN"/>
-        <AsyncLogger name="org.mule.runtime.config" level="WARN"/>
+        <AsyncLogger name="org.mule.runtime.config" level="DEBUG"/>
```
gsfernandes, line 22: Revert this change

PR #9039, opened 2020-07-11
MULE-18543: ComponentAst parameter provide a way to know if value com…

mlischetti:
…es from a default value or DSL

### DIFF
--- a/modules/spring-config/src/main/java/org/mule/runtime/config/internal/model/DefaultComponentParameterAst.java
+++ b/modules/spring-config/src/main/java/org/mule/runtime/config/internal/model/DefaultComponentParameterAst.java
@@ -173,4 +173,14 @@ public class DefaultComponentParameterAst implements ComponentParameterAst {
     return ofNullable(metadata);
   }
 
+  @Override
+  public boolean isDefault() {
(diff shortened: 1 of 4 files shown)


modules/spring-config/src/main/java/org/mule/runtime/config/internal/model/DefaultComponentParameterAst.java
```diff
@@ -173,4 +173,14 @@ public ParameterModel getModel() {
     return ofNullable(metadata);
   }
 
+  @Override
+  public boolean isDefault() {
+    return ofNullable(getRawValue()).map(v -> {
+      if (getModel() != null && getModel().getDefaultValue() != null) {
```
gsfernandes, line 179: I'm pretty sure this is also being done when value is set: getModel().getDefaultValue()
Wouldn't be better to set the isDefault=true when the value was set and populated from the getModel().getDefaultValue()? Or maybe I'm wrong and the default value already comes from the XSD.

gsfernandes: If the default value comes from the XSD we should leave this logic here to be based on the model definition.

PR #9079, opened 2020-07-20
MULE-17863: Avoid file leak in TlsConfiguration

fsgonz:


core/src/main/java/org/mule/api/security/tls/TlsConfiguration.java
```diff
@@ -324,7 +325,7 @@ protected KeyStore loadKeyStore() throws GeneralSecurityException, IOException
     {
         KeyStore tempKeyStore = KeyStore.getInstance(keystoreType);
 
-        InputStream is = IOUtils.getResourceAsStream(keyStoreName, getClass());
+       try(InputStream is = IOUtils.getResourceAsStream(keyStoreName, getClass())) {
```
eze210, line 328: Add new-line before {

PR #9086, opened 2020-07-21
MULE-18583: Exclude org.mule.tests:test-components in the enforcer pl…

anosenzo:
…ugin

### DIFF
--- a/modules/extensions-spring-support/pom.xml
+++ b/modules/extensions-spring-support/pom.xml
@@ -400,4 +400,47 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <profiles>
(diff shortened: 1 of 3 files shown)


modules/extensions-spring-support/pom.xml
```diff
@@ -400,4 +400,47 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <profiles>
+        <profile>
+            <id>release</id>
```
aiannucci, line 406: Is this necessary if we already have it in a parent POM?

anosenzo:
--validate

anosenzo:
--validate

PR #9092, opened 2020-07-21
MULE-18571: Migrate mule/tests/parsers-plugin components so that they use an SDK extension

elrodro83:
* Move pojos to test-components
* Rename remaining processor-chains stuff to test-processor-chains


tests/test-components/pom.xml
```diff
@@ -8,12 +8,12 @@
...
         <groupId>org.mule.extensions</groupId>
         <artifactId>mule-modules-parent</artifactId>
-        <version>1.1.6</version>
+        <version>1.3.0</version>
     </parent>
 
     <groupId>org.mule.tests</groupId>
     <artifactId>test-components</artifactId>
-    <version>1.0.0-SNAPSHOT</version>
+    <version>4.4.0-SNAPSHOT</version>
```
eze210, line 16: Isn't there some property?

elrodro83: not available here, since the parent pom is not the actual parent in the module hierarchy

PR #9094, opened 2020-07-21
MULE-18426: Interception API: Add a callback to source interceptor on…

JuLuzzi:
…ce all event contexts for an event are complete

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/processor/interceptor/AbstractCompletableInterceptorSourceCallbackAdapter.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/processor/interceptor/AbstractCompletableInterceptorSourceCallbackAdapter.java
@@ -132,6 +132,7 @@ public abstract class AbstractCompletableInterceptorSourceCallbackAdapter<T exte
           interceptor.afterCallback(component.getLocation(), interceptionEvent, thrown);
         } finally {
           setContextClassLoader(currentThread, ctxClassLoader, originalTCCL);
+          interceptor.afterTerminated(component.getLocation(), interceptionEvent, thrown);
         }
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/mule/runtime/core/internal/processor/interceptor/AbstractCompletableInterceptorSourceCallbackAdapter.java
```diff
@@ -132,6 +132,7 @@ public AbstractCompletableInterceptorSourceCallbackAdapter(SourceInterceptorFact
           interceptor.afterCallback(component.getLocation(), interceptionEvent, thrown);
         } finally {
           setContextClassLoader(currentThread, ctxClassLoader, originalTCCL);
+          interceptor.afterTerminated(component.getLocation(), interceptionEvent, thrown);
```
elrodro83, line 135: no, this should be done on a termination callback on the event context

core/src/main/java/org/mule/runtime/core/internal/processor/interceptor/AbstractCompletableInterceptorSourceCallbackAdapter.java
```diff
@@ -132,6 +132,7 @@ public AbstractCompletableInterceptorSourceCallbackAdapter(SourceInterceptorFact
           interceptor.afterCallback(component.getLocation(), interceptionEvent, thrown);
         } finally {
           setContextClassLoader(currentThread, ctxClassLoader, originalTCCL);
+          //interceptor.afterTerminated(component.getLocation(), event.getContext());
```
elrodro83, line 135: remove

PR #9096, opened 2020-07-22
MULE-18571: Migrate DependencyInjectionObject so that it uses an SDK

elrodro83:
extension


elrodro83:
--validate

elrodro83:
--validate

PR #9110, opened 2020-07-24
Fix/mule 18582 3.x

pabloperalta:
MULE-18582: Application stuck in "starting" status forever when SFTP configured to "reconnect forever" if connection fails on startup

Added the possibility of canceling a deployment that for now would basically stop al connector's retry policies from retrying. This method should be called before requesting the deploymentLock to stop an application.

### DIFF
--- a/core/src/main/java/org/mule/retry/policies/AbstractPolicyTemplate.java
+++ b/core/src/main/java/org/mule/retry/policies/AbstractPolicyTemplate.java
@@ -21,6 +21,7 @@ import org.mule.retry.notifiers.ConnectNotifier;
 
 import java.io.InterruptedIOException;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -36,6 +37,8 @@ public abstract class AbstractPolicyTemplate implements RetryPolicyTemplate, Mul
     private Map<Object, Object> metaInfo;
 
     private MuleContext muleContext;
+
+    private AtomicBoolean cancelStart = new AtomicBoolean(false);
 
     private volatile boolean lastRetryInterrupted = false;
 
@@ -98,7 +101,7 @@ public abstract class AbstractPolicyTemplate implements RetryPolicyTemplate, Mul
                     }
                 }
             }
-            while (status.isOk());
+            while (status.isOk() && !cancelStart.get());
 
             context.setFailed(cause);
 
@@ -122,7 +125,7 @@ public abstract class AbstractPolicyTemplate implements RetryPolicyTemplate, Mul
             }
         }
     }
-    
+
     public RetryNotifier getNotifier()
     {
         return notifier;
@@ -153,5 +156,16 @@ public abstract class AbstractPolicyTemplate implements RetryPolicyTemplate, Mul
     public boolean isSynchronous()
     {
         return true;
+    }
+
+    protected AtomicBoolean getCancelStart()
+    {
+        return cancelStart;
+    }
+
+    @Override
+    public void cancelStart()
+    {
+        cancelStart.compareAndSet(false, true);
     }
 }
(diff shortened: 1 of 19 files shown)


core/src/main/java/org/mule/retry/policies/AbstractPolicyTemplate.java
```diff
@@ -37,6 +38,8 @@
 
     private MuleContext muleContext;
 
+    private AtomicBoolean cancelStart = new AtomicBoolean(false);
```
IvanAndresFritzler, line 41: Shouldn't this be also named stopRetrying?

core/src/main/java/org/mule/retry/policies/AbstractPolicyTemplate.java
```diff
@@ -98,7 +101,7 @@ public RetryContext execute(RetryCallback callback, WorkManager workManager) thr
                     }
                 }
             }
-            while (status.isOk());
+            while (status.isOk() && !cancelStart.get());
```
IvanAndresFritzler, line 104: same (stopRetrying)

core/src/main/java/org/mule/retry/policies/AbstractPolicyTemplate.java
```diff
@@ -154,4 +157,15 @@ public boolean isSynchronous()
     {
         return true;
     }
+
+    protected AtomicBoolean getCancelStart()
```
IvanAndresFritzler, line 161: same (isStopRetrying)

modules/launcher/src/test/java/org/mule/module/launcher/domain/DefaultMuleDomainTestCase.java
```diff
@@ -0,0 +1,183 @@
...
+
+        // When cancelling start
+        defaultMuleDomain.cancelStart();
+
+        // Connectors's retry policy start is also cancelled
+        verify(retryPolicyTemplate).stopRetrying();
+    }
+
+    @Test
+    public void testCancelStartFiveConectorsWithRetryPolicies()
```
IvanAndresFritzler, line 83: Better to avoid magic numbers and stick to 2 if no extra logic (or cases) involved

modules/launcher/src/test/java/org/mule/module/launcher/domain/DefaultMuleDomainTestCase.java
```diff
@@ -0,0 +1,183 @@
...
+
+        // When cancelling start
+        defaultMuleDomain.cancelStart();
+
+        // Then no error occurs
+
+    }
+
+    @Test
+    public void testStopCancelStartFiveConectors()
```
IvanAndresFritzler, line 134: Same (stick to two)

core/src/main/java/org/mule/retry/policies/AbstractPolicyTemplate.java
pabloperalta: Applied the suggested changes

core/src/main/java/org/mule/retry/policies/AbstractPolicyTemplate.java
pabloperalta: Applied the suggested changes

core/src/main/java/org/mule/retry/policies/AbstractPolicyTemplate.java
pabloperalta: Applied the suggested changes

modules/launcher/src/test/java/org/mule/module/launcher/domain/DefaultMuleDomainTestCase.java
pabloperalta: Applied the suggested changes

modules/launcher/src/test/java/org/mule/module/launcher/domain/DefaultMuleDomainTestCase.java
pabloperalta: Applied the suggested changes

PR #9114, opened 2020-07-24
MULE-18196: Application fails to deploy when having same custom name space as that of custom policy applied

szaffarano:
Add feature flag


facunlk:
--validate

szaffarano:
--validate

PR #9125, opened 2020-07-27
MULE-18601: MetadataCacheKeyGenerator based on ArtifactAST is including

elrodro83:
parameters as part of the hash when it shouldn't


elrodro83:
--validate

PR #9134, opened 2020-07-29
MULE-18623: DynamicConfigurationProvider leaks configuration instances on cache eviction

eze210:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DynamicConfigurationProvider.java
```diff
@@ -217,7 +220,6 @@ protected void registerConfiguration(ConfigurationInstance configuration) {
   public List<ConfigurationInstance> getExpired() {
     return cache.asMap().entrySet().stream().filter(entry -> isExpired(entry.getValue())).map(entry -> {
       cache.invalidate(entry.getKey());
-      unRegisterConfiguration(entry.getValue());
```
marianogonzalez, line 220: why is this the problem? Who else can remove items from the cache?

eze210: There is a list in the parent class which isn't cleaned up

eze210: Sorry, the cache's removalListener

marianogonzalez: that's not what I asked. Your fix is to move this from here to the remove listener. My question is, why does that fix the problem? that makes sense if the item gets removed from the cache from any place other than this. Which place is that?

eze210: The cache has a time-based eviction policy (see `expireAfterAccess(expirationPolicy.getMaxIdleTime(), expirationPolicy.getTimeUnit())` )
It's automatically called on cache maintenance

eze210:
--validate

PR #9152, opened 2020-08-03
MULE-18619: Component models not set for infrastructure types in AST

elrodro83:


modules/extensions-xml-support/src/test/java/org/mule/runtime/extension/internal/loader/validation/DefaultModelValidatorTestCase.java
```diff
@@ -220,17 +221,17 @@ private ExtensionModel getExtensionModelFrom(String modulePath, Set<ExtensionMod
   private Set<ExtensionModel> getDependencyExtensions() {
     ExtensionModel petstore = loadExtension(PetStoreConnector.class, emptySet());
     ExtensionModel heisenberg = loadExtension(HeisenbergExtension.class, emptySet());
-    return ImmutableSet.<ExtensionModel>builder().add(petstore, heisenberg).build();
+    return ImmutableSet.<ExtensionModel>builder().add(petstore, heisenberg, getExtensionModel()).build();
   }
 
-  private ExtensionModel loadExtension(Class extension, Set<ExtensionModel> deps) {
+  private ExtensionModel loadExtension(Class extension, Set<ExtensionModel> extensions) {
```
marianogonzalez, line 227: should this be called depedencyExtensions or something like that? The intent of the parameter is now lost

modules/spring-config/src/main/java/org/mule/runtime/config/internal/model/type/ApplicationModelTypeUtils.java
```diff
@@ -236,38 +238,65 @@ private static void handleNestedParameters(ComponentModel componentModel, Stream
...
-              } else {
-                enrichComponentModels(componentModel, nestedComponents,
-                                      of(extensionModelHelper.resolveDslElementModel(paramModel,
-                                                                                     componentModel
-                                                                                         .getIdentifier())),
-                                      paramModel, extensionModelHelper);
-              }
-
-            }));
+    model.getAllParameterModels()
```
marianogonzalez, line 241: seems like nothing change here other than formatting? Is taht the case? Can it be reverted then?

modules/spring-config/src/main/java/org/mule/runtime/config/internal/model/type/ApplicationModelTypeUtils.java
```diff
@@ -309,26 +334,46 @@ private static void enrichComponentModels(ComponentModel componentModel,
...
+        });
+      }
+
+      @Override
+      public void visitUnion(UnionType unionType) {
+        unionType.getTypes()
+            .forEach(type -> getTypeId(type)
+                .flatMap(paramDsl::getChild)
+                .flatMap(childDsl -> getIdentifier(childDsl))
+                .flatMap(childComponentId -> ofNullable(nestedWrappedComponents.get(childComponentId)))
```
marianogonzalez, line 370: the flatMaps in a row?? ufff...

elrodro83: The revenge of the `Optional`s

modules/spring-config/src/main/java/org/mule/runtime/config/internal/model/type/ApplicationModelTypeUtils.java
elrodro83: no, i turned this around, before it iterated the children and matched those against the parameter models, now it iterates the parameter models and looks it up in the dsl.

PR #9155, opened 2020-08-04
MULE-18639: Simple type nested parameters are not added

fsgonz:


fsgonz:
Sorry. I have to verify something else

PR #9157, opened 2020-08-04
MULE-18639: Simple type nested parameters are not added

fsgonz:


fsgonz:
--validate

modules/spring-config/src/main/java/org/mule/runtime/config/internal/model/type/ApplicationModelTypeUtils.java
```diff
@@ -283,22 +284,49 @@ public void visitUnion(UnionType unionType) {
...
+                                                                             childComp.getMetadata())));
+                    }
+                  });
                 }
+
               });
         });
   }
 
+  private static boolean isSimpleMetadataType(ParameterModel paramModel) {
```
elrodro83, line 326: remove this unused method

fsgonz: Done

PR #9164, opened 2020-08-04
MULE-17447: ResourceReleaser should be loaded and invoked only if any of the classes to be released has been loaded - Fix test (#8222)

pabloperalta:
(cherry picked from commit 72b6503643654b024649268522f434c5b0599659)


pabloperalta:
--validate

PR #9165, opened 2020-08-04
CMTS-26: Validate and fix ValueProviderCacheIdGenerator for complex parameters

lucianoRM:


modules/spring-config/src/main/java/org/mule/runtime/config/api/dsl/model/metadata/ComponentBasedIdHelper.java
```diff
@@ -56,4 +61,50 @@ static String sourceElementNameFromSimpleValue(ComponentAst owner, ComponentPara
...
+      return null;
+    }
+
+    private Void hashForRight(Object o) {
+      if (o instanceof ComponentAst) {
+        final ComponentAst c = (ComponentAst) o;
+        c.getParameters().stream().sorted(comparing(p -> p.getModel().getName())).forEach(p -> {
+          hashBuilder.append(p.getModel().getName());
+          if (p.getModel().getType() instanceof ArrayType) {
+            hashForList((Collection<ComponentAst>) p.getValue().getRight());
```
gsfernandes, line 93: How about Map?

lucianoRM: Aren't maps ObjectTypes? It will call the same function again but with the value.

gsfernandes: You are right

lucianoRM:
Sonar fails because apparently we are accessing to an optional value without checking if it's present first. But we do, so there is no issue there.

PR #9172, opened 2020-08-06
MULE 18653: update Saxon

jcgrethe:


jcgrethe:
> shouldn't you need to do the same on the xml module?

update saxon? it use the version on the parent pom.

marianogonzalez:
> > shouldn't you need to do the same on the xml module?
> 
> update saxon? it use the version on the parent pom.

Check https://github.com/mulesoft/mule-xml-module/blob/3d43c52e752b26ab0292b45f95995c2703ff4477/pom.xml , the Mule 4 version of this carries its own version. Is it necessary to update that one as well?

jcgrethe:
> > > shouldn't you need to do the same on the xml module?
> > 
> > 
> > update saxon? it use the version on the parent pom.
> 
> Check https://github.com/mulesoft/mule-xml-module/blob/3d43c52e752b26ab0292b45f95995c2703ff4477/pom.xml , the Mule 4 version of this carries its own version. Is it necessary to update that one as well?

I don't think so. The problem was only in mule 3 and also as I see mule 4 uses an upper version.

PR #9177, opened 2020-08-07
CMTS-28: Support fetching values by using metadata key resolver

svacas:


modules/tooling-support-parent/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/config/InternalDeclarationSession.java
```diff
@@ -135,6 +140,24 @@ public ValueResult getValues(ParameterizedElementDeclaration component, String p
...
+    return artifactHelper()
+        .findComponentModel(component)
+        .map(cm -> {
+          Optional<ConfigurationInstance> configurationInstance =
+              ofNullable(component.getConfigRef()).flatMap(name -> artifactHelper().getConfigurationInstance(name));
+          MetadataKey metadataKey = buildMetadataKey(cm, component);
+          ClassLoader extensionClassLoader = getClassLoader(artifactHelper().getExtensionModel(component));
+          return withContextClassLoader(extensionClassLoader, () -> {
+            DefaultMetadataContext metadataContext = createMetadataContext(configurationInstance, extensionClassLoader);
+            return new MetadataMediator<>(cm).getMetadataKeys(metadataContext, metadataKey, new ReflectionCache());
```
gsfernandes, line 154: We have already a reflectionCache as an instance field. We could use that one.

modules/tooling-support-parent/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/config/InternalDeclarationSession.java
```diff
@@ -135,6 +140,24 @@ public ValueResult getValues(ParameterizedElementDeclaration component, String p
...
+          });
+        })
+        .orElseGet(() -> MetadataResult.success(MetadataKeysContainerBuilder.getInstance()
```
gsfernandes, line 157: I would say that in this case it should be a failure as the only case for this orElse is if the componentModel is not present or found for the declaration. It would mean that they were trying to resolve keys for a component that is not defined in any of the extension models declared for the session :S

gsfernandes: Please review the same case for types. It should fail if component is not found

modules/tooling-support-parent/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/config/InternalDeclarationSession.java
```diff
@@ -259,15 +273,13 @@ private MetadataKey buildMetadataKey(ComponentModel componentModel, ComponentEle
         .collect(toList());
   }
 
-  private <T extends ComponentModel> Map<String, Object> getComponentElementDeclarationParameters(ComponentElementDeclaration componentElementDeclaration,
-                                                                                                  T model) {
-    Map<String, Object> parametersMap = new HashMap<>();
+  private <T extends ComponentModel> Map<String, String> getParameterValues(ComponentElementDeclaration componentElementDeclaration,
```
gsfernandes, line 276: getMetadataKeyPartsValuesFromComponentDeclaration?

PR #9182, opened 2020-08-10
MULE-18609: Add record count/size of the streams to FlowConstructStatistics

elrodro83:


aiannucci:
--validate

core/src/main/java/org/mule/runtime/core/api/management/stats/AllStatistics.java
```diff
@@ -86,4 +94,36 @@ public synchronized void remove(FlowConstructStatistics stat) {
...
+                                               return statistics;
+                                             });
+  }
+
+  /**
+   * @param componentLocation the location of the component to get the statistics for.
+   * @return the statistics for the component with the provided {@code componentLocation}.
+   * @since 4.4, 4.3.1
+   */
+  public PayloadStatistics getPayloadStatistics(String componentLocation) {
```
szaffarano, line 126: It should be better to receive a Component instead of a String and do the transformation (`component.getLocation().getLocation()`) inside this method to be consistent with `computePayloadStatisticsIfAbsent`

elrodro83: This is on purpose, i don't need this method to be consistent because this method is to be used within ARM, and i don't want to force them to obtain a proper Component, when the may already know the location they want.

PR #9188, opened 2020-08-11
CMTS-15: Implement API for MTS to resolve values (ValueProviders) for a componentModel.parameter

lucianoRM:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/ValueProviderFactory.java
```diff
@@ -71,10 +72,20 @@ ValueProvider createValueProvider() throws ValueResolvingException {
       injectValueProviderFields(resolver);
 
       if (factoryModelProperty.usesConnection()) {
+        Object connection = connectionSupplier.get();
```
gsfernandes, line 75: Create a task on CMTS side to also do this validation on MTS side. If the resolver defines that requires a configuration/connection we should check first if the parameterizableElementDeclaration has the configRef value set and the configRef is part of the session. This should also apply for metadataTypes resolution.

modules/tooling-support-parent/functional-tests/src/test/java/org/mule/runtime/module/tooling/ComponentValueProviderTestCase.java
```diff
@@ -48,24 +86,96 @@ public void actingParameterGroup() {
     final List<String> listValue = asList("one", "two", "three");
     ComponentElementDeclaration elementDeclaration =
         actingParameterGroupOPDeclaration(CONFIG_NAME, stringValue, intValue, listValue);
-    getResultAndValidate(session, elementDeclaration, PROVIDED_PARAMETER_NAME, "stringValue-0-one-two-three");
+    validateValuesSuccess(session, elementDeclaration, PROVIDED_PARAMETER_NAME, "stringValue-0-one-two-three");
   }
 
   @Test
-  public void complexActingParameter() {
+  public void actingParameterGroupWithDefaultValue() {
```
gsfernandes, line 93: We have agreed on not supporting default values when resolving metadata keys or valueProviders. Change the code and test to assert the error here.

lucianoRM: Added TODO in MTS

lucianoRM:
--validate

PR #9208, opened 2020-08-13
MULE-18650: Implement the output stream data flow statistics decorator/factories

elrodro83:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -125,6 +127,9 @@
   @Inject
   private ClusterService clusterService;
 
+  @Inject
+  private PayloadStatisticsCursorDecoratorFactory payloadStatisticsCursorDecoratorFactory;
```
fsgonz, line 131: Maybe use an interface in this case. Just in case you want to override the default statistic cursor decorator in the registry for special envs?

elrodro83:
--validate

PR #9209, opened 2020-08-13
CMTS-53: Avoid runtime metatdata types resolution when key values are…

svacas:
… missing

### DIFF
--- a/modules/tooling-support-parent/functional-tests/src/test/java/org/mule/runtime/module/tooling/ComponentValueProviderTestCase.java
+++ b/modules/tooling-support-parent/functional-tests/src/test/java/org/mule/runtime/module/tooling/ComponentValueProviderTestCase.java
@@ -155,8 +155,7 @@ public class ComponentValueProviderTestCase extends DeclarationSessionTestCase {
                 .getDeclaration())
             .getDeclaration();
     validateValuesFailure(session, operationElementDeclaration, "anyParameter",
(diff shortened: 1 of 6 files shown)


modules/tooling-support-parent/functional-tests/src/test/java/org/mule/runtime/module/tooling/ComponentValueProviderTestCase.java
```diff
@@ -155,7 +155,8 @@ public void getValuesWithWrongDeclaration() {
                 .getDeclaration())
             .getDeclaration();
     validateValuesFailure(session, operationElementDeclaration, "anyParameter",
-                          "There is no extensionModel for extension: WrongExtension", UNKNOWN);
+                          "Unknown error while resolving values for parameter: 'anyParameter'. MuleRuntimeException: There is no extensionModel for extension: WrongExtension",
```
gsfernandes, line 158: Revert this as we need to understand why this error has changed.

modules/tooling-support-parent/functional-tests/src/test/java/org/mule/runtime/module/tooling/MetadataTypesTestCase.java
```diff
@@ -86,6 +87,7 @@ public void operationDynamicTypes() {
                    "}"));
   }
 
+  // TODO BUG Optional levels are required for multi-level keys!
```
gsfernandes, line 90: Create a ticket for this as we have discussed and add the Jira ticket id here.

PR #9215, opened 2020-08-19
Store Procedure execution fails and does not return data back to mule flow

dancazacu1:
Fixed StatementResultIterator.hasNext that was always true for stored procedures and resulted in OutOfMemoryException


modules/db/src/main/java/org/mule/module/db/internal/result/statement/StatementResultIterator.java
```diff
@@ -89,9 +89,11 @@ public boolean hasNext()
                     }
                 }
 
-                if (generatedKeys == null)
+                if (generatedKeys == null && !moveToNextResult() && statement.getUpdateCount() == NO_UPDATE_COUNT)
                 {
-                    moveToNextResult();
+                    updateCount = statement.getUpdateCount();
```
csimonelli, line 94: You are calling statement.getUpdateCount() twice in the same read. I think this change the statement state, so the second time it maybe will read again.

modules/db/src/test/java/org/mule/module/db/sql/executor/StatementResultIteratorTestCase.java
```diff
@@ -286,6 +293,28 @@ public void failsAfterGettingResultSetWhenStreamingIsUnsupported() throws Except
         iterator.next();
     }
 
+    @Test
```
csimonelli, line 296: I think it needs a test with a recordset with more than 1 result.

nicomz:
--validate

PR #9221, opened 2020-08-19
CMTS-30: Support for multi-level resolution on values, changed contra…

gsfernandes:
…ct to receive providerName instead of parameterName

### DIFF
--- a/modules/tooling-support-parent/functional-tests/src/test/java/org/mule/runtime/module/tooling/ComponentValueProviderTestCase.java
+++ b/modules/tooling-support-parent/functional-tests/src/test/java/org/mule/runtime/module/tooling/ComponentValueProviderTestCase.java
@@ -7,6 +7,9 @@
 package org.mule.runtime.module.tooling;
 
 import static java.util.Arrays.asList;
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.collection.IsCollectionWithSize.hasSize;
+import static org.junit.Assert.assertThat;
 import static org.mule.runtime.app.declaration.api.fluent.ElementDeclarer.newParameterGroup;
 import static org.mule.runtime.extension.api.values.ValueResolvingException.INVALID_VALUE_RESOLVER_NAME;
 import static org.mule.runtime.extension.api.values.ValueResolvingException.MISSING_REQUIRED_PARAMETERS;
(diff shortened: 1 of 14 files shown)


gsfernandes:
--validate

gsfernandes:
--validate

gsfernandes:
--validate

modules/tooling-support-parent/functional-tests/src/test/java/org/mule/runtime/module/tooling/ComponentValueProviderTestCase.java
```diff
@@ -167,4 +173,15 @@ public void getValuesOnParameterWithNoValueProvider() {
                           INVALID_VALUE_RESOLVER_NAME);
   }
 
+  @Test
+  public void multiLevelSourceValue() {
+    ComponentElementDeclaration elementDeclaration = sourceWithMultiLevelValue(CONFIG_NAME, "America", "USA");
+    // ProviderName is groupName
+    ValueResult valueResult = getValueResult(session, elementDeclaration, "values");
+    // ValueResult for multi level no matter if previous levels are set it would always return the whole tree
+    assertThat(valueResult.getValues(), hasSize(1));
```
svacas, line 182: just to understand, if the value provider returned more than one continent then the valueResult would contain a tree for each continent, right?

gsfernandes: No, it will have a ValueResult with N Values for each continent, each instance its children with countries and so on. Exactly as MetadataKeys but without partial fetch.
 
ValueProvider doesn't support partialFetch so, it will always return the whole tree as the resolution is not for the parameter but for the "resolverName"... which is not the providerId but it would be the parameterGroup.name instead of parameter.name.

Let me try to clarify this with an example:
Operation(name=pin, parameterGroup(name=locationGroup, parameters[continent, country, city]).
The Java code for the operation is 
void pin(@ ParameterGroup(name=locationGroup) @ OfValue(LocationValueProvider) Location location)

So, the user when configuring this operation in the GUI would have three parameters but as the Location class has three fields, continent, country, city.
The ValueProvider always return the whole tree as it doesn't support "partial fetch".

So, the extension model says in all the parameters, continent, country, city that the valueProviderModel has a providerName=locationGroup...meaning that when the valueProvider is invoked it should say: providerName=locationGroup. The value provider mediator is expecting to be invoked with the providerName instead of the parameter name. 
At the same time each parameter in addition to have the same providerName has a partOrder which defines it in which level of the tree should go its values.

So, we didn't want to "break" this contract from SDK and ValueProvider per se... and changed the values endpoint for MTS to be aligned with ValueProviderService and its mediator. So, it is up to MTS to do the matching for each level and value to filter out the result and have a dialect that works in terms of "parameter" instead of "providerName".

modules/tooling-support-parent/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/config/InternalDeclarationSession.java
```diff
@@ -241,16 +241,16 @@ public void dispose() {
...
                                                                                       ParameterValueResolver parameterValueResolver,
                                                                                       Optional<String> configName) {
-    ValueProviderMediator<T> valueProviderMediator = createValueProviderMediator(componentModel);
+    ValueProviderMediator<T> valueProviderMediator = createValueProviderMediator(containerModel);
     ExtensionResolvingContext context =
         new ExtensionResolvingContext(() -> configName.flatMap(name -> artifactHelper().getConfigurationInstance(name)),
                                       connectionManager);
     try {
-      return resultFrom(valueProviderMediator.getValues(parameterName,
+      return resultFrom(valueProviderMediator.getValues(providerName,
```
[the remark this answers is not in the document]
gsfernandes: No, it is expecting to receive the providerName as it does:
org.mule.runtime.module.extension.internal.value.ValueProviderMediator#getParameters.

PR #9233, opened 2020-08-20
MULE-18689: Error handler flux sinks leak when the error handler isn't executed

eze210:


eze210:
--validate

eze210:
--validate

PR #9245, opened 2020-08-21
CMTS-32: Improve error messages and validations

gsfernandes:


modules/tooling-support-parent/functional-tests/src/test/java/org/mule/runtime/module/tooling/ConnectivityTestingTestCase.java
```diff
@@ -46,7 +46,7 @@ public void testConnectionWrongConfigurationName() {
     ConnectionValidationResult connectionValidationResult = session.testConnection("invalidConfigName");
     assertThat(connectionValidationResult.isValid(), equalTo(false));
     assertThat(connectionValidationResult.getMessage(),
-               equalTo("Could not find a connection provider for configuration: 'invalidConfigName'"));
+               equalTo("Could not perform test connection for configuration: 'invalidConfigName' due to a connection provider is not defined"));
```
lucianoRM, line 49: Be sure to change this is you modify the message from the other PR

modules/tooling-support-parent/test-extension/src/test/java/org/mule/runtime/module/tooling/TestExtensionDeclarationUtils.java
```diff
@@ -310,15 +323,36 @@ public static OperationElementDeclaration multiLevelShowInDslGroupOPDeclaration(
...
     }
     return elementDeclarer.getDeclaration();
+
   }
 
   public static ComponentElementDeclaration<?> invalidComponentDeclaration() {
     return TEST_EXTENSION_DECLARER.newConstruct("invalid").getDeclaration();
   }
 
+  public static ComponentElementDeclaration<?> componentDeclarationWrongConfigRef() {
```
lucianoRM, line 341: I wouldn't add this type of methods here. In any case, add a method that accepts the configName you want to use. 
But this one seems too specific for the test case you are trying. 
At least the idea of this class, was to provide an easy way to declare all elements of the TestExtension. But not to test different cases. That should be the concern of the user class

modules/tooling-support-parent/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/config/InternalDeclarationSession.java
```diff
@@ -100,215 +63,129 @@
...
   InternalDeclarationSession(ArtifactDeclaration artifactDeclaration) {
     this.artifactHelperLazyValue =
         new LazyValue<>(() -> new ArtifactHelper(extensionManager, componentLocator, artifactDeclaration));
+    this.valueProviderExecutorLazyValue =
+        new LazyValue<>(() -> new ValueProviderExecutor(muleContext, connectionManager, expressionManager, reflectionCache,
+                                                        artifactHelper()));
+    this.metadataKeysExecutorLazyValue =
+        new LazyValue<>(() -> new MetadataKeysExecutor(connectionManager, reflectionCache, artifactHelper()));
+    this.metadataComponentExecutorLazyValue =
+        new LazyValue<>(() -> new MetadataComponentExecutor(connectionManager, reflectionCache, artifactHelper()));
```
lucianoRM, line 80: artifactHelper is  being initialized once any of these 3 lazyValues are. That should not be a problem but maybe it would be better for them to accept LazyValue<ArtifactHelper> ?

lucianoRM (changes requested):

modules/tooling-support-parent/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/config/InternalDeclarationSession.java
gsfernandes: Actually the artifactHelper is used before to validate extension model from declaration so at the point the executor is used the artifactHelper is already initialized.

modules/tooling-support-parent/test-extension/src/test/java/org/mule/runtime/module/tooling/TestExtensionDeclarationUtils.java
gsfernandes: Ok, I will change it.

lucianoRM (approved):

PR #9280, opened 2020-08-27
MULE-18047: Add the component that generated a cursor provider (support/4.2.2)

szaffarano:


core-tests/src/test/java/org/mule/runtime/core/internal/streaming/AbstractTroubleshootCursorProviderTestCase.java
```diff
@@ -0,0 +1,108 @@
...
+        {false, false},
+        {false, true},
+        {true, false},
+        {true, true}
+    };
+  }
+
+  @Before
+  public void before() throws NoSuchFieldException, IllegalAccessException {
+    setProperty(TRACK_CURSOR_PROVIDER_CLOSE_PROPERTY, trackStackTrace.toString());
```
elrodro83, line 64: use a SystemProperty rule instead

szaffarano: This was an old code that wasn't removed by mistake.  The cursor provider is set with this flag when instantiated (eg https://github.com/mulesoft/mule/blob/fix/MULE-18047/core-tests/src/test/java/org/mule/runtime/core/internal/streaming/bytes/TroubleshootCursorStreamProviderTestCase.java#L53)

PR #9312, opened 2020-08-31
MULE-18707: End of pipeline is spawned before all exceptions are inte…

fsgonz:
…rcepted in IdempotentRedeliveryPolicy


fsgonz:
--validate

PR #9327, opened 2020-08-31
CMTS-66: withContextClassLoader not correctly defined to propaget Val…

gsfernandes:
…uesResolvingException

### DIFF
--- a/modules/tooling-support-parent/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/value/ValueProviderExecutor.java
+++ b/modules/tooling-support-parent/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/value/ValueProviderExecutor.java
@@ -51,12 +51,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.function.Supplier;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 public class ValueProviderExecutor {
-
-  private static final Logger LOGGER = LoggerFactory.getLogger(ValueProviderExecutor.class);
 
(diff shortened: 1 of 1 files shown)


modules/tooling-support-parent/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/value/ValueProviderExecutor.java
```diff
@@ -91,12 +86,18 @@ public ValueResult resolveValues(ParameterizedModel parameterizedModel,
         return resultFrom(withContextClassLoader(extensionClassLoader, () -> valueProviderMediator.getValues(providerName,
                                                                                                              parameterValueResolver,
                                                                                                              connectionSupplier(context),
-                                                                                                             configSupplier(context))));
+                                                                                                             configSupplier(context)),
+                                                 ValueResolvingException.class,
```
lucianoRM, line 90: If you want, We have a utils method in org.mule.tooling.client.internal.utils.FunctionalUtils that wraps an exception from a Callable. So that the code is cleaner.

lucianoRM (changes requested):

modules/tooling-support-parent/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/value/ValueProviderExecutor.java
gsfernandes: That's on MTS side :(

lucianoRM (approved):

gsfernandes:
--validate

PR #9331, opened 2020-09-01
CMTS-68: DataWeaveExpressionLanguage should be loaded lazily when Dec…

gsfernandes:
…larationSession is deployed

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/api/config/MuleDeploymentProperties.java
+++ b/core/src/main/java/org/mule/runtime/core/api/config/MuleDeploymentProperties.java
@@ -14,9 +14,6 @@ public class MuleDeploymentProperties {
   public static final String DEPLOYMENT_PROPERTY_PREFIX = "mule.application.deployment";
   public static final String MULE_MUTE_APP_LOGS_DEPLOYMENT_PROPERTY = DEPLOYMENT_PROPERTY_PREFIX + ".muteLog";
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mule/runtime/core/api/config/MuleDeploymentProperties.java
```diff
@@ -14,9 +14,6 @@
   public static final String DEPLOYMENT_PROPERTY_PREFIX = "mule.application.deployment";
   public static final String MULE_MUTE_APP_LOGS_DEPLOYMENT_PROPERTY = DEPLOYMENT_PROPERTY_PREFIX + ".muteLog";
 
-  public static final String MULE_DISABLE_TEST_CONNECTION_ON_DEPLOYMENT_PROPERTY =
```
lucianoRM, line 17: This is API, I would not remove it

lucianoRM (changes requested):

core/src/main/java/org/mule/runtime/core/api/config/MuleDeploymentProperties.java
gsfernandes: I have added that in a previous commit, so, it was not released yet. It has been part of a timestamp based build but not yet released. Those are internal "releases" that we have been doing for validating the changes on master.

lucianoRM: Ok, I did not know that.

lucianoRM (approved):

PR #9346, opened 2020-09-03
MULE-18523: fix isDeployedMessageAtLogs when app is deployed after mule started (#9339)

jcgrethe:
* fix isDeployedMessageAtLogs when app is deployed after mule started

* move string to constant


jcgrethe:
--validate

PR #9358, opened 2020-09-03
CMTS-65: Spike Sample Data

svacas:
for review only

### DIFF
--- a/modules/tooling-support-parent/tooling-support/src/main/java/org/mule/runtime/module/tooling/api/artifact/DeclarationSession.java
+++ b/modules/tooling-support-parent/tooling-support/src/main/java/org/mule/runtime/module/tooling/api/artifact/DeclarationSession.java
@@ -12,12 +12,14 @@ import org.mule.runtime.api.connection.ConnectionValidationResult;
 import org.mule.runtime.api.metadata.MetadataKeysContainer;
 import org.mule.runtime.api.metadata.descriptor.ComponentMetadataTypesDescriptor;
 import org.mule.runtime.api.metadata.resolving.MetadataResult;
+import org.mule.runtime.api.sampledata.SampleDataResult;
 import org.mule.runtime.api.value.ValueResult;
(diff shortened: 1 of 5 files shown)


modules/tooling-support-parent/tooling-support/src/main/java/org/mule/runtime/module/tooling/api/artifact/DeclarationSession.java
```diff
@@ -70,6 +72,19 @@
...
+   *
+   * @param component the component whose sample data is required
+   * @return a {@link SampleDataResult} with the sample data message
+   */
+  SampleDataResult getSampleData(ComponentElementDeclaration component);
+
+  /**
+   * Exposes the application Expression Manager.
+   */
+  ExpressionManager getExpressionManager();
```
lucianoRM, line 86: Why do we need this? How will the client of the DeclarationSession need the ExpressionManager?

gsfernandes: We have defined to do it this way just to validate the approach for implementing this feature. I would say that we should create a task to enable access to the org.mule.runtime.api.el.DefaultExpressionLanguageFactoryService from our Agent plugin so we can create a org.mule.runtime.api.el.ExpressionLanguage which part of Mule API instead of exposing the one associated to the session. 
Let's create a Jira to do this on next sprint.

gsfernandes (changes requested):

modules/tooling-support-parent/tooling-support/src/main/java/org/mule/runtime/module/tooling/api/artifact/DeclarationSession.java
svacas: [CMTS-70: Provide access to Expression Language from Tooling Agent](https://www.mulesoft.org/jira/browse/CMTS-64)

PR #9366, opened 2020-09-07
MULE-18684: Unwanted namespace changes when ConnectionException and MULE:CONNECTIVITY errors are involved

IvanAndresFritzler:


tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/HeisenbergOperations.java
```diff
@@ -416,7 +416,7 @@ public String callSaul(@Connection HeisenbergConnection connection) {
   }
 
   @MediaType(TEXT_PLAIN)
-  public String callGusFring() throws HeisenbergException {
+  public String callGusFring(@Connection HeisenbergConnection connection) throws HeisenbergException {
```
elrodro83, line 419: add new operations/components to do your tests rather than modifying already existing ones that may be testing other stuff

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/exception/ExceptionHandlerManagerTestCase.java
```diff
@@ -82,15 +81,6 @@ public void processAndEnrich() {
     assertThat(throwable.getMessage(), is(ERROR_MESSAGE));
   }
 
-  @Test
-  public void handleConnectionException() {
```
elrodro83, line 86: why is this test removed?

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/DefaultExecutionMediatorTestCase.java
```diff
@@ -101,9 +106,19 @@
 @RunWith(Parameterized.class)
 public class DefaultExecutionMediatorTestCase extends AbstractMuleContextTestCase {
```
elrodro83, line 107: too many changes in this test, are we sure we're not breaking anything?

elrodro83 (changes requested):

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/HeisenbergOperations.java
IvanAndresFritzler: This particular operation is throwing a ConnectionException and it was not declaring the handling of a connection, thus not mapping the CONNECTIVITY error and making OperationErrorHandlingTestCase#connectionExceptionThrowsAnConnectivityError() fail after my changes, because the namespace was being resolved to MULE instead of HEISENBERG. I believe that in this case the modification applies, but can be modified to include a new operation if the SDK team thinks that is necessary. Side note: The test passed prior to my changes because the ConnectionException mapping was always added, regardless of the operation declarations.

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/exception/ExceptionHandlerManagerTestCase.java
IvanAndresFritzler: After my changes, ExceptionHandlerManager does no longer handle a ConnectionException in a special way, so I removed the unit test that was verifying that.

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/DefaultExecutionMediatorTestCase.java
IvanAndresFritzler: Well, I share the concern, but on the other hand, most of the changes are a consequence of removing the ConnectionException handling from the ExceptionHandlerManager class that in some cases was completely replacing the exception tree by a new one, discarding Module and Messaging exceptions that are relevant in the context of error handling and troubleshooting (see https://github.com/mulesoft/mule/pull/9366/commits/04ccc49e9e27ef59e7d782ce868ed177ffc00c97). That particular change will allow us to make an additional enhancement regarding the error handling and the troubleshooting that consists on suppressing the ConnectionException (probably in the ModuleExceptionHandler). I'm in the process of locally running some connector tests in order to see if something breaks.

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/HeisenbergOperations.java
ndinu: I agree with Rodro here, if you add this you will have to modify any usage of the operation.

How is this operation throwing a ConnectionException?

IvanAndresFritzler: The operation ends up throwing a ConnectionException because of the declaration of the HeisenbergConnectionExceptionEnricher class as an ExceptionHandler of the HeisenbergExtension, transforming any non MuleFatalException into a ConnectionException.

PR #9368, opened 2020-09-08
CMTS-35: Support invalidate caches (types/values) for a configuration

lucianoRM:


modules/spring-config/src/main/java/org/mule/runtime/config/api/dsl/model/metadata/DeclarationBasedValueProviderCacheIdGenerator.java
```diff
@@ -25,13 +25,26 @@ public DeclarationBasedValueProviderCacheIdGenerator(DslResolvingContext context
...
+
+  //This method was added so that it can be called by reflection and we can keep the logic on how
+  //the ValueProviderCacheId is generated for configs and connections in one place, without changing the
+  //ValueProviderCacheIdGenerator API.
+  //I could not come with a better solution that does not involve refactoring of the whole API. Having
+  //said that, we have encountered multiple scenarios where we are needing more information than the
+  //currently provided by the ValueProviderCacheId (Hierarchical info for example). Meaning that we should
+  //consider refactoring this in the future to make things simpler. MULE-18743
+  private Optional<ValueProviderCacheId> getIdForDependency(ElementDeclaration elementDeclaration) {
+    return elementModelFactory.create(elementDeclaration).flatMap(delegate::resolveIdForInjectedElement);
```
gsfernandes, line 46: How about another interface that extends ValueProviderCacheIdGenerator and provides this capability?

lucianoRM: I thought about that, but all this is in a .api. package and we need it to be api in order to access it from Tooling. Would that be ok?

gsfernandes (approved):

PR #9377, opened 2020-09-08
MULE-18733: Find another solution for HTTPC-50

SebaElizalde:


SebaElizalde:
--validate

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DynamicConfigurationProvider.java
```diff
@@ -334,4 +304,41 @@ public boolean isDynamic() {
         .filter(ob -> ob instanceof ConnectionProviderObjectBuilder)
         .map(ob -> ((ConnectionProviderObjectBuilder) ob).providerModel);
   }
+
+  private static class ResolverResultAndEvent {
+
+    private final Pair<ResolverSetResult, ResolverSetResult> resolverSetResult;
+    private final CoreEvent event;
```
ndinu: Why is it needed to store a reference to the event?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DynamicConfigurationProvider.java
```diff
@@ -244,16 +218,12 @@ protected void registerConfiguration(ConfigurationInstance configuration) {
...
-        cache.remove(entry.getKey());
-        unRegisterConfiguration(entry.getValue());
-        return entry.getValue();
-      }).collect(toImmutableList());
-    } finally {
-      cacheWriteLock.unlock();
-    }
+    ConcurrentMap<ResolverResultAndEvent, ConfigurationInstance> entries = cache.asMap();
+    return entries.entrySet().stream().filter(entry -> isExpired(entry.getValue())).map(entry -> {
+      cache.invalidate(entry.getKey());
```
ndinu, line 223: I think there should be a synchronization between the expiration of entries and the retrieval of entries. So that no config is retrieve right before it is being expired.

ndinu (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DynamicConfigurationProvider.java
```diff
@@ -144,33 +144,55 @@ public ConfigurationInstance get(Event event) {
...
+                                                                         connectionProviderValues);
+      } else {
+        configuration = configurationInstanceFactory.createConfiguration(getName(),
+                                                                         values.getFirst(),
+                                                                         event,
+                                                                         ofNullable(connectionProviderResolver));
+      }
+      registerConfiguration(configuration);
+      return configuration;
+    } catch (MuleException e) {
```
ndinu, line 193: Maybe we can always wrap the exception in a custom one, and then always unwrap it?

SebaElizalde:
See MULE-18773-V2: Find another solution for HTTPC-50 [#9475](https://github.com/mulesoft/mule/pull/9475)

PR #9419, opened 2020-09-15
MULE-18742: Getting ConcurrentModificationException error when disposing logs

szaffarano:


modules/launcher/src/test/java/org/mule/runtime/module/launcher/log4j2/MuleLog4jContextFactoryTestCase.java
```diff
@@ -74,4 +93,36 @@ public void dispose() {
     factory.dispose();
     verify(contextSelector).dispose();
   }
+
+  @Test
+  @Issue("MULE-18742")
+  @Description("If any shutdown callback is cancelled while the log is disposing everything should works")
```
elrodro83, line 99:
```suggestion
  @Description("If any shutdown callback is cancelled while the log is disposing everything should work")
```

elrodro83 (changes requested):

elrodro83 (approved):

PR #9432, opened 2020-09-17
EE-7446: Server Plugins are not able to be injected with a Mule Servi…

gsfernandes:
…ce that provides more than one contract

We cannot register the contracts and the also the service as there would be two instances of the a contract registered twice with different names.


elrodro83 (approved):

gsfernandes:
--validate

gsfernandes:
--validate

elrodro83 (approved):

gsfernandes:
--validate

PR #9439, opened 2020-09-18
MULE-18777: Multiple reconnections of a source must not occur at the same time

gabrieldalborgo:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -323,7 +323,6 @@ public void onException(ConnectionException exception) {
     reconnectionAction
         .doOnSuccess(v -> onReconnectionSuccessful())
         .doOnError(this::onReconnectionFailed)
-        .doAfterTerminate(() -> reconnecting.set(false))
```
marianogonzalez, line 326: why moving this from here to line 312? also, does the `doOn` vs `doAfter` change really fix the race condition or just makes it less likely?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -309,7 +309,7 @@ public void onException(ConnectionException exception) {
                 exception);
 
     Mono<Void> reconnectionAction = sourceAdapter.getReconnectionAction(exception)
-        .map(p -> from(retryPolicyTemplate.applyPolicy(p, retryScheduler)))
+        .map(p -> from(retryPolicyTemplate.applyPolicy(p, retryScheduler)).doOnTerminate(() -> reconnecting.set(false)))
```
marianogonzalez, line 312: where's the test case for this?

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
gabrieldalborgo: I'm working on it. This PR is to validate that this fix does not break any tests

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
gabrieldalborgo: Moving it to line 312 will make it apply only to sources that implement Reconnectable. Other cases are covered by the doWork method (line 548)

gabrieldalborgo: This case is covered by `org.mule.test.module.extension.source.CustomReconnectionSourceTestCase.successfulCustomReconnection`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
gabrieldalborgo: This case is covered by: `org.mule.test.module.extension.source.CustomReconnectionSourceTestCase.successfulCustomReconnection`

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
```diff
@@ -0,0 +1,86 @@
...
+
+  private Scheduler scheduler;
+
+  private static final AtomicReference<Integer> countStartedSources = new AtomicReference<>(0);
+
+  private void delay(Long millis) {
+    try {
+      Thread.sleep(millis);
+    } catch (InterruptedException e) {
+      e.printStackTrace();
```
marianogonzalez, line 44: don't swallow this exception

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
```diff
@@ -0,0 +1,86 @@
...
+import org.mule.runtime.extension.api.runtime.operation.Result;
+import org.mule.runtime.extension.api.runtime.source.Source;
+import org.mule.runtime.extension.api.runtime.source.SourceCallback;
+
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.atomic.AtomicReference;
+
+import javax.inject.Inject;
+
+public class OtherReconnectionSource extends Source<Integer, Void> {
```
marianogonzalez, line 26: all tests in this class are flaky

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
```diff
@@ -0,0 +1,86 @@
...
+  ConnectionProvider<ReconnectableConnection> connectionProvider;
+
+  @Inject
+  SchedulerService schedulerService;
+
+  private final AtomicReference<ScheduledFuture<?>> scheduleWithFixedDelay = new AtomicReference<>();
+
+  private Scheduler scheduler;
+
+  private static final AtomicReference<Integer> countStartedSources = new AtomicReference<>(0);
```
marianogonzalez, line 38: Use AtomicInteger instead

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
marianogonzalez, line 26: all tests in this class seem to be testing the implementation rather than behavior

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
marianogonzalez: this seems to be tackling a symptom rather than the root cause. What's the real problem been solved here? Moving this should have no effect, as no two reconnection tasks should be overlapping. If they do, then the bug is a prior race condition

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
gabrieldalborgo: What I saw, is that the reconnecting variable is being released (at this point) before the source onStart method ends (which runs asyncrhonously sometimes). And if at the source onStart method a connection exception is notified through the source callback, another reconnection overlaps the current one.

marianogonzalez: right, but the ting is that reconnection is always asynchronous, with the only exception of sources that implement the Reconnectable interface. So maybe it's not about moving this but about how could the overlap happen in the first place. I could well argue that it's really the `doWork` method the one which shouldn't be changing the variable and that the actual bug is the overlap, which you're merely hiding here.

I think you should start by getting good functional tests since the issue seems to revolve exclusively around reconnectable sources which is not something I'm seeing properly replicated on your tests cases.

gabrieldalborgo: I just pushed the test that I forgot before

gabrieldalborgo:
--validate

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/reconnection/ReconnectionTestCase.java
```diff
@@ -88,6 +89,23 @@ public void reconnectSource() throws Exception {
     });
   }
 
+  @Test
+  public void doNotStartSourceTwiceAfterExceptionOnReconnection() throws Exception {
+    ((Startable) getFlowConstruct("otherReconnectForever")).start();
+    check(5000, 1000, () -> !capturedEvents.isEmpty());
+    switchOtherFail();
+
+    checkNot(30000, 1000, () -> {
```
ndinu, line 98: I think we should make the test shorter, can we guarantee that the issue is fixed waiting less than 30 secs?

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
```diff
@@ -0,0 +1,87 @@
...
+    delay(1000L);
+    doStart(sourceCallback);
+  }
+
+  private void doStart(SourceCallback<Integer, Void> sourceCallback) throws MuleException {
+    ReconnectableConnection connection = connectionProvider.connect();
+    this.scheduler = schedulerService.ioScheduler();
+
+    if (ReconnectableConnectionProvider.otherFail) {
+      delay(10000L);
```
ndinu, line 53: Does it need to wait this much?

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/reconnection/ReconnectionTestCase.java
gabrieldalborgo: Let me try it

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
gabrieldalborgo: With 200ms (and maybe less) is enough, but I set a ridiculous value to avoid the test became flaky. Do you you think that is not necessary or not a good approach?

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/reconnection/ReconnectionTestCase.java
```diff
@@ -88,6 +89,23 @@ public void reconnectSource() throws Exception {
...
+      synchronized (capturedEvents) {
+        return capturedEvents.stream()
+            .map(event -> (Integer) event.getMessage().getPayload().getValue())
+            .filter(startedSources -> startedSources > 1)
```
SebaElizalde, line 102: Is there any way to also check that reconnection was triggered twice and simultaneously? From what I understand this test can also pass if only 1 reconnection is made (which is unlikely due to the Thread sleeps but could also happen if code is changed later on in the Extension Message Source).

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
ndinu: I think that the wait here will make the tests take longer, and we should avoid it if possible. If less than 200ms is enough, maybe we can use 1 second?

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/reconnection/ReconnectionTestCase.java
gabrieldalborgo: That is what this test does. The variable `startedSources` is counting how many sources are starting at the same time. I will rename it to something like `currentStartedSources`

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/reconnection/ReconnectionTestCase.java
gabrieldalborgo: Done

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
gabrieldalborgo: Done.

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/reconnection/ReconnectionTestCase.java
gabrieldalborgo: Done.

gabrieldalborgo:
--validate

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -581,6 +601,45 @@ public Object getWorkOwner() {
...
+      return delegate.getWorkDescription();
+    }
+
+    @Override
+    public Object getWorkOwner() {
+      return delegate.getWorkOwner();
+    }
+  }
+
+  private interface CompletionCallback {
```
marianogonzalez, line 635: we have a ton of interfaces like this already. Simply reuse `ReactiveReconnectionCallback`

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
```diff
@@ -0,0 +1,85 @@
...
+    countStartedSources.accumulateAndGet(1, Integer::sum);
+    doStart(sourceCallback);
+  }
+
+  private void doStart(SourceCallback<Integer, Void> sourceCallback) throws MuleException {
+    ReconnectableConnection connection = connectionProvider.connect();
+    this.scheduler = schedulerService.ioScheduler();
+
+    if (ReconnectableConnectionProvider.otherFail) {
+      delay(1000L);
```
marianogonzalez, line 52: sleeps make this flaky. Use latches instead

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
```diff
@@ -0,0 +1,85 @@
...
+        sourceCallback.onConnectionException(new ConnectionException(new RuntimeException(), connection));
+      } else {
+        sourceCallback.handle(Result.<Integer, Void>builder().output(countStartedSources.get()).build());
+      }
+    }, 0, 1000, MILLISECONDS));
+  }
+
+  @Override
+  public void onStop() {
+    countStartedSources.accumulateAndGet(-1, Integer::sum);
```
marianogonzalez, line 69: why not simlpy `addAndGet(-1)` ?

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
```diff
@@ -0,0 +1,85 @@
...
+
+public class OtherReconnectionSource extends Source<Integer, Void> {
+
+  @Connection
+  ConnectionProvider<ReconnectableConnection> connectionProvider;
+
+  @Inject
+  SchedulerService schedulerService;
+
+  private final AtomicReference<ScheduledFuture<?>> scheduleWithFixedDelay = new AtomicReference<>();
```
marianogonzalez, line 35: why is this in an atomic reference? I don't see the need and you're taking no value out of it

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  private void doStart(SourceCallback<Integer, Void> sourceCallback) throws MuleException {
+    ReconnectableConnection connection = connectionProvider.connect();
+    this.scheduler = schedulerService.ioScheduler();
+
+    if (ReconnectableConnectionProvider.otherFail) {
+      delay(1000L);
+      sourceCallback.onConnectionException(new ConnectionException(new RuntimeException(), connection));
+      ReconnectableConnectionProvider.otherFail = !ReconnectableConnectionProvider.otherFail;
```
marianogonzalez, line 54: `= false`

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
```diff
@@ -0,0 +1,85 @@
...
+
+  private final AtomicReference<ScheduledFuture<?>> scheduleWithFixedDelay = new AtomicReference<>();
+
+  private Scheduler scheduler;
+
+  private static final AtomicInteger countStartedSources = new AtomicInteger(0);
+
+  @Override
+  public void onStart(SourceCallback<Integer, Void> sourceCallback) throws MuleException {
+    countStartedSources.accumulateAndGet(1, Integer::sum);
```
marianogonzalez, line 43: `addAndGet(1)`

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
```diff
@@ -0,0 +1,85 @@
...
+      sourceCallback.onConnectionException(new ConnectionException(new RuntimeException(), connection));
+      ReconnectableConnectionProvider.otherFail = !ReconnectableConnectionProvider.otherFail;
+      throw new RuntimeException("Fail starting source");
+    }
+
+    scheduleWithFixedDelay.set(this.scheduler.scheduleWithFixedDelay(() -> {
+      if (ReconnectableConnectionProvider.otherFail) {
+        sourceCallback.onConnectionException(new ConnectionException(new RuntimeException(), connection));
+      } else {
+        sourceCallback.handle(Result.<Integer, Void>builder().output(countStartedSources.get()).build());
```
marianogonzalez, line 62: pushing and capturing an integer to the flow is probalby going to lead to a flaky test. Specially since you are using a source which fails exactly once every one a day and you're forcing a sleep which happens to be the exact half of the reconnection frequency.

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
```diff
@@ -0,0 +1,85 @@
...
+import org.mule.runtime.extension.api.runtime.source.Source;
+import org.mule.runtime.extension.api.runtime.source.SourceCallback;
+
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+
+import javax.inject.Inject;
+
+public class OtherReconnectionSource extends Source<Integer, Void> {
```
marianogonzalez, line 27: use a more meaningful name

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/ReconnectableConnectionProvider.java
```diff
@@ -28,6 +28,7 @@
 public class ReconnectableConnectionProvider implements CachedConnectionProvider<ReconnectableConnection> {
 
   public static volatile boolean fail;
+  public static volatile boolean otherFail;
```
marianogonzalez, line 31: use a more meaningful name

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/reconnection/ReconnectionTestCase.java
```diff
@@ -88,6 +89,23 @@ public void reconnectSource() throws Exception {
     });
   }
 
+  @Test
+  public void doNotStartSourceTwiceAfterExceptionOnReconnection() throws Exception {
+    ((Startable) getFlowConstruct("otherReconnectForever")).start();
```
marianogonzalez, line 94: this test depends greatly on `ReconnectableConnectionProvider.otherFail`. The initial value of that flag should be set of `doSetUp()`

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/ReconnectionOperations.java
```diff
@@ -39,6 +40,10 @@ public void switchConnection() {
     fail = !fail;
   }
 
+  public void switchOtherFail() {
```
marianogonzalez, line 43: meaninful name

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
gabrieldalborgo: Done

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
gabrieldalborgo: Done

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
gabrieldalborgo: Done

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
gabrieldalborgo: Done

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
gabrieldalborgo: Done

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/OtherReconnectionSource.java
gabrieldalborgo: Done

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/ReconnectableConnectionProvider.java
gabrieldalborgo: Done

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/ReconnectionOperations.java
gabrieldalborgo: Done

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/reconnection/ReconnectionTestCase.java
gabrieldalborgo: Done

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
gabrieldalborgo: Change rollbacked, isn't the right fix.

gabrieldalborgo:
--validate

ndinu (approved):

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/FallibleReconnectableSource.java
```diff
@@ -0,0 +1,95 @@
...
+
+  private Scheduler scheduler;
+  private ScheduledFuture<?> scheduleWithFixedDelay;
+
+  private static final AtomicInteger countStartedSources = new AtomicInteger(0);
+  private static final Latch latch = new Latch();
+
+  @Override
+  public void onStart(SourceCallback<Void, Void> sourceCallback) throws MuleException {
+    if (countStartedSources.addAndGet(1) > 1) {
```
fsgonz, line 46: Shouldn't this verify that there are two countedStartedSources simultaneously in the onStart instead of substracting in the onStop. What you need to verify is that the onStart is not executed again until the firs one ended.

fsgonz (changes requested):

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/reconnection/ReconnectionTestCase.java
```diff
@@ -88,6 +92,16 @@ public void reconnectSource() throws Exception {
     });
   }
 
+  @Test
+  public void doNotStartSourceTwiceAfterExceptionOnReconnection() throws Exception {
+    ((Startable) getFlowConstruct("reconnectAfterFailure")).start();
+    check(5000, 1000, () -> !capturedEvents.isEmpty());
+    FallibleReconnectableSource.fail = true;
+    checkNot(4000, 500, () -> FallibleReconnectableSource.simultaneouslyStartedSources);
```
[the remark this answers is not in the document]
gabrieldalborgo: No, there isn't. I will change it to 5000 again, so there is no confusion

tests/test-extensions/reconnection-ext/src/main/java/org/mule/extension/test/extension/reconnection/FallibleReconnectableSource.java
```diff
@@ -0,0 +1,102 @@
...
+    } finally {
+      countStartedSources.addAndGet(-1);
+    }
+  }
+
+  private void doStart(SourceCallback<Void, Void> sourceCallback) throws MuleException {
+    if (fail) {
+      await();
+      fail = false;
+      throw new RuntimeException("Fail starting source");
```
fsgonz, line 64: Why make this fail again? In case you want to retry that a new reconnection triggers a new start on the source should this be a onConnectionException.

fsgonz: As per slack, this has the same effect.

fsgonz (approved):

PR #9445, opened 2020-09-21
MULE-18465: PollingSource refresh token does not work with OAuth Authorization Code Grant

ndinu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -304,6 +305,9 @@ public void onException(ConnectionException exception) {
       return;
     }
 
+    refreshTokenIfNecessary(getConfigurationInstance()
+        .flatMap(configurationInstance -> configurationInstance.getConnectionProvider()).orElse(null), exception);
```
marianogonzalez, line 309: add logs stating that the token was refreshed. Same thing goes for all the other similar PRs that you sent

marianogonzalez (changes requested):

marianogonzalez (approved):

PR #9451, opened 2020-09-23
MULE-18778: Sample Data resolution must take into account oauth token refresh

ndinu:
… refresh

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/data/sample/SampleDataProviderMediator.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/data/sample/SampleDataProviderMediator.java
@@ -13,10 +13,12 @@ import static java.util.Optional.ofNullable;
 import static org.mule.runtime.core.internal.event.NullEventFactory.getNullEvent;
 import static org.mule.runtime.core.internal.management.stats.NoOpCursorComponentDecoratorFactory.NO_OP_INSTANCE;
 import static org.mule.runtime.core.internal.util.rx.ImmediateScheduler.IMMEDIATE_SCHEDULER;
+import static org.mule.runtime.module.extension.internal.runtime.connectivity.oauth.ExtensionsOAuthUtils.withRefreshToken;
 import static org.mule.sdk.api.data.sample.SampleDataException.NOT_SUPPORTED;
 import static org.mule.sdk.api.data.sample.SampleDataException.UNKNOWN;
 
 import org.mule.runtime.api.component.Component;
+import org.mule.runtime.api.connection.ConnectionProvider;
 import org.mule.runtime.api.message.Message;
 import org.mule.runtime.api.meta.model.ComponentModel;
 import org.mule.runtime.api.meta.model.EnrichableModel;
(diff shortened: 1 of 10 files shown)


ndinu:
--validate

ndinu:
--validate

ndinu:
--validate

ndinu:
--validate

ndinu:
--validate

ndinu:
--validate

ndinu:
--validate

gabrieldalborgo (approved):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/data/sample/SampleDataProviderMediator.java
```diff
@@ -96,6 +98,28 @@ public Message getSampleData(ParameterValueResolver parameterValueResolver,
...
+   *                                by the  {@link SampleDataProvider}
+   * @param connectionProvider      the connection provider in charge of providing the connection given by the connection supplier.
+   *
+   * @return a {@link Message} carrying the sample data
+   * @throws SampleDataException if an error occurs resolving the sample data
+   */
+  public Message getSampleData(ParameterValueResolver parameterValueResolver,
+                               Supplier<Object> connectionSupplier,
+                               Supplier<Object> configurationSupplier,
+                               ConnectionProvider connectionProvider)
```
marianogonzalez, line 121: you're already providing the config. You should expose this in the API

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/data/sample/SampleDataProviderMediator.java
ndinu: I did not understand this comment, could you rephrase?

ndinu: In order to give me the connection provider, the caller needed to create the configuration. What I can do is to receive a Supplier<ConnectionProvider> and only do it only if necessary.

The util method that makes the refresh of the token will also need to receive the Supplier instead of the provider directly.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/data/sample/SampleDataProviderMediator.java
```diff
@@ -96,6 +98,28 @@ public Message getSampleData(ParameterValueResolver parameterValueResolver,
...
+  }
+
+  /**
+   * Resolves the sample data
+   *
+   * @param parameterValueResolver  parameter resolver required if the associated {@link SampleDataProvider} requires
+   *                                the value of parameters from the same parameter container.
+   * @param connectionSupplier      supplier of connection instances related to the container and used, if necessary,
+   *                                by the {@link SampleDataProvider}
+   * @param configurationSupplier   supplier of connection instances related to the container and used, if necessary,
```
SebaElizalde, line 111: typo, should say supplier of configurations instances

marianogonzalez (approved):

PR #9459, opened 2020-09-24
MULE-18794: Cannot access application resources from KeyResolver for Configurations Part 2

SebaElizalde:


SebaElizalde:
--validate

SebaElizalde:
--validate

tests/test-extensions/classloading-extension/src/main/java/org/mule/test/classloading/ClassLoadingExtension.java
```diff
@@ -11,12 +11,14 @@
...
 import org.mule.runtime.extension.api.annotation.connectivity.ConnectionProviders;
 import org.mule.runtime.extension.api.annotation.dsl.xml.Xml;
+import org.mule.runtime.extension.api.annotation.metadata.MetadataScope;
 import org.mule.test.classloading.api.ClassLoadingHelper;
 
 @Extension(name = "ClassLoading")
 @Xml(prefix = "classloading")
 @ConnectionProviders({CLPoolingConnectionProvider.class, CLCachedConnectionProvider.class, CLNoneConnectionProvider.class})
 @Configurations(CLConfiguration.class)
+@MetadataScope(keysResolver = CLKeysResolver.class)
```
ndinu, line 21: I would set the type resolver for a specific operation you want to modify, or create a new one.

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/classloading/ClassLoadingOnConnectionsTestCase.java
```diff
@@ -29,13 +34,18 @@
...
+
 import org.hamcrest.Matcher;
 import org.hamcrest.core.StringContains;
 import org.junit.Rule;
 import org.junit.Test;
 
 public class ClassLoadingOnConnectionsTestCase extends AbstractExtensionFunctionalTestCase {
 
+  @Inject
+  private MetadataService metadataManager;
```
ndinu, line 47: call the field metadataService

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/classloading/ClassLoadingOnConnectionsTestCase.java
```diff
@@ -72,6 +82,12 @@ public void allOptionalParameterGroup() throws Exception {
     verifyUsedClassLoaders(ALL_OPTIONAL_PARAMETER_GROUP);
   }
 
+  @Test
+  public void operationWithMetadataResolver() throws Exception {
```
ndinu, line 86: I don't feel that this test belongs in this class.

tests/test-extensions/classloading-extension/src/main/java/org/mule/test/classloading/CLKeysResolver.java
```diff
@@ -0,0 +1,50 @@
...
+import org.mule.runtime.api.metadata.MetadataKeyBuilder;
+import org.mule.runtime.api.metadata.MetadataResolvingException;
+import org.mule.runtime.api.metadata.resolving.OutputTypeResolver;
+import org.mule.runtime.api.metadata.resolving.TypeKeysResolver;
+
+import java.util.Set;
+
+import com.google.common.collect.ImmutableSet;
+
+public class CLKeysResolver implements TypeKeysResolver, OutputTypeResolver {
```
ndinu, line 26: If you are going to assert that the operation metadata key resolution and source metadata key resolution are using the correct class loader, I would make them use two different resolvers, and use the addClassLoader method with different values.

ndinu (changes requested):

tests/test-extensions/classloading-extension/src/main/java/org/mule/test/classloading/CLKeysResolver.java
SebaElizalde: They both end up calling the same method, so I don't feel like there is a need to distinguish between sources and operations by making 2 separate tests.

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/classloading/ClassLoadingOnConnectionsTestCase.java
SebaElizalde: will refactor

tests/test-extensions/classloading-extension/src/main/java/org/mule/test/classloading/ClassLoadingExtension.java
SebaElizalde: This is needed for configs to be annotated with the KeysResolver annotation in the model. That then allows the following line to be called in the tests:
`metadataService.getMetadataKeys(builder().globalName("none").build());`
and that will trigger getting the metadata keys.

tests/test-extensions/classloading-extension/src/main/java/org/mule/test/classloading/CLKeysResolver.java
ndinu: I think that calling the same method is an implementation detail, we should cover both cases.

SebaElizalde:
--validate

ndinu (dismissed):

ndinu (approved):

PR #9467, opened 2020-09-28
CMTS-91: Improve ValueProvider caching with providerId information

gsfernandes:


modules/spring-config/src/test/java/org/mule/runtime/config/dsl/model/ValueProviderCacheIdGeneratorTestCase.java
```diff
@@ -400,13 +399,24 @@ public void differentConfigsWithSameParameterGetDifferentHash() throws Exception
...
     when(valueProviderModel.requiresConnection()).thenReturn(true);
     when(valueProviderModel.requiresConfiguration()).thenReturn(true);
     Optional<ValueProviderCacheId> opId1 = computeIdFor(app, OPERATION_LOCATION, PROVIDED_PARAMETER_NAME);
     when(valueProviderModel.getProviderName()).thenReturn("newValueProviderName");
     Optional<ValueProviderCacheId> opId2 = computeIdFor(app, OPERATION_LOCATION, PROVIDED_PARAMETER_NAME);
+    checkIdsAreEqual(opId1, opId2);
+  }
+
+  @Test
+  public void differentValueProviderIdGetsDifferentHash() throws Exception {
```
lucianoRM, line 413: We should test that 2 different operations with the same value providerId and same values, return the same hash

lucianoRM (changes requested):

modules/spring-config/src/test/java/org/mule/runtime/config/dsl/model/ValueProviderCacheIdGeneratorTestCase.java
gsfernandes: https://github.com/mulesoft/mule/pull/9467/files#diff-651f29168e45b3db6280ce600b706537R402

gsfernandes:
--validate

gsfernandes:
--validate

modules/spring-config/src/test/java/org/mule/runtime/config/dsl/model/ValueProviderCacheIdGeneratorTestCase.java
gsfernandes: Ok, I will add another test with this scenario.

gsfernandes: Done, please review this again.

lucianoRM (approved):

modules/spring-config/src/main/java/org/mule/runtime/config/api/dsl/model/metadata/DslElementBasedValueProviderCacheIdGenerator.java
```diff
@@ -101,24 +103,32 @@ public DslElementBasedValueProviderCacheIdGenerator(ComponentLocator<DslElementM
...
 
-    parts.add(resolveValueProviderId(valueProviderModel));
     parts.addAll(resolveActingParameterIds(valueProviderModel, parameterModelsInformation));
     parts.addAll(resolveIdForInjectedElements(containerComponent, valueProviderModel));
+    parts.add(resolveValueProviderId(valueProviderModel));
+    parts.add(aValueProviderCacheId(fromElementWithName(VALUE_PROVIDER).withHashValueFrom(VALUE_PROVIDER)));
 
     String id = resolveDslTag(containerComponent).orElse(getSourceElementName(containerComponent));
-    return of(aValueProviderCacheId(fromElementWithName(id).withHashValueFrom(id).containing(parts)));
+    return of(aValueProviderCacheId(fromElementWithName(id)
```
lucianoRM, line 126: only the namespace and if not present the id? You are assuming that if the namespace is not present, then the id should be the identifier right?

gsfernandes: Yes, there is no other way to associate it to an extension.

PR #9469, opened 2020-09-28
MULE-18803: Premature completion of choice/round-robin route causes

elrodro83:
dropped events


core/src/main/java/org/mule/runtime/core/internal/routing/AbstractSinkRouter.java
```diff
@@ -45,16 +51,12 @@ protected AbstractSinkRouter(Publisher<CoreEvent> publisher, List<ProcessorRoute
    */
   List<Flux<CoreEvent>> collectPublishers() {
     List<Flux<CoreEvent>> routes = new ArrayList<>();
```
eze210, line 53: Avoid shadowing the attribute routes

eze210 (changes requested):

eze210 (approved):

PR #9475, opened 2020-09-29
MULE-18773-V2: Find another solution for HTTPC-50

SebaElizalde:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DynamicConfigurationProvider.java
```diff
@@ -334,4 +318,22 @@ public boolean isDynamic() {
         .filter(ob -> ob instanceof ConnectionProviderObjectBuilder)
         .map(ob -> ((ConnectionProviderObjectBuilder) ob).providerModel);
   }
+
+  /**
+   * Used to preserve exception throwing behaviour in {@link #getConfiguration} as the method {@link #createConfiguration} can no longer throw
+   * checked exception as it is used inside {@link Map#computeIfAbsent}.
+   */
+  private class WrappingRuntimeException extends RuntimeException {
```
ndinu, line 326: private static class

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DynamicConfigurationProvider.java
```diff
@@ -334,4 +318,22 @@ public boolean isDynamic() {
...
+
+    private final Exception wrappedException;
```
ndinu, line 328: You can use the cause instead of creating the instance variable.

SebaElizalde: but getCause() returns a Throwable type instead of an Exception type. I would either need to cast it or change the getConfiguration signature to "throws Throwable".

ndinu: 👍

ndinu (approved):

SebaElizalde:
--validate

SebaElizalde:
--validate

SebaElizalde:
--validate

SebaElizalde:
--validate

SebaElizalde:
--validate

SebaElizalde:
--validate

SebaElizalde:
--validate

PR #9488, opened 2020-10-01
MULE-18817: XML to JSON fail when XML depth is over 64

jcgrethe:


fsgonz (changes requested):
Add a comment before https://github.com/mulesoft/mule/blob/support/3.9.x/modules/json/pom.xml#L92

indicating "In case this version is changed, please verify shaded classes added due to MULE-18817"

fsgonz (approved):

PR #9491, opened 2020-10-01
CMTS-95: Support resources for DeclarationSession

gsfernandes:


modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/api/ArtifactAgnosticServiceBuilder.java
```diff
@@ -44,6 +44,8 @@
    */
   T addDependency(Dependency dependency);
 
+  T addResource(String resourcePath, byte[] content);
```
lucianoRM, line 47: Add javadoc. 
What format should the byte[] have? It will be written as is right? No encoding, or compression?
Also, I think the resourcePath is weird. The content will be copied to the application folder as root right? Why provide the possibility to "create" subdirectories? it seems kind of complex for something that will not be used multiple times. Maybe if they just provide the names it will be easier.

lucianoRM (changes requested):

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/api/ArtifactAgnosticServiceBuilder.java
gsfernandes: Encoding would be passed as parameter too. Compression, I don't think we would need to handle that (at least here in the DeclarationSesssionBuilder).
Paths as they are defined in the declaration, I would keep the possibility to allow setting a classpath resourcePath. I have to add javadoc, the initial PR was to validate the approach and confirm it would work without having to change the way we create the application.

gsfernandes: I think that resourcePath should allow to use directories too. As content is in binary we don't need the encoding.

lucianoRM (approved):

PR #9503, opened 2020-10-05
MULE-18828: Same MetadataCacheId when two configless operations have …

gsfernandes:
…exactly the same category and resolver names


modules/spring-config/src/main/java/org/mule/runtime/config/api/dsl/model/metadata/ComponentAstBasedMetadataCacheIdGenerator.java
```diff
@@ -185,6 +186,11 @@ public ComponentAstBasedMetadataCacheIdGenerator(ComponentLocator<ComponentAst>
                                       .orElse(format("(%s):(%s)", getSourceElementName(component), "Unknown Type"))));
   }
 
+  private Optional<MetadataCacheId> resolveDslTagNamespace(ComponentAst elementModel) {
+    String namespace = elementModel.getIdentifier().getNamespace();
```
lucianoRM, line 190: I think the identifier can be null, we should consider that

lucianoRM (changes requested):

modules/spring-config/src/main/java/org/mule/runtime/config/api/dsl/model/metadata/ComponentAstBasedMetadataCacheIdGenerator.java
gsfernandes: org.mule.runtime.ast.api.util.BaseComponentAstDecorator#toString it is not checking for null. It should always have an identifier, if it is null I would say there is a bug in AST.

gsfernandes: Confirmed by Rodro, it should not be null.

lucianoRM (approved):

PR #9509, opened 2020-10-05
MULE-18777: Multiple reconnections of a source must not occur at the same time - Part 2

gabrieldalborgo:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -323,22 +313,25 @@ public void onException(ConnectionException exception) {
...
-            sink.error(e);
-          }
-        }));
-
-    reconnectionAction
-        .doOnSuccess(v -> onReconnectionSuccessful())
-        .doOnError(this::onReconnectionFailed)
-        .doAfterTerminate(() -> reconnecting.set(false))
-        .subscribe();
+    Executors.newSingleThreadExecutor().submit(() -> {
```
marianogonzalez, line 316: bad usage of scheduler:

* You're creating a scheduler which you never dispose
* All scheduling should be done through the scheduling service

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -747,4 +707,12 @@ private void stopUsingConfiguration(CoreEvent event) {
   public String toString() {
     return this.getClass().getSimpleName() + ": " + Objects.toString(sourceAdapter);
   }
+
+  private class ImmediateExecutor implements Executor {
```
marianogonzalez, line 711: delete and use org.mule.runtime.core.internal.util.rx.ImmediateScheduler#IMMEDIATE_SCHEDULER instead

marianogonzalez (changes requested):
Besides the requested code changes, it's not clear to me how this PR adds all the test coverage that was missing. Is the test case you're adding enough to validate the regression from the original fix? What else is missing?

gabrieldalborgo:
> Besides the requested code changes, it's not clear to me how this PR adds all the test coverage that was missing. Is the test case you're adding enough to validate the regression from the original fix? What else is missing?

The regression you mention is that the reconnection fails and you don't try again despite not reaching the try limit.

There are tests that verify that the reconnection works and a new one that verifies that if the reconnection attempts are exhausted, it is not tried again.

But is not clear to me how to validate the regression scenario, how to represent it in a test. Do you have any suggestion?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -149,6 +148,7 @@
   private SourceAdapter sourceAdapter;
   private RetryPolicyTemplate retryPolicyTemplate;
   private Scheduler retryScheduler;
+  private Scheduler reconnectScheduler;
```
marianogonzalez, line 151: you already have a retry scheduler. Why not reuse that one?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -464,6 +468,14 @@ private void stopSchedulers() {
         retryScheduler = null;
       }
     }
+
+    if (reconnectScheduler != null) {
```
marianogonzalez, line 472: to create and destroy the scheduler on each use is not a good way of handling the lifecycle

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
gabrieldalborgo: Done

marianogonzalez:
> > Besides the requested code changes, it's not clear to me how this PR adds all the test coverage that was missing. Is the test case you're adding enough to validate the regression from the original fix? What else is missing?
> 
> The regression you mention is that the reconnection fails and you don't try again despite not reaching the try limit.
> 
> There are tests that verify that the reconnection works and a new one that verifies that if the reconnection attempts are exhausted, it is not tried again.
> 
> But is not clear to me how to validate the regression scenario, how to represent it in a test. Do you have any suggestion?

Your original fix broke some scenarios, even though all tests passed. That's a clear indication that test coverage is not good enough here. The question is, how are you ensuring that this fix doesn't introduce more regressions? How are you improving the test coverage to consider this fix safe?

gabrieldalborgo:
> > > Besides the requested code changes, it's not clear to me how this PR adds all the test coverage that was missing. Is the test case you're adding enough to validate the regression from the original fix? What else is missing?
> > 
> > 
> > The regression you mention is that the reconnection fails and you don't try again despite not reaching the try limit.
> > There are tests that verify that the reconnection works and a new one that verifies that if the reconnection attempts are exhausted, it is not tried again.
> > But is not clear to me how to validate the regression scenario, how to represent it in a test. Do you have any suggestion?
> 
> Your original fix broke some scenarios, even though all tests passed. That's a clear indication that test coverage is not good enough here. The question is, how are you ensuring that this fix doesn't introduce more regressions? How are you improving the test coverage to consider this fix safe?

I added two more tests around reconnection.

marianogonzalez (approved):

fsgonz (commented):
https://www.mulesoft.org/jira/browse/MULE-17267 I've validated this with changes.

fsgonz (approved):

PR #9525, opened 2020-10-08
MULE-18880: ClassCastException: org.mule.runtime.api.metadata.TypedValue cannot be cast to org.mule.runtime.extension.api.runtime.operation.Result	

elrodro83:


fsgonz (approved):

elrodro83:
--validate

PR #9531, opened 2020-10-08
CMTS-98: DeclarationSession is not failing when parameter values for …

gsfernandes:
…values/metadata/sampleData have an expression

### DIFF
--- a/modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/AbstractParameterResolverExecutor.java
+++ b/modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/AbstractParameterResolverExecutor.java
@@ -10,6 +10,7 @@ import static java.lang.String.format;
 import static java.util.function.Function.identity;
 import static java.util.stream.Collectors.toMap;
 import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
+import static org.mule.runtime.api.meta.ExpressionSupport.NOT_SUPPORTED;
 import static org.mule.runtime.module.extension.internal.util.MuleExtensionUtils.isExpression;
 import static org.mule.runtime.module.tooling.internal.artifact.params.ParameterExtractor.extractValue;
 import org.mule.runtime.api.exception.MuleRuntimeException;
@@ -95,7 +96,7 @@ public class AbstractParameterResolverExecutor {
                                                                             parameterName, parameterGroupName)));
         Object value = extractValue(parameterElement.getValue(),
                                     artifactHelper.getParameterClass(parameterModel, parameterizedElementDeclaration));
-        if (isExpression(value)) {
+        if (!parameterModel.getExpressionSupport().equals(NOT_SUPPORTED) && isExpression(value)) {
           throw new ExpressionNotSupportedException(format("Error resolving value for parameter: '%s' from declaration, it cannot be an EXPRESSION value",
                                                            parameterName));
         }
--- a/modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/AbstractParameterResolverExecutor.java
+++ b/modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/AbstractParameterResolverExecutor.java
@@ -6,11 +6,12 @@
  */
 package org.mule.runtime.module.tooling.internal.artifact;
 
+import static java.lang.String.format;
 import static java.util.function.Function.identity;
 import static java.util.stream.Collectors.toMap;
 import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
+import static org.mule.runtime.module.extension.internal.util.MuleExtensionUtils.isExpression;
 import static org.mule.runtime.module.tooling.internal.artifact.params.ParameterExtractor.extractValue;
-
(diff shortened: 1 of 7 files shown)


modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/AbstractParameterResolverExecutor.java
```diff
@@ -87,9 +93,13 @@ protected ParameterValueResolver parameterValueResolver(ParameterizedElementDecl
         final ParameterModel parameterModel = parameterGroupModel.getParameter(parameterName)
             .orElseThrow(() -> new MuleRuntimeException(createStaticMessage("Could not find parameter with name: '%s' in parameter group: '%s'",
                                                                             parameterName, parameterGroupName)));
-        parametersMap.put(parameterName,
-                          extractValue(parameterElement.getValue(),
-                                       artifactHelper.getParameterClass(parameterModel, parameterizedElementDeclaration)));
+        Object value = extractValue(parameterElement.getValue(),
+                                    artifactHelper.getParameterClass(parameterModel, parameterizedElementDeclaration));
+        if (isExpression(value)) {
```
gsfernandes, line 98: I just realized that I should first check if the parameterModel.isSupportExpression() == true before checking the String content.

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/metadata/MetadataKeyDeclarationResolver.java
```diff
@@ -150,7 +152,12 @@ public MetadataKeyResult resolveKeyResult() {
             .orElseThrow(() -> new MuleRuntimeException(createStaticMessage("Could not find parameter with name: %s in parameter group: %s",
                                                                             parameterName, parameterGroupName)));
         if (parameterModel.getModelProperty(MetadataKeyPartModelProperty.class).isPresent()) {
-          parametersMap.put(parameterName, extractSimpleValue(parameterElement.getValue()));
+          String value = extractSimpleValue(parameterElement.getValue());
+          if (isExpression(value)) {
```
gsfernandes, line 156: Same here. I just realized that I should first check if the parameterModel.isSupportExpression() == true before checking the String content.

gsfernandes:
--validate

gsfernandes:
--validate

gsfernandes:
--validate

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/AbstractParameterResolverExecutor.java
```diff
@@ -87,9 +94,13 @@ protected ParameterValueResolver parameterValueResolver(ParameterizedElementDecl
         final ParameterModel parameterModel = parameterGroupModel.getParameter(parameterName)
             .orElseThrow(() -> new MuleRuntimeException(createStaticMessage("Could not find parameter with name: '%s' in parameter group: '%s'",
                                                                             parameterName, parameterGroupName)));
-        parametersMap.put(parameterName,
-                          extractValue(parameterElement.getValue(),
-                                       artifactHelper.getParameterClass(parameterModel, parameterizedElementDeclaration)));
+        Object value = extractValue(parameterElement.getValue(),
+                                    artifactHelper.getParameterClass(parameterModel, parameterizedElementDeclaration));
+        if (!parameterModel.getExpressionSupport().equals(NOT_SUPPORTED) && isExpression(value)) {
```
lucianoRM, line 99: Even if the parameterModel supportExpressions. Shouldn't we fail? I mean, we won't be able to process them anyways.

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/metadata/MetadataKeyDeclarationResolver.java
```diff
@@ -150,7 +153,12 @@ public MetadataKeyResult resolveKeyResult() {
             .orElseThrow(() -> new MuleRuntimeException(createStaticMessage("Could not find parameter with name: %s in parameter group: %s",
                                                                             parameterName, parameterGroupName)));
         if (parameterModel.getModelProperty(MetadataKeyPartModelProperty.class).isPresent()) {
-          parametersMap.put(parameterName, extractSimpleValue(parameterElement.getValue()));
+          String value = extractSimpleValue(parameterElement.getValue());
+          if (!parameterModel.getExpressionSupport().equals(NOT_SUPPORTED) && isExpression(value)) {
```
lucianoRM, line 157: The same here

lucianoRM (changes requested):

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/AbstractParameterResolverExecutor.java
gsfernandes: Well, in that case the "value" is handled as String and it is up to the resolver to handle it. I added this in order to prevent doing the parsing of the String for each parameter as it has impact in performance.

lucianoRM (approved):

PR #9558, opened 2020-10-14
MULE-18889: Export META-INF/services/* resources in lightweight deployments

elrodro83:
* To make it consistent with the classloader-model.json that is
generated by the packager for heavyweight deployments.


modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/ApplicationDeploymentClassloadingTestCase.java
```diff
@@ -344,6 +345,90 @@ public void deploysLightApplicationWithPluginDependingOnPlugin() throws Exceptio
     executeApplicationFlow("main");
   }
 
+  // TODO
```
gsfernandes, line 348: remove // TODO

gsfernandes (changes requested):

gsfernandes (approved):

gsfernandes (approved):

PR #9573, opened 2020-10-15
MULE-18872: Dropped event causes OOM error

IvanAndresFritzler:


core/src/main/java/org/mule/runtime/core/internal/context/DefaultMuleContext.java
```diff
@@ -266,8 +266,8 @@
 
   static {
     // Log dropped events/errors
-    Hooks.onErrorDropped(error -> LOGGER.debug("ERROR DROPPED " + error));
-    Hooks.onNextDropped(event -> LOGGER.debug("EVENT DROPPED " + event));
+    Hooks.onErrorDropped(error -> LOGGER.debug("ERROR DROPPED {}", (Object) error));
```
elrodro83, line 269: i would log the exception with the full stack trace here. It is already in DEBUG level and the information provided by it may be useful.

elrodro83 (changes requested):

core/src/main/java/org/mule/runtime/core/internal/context/DefaultMuleContext.java
IvanAndresFritzler: Agree (I have been overly conservative).

elrodro83 (approved):

IvanAndresFritzler:
--validate

PR #9585, opened 2020-10-16
MULE-18894: NotSerializableException when persisting ABP decorators.

fsgonz:


core/src/main/java/org/mule/runtime/core/internal/management/stats/PayloadStatisticsCollection.java
```diff
@@ -44,4 +58,15 @@ public Spliterator spliterator() {
     populator.accept(size());
     return super.toArray(object);
   }
+
+  @Override
+  public void writeExternal(ObjectOutput out) throws IOException {
+    out.writeObject(getCollection());
+
```
elrodro83, line 65: remove empty line

core/src/main/java/org/mule/runtime/core/internal/management/stats/PayloadStatisticsCollection.java
```diff
@@ -44,4 +58,15 @@ public Spliterator spliterator() {
...
+  }
+
+  @Override
+  public void readExternal(ObjectInput in) throws IOException, ClassNotFoundException {
+    this.collection = (Collection) in.readObject();
```
elrodro83, line 70: we shouldn't be expecting to encounter this actual class serialized.

core/src/main/java/org/mule/runtime/core/internal/management/stats/PayloadStatisticsList.java
```diff
@@ -45,4 +61,14 @@ public Spliterator spliterator() {
...
   }
 
+  @Override
+  public void writeExternal(ObjectOutput out) throws IOException {
+    out.writeObject(getList());
+  }
+
+  @Override
+  public void readExternal(ObjectInput in) throws IOException, ClassNotFoundException {
+    this.collection = (List) in.readObject();
```
elrodro83, line 71: we shouldn't be expecting to encounter this actual class serialized.

core/src/main/java/org/mule/runtime/core/internal/management/stats/PayloadStatisticsCollection.java
```diff
@@ -6,18 +6,32 @@
...
 import java.util.function.LongConsumer;
 
 import org.apache.commons.collections.collection.AbstractCollectionDecorator;
 
-class PayloadStatisticsCollection<T> extends AbstractCollectionDecorator {
+class PayloadStatisticsCollection<T> extends AbstractCollectionDecorator implements Externalizable {
 
   private final LongConsumer populator;
 
+  public PayloadStatisticsCollection() {
```
elrodro83, line 26: would this be needed if there were a kryo serializer for this class?

core/src/main/java/org/mule/runtime/core/internal/management/stats/PayloadStatisticsList.java
```diff
@@ -14,10 +19,21 @@
...
 import org.apache.commons.collections.list.AbstractListDecorator;
 
-final class PayloadStatisticsList<T> extends AbstractListDecorator {
+final class PayloadStatisticsList<T> extends AbstractListDecorator implements Externalizable {
+
+  private static final long serialVersionUID = -5991764233175594789L;
 
   private final LongConsumer populator;
 
+  public PayloadStatisticsList() {
```
elrodro83, line 28: would this be needed if there were a kryo serializer for this class?

elrodro83 (changes requested):

core/src/main/java/org/mule/runtime/core/internal/management/stats/PayloadStatisticsCollection.java
fsgonz: the void constructor is needed for non-kryo scenarios. Kryo also needs all parameters to be initialized in the void constructor.

core/src/main/java/org/mule/runtime/core/internal/management/stats/PayloadStatisticsList.java
fsgonz: the void constructor is needed for non-kryo scenarios. Kryo also needs all parameters to be initialized in the void constructor.

core/src/main/java/org/mule/runtime/core/internal/management/stats/PayloadStatisticsCollection.java
fsgonz: Needed for the case where we are not using kryo.

elrodro83 (approved):

elrodro83 (approved):

PR #9622, opened 2020-10-22
MULE-18913: Dynamic Configurations are expired even though a paginated operation is still using it for 4.2.x

SebaElizalde:


tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/MoneyLaunderingOperation.java
```diff
@@ -251,6 +251,33 @@ public void close(HeisenbergConnection connection) {
     };
   }
 
+  public PagingProvider<HeisenbergConnection, PersonalInfo> failingConnectionPagedOperation(@Config HeisenbergExtension config,
```
ndinu, line 254: Shouldn't this be checking if the config is disposed in the getPage method?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/MutableConfigurationStats.java
```diff
@@ -51,4 +51,20 @@
    * @since 4.1.6 4.2.2 4.3.0
    */
   int discountRunningSource();
+
+  /**
+   * Increments the return value of {@link #getOpenedStreams()} by one
+   *
+   * @return the new value
+   * @since 4.2.3 4.3.0
```
ndinu, line 59: 4.3.0 is already released, I think it should be 4.2.3 4.3.1 4.4.0

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/MutableConfigurationStats.java
```diff
@@ -51,4 +51,20 @@
...
+   */
```
ndinu, line 60: First be sure that this is the api that will be used in the master branch

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/AbstractReturnDelegate.java
```diff
@@ -188,8 +196,10 @@ protected Message toMessage(Object value, ExecutionContextAdapter operationConte
       if (value.getOutput() instanceof InputStream) {
         ConnectionHandler connectionHandler = (ConnectionHandler) operationContext.getVariable(CONNECTION_PARAM);
         if (connectionHandler != null && supportsStreaming(operationContext.getComponentModel())) {
-          value = value.copy()
-              .output(new ConnectedInputStreamWrapper((InputStream) value.getOutput(), connectionHandler))
+          ConfigurationInstance config = (ConfigurationInstance) operationContext.getConfiguration().get();
```
ndinu, line 199: Do not call the `get` method without checking if the optional is present or not.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/AbstractReturnDelegate.java
```diff
@@ -220,7 +230,9 @@ private Object streamingContent(Object value,
     if (value instanceof InputStream) {
       ConnectionHandler connectionHandler = (ConnectionHandler) operationContext.getVariable(CONNECTION_PARAM);
       if (connectionHandler != null && supportsStreaming(operationContext.getComponentModel())) {
-        value = new ConnectedInputStreamWrapper((InputStream) value, connectionHandler);
+        ConfigurationInstance config = (ConfigurationInstance) operationContext.getConfiguration().get();
```
ndinu, line 233: Try not to repeat this piece of code.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/AbstractReturnDelegate.java
```diff
@@ -280,6 +304,7 @@ public void close() throws IOException {
         in = new ClosedInputStream();
       } finally {
         connectionHandler.release();
+        onClose.run();
```
ndinu, line 307: What happens if close is called multiple times?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/streaming/PagingProviderProducer.java
```diff
@@ -96,6 +93,9 @@ public PagingProviderProducer(PagingProvider<Object, T> delegate,
   @Override
   public List<T> produce() {
     List<T> page = performWithConnection(delegate::getPage);
+    if (isFirstPage) {
+      tryToMutateConfigurationStats(config, (MutableConfigurationStats::addOpenedStream));
```
ndinu, line 97: What if the first page is not produced right after the paging provider is created?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/streaming/PagingProviderProducer.java
```diff
@@ -193,6 +193,7 @@ public void close() {
       if (connectionSupplier != null) {
         safely(connectionSupplier::close, e -> LOGGER.debug("Found exception closing the connection supplier", e));
       }
+      tryToMutateConfigurationStats(config, (MutableConfigurationStats::discountOpenedStream));
```
ndinu, line 196: Are we sure this is called only once?

ndinu (changes requested):

PR #9637, opened 2020-10-23
MULE-18905: PollingSourceWrapper clears the content of the wrong object store

SebaElizalde:


ndinu:
--validate

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -351,31 +358,34 @@ private boolean passesWatermark(DefaultPollItem pollItem) {
               Serializable previousItemWatermark = recentlyProcessedIds.retrieve(itemId);
               if (compareWatermarks(itemWatermark, previousItemWatermark, watermarkComparator) <= 0) {
                 accept = false;
+              } else {
+                recentlyProcessedIds.remove(itemId);
```
ndinu, line 362: You are never comparing the item watermark to the updated watermark.

Let's say that currentWatermark is 5, and updatedWatermark is 10.

Maybe recentlyProcessed has the items with id A and watermark 5, and the same item A but with id 7 comes for processing, you are changing the updated watermark to 7. In the next poll, an item of watermark 8 would be accepted.

tests/test-extensions/petstore-extension/src/main/java/org/mule/test/petstore/extension/WatermarkingPetAdoptionSource.java
```diff
@@ -0,0 +1,97 @@
...
+import org.mule.runtime.extension.api.runtime.source.SourceCallbackContext;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@MetadataScope(outputResolver = PollingSourceMetadataResolver.class)
+@MediaType(TEXT_PLAIN)
+public class WatermarkingPetAdoptionSource extends PollingSource<String, Void> {
+
+  public static final List<String> ALL_WATERMARK_PETS = asList("Anibal", "ANIBAL", "Barbara");
```
ndinu, line 32: Shouldn't this be asList("Anibal", "Anibal", "Barbara");

ndinu (changes requested):

tests/test-extensions/petstore-extension/src/main/java/org/mule/test/petstore/extension/WatermarkingPetAdoptionSource.java
```diff
@@ -0,0 +1,103 @@
...
+import org.mule.runtime.extension.api.runtime.source.SourceCallbackContext;
+
+import java.util.List;
+
+@MetadataScope(outputResolver = PollingSourceMetadataResolver.class)
+@MediaType(TEXT_PLAIN)
+public class WatermarkingPetAdoptionSource extends PollingSource<String, Void> {
+
+  public static int STARTED_POLLS;
+  private static int index;
```
ndinu, line 31: I think index and polls should not be static fields.

ndinu (changes requested):

ndinu (approved):

PR #9667, opened 2020-11-03
MULE-18675: Dynamic Configurations are expired even though a paginated operation is still using it.

SebaElizalde:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DynamicConfigurationProvider.java
```diff
@@ -231,6 +231,7 @@ protected void registerConfiguration(ConfigurationInstance configuration) {
     cacheWriteLock.lock();
     try {
       return cache.entrySet().stream().filter(entry -> isExpired(entry.getValue())).map(entry -> {
+        System.out.println("Config " + entry.getValue() + " is expired. Invalidating and unregistering.");
```
SebaElizalde, line 234: removing this

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultMutableConfigurationStats.java
```diff
@@ -100,4 +109,22 @@ public int discountRunningSource() {
     updateLastUsed();
     return runningSources.decrementAndGet();
   }
+
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public int addActiveComponent() {
+    updateLastUsed();
```
marianogonzalez, line 118: if all the other methods are deprecated in favor of these new set of methods, then the other ones should delegate into this one instead of having what's essentially repeated code.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DynamicConfigurationProvider.java
```diff
@@ -242,7 +243,7 @@ protected void registerConfiguration(ConfigurationInstance configuration) {
 
   private boolean isExpired(ConfigurationInstance configuration) {
     ConfigurationStats stats = configuration.getStatistics();
-    return stats.getRunningSources() == 0 && stats.getInflightOperations() == 0
+    return stats.getRunningSources() == 0 && stats.getInflightOperations() == 0 && stats.getActiveComponents() == 0
```
marianogonzalez, line 246: same as before, it should be enough to just check for `stats.getActiveComponents()`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultExecutionMediator.java
```diff
@@ -102,6 +102,7 @@ public void execute(CompletableComponentExecutor<M> executor,
     final MutableConfigurationStats stats = getMutableConfigurationStats(context);
     if (stats != null) {
       stats.addInflightOperation();
```
marianogonzalez, line 104: remove

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultExecutionMediator.java
```diff
@@ -163,6 +164,7 @@ public void complete(Object value) {
         } finally {
           if (stats != null) {
             stats.discountInflightOperation();
```
marianogonzalez, line 166: remove

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultExecutionMediator.java
```diff
@@ -177,6 +179,7 @@ public void error(Throwable t) {
           } finally {
             if (stats != null) {
               stats.discountInflightOperation();
```
marianogonzalez, line 181: remove

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/result/AbstractReturnDelegate.java
```diff
@@ -293,13 +302,36 @@ protected MediaType resolveMediaType(Object value, MediaType contextMimeType, Ch
     return contextMimeType.withCharset(contextEncoding);
   }
 
+  private Runnable incrementActiveComponent(ExecutionContextAdapter executionContext) {
```
marianogonzalez, line 305: this is inefficient performance wise. You should check if the config is present or not before hand. If it's present, contruct a runnable using the config directly, avoiding the Optional#map call. If it's not present, do not create an empty `Runnable`,  simply pass `null` and have the stream check for that.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -682,11 +682,17 @@ private boolean shouldRunOnThisNode() {
 
   private Optional<ConfigurationInstance> startUsingConfiguration(CoreEvent event) {
     return getConfigurationAndTryToMutateStats(event,
-                                               (mutableConfigurationStats -> mutableConfigurationStats.addRunningSource()));
+                                               (mutableConfigurationStats -> {
+                                                 mutableConfigurationStats.addRunningSource();
```
marianogonzalez, line 686: remove

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -682,11 +682,17 @@ private boolean shouldRunOnThisNode() {
...
+                                                 mutableConfigurationStats.addActiveComponent();
+                                               }));
   }
 
   private void stopUsingConfiguration(CoreEvent event) {
-    getConfigurationAndTryToMutateStats(event, (mutableConfigurationStats -> mutableConfigurationStats.discountRunningSource()));
+    getConfigurationAndTryToMutateStats(event, (mutableConfigurationStats -> {
+      mutableConfigurationStats.discountRunningSource();
```
marianogonzalez, line 693: remove

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/streaming/PagingProviderProducer.java
```diff
@@ -216,6 +226,14 @@ private Object getConnection(ConnectionSupplier connectionSupplier) {
     }
   }
 
+  private void tryToMutateConfigurationStats(ConfigurationInstance configurationInstance,
```
marianogonzalez, line 229: repeated code.. Use the utils one

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/DynamicConfigExpirationTestCase.java
```diff
@@ -85,6 +129,38 @@ public void expirationWorksAfterRestartingSource() throws Exception {
     doNotExpireDynamicConfigWithCustomExpirationUsedBySource();
   }
 
+  @Test
```
marianogonzalez, line 132: these test are not wrong, don't delete them, but they are only helpful for troubleshooting. These test are testing the implementation, not the behavior. This is missing tests that verify that a config which has exceeded its TTL, it's never expired because it still has an open stream. Afterwards, the same test needs to close the stream and verify that the config is indeed expired. Test coverage here is not enough

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/MutableConfigurationStats.java
```diff
@@ -26,29 +26,57 @@
...
+   * @deprecated Use {@link #discountActiveComponent()} instead.
    */
+  @Deprecated
   int discountRunningSource();
+
+  /**
+   * Increments the return value of {@link #getActiveComponents()} by one
+   *
+   * @return the new value
+   * @since 4.2.3 4.3.1
```
ndinu, line 71: Also 4.4.0

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/MutableConfigurationStats.java
```diff
@@ -26,29 +26,57 @@
...
+   */
+  int addActiveComponent();
+
+  /**
+   * Decrements the return value of {@link #getActiveComponents()} by one
+   *
+   * @return the new value
+   * @since 4.2.3 4.3.1
```
ndinu, line 79: Same here

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/result/AbstractReturnDelegate.java
```diff
@@ -293,13 +302,36 @@ protected MediaType resolveMediaType(Object value, MediaType contextMimeType, Ch
...
+    Optional<ConfigurationInstance> config = executionContext.getConfiguration();
+    return config
+        .<Runnable>map(configurationInstance -> () -> tryToMutateConfigurationStats(configurationInstance,
+                                                                                    (MutableConfigurationStats::addActiveComponent)))
+        .orElseGet(() -> () -> {
+        });
+  }
+
+  private Runnable decrementActiveComponent(ExecutionContextAdapter executionContext) {
```
ndinu, line 314: By reading the name of these methods, it seems that it changes the counter, I would change them to reflect that the runnable that modifies it is returned.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/streaming/PagingProviderProducer.java
```diff
@@ -141,6 +147,7 @@ public int getSize() {
       return function.apply(connection);
     } catch (Exception caughtException) {
       if (isFirstPage) {
+        tryToMutateConfigurationStats(config, (MutableConfigurationStats::discountActiveComponent));
```
ndinu, line 150: Can it happen that if we are in the first page, we are actually calling the discountActiveComponent twice?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -550,4 +553,20 @@ public static MutableConfigurationStats getMutableConfigurationStats(ExecutionCo
...
     }
   }
+
+  /**
+   * Calls a consumer of the {@link MutableConfigurationStats} for a given {@link ConfigurationInstance}
+   *
+   * @param configurationInstance the {@link ConfigurationInstance} from where to extract the stats
+   * @param mutableConfigurationStatsConsumer the {@link MutableConfigurationStats} consumer
+   *
+   * @since 4.2.3 4.3.1
```
ndinu, line 563: Also 4.4.0

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultMutableConfigurationStats.java
SebaElizalde: Wouldn't that break the contract for getRunningSources() and getInflightOperations() (see https://github.com/mulesoft/mule-extensions-api/pull/750/files) ? If addRunningSource() delegates to addActiveComponent() then the activeComponents counter would be raised and the runningSources would remain at 0. Then asking for getRunningSources() would return 0.
The deprecated components should preserve their behavior and thus their counters, I think.
I do agree with removing the checks mentioned in other comments.

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/DynamicConfigExpirationTestCase.java
SebaElizalde: Yes I 100% agree. Until https://www.mulesoft.org/jira/browse/MULE-18774 is solved, configs are not expired until the flow finishes, so the only option we have is to check the counter (thus checking implementation rather than behavior). Actually all the dataweave waits are not needed for these tests, they are just there for when the other issues get solved. But I don't think this is too clear, so I will split the tests in 2: 1 group checks behavior and the other group checks implementation with a comment that says to remove when MULE-18774 is fixed.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/streaming/PagingProviderProducer.java
SebaElizalde: No because if an exception is caught on getting the first page, the whole paging provider is disposed and a new one is created later.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultMutableConfigurationStats.java
marianogonzalez: yes, you're right

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/DynamicConfigExpirationTestCase.java
```diff
@@ -72,32 +70,12 @@ public CoreEvent process(CoreEvent event) throws MuleException {
...
-      synchronized (capturedConfigStates) {
-        if (streamingConfig == null) {
-          streamingConfig = (HeisenbergExtension) muleContext.getExtensionManager()
+      synchronized (capturedStats) {
+        if (config == null) {
+          config = (HeisenbergExtension) muleContext.getExtensionManager()
               .getConfiguration("heisenbergWithShortExpiration", event).getValue();
         }
-        capturedConfigStates.add(streamingConfig.getDispose());
+        capturedStats.add(config.getDispose());
```
marianogonzalez, line 78: this doesn't address my initial feedback

marianogonzalez (changes requested):

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/MoneyLaunderingOperation.java
```diff
@@ -251,6 +251,34 @@ public void close(HeisenbergConnection connection) {
     };
   }
 
+  public PagingProvider<HeisenbergConnection, HeisenbergExtension> getConfigFailingPagedOperation(@Config HeisenbergExtension config,
```
marianogonzalez, line 254: what's the rationale behind this?

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/HeisenbergOperations.java
```diff
@@ -543,6 +544,28 @@ public InputStream nameAsStream(@Config HeisenbergExtension config) {
     return new ByteArrayInputStream(sayMyName(config).getBytes());
   }
 
+  @MediaType(value = TEXT_PLAIN, strict = false)
+  public InputStream nameAsStreamConnected(@Config HeisenbergExtension config, @Connection HeisenbergConnection connection) {
```
marianogonzalez, line 548: what's the rationale behind this? This doesn't make much sense and keeps testing the implementation instead of the behavior

marianogonzalez (changes requested):

PR #9668, opened 2020-11-03
MULE-18905-4.3.x: PollingSourceWrapper clears the content of the wrong object store

SebaElizalde:


modules/extensions-spring-support/src/test/resources/models/petstore.json
```diff
@@ -23163,8 +23163,7 @@
             "resolverName": "OUTPUT_PAYLOAD",
             "requiresConnection": false,
             "requiresConfiguration": false
-          },
-          "partialTypeKeyResolver": false
```
ndinu: Why does this change?

tests/test-extensions/petstore-extension/src/main/java/org/mule/test/petstore/extension/WatermarkingPetAdoptionSource.java
```diff
@@ -33,7 +33,7 @@
 
   @Parameter
   @org.mule.runtime.extension.api.annotation.param.Optional(defaultValue = "false")
-  protected boolean watermark;
```
ndinu: Was there any bug that prevents to use this value here, but allows it on master? Is this already identified?

ndinu (approved):

PR #9676, opened 2020-11-04
MULE-18537: Log retries on retry policy

szaffarano:


core/src/main/java/org/mule/runtime/core/internal/retry/policies/SimpleRetryPolicy.java
```diff
@@ -194,4 +197,17 @@ protected AtomicInteger initialValue() {
...
+      LOGGER.info("Retrying execution of event, attempt {} of {}.", attempts,
+                  count != RETRY_COUNT_FOREVER ? valueOf(count) : "unlimited");
+    }
+  }
+
+  private void logRetriesExhausted() {
+    if (LOGGER.isInfoEnabled()) {
+      LOGGER.info("Retry attempts exhausted. Failing...");
+    }
+  }
```
elrodro83, line 212: This could be very verbose, maybe change to DEBUG? what was the original reason for removing this in the first place?

elrodro83 (changes requested):

core/src/main/java/org/mule/runtime/core/internal/retry/policies/SimpleRetryPolicy.java
szaffarano: It wasn't removed.  This piece of code didn't change between `4.2.2` and `4.3.x`.  The retries are logged [here](https://github.com/mulesoft/mule/blob/support/4.2.x/core/src/main/java/org/mule/runtime/core/internal/retry/policies/SimpleRetryPolicy.java#L123) in INFO, but they don't [here](https://github.com/mulesoft/mule/blob/support/4.2.x/core/src/main/java/org/mule/runtime/core/internal/retry/policies/SimpleRetryPolicy.java#L72).

szaffarano: I changed the logging level to DEBUG instead of INFO.

core/src/main/java/org/mule/runtime/core/internal/retry/policies/SimpleRetryPolicy.java
```diff
@@ -194,4 +197,17 @@ protected AtomicInteger initialValue() {
...
+
+  private void logRetrying(long attempts) {
+    if (LOGGER.isDebugEnabled()) {
+      LOGGER.debug("Retrying execution of event, attempt {} of {}.", attempts,
+                  count != RETRY_COUNT_FOREVER ? valueOf(count) : "unlimited");
+    }
+  }
+
+  private void logRetriesExhausted() {
+    if (LOGGER.isDebugEnabled()) {
```
elrodro83, line 209: for logging a constant string there is no need to ask for the level before

elrodro83 (changes requested):

elrodro83 (approved):

PR #9687, opened 2020-11-09
CMTS-98: DeclarationSession is not failing when parameter values for …

gsfernandes:
…values/metadata/sampleData have an expression


modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/metadata/MetadataComponentExecutor.java
```diff
@@ -63,6 +64,8 @@ public MetadataComponentExecutor(ConnectionManager connectionManager, Reflection
       ClassLoader extensionClassLoader = getClassLoader(artifactHelper.getExtensionModel(componentElementDeclaration));
 
       return resolveMetadata(componentModel, optionalConfigurationInstance, metadataKey, extensionClassLoader);
+    } catch (ExpressionNotSupportedException e) {
+      return failure(MetadataFailure.Builder.newFailure(e).withFailureCode(INVALID_METADATA_KEY).onKeys());
```
lucianoRM, line 68: static import

lucianoRM (changes requested):

lucianoRM (approved):

PR #9691, opened 2020-11-10
CMTS-55: Check if ValueProviderModel defines that requires a configuration and connection before calling Runtime Service on MTS side

lucianoRM:


gsfernandes (approved):

lucianoRM:
--validate

PR #9693, opened 2020-11-10
MULE-18631: Need to be able to set a limit on the number of events to be pushed per poll where N >= 1

ndinu:


modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/source/NegativePollingSourceLimitingTestCase.java
```diff
@@ -0,0 +1,50 @@
...
+import org.junit.runners.Parameterized;
+
+import org.mule.tck.junit4.rule.SystemProperty;
+import org.mule.test.module.extension.InvalidExtensionConfigTestCase;
+import org.mule.test.runner.RunnerDelegateTo;
+
+import java.util.Collection;
+
+@RunnerDelegateTo(Parameterized.class)
+public class NegativePollingSourceLimitingTestCase extends InvalidExtensionConfigTestCase {
```
Bardammu, line 22: Is it possible to test that this fails because the limit is not greater than one?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/WatermarkStatus.java
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.module.extension.internal.runtime.source.poll;
+
+public enum WatermarkStatus {
```
Bardammu, line 9: Do you think we should add a Javadoc for this?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/loader/AbstractJavaExtensionModelLoader.java
```diff
@@ -86,6 +88,7 @@
 public class AbstractJavaExtensionModelLoader extends ExtensionModelLoader {
 
   private static final boolean IGNORE_DISABLED = getProperty(DISABLE_SDK_IGNORE_COMPONENT) != null;
+  private final boolean ENABLE_POLLING_SOURCE_LIMIT = getProperty(ENABLE_SDK_POLLING_SOURCE_LIMIT) != null;
```
Bardammu, line 91: This can be static too

ndinu: The tests rely on the value of the property changing. What I can try to do is find a way for the tests to add the parameter to the ExtensionLoadingContext, instead of doing this through the system prop.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -522,6 +582,28 @@ private void updateWatermark(Serializable value, Comparator comparator) {
...
+    }
+
+    updateRecentlyProcessedIds();
+    watermarkObjectStore.store(WATERMARK_ITEM_OS_KEY, value);
+  }
+
+  private void setCurrentWatermarkAsMinimumRejectWatermark(Serializable minimumRejectedByLimitPassingWatermark)
+      throws ObjectStoreException {
+    if (watermarkObjectStore.contains(WATERMARK_ITEM_OS_KEY)) {
+      watermarkObjectStore.retrieve(WATERMARK_ITEM_OS_KEY);
```
SebaElizalde, line 601: Why retrieve if the value is not used? Did you leave this line by accident?

ndinu: My bad, removing it.

SebaElizalde (approved):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/PollingSourceDeclarationEnricher.java
```diff
@@ -60,26 +71,47 @@ protected void onSource(SourceDeclaration source) {
...
+        return parameter;
+      }
+
+      private ParameterDeclaration declareSchedulingStrategyParameter(ClassTypeLoader loader) {
+        ParameterDeclaration parameter = new ParameterDeclaration(SCHEDULING_STRATEGY_PARAMETER_NAME);
+        parameter.setDescription(SCHEDULING_STRATEGY_PARAMETER_DESCRIPTION);
+        parameter.setRequired(true);
+        parameter.setType(loader.load(SchedulingStrategy.class), false);
+        parameter.setExpressionSupport(NOT_SUPPORTED);
+        parameter.addModelProperty(new InfrastructureParameterModelProperty(10));
```
marianogonzalez, line 104: I know this is unrelated to this change, but we need to review what's this sequence number for and why simply 10... These magic numbers are super fragile. Meanwhile, put this into a constant

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConfigurer.java
```diff
@@ -152,4 +155,19 @@ public Source build(ValueResolvingContext context) throws MuleException {
...
+
+  private int resolverMaxItemsPerPoll(ResolverSet resolverSet, ValueResolvingContext context, CoreEvent event)
+      throws MuleException {
+    ValueResolver<?> valueResolver = resolverSet.getResolvers().get(POLLING_SOURCE_LIMIT_PARAMETER_NAME);
+    if (valueResolver == null) {
+      return Integer.MAX_VALUE;
+    } else {
+      int maxItemsPerPoll = (Integer) valueResolver.resolve(context);
+      if (maxItemsPerPoll < 1) {
+        throw new IllegalStateException(format("The %s parameter must have a value greater than 1",
```
marianogonzalez, line 167: `IllegalArgumentException`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -253,25 +269,113 @@ public PollItemStatus accept(Consumer<PollItem<T, A>> consumer) {
...
+      }
+      if (minimumRejectedByLimitPassingWatermark == null ||
+          compareWatermarks(itemWatermark, minimumRejectedByLimitPassingWatermark, watermarkComparator) < 0) {
+        LOGGER.debug("An item that passed all previous validations is being rejected by the poll limit and its watermark" +
+            "value will be stored so that is processed on future polls if sent for processing.");
+        minimumRejectedByLimitPassingWatermark = itemWatermark;
+      }
+    }
+
+    private void saveWatermarkValue(WatermarkStatus watermarkStatus, DefaultPollItem pollItem) {
```
marianogonzalez, line 312: isn't this change actually tackling the watermark issues rather than implementing the limit thing? If the answer is yes, then split into different PRs

tests/test-extensions/petstore-extension/src/main/java/org/mule/test/petstore/extension/PetAdoptionLimitingSource.java
```diff
@@ -0,0 +1,64 @@
...
+import org.mule.runtime.extension.api.runtime.operation.Result;
+import org.mule.runtime.extension.api.runtime.source.PollContext;
+import org.mule.runtime.extension.api.runtime.source.PollingSource;
+import org.mule.runtime.extension.api.runtime.source.SourceCallbackContext;
+
+import java.util.List;
+
+@MetadataScope(outputResolver = PollingSourceMetadataResolver.class)
+@MediaType(TEXT_PLAIN)
+public class PetAdoptionLimitingSource extends PollingSource<String, Integer> {
```
marianogonzalez, line 23: why a new source for this? It'd be great to test it with an existing one since the real use case is injecting this functionality into existing sources

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
ndinu: For the limit feature, I need to test watermarking without impacting the object stores, after that, if the item is not rejected due to the poll limit, the pertinent values will be stored. This is why, the passesWatermark method returns a WatermarkStatus, and later using that status we know what to store in the object stores.

tests/test-extensions/petstore-extension/src/main/java/org/mule/test/petstore/extension/PetAdoptionLimitingSource.java
ndinu: I created a new source because I wanted to know in which poll was each item polled, and I also wanted to have an easy way to configure the watermark of each item. 

I can take some other tests and parameterize them so that they run with other sources using this parameter with a value of 1 or 2 for example, would that be ok?

marianogonzalez (approved):

PR #9710, opened 2020-11-11
MULE-18875: ValueProvider/SampleData optional parameters declared in resolver are not included in the model

gabrieldalborgo:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/ValueProvidersParameterDeclarationEnricher.java
```diff
@@ -155,9 +162,9 @@ private void enrichParameter(OfValues resolverClass,
     paramDeclaration.addModelProperty(propertyBuilder.build());
 
     valueProviderModelConsumer
-        .accept(new ValueProviderModel(getRequiredParametersAliases(resolverParameters, containerParameterNames),
```
ndinu, line 158: I think we should add tests that assert the value provider model other than the check of the test extensions extension model.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/ValueProvidersParameterDeclarationEnricher.java
```diff
@@ -247,11 +254,17 @@ private void enrichParameter(OfValues resolverClass,
...
-        .filter(ExtensionParameter::isRequired)
-        .map(param -> parameterNames.getOrDefault(param.getName(), param.getName()))
+  private List<ParameterModel> getParametersModel(List<ExtensionParameter> parameterDeclarations,
+                                                  Map<String, String> parameterNames,
+                                                  List<ParameterDeclaration> allParameters) {
+    Map<String, Boolean> paramsInfo = parameterDeclarations.stream()
+        .collect(toMap(param -> parameterNames.getOrDefault(param.getName(), param.getName()), ExtensionParameter::isRequired));
+    return allParameters.stream()
+        .filter(param -> paramsInfo.containsKey(param.getName()))
+        .map(param -> new ImmutableParameterModel(param.getName(), param.getDescription(), param.getType(),
```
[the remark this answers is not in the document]
gabrieldalborgo: What do you have in mind? Inherit the default value from the real parameter? or allow to define a new default value at the ValueProvider?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/ValueProvidersParameterDeclarationEnricher.java
```diff
@@ -247,11 +254,17 @@ private void enrichParameter(OfValues resolverClass,
...
+                                                  param.hasDynamicType(), paramsInfo.get(param.getName()),
```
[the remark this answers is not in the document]
gabrieldalborgo: Yes, that is right

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/ValueProvidersParameterDeclarationEnricher.java
gabrieldalborgo: Where would you create those tests?

ndinu: I would create a test for the enricher where the extension declaration is mocked, and then enriched, then assert the enriched declaration.

gabrieldalborgo: Tests added

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/ValueProvidersParameterDeclarationEnricher.java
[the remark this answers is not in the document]
ndinu: I had in mind setting the default value of the real parameter, which is the value that will be taken for values resolution is none is provided.

modules/extensions-support/pom.xml
```diff
@@ -211,6 +218,24 @@
...
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mule.tests</groupId>
+            <artifactId>mule-values-extension</artifactId>
+            <version>4.4.0-SNAPSHOT</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mule.tests</groupId>
```
marianogonzalez, line 228: repeated dependency

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/SampleDataDeclarationEnricher.java
```diff
@@ -156,11 +161,17 @@ private SampleDataProviderModel createSampleDataModel(SampleData annotation, Exe
...
-        .filter(ExtensionParameter::isRequired)
-        .map(param -> parameterNames.getOrDefault(param.getName(), param.getName()))
+  private List<ParameterModel> getParametersModel(List<ExtensionParameter> parameterDeclarations,
+                                                  Map<String, String> parameterNames,
+                                                  List<ParameterDeclaration> allParameters) {
+    Map<String, Boolean> paramsInfo = parameterDeclarations.stream()
+        .collect(toMap(param -> parameterNames.getOrDefault(param.getName(), param.getName()), ExtensionParameter::isRequired));
+    return allParameters.stream()
+        .filter(param -> paramsInfo.containsKey(param.getName()))
+        .map(param -> new ImmutableParameterModel(param.getName(), param.getDescription(), param.getType(),
```
marianogonzalez, line 171: you should use the aliased name here instead of `param.getName()`.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/ValueProvidersParameterDeclarationEnricher.java
```diff
@@ -247,11 +254,19 @@ private void enrichParameter(OfValues resolverClass,
...
-  private List<String> getRequiredParametersAliases(List<ExtensionParameter> parameterDeclarations,
-                                                    Map<String, String> parameterNames) {
-    return parameterDeclarations.stream()
-        .filter(ExtensionParameter::isRequired)
-        .map(param -> parameterNames.getOrDefault(param.getName(), param.getName()))
+  private List<ParameterModel> getParametersModel(List<ExtensionParameter> parameterDeclarations,
+                                                  Map<String, String> parameterNames,
+                                                  List<ParameterDeclaration> allParameters) {
+    Map<String, Boolean> paramsInfo = parameterDeclarations.stream()
+        .collect(toMap(param -> parameterNames.getOrDefault(param.getName(), param.getName()), ExtensionParameter::isRequired));
```
marianogonzalez, line 261: same as before. You're loosing the aliased name. Several test cases should have broken because of this. If not, we have a coverage problem you should fix as part of this PR

marianogonzalez (changes requested):

modules/extensions-support/pom.xml
```diff
@@ -169,6 +169,13 @@
             <classifier>mule-plugin</classifier>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mule.tests</groupId>
+            <artifactId>mule-values-extension</artifactId>
+            <version>4.4.0-SNAPSHOT</version>
+            <classifier>mule-plugin</classifier>
```
marianogonzalez, line 176: why do you need both the mule-plugin and jar dependencies?

marianogonzalez (changes requested):

marianogonzalez (approved):

gabrieldalborgo:
--validate

modules/extensions-spring-support/src/test/resources/models/metadata.json
```diff
@@ -38647,7 +38647,6 @@
       "format": "xml",
       "type": "Object",
       "annotations": {
-        "typeId": "#root:shiporder",
```
marianogonzalez, line 38650: Why was this moved?

marianogonzalez (changes requested):

modules/extensions-spring-support/src/test/resources/models/metadata.json
gabrieldalborgo: I'm not sure, this happens when the extension model is regenerated. It is something that is already in the master. I will create a ticket to review it.
Also, I reverted the changes to this file.

marianogonzalez (approved):

PR #9724, opened 2020-11-13
MULE-18985: ValueProvider and Metadata should return a failureErrorCo…

gsfernandes:
…de=CONNECTION_FAILURE when a connection cannot be established/created from the context

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataInputDelegate.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataInputDelegate.java
@@ -11,21 +11,21 @@ import static java.util.stream.Collectors.toList;
 import static org.mule.metadata.api.utils.MetadataTypeUtils.isCollection;
 import static org.mule.metadata.api.utils.MetadataTypeUtils.isNullType;
 import static org.mule.metadata.java.api.utils.JavaTypeUtils.getType;
+import static org.mule.runtime.api.metadata.resolving.FailureCode.CONNECTION_FAILURE;
 import static org.mule.runtime.api.metadata.resolving.FailureCode.NO_DYNAMIC_TYPE_AVAILABLE;
 import static org.mule.runtime.api.metadata.resolving.MetadataFailure.Builder.newFailure;
 import static org.mule.runtime.api.metadata.resolving.MetadataResult.failure;
 import static org.mule.runtime.api.metadata.resolving.MetadataResult.success;
 import static org.mule.runtime.module.extension.internal.metadata.MetadataResolverUtils.resolveWithOAuthRefresh;
-
 import org.mule.metadata.api.model.MetadataType;
 import org.mule.metadata.java.api.annotation.ClassInformationAnnotation;
+import org.mule.runtime.api.connection.ConnectionException;
 import org.mule.runtime.api.meta.model.ComponentModel;
 import org.mule.runtime.api.meta.model.EnrichableModel;
 import org.mule.runtime.api.meta.model.parameter.ParameterModel;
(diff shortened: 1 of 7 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataInputDelegate.java
```diff
@@ -136,14 +138,16 @@ NamedTypeResolver getParameterResolver(String parameterName) {
           .withReason(NULL_TYPE_ERROR)
           .onParameter(parameter.getName());
       return failure(parameter.getType(), failure);
+    } catch (ConnectionException e) {
+      return failure(newFailure(e).withMessage("Failed to establish connection: " + ExceptionUtils.getMessage(e))
+          .withFailureCode(CONNECTION_FAILURE).onKeys());
```
lucianoRM, line 143: shouldn't this be onComponent()? since we are resolving input metadata? I think the onKeys() should be used only when resolving metadata keys.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataOutputDelegate.java
```diff
@@ -140,6 +143,9 @@
           .withReason(NULL_TYPE_ERROR)
           .onOutputPayload();
       return failure(output.getType(), failure);
+    } catch (ConnectionException e) {
+      return failure(newFailure(e).withMessage("Failed to establish connection: " + ExceptionUtils.getMessage(e))
+          .withFailureCode(CONNECTION_FAILURE).onKeys());
```
lucianoRM, line 148: same here

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataOutputDelegate.java
```diff
@@ -175,6 +181,9 @@
           .withReason(NULL_TYPE_ERROR)
           .onOutputAttributes();
       return failure(failure);
+    } catch (ConnectionException e) {
+      return failure(newFailure(e).withMessage("Failed to establish connection: " + ExceptionUtils.getMessage(e))
+          .withFailureCode(CONNECTION_FAILURE).onKeys());
```
lucianoRM, line 186: and here

lucianoRM (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataOutputDelegate.java
gsfernandes: Good catch. Yes, it should be onComponent().

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataOutputDelegate.java
gsfernandes: Good catch. Yes, it should be onComponent().

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataInputDelegate.java
gsfernandes: Good catch. Yes, it should be onComponent().

gsfernandes:
@lucianoRM @ndinu I have made more changes to handle errors from OAuth Service.

lucianoRM (approved):

lucianoRM (approved):

ndinu (approved):

PR #9733, opened 2020-11-16
MULE-18906: Watermark in polling sources is updated when source is stopping.

SebaElizalde:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -603,14 +607,19 @@ private void setCurrentWatermarkAsMinimumRejectWatermark(Serializable minimumRej
...
     try {
       osClearingLock.lock();
       List<String> strings = idsOnUpdatedWatermark.allKeys();
-      recentlyProcessedIds.clear();
+      if (clearRecentlyProcessed) {
+        recentlyProcessedIds.clear();
+      }
       strings.forEach(key -> {
         try {
+          if (!clearRecentlyProcessed && recentlyProcessedIds.contains(key)) {
```
Bardammu, line 620: We can try to retrieve all the values of `recentlyProcessedIds ` and then call contains for each key. In this way, we avoid making a request for each key to OSv2.
Not in this PR, it's something we should think about.

SebaElizalde:
Re-done at https://github.com/mulesoft/mule/pull/9736.

PR #9734, opened 2020-11-16
MULE-18995: org.mule.sdk.api package should be PARENT_FIRST

ndinu:


gsfernandes (approved):

modules/container/src/main/java/org/mule/runtime/container/internal/ContainerClassLoaderFactory.java
```diff
@@ -193,9 +192,9 @@ protected ArtifactClassLoader createArtifactClassLoader(final ClassLoader parent
    */
   private LookupStrategy getSpecialLookupStrategy(String exportedPackage) {
     // If an extension uses a class provided by the mule-sdk-api artifact, the container classloader should use
-    // the class with which the extension was compiled.
+    // the class with which the extension was compiled only if the class is not present in the distribution.
     if (exportedPackage.startsWith(MULE_SDK_API_PACKAGE)) {
-      return CHILD_FIRST;
```
elrodro83, line 198: Was there no specific reason for this be this way? I think we talked about this in its moment, but don't remember right now.

ndinu: Yes, I though that when both the distribution and the extension had a version of a class, the version of the extension should be used. But I was wrong.

One of the side effects is that extensions that use a different sdk-api version than the distro have issues creating the extension model because the annotations do not match when tested.

elrodro83 (approved):

PR #9745, opened 2020-11-18
MULE-18906-4.3.x: Watermark in polling sources is updated when source is stopping. 

SebaElizalde:


SebaElizalde:
--validate

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -254,7 +256,9 @@ public PollItemStatus accept(Consumer<PollItem<T, A>> consumer) {
       pollItem.validate();
 
       PollItemStatus status;
-      if (!acquireItem(pollItem, callbackContext)) {
+      if (isRequestedToStop()) {
```
ndinu, line 259: We should remove the other call to this method.

SebaElizalde: nice catch. thanks.

ndinu (approved):

PR #9795, opened 2020-11-26
MULE-18765: Dynamic Configurations are expired even though a paginated operation is still using it.

SebaElizalde:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultMutableConfigurationStats.java
```diff
@@ -89,6 +100,7 @@ public int discountInflightOperation() {
   @Override
   public int addRunningSource() {
```
ndinu, line 101: I would implement these deprecated methods as, for example:

addActiveComponent();
return runningSources.incrementAndGet();

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/MutableConfigurationStats.java
```diff
@@ -26,29 +26,57 @@
...
+  @Deprecated
   int discountInflightOperation();
 
   /**
    * Increments the return value of {@link #getRunningSources()} by one
    *
    * @return the new value
    * @since 4.1.6 4.2.2 4.3.0
+   *
+   * @deprecated Use {@link #discountActiveComponent()} instead.
```
ndinu, line 51: addActiveComponent instead of discount

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/MutableConfigurationStats.java
```diff
@@ -26,29 +26,57 @@
    * Increments the return value of {@link #getInflightOperations()} by one
    *
    * @return the new value
+   *
+   * @deprecated Use {@link #discountActiveComponent()} instead.
```
ndinu, line 30: addActiveComponent instead of discount

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/HeisenbergOperations.java
```diff
@@ -611,10 +609,25 @@ public void close(HeisenbergConnection connection) throws MuleException {
     };
   }
 
-  public HeisenbergExtension getConfig(@Config HeisenbergExtension config, @Optional(defaultValue = "false") boolean fail) {
-    if (fail) {
-      throw new MuleRuntimeException(createStaticMessage("Failed returning the config"));
-    }
-    return config;
+  @MediaType(value = TEXT_PLAIN, strict = false)
```
ndinu, line 612: Why do we remove the old operation?

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/MoneyLaunderingOperation.java
```diff
@@ -232,7 +234,7 @@ public void close(HeisenbergConnection heisenbergConnection) throws MuleExceptio
         if (getPageCalls == failOn) {
           throw new ModuleException(CONNECTIVITY, new ConnectionException("Failed to retrieve Page"));
         }
-        if (getPageCalls > 2) {
+        if (getPageCalls > 3) {
```
ndinu, line 237: We have to be careful with this, are we sure we are not impacting other tests?

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/DynamicConfigExpirationTestCase.java
```diff
@@ -31,7 +35,7 @@
 
 public class DynamicConfigExpirationTestCase extends AbstractExtensionFunctionalTestCase {
 
-  private static final int PROBER_TIMEOUT = 20000;
+  private static final int PROBER_TIMEOUT = 30000;
```
ndinu, line 38: Can we make the tests run faster? it seems unnecessary for them to take more than 20 seconds.

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/MoneyLaunderingOperation.java
```diff
@@ -223,7 +223,9 @@ public void close(HeisenbergConnection heisenbergConnection) throws MuleExceptio
     };
   }
 
-  public PagingProvider<HeisenbergConnection, Integer> failAtClosePagedOperation(Integer failOn) {
+  public PagingProvider<HeisenbergConnection, Integer> failingConnectivityPagedOperation(Integer failOn,
```
ndinu, line 226: By changing the old operation, you had to change all the tests that used it, why not create a new one?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/result/AbstractReturnDelegate.java
```diff
@@ -145,10 +149,10 @@ protected Message toMessage(Object value, ExecutionContextAdapter operationConte
       if (resultValue.getOutput() instanceof InputStream) {
         ConnectionHandler connectionHandler = (ConnectionHandler) operationContext.getVariable(CONNECTION_PARAM);
         if (connectionHandler != null && supportsStreaming(operationContext.getComponentModel())) {
-          resultValue = resultValue.copy()
-              .output(componentDecoratorFactory
-                  .decorateOutput(new ConnectedInputStreamWrapper((InputStream) resultValue.getOutput(), connectionHandler),
-                                  event.getCorrelationId()))
+          resultValue = resultValue
+              .copy().output(new ConnectedInputStreamWrapper((InputStream) resultValue.getOutput(), connectionHandler,
```
ndinu, line 153: This still needs to be decorated by the componentDecoratorFactory

ndinu (dismissed):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultMutableConfigurationStats.java
SebaElizalde: sure

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/HeisenbergOperations.java
SebaElizalde: I added to test https://www.mulesoft.org/jira/browse/MULE-18984 and get the config used. But this cannot be done for a streamed operation and also I found another way of getting the config during the tests `config = (HeisenbergExtension) muleContext.getExtensionManager().getConfiguration("heisenbergWithShortExpiration", event).getValue();` making this operation unnecessary. So I removed it to prevent the connector from getting even more bloated with operations and stuff.

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/MoneyLaunderingOperation.java
SebaElizalde: yeah checked it

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/MoneyLaunderingOperation.java
SebaElizalde: For reuse because the new operation I needed was almost identical to this one and I was trying to reduce the bloat on the heisenberg connector, and it was easy to change the tests that used it.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/result/AbstractReturnDelegate.java
SebaElizalde: right, my bad

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/DynamicConfigExpirationTestCase.java
SebaElizalde: sure

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultMutableConfigurationStats.java
```diff
@@ -80,7 +80,7 @@ public int getActiveComponents() {
   @Override
   public int addInflightOperation() {
     updateLastUsed();
```
ndinu, line 82: There is no need to call the "updateLastUsed" method on all the deprecated methods, since the add/discount activeComponente will end up updating it anyway.

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/DynamicConfigExpirationTestCase.java
```diff
@@ -127,14 +132,38 @@ public void expirationWorksAfterRestartingSource() throws Exception {
 
   @Test
   public void dynamicConfigIsExpiredBeforeFlowEnds() throws Exception {
-    flowRunner("dynamicConfigIsExpiredBeforeFlowEnds").withPayload("Walter Blanco").run();
-    assertDisposedStatuses();
+    flowRunner("dynamicConfigIsExpiredBeforeFlowEnds").run();
+    assertDisposedStatuses(Arrays.asList(0, 1));
```
marianogonzalez, line 136: static import

modules/extensions-spring-support/src/test/resources/dynamic-config-expiration.xml
```diff
@@ -8,7 +8,7 @@
       http://www.mulesoft.org/schema/mule/test http://www.mulesoft.org/schema/mule/test/current/mule-test.xsd">
 
     <configuration>
-        <dynamic-config-expiration frequency="1000" timeUnit="MILLISECONDS">
+        <dynamic-config-expiration frequency="500" timeUnit="MILLISECONDS">
```
marianogonzalez, line 11: why reduce this? Can't this make the test flaky when running in jenkins?

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/PagedOperationExecutionTestCase.java
```diff
@@ -121,7 +121,7 @@ public void pagingProviderIsClosedSafelyDuringExceptionOnFirstPage() throws Exce
   @Test
   public void pagingProviderIsClosedSafelyAfterDataSourceIsFullyConsumed() throws Exception {
     resetCounters();
-    flowRunner("consumeFailAtClosePagedOperation").withPayload(4).run();
+    flowRunner("consumeFailAtClosePagedOperation").withPayload(-1).run();
```
marianogonzalez, line 124: why this change? Weird that no other assertion changed

modules/extensions-spring-support/src/test/resources/dynamic-config-expiration.xml
```diff
@@ -140,4 +166,42 @@
         <logger level="ERROR" message="#[ import * from dw::Runtime --- 'Waiting for 10000ms' wait 10000]"/>
     </flow>
 
+    <flow name="dynamicWithShortExpirationForPagedOperation">
+        <heisenberg:failing-connectivity-paged-operation failOn="#[vars.failOn]" failAtClose="false" config-ref="heisenbergWithShortExpiration"/>
+        <test:processor class="org.mule.test.module.extension.DynamicConfigExpirationTestCase$CaptureProcessor"/>
+        <foreach collection="#[payload]">
+            <logger level="ERROR" message="#[payload]" />
```
marianogonzalez, line 173: remove?

modules/extensions-spring-support/src/test/resources/dynamic-config-expiration.xml
```diff
@@ -140,4 +166,42 @@
...
+            <test:processor class="org.mule.test.module.extension.DynamicConfigExpirationTestCase$CaptureProcessor"/>
+            <logger level="ERROR" message="#[%dw 2.0
```
marianogonzalez, line 175: this is unnecesary. The test:processor operation has a waitTime parameter which automatically sleeps

modules/extensions-spring-support/src/test/resources/dynamic-config-expiration.xml
```diff
@@ -140,4 +166,42 @@
...
+                            import * from dw::Runtime
+                            output application/json
+                            ---
+                            'Waiting for 5000 milliseconds' wait 4000]"/>
+        </foreach>
+        <logger level="ERROR" message="#[%dw 2.0
```
marianogonzalez, line 181: same... remove

modules/extensions-spring-support/src/test/resources/dynamic-config-expiration.xml
```diff
@@ -140,4 +166,42 @@
...
+                            output application/json
+                            ---
+                            'Waiting for 5000 milliseconds' wait 4000]"/>
+        <test:processor class="org.mule.test.module.extension.DynamicConfigExpirationTestCase$CaptureProcessor"/>
+    </flow>
+
+    <flow name="dynamicWithShortExpirationForStreamingOperation">
+        <heisenberg:name-as-stream-connected config-ref="heisenbergWithShortExpiration" />
+        <test:processor class="org.mule.test.module.extension.DynamicConfigExpirationTestCase$CaptureProcessor"/>
+        <logger level="ERROR" message="#[%dw 2.0
```
marianogonzalez, line 192: remove these sleeps all over the place

modules/extensions-spring-support/src/test/resources/heisenberg-paged-operation-config.xml
```diff
@@ -53,16 +53,16 @@
     </flow>
 
     <flow name="consumeFailAtClosePagedOperation">
-        <heisenberg:fail-at-close-paged-operation failOn="#[payload]" config-ref="heisenberg"/>
+        <heisenberg:failing-connectivity-paged-operation failOn="#[payload]" config-ref="heisenberg"/>
```
marianogonzalez, line 56: isn't this changing the test condition? You should leave these as they are and add new ones instead

modules/extensions-spring-support/src/test/resources/heisenberg-paged-operation-config.xml
```diff
@@ -53,16 +53,16 @@
...
         <foreach collection="#[payload]">
             <logger level="ERROR" message="#[payload]"/>
         </foreach>
     </flow>
 
     <flow name="failAtClosePagedOperation">
-        <heisenberg:fail-at-close-paged-operation failOn="#[payload]" config-ref="heisenberg">
+        <heisenberg:failing-connectivity-paged-operation failOn="#[payload]" config-ref="heisenberg">
```
marianogonzalez, line 63: same

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/result/AbstractReturnDelegate.java
```diff
@@ -255,7 +263,10 @@ private Object streamingContent(Object value,
       ConnectionHandler connectionHandler = (ConnectionHandler) operationContext.getVariable(CONNECTION_PARAM);
       if (connectionHandler != null && supportsStreaming(operationContext.getComponentModel())) {
         value = componentDecoratorFactory
-            .decorateOutput(new ConnectedInputStreamWrapper((InputStream) value, connectionHandler), correlationId);
+            .decorateOutput(new ConnectedInputStreamWrapper((InputStream) value, connectionHandler,
```
marianogonzalez, line 266: what's changed here? if it's only formatting then revert

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/result/AbstractReturnDelegate.java
```diff
@@ -293,13 +304,36 @@ protected MediaType resolveMediaType(Object value, MediaType contextMimeType, Ch
     return contextMimeType.withCharset(contextEncoding);
   }
 
+  private Runnable getIncrementActiveComponent(ExecutionContextAdapter executionContext) {
```
marianogonzalez, line 307: refactor per comments in `ConnectedInputStreamWrapper`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/result/AbstractReturnDelegate.java
```diff
@@ -293,13 +304,36 @@ protected MediaType resolveMediaType(Object value, MediaType contextMimeType, Ch
...
+    Optional<ConfigurationInstance> config = executionContext.getConfiguration();
+    if (config.isPresent()) {
+      return () -> tryToMutateConfigurationStats(config.get(), (MutableConfigurationStats::addActiveComponent));
+    }
+    return null;
+  }
+
+  private Runnable getDecrementActiveComponent(ExecutionContextAdapter executionContext) {
```
marianogonzalez, line 315: rename: getDecrementActiveComponentTask

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/result/AbstractReturnDelegate.java
```diff
@@ -293,13 +304,36 @@ protected MediaType resolveMediaType(Object value, MediaType contextMimeType, Ch
...
+    private final Runnable onClose;
+    private AtomicBoolean alreadyClosed = new AtomicBoolean(false);
 
-    private ConnectedInputStreamWrapper(InputStream delegate, ConnectionHandler<?> connectionHandler) {
+    private ConnectedInputStreamWrapper(InputStream delegate, ConnectionHandler<?> connectionHandler, Runnable onCreate,
+                                        Runnable onClose) {
       super(delegate);
       this.connectionHandler = connectionHandler;
+      this.onClose = onClose;
+      if (onCreate != null) {
```
marianogonzalez, line 334: this doesn't make sense. If you want this task executed when the instance is created, then call it after doing the `new`... This should not be the constructor's responsibility

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/result/AbstractReturnDelegate.java
```diff
@@ -322,6 +356,9 @@ public void close() throws IOException {
         in = new ClosedInputStream();
       } finally {
         connectionHandler.release();
+        if (onClose != null && alreadyClosed.compareAndSet(false, true)) {
```
marianogonzalez, line 359: this should be a try.finally of it's own.. .this needs to happen even if `connectionHandler.release()` fails

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/streaming/PagingProviderProducer.java
```diff
@@ -83,6 +88,7 @@ public PagingProviderProducer(PagingProvider<Object, T> delegate,
     this.supportsOAuth = supportsOAuth;
     retryPolicy = (RetryPolicyTemplate) executionContext.getRetryPolicyTemplate().orElseGet(NoRetryPolicyTemplate::new);
     connectionSupplierFactory = createConnectionSupplierFactory();
+    tryToMutateConfigurationStats(config, (MutableConfigurationStats::addActiveComponent));
```
marianogonzalez, line 91: this `tryToMutateConfigurationStats` is overkill. Just extract the the stats object in the constructor and make sure it's mutable there. No need to extract it, test it and instantiate a lamba every time

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -550,4 +553,20 @@ public static MutableConfigurationStats getMutableConfigurationStats(ExecutionCo
...
+
+  /**
+   * Calls a consumer of the {@link MutableConfigurationStats} for a given {@link ConfigurationInstance}
+   *
+   * @param configurationInstance the {@link ConfigurationInstance} from where to extract the stats
+   * @param mutableConfigurationStatsConsumer the {@link MutableConfigurationStats} consumer
+   *
+   * @since 4.2.3 4.3.1 4.4.0
+   */
+  public static void tryToMutateConfigurationStats(ConfigurationInstance configurationInstance,
```
marianogonzalez, line 565: delete

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultMutableConfigurationStats.java
SebaElizalde: I thought the intention behind calling the add/discount method directly was to abstract from its implementation, but here we are using the knowledge that it also calls updateLastUsed. Also if I delete the line updateLastUsed it will be just like before de PR review only that we shifted some code around.

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/HeisenbergOperations.java
```diff
@@ -611,10 +609,25 @@ public void close(HeisenbergConnection connection) throws MuleException {
...
+  public InputStream nameAsStreamConnected(@Config HeisenbergExtension config, @Connection HeisenbergConnection connection) {
+    return new InputStream() {
+
+      private int timesRead = 0;
```
marianogonzalez, line 616: `timesRead` -> `bytesRead`

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/MoneyLaunderingOperation.java
marianogonzalez: revert this change. You're adding tests at the expense of others. These are not the same scenario

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/MoneyLaunderingOperation.java
marianogonzalez, line 237: why from 2 to 3?

marianogonzalez (changes requested):

modules/extensions-spring-support/src/test/resources/dynamic-config-expiration.xml
SebaElizalde: It is done to detect expired configs faster, otherwise expiration times need to be longer in order to accomodate for the case where it just polled and it needs to wait another full second for the re-check (this in turn makes the whole test suite take longer). Pretty sure it does not affect any test, and even if it does it would mean that the test was wrong or flaky.

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/PagedOperationExecutionTestCase.java
SebaElizalde: The 4 is a parameter indicating that on the 4th page it would fail, but on the 3rd page everything is consumed and there are no more pages, thus it does not fail. I think the -1 is more explicit of the intention "to never fail". I can revert it tho

modules/extensions-spring-support/src/test/resources/heisenberg-paged-operation-config.xml
SebaElizalde: okay

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/result/AbstractReturnDelegate.java
SebaElizalde: there is change

modules/extensions-spring-support/src/test/resources/dynamic-config-expiration.xml
SebaElizalde: yes but waitTime is not applied when you use a custom processor due to how FunctionalTestProcessor is coded, nor can you access the waitTime variable in the custom processor :/
I can replace the dataweave sleeps with <test:processor waitTime="1000"> but I think it is too much overhead maybe?

marianogonzalez: on the contrary... using DW for sleeping is the overkill thing

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultMutableConfigurationStats.java
marianogonzalez: I don't understand what you mean. This method needs to delegate to `addActiveComponent()` which will update the timestamp. I see no reason why `updateLastUsed()` should be called from here

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultMutableConfigurationStats.java
```diff
@@ -80,6 +90,7 @@ public int addInflightOperation() {
   @Override
   public int discountInflightOperation() {
     updateLastUsed();
```
marianogonzalez, line 92: remove

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultMutableConfigurationStats.java
```diff
@@ -89,6 +100,7 @@ public int discountInflightOperation() {
   @Override
   public int addRunningSource() {
     updateLastUsed();
```
marianogonzalez, line 102: remove

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultMutableConfigurationStats.java
```diff
@@ -98,6 +110,25 @@ public int addRunningSource() {
   @Override
   public int discountRunningSource() {
     updateLastUsed();
```
marianogonzalez, line 112: remove

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/result/AbstractReturnDelegate.java
```diff
@@ -149,13 +148,12 @@ protected Message toMessage(Object value, ExecutionContextAdapter operationConte
...
-              .copy().output(componentDecoratorFactory
-                  .decorateOutput(new ConnectedInputStreamWrapper((InputStream) resultValue.getOutput(), connectionHandler,
-                                                                  getIncrementActiveComponent(operationContext),
-                                                                  getDecrementActiveComponent(operationContext)),
-                                  event.getCorrelationId()))
-              .build();
+          ConnectedInputStreamWrapper streamWrapper =
+              new ConnectedInputStreamWrapper((InputStream) resultValue.getOutput(), connectionHandler,
+                                              getDecrementActiveComponentTask(operationContext));
+          incrementActiveComponent(operationContext);
```
marianogonzalez, line 154: incrementing this is not the return delegate's responsibility:

1) this class is used in other context to transform results to messages, and those shouldn't mess with the counters
2) If something fails AFTER the return delegate execution, nobody decrements it

The incrementing needs to happen in some other place, probably an interceptor which decrements it if the operation fails.

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/result/AbstractReturnDelegate.java
```diff
@@ -228,13 +226,13 @@ protected Message toMessage(Object value, ExecutionContextAdapter operationConte
...
-                                                                                      getIncrementActiveComponent(operationContext),
-                                                                                      getDecrementActiveComponent(operationContext)),
-                                                      cursorProviderFactory, event,
-                                                      operationContext.getComponent().getLocation()))
+          ConnectedInputStreamWrapper streamWrapper =
+              new ConnectedInputStreamWrapper(componentDecoratorFactory.decorateOutput((InputStream) result.getOutput(),
+                                                                                       event.getCorrelationId()),
+                                              connectionHandler, getDecrementActiveComponentTask(operationContext));
+          incrementActiveComponent(operationContext);
+          result = result.copy().output(StreamingUtils.streamingContent(streamWrapper, cursorProviderFactory, event,
```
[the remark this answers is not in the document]
SebaElizalde: there is another streamingContent method in this class.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -569,7 +574,42 @@ private boolean shouldUsePrecalculatedContext(CoreEvent event) {
   }
 
   protected void executeOperation(ExecutionContextAdapter<T> operationContext, ExecutorCallback callback) {
-    executionMediator.execute(componentExecutor, operationContext, callback);
+    final MutableConfigurationStats stats = getMutableConfigurationStats(operationContext);
```
marianogonzalez, line 577: feels like all of this should be happening inside the mediator. Actually, if you look at the mediators implementation, it will obtain the stats again and call addInfightOperation. That should not happen anymore. Move this into the mediator implementation and avoid unnecesary interactions with the stats object

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -569,7 +574,42 @@ private boolean shouldUsePrecalculatedContext(CoreEvent event) {
...
+
+      @Override
+      public void error(Throwable e) {
+        stats.discountActiveComponent();
+        callback.error(e);
+      }
+    };
+  }
+
+  private Boolean isConnectedStreamingOperation() {
```
marianogonzalez, line 605: move this to org.mule.runtime.module.extension.internal.util.MuleExtensionUtils

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/streaming/PagingProviderProducer.java
```diff
@@ -83,6 +88,9 @@ public PagingProviderProducer(PagingProvider<Object, T> delegate,
     this.supportsOAuth = supportsOAuth;
     retryPolicy = (RetryPolicyTemplate) executionContext.getRetryPolicyTemplate().orElseGet(NoRetryPolicyTemplate::new);
     connectionSupplierFactory = createConnectionSupplierFactory();
+    ConfigurationStats configurationStats = config.getStatistics();
+    mutableStats =
```
marianogonzalez, line 92: this ternary (or something very similar) is showing up in a lot of places. Extract to `org.mule.runtime.module.extension.internal.util.MuleExtensionUtils`

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
SebaElizalde: Moved it to the mediator, but the call to addInfightOperation is still necessary there to avoid breaking the api.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultExecutionMediator.java
```diff
@@ -101,6 +103,7 @@ public void execute(CompletableComponentExecutor<M> executor,
 
     final MutableConfigurationStats stats = getMutableConfigurationStats(context);
     if (stats != null) {
+      stats.addActiveComponent();
       stats.addInflightOperation();
```
marianogonzalez, line 107: Shouldn't this be removed? You already counted the active component. Why count twice?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultExecutionMediator.java
```diff
@@ -152,6 +155,9 @@ public void complete(Object value) {
         // after() method cannot be invoked in the finally. Needs to be explicitly called before completing the callback.
         // Race conditions appear otherwise, specially in connection pooling scenarios.
         if (stats != null) {
+          if (!isConnectedStreamingOperation(operationModel)) {
+            stats.discountActiveComponent();
+          }
           stats.discountInflightOperation();
```
marianogonzalez, line 161: same here. This line should go

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultExecutionMediator.java
```diff
@@ -173,6 +179,7 @@ public void error(Throwable t) {
         } finally {
           if (stats != null) {
             stats.discountInflightOperation();
```
marianogonzalez, line 181: same

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -550,4 +554,19 @@ public static MutableConfigurationStats getMutableConfigurationStats(ExecutionCo
       return null;
     }
   }
+
+  /**
+   * @return whether the {@link ComponentModel} defines a streaming operation that uses a connection.
+   *
+   * @since 4.2.3 - 4.3.0
+   */
+  public static Boolean isConnectedStreamingOperation(ComponentModel componentModel) {
```
marianogonzalez, line 563: don't use boxed boolean

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DefaultExecutionMediator.java
SebaElizalde: They count different things, I left the old one to preserve the ConfigurationStats api which states:
  /**
   * @return How many currently executing operations are making use of the referenced configuration
   *
   * @deprecated Use {@link #getActiveComponents()} instead.
   */
  @Deprecated
  int getInflightOperations();

marianogonzalez (approved):

PR #9804, opened 2020-11-30
Connection is not properly invalidated during a source reconnection

gabrieldalborgo:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -311,12 +312,15 @@ public void onException(ConnectionException exception) {
                        sourceAdapter.getName(), getLocation().getRootContainerName()),
                 exception);
 
+    if (!(sourceAdapter.getDelegate() instanceof Reconnectable)) {
```
ndinu, line 315: Instead of getting the delegate, and checking if it is a Reconnectable, I think it is better to check if sourceAdapter.getReconnectionAction(exception) is empty.

fsgonz (approved):

PR #9812, opened 2020-12-01
CMTS-130: ValuesResolver should return a unique failure code when opt…

gsfernandes:
…ions defined by acting parameters value are not valid


modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/AbstractParameterResolverExecutor.java
```diff
@@ -42,7 +42,7 @@
   protected final ReflectionCache reflectionCache;
   protected final ArtifactHelper artifactHelper;
 
-  public static final String INVALID_PARAMETER_VALUE = "INVALID_PARAMETER_VALUE";
+  public static final String INVALID_ACTING_PARAMETER_VALUE = "INVALID_ACTING_PARAMETER_VALUE";
```
lucianoRM, line 45: I don't get why the renaming. is the INVALID_PARAMETER_VALUE used somewhere else and we want to distinguish them?

PR #9829, opened 2020-12-04
MULE-19032: While doing test connectivity for OCS connections if the token has been expired a refresh token is not triggered

ndinu:


marianogonzalez (changes requested):
Issue ID needs to be MULE

marianogonzalez (approved):

PR #9834, opened 2020-12-09
MULE-19031: fix variable escaping

jcgrethe:


fsgonz (changes requested):
Can we add a test? And also if possible a test involving the target in the response in http (which uses the enricher).

fsgonz (approved):

PR #9835, opened 2020-12-09
CMTS-99: Support default values for DeclarationServices

gsfernandes:


gsfernandes:
--validate

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/metadata/MetadataKeyDeclarationResolver.java
```diff
@@ -43,20 +50,34 @@
  */
 public class MetadataKeyDeclarationResolver {
 
+  private static final Logger LOGGER = LoggerFactory.getLogger(MetadataKeyDeclarationResolver.class);
```
lucianoRM, line 53: static import

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/metadata/MetadataKeyDeclarationResolver.java
```diff
@@ -68,15 +89,31 @@ public MetadataKeyResult resolveKeyResult() {
...
     for (MetadataKeyInfo keyInfo : keyPartModelsInfo) {
-      String id;
+      String id = null;
+      boolean isNullValue = false;
       if (keyPartValues.containsKey(keyInfo.parameterModel.getName())) {
         id = keyPartValues.get(keyInfo.parameterModel.getName());
+        isNullValue = id == null;
+      } else if (!partialKey && keyInfo.parameterModel.getDefaultValue() != null) {
+        String defaultValue = keyInfo.parameterModel.getDefaultValue().toString();
+        if (expressionManager != null && isExpression(defaultValue)) {
```
lucianoRM, line 99: why this validation? When is the expressionManager going to be null? and why shouldn't we fail if that is the case?

lucianoRM (changes requested):

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/metadata/MetadataKeyDeclarationResolver.java
gsfernandes: We use this resolver from MTS to validate that the key is "complete" before calling Runtime. So, on that scenario we won't need to resolve the expression.

lucianoRM (approved):

PR #9839, opened 2020-12-10
MULE-19039: Foreach need to use different correlationId for every iteration

pabloperalta:


core/src/main/java/org/mule/routing/outbound/AbstractMessageSequenceSplitter.java
```diff
@@ -153,7 +159,7 @@ public Object transform(Object input)
                 if ((!correlationSet && (enableCorrelation == CorrelationMode.IF_NOT_SET))
                         || (enableCorrelation == CorrelationMode.ALWAYS))
                 {
-                    setMessageCorrelationId(message, correlationId, correlationSequence);
+                        setMessageCorrelationId(message, correlationId, correlationSequence);
```
szaffarano, line 162: Please indent.

PR #9857, opened 2020-12-11
CMTS-133: If a deployment exception or lifecycle exception is thrown …

gsfernandes:
…when creating the backed application for a design session the exceptions should be thrown instead of failures with UNKNOWN error code

### DIFF
--- a/modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/sampledata/SampleDataExecutor.java
+++ b/modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/sampledata/SampleDataExecutor.java
@@ -6,6 +6,7 @@
  */
 package org.mule.runtime.module.tooling.internal.artifact.sampledata;
 
+import static com.google.common.base.Throwables.propagateIfPossible;
 import static java.lang.String.format;
 import static java.util.Optional.empty;
 import static java.util.Optional.ofNullable;
@@ -37,6 +38,8 @@ import org.mule.sdk.api.data.sample.SampleDataException;
 
 import java.util.Optional;
 import java.util.function.Supplier;
+
+import com.google.common.base.Throwables;
 
 public class SampleDataExecutor extends AbstractParameterResolverExecutor {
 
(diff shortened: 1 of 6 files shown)


modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/sampledata/SampleDataExecutor.java
```diff
@@ -81,9 +83,9 @@ public SampleDataResult getSampleData(ComponentModel componentModel, ComponentEl
       if (cause instanceof SampleDataException) {
         failureBuilder.withFailureCode(((SampleDataException) cause).getFailureCode());
       }
```
lucianoRM, line 85: This failure is not needed anymore since the exception will be thrown

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/value/ValueProviderExecutor.java
```diff
@@ -87,9 +89,9 @@ public ValueResult resolveValues(ParameterizedModel parameterizedModel,
       if (cause instanceof ValueResolvingException) {
         failureBuilder.withFailureCode(((ValueResolvingException) cause).getFailureCode());
       }
```
lucianoRM, line 91: same here

lucianoRM (changes requested):

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/sampledata/SampleDataExecutor.java
gsfernandes: No, actually the way the connector has to return a "failure" is by throwing this exception which has a an error code. If any other exception is thrown we should propagate it but this transformation and exception handling is needed here.

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/value/ValueProviderExecutor.java
gsfernandes: No, actually the way the connector has to return a "failure" is by throwing this exception which has a an error code. If any other exception is thrown we should propagate it but this transformation and exception handling is needed here.

lucianoRM (approved):

PR #9859, opened 2020-12-15
MULE-19055: SampleDataProvider validator is not validating generic

marianogonzalez:


ndinu (dismissed):

ndinu (dismissed):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/SampleDataModelValidator.java
```diff
@@ -144,6 +157,38 @@ private void validateResolver(ConnectableComponentModel model,
...
+        .map(this::asString)
+        .collect(joining(",")) + ">";
+  }
+
+  private String asString(Type type) {
+    if (type instanceof ParameterizedTypeImpl) {
+      ParameterizedTypeImpl parameterizedType = (ParameterizedTypeImpl) type;
+      return parameterizedType.getRawType().getName() + asGenericSignature(asList(parameterizedType.getActualTypeArguments()));
+    } else {
+      return type != null ? type.getTypeName() : Object.class.getName();
```
ndinu, line 188: Maybe when the type is null, and is the generic of the SampleDataProvider, we should add an error that states that the SampleDataProviderGenerics must be explicit.

marianogonzalez: This is for cases in which one of the provider's generic types is a Map or some other parameterized type where we cannot take it upon ourselves to demand proper use of those clases

ndinu (approved):

marianogonzalez:
--validate

marianogonzalez:
--validate

marianogonzalez:
--validate

marianogonzalez:
--validate

marianogonzalez:
--validate

PR #9864, opened 2020-12-17
CMTS-139: DeclarationSession should fail to be created if a required …

gsfernandes:
…parameter for a configuration/connection is not present


ndinu (approved):

gsfernandes:
--validate

PR #9865, opened 2020-12-17
CMTS-140: MetadataKeys and ValueProviders should preserver order of v…

gsfernandes:
…alues from connector's resolver

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/ValueProviderMediator.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/ValueProviderMediator.java
@@ -7,6 +7,7 @@
 package org.mule.runtime.module.extension.internal.value;
 
 import static java.lang.String.format;
+import static java.util.stream.Collectors.toCollection;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
 import static org.mule.runtime.extension.api.values.ValueResolvingException.INVALID_VALUE_RESOLVER_NAME;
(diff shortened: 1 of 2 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/ValueProviderMediator.java
```diff
@@ -156,7 +159,7 @@ public ValueProviderMediator(T containerModel, Supplier<MuleContext> muleContext
     return valueSet.stream()
         .map(option -> cloneAndEnrichValue(option, parameters))
         .map(ValueBuilder::build)
-        .collect(toSet());
+        .collect(toCollection(LinkedHashSet::new));
```
lucianoRM, line 162: I don't think this is correct. Yes, once the LinkedHashSet is created, the order will be preserved, but if the ValueProvider returns a plain HashSet, results will be in a different order every time. 
Shouldn't we sort the results from the valueSet too? Just in case the VP does not return an ordered set? 
Maybe this is like this by design and we want to keep the order of the results only if the VP wants them in order. Otherwise, we do not care.

lucianoRM (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/ValueProviderMediator.java
gsfernandes: Exactly, that's the case. As an initial support with the current API we would like to preserve the order but that only happens if the resolver/provider has returned a sorted set. If not, it is exactly as it is right now, order is not warrantied therefore each call to the resolver/provider would return different orders because it doesn't matter.

lucianoRM (approved):

gsfernandes:
--validate

PR #9876, opened 2020-12-22
MULE-19073: Exception in Catch Exception Strategy does not roll back transaction

pabloperalta:


core/src/main/java/org/mule/exception/TemplateMessagingExceptionStrategy.java
```diff
@@ -191,12 +192,12 @@ protected MuleEvent route(MuleEvent event, Exception t)
             catch (Exception e)
             {
                 logFatal(event, e);
+                throw new MessagingException(CoreMessages.createStaticMessage("There was an error processing the catch strategy"), event, e);
```
fsgonz, line 195: static import

core/src/test/java/org/mule/exception/CatchMessagingExceptionStrategyTestCase.java
```diff
@@ -115,9 +117,15 @@ public void testHandleExceptionWithConfiguredMessageProcessors() throws Exceptio
     @Test
     public void testHandleExceptionWithMessageProcessorsChangingEvent() throws Exception
     {
+        when((Class<String>) mockMuleEvent.getMessage().getDataType().getType()).thenReturn(String.class);
```
fsgonz, line 120: Why is this needed?

modules/cxf/src/test/resources/catch-exception-strategy-flow-conf-httpn.xml
```diff
@@ -83,7 +83,7 @@
             <processor-chain>
                 <script:component>
                     <script:script engine="groovy">
-                        return message.getOriginalPayload()
+                        return new String(message.getOriginalPayload())
```
fsgonz, line 86: Why is this needed?

fsgonz (changes requested):

modules/cxf/src/test/resources/catch-exception-strategy-flow-conf-httpn.xml
pabloperalta: No longer needed

core/src/test/java/org/mule/exception/CatchMessagingExceptionStrategyTestCase.java
pabloperalta: No longer needed

mbuchwald (approved):

PR #9889, opened 2020-12-28
MULE-19088: NPE executing an intercepted extension operation

IvanAndresFritzler:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/DefaultExecutionContext.java
```diff
@@ -181,8 +183,17 @@ public CoreEvent getEvent() {
   @Override
   public void changeEvent(CoreEvent updated) {
     requireNonNull(event);
-    event = updated;
     securityContext = event.getSecurityContext();
```
elrodro83, line 186: should get this securityContext from `updated` rather than `event`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/DefaultExecutionContext.java
```diff
@@ -181,8 +183,17 @@ public CoreEvent getEvent() {
...
+    // Prevents possible lack of SdkContext if this context was generated by an interceptor (see MULE-19088)
```
elrodro83, line 187: in the comment, point to the place in the code where the interceptor is creating the event without the context

elrodro83 (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/DefaultExecutionContext.java
IvanAndresFritzler: Nice catch :)

elrodro83 (approved):

IvanAndresFritzler:
--validate

PR #9891, opened 2020-12-28
MULE-18976: Implement Feature Flagging (add system property support)

szaffarano:


core/src/main/java/org/mule/runtime/core/internal/config/FeatureFlaggingServiceBuilder.java
```diff
@@ -49,9 +51,17 @@ public FeatureFlaggingService build() {
 
     final String id = context.getConfiguration().getId();
     configurations.forEach((feature, p) -> {
-      boolean enabled = p.test(context);
-
-      LOGGER.debug("Setting feature {} = {} for artifact [{}]", feature, enabled, id);
+      boolean enabled;
+
+      String systemPropertyName = feature.getOverridingSystemPropertyName().orElse(null);
+      if (systemPropertyName != null && getProperty(systemPropertyName) != null) {
```
elrodro83, line 57: why `systemPropertyName != null` here if you already have an optional in the previous line?

elrodro83 (changes requested):

elrodro83 (approved):

PR #9893, opened 2020-12-28
MULE-19090: SampleDataModelValidator yields false negatives when operation is paginated

marianogonzalez:


marianogonzalez:
--validate

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/SampleDataModelValidator.java
```diff
@@ -181,13 +206,95 @@ private String asGenericSignature(List<Type> types) {
...
+      return true;
+    }
+
+    return true;
+  }
+
+  private Pair<Type, Type> getPagingProviderGenerics(ConnectableComponentModel model) {
+    return model.getModelProperty(ImplementingMethodModelProperty.class)
+        .map(mp -> {
+          ParameterizedType type = (ParameterizedType) mp.getMethod().getGenericReturnType();
```
ndinu, line 269: Will this always be a ParameterizedType?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/SampleDataModelValidator.java
```diff
@@ -181,13 +206,95 @@ private String asGenericSignature(List<Type> types) {
...
+          return new Pair<>(type.getActualTypeArguments()[0], type.getActualTypeArguments()[1]);
```
ndinu, line 270: Should we check the size of the array?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/SampleDataModelValidator.java
marianogonzalez: yes. At this point you already now that the type will be a `PagingProvider` which is parameterized

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/SampleDataModelValidator.java
marianogonzalez: you already know this will be a `PagingProvider`, thus you already know the array will have two positions.

ndinu (approved):

aiannucci:
--validate

PR #9900, opened 2020-12-30
MULE-18559: Support @RuntimeVersion annotation

Bardammu:
This commit makes it possible to inject a instance of MuleVersion on fields annotated with
@RuntimeVersion

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/RefNameDeclarationEnricher.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/RefNameDeclarationEnricher.java
@@ -6,24 +6,17 @@
  */
 package org.mule.runtime.module.extension.internal.loader.enricher;
 
-import static org.apache.commons.collections.CollectionUtils.isEmpty;
-import static org.reflections.ReflectionUtils.getAllFields;
-import static org.reflections.ReflectionUtils.withAnnotation;
-
-import org.mule.runtime.api.meta.model.declaration.fluent.BaseDeclaration;
+import org.mule.runtime.api.meta.model.ModelProperty;
 import org.mule.runtime.api.meta.model.declaration.fluent.ConfigurationDeclaration;
 import org.mule.runtime.api.meta.model.declaration.fluent.ConnectionProviderDeclaration;
-import org.mule.runtime.api.meta.model.declaration.fluent.SourceDeclaration;
 import org.mule.runtime.extension.api.annotation.param.RefName;
 import org.mule.runtime.extension.api.declaration.fluent.util.IdempotentDeclarationWalker;
 import org.mule.runtime.extension.api.exception.IllegalConfigurationModelDefinitionException;
 import org.mule.runtime.extension.api.loader.DeclarationEnricher;
 import org.mule.runtime.extension.api.loader.ExtensionLoadingContext;
-import org.mule.runtime.module.extension.internal.loader.java.property.ImplementingTypeModelProperty;
 import org.mule.runtime.module.extension.internal.loader.java.property.RequireNameField;
 
 import java.lang.reflect.Field;
-import java.util.Collection;
 
 /**
  * A {@link DeclarationEnricher} which looks for configurations with fields annotated with {@link RefName}.
@@ -56,16 +49,17 @@ public final class RefNameDeclarationEnricher extends AbstractAnnotatedFieldDecl
   }
 
   @Override
-  protected void doEnrich(BaseDeclaration<?> declaration) {
-    declaration.getModelProperty(ImplementingTypeModelProperty.class).ifPresent(typeProperty -> {
-      Collection<Field> fields = getAllFields(typeProperty.getType(), withAnnotation(RefName.class));
-      if (isEmpty(fields)) {
-        return;
-      }
+  protected ModelProperty getModelProperty(Field field) {
+    return new RequireNameField(field);
+  }
 
-      validate(fields, typeProperty, RefName.class, String.class);
+  @Override
+  protected Class getAnnotation() {
+    return RefName.class;
+  }
 
-      declaration.addModelProperty(new RequireNameField(fields.iterator().next()));
-    });
+  @Override
+  protected Class getImplementingClass() {
+    return String.class;
   }
 }
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/RefNameDeclarationEnricher.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/RefNameDeclarationEnricher.java
@@ -6,28 +6,24 @@
  */
 package org.mule.runtime.module.extension.internal.loader.enricher;
 
-import static java.util.stream.Collectors.toList;
 import static org.apache.commons.collections.CollectionUtils.isEmpty;
-import static org.mule.runtime.extension.api.loader.DeclarationEnricherPhase.POST_STRUCTURE;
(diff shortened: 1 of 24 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/RuntimeVersionDeclarationEnricher.java
```diff
@@ -0,0 +1,100 @@
...
+    declaration.getModelProperty(ImplementingTypeModelProperty.class).ifPresent(p -> {
+      ImplementingTypeModelProperty typeProperty = (ImplementingTypeModelProperty) p;
+      Collection<Field> fields = getAllFields(typeProperty.getType(), withAnnotation(RuntimeVersion.class));
+      if (isEmpty(fields)) {
+        return;
+      }
+
+      if (fields.size() > 1) {
+        throw new IllegalConfigurationModelDefinitionException(String
+            .format("Only one field is allowed to be annotated with @%s, but class '%s' has %d fields "
```
marianogonzalez, line 82: static import

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/property/RuntimeVersionModelProperty.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package org.mule.runtime.module.extension.internal.loader.java.property;
+
+import org.mule.runtime.api.meta.model.config.ConfigurationModel;
+import org.mule.sdk.api.annotation.param.RuntimeVersion;
+
+import java.lang.reflect.Field;
+
+/**
+ * A custom model property used to indicate that the class implementing a {@link ConfigurationModel} has a {@link Field}
```
marianogonzalez, line 15: what is a custom model property? Did you mean private?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/InjectedFieldsModelValidator.java
```diff
@@ -92,8 +96,10 @@ protected void onOperation(HasOperationModels owner, OperationModel model) {
 
           @Override
           protected void onConnectionProvider(HasConnectionProviderModels owner, ConnectionProviderModel model) {
-            validateFields(model, getImplementingType(model), DefaultEncoding.class);
-            validateFields(model, getImplementingType(model), RefName.class);
+            Optional<Class> implementingType = getImplementingType(model);
+            validateFields(model, implementingType, DefaultEncoding.class, String.class);
```
marianogonzalez, line 100: instead of calling this method 3 times, shouldn't it be refactored so that all the annotations and types are passed together and the list of fields is iterated only once?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConfigurer.java
```diff
@@ -114,7 +117,9 @@ protected Object instantiateObject() {
           @Override
           public Object build(ValueResolvingContext context) throws MuleException {
             Object source = build(resolverSet.resolve(context));
+            MuleVersion muleVersion = new MuleVersion(MuleExtensionModelProvider.MULE_VERSION);
```
marianogonzalez, line 120: Static import

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/RuntimeVersionDeclarationEnricher.java
```diff
@@ -0,0 +1,100 @@
...
+                + "with such annotation. Offending fields are: [%s]", RuntimeVersion.class.getSimpleName(),
+                    typeProperty.getType().getName(),
+                    fields.size(), Joiner.on(", ").join(fields.stream().map(Field::getName).collect(toList()))));
+      }
+
+      final Field runtimeVersionField = fields.iterator().next();
+      if (!MuleVersion.class.equals(runtimeVersionField.getType())) {
+        throw new IllegalConfigurationModelDefinitionException(String
+            .format("Class '%s' declares the field '%s' which is annotated with @%s and is of type '%s'. Only "
+                + "fields of type String are allowed to carry such annotation", typeProperty.getType().getName(),
```
gabrieldalborgo, line 92: Should it be MuleVersion instead of String?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/RuntimeVersionDeclarationEnricher.java
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * {@link DeclarationEnricher} which looks at classes with fields annotated with {@link RuntimeVersion}. It validates
+ * that the annotations is used properly and if so it adds a {@link RuntimeVersionModelProperty}.
+ * <p>
+ * If the {@link RuntimeVersion} annotation is used in a way which breaks the rules set on its javadoc, an
+ * {@link IllegalConfigurationModelDefinitionException} will be thrown.
+ *
+ * @since 4.4
+ */
+public class RuntimeVersionDeclarationEnricher implements DeclarationEnricher {
```
marianogonzalez, line 43: this is basically a copy paste from `RefNameDeclarationEnricher`. Create a base class for both and avoid code duplication

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/RefNameDeclarationEnricher.java
```diff
@@ -60,33 +51,21 @@ public void onConfiguration(ConfigurationDeclaration declaration) {
...
+
     }.walk(extensionLoadingContext.getExtensionDeclarer().getDeclaration());
   }
 
-  private void doEnrich(BaseDeclaration<?> declaration) {
-
+  @Override
+  protected void doEnrich(BaseDeclaration<?> declaration) {
     declaration.getModelProperty(ImplementingTypeModelProperty.class).ifPresent(typeProperty -> {
       Collection<Field> fields = getAllFields(typeProperty.getType(), withAnnotation(RefName.class));
```
marianogonzalez, line 61: there's still a lot of duplicated code. Can't the part of fetching the ImplementType and extracting the fields also be reused by simply passing the Annotation class to filter upon?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/InjectedFieldsModelValidator.java
marianogonzalez: can you explain this change? this doesn't provide the proposed optimization

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/InjectedFieldsModelValidator.java
Bardammu: I will have to pass the `ConnectionProviderModel`, the `implementingType`, and a list or map with the types and annotations. It looks quite ugly to me. Is there any benefit between calling three times a  method and a foreach that requires creating the list and map?

On the other hand it's not exactly the same of parameters, so I'll end up with code duplication anyway

marianogonzalez (approved):

PR #9923, opened 2021-01-04
MULE-19107: fix null pointer exception when a header is null

jcgrethe:


modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseHeaderBuilder.java
```diff
@@ -62,6 +62,12 @@ else if (headerValue instanceof String[])
         }
         else
         {
+            if (headerValue == null)
+            {
+                LOGGER.warn("The header " + headerName + " is null. Removing it from the response.");
```
eze210, line 67: Where is it removed? Is the header removed if you do this?
builder.addHeader("name", "not null");
builder.addHeader("name", null);

modules/http/src/test/java/org/mule/module/http/internal/listener/HttpResponseHeaderBuilderTestCase.java
```diff
@@ -0,0 +1,16 @@
+package org.mule.module.http.internal.listener;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertTrue;
+
+public class HttpResponseHeaderBuilderTestCase {
+
+    @Test
+    public void testNullHeader()
```
eze210, line 10: nullHeaderValueIsIgnored()

modules/http/src/test/java/org/mule/module/http/internal/listener/HttpResponseHeaderBuilderTestCase.java
```diff
@@ -0,0 +1,16 @@
+package org.mule.module.http.internal.listener;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertTrue;
+
+public class HttpResponseHeaderBuilderTestCase {
+
+    @Test
```
eze210, line 9: add @Issue("MULE-19017")

eze210 (changes requested):

modules/http/src/main/java/org/mule/module/http/internal/listener/HttpResponseHeaderBuilder.java
jcgrethe: Is not remove, the header is not added to the list.

eze210 (approved):

PR #9949, opened 2021-01-07
MULE-19122: Access Tokens are overwritten when using dynamic configs with client credentials

SebaElizalde:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/oauth/clientcredentials/ClientCredentialsOAuthHandler.java
```diff
@@ -134,7 +134,7 @@ private ClientCredentialsOAuthDancer createDancer(ClientCredentialsConfig config
         .responseExpiresInExpr(grantType.getExpirationRegex())
         .responseAccessTokenExpr(grantType.getAccessTokenExpr())
         .withClientCredentialsIn(toCredentialsLocation(grantType.getCredentialsPlacement()))
-        .resourceOwnerIdTransformer(ownerId -> ownerId + "-" + config.getOwnerConfigName());
+        .resourceOwnerIdTransformer(ownerId -> ownerId + "-" + config.getConfigIdentifier());
```
marianogonzalez, line 137: this isn't enough to generate the disambiguation you want. You could have two dynamic instances which are using the exact same OAuth configuration but differing in something else.

tests/test-extensions/test-oauth-extension/src/main/java/org/mule/test/oauth/ClientCredentialsConfig.java
```diff
@@ -6,13 +6,24 @@
...
+import org.mule.runtime.api.lifecycle.Disposable;
 import org.mule.runtime.extension.api.annotation.Configuration;
 import org.mule.runtime.extension.api.annotation.Operations;
 import org.mule.runtime.extension.api.annotation.connectivity.ConnectionProviders;
 
 @ConnectionProviders(TestOAuthClientCredentialsProvider.class)
 @Operations(TestOAuthOperations.class)
 @Configuration(name = "client-credentials")
-public class ClientCredentialsConfig {
+public class ClientCredentialsConfig implements Disposable {
```
marianogonzalez, line 17: where is the test case for this?

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/oauth/clientcredentials/ClientCredentialsOAuthHandler.java
SebaElizalde: But if they share the same OAuth config is it not okay for them to share the token? How does the OAuth protocol handle this?
Also two dynamic configs with the same oauth config are also currently sharing the same dancer:
`ClientCredentialsOAuthDancer dancer = dancers.get(config.getConfigIdentifier());`
Is that an issue as well?

tests/test-extensions/test-oauth-extension/src/main/java/org/mule/test/oauth/ClientCredentialsConfig.java
SebaElizalde: see https://github.com/mulesoft/mule-integration-tests/pull/1332/files

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/oauth/clientcredentials/ClientCredentialsOAuthHandler.java
marianogonzalez: Ok. I better understand your solution now. Disregard

marianogonzalez: The answer is it depends. Some Service provders like salesforce federate their apps, so the same set of credentials are good in both the testing and productin environments, but the tokens are not. Isn't it safer to actually segregate by configuration instance?

marianogonzalez: also, why is this problem exclusive to client credentials? Doesn't Authorization Code and OCS potentially have the same problem?

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/oauth/clientcredentials/ClientCredentialsOAuthHandler.java
SebaElizalde: For Authorization Code and OCS, dynamic parameters are not allowed, thus the config name would match to only 1 oauth config. But, from what you are saying now, I see that it can be a problem for those too, because other parts of the config can be dynamic so the same config name can match multiple config instance, and we dont want them to share the same token. I'll try to apply the segregation by config instance for all grant types.

marianogonzalez (approved):

PR #9950, opened 2021-01-07
MULE-19115: Optional parameter declared and annotated on SampleDataProvider is still defined as required in the model

gabrieldalborgo:


tests/test-extensions/sample-data-extension/src/main/java/org/mule/test/data/sample/extension/provider/GroupTestSampleDataProvider.java
```diff
@@ -8,7 +8,7 @@
 
 import static org.mule.test.data.sample.extension.SampleDataExtension.adaptLegacy;
 
-import org.mule.runtime.extension.api.annotation.param.Optional;
+import org.mule.sdk.api.annotation.param.Optional;
```
marianogonzalez, line 11: make sure to have tests that cover both cases... old and new `@Optional` annotations

marianogonzalez (changes requested):

tests/test-extensions/sample-data-extension/src/main/java/org/mule/test/data/sample/extension/provider/GroupTestSampleDataProvider.java
gabrieldalborgo: Yes, I had check it. There are both cases for VP's and SampleData

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/loader/java/type/ExtensionParameter.java
```diff
@@ -79,23 +80,23 @@ default boolean shouldBeAdvertised() {
...
     final java.util.Optional<Optional> annotation = getAnnotation(Optional.class);
     if (annotation.isPresent()) {
-      final Optional optionalAnnotation = annotation.get();
-      final String defaultValue = optionalAnnotation.defaultValue();
-      if (!defaultValue.equals(Optional.NULL)) {
-        optionalDefaultValue = java.util.Optional.of(defaultValue);
-      }
+      return getDefaultValue(annotation.get());
+    }
+    final java.util.Optional<org.mule.sdk.api.annotation.param.Optional> sdkAnnotation =
```
ndinu, line 94: If both are present, do you think we should use the value from the extensions-api annotation?

gabrieldalborgo: You're right. The other way makes more sense to me. I'll change it

ndinu (approved):

marianogonzalez (approved):

PR #9951, opened 2021-01-07
MULE-19123: CXF Module fails when request body is empty - Mule 3.9.4

pabloperalta:


modules/xml/src/test/java/org/mule/module/xml/util/XmlUtilsTestCase.java
```diff
@@ -6,10 +6,16 @@
  */
 package org.mule.module.xml.util;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.*;
```
szaffarano, line 10: Avoid using `*`

modules/xml/src/test/java/org/mule/module/xml/util/XmlUtilsTestCase.java
```diff
@@ -6,10 +6,16 @@
...
 import static org.junit.Assert.assertEquals;
+import static org.mule.module.xml.util.XMLUtils.*;
+import static org.mule.util.xmlsecurity.XMLSecureFactories.*;
```
szaffarano, line 13: Avoid using `*`

szaffarano (approved):

modules/xml/src/test/java/org/mule/module/xml/util/XmlUtilsTestCase.java
pabloperalta: Fixed

modules/xml/src/test/java/org/mule/module/xml/util/XmlUtilsTestCase.java
pabloperalta: Fixed

PR #9954, opened 2021-01-08
MULE-19123: CXF Module fails when request body is empty - Mule 3.9.4 (#9951)

pabloperalta:
(cherry picked from commit 283a75d02e2a00af340e8457e6214f6590f23b6f)


pabloperalta:
Thought of a better solution

PR #9962, opened 2021-01-11
MULE-18873: Scatter-Gather routes are dropping events during Flow-Ref…

IvanAndresFritzler:
… processing

(cherry picked from commit 79ce3b61d30e5d3a37c83faab8ac3f8cb99a0a33)


src/main/resources/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,31 @@
+Manifest-Version: 1.0
```
eze210, line 1: Why is this manifest needed?

eze210 (changes requested):

src/main/resources/META-INF/MANIFEST.MF
IvanAndresFritzler: Removing...

eze210 (approved):

PR #9976, opened 2021-01-12
MULE-19129: Race Condition in notification policy could make notifica…

mbuchwald:
…tions to not be send to listeners (#9968)


mlischetti (approved):

mbuchwald:
--validate

PR #9984, opened 2021-01-14
CMTS-155: Improve logs for DeclarationSession

gsfernandes:


modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/metadata/MetadataExecutor.java
```diff
@@ -73,7 +81,23 @@ protected MetadataContext createMetadataContext(Optional<ConfigurationInstance>
...
                                       new DefaultMetadataCache(),
-                                      new JavaTypeLoader(extensionClassLoader));
+                                      new JavaTypeLoader(extensionClassLoader)) {
+
+      @Override
+      public <C> Optional<C> getConnection() throws ConnectionException {
+        long startTime = currentTimeMillis();
+        try {
+          if (LOGGER.isDebugEnabled()) {
+            LOGGER.debug("Retrieving connection");
```
lucianoRM, line 91: I would change the messages to something easily readable without the need to read the whole message. 
Something like: 
Get Connection STARTED for ....
...
Get Connection FINISHED for ....
 
The same for the other methods

lucianoRM (changes requested):

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/metadata/MetadataExecutor.java
gsfernandes: Done!

lucianoRM (approved):

PR #10018, opened 2021-01-20
MULE-18561: Adjust sdk-api implicit argument handling

ndinu:


ndinu:
--validate

ndinu:
--validate

ndinu:
--validate

modules/extensions-spring-support/src/test/resources/models/semantic-terms-extension.json
```diff
@@ -3453,40 +3453,6 @@
...
-                    "order": 8
-                  },
-                  "allowedStereotypeModels": [
-                    {
-                      "type": "NTLM_DOMAIN",
-                      "namespace": "SEMANTIC-TERMS"
-                    }
-                  ],
-                  "semanticTerms": [],
-                  "name": "ntlmDomain",
```
ndinu, line 3487: Does anyone know why this parameter was added in the model? I see no reference to the class it is in.

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/capability/xml/description/ParameterDescriptionDocumenter.java
```diff
@@ -44,7 +45,8 @@ void document(ParameterizedDeclaration<?> parameterized, Element method, MethodD
 
     if (method instanceof ExecutableElement) {
       ((ExecutableElement) method).getParameters().stream()
-          .filter(e -> e.getAnnotation(ParameterGroup.class) != null)
+          .filter(e -> e.getAnnotation(ParameterGroup.class) != null
```
marianogonzalez, line 48: I'm putting this comment here but also take this as feedback for the entire PR (and for future Forward compatibility PRs as well). I know this is how we've been doing it in the past (me included) but I think that this pattern of having if clauses with `||` operators checking for one annotation or the other, or having collections in which we aggregate annotated components, followed by and `addAll`/`putAll` doing the same for another annotation, is getting totally out of hand and really affecting the code's readability. 

Please extract all of those things to other methods (either private in the same class, or at an utils class, doesn't matter) so that instead of having `if (isAnnotatedWith(X) or || isAnnotatedWithY)` we can have `if (someSemanticIsPresent()` instead. I think it's really important we keep syntax and semantics decoupled.

marianogonzalez (changes requested):

PR #10034, opened 2021-01-22
MULE-19161: Execution type blocking is ignored for defining if an operation is blocking.

fsgonz:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/utils/ModelLoaderUtils.java
```diff
@@ -45,6 +47,14 @@ public static boolean isRoute(ExtensionParameter parameter) {
   }
 
   public static boolean isNonBlocking(MethodElement method) {
+    return isCompletionCallbackIn(method) && !isBlockingExecutionType(method);
```
marianogonzalez, line 50: this is incorrect. The semantic of this method is to indicate if the operation is non blocking and therefore executes asynchronously. Adding the `@Execution(BLOCKING)` annotation does not change that. The only effect of that is sending the initial portion of the operation in the `IO` pool instead of `CPU_LIGHT`. The operation doesn't become blocking because of that. The only way a non blocking operation becomes blocking is if using a connection that is participating in an active transaction

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/utils/ModelLoaderUtils.java
fsgonz: Ok.

fsgonz:
Closed.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/utils/ModelLoaderUtils.java
facunlk: @marianogonzalez isn't that against what the documentation of the execution type says? [Here](https://github.com/mulesoft/mule-api/blob/master/src/main/java/org/mule/runtime/api/meta/model/operation/ExecutionType.java#L33) it states the following:
```
  /**
   * Blocking processing that use {@link Thread#sleep(long)}, {@link Lock#lock()}, blocking IO operations
   * or any other technique that blocks the current thread during processing.
   */
  BLOCKING
```
Based on that, it is fair to assume that developers using the SDK will expect the operation to be blocking.

marianogonzalez: That member is an option inside of a larger enum which says:

```
In order for Mule to determine the best way to execute different components, it
needs to know the type of work the components will be performing.
 ```
 
 This annotation is to describe the type of work. Not to tell the runtime what to make of that information. Extensions have zero control over how they are executed, by design.  A non blocking operation could still perform CPU intensive or IO work before it yields control to the OS, but that doesn't mean the operation isn't non blocking anymore, it just means that it might block for a while before unblocking

marianogonzalez: Additionally, there's another article in the SDK documentation which explains the heuristics that the SDK uses to automatically determine the correct execution type without user intervention (actually, the `@ExecutionType` is very rarely used). 

That article adds additional context on what this annotation means and when it should be used

PR #10083, opened 2021-02-09
MULE-19127: Fix: when flows are not stopped and apps are restarted, flow stopped …

sofiamorseletto:
…state should not be persisted


fsgonz (approved):

sofiamorseletto:
--validate

PR #10093, opened 2021-02-11
MULE-19185: Avoid multiple construction of CompositeClassloader in operation calls

eze210:


eze210:
This PR was sent to validate the performance degradation fixed in #10094

PR #10099, opened 2021-02-12
MULE-19124: ExceptionNotification doesn't have flow name.

SebaElizalde:


core/src/main/java/org/mule/runtime/core/api/exception/SystemExceptionHandler.java
```diff
@@ -25,6 +27,14 @@
...
    */
   void handleException(Exception exception);
+
+  /**
+   * Take some action when a system exception has occurred (i.e., there was no message in play when the exception occurred).
+   *
+   * @param exception which occurred
+   * @param componentLocation the {@link ComponentLocation} which produced the exception
+   */
+  void handleException(Exception exception, ComponentLocation componentLocation);
```
ndinu, line 37: We cannot add a new method to an existing  API interface, not at least without a default implementation that makes sense.

ndinu (changes requested):

fsgonz (approved):

mlischetti (approved):

core/api-changes.json
```diff
@@ -75,6 +75,15 @@
           "classSimpleName": "ReactiveInterceptor",
           "elementKind": "interface",
           "justification": "MULE-18177: This should have neven been made privileged"
+        },
+        {
+        "code": "java.method.addedToInterface",
```
ndinu, line 80: Is this necessary after adding the default implementation?

SebaElizalde: let me check

core/src/main/java/org/mule/runtime/core/api/exception/SystemExceptionHandler.java
```diff
@@ -25,6 +27,16 @@
    * @param exception which occurred
    */
   void handleException(Exception exception);
+
+  /**
+   * Take some action when a system exception has occurred (i.e., there was no message in play when the exception occurred).
+   *
+   * @param exception which occurred
+   * @param componentLocation the {@link ComponentLocation} which produced the exception
+   */
```
ndinu, line 36: add @since 4.4.0

core/src/main/java/org/mule/runtime/core/api/exception/SystemExceptionHandler.java
```diff
@@ -25,6 +27,16 @@
    * @param exception which occurred
    */
   void handleException(Exception exception);
+
+  /**
+   * Take some action when a system exception has occurred (i.e., there was no message in play when the exception occurred).
+   *
+   * @param exception which occurred
+   * @param componentLocation the {@link ComponentLocation} which produced the exception
```
ndinu, line 35: add since annotation.

ndinu (approved):

PR #10104, opened 2021-02-17
MULE-19157: Exclusive Parameter Group with one required fails validation when one of the optional parameters is a POJO.

SebaElizalde:


modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/parameter/group/ParameterGroupTestCase.java
```diff
@@ -1,28 +0,0 @@
...
-
-public class ParameterGroupTestCase extends AbstractExtensionFunctionalTestCase {
-
-  @Override
-  protected String getConfigFile() {
-    return "values/some-parameter-group-config.xml";
-  }
-
-  @Test
-  public void test() throws Exception {
```
SebaElizalde, line 24: Moved test to ParameterGroupExclusiveOptionalsOneRequiredTestCase

tests/test-extensions/some-extension/src/main/java/org/mule/test/some/extension/SomeOps.java
```diff
@@ -44,11 +44,18 @@ public void someOp(@Connection String conn, @Config SomeExtension ext) {}
    * <it>isOneRequired</it> property.
    *
    * @param oneParameterGroup some test operation config with a isOneRequired exclusive-optional configuration
-   * @return one of the configs arguments
+   * @return the config
    */
-  public String oneRequiredParameterResolverOperation(@ParameterGroup(
+  public SomeParameterGroupOneRequiredConfig oneRequiredParameterResolverOperationDslTrue(@ParameterGroup(
```
ndinu, line 49: Any particular reason not to create a new one instead of changing the existing one?

tests/test-extensions/some-extension/src/main/java/org/mule/test/some/extension/SomeEmittingSource.java
```diff
@@ -44,14 +45,19 @@
   Integer times;
   private Scheduler emitterScheduler;
 
+  @ParameterGroup(name = "Awesome Parameter Group")
+  SomeParameterGroupOneRequiredConfig someParameterGroup;
```
ndinu, line 49: Won't this make you change al tests that use this source?

SebaElizalde: yeah, already changed the 2 xml files within the mule project that used it. Could not find any usages of this test extension source outside of it.

tests/test-extensions/some-extension/src/main/java/org/mule/test/some/extension/SomeOps.java
SebaElizalde: Just for better clarity and readability. Also to avoid repetition.

ndinu: This also means that every usage of this operation probably needs to change. I would create a new one.

tests/test-extensions/some-extension/src/main/java/org/mule/test/some/extension/SomeEmittingSource.java
ndinu: I would create a new source that extends from this one and has this parameter group.

SebaElizalde: changed

SebaElizalde:
--validate

ndinu (approved):

PR #10122, opened 2021-02-19
MULE-19226: Policy not able to inject internal dependencies for artifacts that shares with the application

IvanAndresFritzler:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -464,9 +465,9 @@ public static ExtensionModel loadExtension(Class<?> clazz, Map<String, Object> p
     return new DefaultJavaExtensionModelLoader().loadExtensionModel(clazz.getClassLoader(), dslResolvingContext, params);
   }
 
-  public static String getImplicitConfigurationProviderName(ExtensionModel extensionModel,
+  public static String getImplicitConfigurationProviderName(MuleContext muleContext, ExtensionModel extensionModel,
```
elrodro83, line 468: get the id (better called `prefix`) directly instead of passing the `MuleContext` around

elrodro83 (changes requested):

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
```diff
@@ -58,8 +58,8 @@ public CompositeArtifactExtensionManager(ExtensionManager parentExtensionManager
     this.childExtensionManager = childExtensionManager;
 
     extensionModels = new HashSet<>();
-    extensionModels.addAll(parentExtensionManager.getExtensions());
     extensionModels.addAll(childExtensionManager.getExtensions());
+    extensionModels.addAll(parentExtensionManager.getExtensions());
```
marianogonzalez, line 62: why this change? this looks extra suspicious considering you're adding into a HashSet

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
```diff
@@ -132,12 +132,10 @@ public ConfigurationInstance getConfiguration(String configurationProviderName,
   public Optional<ConfigurationProvider> getConfigurationProvider(ExtensionModel extensionModel, ComponentModel componentModel) {
     Optional<ConfigurationProvider> configurationModel =
         childExtensionManager.getConfigurationProvider(extensionModel, componentModel);
-
-    if (!configurationModel.isPresent()) {
+    if (!configurationModel.isPresent() && !childExtensionManager.getExtension(extensionModel.getName()).isPresent()) {
```
marianogonzalez, line 135: pretty sure this breaks backwards compatibility when a config is defined at the domain level

modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/api/policy/PolicyTemplate.java
```diff
@@ -26,13 +26,8 @@
    */
   void dispose();
 
-  /**
-   * @return plugins deployed only inside the policy template
-   */
-  List<ArtifactPlugin> getArtifactPlugins();
```
marianogonzalez, line 32: Can't remove this. This is API

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -464,9 +465,10 @@ public static ExtensionModel loadExtension(Class<?> clazz, Map<String, Object> p
     return new DefaultJavaExtensionModelLoader().loadExtensionModel(clazz.getClassLoader(), dslResolvingContext, params);
   }
 
-  public static String getImplicitConfigurationProviderName(ExtensionModel extensionModel,
+  public static String getImplicitConfigurationProviderName(String prefix, ExtensionModel extensionModel,
                                                             ConfigurationModel implicitConfigurationModel) {
-    return format("%s-%s-implicit", extensionModel.getName(), implicitConfigurationModel.getName());
+    Preconditions.checkArgument(isNotEmpty(prefix), "A prefix must be provided");
```
marianogonzalez, line 470: static import

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -464,9 +465,10 @@ public static ExtensionModel loadExtension(Class<?> clazz, Map<String, Object> p
...
+    return format("%s-%s-%s-implicit", prefix, extensionModel.getName(), implicitConfigurationModel.getName());
```
marianogonzalez, line 471: doesn't this break for the domain use case?

marianogonzalez (changes requested):

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
IvanAndresFritzler: Classloaders for each extension are set using a property of their extension models. This change aims to use the policy classloader when the policy is executing and not the application one (note that the set uses the extension name as the key, so the order matters, because in the case of an application declaring the same extension than a policy, the policy extension model was not added to this set).

modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/api/policy/PolicyTemplate.java
IvanAndresFritzler: It's not removed (see that I also renamed getOwnArtifactPlugins to getArtifactPlugins)

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
marianogonzalez: This is not the correct way of achieving what you want:

* This is a HashSet, order is not guaranteed. This is would actually be flaky on different depending on the JDK version/vendor to use
* If you want a specific classloader to be used, then you need to reference it specifically (usually by setting it as the TCCL).

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
marianogonzalez: You can have the case in which both the child and parent both define an extension, but the child uses it implicitly while the parent uses it explicitly. In that case, one of two things can happen:

1) This change is superflous as it ends up looking into the parent anyway
2) it returns an empty Optional when before it would had returned the parents provider, therefore breaking backwards compatibility

IvanAndresFritzler:
--validate

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
IvanAndresFritzler: After a refactor in order to address the backward compatibility issue, this discussion no longer applies.

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/CompositeArtifactExtensionManager.java
IvanAndresFritzler: The code has been refactored and the CompositeArtifactExtensionManager is no longer modified.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
IvanAndresFritzler: Yes, it was breaking the domain use case. Code has been refactored in order to avoid this.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
```diff
@@ -134,6 +142,18 @@ protected void validateOperationConfiguration(ConfigurationProvider configuratio
     }
   }
 
+  private void validatePolicyIsolation(ConfigurationProvider configurationProvider) {
+    if (featureFlaggingService.isEnabled(MuleRuntimeFeature.ENABLE_POLICY_ISOLATION)
+        && muleContext.getArtifactType().equals(ArtifactType.POLICY)
```
elrodro83, line 147: Static imports

modules/spring-config/src/main/java/org/mule/runtime/config/internal/MuleArtifactContext.java
```diff
@@ -803,7 +803,14 @@ private static void configureSplitterExceptionHandlingFeature() {
     FeatureFlaggingRegistry ffRegistry = FeatureFlaggingRegistry.getInstance();
 
     ffRegistry.registerFeature(HANDLE_SPLITTER_EXCEPTION,
-                               ctx -> ctx.getConfiguration().getMinMuleVersion().map(v -> v.atLeast("4.4.0")).orElse(false));
+                               muleContext -> {
```
elrodro83, line 806: why?

modules/spring-config/src/main/java/org/mule/runtime/config/internal/model/properties/PropertiesResolverUtils.java
```diff
@@ -341,7 +341,14 @@ public static void configurePropertiesResolverFeatureFlag() {
       FeatureFlaggingRegistry ffRegistry = FeatureFlaggingRegistry.getInstance();
 
       ffRegistry.registerFeature(HONOUR_RESERVED_PROPERTIES,
-                                 ctx -> ctx.getConfiguration().getMinMuleVersion().map(v -> v.newerThan("4.2.2")).orElse(false));
+                                 muleContext -> {
```
elrodro83, line 344: why?

elrodro83 (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/ImplicitConfigurationProviderFactory.java
```diff
@@ -41,4 +41,17 @@ ConfigurationProvider createImplicitConfigurationProvider(ExtensionModel extensi
...
+  /**
+   * Returns an implicit {@link ConfigurationProvider} name, that can be used to look it up.
+   *
+   * @param extensionModel     The configurable {@link ExtensionModel}.
+   * @param configurationModel The {@link ConfigurationModel} that represents the extensionModel configuration.
+   * @param muleContext        The corresponding {@link MuleContext}.
+   * @return The {@link ConfigurationProvider} name.
+   * @see org.mule.runtime.module.extension.internal.manager.ExtensionRegistry#getConfigurationProvider(String)
+   */
+  String resolveImplicitConfigurationProviderName(ExtensionModel extensionModel, ConfigurationModel configurationModel,
```
elrodro83, line 54: add `@since`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultImplicitConfigurationProviderFactory.java
```diff
@@ -113,4 +119,47 @@ private boolean needsDynamicConnectionProvider(ExtensionModel extensionModel, Co
...
+   * {@link MuleRuntimeFeature#ENABLE_POLICY_ISOLATION} feature flag.
+   * 
+   * @param extensionModel     The configurable {@link ExtensionModel}.
+   * @param configurationModel The {@link ConfigurationModel} that represents the extensionModel configuration.
+   * @param muleContext        The corresponding {@link MuleContext}.
+   * @return The {@link ConfigurationProvider} name.
+   */
+  private String featureFlaggedPolicyExtensionsImplicitConfigurationProviderName(ExtensionModel extensionModel,
+                                                                                 ConfigurationModel configurationModel,
+                                                                                 MuleContext muleContext) {
```
elrodro83, line 147: have the mule context injected into this class instead of receiving it as parameter

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/DefaultExtensionManager.java
```diff
@@ -86,6 +86,11 @@
   public void initialise() throws InitialisationException {
     extensionRegistry = new ExtensionRegistry(new DefaultRegistry(muleContext));
     extensionActivator = new ExtensionActivator(muleContext);
+    try {
+      initialiseIfNeeded(implicitConfigurationProviderFactory, muleContext);
+    } catch (MuleException e) {
```
elrodro83, line 91: catch the more specific `InitialisationException`

elrodro83: and actaully, no need to catch since this method is already throwing that!

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultImplicitConfigurationProviderFactory.java
elrodro83: review all the chained calls where the mule context is being passed around

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java
```diff
@@ -134,6 +142,18 @@ protected void validateOperationConfiguration(ConfigurationProvider configuratio
     }
   }
 
+  private void validatePolicyIsolation(ConfigurationProvider configurationProvider) {
+    if (featureFlaggingService.isEnabled(ENABLE_POLICY_ISOLATION)
+        && muleContext.getArtifactType().equals(POLICY)
+        && configurationProvider.getExtensionModel() != extensionModel) {
+      throw new IllegalOperationException(format(
```
elrodro83, line 149: add a TODO to make this validation at the AST level, before actually initializing the app: https://www.mulesoft.org/jira/browse/MULE-19350

elrodro83 (changes requested):

IvanAndresFritzler:
--validate

elrodro83 (dismissed):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/ImplicitConfigurationProviderFactory.java
```diff
@@ -39,6 +37,17 @@ ConfigurationProvider createImplicitConfigurationProvider(ExtensionModel extensi
...
+  /**
+   * Returns an implicit {@link ConfigurationProvider} name, that can be used to look it up.
+   *
+   * @param extensionModel     The configurable {@link ExtensionModel}.
+   * @param configurationModel The {@link ConfigurationModel} that represents the extensionModel configuration.
+   * @return The {@link ConfigurationProvider} name.
+   * @see org.mule.runtime.module.extension.internal.manager.ExtensionRegistry#getConfigurationProvider(String)
+   * @since 4.4.0
+   */
+  String resolveImplicitConfigurationProviderName(ExtensionModel extensionModel, ConfigurationModel configurationModel);
```
marianogonzalez, line 51: why is this moved into this factory and exposed in this interface? it's weird that we have the need to expose this as an interface method. Sounds like an implementation detail that shouldn't be exposed by a factory class. This is a code smell. Double check if this responsibility doesn't belong somewhere else. I don't understand the reason why it should be moved here

marianogonzalez (changes requested):

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultApplicationPolicyInstance.java
```diff
@@ -137,6 +136,65 @@ private void initPolicyContext() throws InitialisationException {
...
+
+  /**
+   * HTTP and Sockets extension models must be added if found in the parent artifact extensions (backward compatibility for API
+   * Gateway).
+   * 
+   * @return Set containing the parent artifact HTTP and Sockets extension models (if present).
+   */
+  private Set<ExtensionModel> getInheritedExtensionModels() {
+    Set<ExtensionModel> inheritedExtensionModels = new HashSet<>(2);
+    application.getRegistry().<ExtensionManager>lookupByName(OBJECT_EXTENSION_MANAGER).get().getExtension("HTTP")
```
elrodro83, line 191: avoid looking up the extension manager twice in the registry

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/DefaultExtensionManager.java
```diff
@@ -76,6 +78,9 @@
   @Inject
   private ExpressionManager expressionManager;
 
+  @Inject
+  FeatureFlaggingService featureFlaggingService;
```
elrodro83, line 82: private

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultImplicitConfigurationProviderFactory.java
```diff
@@ -41,58 +43,63 @@
  */
 public final class DefaultImplicitConfigurationProviderFactory implements ImplicitConfigurationProviderFactory {
 
+  @Inject
+  MuleContext muleContext;
```
elrodro83, line 47: private

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultImplicitConfigurationProviderFactory.java
```diff
@@ -41,58 +43,63 @@
...
+  @Inject
+  FeatureFlaggingService featureFlaggingService;
```
elrodro83, line 49: private

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -125,6 +130,9 @@
  */
 public class MuleExtensionUtils {
 
+  private static final String IMPLICIT_CONFIGURATION_SUFFIX = "-implicit";
+  public static final String HYPHEN = "-";
```
elrodro83, line 134: isnt there a definition for this somewhere else already?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -125,6 +130,9 @@
  */
 public class MuleExtensionUtils {
 
+  private static final String IMPLICIT_CONFIGURATION_SUFFIX = "-implicit";
```
elrodro83, line 133: `HYPHEN + "implicit"`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -481,9 +488,68 @@ public static ExtensionModel loadExtension(Class<?> clazz, Map<String, Object> p
...
+   *
+   * @param extensionModel     The configurable {@link ExtensionModel}.
+   * @param configurationModel The {@link ConfigurationModel} that represents the extension configuration.
+   * @return The {@link org.mule.runtime.extension.api.runtime.config.ConfigurationProvider} name.
+   */
+  private static String getFeatureFlaggedPolicyExtensionsImplicitConfigurationProviderName(ExtensionModel extensionModel,
+                                                                                           ConfigurationModel configurationModel,
+                                                                                           String extendedArtifactId,
+                                                                                           FeatureFlaggingService featureFlaggingService) {
+    if (featureFlaggingService != null && featureFlaggingService.isEnabled(ENABLE_POLICY_ISOLATION)) {
```
elrodro83, line 525: can `featureFlaggingService ` be null?

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/manager/DefaultExtensionManagerTestCase.java
```diff
@@ -151,7 +152,7 @@
   @Before
   public void before() throws MuleException {
     muleContext = mockContextWithServices();
-
+    when(muleContext.getArtifactType()).thenReturn(ArtifactType.APP);
```
elrodro83, line 155: static import

elrodro83 (changes requested):

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultApplicationPolicyInstance.java
```diff
@@ -137,6 +136,66 @@ private void initPolicyContext() throws InitialisationException {
     }
   }
 
+  /**
+   * Applies the {@link MuleRuntimeFeature#ENABLE_POLICY_ISOLATION} feature to the policy artifact plugins list.
+   *
+   * @return The policy artifact plugins.
+   */
+  private List<ArtifactPlugin> getFeatureFlaggedArtifactPlugins(ArtifactDescriptor policyArtifactDescriptor) {
+    if (FeatureFlaggingUtils.isFeatureEnabled(ENABLE_POLICY_ISOLATION, policyArtifactDescriptor)) {
```
elrodro83, line 145: static import

elrodro83 (approved):

marianogonzalez (dismissed):

marianogonzalez (approved):

elrodro83 (approved):

PR #10151, opened 2021-02-25
CMTS-180: Use parameterValueResolver with default values in order to …

gsfernandes:
…support type convertion


svacas (approved):

gsfernandes:
--validate

PR #10159, opened 2021-02-26
MULE-18725: fix report error with soap exception

jcgrethe:


modules/ws/src/main/java/org/mule/module/ws/consumer/WSConsumer.java
```diff
@@ -194,7 +196,7 @@ protected MuleEvent processNext(MuleEvent event) throws MuleException
 
                         event.getMessage().setPayload(soapFault.getDetail());
 
-                        throw new SoapFaultException(event, soapFault, this);
+                        throw new SoapFaultException(event, soapFault, callerMessageProcessor);
```
elrodro83, line 199: can you use `WSConsumer.this` to avid putting an attribute in this inner class?

elrodro83 (changes requested):

elrodro83 (approved):

PR #10167, opened 2021-03-02
MULE-19185: Avoid multiple construction of CompositeClassloader in operation calls

eze210:


eze210:
It was a test to generate a distribution for performance, the actual PR is #10094

PR #10187, opened 2021-03-15
MULE-19262: Operation content InputStream is not reset after the access token is refreshed

ndinu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/streaming/CursorResetHandler.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * @since 4.3.1 4.4.0
+ */
+public class CursorResetHandler {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(CursorResetHandler.class);
+
+  private final List<Pair<Cursor, Long>> cursorPositions;
+
+  public CursorResetHandler(List<Pair<Cursor, Long>> cursorPositions) {
```
marianogonzalez, line 29: this class should just receive the cursor. Clients are not responsible for building the data structures that it needs under hood. The list of pairs should be built here

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/streaming/CursorResetHandler.java
ndinu: Agree, I have just changed this.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/streaming/CursorResetHandler.java
```diff
@@ -0,0 +1,58 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(CursorResetHandler.class);
+
+  private final List<Pair<Cursor, Long>> cursorPositions;
+
+  public CursorResetHandler(List<Cursor> cursors) {
+    this.cursorPositions = getCursorPositions(cursors);
+  }
+
+  private List<Pair<Cursor, Long>> getCursorPositions(List<Cursor> cursors) {
+    List<Pair<Cursor, Long>> cursorPositions = new LinkedList<>();
```
marianogonzalez, line 35: `use new ArrayList<>(cursors.size())` instead

marianogonzalez (changes requested):

marianogonzalez (approved):

PR #10190, opened 2021-03-16
MULE-19264: configured parameters are not injected at all

pabloperalta:


tests/test-extensions/petstore-extension/src/main/java/org/mule/test/petstore/extension/PetStoreConnector.java
```diff
@@ -64,6 +64,9 @@
   @Optional
   private Aquarium aquarium;
 
+  @ParameterGroup(name = "Advanced Leash Configuration", showInDsl = true)
+  public AdvancedLeashConfiguration advancedLeashConfiguration;
```
elrodro83, line 68: add a comment to make it clear that the odd thing about this are the spaces in the name

elrodro83 (changes requested):

marianogonzalez (changes requested):
what's the purpose of this PR? This changes no code and makes no assertion over the added parameter group other than standard serialization

tests/test-extensions/petstore-extension/src/main/java/org/mule/test/petstore/extension/PetStoreConnector.java
pabloperalta: Done

pabloperalta:
This PR is about a bug in the AST that resulted in configured parameters being ignored. This happened because in the case of a ParameterGroup name that was not camel case the values were not being set. I have added a test in mule-integration-tests: https://github.com/mulesoft/mule-integration-tests/pull/1401/files#diff-e7d47bf91d8b68d83baa41f64c27281fbc479ec7721a51de3544adb67d336fefR709

elrodro83 (approved):

PR #10203, opened 2021-03-22
MULE-19255: Support new SDK API Value Providers

Bardammu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/ValueProvidersParameterDeclarationEnricher.java
```diff
@@ -278,4 +286,40 @@ private void enrichParameter(OfValues resolverClass,
     }
     return parameterNames;
   }
+
+  private Optional<?> getOfValueAnnotation(ParameterDeclaration parameterDeclaration) {
```
marianogonzalez, line 290: this goes against the pattern that we have defined and used across all other FC features. We want one single code path capable of handling both cases. If an SDK-API implementation is found, then use it. If a legacy implementation is used, then use an adapter wrapper so that it can be treated as as if it were an sdk-api one. `LegacyResultAdapter` and `LegacyPollingSourceAdapter` are examples of this pattern. 

Using wildcard and `Either` classes are not a scalable pattern

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/EitherOfValue.java
```diff
@@ -0,0 +1,54 @@
...
+ * Encapsulates either a {@link OfValues} annotation or a {@link org.mule.runtime.extension.api.annotation.values.OfValues}
+ * annotation in order to be able to use a single object to represent both.
+ * <p>
+ * If the {@link EitherOfValue} is used with an annotation other than {@link OfValues} or a
+ * {@link org.mule.runtime.extension.api.annotation.values.OfValues}, an {@link IllegalModelDefinitionException} exception is
+ * thrown.
+ *
+ * @since 4.4.0
+ */
+public class EitherOfValue {
```
marianogonzalez, line 24: Replace with an adapter class

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/SdkValueAdapter.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Adapter of the legacy {@link Value} to {@link org.mule.sdk.api.values.Value}
+ *
+ * @since 4.4.0
+ */
+public class SdkValueAdapter implements Value {
```
marianogonzalez, line 19: This should be the other way around. Adapt the extension-api verion of it into an sdk-api version as any new features will be added on the `sdk-api` one. This class should actually be a  `LegacyValueAdapter`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/SdkValueProviderAdapter.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Adapter of the legacy {@link ValueProvider} to {@link org.mule.sdk.api.values.ValueProvider}
+ *
+ * @since 4.4.0
+ */
+public class SdkValueProviderAdapter implements ValueProvider {
```
marianogonzalez, line 21: Same as before. Invert the adapter direction

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/ValueProvidersParameterDeclarationEnricher.java
Bardammu: How can we use an adapter with an annotation o or a class?
We are using Either right now by the way: https://github.com/mulesoft/mule/commit/8eceaac2b1422e2a1c29b1897a16eaba28bc38cc

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/EitherOfValue.java
Bardammu: OfValue is an annotation, how can I use an adapter here?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/ValueProvidersParameterDeclarationEnricher.java
```diff
@@ -278,4 +282,44 @@ private void enrichParameter(OfValues resolverClass,
...
+    Optional<OfValues> legacyAnnotation = getAnnotation(parameterDeclaration, OfValues.class);
+    Optional<org.mule.sdk.api.annotation.values.OfValues> sdkAnnotation =
+        getAnnotation(parameterDeclaration, org.mule.sdk.api.annotation.values.OfValues.class);
+
+    if (legacyAnnotation.isPresent() && sdkAnnotation.isPresent()) {
+      throw new IllegalModelDefinitionException(format("Parameter %s cannot be annotated with both %s and %s",
+                                                       parameterDeclaration.getName(),
+                                                       OfValues.class.getName(),
+                                                       org.mule.sdk.api.annotation.values.OfValues.class.getName()));
+    } else if (legacyAnnotation.isPresent()) {
```
marianogonzalez, line 296: `IllegalModelDefinitionException` is both present at the same time

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/ValueProvidersParameterDeclarationEnricher.java
```diff
@@ -278,4 +282,44 @@ private void enrichParameter(OfValues resolverClass,
...
+    } else if (sdkAnnotation.isPresent()) {
+      return of(new OfValueInformation(sdkAnnotation.get().value(), sdkAnnotation.get().open()));
+    } else {
+      return empty();
+    }
+  }
+
+  private Optional<OfValueInformation> getOfValueInformation(ParameterGroupDeclaration parameterGroupDeclaration,
+                                                             AnnotatedElement annotatedElement) {
+    OfValues legacyAnnotation = annotatedElement.getAnnotation(OfValues.class);
```
marianogonzalez, line 307: `IllegalModelDefinitionException` is both present at the same time

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/LegacyValueAdapter.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @Override
+  public String getDisplayName() {
+    return value.getDisplayName();
+  }
+
+  @Override
+  public Set<Value> getChilds() {
+    Set<Value> values = new HashSet<>();
```
marianogonzalez, line 40: Use a `LinkedHashSet` to respect the delegate's iteration order

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/LegacyValueAdapter.java
```diff
@@ -0,0 +1,49 @@
...
+    value.getChilds().forEach(v -> values.add(new LegacyValueAdapter(v)));
+    return values;
```
marianogonzalez, line 42: The delegate returns  an `unmodifiableSet` and so should this.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/LegacyValueAdapter.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @Override
+  public String getDisplayName() {
+    return value.getDisplayName();
+  }
+
+  @Override
+  public Set<Value> getChilds() {
+    Set<Value> values = new HashSet<>();
+    value.getChilds().forEach(v -> values.add(new LegacyValueAdapter(v)));
```
marianogonzalez, line 41: not a big deal since this is design time exclusive, but this would be more efficient as value.getChilds().stream().map(LegacyValueAdapter::new).collect(LinkedHashSet::new)

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/LegacyValueProviderAdapter.java
```diff
@@ -0,0 +1,44 @@
...
+  private final org.mule.runtime.extension.api.values.ValueProvider valueProvider;
+
+  public LegacyValueProviderAdapter(org.mule.runtime.extension.api.values.ValueProvider valueProvider) {
+    this.valueProvider = valueProvider;
+  }
+
+  @Override
+  public Set<Value> resolve() throws ValueResolvingException {
+    try {
+      Set<Value> values = new HashSet<>();
```
marianogonzalez, line 32: same comments as with the Value adapter

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/ConfigurationProviderToolingAdapter.java
```diff
@@ -210,7 +198,7 @@ private MetadataContext getMetadataContext(ClassLoader classLoader) {
   private interface WithConnectionProviderCallable<T> {
 
     T call(ConnectionProvider connectionProvider, ConnectionProviderModel connectionProviderModel)
-        throws ValueResolvingException;
+        throws org.mule.sdk.api.values.ValueResolvingException;
```
ndinu, line 201: I think that you do not need the fqdn here, just ValueResolvingException, since the import has changed.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/ValueProviderFactory.java
```diff
@@ -125,4 +127,19 @@ private void injectValueProviderFields(ValueProvider resolver) throws ValueResol
...
+
+  private ValueProvider adaptResolver(Object resolverObject) throws ValueResolvingException {
+    if (resolverObject instanceof ValueProvider) {
+      return (ValueProvider) resolverObject;
+    } else if (resolverObject instanceof org.mule.runtime.extension.api.values.ValueProvider) {
+      return new SdkValueProviderAdapter((org.mule.runtime.extension.api.values.ValueProvider) resolverObject);
+    } else {
+      throw new ValueResolvingException(format("An error occurred trying to create a ValueProvider: %s should implement %s or %s",
+                                               resolverObject.getClass().getName(),
+                                               ValueProvider.class.getSimpleName(),
```
ndinu, line 139: If you use the .getSimpleName() I think you will get "ValueProvider" on both calls.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/ValueProviderMediator.java
```diff
@@ -110,9 +108,9 @@ public ValueProviderMediator(T containerModel, Supplier<MuleContext> muleContext
    * @return a {@link Set} of {@link Value} correspondent to the given parameter
    * @throws ValueResolvingException if an error occurs resolving {@link Value values}
    */
-  public Set<Value> getValues(String parameterName, ParameterValueResolver parameterValueResolver,
-                              Supplier<Object> connectionSupplier, Supplier<Object> configurationSupplier,
-                              ConnectionProvider connectionProvider)
+  public Set<org.mule.sdk.api.values.Value> getValues(String parameterName, ParameterValueResolver parameterValueResolver,
```
ndinu, line 111: Would this bring problems for tooling? I think they use this class directly. @gsfernandes

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/ValueProviderUtils.java
```diff
@@ -118,17 +141,22 @@ static ValueBuilder cloneAndEnrichValue(Value value, Map<Integer, String> partOr
...
     } catch (Exception e) {
       throw new IllegalStateException(format("There was an error creating an instance of %s to retrieve the Id of the provider",
                                              valueProviderClazz.getName()),
                                       e);
     }
-    return valueProvider.getId();
+
+    return valueProviderObject instanceof org.mule.runtime.extension.api.values.ValueProvider
+        ? ((org.mule.runtime.extension.api.values.ValueProvider) valueProviderObject).getId()
+        : ((ValueProvider) valueProviderObject).getId();
```
ndinu, line 159: I think we should also check that it is a class that implements ValueProvider instead of assuming it because it is not an sdk VP.

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/value/ValueProviderExecutor.java
```diff
@@ -78,21 +83,17 @@ public ValueResult resolveValues(ParameterizedModel parameterizedModel,
...
+                                                                 parameterValueResolver, context),
+                                                 org.mule.sdk.api.values.ValueResolvingException.class,
                                                  e -> {
                                                    throw new ExecutorExceptionWrapper(e);
                                                  }));
       } finally {
         context.dispose();
       }
-    } catch (ValueResolvingException e) {
+    } catch (org.mule.sdk.api.values.ValueResolvingException e) {
```
ndinu, line 96: Is there any reference to the old ValueResolvingException? If not, change the import to the new one.

core/src/main/java/org/mule/runtime/core/internal/value/MuleValueProviderService.java
```diff
@@ -90,15 +99,48 @@ private ValueProvider findValueProvider(Location location, String providerName)
 
     Object component = findComponent(realLocation);
 
-    if (component instanceof ComponentValueProvider) {
-      return () -> ((ComponentValueProvider) component).getValues(providerName);
+    if (component instanceof SdkComponentValueProvider) {
+      return new ValueProvider() {
```
marianogonzalez, line 103: Wrong adapter direction and wrong adaptation place. There must be one single and unique place at which the value providers are adapted, and that's where they're instantiated. Cannot do adaptation at the point they're used. Also, the old needs to be adapted into the new. Never adapt the new into the old

core/src/main/java/org/mule/runtime/core/internal/value/MuleValueProviderService.java
```diff
@@ -90,15 +99,48 @@ private ValueProvider findValueProvider(Location location, String providerName)
...
+
+        @Override
+        public String getId() {
+          return null;
+        }
+      };
     }
 
-    if (component instanceof ConfigurationParameterValueProvider) {
+    if (component instanceof SdkConfigurationParameterValueProvider) {
```
marianogonzalez, line 117: same as above

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/ValueProvidersParameterDeclarationEnricher.java
```diff
@@ -278,4 +282,44 @@ private void enrichParameter(OfValues resolverClass,
...
     return parameterNames;
   }
+
+  private Optional<OfValueInformation> getOfValueInformation(ParameterDeclaration parameterDeclaration) {
+    Optional<OfValues> legacyAnnotation = getAnnotation(parameterDeclaration, OfValues.class);
+    Optional<org.mule.sdk.api.annotation.values.OfValues> sdkAnnotation =
+        getAnnotation(parameterDeclaration, org.mule.sdk.api.annotation.values.OfValues.class);
+
+    if (legacyAnnotation.isPresent() && sdkAnnotation.isPresent()) {
+      throw new IllegalModelDefinitionException(format("Annotations %s and %s are both present at the same time on parameter %s",
```
marianogonzalez, line 292: also specifiy the component in the error message

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/ValueProviderFactory.java
```diff
@@ -125,4 +127,19 @@ private void injectValueProviderFields(ValueProvider resolver) throws ValueResol
           + missingParameters, MISSING_REQUIRED_PARAMETERS);
     }
   }
+
+  private ValueProvider adaptResolver(Object resolverObject) throws ValueResolvingException {
+    if (resolverObject instanceof ValueProvider) {
+      return (ValueProvider) resolverObject;
+    } else if (resolverObject instanceof org.mule.runtime.extension.api.values.ValueProvider) {
```
marianogonzalez, line 134: wrong adapter direction

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/ValueProviderMediator.java
marianogonzalez: yep. This ties back into the prior comment of deprecation/replacement being incomplete. This breaks backwards compatibility

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
```diff
@@ -111,7 +108,7 @@
  * @since 4.0
  */
 public abstract class ExtensionComponent<T extends ComponentModel> extends AbstractComponent
-    implements MuleContextAware, MetadataKeyProvider, MetadataProvider<T>, ComponentValueProvider,
+    implements MuleContextAware, MetadataKeyProvider, MetadataProvider<T>, SdkComponentValueProvider,
```
marianogonzalez, line 111: This breaks backwards compatibility. What happens with current clients which are casting to `ComponentValueProvider`?  Where's the adaptation layer for that?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/ConfigurationProviderToolingAdapter.java
```diff
@@ -63,14 +62,14 @@
...
- * <li>{@link ConfigurationParameterValueProvider}, to resolve {@link Value values} associated to a configuration and their
+ * <li>{@link SdkConfigurationParameterValueProvider}, to resolve {@link Value values} associated to a configuration and their
  * related connection</li>
  * </ul>
  *
  * @since 4.0
  */
 public final class ConfigurationProviderToolingAdapter extends StaticConfigurationProvider
-    implements MetadataKeyProvider, ConfigurationParameterValueProvider {
+    implements MetadataKeyProvider, SdkConfigurationParameterValueProvider {
```
marianogonzalez, line 72: same as before. Incompatible change

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DynamicConfigurationProvider.java
```diff
@@ -73,7 +73,7 @@
  * @since 4.0.0
  */
 public final class DynamicConfigurationProvider extends LifecycleAwareConfigurationProvider
-    implements ExpirableConfigurationProvider, ConfigurationParameterValueProvider {
+    implements ExpirableConfigurationProvider, SdkConfigurationParameterValueProvider {
```
marianogonzalez, line 76: incompatible

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/ValueProviderUtils.java
```diff
@@ -39,6 +40,18 @@ private ValueProviderUtils() {
 
   }
 
+  static org.mule.runtime.extension.api.values.ValueBuilder cloneAndEnrichValue(org.mule.runtime.api.value.Value value,
```
marianogonzalez, line 43: why is this needed? Doesn't the need for this disappears if adaptation is done in a single place and in the correct direction?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/ValueProviderUtils.java
marianogonzalez: The entire code needs to be writteng in terms of the SDK API. We shouldn't have these type of checks in any where else other than the ValuePRoviderFactory.

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/value/ValueProviderExecutor.java
```diff
@@ -180,4 +181,18 @@ private ValueProviderMediator createValueProviderMediator(ParameterizedModel par
     return empty();
   }
 
+  private Set<Value> getValues(String providerName,
+                               ValueProviderMediator<?> valueProviderMediator,
+                               ParameterValueResolver parameterValueResolver,
+                               ExtensionResolvingContext context)
+      throws org.mule.sdk.api.values.ValueResolvingException {
```
marianogonzalez, line 188: use qualified import. Also, remember that the legacy Exception extends this one, so ou can use them interchangealy.

marianogonzalez (changes requested):

core/src/main/java/org/mule/runtime/core/internal/value/MuleValueProviderService.java
Bardammu: So, should mule uses the SDK value or the SDK return the legacy Value?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/ValueProviderUtils.java
Bardammu: Tooling client uses this utils

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/ValueProviderFactory.java
Bardammu: This is the opposite of what I did in the previous commits. where I was adapting the other way and we agree that it was wrong.

marianogonzalez: yes, sorry, got confused by the classnames. `SdkValueProviderAdapter` should be called `LegacyValueProviderAdapter` ?

core/src/main/java/org/mule/runtime/core/internal/value/MuleValueProviderService.java
marianogonzalez: not following

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/ValueProviderUtils.java
marianogonzalez: They shouldn't. Please create a ticket for us to provide an API for whatever internal functionality they are using and add a TODO here.

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
```diff
@@ -423,29 +423,33 @@ public void setMuleContext(MuleContext context) {
...
     } catch (MuleRuntimeException e) {
       Throwable rootException = getRootException(e);
       if (rootException instanceof ValueResolvingException) {
-        throw (ValueResolvingException) rootException;
+        throw (org.mule.runtime.extension.api.values.ValueResolvingException) rootException;
       } else {
-        throw new ValueResolvingException("An unknown error occurred trying to resolve values. " + e.getCause().getMessage(),
-                                          UNKNOWN, e);
+        throw new org.mule.runtime.extension.api.values.ValueResolvingException("An unknown error occurred trying to resolve values. "
+            + e.getCause().getMessage(),
```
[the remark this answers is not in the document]
Bardammu: No, it's checking that there is a known ValueResolvingException as a root cause and throwing UNKNOWN otherwise.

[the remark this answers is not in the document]
marianogonzalez: but if the root exception is a `org.mule.sdk.api.values.ValueResolvingException` doesn't it get wrapped as an UNKNOWN legacy exception and the true error code gets lost?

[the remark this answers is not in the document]
marianogonzalez: Additionally, this shouldn't be throwing a `org.mule.runtime.extension.api.values.ValueResolvingException` but a `org.mule.sdk.api.values.ValueResolvingException`. We can't however make that change as tooling is using non API components and then catching for that specific exception. Add a TODO here also pointing to the jira you created explaining that this change needs to happen

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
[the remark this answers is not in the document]
Bardammu: `getValues` from  `ValueProviderMediator`  throws the legacy exception I can't change that since it's also used by tooling, so it's not possible to get  `org.mule.sdk.api.values.ValueResolvingException` here. The `ValueProviderMediator` catches the SDK exception and throws the legacy exception.

Regarding the second comment besides tooling using the method, this is defined on the ComponentValueProvider API so we should actually deprecate the method and create a new one.

marianogonzalez (approved):

PR #10217, opened 2021-03-25
MULE-19127: Fixing condition check to prevent previously stopped flows from restarting

sofiamorseletto:


modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/FlowStoppedDeploymentPersistenceListenerTestCase.java
```diff
@@ -94,18 +94,14 @@ public void onStartShouldSaveDeploymentPropertyAsTrue() throws Exception {
   }
 
   @Test
-  @Description("checkIfFlowShouldStart method should check deployment properties")
-  public void checkIfFlowShouldStartMethodMustSetShouldStartPropertyAsFalseWhenDeploymentPropertyIsFalse() throws Exception {
+  @Description("shouldStart method should check deployment properties")
```
elrodro83, line 97: add missing `@Issue` annotation

elrodro83 (changes requested):

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/FlowStoppedDeploymentPersistenceListenerTestCase.java
sofiamorseletto: The issue annotation is on the class [name](https://github.com/mulesoft/mule/blob/28081e8bd24d3e48331f7eb9dc6d2a16421bd152/modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/FlowStoppedDeploymentPersistenceListenerTestCase.java#L36), since this PR is a fix on the original issue the issue is still the same. Should I still add the Issue on each method anyways?

elrodro83: no, it's ok then

elrodro83 (approved):

PR #10220, opened 2021-03-26
MULE-19289: Original exception lost on connectivity error in TransactionBindingDelegate#bindResource

elrodro83:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/transaction/TransactionBindingDelegate.java
```diff
@@ -120,7 +120,11 @@ public void invalidate() {
       }
     } finally {
       if (!bound) {
-        connectionHandler.release();
+        try {
+          connectionHandler.release();
+        } catch (Exception e) {
+          // Nothing to do
```
fsgonz, line 126: I would log a warning in an general exception is enable easier troubleshooting in case something results in a leak.

fsgonz (approved):

PR #10228, opened 2021-03-30
MULE-19290: Pooling profile initialization policy is not being honored.

SebaElizalde:


core-tests/src/test/java/org/mule/runtime/core/internal/connection/PoolingConnectionManagementStrategyTestCase.java
```diff
@@ -142,10 +155,60 @@ public void failDueToInvalidConnection() throws ConnectionException {
...
+    initStrategy();
+    verify(connectionProvider, times(0)).connect();
+    verify(connectionProvider, times(0)).disconnect(any());
+  }
+
+  @Test
+  public void initializationPolicyIsLimitedByMaxIdle() throws ConnectionException {
+    poolingProfile = new PoolingProfile(5, 3, DEFAULT_MAX_POOL_WAIT, DEFAULT_POOL_EXHAUSTED_ACTION, INITIALISE_ALL);
+    initStrategy();
+    verify(connectionProvider, times(5)).connect();
```
marianogonzalez, line 195: these two assertions are repeated all over the place. Extract to a method which receives two ints

core-tests/src/test/java/org/mule/runtime/core/internal/connection/PoolingConnectionManagementStrategyTestCase.java
```diff
@@ -96,11 +98,14 @@ public void poolingListenerFailsOnBorrow() throws Exception {
       assertThat(e.getCause(), is(sameInstance(exception)));
       verify(connectionProvider).disconnect(any(Lifecycle.class));
     }
-
   }
 
   @Test
   public void connectionsDependenciesInjected() throws Exception {
+    initStrategy();
```
marianogonzalez, line 105: why adding this here? This already happens on the before()

core-tests/src/test/java/org/mule/runtime/core/internal/connection/PoolingConnectionManagementStrategyTestCase.java
```diff
@@ -124,6 +129,10 @@ public void exhaustion() throws Exception {
 
   @Test
   public void release() throws Exception {
+    initStrategy();
```
marianogonzalez, line 132: why adding this here? This already happens on the before()

core-tests/src/test/java/org/mule/runtime/core/internal/connection/PoolingConnectionManagementStrategyTestCase.java
```diff
@@ -134,6 +143,10 @@ public void release() throws Exception {
 
   @Test(expected = ConnectionException.class)
   public void failDueToInvalidConnection() throws ConnectionException {
+    initStrategy();
```
marianogonzalez, line 146: why adding this here? This already happens on the before()

core-tests/src/test/java/org/mule/runtime/core/internal/connection/PoolingConnectionManagementStrategyTestCase.java
```diff
@@ -142,10 +155,60 @@ public void failDueToInvalidConnection() throws ConnectionException {
 
   @Test(expected = ConnectionException.class)
   public void failDueToNullConnectionValidationResult() throws ConnectionException {
+    initStrategy();
```
marianogonzalez, line 158: why adding this here? This already happens on the before()

core/src/main/java/org/mule/runtime/core/internal/connection/PoolingConnectionManagementStrategy.java
```diff
@@ -106,9 +112,31 @@ public void close() throws MuleException {
...
   }
 
+  protected void applyInitialisationPolicy(GenericObjectPool pool) {
+    int initialConnections = 0;
+    switch (poolingProfile.getInitialisationPolicy()) {
+      case INITIALISE_ONE:
+        initialConnections = 1;
+        break;
+      case INITIALISE_ALL:
+        initialConnections = poolingProfile.getMaxActive();
```
marianogonzalez, line 127: this should actually be min(getMaxIdle(), getMaxActive())

marianogonzalez (changes requested):
fsdfg

core/src/main/java/org/mule/runtime/core/internal/connection/PoolingConnectionManagementStrategy.java
```diff
@@ -106,9 +112,31 @@ public void close() throws MuleException {
     config.timeBetweenEvictionRunsMillis = poolingProfile.getEvictionCheckIntervalMillis();
     GenericObjectPool genericPool = new GenericObjectPool(new ObjectFactoryAdapter(), config);
 
+    applyInitialisationPolicy(genericPool);
+
     return genericPool;
   }
 
+  protected void applyInitialisationPolicy(GenericObjectPool pool) {
+    int initialConnections = 0;
```
ndinu, line 121: I would make the "=0" explicit in the switch for the case INITIALIZE_NONE.

core-tests/src/test/java/org/mule/runtime/core/internal/connection/PoolingConnectionManagementStrategyTestCase.java
```diff
@@ -142,10 +155,60 @@ public void failDueToInvalidConnection() throws ConnectionException {
...
+    connection1 = strategy.getConnectionHandler();
+    connection2 = strategy.getConnectionHandler();
+
     when(connectionProvider.validate(anyVararg())).thenReturn(null);
     strategy.getConnectionHandler().getConnection();
   }
 
+  @Test
+  public void defaultInitializationPolicyIsHonored() throws ConnectionException {
```
ndinu, line 167: I would also assert how many elements are idle in the pool for all these new tests.

core-tests/src/test/java/org/mule/runtime/core/internal/connection/PoolingConnectionManagementStrategyTestCase.java
SebaElizalde: removed it from the before() because some tests need to configure additional things before initializing the strategy

core-tests/src/test/java/org/mule/runtime/core/internal/connection/PoolingConnectionManagementStrategyTestCase.java
ndinu: You are right, dismiss this comment 😄

core-tests/src/test/java/org/mule/runtime/core/internal/connection/PoolingConnectionManagementStrategyTestCase.java
marianogonzalez: if only a couple of tests need a different setting then leave it on before and let that handful of test redefine only when needed

SebaElizalde: It is more than just a couple of tests. It is shorter this way and it also avoids reseting and re-initializing objects in the 6 new tests. Plus I think is is better for the @before method to just contain code that is common to ALL tests.

marianogonzalez (approved):

core/src/main/java/org/mule/runtime/core/internal/connection/PoolingConnectionManagementStrategy.java
```diff
@@ -106,9 +114,43 @@ public void close() throws MuleException {
...
+  protected void applyInitialisationPolicy(GenericObjectPool pool) {
+    int initialConnections;
+    switch (poolingProfile.getInitialisationPolicy()) {
+      case INITIALISE_NONE:
+        initialConnections = 0;
+        break;
+      case INITIALISE_ONE:
+        initialConnections = 1;
+        break;
+      case INITIALISE_ALL:
```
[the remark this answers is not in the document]
ndinu: 👍

PR #10232, opened 2021-03-31
MULE-19295: Add getRootId to EventContext

mbuchwald:


core/src/main/java/org/mule/runtime/core/internal/event/DefaultEventContext.java
```diff
@@ -345,6 +350,11 @@ public String getId() {
       return id;
     }
 
+    @Override
+    public String getRootId() {
+      return this.parent.getRootId();
```
eze210, line 355: you have access to the root ctx from here (this.root) to avoid this second recursion

eze210 (changes requested):

eze210 (approved):

PR #10240, opened 2021-03-31
MULE-19220: Request tracking fields should be set in the same thread as the operation execution

eze210:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -783,8 +786,39 @@ private void prepareAndExecuteOperation(CoreEvent event, Supplier<ExecutorCallba
...
+    try {
+      executeOperation(operationContext, mapped(callbackSupplier.get(), operationContext,
+                                                isTargetWithPolicies(event) ? valueReturnDelegate : returnDelegate));
+    } finally {
+      unsetCurrentLocation(wasProcessorPathSet);
+    }
+  }
+
+  private boolean setCurrentLocation() {
+    ComponentLocation componentLocation = getLocation();
```
marianogonzalez, line 800: don't do this on every execution. Cannot you do this on the `initialise()` and cache the result?

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
eze210: No, because this method sets the location in the MDC, which is thread local. We need to set it in the thread that is going to execute the operation, and we don't know which is that thread on `initialise()`.
Probably the method should be renamed to "addComponentLocationToMDC()"

marianogonzalez: That's not what I meant. The MDC you need to set everytime, sure. But the resolved location is fixed. What I meant is that you shouldn't be calculating the location on every single request. Makes sense?

core/src/main/java/org/mule/runtime/core/privileged/processor/chain/AbstractMessageProcessorChain.java
```diff
@@ -341,38 +340,20 @@ private ReactiveProcessor applyInterceptors(List<ReactiveInterceptor> intercepto
     interceptors.addAll(additionalInterceptors);
 
     // #4 Wrap execution, before processing strategy, on flow thread.
-    interceptors.add((processor, next) -> {
-      String processorPath;
-      if (processor instanceof Component && ((Component) processor).getLocation() != null) {
-        processorPath = ((Component) processor).getLocation().getLocation();
```
marianogonzalez, line 347: by moving this logic out of this interceptor chain into the `ComponentMessageProcessor` class you're making this logic only available for SDK based component. What about hand crafted components like the `mule` and `ee` base language elements?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -783,8 +788,41 @@ private void prepareAndExecuteOperation(CoreEvent event, Supplier<ExecutorCallba
...
+      unsetCurrentLocation(wasProcessorPathSet);
+    }
+  }
+
+  private boolean setCurrentLocation() {
+    if (MDC.get(PROCESSOR_PATH_MDC_KEY) != null) {
+      return false;
+    }
+
+    if (processorPath == null) {
```
marianogonzalez, line 806: you have a potential race condition here with the first couple of events as processorPath needs to be set for the first time. I think that the processorPath field should be initialised on the `initialise()` phase.

marianogonzalez (changes requested):

core/src/main/java/org/mule/runtime/core/privileged/processor/chain/AbstractMessageProcessorChain.java
eze210: reverted this part

marianogonzalez (approved):

PR #10243, opened 2021-04-05
MULE-19205: Migrate validation in ScatterGatherRouter#initialize

elrodro83:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/CoreComponentBuildingDefinitionProvider.java
```diff
@@ -186,7 +187,6 @@
   private static final String FLOW = "flow";
   private static final String FLOW_REF = "flow-ref";
   private static final String EXCEPTION_LISTENER_ATTRIBUTE = "exceptionListener";
-  private static final String SCATTER_GATHER = "scatter-gather";
   private static final String PARALLEL_FOREACH = "parallel-foreach";
```
fsgonz, line 190: The parallel foreach can also take the one from CoreDslConstants

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/ScatterGatherRoutes.java
```diff
@@ -0,0 +1,66 @@
...
+import org.mule.runtime.ast.api.validation.ValidationResultItem;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Predicate;
+
+@IgnoreOnLazyInit
+public class ScatterGatherRoutes implements Validation {
+
+  private static final String SCATTER_GATHER_ELEMENT = "scatter-gather";
```
fsgonz, line 30: use the one in CoreDslConstants if possible.

fsgonz (changes requested):

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/ScatterGatherRoutes.java
elrodro83: CoreDslConstants is in spring-config, which is not a dependency of this module

fsgonz: Ok!

fsgonz (approved):

PR #10246, opened 2021-04-06
MULE-19103: introduce ConnectivitySchema modeling and serialization

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/connectivity/platform/schema/ConnectivitySchema.java
```diff
@@ -0,0 +1,351 @@
...
+
+    /**
+     * Sets the schema's GAV
+     *
+     * @param groupId    a valid Maven {@code groupId}
+     * @param artifactId a valid Maven {@code artifactId}
+     * @param version    a valid Maven {@code version}
+     * @return {@code this} builder
+     */
+    public Builder setGav(String groupId, String artifactId, String version) {
```
ndinu, line 50: I would name the builder methods without the `set`.

setGav -> gav
addParameter -> withParameter

ndinu (approved):

PR #10262, opened 2021-04-13
 MULE-19103: add the concept of semantic terms into the ExtensionModel

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/semantic/SemanticTermsEnricher.java
```diff
@@ -0,0 +1,96 @@
...
+
+      @Override
+      protected void onSource(SourceDeclaration declaration) {
+        extractType(declaration).ifPresent(type -> addSemanticTerms(declaration, type));
+      }
+
+      @Override
+      protected void onParameter(ParameterGroupDeclaration parameterGroup, ParameterDeclaration parameter) {
+        extractDeclaredParameter(parameter).ifPresent(e -> addSemanticTerms(parameter, e));
+        Set<String> typeTerms = getSemanticTerms(parameter.getType());
```
ndinu, line 69: The type annotation has an immutable set, I think that is good because that way you cannot mutate the annotation and therefore the metadataType that is annotated.

Is there something that I am missing, or will the add terms fail because it is immutable?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/semantic/SemanticTermsEnricher.java
```diff
@@ -0,0 +1,96 @@
...
+              typeTerms.remove(SECRET);
+              typeTerms.add(SCALAR_SECRET);
+            }
+          });
+        }
+      }
+    }.walk(extensionLoadingContext.getExtensionDeclarer().getDeclaration());
+  }
+
+  private void addTermIfPresent(Set<String> terms, WithSemanticTermsDeclaration declaration, String term, String mappedTerm) {
```
ndinu, line 87: Do we want to maintain both terms in the set? for example NTLM_PROXY_CONFIGURATION and NTLM_PROXY_CONFIGURATION_PARAMETER, or should it be replaced?

marianogonzalez: `NTLM_PROXY_CONFIGURATION` is kept as a type term, while `NTLM_PROXY_CONFIGURATION_PARAMETER` gets added on the parameter

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/semantic/SemanticTermsEnricher.java
marianogonzalez: good catch

ndinu (approved):

PR #10263, opened 2021-04-14
MULE-19328: Inconsistent flow's processing strategy state after stop

IvanAndresFritzler:
…er to zero

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/processor/strategy/StreamEmitterProcessingStrategyFactory.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/processor/strategy/StreamEmitterProcessingStrategyFactory.java
@@ -103,7 +103,7 @@ public class StreamEmitterProcessingStrategyFactory extends AbstractStreamProces
   }
 
 
-  static class StreamEmitterProcessingStrategy extends AbstractReactorStreamProcessingStrategy implements Stoppable {
+  static class StreamEmitterProcessingStrategy extends AbstractReactorStreamProcessingStrategy {
 
     private static final Logger LOGGER = getLogger(StreamEmitterProcessingStrategy.class);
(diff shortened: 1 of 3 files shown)


tests/component-plugin/src/main/java/org/mule/functional/api/flow/FlowRunner.java
```diff
@@ -371,6 +371,14 @@ public void runExpectingException(Matcher<Throwable> causeMatcher, Matcher<CoreE
     }
   }
 
+  /**
+   * Stops the flow and starts it again.
+   */
+  public void restartFlow() throws MuleException {
```
elrodro83, line 377: remove this (see comments in the tests pr)

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/StreamEmitterProcessingStrategyFactory.java
```diff
@@ -147,6 +147,7 @@ public void stop() {
       if (allSchedulersStopped()) {
         stopSchedulersIfNeeded();
       }
+      super.stop();
```
elrodro83, line 150: should this be added to other implementations of PSs?

elrodro83 (changes requested):

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/StreamEmitterProcessingStrategyFactory.java
IvanAndresFritzler: AbstractReactorStreamProcessingStrategy is the one that introduces the inflightEvents concept and was not implementing Stoppable. I did not found other PS extending such class or incorporating a similar concept.

elrodro83 (approved):

PR #10268, opened 2021-04-15
MULE-19347: Transaction bind fails because CachedConnectionHandler re…

emariotti3:
…turns cached invalid connection.

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/transaction/TransactionBindingDelegate.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/transaction/TransactionBindingDelegate.java
@@ -10,6 +10,7 @@ import static java.lang.String.format;
 import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
 import static org.mule.runtime.extension.api.util.NameUtils.getComponentModelTypeName;
 import static org.slf4j.LoggerFactory.getLogger;
+import static org.mule.runtime.core.api.util.ExceptionUtils.extractConnectionException;
 
 import org.mule.runtime.api.connection.ConnectionException;
 import org.mule.runtime.api.connection.ConnectionHandler;
@@ -123,10 +124,15 @@ public class TransactionBindingDelegate {
                                                                   extensionModel.getName(),
                                                                   connection.getClass().getName())));
       }
+    } catch (Exception e) {
+      if (extractConnectionException(e).isPresent()) {
+        connectionHandler.invalidate();
+      }
+      throw e;
     } finally {
       if (!bound) {
         try {
-          connectionHandler.invalidate();
+          connectionHandler.release();
         } catch (Exception e) {
           final String msg = "Ignored '" + e.getClass().getName() + ": " + e.getMessage() + "' during connection release";
           if (LOGGER.isDebugEnabled()) {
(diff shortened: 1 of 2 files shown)


emariotti3:
Test at https://github.com/mulesoft/mule-jms-connector/tree/MULE-19347

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/transaction/TransactionBindingDelegate.java
```diff
@@ -126,7 +126,7 @@ public void invalidate() {
     } finally {
       if (!bound) {
         try {
-          connectionHandler.release();
+          connectionHandler.invalidate();
```
marianogonzalez, line 129: this is assuming that the binding failed because of a connectivity issue, but that's not necessarily the case. The binding can fail for a number of reasons, like the transaction being already bound to an incompatible resource type or other things. In that case, you're severing a connection that might be perfectly valid, leading to issues. 

You can only do this if you're certain that a connection problem occurred, in which case, I would challenge if this is the right place to do this.

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -1090,14 +1092,26 @@ protected boolean isAsync() {
     } else {
       Optional<ConfigurationInstance> staticConfig = getStaticConfiguration();
       if (staticConfig.isPresent()) {
-        RetryPolicyTemplate resolvedRetryPolicyTemplate = getRetryPolicyTemplate(staticConfig);
-        return resolvedRetryPolicyTemplate.isEnabled() && resolvedRetryPolicyTemplate.isAsync();
+        return isAsyncExecutableBasedOn(staticConfig);
```
marianogonzalez, line 1095: changes to this class don't belong here, but in the PR that @fsgonz is working on

modules/spring-config/src/main/java/org/mule/runtime/config/internal/MuleArtifactContext.java
```diff
@@ -157,6 +158,7 @@
     configureBatchFixedAggregatorTransactionRecordBuffer();
     configureComputeConnectionErrorsInStats();
     configureToStringTransformerTransformIteratorElements();
+    configureResolveExuectionModeBasedOnEnabledReconnectionStrategy();
```
marianogonzalez, line 161: doesn't belong here

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/transaction/TransactionBindingDelegate.java
```diff
@@ -123,6 +124,11 @@ public void invalidate() {
                                                                   extensionModel.getName(),
                                                                   connection.getClass().getName())));
       }
+    } catch (Exception e) {
```
marianogonzalez, line 127: this changes looks good and I understand there's a functional test in the jms connector, at the very least a unit test should exist over this

marianogonzalez (changes requested):

marianogonzalez (approved):

PR #10274, opened 2021-04-20
MULE-19356: Avoid managing source lifecycle from flow when it has no source

elrodro83:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/config/DefaultComponentInitialStateManager.java
```diff
@@ -25,9 +25,10 @@
 
   @Override
   public boolean mustStartMessageSource(Component component) {
-    if (configurationProperties.resolveProperty(DISABLE_SCHEDULER_SOURCES_PROPERTY).isPresent()) {
+    if (configurationProperties.resolveProperty(MuleSystemProperties.DISABLE_SCHEDULER_SOURCES_PROPERTY).isPresent()) {
```
eze210, line 28: static import

eze210 (changes requested):

modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/config/DefaultComponentInitialStateManager.java
eze210: ok

eze210 (approved):

eze210 (approved):

PR #10281, opened 2021-04-22
MULE-19100 : Provide a reusable mechanism for using object fields as acting parameters

ndinu:


ndinu:
--validate

marianogonzalez (approved):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/InjectableParameterResolver.java
```diff
@@ -0,0 +1,161 @@
...
+    parameterValue = resolvedParameterValues.get(parameterName);
+    InjectableParameterInfo injectableParameterInfo = injectableParametersMap.get(parameterName);
+
+    if (parameterValue != null) {
+      try {
+        parameterValue = expressionManager
+            .evaluate("#[payload]",
+                      DataType.fromType(getClassFromType(injectableParameterInfo.getType())),
+                      BindingContext.builder()
+                          .addBinding("payload", new TypedValue(parameterValue, DataType.fromObject(parameterValue))).build())
```
marianogonzalez, line 78: what about the mediatType? Doesnt the `fromObject` method return  `application/java` for a String containing a Json. Are there any tests using this with json payloads?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/InjectableParameterResolver.java
```diff
@@ -0,0 +1,161 @@
...
+            .getValue();
+      } catch (ClassNotFoundException e) {
+        LOGGER.debug("Transformation of injectable parameter '{}' failed, the same value of the resolution will be used.",
```
marianogonzalez, line 81: why is a class cast something that you can just ignore and continue as is? Shouldn't this actually raise an error to be dealt with?

ndinu: It is a class not found exception. If the current class loader does not have the class being looked for, we will do a best effort and try to inject the value as is to the value provider.

PR #10282, opened 2021-04-22
MULE-19342: Parallelism problem in blocking ops processing.

fsgonz:


elrodro83 (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -696,7 +692,10 @@ public ProcessingType getProcessingType() {
 
         @Override
         public boolean isAsync() {
```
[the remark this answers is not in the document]
fsgonz: If there is logic in using this in the proactor is the defintion in the jdoc of the OperationInnerProcessor
"An async processor is one that eventually changes the thread where it is executing.". That would be the criteria for adding parallelism or not in the proactor.
The problem is that the isAsync in the owner class has other semantics (it indicates that the processor *would* eventually change the thread, not that it would in all cases). I'll change the name of the methods to make it clear.

elrodro83 (changes requested):

elrodro83 (dismissed):

fsgonz:
--validate

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/ComponentProcessor.java
```diff
@@ -0,0 +1,38 @@
...
+   */
+  boolean isBlocking();
+
+  /**
+   * Indicates that the processor may change the thread where it is executing but not always. For example if a reconnection
+   * strategy is applied and a connection problem happens.
+   *
+   * @return {@code} if the processor may be asynchronous.
+   */
+  boolean mayBeAsync();
```
marianogonzalez, line 37: this name is misleading and could be interpreted as something that is only important when `isBlocking()` returns false. I would rename this to `mayJumpThreads` or something like that

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/ComponentProcessor.java
```diff
@@ -0,0 +1,38 @@
...
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.internal.processor.strategy;
+
+import org.mule.runtime.core.api.processor.ReactiveProcessor;
+
+/**
+ * Specialization of {@link ReactiveProcessor} that allows to perform certain optimizations when the processor is not blocking, as
```
marianogonzalez, line 12: allows to perform -> allows performing

marianogonzalez, line 12: I think this contract needs to be better defined. If the purpose is to allow for optimizations, then the interface name should reflect that an execution type is what's being modeled. If what this is modeling is the execution of a mule component, then the jdoc should reflect that and execution optimizations is only ONE aspect of that

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/ComponentProcessor.java
```diff
@@ -0,0 +1,38 @@
...
+ * indicated by its {@link #isBlocking()} method and that evaluates if a thread switch may happen, as indicated by its
```
marianogonzalez, line 13: this sentence is confusing. It's not clear if `mayBeAsync()` should only be considered whtn the operation is blocking, or if the `and` is a typo which should be removed.

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/ComponentProcessor.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public interface ComponentProcessor extends ReactiveProcessor {
+
+  /**
+   * A blocking processor is one that will perform blocking calls. So it is not guaranteed that a thread switch happens before
+   * that call or that the operation returns without blocking.
+   *
+   * @return {@code} if the processor is blocking.
+   */
+  boolean isBlocking();
```
marianogonzalez, line 29: I think that splitting one single semantic into two different methods is a bad practice which opens the door to inconsistent states, not to mention that it forces a very specific consumption contracts.

I would replace these two methods by a single which returns an enum with values `FIX_THREAD_BLOCKING, THEAD_JUMPING_BLOCKING, ASYNC`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -825,73 +827,18 @@ private void unsetCurrentLocation(boolean wasProcessorPathSet) {
 
   private void initRetryPolicyResolver() {
     Optional<ConfigurationInstance> staticConfig = getStaticConfiguration();
+    RetryPolicyTemplateResolver policyResolver =
+        getRetryPolicyTemplateResolver();
```
marianogonzalez, line 831: single line

marianogonzalez (changes requested):

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/ComponentProcessor.java
fsgonz: only isBlocking was left as this is the only aspect from the operation/component which should be taken into account for the optimization.

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/ComponentProcessor.java
fsgonz: only isBlocking was left as this is the only aspect from the operation/component which should be taken into account for the optimization. The jdoc and the interface was renamed again.

fsgonz:
--validate

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/ComponentProcessor.java
fsgonz: I've used the original motivation for the introduction of this inner processor to write the jdoc (originally named OperationInnerProcessor MULE-17939). I've changed the name to ComponentInnerProcessor. The idea is to apply the processing strategy only to the operation when there is a policy involved.

fsgonz: Second option:
Specialization of {@link ReactiveProcessor} that allows execution of a mule component taking into account its processing strategy when it has to be decoupled from a policy (see MULE-17939). This allows performing certain optimizations when the processor is not blocking, as indicated by its {@link #isBlocking()} method.

marianogonzalez (approved):

PR #10290, opened 2021-04-26
MULE-19370: Avoid using reactor operators for callbacks that may be passed when subscribing

elrodro83:


core/src/main/java/org/mule/runtime/core/privileged/processor/chain/AbstractMessageProcessorChain.java
```diff
@@ -200,7 +205,7 @@ public CoreEvent process(CoreEvent event) throws MuleException {
                                             }))
                                                 .map(result -> result.reduce(me -> {
```
IvanAndresFritzler, line 206: I think that this duplicates MessageProcessors.propagateErrorResponseMapper() -wich btw doesn't have the identity() sugar-

IvanAndresFritzler (changes requested):

IvanAndresFritzler (approved):

IvanAndresFritzler (approved):

PR #10297, opened 2021-04-28
MULE-19302: Remove lazyInit concept from LifecycleAwareConfigurationInstance

elrodro83:


elrodro83:
--validate

core/src/main/java/org/mule/runtime/core/internal/connection/DefaultConnectivityTesterFactory.java
```diff
@@ -0,0 +1,178 @@
...
+
+  @Override
+  public ConnectivityTester create(String name) {
+    final Lock testConnectivityLock = lockFactory.createLock(this.getClass().getName() + "-testConnectivity-" + name);
+    return new ConnectivityTester() {
+
+      @Override
+      public void testConnectivity(ConnectionProvider provider, ConfigurationInstance configurationInstance)
+          throws MuleException {
+        if (provider instanceof NoConnectivityTest || !doTestConnectivity.get()) {
```
fsgonz, line 77: Is there a way to avoid instanceof without changing api? It becomes a problem if some day we want to decorate the provider.

core/src/main/java/org/mule/runtime/core/internal/connection/DefaultConnectivityTesterFactory.java
```diff
@@ -0,0 +1,178 @@
...
+        Scheduler retryScheduler = schedulerService.ioScheduler();
+        RetryPolicyTemplate retryTemplate = connectionManager.getRetryTemplateFor(provider);
+        ReconnectionConfig reconnectionConfig = connectionManager.getReconnectionConfigFor(provider);
+        final Latch latch = new Latch();
+        RetryCallback retryCallback = new RetryCallback() {
+
+          @Override
+          public void doWork(RetryContext context) throws Exception {
+            try {
+              Lock lock = testConnectivityLock;
```
fsgonz, line 90: is this variable really needed?

core/src/main/java/org/mule/runtime/core/internal/connection/DefaultConnectivityTesterFactory.java
```diff
@@ -0,0 +1,178 @@
...
+              if (lock != null) {
+                final boolean lockAcquired = lock.tryLock();
+                if (lockAcquired) {
+                  LOGGER.debug("Doing testConnectivity() for config '{}'", name);
+                  try {
+                    ConnectionValidationResult result = connectionManager.testConnectivity(configurationInstance);
+                    if (result.isValid()) {
+                      context.setOk();
+                    } else {
+                      if ((reconnectionConfig.isFailsDeployment())) {
```
fsgonz, line 100: if this does not affect performance (I don't think so), I would extract the case where the test connectivity fails at least to a method to make the general doWork method more declarative and avoid problems if this logics gets more troublesome.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/lazy/NoOpConnectivityTesterFactory.java
```diff
@@ -0,0 +1,36 @@
...
+package org.mule.runtime.config.internal.lazy;
+
+import org.mule.runtime.api.connection.ConnectionProvider;
+import org.mule.runtime.api.exception.MuleException;
+import org.mule.runtime.core.api.util.func.CheckedRunnable;
+import org.mule.runtime.core.internal.connection.ConnectivityTester;
+import org.mule.runtime.core.internal.connection.ConnectivityTesterFactory;
+import org.mule.runtime.extension.api.runtime.config.ConfigurationInstance;
+
+public class NoOpConnectivityTesterFactory implements ConnectivityTesterFactory {
```
fsgonz, line 16: add jdoc.

fsgonz (changes requested):

core/src/main/java/org/mule/runtime/core/internal/connection/DefaultConnectivityTesterFactory.java
fsgonz: The ideal would be to avoid the if entirely with some kind of double dispatch but maybe we can avoid the instanceof and avoid problems if some day we have to wrap the ConnectionProvider.

elrodro83: `NoConnectivityTest` comes from the sdk, not runtime, so cannot do that without extending its API which would have a much broader impact.

elrodro83: Actually, seeing how this interface is used in the connectors, an annotation would have been better. But then again, this logic in the runtime would be very similiar. Keep in mind that the design entention of the sdk is to make life easier to the people writing connectors, not the runtime.

fsgonz: Ok. Even if this is used by the connectors the best way would not to depend of the interface with an instanceof to decide one thing or the other. This cannot be changed for this PR but should be taken into account.

fsgonz (approved):

PR #10300, opened 2021-04-29
MULE-18580: ZonedDateTime Parameters are not parsed from Strings

dabreu:


tests/test-extensions/some-extension/src/main/java/org/mule/test/some/extension/SomeExtension.java
```diff
@@ -7,22 +7,37 @@
 
 package org.mule.test.some.extension;
 
+import org.mule.runtime.extension.api.annotation.Configuration;
+import org.mule.runtime.extension.api.annotation.Configurations;
```
ndinu, line 11: remove unused import

tests/test-extensions/some-extension/src/main/java/org/mule/test/some/extension/SomeExtension.java
```diff
@@ -7,22 +7,37 @@
...
 import org.mule.runtime.extension.api.annotation.dsl.xml.Xml;
 import org.mule.runtime.extension.api.annotation.error.ErrorTypes;
+import org.mule.sdk.api.annotation.param.Optional;
+import org.mule.sdk.api.annotation.param.Parameter;
 import org.mule.test.heisenberg.extension.HeisenbergErrors;
 
+import java.time.ZonedDateTime;
+
 @Extension(name = "SomeExtension")
+@Configuration(name = "config")
```
SebaElizalde, line 26: why is this needed?

tests/test-extensions/some-extension/src/main/java/org/mule/test/some/extension/SomeOps.java
```diff
@@ -26,8 +29,27 @@
...
 
+  public SomeExtension retrieveConfiguration(@Config SomeExtension config) {
+    return config;
+  }
+
+  public ZonedDateTime retrieveOperationZonedDateTime() {
+    return zonedDateTime;
+  }
+
+  public boolean isAfter(ZonedDateTime zonedDateTimeParam, @Config SomeExtension config) {
```
SebaElizalde, line 46: it is not clear to me the purpose of this operation it just tests the isAfter method of the Java class

SebaElizalde (changes requested):

tests/test-extensions/some-extension/src/main/java/org/mule/test/some/extension/SomeExtension.java
dabreu: will do

tests/test-extensions/some-extension/src/main/java/org/mule/test/some/extension/SomeExtension.java
dabreu: I added the name in order to reference the config on the parameter-value-resolver.xml file

tests/test-extensions/some-extension/src/main/java/org/mule/test/some/extension/SomeOps.java
dabreu: The intention of this method was to add a test scenario to validate that a ZonedDateTime operation parameter/argument is being correctly parsed from the configured string on the flow xml file.

SebaElizalde: Maybe make the test just return the parameter then? Plus adding a parameter to the SomeOps class like you did:
```
  @Parameter
  @Optional
  private ZonedDateTime zonedDateTime;
```
 makes all methods/operations defined in that class also have that parameter already. So maybe this can be simplified so there are less parameters and it is not so confusing.

tests/test-extensions/some-extension/src/main/java/org/mule/test/some/extension/SomeExtension.java
SebaElizalde: oh, it became required because you added an optional parameter to the config?

tests/test-extensions/some-extension/src/main/java/org/mule/test/some/extension/SomeExtension.java
```diff
@@ -7,22 +7,36 @@
...
 @Export(classes = CustomConnectionException.class)
-@Sources({SomeEmittingSource.class, AnotherEmittingSource.class, YetAnotherEmittingSource.class})
+@Sources({SomeEmittingSource.class, AnotherEmittingSource.class, YetAnotherEmittingSource.class,
+    ParameterEmittingSource.class})
 @Xml(namespace = "http://www.mulesoft.org/schema/mule/some", prefix = "some")
 public class SomeExtension {
 
+  @Parameter
+  @Optional
+  private ZonedDateTime zonedDateTime;
```
SebaElizalde, line 37: Also if there aren't any differences in picking up a value from a config or a source or a parameter, maybe it is not needed to test all possible combinations (config + source + operation + another operation), but it does not hurt either... so I leave that up to you.

tests/test-extensions/some-extension/src/main/java/org/mule/test/some/extension/SomeOps.java
dabreu: Yes, makes sense, I will simplify the operation as you suggested

SebaElizalde (approved):

PR #10302, opened 2021-04-30
MULE-19380: Tooling: Do a proper lazy init in AbstractArtifactAgnosticService

elrodro83:


modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/AbstractArtifactAgnosticService.java
```diff
@@ -63,7 +72,26 @@ protected Application getStartedApplication() throws ApplicationStartingExceptio
...
         application.install();
-        application.init();
+
+        if (isLazyInit()) {
+          try {
+            application.lazyInit(!isLazyInitEnableXmlValidations());
+            // Force the initialization of the app components so that the registry is populated with the configs
+            application.getRegistry().lookupByType(LazyComponentInitializer.class)
+                .ifPresent(lazyInit -> lazyInit
+                    .initializeComponents((Predicate<ComponentAst>) comp -> comp.getModel(ConfigurationModel.class).isPresent()));
```
gsfernandes, line 82: We do need to enable also every global element besides Configurations.

gsfernandes: The idea at this point is that the application should only has globalElements (topLevel pojos) or configurations.

gsfernandes (changes requested):

elrodro83:
--validate

gsfernandes (approved):

elrodro83:
--validate

gsfernandes (approved):

PR #10312, opened 2021-05-03
MULE-19404: XmlSdk: Use a pool of xslt tranformers in the extModel generation

elrodro83:


modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/loader/XmlExtensionLoaderDelegate.java
```diff
@@ -228,7 +232,11 @@ private static ParameterRole getRole(final String role) {
   private static final String XML_SUFFIX = ".xml";
   private static final String TYPES_XML_SUFFIX = "-catalog" + XML_SUFFIX;
 
-  final Set<ComponentIdentifier> NOT_GLOBAL_ELEMENT_IDENTIFIERS =
+  private static final PoolService<Transformer> FOR_TNS_XSTL_TRANSFORMER_POOL =
+      new ConcurrentPool<>(new ConcurrentLinkedQueueCollection<>(), new ForTnsTransformerFactory(), 1,
+                           max(1, getRuntime().availableProcessors() / 2), false);
```
eze210, line 237: Use a constant. Is that size arbitrary or has a why?

eze210 (changes requested):
Add a test

eze210 (approved):

PR #10321, opened 2021-05-05
MULE-19390: ExclusiveOptionals PG with Dsl True fails when there are repeated names

SebaElizalde:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -584,4 +601,70 @@ public static boolean isConnectedStreamingOperation(ComponentModel componentMode
...
+   *
+   * @since 4.4.0
+   */
+  public static void checkParameterGroupExclusiveness(Optional<ParameterizedModel> model,
+                                                      List<ParameterGroupModel> groups,
+                                                      Map<String, ?> parameters, Map<String, String> aliasedParameterNames)
+      throws ConfigurationException {
+    Set<String> explodedParameterNames = new HashSet<>();
+    Set<String> topLevelParameterNames = parameters.entrySet().stream()
+        .flatMap(entry -> {
```
ndinu, line 624: Wouldn't this just be a map?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -584,4 +601,70 @@ public static boolean isConnectedStreamingOperation(ComponentModel componentMode
...
+          if (entry.getValue() instanceof ParameterValueResolver) {
+            try {
+              ((ParameterValueResolver) entry.getValue()).getParameters().keySet()
+                  .forEach(k -> explodedParameterNames.add(aliasedParameterNames.getOrDefault(k, k)));
```
ndinu, line 628: I think it is weird to do processing and a adding values to a list in the map function. It should only take care of the mapping of the item.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -584,4 +601,70 @@ public static boolean isConnectedStreamingOperation(ComponentModel componentMode
...
+          aliasedParameterNames.getOrDefault(key, key);
+          return Stream.of(key);
+        })
+        .collect(toSet());
+
+    Set<String> parameterNames;
+    for (ParameterGroupModel group : groups) {
+      for (ExclusiveParametersModel exclusiveModel : group.getExclusiveParametersModels()) {
+        if (group.isShowInDsl()) {
+          parameterNames = explodedParameterNames;
```
ndinu, line 643: If we do this, and there was a pojo that is not in a ShowInDsl group, we would get its fields as parameter names here.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -584,4 +601,70 @@ public static boolean isConnectedStreamingOperation(ComponentModel componentMode
...
+   * @param model                 the model that owns the parameter groups
+   * @param groups                the parameter groups in the model whose exclusiveness must be honored
+   * @param parameters            the resolved parameters
+   * @param aliasedParameterNames the alias for each parameter if it has one
+   *
+   * @throws ConfigurationException if the group exclusiveness is not honored by the resolved parameters.
+   *
+   * @since 4.4.0
+   */
+  public static void checkParameterGroupExclusiveness(Optional<ParameterizedModel> model,
```
ndinu, line 618: What is the reason to move this to a util class?

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/util/MulExtensionUtilsTestCase.java
```diff
@@ -65,6 +80,144 @@ public void optionalWithDefaultValue() {
...
+    parameterResolvers.put("anotherParameter", mock(StaticValueResolver.class));
+    parameterResolvers.put("repeatedNameParameter", mock(StaticValueResolver.class));
+    ObjectBuilderValueResolver valueResolver = mock(ObjectBuilderValueResolver.class);
+    when(valueResolver.getParameters()).thenReturn(parameterResolvers);
+    Map<String, ObjectBuilderValueResolver> parameters = new HashMap<>();
+    parameters.put("complexParameter", valueResolver);
+    Map<String, String> aliasedParameterNames = new HashMap<>(0);
+    checkParameterGroupExclusiveness(Optional.of(mock(OperationModel.class)), groups, parameters, aliasedParameterNames);
+  }
+
```
ndinu, line 220: Add a test where there is another parameter group, that is not show in DSL, that has a parameter that is a POJO, and has a field with the name of a parameter in the exclusive optional parameter group that is Show in DSL.

ndinu (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
SebaElizalde: previously, it was a public method inside a class but it used no internal fields or state of the class. it checks things over the model so it made sense to me to move it,

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -584,4 +601,82 @@ public static boolean isConnectedStreamingOperation(ComponentModel componentMode
...
+          .orElseThrow(() -> new ConfigurationException(createStaticMessage("Could not find ExtensionParameterDescriptorModelProperty for the parameter group %s",
+                                                                            group.getName())));
+      String containerName = property.getExtensionParameter().getName();
+      try {
+        Object parameter = parameters.get(containerName);
+        if (parameter == null) {
+          throw new ConfigurationException(createStaticMessage("Was expecting a parameter with name [%s] among the resolved parameters",
+                                                               containerName));
+        }
+        return ((ParameterValueResolver) parameters.get(containerName)).getParameters().keySet().stream()
```
ndinu, line 669: I would make a check before casting this, so that we get a better error message of this is not a ParameterValueResolver.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -584,4 +601,82 @@ public static boolean isConnectedStreamingOperation(ComponentModel componentMode
...
+          }
+        }
+      }
+    }
+  }
+
+  private static Set<String> resolveParameterNames(ParameterGroupModel group, Map<String, ?> parameters,
+                                                   Map<String, String> aliasedParameterNames)
+      throws ConfigurationException {
+    Set<String> topLevelParameterNames = new HashSet<>();
```
ndinu, line 652: Shouldn't we only do this if group is not show in DSL?

SebaElizalde: yeah

tests/test-extensions/some-extension/src/main/java/org/mule/test/some/extension/SomeExtension.java
```diff
@@ -7,36 +7,22 @@
...
 import org.mule.runtime.extension.api.annotation.error.ErrorTypes;
-import org.mule.sdk.api.annotation.param.Optional;
-import org.mule.sdk.api.annotation.param.Parameter;
 import org.mule.test.heisenberg.extension.HeisenbergErrors;
 
-import java.time.ZonedDateTime;
-
 @Extension(name = "SomeExtension")
-@Configuration(name = "config")
+@Configurations({ParameterGroupConfig.class, ParameterGroupDslConfig.class})
```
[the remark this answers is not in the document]
SebaElizalde: yup, they were addressed

ndinu (approved):

PR #10322, opened 2021-05-05
MULE-19388: ValueProviderModels from sdk-api usages must provide a SinceMuleVersionModelProperty

ndinu:


modules/extensions-spring-support/src/test/resources/models/test-oauth-ocs.json
```diff
@@ -6472,7 +6472,8 @@
                     "providerId": "org.mule.test.oauth.OAuthValuesProvider",
                     "requiresConfiguration": false,
                     "requiresConnection": true,
-                    "isOpen": true
+                    "isOpen": true,
+                    "modelProperties": {}
```
lucianoRM, line 6476: Just a question. These empty model properties were added here only because the test that uses this file compares Json, right? 
Are we still maintaining backwards compatibility with models serialized before this change? I can see that the constructor el ValueProviderModel is handling this but just wanted to make sure.

ndinu: Discussed offline

SebaElizalde (approved):

PR #10338, opened 2021-05-13
MULE-10052 - Added Extension's validator to warn the case where any @…

dabreu:
…Parameter present in the Extension class will be ignored if the extension is annotated with @Configurations

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/IgnoredExtensionParameterModelValidator.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/IgnoredExtensionParameterModelValidator.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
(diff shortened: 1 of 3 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/IgnoredExtensionParameterModelValidator.java
```diff
@@ -0,0 +1,69 @@
...
+  @Override
+  public void validate(ExtensionModel extensionModel, ProblemsReporter problemsReporter) {
+    if (!isCompiletime(extensionModel)) {
+      return;
+    }
+
+    extensionModel.getModelProperty(ExtensionTypeDescriptorModelProperty.class)
+        .map(descriptor -> descriptor.getType())
+        .ifPresent(type -> {
+          checkIgnoredParameters(extensionModel, (ExtensionElement) type, problemsReporter);
```
ndinu, line 48: I would make a check before casting directly to ExtensionElement.

dabreu: done!

ndinu (approved):

PR #10339, opened 2021-05-13
MULE-19390-4.3.x: ExclusiveOptionals PG with Dsl True fails when there are repeated names

SebaElizalde:


ndinu (approved):
Remember to remove the  -4.3.x from the title when squash and merging.

PR #10351, opened 2021-05-17
MULE-19199: LazyInit: Remove the flag set into the configuration properties

elrodro83:


core/src/main/java/org/mule/runtime/core/api/context/DefaultMuleContextFactory.java
```diff
@@ -161,19 +152,15 @@ public MuleContext createMuleContext(final List<ConfigurationBuilder> configurat
...
-
-      @Override
-      public void configure(MuleContext muleContext) throws ConfigurationException {
-        // Configure with startup properties
-        if (properties != null && !properties.isEmpty()) {
-          new SimpleConfigurationBuilder(properties).configure(muleContext);
-        }
+    return doCreateMuleContext(contextBuilder, muleContext -> {
+      // Configure with startup properties
+      if (properties != null && !properties.isEmpty()) {
```
fsgonz, line 157: use MapUtils.isEmpty

fsgonz (changes requested):

fsgonz (approved):

PR #10352, opened 2021-05-17
MULE-19426: BlockingProcessingStrategy not working for non-blocking operations with execution type BLOCKING	

eze210:


core/src/main/java/org/mule/runtime/core/internal/processor/strategy/BlockingProcessingStrategyFactory.java
```diff
@@ -81,5 +84,26 @@ public ReactiveProcessor onProcessor(ReactiveProcessor processor) {
...
+     * @return true if a Mono.block call is needed to wait for operation completion.
+     */
+    private static boolean needsMonoBlock(ReactiveProcessor processor) {
+      try {
+        // This method is added in the SDK, and it's return value is false for those operations which include a
+        // completion callback. That completion callback is the way to implement a non-blocking operations, so if this
+        // method returns true, we'll need a Mono.block call to wait the operation completion.
+        Method isBlockingMethod = processor.getClass().getDeclaredMethod("isBlocking");
+        isBlockingMethod.setAccessible(true);
+        return !(boolean) isBlockingMethod.invoke(processor);
```
elrodro83, line 103: This is expensive and being done for every call. Cache this result or precalculate and make it available some other way.

elrodro83 (changes requested):

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/BlockingProcessingStrategyFactory.java
eze210: Reflection is baaad

eze210:
Addressed in MULE-19431

PR #10353, opened 2021-05-17
MULE-19426: BlockingProcessingStrategy not working for non-blocking operations with execution type BLOCKING

eze210:


eze210:
Addressed in MULE-19431

PR #10356, opened 2021-05-18
MULE-19101: Support object field acting parameters in SampleData

Bardammu:


modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/enricher/SampleDataDeclarationEnricherTestCase.java
```diff
@@ -135,7 +165,21 @@ private void assertWithRequiredParameter(OperationDeclaration operationDeclarati
     }
   }
 
+  private void assertAliasedParameter(OperationDeclaration operationDeclaration, String alias, String name) {
```
ndinu, line 168: Was this test just moved, or has it changed?

ndinu (changes requested):

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/enricher/SampleDataDeclarationEnricherTestCase.java
Bardammu: Just moved,  I usially prefer the auxiliary methods and the end of the file

ndinu (approved):

ndinu (approved):

ndinu (approved):

PR #10357, opened 2021-05-18
MULE-19402: FeatureFlaggingService improvement

IvanAndresFritzler:


IvanAndresFritzler:
--validate

IvanAndresFritzler:
--validate

core/src/main/java/org/mule/runtime/core/api/config/FeatureFlaggingRegistry.java
```diff
@@ -70,35 +71,69 @@ private FeatureFlaggingRegistry() {}
...
   public void registerFeature(Feature feature, Predicate<MuleContext> condition) {
+    validate(feature, condition);
+    Predicate<MuleContext> added = configurations.putIfAbsent(feature, condition);
+    if (added != null) {
+      throw new IllegalArgumentException(format(FEATURE_ALREADY_REGISTERED, feature));
+    }
+  }
+
+  /**
+   * Registers a {@link Predicate} associated with a String which represents a given feature. The {@link Predicate} will be
```
elrodro83, line 86: what do you mean by `associated with a String`? where is a String here?

core/src/main/java/org/mule/runtime/core/api/config/FeatureFlaggingRegistry.java
```diff
@@ -70,35 +71,69 @@ private FeatureFlaggingRegistry() {}
...
+   * evaluated at deployment time, exposing all the features through a per-application overridden {@link FeatureFlaggingService}
+   *
+   * @see FeatureFlaggingService
+   *
+   * @param feature   Name representing the registered feature
```
elrodro83, line 91: `name` is misleading

core/src/main/java/org/mule/runtime/core/internal/config/FeatureFlaggingServiceBuilder.java
```diff
@@ -19,56 +19,123 @@
 import org.mule.runtime.api.config.Feature;
 import org.mule.runtime.core.api.MuleContext;
 import org.mule.runtime.api.config.FeatureFlaggingService;
+import org.mule.runtime.core.api.config.FeatureContext;
 import org.slf4j.Logger;
 
 /**
- * This builder creates a {@link FeatureFlaggingService}.
+ * <p>
```
elrodro83, line 26: remove this separator at the beginning

elrodro83 (changes requested):

core/src/main/java/org/mule/runtime/core/api/config/FeatureFlaggingRegistry.java
IvanAndresFritzler: I just copied the deprecated method documentation, but this is a confusing text indeed... will update both.

elrodro83 (approved):

PR #10359, opened 2021-05-18
MULE-19430: Allow XA to suspend multiple transactions

jcgrethe:


mbuchwald (approved):

core/src/main/java/org/mule/runtime/core/api/transaction/TransactionCoordination.java
```diff
@@ -27,7 +29,7 @@
    * {@link #bindTransaction(Transaction)}, it may be more consistent to have it as an instance variable.
    */
   private final ThreadLocal<Transaction> transactions = new ThreadLocal<>();
-  private final ThreadLocal<Transaction> suspendedTransaction = new ThreadLocal<>();
+  private final ThreadLocal<Stack<Transaction>> suspendedTransaction = new ThreadLocal<>();
```
elrodro83, line 32: USe an `ArrayDeque` instead of `Stack`

PR #10362, opened 2021-05-19
MULE-19430: change stack for ArrayDeque

jcgrethe:


core/src/main/java/org/mule/runtime/core/api/transaction/TransactionCoordination.java
```diff
@@ -29,7 +29,7 @@
    * {@link #bindTransaction(Transaction)}, it may be more consistent to have it as an instance variable.
    */
   private final ThreadLocal<Transaction> transactions = new ThreadLocal<>();
-  private final ThreadLocal<Stack<Transaction>> suspendedTransaction = new ThreadLocal<>();
+  private final ThreadLocal<ArrayDeque<Transaction>> suspendedTransaction = new ThreadLocal<>();
```
elrodro83, line 32: declare as `Deque `

elrodro83 (changes requested):

PR #10373, opened 2021-05-26
MULE-19431: Transaction interceptors are being nested

eze210:


core/src/main/java/org/mule/runtime/core/internal/processor/strategy/BlockingProcessingStrategyFactory.java
```diff
@@ -81,5 +81,21 @@ public ReactiveProcessor onProcessor(ReactiveProcessor processor) {
       }
     }
 
+    /**
+     * This strategy adds a Mono.block call in order to preserve the thread, because it's a precondition to transactions to work.
```
fsgonz, line 85: precondition *for* transactions to work. I would remove the comma before because and the split the two last sentences.
"It  allows us to make a performance optimization for those operations.  This method is intended to detect which operations..."

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/DeferredExecutorCallbackTestCase.java
```diff
@@ -0,0 +1,108 @@
...
+import static java.lang.Thread.currentThread;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mule.runtime.core.api.util.ClassUtils.withContextClassLoader;
+import org.mule.runtime.extension.api.runtime.operation.CompletableComponentExecutor.ExecutorCallback;
+
+import org.junit.Test;
+
+public class DeferredExecutorCallbackTestCase {
```
fsgonz, line 18: Allure annotations.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DeferredExecutorCallback.java
```diff
@@ -0,0 +1,97 @@
...
+
+  @Override
+  public synchronized void close() throws Exception {
+    isEnabled = true;
+
+    if (!isErrorCalled && !isCompleteCalled) {
+      return;
+    }
+
+    Thread currentThread = currentThread();
```
fsgonz, line 83: maybe add the deferred operation to a different method so that this method is not so long.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DeferredExecutorCallback.java
```diff
@@ -0,0 +1,97 @@
...
+ *   }
+ *
+ *   // When the execution goes out of the try, delegate.complete(someValue) is called. It's useful because the
+ *   // interception logic isn't applied here.
+ * </code>
+ *
+ * The methods of this class are synchronized because the callback completion can be invoked from other thread than the close
+ * method. This shouldn't introduce a performance degradation because they're only called from two threads, and only once.
+ */
+class DeferredExecutorCallback implements ExecutorCallback, AutoCloseable {
```
fsgonz, line 34: Just a comment: I think it's ok to leave this a package class (preferable to an inner class) as this will be used only in this package.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/FutureExecutionCallbackAdapter.java
```diff
@@ -0,0 +1,33 @@
...
+package org.mule.runtime.module.extension.internal.runtime.operation;
+
+import org.mule.runtime.extension.api.runtime.operation.CompletableComponentExecutor.ExecutorCallback;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * ExecutorCallback adapter which delegates to the corresponding methods in a CompletableFuture.
+ */
+class FutureExecutionCallbackAdapter implements ExecutorCallback {
```
fsgonz, line 16: Just a comment: I think it's ok to leave this a package class (preferable to an inner class) as this will be used only in this package.

fsgonz (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/DeferredExecutorCallback.java
eze210: Yes, the idea was adding JavaDocs without increasing the length of the DefaultExecutorMediator code

fsgonz (dismissed):

fsgonz (approved):

PR #10374, opened 2021-05-26
MULE-19435 Wrap exception with MuleRuntimeException,

spagssfdc:
This prevents flux from getting stuck when a dynamic flow does not exist.


spagssfdc:
--validate

PR #10375, opened 2021-05-26
MULE-19442 - Operation Input Stream parameter of Parameter Group with…

dabreu:
… showInDsl=true is not reset after refreshing Oauth token

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
@@ -143,23 +143,25 @@ import org.mule.runtime.module.extension.internal.runtime.result.ValueReturnDele
 import org.mule.runtime.module.extension.internal.runtime.result.VoidReturnDelegate;
 import org.mule.runtime.module.extension.internal.runtime.streaming.CursorResetInterceptor;
 import org.mule.runtime.module.extension.internal.runtime.transaction.ExtensionTransactionFactory;
+import org.mule.runtime.module.extension.internal.util.IntrospectionUtils;
 import org.mule.runtime.module.extension.internal.util.ReflectionCache;
 
 import java.io.InputStream;
 import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.function.Supplier;
 
 import javax.inject.Inject;
@@ -167,6 +169,7 @@ import javax.inject.Inject;
 import org.reactivestreams.Publisher;
 import org.slf4j.Logger;
 import org.slf4j.MDC;
+
 import reactor.core.publisher.Flux;
 import reactor.util.context.Context;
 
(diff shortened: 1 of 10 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -195,58 +198,45 @@
...
+  protected CompletableComponentExecutor componentExecutor;
+  protected ReturnDelegate returnDelegate;
+  protected PolicyManager policyManager;
+  private Optional<TransactionConfig> transactionConfig;
   @Inject
   private ErrorTypeLocator errorTypeLocator;
-
   @Inject
   private Collection<ExceptionContextProvider> exceptionContextProviders;
-
```
marianogonzalez, line 228: don't remove this separations

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/streaming/CursorResetInterceptor.java
```diff
@@ -72,4 +68,28 @@ public Throwable onError(ExecutionContext<OperationModel> executionContext, Thro
 
     return exception;
   }
+
+  private List<Cursor> getParameterCursors(ExecutionContext<OperationModel> ctx) {
```
marianogonzalez, line 72: this introduces a performance regression. With this code, the list of parameters is iterated on every single operation execution, when the reality is that the stream parmeters are always the same. This is why the current implementation receives the exact list of parameter names to be fetched. This solution must respect that. This interceptor must have direct access to the parameters it needs to fetch

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/IntrospectionUtils.java
```diff
@@ -1675,4 +1678,71 @@ public static void setValueIntoField(Object target, Object value, Field field) {
...
+   * @param componentModel the component model used to filter the parameters
+   * @param filter         the filter to apply to the group model parameters
+   * @return The map having the parameter group as key and a set of its parameters matching the filter as value
+   */
+  public static Map<ParameterGroupModel, Set<ParameterModel>> getFilteredParameters(ComponentModel componentModel,
+                                                                                    Predicate<ParameterModel> filter) {
+    Map<ParameterGroupModel, Set<ParameterModel>> filteredParameters = new HashMap<>();
+    componentModel.getParameterGroupModels().forEach(parameterGroupModel -> {
+      parameterGroupModel.getParameterModels().stream()
+          .filter(p -> filter.test(p))
```
marianogonzalez, line 1694: .filter(filter)

marianogonzalez (changes requested):

tests/test-extensions/petstore-extension/src/main/java/org/mule/test/petstore/extension/PetStoreOperations.java
```diff
@@ -134,6 +135,38 @@ public Boolean isSplitterExceptionHandled() {
...
+  public List<String> getPetsWithParameterGroupShowDsl(@Connection PetStoreClient client,
+                                                       @Config PetStoreConnector config,
+                                                       @ParameterGroup(name = "Owner", showInDsl = true) PetOwner owner) {
+    if (owner.getAddress() != null) {
+      IOUtils.toString(owner.getAddress().getValue());
+    }
+    return getPets(client, config, owner.getName(), owner.getSignature());
+  }
+
+  public List<String> getPetsWithParameterGroupShowDslParameterWithAlias(@Connection PetStoreClient client,
```
SebaElizalde, line 153: you can join this operation with the one above to simplify and just pass null parameters when necessary, no?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/streaming/CursorResetInterceptor.java
dabreu: Oh, let me explain because it might not be clear, but this solution is similar to the existing one, where the stream parameters are pre-filtered.
Now, the constructor of CursorResetInterceptor  class receives as an argument the map with the stream parameters already filtered (done on ComponentMessageProcessor). Then, the 'before' method uses this pre-filtered Map and iterates it to fetch the value of each of the stream parameters from the context. It only iterates that map, not all the parameters. It is similar to the previous approach, but instead of using a list now it uses a map.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
dabreu: done

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/IntrospectionUtils.java
dabreu: changed

tests/test-extensions/petstore-extension/src/main/java/org/mule/test/petstore/extension/PetStoreOperations.java
dabreu: makes sense, changed

dabreu:
--validate

marianogonzalez (approved):

ndinu (approved):

PR #10376, opened 2021-05-26
MULE-19435 Wrap exception with MuleRuntimeException.

spagssfdc:
When a dynamic flow does not exist, throw a MuleRuntimeException instead of calling Flux.error so flux does not get blocked.


spagssfdc:
--validate

PR #10380, opened 2021-05-29
MULE-19263: validate CompletionCallback's generic types

Bardammu:
This commit modify the validation at compile time of operation's return type  to
forbide the existence of CompletionCallback with Void type for output and non Void type
for output attributes.

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/TypeWrapper.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/TypeWrapper.java
@@ -59,14 +59,15 @@ public class TypeWrapper implements Type {
   private LazyValue<Boolean> instantiable;
 
   public TypeWrapper(Class<?> aClass, ClassTypeLoader typeLoader) {
-    this.aClass = aClass;
+    this.aClass = aClass != null ? aClass : Object.class;
     instantiable = new LazyValue<>(() -> IntrospectionUtils.isInstantiable(aClass, new ReflectionCache()));
     this.type = aClass;
     this.typeLoader = typeLoader;
   }
 
   public TypeWrapper(ResolvableType resolvableType, ClassTypeLoader typeLoader) {
-    this.aClass = resolvableType.resolve();
+    Class<?> resolvedTypeClass = resolvableType.resolve();
+    this.aClass = resolvedTypeClass != null ? resolvedTypeClass : Object.class;
     instantiable = new LazyValue<>(() -> IntrospectionUtils.isInstantiable(aClass, new ReflectionCache()));
     this.type = resolvableType.getType();
     this.generics = new ArrayList<>();
@@ -160,7 +161,7 @@ public class TypeWrapper implements Type {
 
   @Override
   public boolean isAssignableFrom(Class<?> clazz) {
-    return aClass != null && aClass.isAssignableFrom(clazz);
+    return aClass.isAssignableFrom(clazz);
   }
 
   @Override
@@ -174,7 +175,7 @@ public class TypeWrapper implements Type {
 
   @Override
   public boolean isAssignableTo(Class<?> clazz) {
-    return aClass != null && clazz.isAssignableFrom(aClass);
+    return clazz.isAssignableFrom(aClass);
   }
 
   @Override
@@ -193,7 +194,7 @@ public class TypeWrapper implements Type {
 
   @Override
   public boolean isSameType(Class<?> clazz) {
-    return aClass != null && aClass.equals(clazz);
+    return aClass.equals(clazz);
   }
 
   @Override
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/TypeWrapper.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/TypeWrapper.java
@@ -193,7 +193,7 @@ public class TypeWrapper implements Type {
 
   @Override
   public boolean isSameType(Class<?> clazz) {
-    return aClass.equals(clazz);
+    return aClass != null && aClass.equals(clazz);
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/OperationReturnTypeModelValidator.java
```diff
@@ -88,9 +92,16 @@ private void validateNonBlockingCallback(MethodElement<? extends Type> operation
     operationMethod.getParameters().stream()
         .filter(p -> p.getType().isSameType(CompletionCallback.class))
         .findFirst().ifPresent(p -> {
-          if (p.getType().getGenerics().isEmpty()) {
+          List<TypeGeneric> generics = p.getType().getGenerics();
+          if (generics.isEmpty()) {
             problemsReporter.addError(new Problem(p, format(MISSING_GENERICS_ERROR_MESSAGE, operationModel.getName(),
                                                             extensionModel.getName(), CompletionCallback.class.getName())));
+          } else {
+            if (generics.get(0).getConcreteType().isSameType(Void.class) &&
```
ndinu, line 100: This restriction also applies to blocking operations, and operations that return message collection if it uses Result or List<Result> return type.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/OperationReturnTypeModelValidator.java
```diff
@@ -88,9 +92,16 @@ private void validateNonBlockingCallback(MethodElement<? extends Type> operation
...
+                !generics.get(1).getConcreteType().isSameType(Void.class)) {
+              problemsReporter.addError(new Problem(p, format(INVALID_GENERICS_ERROR_MESSAGE, operationModel.getName(),
```
ndinu, line 102: We should debate this, but I think this should only be done if compiling.

marianogonzalez: This breaks backwards compatibility

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/OperationReturnTypeModelValidator.java
ndinu: Do you suggest to change these errors for warnings? Or do you see viable to fail when compiling?

ndinu (approved):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/TypeWrapper.java
```diff
@@ -193,7 +193,7 @@ public boolean isSameType(Type type) {
 
   @Override
   public boolean isSameType(Class<?> clazz) {
-    return aClass.equals(clazz);
+    return aClass != null && aClass.equals(clazz);
```
marianogonzalez, line 196: how could this ever happen? How would somebody create a type wrapper around a null class?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/OperationReturnTypeModelValidator.java
```diff
@@ -123,4 +143,13 @@ private IllegalModelDefinitionException missingReturnTypeException(ExtensionMode
     throw new IllegalOperationModelDefinitionException(format("Operation '%s' in Extension '%s' is missing a return type",
                                                               operationModel.getName(), model.getName()));
   }
+
+  private void validateGenerics(ExtensionModel extensionModel, NamedObject namedObject, OperationModel operationModel,
```
marianogonzalez, line 147: why is the `namedObject` parameter needed? Seems redundant. This method should operate over the operationModel entirely

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/validation/OperationReturnTypeModelValidator.java
Bardammu: In the case of the CompletionCallback it also operates over the operation's extensionParameter. I can use the operationModel on the ProblemReporter but I didn't want to change how it was being used(when there is no generics for example).

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/TypeWrapper.java
Bardammu: It happens when the type is `?`. If I don't make this check I'll get NPE when doing `generics.get(0).getConcreteType().isSameType(Void.class`

marianogonzalez: but then shouldn't the same check be added everywhere? Why not using `Object.class` so that it acts as a NullObject pattern?

Bardammu: The same check it's now in the others methods:
`public boolean isAssignableTo(Class<?> clazz)` and `public boolean isAssignableFrom(Class<?> clazz)`
It looks like someone forgot to add it to `public boolean isSameType(Class<?> clazz)`

I guess it's possible to assign `Object.class` to `aClass` if the constructor gets a `null`, do you think it's a better solution?

marianogonzalez: the check is indeed present in those other methods, but it's not present in all the rest, meaning that there's a potential NPE if any other change starts invoking a method that's not being invoked right now. I think we should completely prevent the NPE situation altogether and using a NullObject is both the easiest and safest way of doing that.

marianogonzalez (approved):

PR #10388, opened 2021-06-01
CMTS-217: Enable OS MetadataCacheManager for Metadata on DeclarationS…

gsfernandes:
…ession

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/metadata/cache/DefaultPersistentMetadataCacheManager.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/metadata/cache/DefaultPersistentMetadataCacheManager.java
@@ -96,10 +96,11 @@ public class DefaultPersistentMetadataCacheManager implements MetadataCacheManag
     return withKeyLock(id, key -> {
       try {
         if (metadataStore.get().contains(key)) {
+          LOGGER.debug(format("Retrieving cache from OS with ID '%s'", id));
           return metadataStore.get().retrieve(key);
         }
 
-        LOGGER.debug("Creating new cache " + id);
+        LOGGER.debug(format("Creating new cache in OS with ID '%s'", id));
         DefaultMetadataCache metadataCache = new DefaultMetadataCache();
         metadataStore.get().store(key, metadataCache);
         return metadataCache;
@@ -117,7 +118,7 @@ public class DefaultPersistentMetadataCacheManager implements MetadataCacheManag
   public void updateCache(String id, MetadataCache cache) {
     withKeyLock(id, key -> {
       try {
-        LOGGER.debug("updateCache Key: " + id);
+        LOGGER.debug(format("Updating cache in OS with ID '%s'", id));
         if (metadataStore.get().contains(key)) {
           metadataStore.get().remove(key);
         }
@@ -139,6 +140,7 @@ public class DefaultPersistentMetadataCacheManager implements MetadataCacheManag
         if (isBlank(keyHash)) {
           clearMetadataCaches();
         } else {
+          LOGGER.debug(format("Removing cache in OS with ID '%s'", key));
           metadataStore.get().remove(key);
         }
       } catch (ObjectDoesNotExistException e) {
@@ -177,6 +179,7 @@ public class DefaultPersistentMetadataCacheManager implements MetadataCacheManag
 
   private void clearMetadataCaches() {
     try {
+      LOGGER.debug("Clearing cache from OS");
       metadataStore.get().clear();
     } catch (ObjectStoreException e) {
       String msg = format("An error occurred while clearing MetadataCaches: %s", e.getMessage());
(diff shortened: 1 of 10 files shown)


modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/api/artifact/DeclarationSession.java
```diff
@@ -77,6 +77,14 @@
    */
   MetadataResult<ComponentMetadataTypesDescriptor> resolveComponentMetadata(ComponentElementDeclaration component);
 
+  /**
+   * Disposes the {@link org.mule.runtime.api.metadata.MetadataCache} associated to the component metadata resolution which is
+   * provided in order to store resources cross resolution by resolvers.
+   *
+   * @param component the component whose has a dynamic metadata types.
+   */
+  void disposeMetadataCache(ComponentElementDeclaration component);
```
lucianoRM, line 86: Shouldn't it be at the ToolingClient level? If the cache is shared between sessions

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/utils/ArtifactHelper.java
```diff
@@ -108,7 +110,16 @@ public void visit(ConfigurationElementDeclaration declaration) {
   }
 
   public Optional<ConfigurationInstance> getConfigurationInstance(String configName) {
-    return findConfigurationProvider(configName).map(cp -> cp.get(getNullEvent()));
+    return findConfigurationProvider(configName).map(cp -> {
```
lucianoRM, line 113: Why is this?

lucianoRM (changes requested):

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/api/artifact/DeclarationSession.java
gsfernandes: Yes, it is shared between session but cannot be accessed without a Session :(.
Why? Because we need to get the key from component, either by its resolver or by its configuration/connection data.

gsfernandes:
--validate

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/utils/ArtifactHelper.java
elrodro83: correct, creating an event is not something to be done lightly and it must be properly disposed/finished afterwards.

core/src/main/java/org/mule/runtime/core/internal/metadata/cache/DefaultPersistentMetadataCacheManager.java
```diff
@@ -139,6 +140,7 @@ public void dispose(String keyHash) {
         if (isBlank(keyHash)) {
           clearMetadataCaches();
         } else {
+          LOGGER.debug(format("Removing cache in OS with ID '%s'", key));
```
lucianoRM, line 143: Check if debug level is enabled in the logger before logging.  Also, is string formatting necessary or does the .debug method accept extra arguments for formatting?

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/config/DefaultDeclarationSessionBuilder.java
```diff
@@ -41,9 +44,14 @@ public DefaultDeclarationSessionBuilder(DefaultApplicationFactory defaultApplica
              String.valueOf(!valueOf(getProperty(MULE_FORCE_TOOLING_APP_LOGS_DEPLOYMENT_PROPERTY, "false"))))
         .put(MULE_LAZY_INIT_DEPLOYMENT_PROPERTY, TRUE)
         .put(MULE_LAZY_INIT_ENABLE_DSL_DECLARATION_VALIDATIONS_DEPLOYMENT_PROPERTY, TRUE)
+        .put(SHARED_PARTITIONED_PERSISTENT_OBJECT_STORE_PATH, getToolingWorkingDir().getAbsolutePath())
         .build();
   }
 
+  private File getToolingWorkingDir() {
+    return new File(getExecutionFolder(), "tooling");
```
lucianoRM, line 52: Does this work if "tooling" folder does not exist?

lucianoRM (changes requested):

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/config/DefaultDeclarationSessionBuilder.java
gsfernandes: Yes, it will create the directory.

core/src/main/java/org/mule/runtime/core/internal/metadata/cache/DefaultPersistentMetadataCacheManager.java
gsfernandes: Sure, I forgot to check if it debug severity is enabled. I would do the changes.

lucianoRM (dismissed):

lucianoRM (approved):

PR #10391, opened 2021-06-02
MULE-17309: Fixing description for Until Successful construct.

asanguinetti:
Where it previously said "asynchronous" now it says "synchronous".

As can be seen in the commits, the same string appeared in different places. One of them seems to be a candidate for removal. I'm proposing this change in [MULE-19462](https://www.mulesoft.org/jira/browse/MULE-19462).

### DIFF
--- a/modules/extensions-support/src/test/resources/mule-extension-model.json
+++ b/modules/extensions-support/src/test/resources/mule-extension-model.json
@@ -2704,7 +2704,7 @@
         }
       ],
       "name": "untilSuccessful",
-      "description": "Attempts to route a message to the message processor it contains in a synchronous manner. Routing is considered successful if no exception has been raised and, optionally, if the response matches an expression",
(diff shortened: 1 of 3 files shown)


modules/extensions-support/src/test/resources/mule-extension-model.json
```diff
@@ -2704,7 +2704,7 @@
         }
       ],
       "name": "untilSuccessful",
-      "description": "Attempts to route a message to the message processor it contains in an asynchronous manner. Routing is considered successful if no exception has been raised and, optionally, if the response matches an expression",
+      "description": "Attempts to route a message to the message processor it contains in a synchronous manner. Routing is considered successful if no exception has been raised and, optionally, if the response matches an expression",
```
elrodro83, line 2707: `to the message processor it contains ` -> `to its inner chain `

elrodro83 (changes requested):

elrodro83 (approved):

PR #10397, opened 2021-06-03
CCNS-33: Allow to resolve values for an object field using DeclarationSession API

lucianoRM:


modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/AbstractParameterResolverExecutor.java
```diff
@@ -93,11 +101,26 @@ protected ParameterValueResolver parameterValueResolver(ParameterizedElementDecl
...
+        // TODO: CCNS-26. Improve this.
+        if (value instanceof String && isExpression(value)) {
+
+          if (!parameterModel.getExpressionSupport().equals(NOT_SUPPORTED) && !overrideExpressionSupport) {
+            throw new ExpressionNotSupportedException(format("Error resolving value for parameter: '%s' from declaration, it cannot be an EXPRESSION value",
+                                                             parameterName));
+          }
+
+          value = new TypeSafeExpressionValueResolver<>((String) value,
+                                                        TypedValue.class,
```
gsfernandes, line 116: mmm, I don't think this is correct. I would assume that here it has to go directly the class implementation from the connector's code definition. Even it is weird to see here a TypedValue.class.  Could you explain this change?

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/AbstractParameterResolverExecutor.java
```diff
@@ -73,7 +80,8 @@ protected ParameterValueResolver parameterValueResolver(ParameterizedElementDecl
   }
 
   protected Map<String, Object> parametersMap(ParameterizedElementDeclaration parameterizedElementDeclaration,
-                                              ParameterizedModel parameterizedModel)
+                                              ParameterizedModel parameterizedModel,
+                                              boolean overrideExpressionSupport)
```
gsfernandes, line 84: I would say that if we go with expressions for ValueProvider body level, we would need it for SampleData too and later for Metadata resolution. Therefore, I would say this by default the new behaviour.

gsfernandes (changes requested):

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/AbstractParameterResolverExecutor.java
lucianoRM: This is to avoid having to go through the class. Also, keep in mind that this is going to be used as input for resolving another expression. 
Let's say that the class defined in the connector is a String, and that the expression received resolves to a JsonObject. If we resolve this expression directly to a String, when using it as input for the expression that extracts it's fields, that fails. That is becuase DW will have an string as input and is not able to detect it as a Json representation

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/AbstractParameterResolverExecutor.java
lucianoRM: Maybe we can change that in the future. I would not want to accept expressions for ValueProviders yet

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/AbstractParameterResolverExecutor.java
gsfernandes: We have discussed this with MA and decided that we could represent input as DW in order to create the value resolver. Same approach could be used in order to resolve either acting parameters for cache or ask DW to generated a minimal representation of a DW script based on the extraction expressions that we would be using for getting acting parameter values.

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/AbstractParameterResolverExecutor.java
gsfernandes: Ok, makes sense.

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/AbstractParameterResolverExecutor.java
```diff
@@ -93,11 +99,21 @@ protected ParameterValueResolver parameterValueResolver(ParameterizedElementDecl
...
                                                                             parameterName, parameterGroupName)));
-        Object value = extractValue(parameterElement.getValue(),
-                                    artifactHelper.getParameterClass(parameterModel, parameterizedElementDeclaration));
-        if (!parameterModel.getExpressionSupport().equals(NOT_SUPPORTED) && isExpression(value)) {
-          throw new ExpressionNotSupportedException(format("Error resolving value for parameter: '%s' from declaration, it cannot be an EXPRESSION value",
-                                                           parameterName));
+        Class<?> parameterClass = artifactHelper.getParameterClass(parameterModel, parameterizedElementDeclaration);
+        Object value = extractValue(parameterElement.getValue(), parameterClass);
+
+        // TODO: CCNS-26. Improve this.
```
gsfernandes, line 105: Do we still need this task/TODO?

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/params/ParameterExtractor.java
```diff
@@ -36,7 +37,12 @@
 
   private Object value;
 
-  public static <T> T extractValue(ParameterValue parameterValue, Class<T> type) {
+  // TODO: CCNS-26. This approach is no longer valid. e.g: If the expected type is an stream, this fails.
```
gsfernandes, line 40: Should we use DW instead?

gsfernandes: Add a test (Ignored with CCNS-26)

gsfernandes (changes requested):

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/AbstractParameterResolverExecutor.java
lucianoRM: As discussed, we do

modules/tooling-support/src/main/java/org/mule/runtime/module/tooling/internal/artifact/AbstractParameterResolverExecutor.java
lucianoRM: Done in the next commit

gsfernandes (approved):

lucianoRM:
--validate

PR #10400, opened 2021-06-04
MULE-19470: MavenBundleDescriptorLoader doesn't resolve properties in pom

elrodro83:


eze210 (approved):

gsfernandes (changes requested):
This is not needed, a change in how MTS loads the extension models has been made in order to avoid reading these values and get them from the dependency resolution context.
This change won't work when pom has not a default value for properties:
https://github.com/mulesoft/mule-runtime-tooling-client/pull/1892

PR #10419, opened 2021-06-15
CCNS-26: Change ValueProviderExecutor to resolve parameter as expected by SDK Mediator

lucianoRM:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/InjectableParameterResolver.java
```diff
@@ -65,25 +65,18 @@ public InjectableParameterResolver(ParameterizedModel parameterizedModel,
...
-                     parameterName);
-      }
-
+    try {
+      parameterValue = expressionManager
+          .evaluate("#[ payload." + parameterName + "]",
+                    DataType.fromType(getType(injectableParameterInfo.getType())),
+                    expressionResolvingContext)
+          .getValue();
+    } catch (IllegalArgumentException e) {
```
marianogonzalez, line 76: shouldn't this just catch any exception?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/InjectableParameterResolver.java
```diff
@@ -65,25 +65,18 @@ public InjectableParameterResolver(ParameterizedModel parameterizedModel,
...
+      LOGGER.debug("Transformation of injectable parameter '{}' failed, the same value of the resolution will be used.",
```
marianogonzalez, line 77: append the exception message and add it as a cause

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/InjectableParameterResolver.java
lucianoRM: I just tried to mimic the previous behavior. Before we only catched a ClassNotFoundException and returned null in that case. But any other exception was re thrown. 
But you are right that this IllegalArgumentException could make no sense.

marianogonzalez (dismissed):

lucianoRM:
--validate

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/value/ValueProviderUtilsTestCase.java
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  @Test
+  public void sanitizeExpressionWithAttribute() {
+    String sanitized = sanitizeExpression("one.two.@three");
+    assertThat(sanitized, equalTo("one_.'two'.@'three'"));
+  }
+
+  @Test
+  public void sanitizeAlreadySanitizedExpression() {
```
ndinu, line 36: What would happen if we sanitize object."field.with.dot" ?

lucianoRM:
--validate

lucianoRM:
--validate

ndinu (approved):

marianogonzalez (changes requested):
COmmit needs to point to a MULE issue

lucianoRM:
Done!

marianogonzalez (approved):

lucianoRM:
--validate

lucianoRM:
--validate

lucianoRM:
--validate

PR #10422, opened 2021-06-16
MULE-19485: Error is not correctly handled by Source Policy when execute-next is within a subflow

mbuchwald:


mbuchwald:
--validate

core/src/main/java/org/mule/runtime/core/internal/policy/PolicyNextActionMessageProcessor.java
```diff
@@ -103,30 +108,48 @@ public CoreEvent process(CoreEvent event) throws MuleException {
   private OnExecuteNextErrorConsumer errorConsumer(PolicyEventMapper policyEventMapper,
                                                    PolicyNotificationHelper notificationHelper) {
 
-    if (isWithinSourcePolicy(getLocation())) {
```
elrodro83, line 106: if you remove this pre-check you will add complexity to every error handling, affecting perfromance

elrodro83 (changes requested):

core/src/main/java/org/mule/runtime/core/internal/policy/PolicyNextActionMessageProcessor.java
```diff
@@ -103,30 +108,48 @@ public CoreEvent process(CoreEvent event) throws MuleException {
...
+    List<FlowStackElement> elements = me.getEvent().getFlowCallStack().getElements();
+    if (elements.size() != 1) {
+      return false;
+    }
+
+    Location completeLocation = builderFromStringRepresentation(elements.get(0).getProcessorPath().split(" ")[0]).build();
+    List<String> parts = completeLocation.getParts();
+    boolean isSubflowWithinASource = completeLocation.getGlobalName().equals(getRootContainerLocation().toString())
+        && parts.size() >= 1 && parts.get(0).equals(SOURCE_POLICY_PART_IDENTIFIER);
+    return isSubflowWithinASource;
```
elrodro83, line 152: return directly, without aux var

elrodro83 (changes requested):

core/src/main/java/org/mule/runtime/core/internal/policy/PolicyNextActionMessageProcessor.java
mbuchwald: I added the variable just for legibility. I can remove it, tho

elrodro83 (approved):

PR #10431, opened 2021-06-17
EE-7789: Oauth Token Expiration - refresh dance sends redirect_uri pa…

sofiamorseletto:
…rameter in the query

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/oauth/authcode/AuthorizationCodeOAuthHandler.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/oauth/authcode/AuthorizationCodeOAuthHandler.java
@@ -10,6 +10,7 @@ import static java.lang.String.format;
 import static java.util.Collections.emptyList;
 import static java.util.Optional.empty;
 import static java.util.Optional.of;
+import static org.mule.runtime.api.config.MuleRuntimeFeature.NO_OAUTH_REDIRECT_URI;
 import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
 import static org.mule.runtime.api.util.Preconditions.checkArgument;
 import static org.mule.runtime.core.api.event.EventContextFactory.create;
@@ -21,6 +22,7 @@ import static org.mule.runtime.module.extension.internal.runtime.connectivity.oa
 import static reactor.core.publisher.Mono.from;
 
 import org.mule.runtime.api.artifact.Registry;
+import org.mule.runtime.api.config.FeatureFlaggingService;
 import org.mule.runtime.api.exception.MuleException;
 import org.mule.runtime.api.exception.MuleRuntimeException;
 import org.mule.runtime.api.lifecycle.InitialisationException;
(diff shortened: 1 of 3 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/oauth/authcode/AuthorizationCodeOAuthHandler.java
```diff
@@ -197,6 +205,11 @@ private AuthorizationCodeOAuthDancer createDancer(AuthorizationCodeConfig config
         .customParameters(config.getCustomParameters())
         .customParametersExtractorsExprs(getParameterExtractors(config));
 
+    if (!featureFlaggingService.isEnabled(NO_OAUTH_REDIRECT_URI)) {
```
marianogonzalez, line 208: I don't think the fix should be applied here but in the AuthorizationCodeDancer in the oauth-service. By applying it here, the issue still persists if somebody tries to consume the Dropbox API directly using the Http Connector. The AuthorizationCodeDancerBuilder should expose a method that allows us to configure this behavior. This class should merely use the feature flag to determine how to configure that. At the same time, the dancer's default behavior should also depend on the feature flag

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/connectivity/oauth/authcode/AuthorizationCodeOAuthHandler.java
sofiamorseletto: Hi, Mariano! A change has also been made on the `DefaultAuthorizationCodeOAuthDancer` of the Oauth-client, you can check it on this PR: https://github.com/mulesoft/mule-oauth-client/pull/69 . If no additional parameters are added the client does not send the redirect_uri parameter when refreshing the token. So even if somebody tries to consume the Dropbox API using the HTTP Connector, the parameters wouldn't have been added. Does this PR make sense to you considering also the change on https://github.com/mulesoft/mule-oauth-client/pull/69 ?

marianogonzalez: Hello Sofia. I didn't get the notification for that other PR, thanks for sending it over. I added some additional comments on that one too. I still don't think this is the right solution. Actually, now that I think about it I see additional problems with this solution:

* The expectation when using a cloud connector is that it works out of the box. This shouldn't depend on a feature flag. The only reason the feature flag is actually needed is the solution not being backwards compatible. It should be up to each connector to specify whether the redirect_uri should be provided or not. The runtime cannot take that decision on its own.
* This solution is that it wouldn't work if the Dropbox connector were to be used alongside another one which does need the redirect_uri parameter, as the feature flag would affect one of the connectors all the time.

marianogonzalez (changes requested):
The jira issue should be moved to the MULE project as the fix is going on the open source modules and the EE project is not public

PR #10432, opened 2021-06-17
MULE-19527: When creating a ValueProvider if the actingParameters have a DW keyword it would fail to create the ValueProvider

ndinu:


gsfernandes (dismissed):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/InjectableParameterResolver.java
```diff
@@ -130,29 +128,54 @@ private BindingContext createBindingContext(ParameterValueResolver parameterValu
...
-                              + getWithBindingPrefix(injectableParameterInfo.getExtractionExpression()))
+                              + BINDING_IDENTIFIER_ACCESS
+                              + sanitizeExpression(injectableParameterInfo.getExtractionExpression()))
                           .collect(Collectors.joining(", ")));
     expression.append(EXPRESSION_SUFFIX);
 
     return expression.toString();
   }
 
+  private String sanitizeExpression(String extractionExpression) {
```
gsfernandes, line 162: I was wondering if this would work with "attributes" but I see it would:
```account.@type``` is how an attribute is accessed in DW.

gsfernandes (approved):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/InjectableParameterResolver.java
ndinu: Good catch, I have just tried this. The extraction expression will have to be `account.@"type"`.

ndinu: I think that needs to be taken into account from the connector, since the SDK asks for a DW selector for the value and using type is not valid. I think the best effort made for the component parameter name is enough.

gsfernandes: Ok, makes sense. Nice work for documentation

PR #10457, opened 2021-06-30
Avoid parsers from configuring specific logic for textBody parameters

elrodro83:
* Impact analysis

### DIFF
--- a/modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/spring/ComponentConfigurationBuilder.java
+++ b/modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/spring/ComponentConfigurationBuilder.java
@@ -14,7 +14,6 @@ import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toMap;
 import static java.util.stream.Stream.concat;
 import static org.mule.runtime.api.util.MuleSystemProperties.DEFAULT_SCHEDULER_FIXED_FREQUENCY;
-import static org.mule.runtime.ast.api.ComponentAst.BODY_RAW_PARAM_NAME;
 import static org.mule.runtime.config.internal.dsl.spring.CommonComponentBeanDefinitionCreator.areMatchingTypes;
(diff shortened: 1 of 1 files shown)


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ExtensionDefinitionParser.java
```diff
@@ -674,31 +598,18 @@ protected ClassLoader getContextClassLoader() {
...
-  }
-
   /**
    * Registers a definition for parsing the given {@code parameterModel} as an element attribute
    *
    * @param parameterModel a {@link ParameterModel}
    * @return an {@link AttributeDefinition.Builder}
    */
-  protected AttributeDefinition.Builder parseAttributeParameter(ParameterModel parameterModel) {
+  protected AttributeDefinition.Builder parseSimpleParameter(ParameterModel parameterModel) {
```
marianogonzalez, line 607: just because the parameter is being parsed from an attribute, it doesn't mean it's simple (or at least what the SDK referts to as a simple parameter)

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ExtensionDefinitionParser.java
```diff
@@ -296,35 +263,14 @@ public void visitAnyType(AnyType anyType) {
...
 
         @Override
         public void visitArrayType(ArrayType arrayType) {
-          if (!parseAsContent(arrayType)) {
-            parseCollectionParameter(parameter, arrayType, paramDsl);
-          }
-        }
-
-        private boolean parseAsContent(MetadataType type) {
-          if (isContent) {
```
marianogonzalez, line 314: where is the code that replaces this behavior?

marianogonzalez (changes requested):

PR #10458, opened 2021-06-30
MULE-19427:When using cluster in mule3 and a RoutingException is rais… 

jcgrethe:
…ed by the SFTP transport, the route and failingMessageProcessor cannot be serialized.

### DIFF
--- a/core/src/main/java/org/mule/api/MessagingException.java
+++ b/core/src/main/java/org/mule/api/MessagingException.java
@@ -11,8 +11,6 @@ import static java.lang.String.format;
 import static org.apache.commons.lang.StringUtils.abbreviate;
 import static org.mule.util.ClassUtils.isConsumable;
 
-import org.apache.commons.lang.SerializationException;
-import org.apache.commons.lang.SerializationUtils;
 import org.mule.VoidMuleEvent;
 import org.mule.api.processor.MessageProcessor;
 import org.mule.config.DefaultMuleConfiguration;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mule/api/MessagingException.java
```diff
@@ -20,10 +22,8 @@
 import org.mule.routing.filters.WildcardFilter;
 import org.mule.transport.NullPayload;
 
-import java.io.IOException;
-import java.io.ObjectInputStream;
-import java.io.ObjectOutputStream;
-import java.io.Serializable;
+import java.io.*;
```
eze210, line 25: Import what you use

core/src/main/java/org/mule/api/MessagingException.java
```diff
@@ -390,17 +390,34 @@ protected void extractMuleMessage(MuleEvent event)
     private void writeObject(ObjectOutputStream out) throws Exception
     {
         out.defaultWriteObject();
-        if (this.failingMessageProcessor instanceof Serializable)
-        {
+        if (isSerializable(this.failingMessageProcessor)) {
```
eze210, line 393: Brackets in new line

core/src/main/java/org/mule/api/MessagingException.java
```diff
@@ -390,17 +390,34 @@ protected void extractMuleMessage(MuleEvent event)
...
+        if (!(object instanceof Serializable)) {
+            return false;
+        }
+
+        Serializable serializableObject = (Serializable) object;
+        try {
+            Field[] fields = serializableObject.getClass().getDeclaredFields();
+            for (Field field : fields) {
+                field.setAccessible(true);
+                if (!(field.get(serializableObject) instanceof Serializable)) {
```
eze210, line 411: Shouldn't this be a recursive call?

eze210: Also, add tests for the recursive case (serializable with a serializable field that has a non-serializable field)

eze210 (changes requested):

eze210 (approved):

PR #10463, opened 2021-07-06
Mule 19137

Bardammu:


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/resolver/ValueResolverFactory.java
```diff
@@ -51,31 +56,38 @@ public ValueResolverFactory(DslSyntaxResolver dslResolver) {
...
                                  ExpressionSupport expressionSupport, boolean required,
                                  Set<ModelProperty> modelProperties,
                                  boolean acceptsReferences, boolean content) {
     if (value instanceof ValueResolver) {
       return (ValueResolver<T>) value;
     }
 
-    ValueResolver<T> resolver;
+    ClassLoader extensionClassLoader = getClassLoader(extensionModel);
+    CompositeClassLoader compositeClassLoader = from(extensionClassLoader, Thread.currentThread().getContextClassLoader());
```
marianogonzalez, line 87: by doing this here you're only fixing the ValueProvider case. However, the jira ticket does say that all ValueProvider implementations should be reviewed looking for other potential occurrences of this problem.

Where else is `ExtensionMetadataTypeUtils.getType` used? Can we do this at a point in which we can be sure this doens't happen in any ValueProvider or in featuers such as sample data, DataSense, etc?

marianogonzalez (changes requested):

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/resolver/ValueResolverFactory.java
Bardammu: Which value provider implementation? I’ve reviewed what I found in the uber and didn’t find any other problem.
Am I ignoring something?

ExtensionMetadataTypeUtils.getType It’s used in several places , for example:

NameClashModelValidator
ContentParameterModelValidator
SubtypesModelValidator
MetadataComponentModelValidator

SchedulingStrategyParsingDelegate
CharsetValueResolverParsingDelegate
ReflectiveExpressionFunctionExecutor
ConfigOverrideValueResolverWrapper
CustomStaticTypeDeclarationEnricher
ModelLoaderUtils

We can’t force the use of extension classloader there.

marianogonzalez: Any ValueResolver which produces a value of a type which shouldn't necessarily be exported has the potential to have the same problem. For example:

* Anything that uses an `ObjectBuilder`
* Anything which could resolve an enum value (e.g: resolving acting parameters for SampleDataProviders or TypeResolvers, the `StringToEnum` transformer, from your list `ReflectiveExpressionFunctionExecutor` is a candidate)
* If `ModelLoaderUtils` uses this class, who else uses `ModelLoaderUtils`?

It doesn't make sense to go do the CL switch on each and everyone of these components. We should also do a best effort to do this at a high level place which minimizes the possibility of missed cases which translate in future bugs. What the issue ask is to not only fix this specific problem but to also evaluate other situations which might force to export packages unnecessarily

gsfernandes: Hi guys! After debugging this and adding some comments to the Jira's issue I found you have had this conversation and changes already proposed here.
I do agree with MG that the change in ValueResolverFactory is a no go but on the other hand I see that the parser defintion from SDK is assuming that the TCCL is always the expected one. See for instance:
```        new ObjectTypeParameterParser(definitionBuilder, elementName, elementNamespace, type, getContextClassLoader(),
                                      dslResolver, parsingContext).parse()
                                          .forEach(this::addDefinition);
```
https://github.com/mulesoft/mule/blob/5e70c46b4b9be32e4088ffbd42b361608578b1f0/modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ExtensionDefinitionParser.java#L818
This assumption is valid due to the BuildingDefinitionProvider is the one that sets the TCCL with the extension classloader:
https://github.com/mulesoft/mule/blob/478ec749401ba373df25ec9d208f3d70fe2d5116/modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/DefaultExtensionBuildingDefinitionProvider.java#L100

What is missed is that the invocation to the ValueResolverFactory happens after the parsers are defined and the TCCL at that point when this code is invoked is always the application/domain classloader.
So, I would propose to wrap calls to the valueResolverFactory.of(...) from the ExtensionDefinitionParser with the TCCL that is correctly set at the moment that the parser is building the definition. Be aware that most of the invocation/code to the valueResolverFactory is later executed as it is provided as a lambda function to the parsers defintions.
What do you think?

PR #10474, opened 2021-07-14
MULE-19581 - Add Polling Sources logging

dabreu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -301,8 +303,12 @@ public PollItemStatus accept(Consumer<PollItem<T, A>> consumer) {
         }
       }
 
+      String itemId = getItemId(pollItem);
       if (status != ACCEPTED || currentPollItemLimitApplied) {
+        LOGGER.debug(String.format(REJECTED_ITEM_MESSAGE, itemId, status));
```
ndinu, line 308: Instead of String.format and %s , use the {} place holder and pass the strings to the debug method, this way the formatting of the string only occurs when the logging happens.

ndinu (dismissed):

ndinu (approved):

PR #10480, opened 2021-07-15
MULE-19587: ComponentAst getParameter should be consistent with getParameters

eze210:


eze210:
--validate

eze210:
We'll avoid the filter instead

PR #10497, opened 2021-07-19
MULE-19601: Features are not reinserted after FeatureFlaggingRegistryTestCase

fsgonz:


core-tests/src/test/java/org/mule/runtime/core/api/config/FeatureFlaggingRegistryTestCase.java
```diff
@@ -25,26 +25,42 @@
...
+import java.util.function.Predicate;
+
 @SmallTest
 @Feature(DEPLOYMENT_CONFIGURATION)
 @Story(FEATURE_FLAGGING)
 public class FeatureFlaggingRegistryTestCase {
 
   private FeatureFlaggingRegistry featureFlaggingRegistry;
 
+  Map<org.mule.runtime.api.config.Feature, Predicate<MuleContext>> configurations = new HashMap<>();
```
elrodro83, line 43: use fqcn for allure and keep the actual test code as clean as possible

core-tests/src/test/java/org/mule/runtime/core/api/config/FeatureFlaggingRegistryTestCase.java
```diff
@@ -25,26 +25,42 @@
...
+  Map<org.mule.runtime.api.config.Feature, Predicate<FeatureContext>> featureFlagConfigurations = new HashMap<>();
+
   @Rule
   public ExpectedException expectedException = ExpectedException.none();
 
   @Before
   public void setUp() {
     featureFlaggingRegistry = FeatureFlaggingRegistry.getInstance();
+    // To reset the feature flags after the test.
+    featureFlaggingRegistry.getFeatureConfigurations().forEach((feature, condition) -> configurations.put(feature, condition));
```
elrodro83, line 53: is there no other tests where this may happen? is it worth extracting this to a Rule for reuse?

elrodro83 (changes requested):

core-tests/src/test/java/org/mule/runtime/core/api/config/FeatureFlaggingRegistryTestCase.java
fsgonz: This depends on the order of execution of the tests. The problem is that the maps in the feature flag services are static. This is the only tests where the feature flag registration is used. I don't think creating a rule is the correct way. This probably should be tackled by refactoring the feature flag registry.

elrodro83 (approved):

PR #10501, opened 2021-07-19
MULE-19607: Avoid setting inexisting infrastructure fileds on every component

elrodro83:


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/operation/OperationDefinitionParser.java
```diff
@@ -30,6 +40,31 @@ public OperationDefinitionParser(Builder definition, ExtensionModel extensionMod
     super(definition, extensionModel, operationModel, dslSyntaxResolver, parsingContext);
   }
 
+  @Override
+  protected Builder doParse(Builder definitionBuilder) throws ConfigurationException {
```
marianogonzalez, line 44: why this? what if we ever wanted to add this capability to a construct or a source?

marianogonzalez (changes requested):

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/operation/OperationDefinitionParser.java
```diff
@@ -30,6 +35,25 @@ public OperationDefinitionParser(Builder definition, ExtensionModel extensionMod
...
   }
 
+  @Override
+  protected boolean hasErrorMappingsGroup() {
+    return !getComponentModel().getModelProperty(NoErrorMappingModelProperty.class).isPresent();
+  }
+
+  @Override
+  protected boolean hasOutputGroup() {
+    AtomicBoolean isVoid = new AtomicBoolean();
```
marianogonzalez, line 45: not a big deal.. but overkill IMO. A simple instanceof would do in this case...

marianogonzalez (dismissed):

marianogonzalez (approved):

PR #10509, opened 2021-07-21
MULE-19616: Content attributes of dsl groups not set

elrodro83:


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ExtensionDefinitionParser.java
```diff
@@ -310,7 +310,7 @@ public void visitArrayType(ArrayType arrayType) {
 
         private boolean parseAsContent(MetadataType type) {
           if (isContent) {
-            parseAttributeParameter(parameter);
+            parseObjectParameter(parameter, paramDsl);
```
marianogonzalez, line 313: content parameters are not necesssarily objects. As a matter of fact, most of the times they're just streams. Method names should be reviewed

marianogonzalez (changes requested):

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ExtensionDefinitionParser.java
```diff
@@ -547,10 +549,11 @@ public void visitArrayType(ArrayType arrayType) {
 
       private boolean parseAsContent(boolean isContent, MetadataType type) {
         if (isContent) {
-          parseFromTextExpression(fieldName, fieldDsl.get(),
-                                  () -> value -> valueResolverFactory.of(fieldName, type, value, defaultValue,
-                                                                         expressionSupport, false, emptySet(),
-                                                                         false, isContent));
+          parseObjectParameter(fieldName, fieldName, type, fieldDsl.get(), defaultValue, expressionSupport, false, emptySet(),
```
marianogonzalez, line 552: misleading method name: this is no necessarily an object

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ExtensionDefinitionParser.java
```diff
@@ -738,12 +741,35 @@ protected void parseFromTextExpression(String key, DslElementSyntax paramDsl, Su
...
+                                                                                type,
+                                                                                value,
+                                                                                defaultValue,
+                                                                                expressionSupport,
+                                                                                required,
+                                                                                modelProperties,
+                                                                                !allowedStereotypes.isEmpty(),
+                                                                                content));
+    } else {
+      parseObjectParameter(key, name, (ObjectType) type,
```
marianogonzalez, line 760: shouldn't this one be renamed to remain consistent with the change in line 772?

marianogonzalez (changes requested):

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ExtensionDefinitionParser.java
elrodro83: actually, this one does roefer to an object type because this is in the `else` of an `if(content)`

marianogonzalez (approved):

PR #10518, opened 2021-07-22
MULE-19621: Source success callback parameters are set into the error

elrodro83:
callback as well


modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/validation/InvalidExtensionConfigTestCase.java
```diff
@@ -55,7 +55,7 @@ public void heisenbergDefaultConfigNegative() throws Exception {
     expectedException.expect(ConfigurationException.class);
     expectedException
         .expectMessage("[validation/heisenberg-default-illegal-config.xml:21]: "
-            + "Element <heisenberg:config> is missing required parameter 'knownAddresses'.");
+            + "Element <heisenberg:config> is missing required parameter 'enemies'.");
```
marianogonzalez, line 58: why changing this scenario instead of addding a new one?

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/validation/InvalidExtensionConfigTestCase.java
```diff
@@ -69,20 +69,20 @@ public void petStoreMissingRequiredParameterInsidePojo() throws Exception {
   }
 
   @Test
+  // TODO MULE-19350 Update this test
   public void operationWithExpressionConfigReference() throws Exception {
-    expectedException.expect(ConfigurationException.class);
+    expectedException.expect(InitialisationException.class);
```
marianogonzalez, line 74: doesn't this break compatibility?

modules/extensions-spring-support/src/test/resources/models/heisenberg.json
```diff
@@ -1648,7 +1648,7 @@
                 }
               },
               "hasDynamicType": false,
-              "required": true,
+              "required": false,
```
marianogonzalez, line 1651: don't change the existing components to fit your scenario. Please create a new component instead.

modules/extensions-spring-support/src/test/resources/validation/source-with-expression-config-ref.xml
```diff
@@ -12,18 +12,17 @@
...
                        dateOfDeath="2011-09-07T00:00:00-05:00"
                        money="0"
-                       recipe="recipes"
                        initialHealth="CANCER"
-                       finalHealth="CANCER"
-                       monthlyIncomes="monthlyIncomes"
-                       candidateDoors="candidateDoors"
-                       enemies="#[mel:[]]">
+                       finalHealth="CANCER">
+        <heisenberg:enemies/>
```
marianogonzalez, line 17: by doing this you're no longer testing what this tests verifies. There's already another test for the inlining case

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/model/PersonalInfo.java
```diff
@@ -60,6 +60,7 @@
   private Calendar dateOfGraduation;
 
   @Parameter
+  @Optional
```
marianogonzalez, line 63: add a new parametr instead

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/HeisenbergSource.java
```diff
@@ -90,6 +90,9 @@
   public static volatile boolean receivedInlineOnSuccess;
   public static volatile boolean receivedInlineOnError;
 
+  public static volatile PersonalInfo receivedInlineOnSuccessData;
```
marianogonzalez, line 93: where's the assertion that consumes these?

marianogonzalez (changes requested):

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/validation/InvalidExtensionConfigTestCase.java
elrodro83: this test was just an old test with the assertions added at some point. There was no relationship between the name of the test and the expected error.

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/validation/InvalidExtensionConfigTestCase.java
elrodro83: no, this is becuasoe of a change in the test, not from change in the code

modules/extensions-spring-support/src/test/resources/validation/source-with-expression-config-ref.xml
elrodro83: the test is called `source-with-expression-config-ref.xml`, what was being tested before has nothing to do.

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/model/PersonalInfo.java
elrodro83: if i add a new parameter intead of this, this would still be required and wouldmn't fit the scenario it tries to reproduce

tests/test-extensions/heisenberg-extension/src/main/java/org/mule/test/heisenberg/extension/HeisenbergSource.java
elrodro83: just added it for conssitency with the error callback

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/validation/InvalidExtensionConfigTestCase.java
marianogonzalez: No. If an invalid extension definition is provided, the current behavior is to throw `ConfigurationException`. This assertion means  that either behavior changed and a different exception is been thrown, or that this test cases is now failing for a different reason, which means the original scenario is no longer tested

marianogonzalez (dismissed):

marianogonzalez (approved):

marianogonzalez (approved):

PR #10537, opened 2021-07-26
MULE-19617: Disambiguate providerName when multiple multilevel FieldVPs are declared

ndinu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/ValueProvidersParameterDeclarationEnricher.java
```diff
@@ -248,6 +250,7 @@ private void enrichParameterFields(List<FieldValues> fieldsValues, ParameterDecl
           .ifPresent(field -> requiresConfiguration.set(true));
 
       int partOrder = 1;
+      String providerName = name + "_" + hash(fieldValues.targetSelectors());
```
lucianoRM, line 253: I know that it is a long shot but, what happens if there is a collision? 
Also, I would change the suffix to #hash or $hash, or directly <hash>, I think that it looks better. And, if possible, only use positive numbers.
Just a comment, not required for approval.

lucianoRM: I would externalize the logic of generating and handling providerNames with suffix so that it's together.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/value/ValueProviderMediator.java
```diff
@@ -263,7 +263,9 @@ public ValueProviderMediator(T containerModel, Supplier<MuleContext> muleContext
         parameterModels.add(parameterModel);
       } else if (parameterModel.getFieldValueProviderModels()
           .stream()
-          .anyMatch(fieldValueProviderModel -> fieldValueProviderModel.getProviderName().equals(valueName))) {
+          .anyMatch(fieldValueProviderModel -> fieldValueProviderModel.getProviderName()
+              .substring(0, fieldValueProviderModel.getProviderName().lastIndexOf("_")).equals(valueName))) {
```
lucianoRM, line 267: I would externalize the logic of generating and handling providerNames with suffix so that it's together.

lucianoRM (changes requested):

lucianoRM (dismissed):

lucianoRM (approved):

PR #10550, opened 2021-07-28
DEL-3317: Add the runtime descriptor generation mechanism on the distro assembly

garodriguezlp:
- Add todo to review mule assembly plugin config later

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -234,7 +234,7 @@
 
         <commonsCollectionsVersion>3.2.2</commonsCollectionsVersion>
         <groovyVersion>3.0.7</groovyVersion>
-        <assembly.verifier.version>1.4</assembly.verifier.version>
+        <assembly.verifier.version>2.0.0</assembly.verifier.version>
 
         <allureJunitVersion>2.8.1</allureJunitVersion>
         <multithreadedtcVersion>1.01</multithreadedtcVersion>
@@ -595,6 +595,8 @@
                         <skip>${skipVerifications}</skip>
                     </configuration>
 
+                    <!-- // @todo[question]: Given that the plugin now uses Groovy 3.0.8, should this be set here?-->
+                    <!-- ref: https://github.com/mulesoft/mule-assembly-verifier-plugin/blob/31e94d9efcf987b935191f3af17633147d7c8e43/pom.xml#L38-L42 -->
                     <dependencies>
                         <!--
                             Groovy dependency is handled by Mule differently,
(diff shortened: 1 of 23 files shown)


pom.xml
```diff
@@ -1583,6 +1585,7 @@
                         </dependency>
                     </dependencies>
 
+                    <!-- // @todo: Why is the execution configure here and not here: https://github.com/mulesoft/mule-ee-distributions/blob/e6370956039d68d2f9f2bfea00dab46bae685ad8/standalone/pom.xml#L138-->
```
aiannucci, line 1588: It should be configured both in mule-distributions and mule-ee-distributions

aiannucci (changes requested):

pom.xml
```diff
@@ -1571,6 +1571,8 @@
                         <skip>${skipVerifications}</skip>
                     </configuration>
 
+                    <!-- // @todo[question]: Given that the plugin now uses Groovy 3.0.8, should this be set here?-->
+                    <!-- ref: https://github.com/mulesoft/mule-assembly-verifier-plugin/blob/31e94d9efcf987b935191f3af17633147d7c8e43/pom.xml#L38-L42 -->
```
garodriguezlp, line 1575: @aiannucci Ale, did I dream it, our we talk about deleting this configuration dependency? If so, are we 100% sure this won't mess up with anything Groovy-related on Mule's side?

pom.xml
garodriguezlp: I agree on that the execution should be configured in the Maven module where it's used, but I wonder if this change could break other unknown to us usages, let's say in [`tooling`](https://github.com/mulesoft/mule-tooling-api/blob/master/pom.xml#L441-L445). BTW: I haven't check in `tooling` has an ancestry relation with `mule`

garodriguezlp: Ale, I just found that the execution configuration was moved from `mule-distributions` 14 days ago [take a look](https://github.com/mulesoft/mule-distributions/pull/629/files) to `mule` [take a look](https://github.com/mulesoft/mule/pull/10470/files) and also the groovy version override was part of that change.

The change we've been talking about is somehow the rollback of that change. Might I be missing something? Any idea on which was the motivation of [MULE-19578](https://www.mulesoft.org/jira/browse/MULE-19578) ?

pom.xml
garodriguezlp: [see](https://github.com/mulesoft/mule/pull/10550#discussion_r679568606)

pom.xml
aiannucci: Yeah, it seems that something is wrong there. It should be moved back to mule-distributions and mule-ee-distributions. Because `pluginManagement` is meant only for version handling.

pom.xml
aiannucci: Sure, we should delete it

anosenzo (approved):

aiannucci (approved):

PR #10585, opened 2021-08-02
MULE-19624: Inner chains of SDK operations started eagerly even if th…

jcgrethe:
…e containing flow has initialState=stopped


elrodro83 (changes requested):
This cherry pick is missing the test cases and the logging in ImmutableProcessorChainExecutor

elrodro83 (approved):

PR #10589, opened 2021-08-03
MULE-19682: Feature flags evaluation must ignore semver suffixes

IvanAndresFritzler:


core-tests/src/test/java/org/mule/runtime/core/api/config/FeatureContextTestCase.java
```diff
@@ -0,0 +1,41 @@
...
+import io.qameta.allure.Issue;
+import io.qameta.allure.Story;
+import org.junit.Test;
+
+@SmallTest
+@io.qameta.allure.Feature(DEPLOYMENT_CONFIGURATION)
+@Story(FEATURE_FLAGGING)
+public class FeatureContextTestCase {
+
+  private static final List<String> suffixedVersions = Arrays.asList("4.4.0-SNAPSHOT", "4.4.0-rc1", "4.4.0-rc1-SNAPSHOT");
```
elrodro83, line 29: make a test for each value instead of a test that verifies many scenarios

elrodro83: or make the test parameterized

core/src/main/java/org/mule/runtime/core/api/config/FeatureContext.java
```diff
@@ -26,7 +27,16 @@
 
   public FeatureContext(MuleVersion artifactMinMuleVersion, String artifactName) {
     this.artifactName = artifactName;
-    this.artifactMinMuleVersion = artifactMinMuleVersion;
+    // Feature flag evaluations must ignore suffixes
+    if (artifactMinMuleVersion != null) {
```
elrodro83, line 31: better to add this logic to `org.mule.runtime.api.meta.MuleVersion`, add an overloaded method with a flag to indicate to ignore the suffixes

elrodro83 (changes requested):

core/src/main/java/org/mule/runtime/core/api/config/FeatureContext.java
IvanAndresFritzler: I will add a method to the MuleVersion in order to get it without the suffixes, and adjust the feature flags predicates accordingly.

core/src/main/java/org/mule/runtime/core/internal/context/DefaultMuleContext.java
```diff
@@ -1199,7 +1199,7 @@ public void removeListener(MuleContextListener listener) {
   private static void configureSplitterExceptionHandlingFeature() {
     FeatureFlaggingRegistry featureFlaggingRegistry = FeatureFlaggingRegistry.getInstance();
     featureFlaggingRegistry.registerFeatureFlag(HANDLE_SPLITTER_EXCEPTION, featureContext -> featureContext
-        .getArtifactMinMuleVersion().filter(muleVersion -> muleVersion.atLeast("4.4.0")).isPresent());
+        .getArtifactMinMuleVersion().filter(muleVersion -> muleVersion.withoutSuffixes().atLeast("4.4.0")).isPresent());
```
elrodro83, line 1202: if the idea is that mule apps follow semantic versioning, then the muleVerison received in these callbacks must already be without suffixes. Otherwise you need to repeat the call to `withoutSuffixes()` on every usage.

elrodro83 (changes requested):

elrodro83 (approved):

PR #10617, opened 2021-08-09
MULE-19483: Since Mule 4.4 XML module would fail to generate the (#10…

elrodro83:
…616)

extension model if the tns:namespace is not present in schemaLocations


elrodro83:
--validate

mbuchwald (approved):

elrodro83:
--validate

PR #10626, opened 2021-08-11
MULE-19226: Policy not able to inject internal dependencies for artifacts that shares with the application

IvanAndresFritzler:


IvanAndresFritzler:
Closing this PR since it's not needed (a mistake has been made while triaging an issue, and this PR would not fix it)

PR #10634, opened 2021-08-12
MULE-19702: rootId is modified through a batch job when it should be the same as the correlationId

mbuchwald:


mbuchwald:
--validate

eze210 (approved):

PR #10639, opened 2021-08-17
MULE-19686: Provide an API to get a DefaultApplicationClassLoaderBuilder

elrodro83:
instance


elrodro83:
--validate

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/MuleArtifactResourcesRegistry.java
```diff
@@ -136,9 +136,8 @@ private MuleArtifactResourcesRegistry(ModuleRepository moduleRepository) {
     LicenseValidator licenseValidator = discoverLicenseValidator(currentThread().getContextClassLoader());
 
     domainManager = new DefaultDomainManager();
-    this.domainClassLoaderFactory = trackDeployableArtifactClassLoaderFactory(
-                                                                              new DomainClassLoaderFactory(containerClassLoader
-                                                                                  .getClassLoader(), new DefaultNativeLibraryFinderFactory()));
+    this.domainClassLoaderFactory =
+        trackDeployableArtifactClassLoaderFactory(domainClassLoaderFactory(name -> getAppDataFolder(name)));
```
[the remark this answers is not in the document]
fsgonz: Notice that this method (which is a default in DefaultNativeLibraryFinderFactory) can be replaced for the lambda reference, even if the previous suggestion is denied.

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/MuleArtifactResourcesRegistry.java
```diff
@@ -153,13 +152,12 @@ private MuleArtifactResourcesRegistry(ModuleRepository moduleRepository) {
...
+
     DeployableArtifactClassLoaderFactory<ApplicationDescriptor> applicationClassLoaderFactory =
-        trackDeployableArtifactClassLoaderFactory(new MuleApplicationClassLoaderFactory(new DefaultNativeLibraryFinderFactory()));
-    pluginDependenciesResolver = new BundlePluginDependenciesResolver(artifactPluginDescriptorFactory);
+        trackDeployableArtifactClassLoaderFactory(applicationClassLoaderFactory(name -> getAppDataFolder(name)));
     pluginClassLoadersFactory = new DefaultRegionPluginClassLoadersFactory(artifactPluginClassLoaderFactory, moduleRepository);
     applicationClassLoaderBuilderFactory =
-        new ApplicationClassLoaderBuilderFactory(applicationClassLoaderFactory, this.artifactPluginClassLoaderFactory,
-                                                 pluginClassLoadersFactory);
+        new ApplicationClassLoaderBuilderFactory(applicationClassLoaderFactory, pluginClassLoadersFactory);
```
fsgonz, line 160: Why in the case of api is directly the factory and the case of the deployable artifact classloader there is another layer (the provider)?

modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/api/builder/ApplicationClassLoaderBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * {@link ArtifactClassLoader} builder for class loaders required by {@link Application} artifacts
+ * 
+ * @since 4.5
+ */
+public interface ApplicationClassLoaderBuilder {
+
+  /**
+   * @param domain the domain artifact to which the application that is going to use this classloader belongs.
```
fsgonz, line 25: This jdoc has to be modified. It appears that the setDomain jdoc remains.

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/DefaultApplicationFactory.java
```diff
@@ -140,9 +142,9 @@ public Application createArtifact(ApplicationDescriptor descriptor) throws IOExc
         applicationClassLoaderBuilderFactory.createArtifactClassLoaderBuilder();
     MuleDeployableArtifactClassLoader applicationClassLoader =
         artifactClassLoaderBuilder
-            .setDomain(domain)
+            .setDomainParentClassLoader((ArtifactClassLoader) domain.getArtifactClassLoader().getClassLoader().getParent())
```
fsgonz, line 145: Is this logic something that should be resolved here or the builder should resolve it when build is invoked? See the comment regarding the jdoc method for the setDomainParentClassLoader which is not clear.

fsgonz: I know there is a problem described in the ArtifactClassloader#getClassloader jdoc method, but isn't there a way to obtain this through a new method (even if we have to add this to the api, so this cast is needed here or in the builder). Not sure about that.

modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/policy/DefaultPolicyTemplateFactoryTestCase.java
```diff
@@ -40,14 +39,15 @@
...
 
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
+import io.qameta.allure.Issue;
```
fsgonz, line 49: shouldn't the io import go before the org.junit imports?

modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/api/builder/ApplicationClassLoaderBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+   * @return the builder
+   */
+  ApplicationClassLoaderBuilder setDomainParentClassLoader(ArtifactClassLoader domainArtifactClassLoader);
+
+  /**
+   * @param artifactPluginDescriptors set of plugins descriptors that will be used by the application.
```
fsgonz, line 31: Maybe remove "set of" as these are varargs. "plugins descriptors that will be used by the application."

modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/api/builder/ApplicationClassLoaderBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * @param artifactDescriptor the descriptor of the artifact for which the class loader is going to be created.
+   * @return the builder
+   */
+  ApplicationClassLoaderBuilder setArtifactDescriptor(ArtifactDescriptor artifactDescriptor);
+
+  /**
+   * Creates a new {@code MuleDeployableArtifactClassLoader} using the provided configuration. It will create the proper class
+   * loader hierarchy and filters so application classes, resources, plugins and it's domain resources are resolve correctly.
```
fsgonz, line 44: so -> so that
it's -> its

modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/api/builder/ApplicationClassLoaderBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+   *
+   * @return a {@code MuleDeployableArtifactClassLoader} created from the provided configuration.
+   * @throws IOException exception cause when it was not possible to access the file provided as dependencies
```
fsgonz, line 47: "exception thrown when it is not possible to access the file provided as dependencies"

modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/api/builder/DomainClassLoaderBuilder.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * {@link ArtifactClassLoader} builder for class loaders required by {@link Domain} artifacts
+ *
+ * @since 4.5
+ */
+public interface DomainClassLoaderBuilder {
+
+  /**
+   * @param artifactPluginDescriptors set of plugins descriptors that will be used by the application.
```
fsgonz, line 23: maybe remove "set of" in the jdoc as indicated before (suggestion)

fsgonz (changes requested):

modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/api/plugin/resolver/DuplicateExportedPackageException.java
```diff
@@ -15,6 +15,8 @@
  */
 public class DuplicateExportedPackageException extends PluginResolutionError {
 
+  private static final long serialVersionUID = 1705370167352289806L;
```
fsgonz, line 18: same comment as before regarding the serialVersionUID.

modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/api/builder/ApplicationClassLoaderBuilder.java
elrodro83: will try to just remove the declared exception

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/MuleArtifactResourcesRegistry.java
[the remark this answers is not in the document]
elrodro83: thing is that the default only makes sense when within the runtime. The intention to this api is to expose this functionality outside of the runtime, in which case the proposed default wouldn't make sense.

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/application/DefaultApplicationFactory.java
elrodro83: the idea is for the api to be used in an external environemnt where you may have just the classloader for the domain, but not any of the domain structure. As this is, getting the classloader for a domain is an internal runtime thing to do, not something to bother external clients with.

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/MuleArtifactResourcesRegistry.java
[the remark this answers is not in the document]
fsgonz: Ok

fsgonz (approved):

PR #10641, opened 2021-08-17
MULE-19667: Create tests for the newly added Source Lifecycle logging

ndinu:


modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/source/PollingSourceWrapperTestCase.java
```diff
@@ -223,30 +223,6 @@ private void setComponentLocationMock() throws Exception {
...
-      }
-    };
-    doAnswer(answer).when(logger).debug(anyString(), (Object) any());
-    doAnswer(answer).when(logger).debug(anyString(), any(), any());
-    doAnswer(answer).when(logger).trace(anyString(), any(), any());
-    doAnswer(answer).when(logger).trace(anyString(), (Object[]) any());
-    return logger;
-  }
-
   private String formatMessage(String message, Object... args) {
```
dabreu, line 226: I think that this method could now be removed from this class, given that it is used on the LoggingTestUtils.java class, right?

ndinu: You are right, I am going to remove it.

dabreu (approved):

PR #10649, opened 2021-08-17
MULE-19500: Support new SDK API deprecated and dsl.xml packages

Bardammu:


ndinu (commented):
I think that the all the concepts from the enrichers should be added to the new extension parsers API, instead of asking for the annotations in the enrichers.

Bardammu:
> I think that the all the concepts from the enrichers should be added to the new extension parsers API, instead of asking for the annotations in the enrichers.

Is there any example of how to use the parsers from the enrichers?

ndinu:
> > I think that the all the concepts from the enrichers should be added to the new extension parsers API, instead of asking for the annotations in the enrichers.
> 
> Is there any example of how to use the parsers from the enrichers? Does the framework contemplate these cases?, if not we should address the issue in another task in my opinion.

IMO the enricher is deleted, the concept is added to the parser and the model loader is the one in charge of adding the information to the model.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/ConnectionProviderModelLoaderDelegate.java
```diff
@@ -42,6 +43,9 @@ void declareConnectionProviders(HasConnectionProviderDeclarer declarer, List<Con
           .withConnectionManagementType(parser.getConnectionManagementType())
           .supportsConnectivityTesting(parser.supportsConnectivityTesting());
 
+      ConnectionProviderDeclaration ConnectionProviderDeclaration = providerDeclarer.getDeclaration();
```
ndinu, line 46: start var name with lowercase

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/ExtensionModelParser.java
```diff
@@ -103,4 +105,14 @@
...
+  /**
+   * @return the extension's {@link DeprecationModel} if one was defined
+   */
+  Optional<DeprecationModel> getDeprecationModel();
+
+
+  /**
+   * @return the extension's {@link XmlDslModel}
+   */
+  XmlDslModel getXmlDslModel();
```
ndinu, line 117: I'm thinking if this method should provide only the information that the annotation provides instead, and the loader would create the XmlDslModel then.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaExtensionModelParser.java
```diff
@@ -136,4 +145,38 @@ public LicenseModelProperty getLicenseModelProperty() {
...
   }
+
+  @Override
+  public Optional<DeprecationModel> getDeprecationModel() {
+    return JavaExtensionModelParserUtils.getDeprecationModel(extensionElement);
+  }
+
+  @Override
+  public XmlDslModel getXmlDslModel() {
+    String extensionName = extensionElement.getName();
```
ndinu, line 156: Same comment as in the interface definition, maybe it should just provide the annotations information.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaExtensionModelParserUtils.java
```diff
@@ -213,6 +222,66 @@ private static void checkAnnotationsNotUsedMoreThanOnce(List<? extends Extension
     return connectionParameter;
   }
 
+  public static Optional<DeprecationModel> getDeprecationModel(ExtensionParameter extensionParameter) {
```
ndinu, line 225: Not saying that we need to do this now, but If we start adding all the parsing utils here maybe it will get a bit crowded. In the future maybe we create a util package here.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaFunctionModelParser.java
```diff
@@ -115,6 +121,36 @@ private void checkIsNotAnExtension() {
     }
   }
 
+  @Override
+  public Optional<DeprecationModel> getDeprecationModel() {
+    Optional<Method> method = functionElement.getMethod();
```
ndinu, line 126: Shouldn't we call the utils here?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaParameterModelParser.java
```diff
@@ -143,15 +150,65 @@ public ExpressionSupport getExpressionSupport() {
...
+        dslConfiguration = empty();
+      }
     }
 
     return dslConfiguration;
   }
 
+  @Override
+  public Optional<DeprecationModel> getDeprecationModel() {
+    Optional<Deprecated> legacyAnnotation = parameter.getAnnotation(Deprecated.class);
```
ndinu, line 182: Shouldn't we call the utils here?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaParameterModelParser.java
```diff
@@ -143,15 +150,65 @@ public ExpressionSupport getExpressionSupport() {
...
+      Optional<ParameterDsl> legacyAnnotation = parameter.getAnnotation(ParameterDsl.class);
+      Optional<org.mule.sdk.api.annotation.dsl.xml.ParameterDsl> sdkAnnotation =
+          parameter.getAnnotation(org.mule.sdk.api.annotation.dsl.xml.ParameterDsl.class);
+      if (legacyAnnotation.isPresent() && sdkAnnotation.isPresent()) {
+        throw new IllegalParameterModelDefinitionException(format("Parameter '%s' is annotated with '@%s' and '@%s' at the same time",
+                                                                  parameter.getName(),
+                                                                  ParameterDsl.class.getName(),
+                                                                  org.mule.sdk.api.annotation.dsl.xml.ParameterDsl.class
+                                                                      .getName()));
+      } else if (legacyAnnotation.isPresent()) {
```
ndinu, line 162: This is a personal opinion, but I think it is better to create a Pojo that contains the info from the annotation, and create the builder with the info from the pojo.

PR #10683, opened 2021-08-20
MULE-19687: Add a MulePluginDescriptorLoaderRepository to

elrodro83:
deployment-model-impl


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/artifact/MavenClassLoaderModelLoader.java
```diff
@@ -70,15 +71,15 @@ public ClassLoaderModel load(File artifactFile, Map<String, Object> attributes,
     long stamp = lock.readLock();
     try {
       MavenConfiguration updatedMavenConfiguration = getMavenConfig();
-      if (!mavenRuntimeConfig.equals(updatedMavenConfiguration)) {
+      if (mavenRuntimeConfig == null || !Objects.equal(mavenRuntimeConfig, updatedMavenConfiguration)) {
```
fsgonz, line 74: I think you can remove the null verificaton mavenRuntimeConfig == null. updatedMavenConfiguration will not be null, so it will enter whenever mavenRuntimeConfig is null (or there is another problem and mavenRuntimeConfig will be assigned null.

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/artifact/MavenClassLoaderModelLoader.java
```diff
@@ -70,15 +71,15 @@ public ClassLoaderModel load(File artifactFile, Map<String, Object> attributes,
...
         long writeStamp = lock.tryConvertToWriteLock(stamp);
         if (writeStamp == 0L) {
           lock.unlockRead(stamp);
           stamp = lock.writeLock();
         } else {
           stamp = writeStamp;
         }
-        if (!mavenRuntimeConfig.equals(updatedMavenConfiguration)) {
+        if (mavenRuntimeConfig == null || !Objects.equal(mavenRuntimeConfig, updatedMavenConfiguration)) {
```
fsgonz, line 82: same here.

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/plugin/PluginMavenClassLoaderModelLoader.java
```diff
@@ -40,17 +40,17 @@
...
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
-import com.google.common.collect.ImmutableList;
 import org.apache.maven.model.Model;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.collect.ImmutableList;
```
fsgonz, line 52: I think this change in order in the imports is not ok.

modules/deployment-model/src/main/java/org/mule/runtime/deployment/model/api/artifact/DescriptorLoaderRepositoryFactory.java
```diff
@@ -0,0 +1,28 @@
...
+import org.mule.runtime.module.artifact.api.descriptor.DescriptorLoaderRepository;
+
+/**
+ * Factory to create instances of {@link DescriptorLoaderRepository}.
+ * 
+ * @since 4.5
+ */
+public class DescriptorLoaderRepositoryFactory {
+
+  private final SpiServiceRegistry serviceRegistry = new SpiServiceRegistry();
```
fsgonz, line 20: I think this is more like a constant than a final instance field. Do you want to share this among all the DescriptorLoaderRepository's created by each factory but not for all the factories?

fsgonz (dismissed):

elrodro83:
--validate

fsgonz (approved):

PR #10719, opened 2021-08-27
MULE-18561:Adjust sdk-api implicit argument handling

ndinu:


ndinu:
--validate

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/RefNameDeclarationEnricher.java
```diff
@@ -54,8 +59,13 @@ protected ModelProperty getModelProperty(Field field) {
   }
 
   @Override
-  protected Class getAnnotation() {
-    return RefName.class;
+  protected Predicate<Field> getFieldHasAnnotationPredicate() {
+    return withAnnotation(RefName.class);
```
dabreu, line 63: shouldn't be checked the annotation org.mule.sdk.api.annotation.param.RefName here?

ndinu: This will be added when the `org.mule.sdk.api.annotation.param` package is implemented. For the scope of each task I think that it is the best to keep it as to what it is necessary for the task to be complemeted, and if a new case is taken into account, implement it extensively.

ndinu:
--validate

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/enricher/RefNameDeclarationEnricher.java
dabreu: got it, agree

dabreu (approved):

PR #10755, opened 2021-09-08
MULE-19724: Exclude mule-sdk-api from the extensions parent of the SDK

marianogonzalez:


ndinu (approved):

ndinu:
--validate

ndinu:
--validate

ndinu:
--validate

ndinu:
PR https://github.com/mulesoft/mule-extensions-parent/pull/160 was reverted so that the Hawk build would stop failing. This is because that PR also needed this PR merged. 

The validation in this PR was still failing with the changes in the parent, but since the parent repo does not run the Hawk it was able to be merged.

We need to revert that revert and merge it with this PR after the validation passes taking into account the exclusion (it may be green but the exclude is not in place right now)

PR #10757, opened 2021-09-08
DEL-3689: Fix old mule artifacts revapi versions pointing to SNAPSHOT…

anosenzo:
… after branching support/4.4.x

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -227,7 +227,7 @@
 
         <formatterConfigPath>formatter.xml</formatterConfigPath>
         <formatterGoal>validate</formatterGoal>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -227,7 +227,7 @@
 
         <formatterConfigPath>formatter.xml</formatterConfigPath>
         <formatterGoal>validate</formatterGoal>
-        <oldMuleArtifactVersion>4.4.0-SNAPSHOT</oldMuleArtifactVersion>
+        <oldMuleArtifactVersion>4.3.0</oldMuleArtifactVersion>
```
elrodro83, line 230: this should be 4.4 actually, right?

elrodro83 (changes requested):

elrodro83 (approved):

PR #10766, opened 2021-09-13
MULE-19146: Creating configResources from URL instead of InputStream so we have that information available on the AST.

asanguinetti:


asanguinetti:
--validate

elrodro83 (approved):

asanguinetti:
--validate

PR #10771, opened 2021-09-14
HTTPC-150: As a dev I want to be able to change the read timeout for the http listener

sofiamorseletto:


sofiamorseletto:
--validate

modules/http-api/src/main/java/org/mule/runtime/http/api/server/HttpServerConfiguration.java
```diff
@@ -27,16 +29,31 @@
   private final int connectionIdleTimeout;
   private final String name;
   private final Supplier<Scheduler> schedulerSupplier;
+  private final long readTimeout;
 
+  private static final long DEFAULT_READ_TIMEOUT_MILLIS = 30000L;
+
+  /**
+   * @deprecated since 4.5.0, use
```
eze210, line 37: 4.5.0, 4.4.1, 4.3.1? We need to discuss which is the version here.
Also, it's missing the `@since` annotation in the new methods

eze210 (changes requested):

modules/http-api/src/main/java/org/mule/runtime/http/api/server/HttpServerConfiguration.java
```diff
@@ -181,6 +187,7 @@ public Builder setName(String name) {
      *
      * @param readTimeout timeout value (in milliseconds)
      * @return this builder
+     * @since 4.3.1
```
eze210, line 190: 4.3.1, 4.4.1, 4.5.0

eze210 (changes requested):

eze210 (approved):

PR #10781, opened 2021-09-15
MULE-19678 - When a ConnectionException is thrown during source start…

dabreu:
…, invalidate the associated connection if present


dabreu:
--validate

ndinu (approved):

PR #10785, opened 2021-09-15
Mule 19508

Bardammu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaExtensionModelParserUtils.java
```diff
@@ -364,4 +407,32 @@ private static ExternalLibraryModel parseExternalLib(ExternalLib externalLibAnno
 
     return builder.build();
   }
+
+  private static <R extends Annotation, S extends Annotation, T> Optional<T> getAnnotationValueFromParameter(
```
ndinu, line 411: I would rename it to getAnnotationValueFromElement

ndinu (approved):

PR #10800, opened 2021-09-20
sdk-11: Logging show successful reconnection message when reconnection attempts are exhausted.

SebaElizalde:


SebaElizalde:
--validate

PR #10822, opened 2021-09-24
SDK-11: Logging show successful reconnection message when reconnection attempts are exhausted.

SebaElizalde:


modules/extensions-support/pom.xml
```diff
@@ -238,6 +238,12 @@
             <version>4.5.0-SNAPSHOT</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mule.runtime</groupId>
+            <artifactId>mule-core-tests</artifactId>
+            <version>4.5.0-SNAPSHOT</version>
```
marianogonzalez, line 244: use project.version

core/src/main/java/org/mule/runtime/core/internal/retry/policies/SimpleRetryPolicy.java
```diff
@@ -71,7 +71,7 @@ public SimpleRetryPolicy(long frequency, int retryCount) {
     net.jodah.failsafe.RetryPolicy<Object> actingPolicy = new net.jodah.failsafe.RetryPolicy<>()
         .handleIf(shouldRetry)
         .withMaxRetries(count != RETRY_COUNT_FOREVER ? count : -1)
-        .withDelay(frequency)
+        .withDelay(frequency.isZero() ? ofMillis(1) : frequency)
```
marianogonzalez, line 74: why this change? how does it changes behavior?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -542,60 +574,52 @@ public FlowConstruct getFlowConstruct() {
...
+        // affect the connection's invalidation
+        extractConnectionException(e).ifPresent(connectionException -> invalidateConnection(connectionException));
+        stopSource();
+      } catch (Exception eStop) {
+        e.addSuppressed(eStop);
+      }
+      try {
+        disposeSource();
+      } catch (Exception eDispose) {
+        e.addSuppressed(eDispose);
```
marianogonzalez, line 599: why?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -542,60 +574,52 @@ public FlowConstruct getFlowConstruct() {
...
-    @Override
-    public Object getWorkOwner() {
-      return ExtensionMessageSource.this;
+    if (statistics != null && statistics.isEnabled() && computeConnectionErrorsInStats()) {
+      statistics.getApplicationStatistics().incConnectionErrors();
     }
   }
 
+  private boolean computeConnectionErrorsInStats() {
+    return featureFlaggingService.isEnabled(COMPUTE_CONNECTION_ERRORS_IN_STATS);
```
marianogonzalez, line 620: why? What I would expect from a feature flag is recovering the current behavior rather than enabling this

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -204,20 +217,41 @@ private synchronized void createSource(boolean restarting) throws Exception {
...
+        .applyPolicy(futureSupplier, t -> true, t -> computeStats(), NULL_THROWABLE_CONSUMER, identity(), retryScheduler)
+        .whenComplete((v, e) -> {
+          if (e != null) {
+            onFailure.accept(e);
+          } else {
+            onSuccess.run();
+          }
+        });
     try {
-      if (restarting && retryPolicyTemplate.isAsync()) {
```
marianogonzalez, line 210: this looks like a fundamental departure of a behavior that was explicitly put in place to avoid race conditions. @fsgonz can you please review this? 

@SebaElizalde why is the restarting condition no longer relevant? Why is the IMMEDIATE_SCHEDULER no longer used?

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
fsgonz: This has been change since introduced originally (MULE-17267). The idea then was to perform the retry in another thread in case of restarting (actually this was already changed in MULE-18777). Does this not fail for the app in the ticket associated to MULE-17267?

core/src/main/java/org/mule/runtime/core/internal/retry/policies/SimpleRetryPolicy.java
SebaElizalde: jodah does not allow a 0 as a valid parameter value, however we do allow 0 values for the frequency param in our retry policy

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
SebaElizalde: I kept this method exactly as it was before (only added the future as a param) see above the deleted lines above for reference.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
SebaElizalde: what do you mean? This is just checking if the feature isEnabled or not, it is not enabling it

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
SebaElizalde: There is no app associated to MULE-17267 sadly.
Indeed it was changed in MULE-18777: https://github.com/mulesoft/mule/commit/e19925827c5249e9435122c5fe2ab7b4f4bca45b (line 209), which makes it confusing as to what behaviour should be honoured. Perhaps line 314 of that PR causes that the reconnection to always executes in another thread regardless of the policy.
In any case, we could now handle sync/async with the future instead of the scheduler?

fsgonz: We've verified that the thread switch in the case of reconnection happens before that, so it won't be a problem.

PR #10824, opened 2021-09-24
MULE-19804: Migrate uses of commons-collections to streams api

elrodro83:
* Move leftovers to mvel module
* Deprecate exported classes that are not being used


core/src/main/java/org/mule/runtime/core/internal/streaming/object/iterator/CompositeProducer.java
```diff
@@ -35,7 +33,7 @@
    * @param producers a list of {@link Producer}
    */
   public CompositeProducer(List<Producer<T>> producers) {
-    if (CollectionUtils.isEmpty(producers)) {
+    if (producers == null || producers.isEmpty()) {
```
fsgonz, line 36: Is this change because of performance? It only avoids a method invocation.

fsgonz (commented):
Is there a way to verify that in every case CollectionUtils.isEmpty was replaced by non null safe isEmpty we won't have any problem?

core/src/main/java/org/mule/runtime/core/internal/streaming/object/iterator/CompositeProducer.java
elrodro83: more or less... the idea is to eventually remove the dependency on commons-collections. With this, there would only remain some base decorators and some other specific things that must be reviewed case by case in order to finish that.

elrodro83:
i reviewed how the collections thta were teste were created to have some reassurance that the change wouldn't break

fsgonz (approved):

PR #10828, opened 2021-09-27
MULE-19795: Make functional tests run with java 11

ndinu:


Bardammu (approved):

tests/runner/src/main/java/org/mule/test/runner/utils/RunnerModuleUtils.java
```diff
@@ -94,16 +121,495 @@ public static void assureSdkApiInClassLoader(ClassLoader extensionClassLoader,
...
+      for (Map.Entry<TypeDescription, byte[]> classEntry : getSdkClassMap(zipInputStream, addedClasses).entrySet()) {
+
+        try {
+          Map<TypeDescription, byte[]> mapEntry = new HashMap<>();
+          mapEntry.put(classEntry.getKey(), classEntry.getValue());
+          ClassLoadingStrategy.Default.INJECTION.load(extensionClassLoader, mapEntry);
+          addedClasses.add(classEntry.getKey().getName());
+          classAdded = true;
+        } catch (Exception e) {
+          // Problem found adding class to class loader.
```
marianogonzalez, line 159: Don't hide the exception. At the very least log it, but why wouldn't it be thrown?

marianogonzalez (changes requested):

tests/runner/src/main/java/org/mule/test/runner/utils/RunnerModuleUtils.java
ndinu: Some classes need to be added in certain order to the class loader, so in some cases this is expected to fail, and the class will be loaded on another try. I can log this in debug level if you think that is the correct approach.

marianogonzalez: Yes. If that's the case, you should make a DEBUG log when that happens and you should also DEBUG log when a class is successfully loaded, so that you can actually verify if it worked on the next attempt. Any exception other than that should result in failure

tests/runner/src/main/java/org/mule/test/runner/utils/RunnerModuleUtils.java
```diff
@@ -155,8 +159,9 @@ private static void addSdkApiClassesDynamically(ClassLoader extensionClassLoader
           ClassLoadingStrategy.Default.INJECTION.load(extensionClassLoader, mapEntry);
           addedClasses.add(classEntry.getKey().getName());
           classAdded = true;
+          LOGGER.debug("Class {} was succesfully added to the extension classloader.", classEntry.getKey().getName());
         } catch (Exception e) {
-          // Problem found adding class to class loader.
+          LOGGER.debug("Class {} failed to be added to the extension classloader.", classEntry.getKey().getName());
```
marianogonzalez, line 164: also log the exception and its message

marianogonzalez: is there a way to separate the wrong class order case from other errors which shouldn't be ignored?

marianogonzalez (changes requested):

marianogonzalez (approved):

PR #10831, opened 2021-09-29
MULE-19814: Add boolean to indicate if a MIME Type is set for Set Var…

hyerans:
…iable

- Add boolean to check if MIME type is set
- Update the if condition to use the content type information when the mime type is missing
- Added Test


tests/integration/src/test/resources/set-variable-mime-type-config.xml
```diff
@@ -0,0 +1,35 @@
...
+
+        <component class="org.mule.transformers.simple.SetVariableMimeTypeTestCase$FlowVariableDataTypeAccessor"/>
+    </flow>
+
+    <flow name="contentTypeOnly">
+        <vm:inbound-endpoint path="testInput2" exchange-pattern="request-response"/>
+
+        <set-variable variableName="testVariable" value="#['test']" encoding="UTF-16"/>
+
+        <set-property propertyName="Content-Type" value="application/json;charset=UTF-16BE"/>
```
sofiamorseletto, line 21: I don't think this is reproducing the payload test case, I think the ideal thing to do would be to send the Content type on the client request instead of setting it on the app

tests/integration/src/test/java/org/mule/transformers/simple/SetVariableMimeTypeTestCase.java
```diff
@@ -0,0 +1,77 @@
...
+    @Test
+    public void requestWithoutAnyDataTypeSet() throws Exception
+    {
+        LocalMuleClient client = muleContext.getClient();
+
+        MuleMessage response = client.send("vm://testInput3", TEST_MESSAGE, null);
+
+        String mimeType = response.getDataType().getMimeType();
+
+        assertThat(mimeType, equalTo(MimeTypes.ANY));
```
sofiamorseletto, line 66: Please specify the exact mime type that the test is expecting (*/*)

sofiamorseletto (changes requested):

tests/integration/src/test/resources/set-variable-mime-type-config.xml
hyerans: Removed this property from the app and added setOutboundProperty to request instead.

tests/integration/src/test/java/org/mule/transformers/simple/SetVariableMimeTypeTestCase.java
hyerans: Adding another assertion for "\*/\*" to add more clarity

tests/integration/src/test/resources/set-variable-mime-type-config.xml
sofiamorseletto: Notice that you're adding the property to the response not to the request, after the flow has run. Maybe there's some way to add the content-type when sending the request

hyerans: Used props map instead to have a more relevant test case as suggested.

sofiamorseletto (approved):

PR #10832, opened 2021-09-29
MULE-19516: Support new SDK API error package. Support MuleErrors enums

Bardammu:


core/src/main/java/org/mule/runtime/core/internal/connection/ErrorTypeHandlerConnectionProviderWrapper.java
```diff
@@ -129,7 +131,15 @@ public RetryPolicyTemplate getRetryPolicyTemplate() {
     return errorTypeRepository.getErrorType(getIdentifier(errorType));
   }
 
+  private Optional<ErrorType> getErrorType(org.mule.sdk.api.error.ErrorTypeDefinition errorType) {
```
marianogonzalez, line 134: same here. The goal is to reduce all thing to a single interface. We should make sure to perform all the necessary adaptations so that our code can deal only in terms of one single interface.

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/ConnectivityErrorPropagationTestCase.java
```diff
@@ -57,19 +58,22 @@ public void failAtConnectWithConnectionException() {
   }
 
   @Test
+  @Ignore
```
marianogonzalez, line 61: unignore these

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/exception/ModuleExceptionHandler.java
```diff
@@ -53,6 +55,13 @@ public ModuleExceptionHandler(ComponentModel componentModel, ExtensionModel exte
           .name(errorDefinition.getType())
           .build());
 
+      String errorTypeDefinitionName;
```
marianogonzalez, line 58: instead of doing this, you should just overwrite the `toString()` implementation in the adapter

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/exception/ModuleExceptionHandler.java
```diff
@@ -116,7 +133,8 @@ private boolean isAllowedError(ErrorType errorType) {
 
   private Throwable getExceptionCause(Throwable throwable) {
     // For subclasses of ModuleException, we use it as it already contains additional information
-    if (throwable.getClass().equals(ModuleException.class)) {
+    if (throwable.getClass().equals(ModuleException.class) ||
```
marianogonzalez, line 136: same as before

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/DefaultExtensionModelFactoryTestCase.java
```diff
@@ -389,7 +389,7 @@ private void assertSinglePrimaryContentParameter(ExtensionModel extensionModel,
   @Export(classes = {HeisenbergExtension.class, HeisenbergException.class}, resources = "methRecipe.json")
   @SubTypeMapping(baseType = Weapon.class, subTypes = {Ricin.class})
   @SubTypeMapping(baseType = Investment.class, subTypes = {CarWash.class, CarDealer.class})
-  @ErrorTypes(HeisenbergErrors.class)
+  @org.mule.sdk.api.annotation.error.ErrorTypes(HeisenbergErrors.class)
```
marianogonzalez, line 392: I don't understand the value of this. What I would expect here is for the new annotation to use a different set of errors. Otherwise, you're not sure where the definition is coming from. As a matter of fact, I even have mixed feeling on whether this should be permitted or result in an error. I think the current tests should remain untouched, and you should use the new API to add new errors while adding new tests which assert on those.

marianogonzalez (changes requested):

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/DefaultExtensionModelFactoryTestCase.java
Bardammu: I just wanted to check that I don't break the model. I actually modified the errors to be SDK-api error, so they are new in a way
I should add new tests, that's correct

core/src/main/java/org/mule/runtime/core/internal/connection/ErrorTypeHandlerConnectionProviderWrapper.java
```diff
@@ -118,6 +118,8 @@ public RetryPolicyTemplate getRetryPolicyTemplate() {
   private Optional<ErrorType> getErrorType(Throwable exception) {
     if (exception instanceof ModuleException) {
       return getErrorType(((ModuleException) exception).getType());
+    } else if (exception instanceof org.mule.sdk.api.exception.ModuleException) {
```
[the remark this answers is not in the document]
Bardammu: It'll require to move the adapters to `mule-extensions-api`, do you think that it's a better approach?

core/src/main/java/org/mule/runtime/core/internal/connection/ErrorTypeHandlerConnectionProviderWrapper.java
Bardammu: But at this point we are getting the Exception from the connector, shouldn't we deal with both interfaces here?

PR #10843, opened 2021-10-04
MULE-19503 - Support new SDK API Execution annotation package

dabreu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/SourceTypeWrapper.java
```diff
@@ -99,9 +101,16 @@ public SourceTypeWrapper(Class<T> aClass, ClassTypeLoader typeLoader) {
...
+                                                             aClass.getName(), annotationTypes[0].getSimpleName()));
     } else {
       return of(new MethodWrapper(methods.iterator().next(), typeLoader));
     }
   }
+
+  private Collection<Method> getMethodAnnotatedWith(Class<?> searchClass, Class<? extends Annotation>... annotationTypes) {
+    Collection<Method> methods = new ArrayList<>();
+    Stream.of(annotationTypes)
+        .forEach(annotationType -> methods.addAll(getMethodsAnnotatedWith(searchClass, annotationType, false)));
```
marianogonzalez, line 113: replace with `.map().collect(toList())` for efficiency

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/source/SourceExecutionSdkApiTestCase.java
```diff
@@ -0,0 +1,70 @@
...
+
+  @Override
+  protected String getConfigFile() {
+    return "source/source-notification-sdk-api-config.xml";
+  }
+
+  @Test
+  public void onSuccessMethodCall() throws Exception {
+    startFlow("onSuccessMethodCallFlow");
+    check(TIMEOUT, DELAY, () -> PetStoreSimpleSourceWithSdkApi.ON_SUCCESS_CALL_COUNT > 0);
```
marianogonzalez, line 43: use static imports

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/SourceTypeWrapper.java
```diff
@@ -99,9 +101,16 @@ public SourceTypeWrapper(Class<T> aClass, ClassTypeLoader typeLoader) {
     } else if (methods.size() > 1) {
       throw new IllegalSourceModelDefinitionException(
                                                       format("Source declared in class '%s' declares more than one method annotated with '%s'",
-                                                             aClass.getName(), annotationType.getSimpleName()));
+                                                             aClass.getName(), annotationTypes[0].getSimpleName()));
```
marianogonzalez, line 104: Can you explain the  `[0]`? what if the repetition comes from another annotation?

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/SourceTypeWrapper.java
dabreu: Yes, it is a private method that is intended only to be called within the class. The parameter is the list containing the same annotations, the legacy one and the sdk api one, and as they have the same name , that's why I used the first one.

marianogonzalez (approved):

PR #10852, opened 2021-10-06
MULE-19831: Nested transactions + flow ref hangs due to a deadlock

mbuchwald:


core/src/main/java/org/mule/runtime/core/api/transaction/TransactionCoordination.java
```diff
@@ -32,6 +32,10 @@
   private final ThreadLocal<Deque<Transaction>> suspendedTransaction = new ThreadLocal<>();
   private final ThreadLocal<Deque<Transaction>> isolatedTransactions = new ThreadLocal<>();
 
+  public boolean runningNestedTransaction() {
+    return suspendedTransaction.get() != null && suspendedTransaction.get().size() >= 1;
```
elrodro83, line 36: call `get` just once

elrodro83 (changes requested):

PR #10872, opened 2021-10-18
MULE-19857: dispose maybe called twice from MuleContainer.stop() and so avoid hooks 

mbuchwald:
This is a copy of #10816  (it won't run validations from its branch name)


mbuchwald:
I wanted to test this but there isn't a way to change the `defaultLogManagerFactory` (static) attribute, and adding a setter just for testing with no real effect but to force a potential issue while testing wouldn't be much of a proof either way

PR #10873, opened 2021-10-18
MULE-19858: XML-SDK extension model generation is loading only mule extension model

elrodro83:


modules/deployment-model/src/test/java/org/mule/runtime/deployment/model/api/artifact/extension/ExtensionModelDiscovererTestCase.java
```diff
@@ -0,0 +1,84 @@
...
+import org.mule.tck.junit4.AbstractMuleTestCase;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import org.junit.Test;
+
+import io.qameta.allure.Description;
+import io.qameta.allure.Issue;
+
+public class ExtensionModelDiscovererTestCase extends AbstractMuleTestCase {
```
fsgonz, line 36: Can we add a @Story and @Feature for this test?

fsgonz (changes requested):

PR #10892, opened 2021-10-19
MULE-19352: DynamicConfigWithStatefulOperationConfigurationOverride validation

elrodro83:


runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/DynamicConfigWithStatefulOperationConfigurationOverride.java
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  @Override
+  public Predicate<List<ComponentAst>> applicable() {
+    return currentElemement(component -> !getConfigOverrideParams(component).isEmpty());
+  }
+
+  @Override
+  public Optional<ValidationResultItem> validate(ComponentAst component, ArtifactAst artifact) {
+    return getConfigOverrideParams(component)
```
ndinu, line 55: Regarding the comment in https://www.mulesoft.org/jira/browse/MULE-17906 , the validation should check that the parameter is defined at field level in the operation class. For this, I think that the presence of the FieldOperationParameterModelProperty in the parameter can be checked

elrodro83:
--validate

PR #10931, opened 2021-10-28
MULE-19505: Support new SDK API Source annotation package

ndinu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/SourceModelLoaderDelegate.java
```diff
@@ -104,10 +123,63 @@ void declareMessageSources(ExtensionDeclarer extensionDeclarer,
...
+      case NOT_SUPPORTED:
+        runsOnPrimaryNodeOnly = true;
+        break;
+      default:
+        runsOnPrimaryNodeOnly = true;
+    }
+    sourceDeclarer.runsOnPrimaryNodeOnly(runsOnPrimaryNodeOnly);
+  }
+
+  private void declareBackPressureSupport(ExtensionDeclarer extensionDeclarer, SourceDeclarer sourceDeclarer,
```
marianogonzalez, line 153: this should still happen on an enricher, only that one that's fixed in the ExtensionModelFactory and not java specific

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/SourceModelLoaderDelegate.java
ndinu: So in the parsing I add the model property, and consume it in the enricher?

What about the cluster support, does it also need to still be added in an enricher for all extensions?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/SourceModelLoaderDelegate.java
```diff
@@ -104,10 +108,35 @@ void declareMessageSources(ExtensionDeclarer extensionDeclarer,
...
+      sourceDeclarer.withModelProperty(parser.getBackPressureStrategyModelProperty().orElse(getDefault()));
+
       sourceDeclarers.put(parser, sourceDeclarer);
     }
   }
 
+  private void declareClusterSupport(SourceDeclarer sourceDeclarer, Optional<SourceClusterSupport> sourceClusterSupport) {
+    boolean runsOnPrimaryNodeOnly;
+    SourceClusterSupport resultingSourceClusterSupport;
+    switch (sourceClusterSupport.orElse(DEFAULT_ALL_NODES)) {
```
marianogonzalez, line 122: this switch doesn't belong there. The loader should just set sourceClusterSupport on the declarer. The logic about how that affects the `runsOnPrimaryNodeOnly` is a language level concern.

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/SourceModelLoaderDelegate.java
ndinu: I would take the configuration of the `runsOnPrimaryNodeOnly` to the enricher right? 

Regarding the switch, I would still need it to get the adequate SourceClusterSupport for the model property.

marianogonzalez: right. The point is that setting the  `runsOnPrimayNodeOnly` property in the `SourceModel` is not something that should be left up to the loader since many things depend on that. To clarify, the problem is not the switch itself, but its location

ndinu:
--validate

marianogonzalez (approved):

PR #10963, opened 2021-11-11
MULE-19918: Add maven profile that skip tests on projects we don't want to test with hawk.

pabloperalta:


pabloperalta:
--validate

elrodro83 (approved):

PR #10964, opened 2021-11-11
Adding SCOPE to the location part identifier type filter used for getting the processing strategies.

asanguinetti:
(This branch is just for testing, it's not the idea to merge it).

- This should fix the case of processors inside subflows not getting the processing strategy of the root flow.


asanguinetti:
--validate

PR #10976, opened 2021-11-13
MULE-19924: Processors in subflows are not getting the processing strategy of the root flow

asanguinetti:
- Adding top level scopes (i.e.: subflows) to the getProcessingStrategy filter.
- This is a temporary fix. There is [MULE-19930](https://www.mulesoft.org/jira/browse/MULE-19930) to properly fix this.


asanguinetti:
--validate

elrodro83 (approved):

PR #10977, opened 2021-11-15
Log more details about open sinks

ca-stefan-cordes:
Stopping one of our workers shows 
```
WARN    org.mule.runtime.core.internal.processor.strategy.StreamPerThreadSink TX Subscribers of ProcessingStrategy for flow 'upsert-customer' not completed in 60000 ms
WARN    org.mule.runtime.core.internal.processor.strategy.StreamPerThreadSink TX Subscribers of ProcessingStrategy for flow 'verify-if-customer-update-needed' not completed in 60000 ms
```

but just the flow-name without count of sinks and/or cache content makes it impossible to find the root cause of pending sinks.

So I added more logging information in case of disposableSinks still filled.


rbourbonnavarro:
Hi @ca-stefan-cordes, does this have an associated MULE ticket in Jira?

ca-stefan-cordes:
No, but a case:
00320892	BitronixTransactionManager did not shutdown and left DB2 Indoubt Thread open

PR #10982, opened 2021-11-15
MULE-19931: mapReduceAnnotation function family should traverse class…

marianogonzalez:
… hierarchy


dabreu:
I believe that it'd be good to have a test that verifies the fix, what do you think?

marianogonzalez:
@dabreu I believe that the current functional coverage and the ext model serialization tests are enough, but having unit tests certainly doesn't hurt. Adding one

dabreu (approved):

PR #10985, opened 2021-11-15
MULE-19921: Redelivery Policy: infinite loop with blank message ID in a source configured with transactions

rbourbonnavarro:


modules/core-components/src/main/java/org/mule/runtime/core/internal/processor/IdempotentRedeliveryPolicy.java
```diff
@@ -263,6 +265,11 @@ public CoreEvent process(CoreEvent event) throws MuleException {
         incrementCounter(messageId, createMessagingException(event, ex));
         throw ex;
       }
+    } catch (ObjectStoreException e) {
```
elrodro83, line 268: what if there was another problem in the object store not related to the key being null?

elrodro83 (changes requested):

elrodro83 (approved):

rbourbonnavarro:
--validate

rbourbonnavarro:
--validate

rbourbonnavarro:
--validate

PR #11021, opened 2021-12-01
MULE-19962: xml-sdk properties not accessible within a second nested private operation

Bardammu:


dabreu (dismissed):

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/processor/ModuleOperationMessageProcessor.java
```diff
@@ -358,18 +358,21 @@ private CoreEvent createEventWithParameters(CoreEvent event) {
...
     // If this operation is called from an outer operation, we need to obtain the config from the previous caller in order to
     // populate the event variables as expected.
-    if (properties.isEmpty() && event.getVariables().containsKey(MODULE_OPERATION_CONFIG_REF)) {
-      resolvedProperties = parseParameters(createPropertiesFromConfigName((String) event.getVariables()
-          .get(MODULE_OPERATION_CONFIG_REF).getValue()), allProperties);
-    } else {
-      resolvedProperties = properties;
+    Map<String, Pair<Object, MetadataType>> resolvedProperties = properties;
+    if (event.getVariables().containsKey(MODULE_OPERATION_CONFIG_REF)) {
+      builder.addVariable(MODULE_OPERATION_CONFIG_REF, event.getVariables().get(MODULE_OPERATION_CONFIG_REF).getValue());
```
elrodro83, line 365: avoid the overhead of querying the variables map more than once, extract `event.getVariables().get(MODULE_OPERATION_CONFIG_REF)` to a variable and use `v != null` in the `if` above

elrodro83 (changes requested):

elrodro83 (approved):

PR #11027, opened 2021-12-01
MULE-19962: xml-sdk properties not accessible within a second nested private operation (#11021)

Bardammu:
…private operation (#11021)


dabreu (approved):

Bardammu:
--validate

PR #11039, opened 2021-12-07
MULE-19521: Support new SDK API runtime package.

SebaElizalde:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceAdapter.java
```diff
@@ -521,12 +522,10 @@ public Object getDelegate() {
   }
 
   Optional<Publisher<Void>> getReconnectionAction(ConnectionException e) {
-    if (sourceInvokationTarget.get() instanceof Reconnectable) {
-      return of(
-                create(sink -> ((Reconnectable) sourceInvokationTarget.get()).reconnect(e,
-                                                                                        new ReactiveReconnectionCallback(sink))));
+    Reconnectable adapter = SdkReconnectableAdapter.from(sourceInvokationTarget.get());
+    if (adapter != null) {
```
marianogonzalez, line 526: bug. This will never (and should never) return null. If the invocation target is not a Reconnectable then it throws exception. Fix and add unit test

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceAdapter.java
SebaElizalde: What you suggest is that the SdkReconnectableAdapter.from(...) method should throw an Exception which is then caught here and in that case return empty()? @marianogonzalez

marianogonzalez (approved):

SebaElizalde:
--validate

SebaElizalde:
--validate

ndinu (approved):

SebaElizalde:
--validate

SebaElizalde:
--validate

PR #11045, opened 2021-12-09
MULE-19960: Refactor how transformerResolvers/converters are discovered

elrodro83:


core/src/main/java/org/mule/runtime/core/internal/message/DefaultMessageBuilder.java
```diff
@@ -527,8 +531,7 @@ public SerializedDataHandler(String name, DataHandler handler, MuleContext muleC
           } else if (currentMuleContext.get() != null) {
             try {
               DataType source = fromObject(theContent);
-              Transformer transformer =
-                  ((MuleContextWithRegistry) muleContext).getRegistry().lookupTransformer(source, DataType.BYTE_ARRAY);
+              Transformer transformer = transformersRegistry.lookupTransformer(source, DataType.BYTE_ARRAY);
```
eze210, line 534: static import

eze210 (changes requested):

core/src/main/java/org/mule/runtime/core/privileged/transformer/TransformerUtils.java
```diff
@@ -131,12 +135,23 @@ public static void checkTransformerReturnClass(Transformer transformer, Object v
     }
   }
 
+  /**
+   * @deprecated since 4.5
```
eze210, line 139: Why is this deprecated without a replacement?

eze210 (approved):

PR #11188, opened 2022-01-11
MULE-20032: Fix messages from pre-built back pressure exceptions in AbstractPipeline.

asanguinetti:


core/src/main/java/org/mule/runtime/core/internal/construct/FlowBackPressureException.java
```diff
@@ -21,7 +21,6 @@
  */
 public abstract class FlowBackPressureException extends MuleException {
 
-  static final String BACK_PRESSURE_ERROR_MESSAGE = "Flow '%s' is unable to accept new events at this time. Reason: %s";
```
asanguinetti, line 24: This change is unrelated, but I just removed it because it was unused. The message is actually taken from `CoreMessages` for i18n.

core/src/main/java/org/mule/runtime/core/internal/construct/FlowBackPressureException.java
```diff
@@ -44,17 +43,27 @@ public FlowBackPressureException(Component flow, BackPressureReason reason, Thro
     this.flow = flow;
   }
 
+  /**
+   * @return An instance of the right type of {@link FlowBackPressureException} according to the given {@link BackPressureReason}.
+   */
+  public static FlowBackPressureException createFlowBackPressureException(Component flow, BackPressureReason reason) {
+    return createFlowBackPressureException(flow, reason, null);
```
asanguinetti, line 50: Note that I'm passing `null` as cause here in order to reuse the other pre-existing version which expects the three parameters. One possible concern is that by doing this we are never using the two parameters constructor from `FlowBackPressureException` and its subclasses. But in the end, they all end up calling the `MuleException` constructor passing a `null` value when there is no cause.

If I wanted to avoid this I would either have to duplicate the switch-case, or make the code a lot less readable.

core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
```diff
@@ -227,18 +225,8 @@ public ProcessingStrategy getProcessingStrategy() {
...
-                               new FlowBackPressureEventsAccumulatedException(this, EVENTS_ACCUMULATED));
-    backPressureExceptions.put(MAX_CONCURRENCY_EXCEEDED,
-                               new FlowBackPressureMaxConcurrencyExceededException(this, EVENTS_ACCUMULATED));
-    backPressureExceptions.put(REQUIRED_SCHEDULER_BUSY,
-                               new FlowBackPressureRequiredSchedulerBusyException(this, EVENTS_ACCUMULATED));
-    backPressureExceptions.put(REQUIRED_SCHEDULER_BUSY_WITH_FULL_BUFFER,
-                               new FlowBackPressureRequiredSchedulerBusyWithFullBufferException(this, EVENTS_ACCUMULATED));
-
-    this.backPressureExceptions = unmodifiableMap(backPressureExceptions);
+    this.backPressureExceptions = unmodifiableMap(Arrays.stream(BackPressureReason.values())
```
elrodro83, line 228: static import

elrodro83 (changes requested):

elrodro83 (approved):

PR #11206, opened 2022-01-13
MULE-20042: When a bean fails creation, it is attempted to create again for disposal

elrodro83:


core/src/main/java/org/mule/runtime/core/internal/registry/AbstractRegistry.java
```diff
@@ -145,11 +147,13 @@ private boolean hasStartFailed() {
...
-        logger.warn(String.format("Could not apply shutdown lifecycle to object '%s' after being unregistered.", key), e);
+    if (object != null) {
+      try {
+        getLifecycleManager().applyPhase(object, getLifecycleManager().getCurrentPhase(), Disposable.PHASE_NAME);
+      } catch (Exception e) {
+        if (logger.isWarnEnabled()) {
+          logger.warn(format("Could not apply shutdown lifecycle to object '%s' after being unregistered.", key), e);
+        }
       }
     }
```
mbuchwald, line 158:
```suggestion
    if (object == null) {
        return null;
    }
    try {
        getLifecycleManager().applyPhase(object, getLifecycleManager().getCurrentPhase(), Disposable.PHASE_NAME);
   } catch (Exception e) {
        if (logger.isWarnEnabled()) {
          logger.warn(format("Could not apply shutdown lifecycle to object '%s' after being unregistered.", key), e);
        }
   }
```

mbuchwald (changes requested):

mbuchwald (approved):

PR #11238, opened 2022-01-17
EE-7870: Move streaming manager to use memory manager buffers

IvanAndresFritzler:


core/src/main/java/org/mule/runtime/core/api/streaming/DefaultStreamingManager.java
```diff
@@ -108,8 +115,10 @@ private ByteBufferManager createByteBufferManager() throws InitialisationExcepti
                                                                    BUFFER_MANAGER_FACTORY_CLASS)),
                                         e, this);
     }
-
-    return factory.create();
+    ByteBufferManager byteBufferManager = factory.create();
+    byteBufferManager.setByteBufferProvider(memoryManagementService.getByteBufferProvider(getByteBufferProviderName(), HEAP));
```
marianogonzalez, line 119: why keep having the ByteBufferManager at all?  Looks like what the ByteBuferProvider should be able to do everything it does

marianogonzalez (changes requested):

IvanAndresFritzler:
--validate

IvanAndresFritzler:
--validate

PR #11239, opened 2022-01-17
MULE-20048: Rework MULE-19803 to leave AST as-is

elrodro83:


modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/parameter/value/ParameterWhitespaceTrimmingTestCase.java
```diff
@@ -129,7 +143,7 @@ public void whitespacesAreNotTrimmedForPojoTextExpression() throws Exception {
   @Test
   public void whitespacesAreTrimmedForPojoTextCDATA() throws Exception {
```
dabreu, line 144: I think that the name of the test should be changed to whitespacesAreNotTrimmedForPojoTextCDATA if cdata is not going to be trimmed.

SebaElizalde (approved):

PR #11241, opened 2022-01-18
MULE-20051: Break inheritance to run less tests

jcgrethe:


modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/ApplicationRedeploymentTestCase.java
```diff
@@ -208,8 +252,6 @@ public void redeploysInvalidExplodedAppAfterSuccessfulDeploymentOnStartup() thro
   }
 
   @Test
-  @Ignore("MULE-16403")
```
jcgrethe, line 211: Seems that it's not failing locally. Also, the issue was closed.

elrodro83 (dismissed):

elrodro83 (approved):

PR #9347, opened 2020-09-03
MULE-18654: Decorate operation input streams with data flow statistics decorator

elrodro83:

PR #11252, opened 2022-01-20
MULE-20054: Optimize decoration of input streams

asanguinetti:
It has been observed that change #9347 introduced a performance regression of about 3~4%.
Upon closer inspection, it seems most of the regression came from the fact that we are now decorating with `ObjectArgumentResolverDecorator` a lot of arguments that we were not decorating before (for example, arguments of type `List`). This change was done so that these arguments were also decorated by the new `CursorComponentDecoratorFactory`.

However, the `ObjectArgumentResolverDecorator` was added on #8528 because some arguments of `Object` type required cursor resolution.

For many of the new arguments that we now want to decorate because of the `CursorComponentDecoratorFactory` (like `List`), we don't really seem to be needing the cursor resolution. I'm proposing to use a different `ArgumentResolverDecorator` that decorates with `CursorComponentDecoratorFactory` but does not perform cursor resolution.

Moreover, it doesn't seem that we need to decorate all arguments with this, only those that are annotated as `Content`, because we were using `NO_OP_INSTANCE` as `CursorComponentDecoratorFactory` for everything else anyway.

[Here](https://stats.perf.anypoint.mulesoft.com/goto/8c61069eaa233dcb026d9c9eded18157) you can see the result of the performance executions of these same changes applied on a 4.4.0-20211129 (that one was selected for stability). The version labeled **MULE-20054b** would be containing only the first commit of this PR (0e175ce) and not the second one (b022efb).

Version | TPS | % to baseline
-- | -- | --
4.4.0-20211129 | 10492 | 0.00%
4.4.0-20211129 MULE-20054b | 10754 | 2.50%
4.4.0-20211129 MULE-20054 | 10816 | 3.09%

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/loader/java/type/ExtensionParameter.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/loader/java/type/ExtensionParameter.java
@@ -21,6 +21,7 @@ import org.mule.runtime.api.tls.TlsContextFactory;
 import org.mule.runtime.core.api.retry.policy.RetryPolicyTemplate;
 import org.mule.runtime.extension.api.annotation.param.Config;
 import org.mule.runtime.extension.api.annotation.param.Connection;
+import org.mule.runtime.extension.api.annotation.param.Content;
 import org.mule.runtime.extension.api.annotation.param.DefaultEncoding;
 import org.mule.runtime.extension.api.annotation.param.Optional;
 import org.mule.runtime.extension.api.client.ExtensionsClient;
@@ -132,6 +133,14 @@ public interface ExtensionParameter extends WithType, WithAnnotations, NamedObje
       return getDefaultValue(annotation.get());
     }
     return java.util.Optional.empty();
+  }
+
+  /**
+   * @return Whether this parameter is considered content.
+   */
+  default boolean isContent() {
+    return isAnnotatedWith(Content.class)
+        || isAnnotatedWith(org.mule.sdk.api.annotation.param.Content.class);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/MethodArgumentResolverDelegate.java
```diff
@@ -356,8 +362,10 @@ private void initArgumentResolvers() {
           }
         }
       }
-    } else {
+    } else if (Object.class.equals(argumentType)) {
```
fsgonz, line 365: MULE-17903 to verify why this was done, as indicated in the PR full explanation.

fsgonz (approved):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/loader/java/type/ExtensionParameter.java
```diff
@@ -134,6 +135,14 @@ default boolean isRequired() {
     return java.util.Optional.empty();
   }
 
+  /**
+   * @return Whether this parameter is considered content.
+   */
+  default boolean isContent() {
```
marianogonzalez, line 141: this shouldn't be done here nor it should depend on the extension code but the extension model

marianogonzalez (changes requested):
This solution is trying to optimize #9234 . The problem is that had I seen that PR originally, I woudl have rejected it:

1) The entire solution is entirely based on specific resolvers of the Java SDK. This means that it won't work any of the other SDK flavours 
2) It's based on introspecting the extension's code. It should instead be inspecting the resulting ExtensionModel.
3) it forces adding similar logic for sources, functions, etc. This should all be centralized on the returning piece of it, not on each usage.
4) it forces propagating concepts such as content to components that should not have it

The correct solution here is to revert #9234 and reimplement it completely. This solution does not scale and does not align with the refactors we're doing in the SDK to do reuse. Please write/refer to a spec in which the design for this was discussed.

Thank you

asanguinetti:
I'm closing this PR because we are going to go with the revert option.

PR #11258, opened 2022-01-24
[DO NOT MERGE] test SE-24213

eze210:


eze210:
Note to write the test:
```
<xmlsdk-operation config-ref="configRef">
	<xmlsdk-primary-content-parameter><![CDATA[Test]]></xmlsdk-primary-content-parameter>
</xmlsdk-operation>
```

PR #11296, opened 2022-02-03
MULE-19984: Implement a new LifecycleObjectSorter and dependency resolvers to reduce app deployment time

hyerans:


core/src/main/java/org/mule/runtime/core/internal/lifecycle/phases/LifecycleObjectSorter.java
```diff
@@ -30,4 +31,11 @@
    * @return The sorted list of objects previously added through {@link #addObject(String, Object)}
    */
   List<Object> getSortedObjects();
+
+  /**
+   * Provides the order of objects as reference for initialise/dispose phases
+   * 
+   * @param lookupObjectsForLifecycle lifecycle object list which is ordered based on the type
+   */
+  void setLifeCycleObjectNameOrderMap(Map<String, Object> lookupObjectsForLifecycle);
```
mbuchwald, line 40: Add default empty method, so you don't have to redefine it empty in other cases later!

modules/spring-config/src/main/java/org/mule/runtime/config/internal/AutoDiscoveredDependencyResolver.java
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.config.internal;
```
elrodro83, line 7: move this and related cpde to its own subpackage

modules/spring-config/src/main/java/org/mule/runtime/config/internal/BeanVertexWrapper.java
```diff
@@ -0,0 +1,49 @@
...
+ * BeanVertexWrapper helps the comparison/equality check among non-proxy vs. proxy or proxy vs. proxy when building a dependency
+ * graph.
+ */
+public class BeanVertexWrapper {
+
+  final Object wrappedObject;
+  final String beanName;
+
+  public BeanVertexWrapper(String beanName, Object o) {
+    requireNonNull(o, "bean should not be null");
```
elrodro83, line 21: should -> must

core/src/main/java/org/mule/runtime/core/internal/lifecycle/phases/LifecycleObjectSorter.java
hyerans: done!

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DeclaredDependencyResolver.java
```diff
@@ -0,0 +1,39 @@
...
+
+  public List<BeanVertexWrapper> getDeclaredDependencies(Object object) {
+    List<BeanVertexWrapper> declaredDependencies = new ArrayList<>();
+    if (object instanceof InjectedDependenciesProvider) {
+      ((InjectedDependenciesProvider) object).getInjectedDependencies()
+          .forEach(dependency -> dependency
+              .reduce(type -> Stream.of(springRegistry.getBeanNamesForType(dependency.getLeft()))
+                  .map(beanName -> new BeanVertexWrapper(beanName, springRegistry.get(beanName)))
+                  .collect(toList()), beanName -> asList(new BeanVertexWrapper(beanName, springRegistry.get(beanName))))
+              .forEach(v -> declaredDependencies.add(v)));
```
elrodro83, line 35: use a collector instead of forEach + add

core/src/main/java/org/mule/runtime/core/internal/lifecycle/phases/DefaultLifecycleObjectSorter.java
```diff
@@ -11,6 +11,7 @@
 import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
```
elrodro83, line 14: remove unused import

core/src/main/java/org/mule/runtime/core/internal/lifecycle/phases/NotInLifecyclePhase.java
```diff
@@ -7,9 +7,11 @@
 package org.mule.runtime.core.internal.lifecycle.phases;
 
 import static java.util.Collections.emptyList;
+
 import org.mule.runtime.api.lifecycle.Initialisable;
 
 import java.util.List;
+import java.util.Map;
```
elrodro83, line 14: remove unused import

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphBeanDependencyResolver.java
```diff
@@ -0,0 +1,151 @@
...
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import javax.inject.Inject;
+
+import org.springframework.beans.factory.NoSuchBeanDefinitionException;
+import org.springframework.beans.factory.config.BeanDefinition;
+
+public class DependencyGraphBeanDependencyResolver implements BeanDependencyResolver {
```
elrodro83, line 25: javadoc

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphBeanDependencyResolver.java
```diff
@@ -0,0 +1,151 @@
...
+   * Add only direct children of current object
+   */
+  private void addDirectChild(String key, Object childObject, DependencyNode parent, Set<DependencyNode> processedKeys) {
+    if (!processedKeys.add(new DependencyNode(key, childObject))) {
+      return;
+    }
+    parent.addChild(new DependencyNode(key, childObject));
+  }
+
+
```
elrodro83, line 149: remove empty lines

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
```diff
@@ -0,0 +1,170 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.jgrapht.alg.cycle.CycleDetector;
+import org.jgrapht.graph.DefaultDirectedGraph;
+import org.jgrapht.graph.DefaultEdge;
+import org.jgrapht.traverse.TopologicalOrderIterator;
+
+public class DependencyGraphLifecycleObjectSorter implements LifecycleObjectSorter {
```
elrodro83, line 31: javadoc

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
```diff
@@ -0,0 +1,170 @@
...
+
+    // get (direct) prerequisite objects for the current object
+    List<Pair<String, Object>> prerequisiteObjects = resolver.getDirectBeanDependencies(beanName);
+
+    // add direct prerequisites to the graph & create edges(current object -> prerequisite)
+    if (prerequisiteObjects.isEmpty()) {
+      return;
+    }
+    prerequisiteObjects.forEach(
+                                (prerequisite) -> {
```
elrodro83, line 78: don't need the parenthesis around `(prerequisite)`

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
```diff
@@ -0,0 +1,170 @@
...
+   * should be initialised/disposed first will come first.
+   * 
+   * @return A list with all the objects sorted based on the lifecycle type
+   */
+  @Override
+  public List<Object> getSortedObjects() {
+    return dependencyGraphs.stream().map(x -> {
+      List<BeanVertexWrapper> sortedObjects = newArrayList(new TopologicalOrderIterator<>(x, new Comparator<BeanVertexWrapper>() {
+
+        //
```
elrodro83, line 127: remove

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyNode.java
```diff
@@ -0,0 +1,49 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.config.internal;
+
+import org.mule.runtime.api.util.Pair;
+
+import java.util.LinkedList;
+import java.util.List;
+
+public class DependencyNode {
```
elrodro83, line 14: javadoc

modules/spring-config/src/main/java/org/mule/runtime/config/internal/context/MuleArtifactContext.java
```diff
@@ -617,12 +644,40 @@ protected SpringMuleContextServiceConfigurator createServiceConfigurator(Default
   @Override
   protected void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory) {
     Optional<ComponentAst> configurationOptional = findComponentDefinitionModel(applicationModel, CONFIGURATION_IDENTIFIER);
-    if (configurationOptional.isPresent()) {
-      return;
+    if (!configurationOptional.isPresent()) {
```
elrodro83, line 647: this needs a rebase

modules/spring-config/src/main/java/org/mule/runtime/config/internal/registry/BeanDependencyResolver.java
```diff
@@ -23,4 +25,6 @@
    */
   List<Object> resolveBeanDependencies(String beanName);
 
+  ConfigurationDependencyResolver getConfigurationDependencyResolver();
```
elrodro83, line 28: javadoc, `@since`

modules/spring-config/src/test/java/org/mule/runtime/config/internal/AutoDiscoveredDependencyResolverTestCase.java
```diff
@@ -0,0 +1,37 @@
...
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+import org.mule.runtime.config.internal.registry.AbstractSpringRegistry;
+
+import io.qameta.allure.Description;
+import org.junit.Before;
+import org.junit.Test;
+
+public class AutoDiscoveredDependencyResolverTestCase {
```
elrodro83, line 19: allure

modules/spring-config/src/test/java/org/mule/runtime/config/internal/AutoDiscoveredDependencyResolverTestCase.java
```diff
@@ -0,0 +1,37 @@
...
+
+  AutoDiscoveredDependencyResolver autoDiscoveredDependencyResolver;
```
elrodro83, line 21: private

modules/spring-config/src/test/java/org/mule/runtime/config/internal/BeanVertexWrapperTestCase.java
```diff
@@ -0,0 +1,133 @@
...
+package org.mule.runtime.config.internal;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+public class BeanVertexWrapperTestCase {
```
elrodro83, line 16: allure

modules/spring-config/src/test/java/org/mule/runtime/config/internal/DeclaredDependencyResolverTestCase.java
```diff
@@ -0,0 +1,39 @@
...
+import static org.mockito.Mockito.verify;
+
+import org.mule.runtime.config.internal.registry.AbstractSpringRegistry;
+import org.mule.runtime.core.internal.lifecycle.InjectedDependenciesProvider;
+
+import io.qameta.allure.Description;
+import org.junit.Before;
+import org.junit.Test;
+
+public class DeclaredDependencyResolverTestCase {
```
elrodro83, line 20: allure

modules/spring-config/src/test/java/org/mule/runtime/config/internal/DependencyGraphBeanDependencyResolverTestCase.java
```diff
@@ -0,0 +1,77 @@
...
+
+import org.mule.runtime.config.internal.dsl.model.ConfigurationDependencyResolver;
+import org.mule.runtime.config.internal.registry.AbstractSpringRegistry;
+
+import io.qameta.allure.Description;
+import org.junit.Before;
+import org.junit.Test;
+
+
+public class DependencyGraphBeanDependencyResolverTestCase {
```
elrodro83, line 21: allure

modules/spring-config/src/test/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorterTestCase.java
```diff
@@ -0,0 +1,191 @@
...
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+import io.qameta.allure.Description;
+import org.junit.Before;
+import org.junit.Test;
+
+
+public class DependencyGraphLifecycleObjectSorterTestCase {
```
elrodro83, line 42: allure

modules/spring-config/src/test/java/org/mule/runtime/config/internal/dsl/model/ConfigurationDependencyResolverTestCase.java
```diff
@@ -0,0 +1,73 @@
...
+
+import com.google.common.collect.Sets;
+import io.qameta.allure.Description;
+import org.hamcrest.Matchers;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ConfigurationDependencyResolverTestCase {
+
+  ConfigurationDependencyResolver configurationDependencyResolver;
```
elrodro83, line 30: private

modules/spring-config/src/test/java/org/mule/runtime/config/internal/dsl/model/ConfigurationDependencyResolverTestCase.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Collection;
+import java.util.Set;
+
+import com.google.common.collect.Sets;
+import io.qameta.allure.Description;
+import org.hamcrest.Matchers;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ConfigurationDependencyResolverTestCase {
```
elrodro83, line 28: allure

elrodro83 (dismissed):

modules/spring-config/src/main/java/org/mule/runtime/config/internal/BeanWrapper.java
```diff
@@ -0,0 +1,48 @@
...
+package org.mule.runtime.config.internal;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * BeanWrapper helps the comparison/equality check among non-proxy vs. proxy or proxy vs. proxy when building a dependency graph.
+ */
+public class BeanWrapper {
+
+  final Object wrappedObject;
```
eze210, line 16: private

modules/spring-config/src/main/java/org/mule/runtime/config/internal/BeanWrapper.java
```diff
@@ -0,0 +1,48 @@
...
+  final String name;
```
eze210, line 17: private

modules/spring-config/src/main/java/org/mule/runtime/config/internal/BeanWrapper.java
```diff
@@ -0,0 +1,48 @@
...
+  public String getName() {
+    return name;
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (wrappedObject.equals(obj)) {
+      return true;
+    } else if (obj instanceof BeanWrapper) {
+      return this.getWrappedObject().hashCode() == ((BeanWrapper) obj).getWrappedObject().hashCode();
```
eze210, line 38: Inside the class, use the field instead of the getter

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
```diff
@@ -0,0 +1,191 @@
...
+ */
+public class DependencyGraphLifecycleObjectSorter implements LifecycleObjectSorter {
+
+  private List<DefaultDirectedGraph<BeanWrapper, DefaultEdge>> dependencyGraphs;
+  private DependencyGraphBeanDependencyResolver resolver;
+  protected Class<?>[] orderedLifecycleTypes;
+  private Map<String, Integer> lifecycleObjectNameOrderMap;
+
+  public DependencyGraphLifecycleObjectSorter(DependencyGraphBeanDependencyResolver resolver, Class<?>[] orderedLifecycleTypes) {
+    this.dependencyGraphs = new ArrayList<>();
```
eze210, line 51: Pass the orderedLifecycleTypes.length to the ArrayList constructor

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
```diff
@@ -0,0 +1,191 @@
...
+    this.resolver = resolver;
+    this.orderedLifecycleTypes = orderedLifecycleTypes;
```
eze210, line 53: QQ: Do you want to copy the array or just hold a reference? Take into account that the parameter is mutable and you are using the field in other methods of this class

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
```diff
@@ -0,0 +1,191 @@
...
+    for (int i = 0; i < orderedLifecycleTypes.length; i++) {
+      DefaultDirectedGraph<BeanWrapper, DefaultEdge> graph = new DefaultDirectedGraph(DefaultEdge.class);
```
eze210, line 55: new DefaultDirectedGraph<>

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
```diff
@@ -0,0 +1,191 @@
...
+      dependencyGraphs.add(graph);
+    }
+    this.lifecycleObjectNameOrderMap = new HashMap<>();
+  }
+
+  /**
+   * Building a single dependency graph(bucket) for each lifecycle type
+   * 
+   * @param beanName
```
eze210, line 64: Missing parameters jdoc

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
```diff
@@ -0,0 +1,191 @@
...
+  public void addObject(String beanName, Object currentObject) {
+
+    requireNonNull(currentObject, "currentObject cannot be null");
+
+    if (stream(orderedLifecycleTypes).noneMatch(x -> x.isInstance(currentObject))) {
+      return;
+    }
+
+    DefaultDirectedGraph<BeanWrapper, DefaultEdge> dependencyGraph = getDependencyGraphForLifecycleType(currentObject);
+    CycleDetector cycleDetector = new CycleDetector(dependencyGraph);
```
eze210, line 77: Provide the parameterized type to this generic

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
```diff
@@ -0,0 +1,191 @@
...
+
+    BeanWrapper currentVertex = new BeanWrapper(beanName, currentObject);
+    dependencyGraph.addVertex(currentVertex);
+
+    // get (direct) prerequisite objects for the current object
+    // List<Pair<String, Object>> prerequisiteObjects = resolver.getDirectBeanDependencies(beanName);
```
eze210, line 83: Remove commented code

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  /**
+   * Provides the index of the graph(bucket) the current object should be added to
+   * 
+   * @param currentObject
+   * @return
+   */
+  private int getDependencyGraphIndex(Object currentObject) {
+    Class<?> clazz = stream(orderedLifecycleTypes).filter(x -> x.isInstance(currentObject)).findFirst().get();
```
eze210, line 121: This method is walking through the orderedLifecycleTypes three times: stream() + asList() + indexOf().
Given you're working on performance, I think it would be preferable to write a single for loop.

If we decide to let it as is, please ensure the Optional returned by findFirst() is present before using get(), or use orElse() instead

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
hyerans: Just holding a reference would be fine. I will mark it as final.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
hyerans: Replaced it with a single for loop :)

modules/spring-config/src/main/java/org/mule/runtime/config/internal/BeanWrapper.java
```diff
@@ -0,0 +1,48 @@
...
+    } else if (obj instanceof BeanWrapper) {
+      return this.wrappedObject.hashCode() == ((BeanWrapper) obj).wrappedObject.hashCode();
+    } else {
+      return false;
+    }
+  }
+
+  @Override
+  public int hashCode() {
+    return wrappedObject.hashCode();
```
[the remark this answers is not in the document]
eze210: If you didn't test it, please create a ticket to do once the microbench ticket is done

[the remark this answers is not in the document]
eze210: If you create the ticket, please also add a TODO in this line with a reference to that ticket

eze210 (changes requested):

eze210:
Please deprecate SpringLifecycleObjectSorter, since you replaced it

modules/spring-config/src/main/java/org/mule/runtime/config/internal/BeanWrapper.java
```diff
@@ -0,0 +1,48 @@
...
+  public String getName() {
+    return name;
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (wrappedObject.equals(obj)) {
+      return true;
+    } else if (obj instanceof BeanWrapper) {
+      return this.wrappedObject.hashCode() == ((BeanWrapper) obj).wrappedObject.hashCode();
```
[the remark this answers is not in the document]
hyerans: this.wrappedObject.equals(((BeanWrapper) obj).wrappedObject) was not able to tell if two proxy objects are same when they wrap the same object. Using the hashCodes kind of solved this problem.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
```diff
@@ -0,0 +1,199 @@
...
+                                    String preReqName = prerequisite.getName();
+                                    Object preReqObject = prerequisite.getWrappedObject();
+
+                                    BeanWrapper preReqVertex = new BeanWrapper(preReqName, preReqObject);
+                                    dependencyGraph.addVertex(preReqVertex);
+
+                                    // remove any additional edge that creates cycle
+                                    if (!dependencyGraph.containsEdge(preReqVertex, current)) {
+                                      dependencyGraph.addEdge(current, preReqVertex);
+                                      if (cycleDetector.detectCycles()) {
```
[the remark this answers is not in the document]
hyerans: W-10704588

modules/spring-config/src/main/java/org/mule/runtime/config/internal/BeanWrapper.java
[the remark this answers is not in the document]
hyerans: [W-10704535](https://gus.lightning.force.com/a07EE00000ojT9BYAU)

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+  }
+
+  /**
+   * Provides the index of the graph(bucket) the current object should be added to
+   * 
+   * @param currentObject current object that is going to be added to the graph(bucket)
+   * @return index of the relevant dependency graph(bucket)
+   */
+  private int getDependencyGraphIndex(Object currentObject) {
```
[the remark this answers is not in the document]
pabloperalta: The list of lifecycle types is hardcoded by phase and each phase defines it's own list (initialisation phase has a list and dispose has another list). Initializable or Disposable are the last on each of it corresponding list while every other element implements the given class. Any given object should be in the earliest bucket possible. If no buckets match, the is always the last one that should be the "default" bucket (such as initializable). If the object doesn't even match that then it is of no interest for the phase (It is not initializable, we don't need to initialise it)

Disposable and Initializable will not be in the same orderedLifecycleTypes list.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/resolvers/DependencyGraphBeanDependencyResolver.java
```diff
@@ -0,0 +1,212 @@
...
+        .anyMatch(x -> x.equals(currentObject.getWrappedObject()))) {
+      return emptyList();
+    }
+    visitedComponentsForBuckets.get(bucketIndex).add(currentObject);
+
+    // If the component was already processed for any other buckets,
+    // it returns the direct dependencies that were saved before to build a dependency graph for the current bucket
+    if (visitedComponents.containsKey(currentObject)) {
+      return visitedComponents.get(currentObject);
+    }
```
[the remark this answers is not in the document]
pabloperalta: Buckets define an order that the graph does not. Some objects of certain types should be initialized before other objects of other types. We have no strategy to generate an edge for these dependencies so we use a bucket sort. Now, when we get an object as an input, we find the bucket it belongs to. Now we need to know which object this one depends on regardless of the bucket. Maybe this second object is actually a dependency for many other objects as well. Some of those might not end up in this bucket.
Each bucket has its own objects to take care of and uses a graph to do so. If an object "A" shows up many times as a dependency, when we get "A"'s dependencies we would get the same list over and over again so we try to do it only once and save the result to reuse it.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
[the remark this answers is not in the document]
asanguinetti: I see... I thought there was one list with the different types to check for each phase, when there is actually one list per phase with all the associated types that must be considered. My bad.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/resolvers/DependencyGraphBeanDependencyResolver.java
```diff
@@ -0,0 +1,212 @@
...
+
+  /**
+   * Provides all the dependencies/required components for the object provided
+   *
+   * @param beanName    the name of the bean to resolve dependencies
+   * @param bucketIndex the bucket(graph) we calculate the dependencies for
+   * @return a map with any relevant object as key, direct dependencies of the key as value
+   *
+   */
+  public Map<BeanWrapper, List<BeanWrapper>> getTransitiveDependencies(String beanName, int bucketIndex) {
```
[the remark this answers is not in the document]
pabloperalta: It is called one per object that fits the bucket. If the object has already been processed inside this bucket, then we don't need to calculate its dependencies so we return an empty list since all nodes and edges that could come from calculating should already be present in the graph.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/resolvers/DependencyGraphBeanDependencyResolver.java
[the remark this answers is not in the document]
hyerans: The original version creates buckets for classes that should be initialized and checked with for loop if each object from the lookup is the instance of the bucket. If object A is an instance of bucket 3 and 11, it will be in bucket 3 in the original implementation. After ordering them, all the required prerequisites for each object will be initialized repeatedly regardless of their existence. This is where we get 80K repeated objects during initialization phase.
 
At first, we made a single graph for each phase, the order of allowed classes still matters so we reintroduce the buckets. Also, there were some objects like MuleContext that could be the prerequisite of many objects across the buckets, even in the same bucket. Caching the result and checking the existence in the bucket helps reduce repetition.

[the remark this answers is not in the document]
asanguinetti: Ok, I think what confused me was not knowing whether we would be returning all dependencies regardless of whether they belong to the same bucket or even the same phase. Based on what you are saying it seems we are returning all of them.
So if we are on the initialization phase, and A is initializable and depends on B, but B is not initializable, we would be returning B as dependency anyway (even though we wouldn't need to initialize it). Is that right?
For this reason, the dependencies would work for all buckets (and maybe all phases too?).

modules/spring-config/src/main/java/org/mule/runtime/config/internal/resolvers/DependencyGraphBeanDependencyResolver.java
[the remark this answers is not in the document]
asanguinetti: Yes, you are right I forgot `getDirectBeanDependencies` would return an empty list if already processed. However if this happens very often for the same object (I mean, receiving an already processed object), maybe we could check that here too and avoid the creation of all these helper objects (mainly `processedComponents` and `queue`). But I wouldn't do it unless we can confirm that this is indeed called many times.

Have you considered the case in which it was processed but for a different bucket? in that case we would be actually rebuilding a result that could be cached, right? But maybe the memory footprint could be a concern in that case, unless we use a bounded cache. Again, I wouldn't change anything unless it can be confirmed that it makes a difference.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/BeanWrapper.java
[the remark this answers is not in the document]
asanguinetti: This doesn't really make sense, if the wrapped objects are the same then `this.wrappedObject.equals(((BeanWrapper) obj).wrappedObject)` should return true, because you would be comparing the wrapped objects, not the wrappers.
If using the `hashCode` changes anything then it means the issue is coming from the fact that two distinct objects may have the same `hashCode`, which is perfectly valid, but I don't think that we should make our equality check be dependant on that (because of the possible collisions that Eze mentioned).
Are you absolutely sure that you have already tried Eze's suggestion and not something else?

modules/spring-config/src/main/java/org/mule/runtime/config/internal/resolvers/DependencyGraphBeanDependencyResolver.java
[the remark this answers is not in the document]
hyerans: Yes, if A(initializable)->B(not initializable), then we will return that information as well because of the edge case A->B->C(initializable). In the case of A->B->C, we need to have the information that A indirectly relies C during the phase.

core/src/main/java/org/mule/runtime/core/internal/lifecycle/phases/LifecycleObjectSorter.java
```diff
@@ -33,9 +33,9 @@
   List<Object> getSortedObjects();
 
   /**
-   * Provides the order of objects as reference for initialise/dispose phases
+   * Provides objects that should be initialized during the initialise/dispose phases
    * 
    * @param lookupObjectsForLifecycle lifecycle object list which is ordered based on the type
    */
-  default void setLifeCycleObjectNameOrderMap(Map<String, Object> lookupObjectsForLifecycle) {};
+  default void setLifeCycleObjectList(Map<String, Object> lookupObjectsForLifecycle) {};
```
asanguinetti, line 40: This is not what I meant. Here you changed the name of the method but it is still receiving a `Map`, so the problem is still there. Also you removed the fact that the order of the components is important from the Javadoc.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
```diff
@@ -45,7 +44,7 @@
   private List<DefaultDirectedGraph<BeanWrapper, DefaultEdge>> dependencyGraphs;
   private DependencyGraphBeanDependencyResolver resolver;
   protected final Class<?>[] orderedLifecycleTypes;
-  private Map<String, Integer> lifecycleObjectNameOrderMap;
+  private List<String> lifecycleObjectList;
```
asanguinetti, line 47: This one was better off being a `Map` from component to priority (for tie-breaking). Having it internally as a list should be less performant because each look up is O(N) (and we probably need to do many of those).
The only scenario in which this could be better is if the Maps are very small (less than 5 elements) and the hash function of the keys is very expensive. In that case we can use a `SmallMap`.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/resolvers/DependencyGraphBeanDependencyResolver.java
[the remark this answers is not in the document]
hyerans: Yes, the duplicate check/caching is happening when calculating direct dependencies only at this moment. I will put caching the result of getTransitiveDependencies as subtask of benchmark ticket. Since we are going through the objects from the lookup and lookup doesn't have much repetition, getTransitiveDependencies won't be called many times for the same object.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
```diff
@@ -175,25 +175,25 @@ public int compare(BeanWrapper o1, BeanWrapper o2) {
...
+   *
    * @param lookupObjectsForLifecycle lifecycle object list which is ordered based on the type
    */
-  public void setLifeCycleObjectNameOrderMap(Map<String, Object> lookupObjectsForLifecycle) {
-    int index = 0;
+  public void setLifeCycleObjectList(Map<String, Object> lookupObjectsForLifecycle) {
     for (Map.Entry<String, Object> entry : lookupObjectsForLifecycle.entrySet()) {
       String objectName = entry.getKey();
-      lifecycleObjectNameOrderMap.put(objectName, index++);
+      lifecycleObjectList.add(objectName);
```
asanguinetti, line 186: Here we should convert from `List` of ordered components to `Map` from name to position (priority). What we were doing before was OK, but instead of iterating the `entrySet` we would be iterating the list that comes as parameter.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
```diff
@@ -150,8 +150,8 @@ private int getDependencyGraphIndex(Object currentObject) {
 
         @Override
         public int compare(BeanWrapper o1, BeanWrapper o2) {
-          if (getLifeCycleObjectNameOrderMap().getOrDefault(o1.getName(), -1) > getLifeCycleObjectNameOrderMap()
-              .getOrDefault(o2.getName(), -1)) {
+          if (getLifeCycleObjectList().indexOf(o1.getName()) > getLifeCycleObjectList()
+              .indexOf(o2.getName())) {
```
asanguinetti, line 154: doing the `indexOf` on an `ArrayList` here is O(N) now, vs. the previous O(1) when having the `Map`. We should revert to the previous internal representation.

asanguinetti (changes requested):

modules/spring-config/src/main/java/org/mule/runtime/config/internal/resolvers/DependencyGraphBeanDependencyResolver.java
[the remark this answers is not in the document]
asanguinetti: Works for me.

core/src/main/java/org/mule/runtime/core/internal/lifecycle/phases/LifecycleObjectSorter.java
hyerans: I reverted the change and changed the input parameter instead!

modules/spring-config/src/main/java/org/mule/runtime/config/internal/BeanWrapper.java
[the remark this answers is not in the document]
hyerans: In that case, I need help to move away from the hashcode comparison. If I use `this.wrappedObject.equals(((BeanWrapper) obj).wrappedObject)` as suggested, I have seen some actual cases where we cannot detect the same objects if they are in different forms. There are some examples in `BeanWrapperTestCase.java`, such as ```
sameProxyInWrapperComparisonTest
sameProxyInTwoDifferentWrappersComparisonTest
objectInProxyAndWrapperComparisonTest
```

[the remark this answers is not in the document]
Comment (asanguinetti) [reply modules/spring-config/src/main/java/org/mule/runtime/config/internal/BeanWrapper.java:38]: I think we need to understand what happens in the real case, because in the tests you are using a `FakeProxy` which has its own `equals` implementation which is delegating the equals against the real object directly. So, two different proxies that delegate to the same real object will have the same `hashCode` but will not compare equal. Not only that, I think your `FakeProxy` will not compare equal to itself (`x.equals(x)`), which violates the reflexivity property of the equals contract.

**Review by asanguinetti — APPROVED** [review]

**Review by eze210 — APPROVED** [review]

File: modules/spring-config/src/main/java/org/mule/runtime/config/internal/AutoDiscoveredDependencyResolver.java [hunk 43]
```diff
@@ -0,0 +1,31 @@
...
+
+import static java.util.stream.Collectors.toList;
+
+import org.mule.runtime.config.internal.registry.AbstractSpringRegistry;
+
+import java.util.List;
+
+public class AutoDiscoveredDependencyResolver {
```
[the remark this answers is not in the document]
hyerans: added javadoc

modules/spring-config/src/main/java/org/mule/runtime/config/internal/registry/SpringRegistryLifecycleManager.java
```diff
@@ -136,7 +141,19 @@ public void applyLifecycle(Object o) throws LifecycleException {
 
     @Override
     public LifecycleObjectSorter newLifecycleObjectSorter() {
-      return new SpringLifecycleObjectSorter(orderedLifecycleTypes, getSpringRegistry());
+      // if (!getSpringRegistry().springContextInitialised.get()) {
+      // return new SpringLifecycleObjectSorter(orderedLifecycleTypes, getSpringRegistry());
+      // } else {
```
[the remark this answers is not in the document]
hyerans: removed the commented code

modules/spring-config/src/main/java/org/mule/runtime/config/internal/AutoDiscoveredDependencyResolver.java
hyerans: all the resolvers are now in the resolvers package

modules/spring-config/src/main/java/org/mule/runtime/config/internal/BeanVertexWrapper.java
hyerans: Changed the name to BeanWrapper and the message is 'bean must not be null' now.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DeclaredDependencyResolver.java
hyerans: done!

core/src/main/java/org/mule/runtime/core/internal/lifecycle/phases/DefaultLifecycleObjectSorter.java
hyerans: remove unused import

core/src/main/java/org/mule/runtime/core/internal/lifecycle/phases/NotInLifecyclePhase.java
hyerans: removed unused import

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphBeanDependencyResolver.java
hyerans: added javadoc

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorter.java
hyerans: added javadoc

modules/spring-config/src/main/java/org/mule/runtime/config/internal/DependencyNode.java
hyerans: added javadoc

modules/spring-config/src/main/java/org/mule/runtime/config/internal/context/MuleArtifactContext.java
hyerans: rebase done for PR 11290

modules/spring-config/src/main/java/org/mule/runtime/config/internal/registry/BeanDependencyResolver.java
hyerans: added javadoc with `@since`

modules/spring-config/src/test/java/org/mule/runtime/config/internal/AutoDiscoveredDependencyResolverTestCase.java
hyerans: added allure feature/story/issue

modules/spring-config/src/test/java/org/mule/runtime/config/internal/BeanVertexWrapperTestCase.java
hyerans: added allure feature/story/issue

modules/spring-config/src/test/java/org/mule/runtime/config/internal/DeclaredDependencyResolverTestCase.java
hyerans: added allure feature/story/issue

modules/spring-config/src/test/java/org/mule/runtime/config/internal/DependencyGraphBeanDependencyResolverTestCase.java
hyerans: added allure feature/story/issue

modules/spring-config/src/test/java/org/mule/runtime/config/internal/DependencyGraphLifecycleObjectSorterTestCase.java
hyerans: added allure feature/story/issue

modules/spring-config/src/test/java/org/mule/runtime/config/internal/dsl/model/ConfigurationDependencyResolverTestCase.java
hyerans: added allure feature/story/issue

modules/spring-config/src/test/java/org/mule/runtime/config/internal/AutoDiscoveredDependencyResolverTestCase.java
hyerans: added private

modules/spring-config/src/test/java/org/mule/runtime/config/internal/dsl/model/ConfigurationDependencyResolverTestCase.java
hyerans: added private

PR #9338, opened 2020-09-02
MULE-18655: Decorate source response streams with data flow statistic…

fsgonz:
…s decorator

PR #9222, opened 2020-08-19
MULE-18649: Spike: Check how sources generating list of messages are

elrodro83:
handled

PR #11310, opened 2022-02-11
W-10619608: Remove support for Payload Statistics

asanguinetti:
### General Approach
This should be mostly a revert of the Epic [EE-7398](https://www.mulesoft.org/jira/browse/). However, reverting each of those changes manually one by one would take a long time unnecessarily because some were just refactors or fixes over the previous ones.
The approach taken was a little bit of a hybrid between reverts and some manual cleanup starting from some of the core classes and looking at their usage.
Another reason why we didn't want to go with full revert is that we needed to keep a few things for semantic versioning and for compatibility with current users like the monitoring team.

### Removing Payload Statistics
That being said. I started by removing the `PayloadStatisticsCursorComponentDecoratorFactory` and replacing its usages with the `NoOpCursorComponentDecoratorFactory`. This was more or less straightforward, however we were still leaving a lot of the underlying support lying around with no apparent purpose, additionally, the performance issues mentioned in #11252 were still there.

### The Big Revert
Next I made sure to at least revert these changes (not fully, but as much as possible):

- #9208
- #9335
- #9222
- #9306
- #9347
- #9338

I didn't revert changes on `ResolverUtils` or `MessageUtils`, nor I removed completely the `CursorDecoratorFactory` and `CursorComponentDecoratorFactory` (nor their NoOp implementations). I wanted to make sure there were no usages left (all related changes had been reverted) before removing them manually.
All these reverts I squashed them together in a single commit because it was too chaotic to analyse them separately for correctness.

### Final Cleanup and Adjustments
After that the only thing remaining was to finally remove the NoOp implementations and the interfaces that were not API.

### Notes for the Reviewers

- There are some things that I marked as deprecated because removing them could break semantic versioning. If you think it could be possible/worth removing them anyway, let me know.
- ~~I didn't revert anything from `ResolverUtils` because I thought those changes could be useful for something else. The changes are mostly related to adding an additional parameter to some methods which is a mapping function to apply to the value after resolution.~~ (This was eventually changed after suggestions from reviewers)
- ~~When doing final testing I realized that I had actually broken what was fixed in #11209 because I had completely removed `MessageUtils#decorateInput`, however there was this thing that was not related to the `CursorComponentDecoratorFactory`, which was wrapping the `CursorStream` inside an `UnclosableCursorStream`. I ended up bringing back the method and its usages, but without the need for the actual `componentDecoratorFactory` and `correlationId` parameters. I don't know if it makes too much sense to leave it like this, but I don't know if it was a good idea to incur into a bigger refactor inside a PR that is supposed to be about removing things only.~~ (Changed as suggested by reviewers)
- Finally, please check if it would make sense to keep anything of what I would be removing. I think that having everything on a single PR can help identifying that. If there is anything that we don't want to remove because it could be useful in the short term, let me know and I will update the PR or generate an alternative one. Keep in mind that if we wanted to keep the `CursorComponentDecoratorFactory` support, we would have to come up with a way to avoid the issues mentioned in #11252 (not the ones about performance, because we actually have a solution for that, but the ones concerning reusability). I don't know if it would be possible to save that by removing the `isContent` checks for example.


core/src/main/java/org/mule/runtime/core/internal/util/message/MessageUtils.java
```diff
@@ -517,31 +477,14 @@ private static Message toMessage(Result<?, ?> result, DataType dataType, Object
...
-      return componentDecoratorFactory.decorateInput((Collection) v, eventCorrelationId);
-    } else if (v instanceof Iterator) {
-      return componentDecoratorFactory.decorateInput((Iterator) v, eventCorrelationId);
-    } else if (v instanceof CursorStreamProvider) {
-      return new InputDecoratedCursorStreamProvider((CursorStreamProvider) v,
-                                                    componentDecoratorFactory,
-                                                    eventCorrelationId);
+  public static Object decorateInput(Object v) {
+    if (v instanceof CursorStream) {
+      return new UnclosableCursorStream((CursorStream) v);
```
asanguinetti, line 487: Had to leave this part because of #11209.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/objectbuilder/DefaultObjectBuilder.java
```diff
@@ -121,28 +116,12 @@ public boolean isDynamic() {
...
-                                                                      componentDecoratorFactory)
-                                                 : identity())
-                             : entry.getValue().isContent() && componentDecoratorFactory != null
-                                 ? mapTypeValue(resolvedValue,
-                                                v -> decorateInput(v, context.getEvent()
-                                                    .getCorrelationId(),
-                                                                   componentDecoratorFactory))
-                                 : resolvedValue);
+                         context == null || context.resolveCursors() ? resolveCursor(resolvedValue, MessageUtils::decorateInput)
+                             : resolvedValue);
```
asanguinetti, line 124: Note that now we are mapping with `MessageUtils#decorateInput` every time we do cursor resolution, regardless of the value being content or not. This is because `CursorStream`s need to be wrapped in `UnclosableCursorStreams` (see #11209). Please check if this is correct, because maybe I'm doing it in cases in which we shouldn't.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/objectbuilder/ParameterGroupObjectBuilder.java
```diff
@@ -101,22 +84,15 @@ public T build(ResolverSetResult result) throws MuleException {
...
-        final boolean isContent = field.isAnnotatedWith(Content.class)
-            || field.isAnnotatedWith(org.mule.sdk.api.annotation.param.Content.class);
-        Object resolvedValue = resolveValue(new StaticValueResolver<>(parameters
-            .apply(name)), context);
-        Object value = context == null || context.resolveCursors()
-            ? resolveCursor(resolvedValue,
-                            isContent ? v -> decorateInput(v, context.getEvent().getCorrelationId(), componentDecoratorFactory)
-                                : identity())
+        Object resolvedValue = resolveValue(new StaticValueResolver<>(parameters.apply(name)), context);
+        Object value = context == null || context.resolveCursors() ? resolveCursor(resolvedValue, MessageUtils::decorateInput)
```
asanguinetti, line 95: Note that now we are mapping with `MessageUtils#decorateInput` every time we do cursor resolution, regardless of the value being content or not. This is because `CursorStream`s need to be wrapped in `UnclosableCursorStream` (see #11209). Please check if this is correct, because maybe I'm doing it in cases in which we shouldn't.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/MethodArgumentResolverDelegate.java
```diff
@@ -336,28 +326,26 @@ private void initArgumentResolvers() {
...
       if (parameter.getParameterizedType() instanceof ParameterizedType) {
         Type generic = ((ParameterizedType) parameter.getParameterizedType()).getActualTypeArguments()[0];
         if (generic instanceof Class) {
           Class<?> genericClass = (Class<?>) generic;
-          if (CursorProvider.class.isAssignableFrom(genericClass) || Object.class.equals(genericClass)) {
-            resolver = new TypedValueCursorArgumentResolverDecorator(resolver, componentDecoratorFactory);
-          } else {
-            resolver = new TypedValueArgumentResolverDecorator(resolver, componentDecoratorFactory);
+          if (CursorProvider.class.isAssignableFrom(genericClass)
+              || InputStream.class.isAssignableFrom(genericClass)
```
asanguinetti, line 341: Had to add the `InputStream` to the classes that will be decorated with `TypedValueCursorArgumentResolverDecorator` because we want to decorate with `UnclosableCursorStream` (see #11209). Previously we had the `TypedValueArgumentResolverDecorator`, but I think we don't need to have a different one now, because it will be doing exactly the same thing, and also, we only need to care about `InputStreams` and not everything else. But please check if I'm mistaken.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/objectbuilder/DefaultObjectBuilder.java
marianogonzalez: pretty sure this case shuold not be wrapped. @dabreu @ndinu  can you confirm?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/objectbuilder/ParameterGroupObjectBuilder.java
marianogonzalez: same here @ndinu @dabreu

core/src/main/java/org/mule/runtime/core/internal/util/message/MessageUtils.java
marianogonzalez: if this is still needed then change signature to be more clear. It should be called asUncloseableStream and should explicitly receive an InputStream.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/MethodArgumentResolverDelegate.java
marianogonzalez: @ndinu @dabreu  please review this

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/MethodArgumentResolverDelegate.java
```diff
@@ -451,61 +435,25 @@ public String toString() {
...
+    public ObjectArgumentResolverDecorator(ArgumentResolver<Object> decoratee) {
       super(decoratee);
-      this.componentDecoratorFactory = componentDecoratorFactory;
     }
 
     @Override
-    protected Object decorate(Object value, String eventCorrelationId) {
-      return decorateInput(resolveCursor(value), eventCorrelationId, componentDecoratorFactory);
+    protected Object decorate(Object value) {
+      return resolveCursor(value, MessageUtils::decorateInput);
```
marianogonzalez, line 456: is this exactly as it was before introducing the the whole decorators thing? This lambda thing seems less performant than what we had in 4.3

marianogonzalez: review all remaining uses of `MessageUtils::decorateInput`. Make sure they're absolutelly necessary. This kind of lambda expressions and stack trace changes are expensive in this context.

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/MethodArgumentResolverDelegate.java
asanguinetti: I can answer that before the whole decorators thing, the `resolveCursor` did not have this variant that takes the additional `UnaryOperator` parameter. When I profiled this I saw some small performance penalty but not really big enough to be concerning. Of course, if we can remove it, that would be ideal.

The thing is, right now it is our way of doing the wrapping with `UnclosableCursorStream` without spreading the logic of `TypedValue` and `CursorProvider` resolution to `DefaultObjectBuilder`, `ParameterGroupObjectBuilder` and `MethodArgumentResolverDelegate`. But maybe we can do this decoration at a different, unique, place?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/objectbuilder/DefaultObjectBuilder.java
ndinu: I think that with the processing done in MethodArgumentResolverDelegate, the case is covered.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/objectbuilder/ParameterGroupObjectBuilder.java
ndinu: same answer as above.

core/src/main/java/org/mule/runtime/core/internal/util/message/MessageUtils.java
asanguinetti: I considered making it receive an InputStream (or even CursorStream) but, if we do that, then we would have to place the `instanceof` check on many different places (basically all places that are now calling this method). However they may be reduced if the wrapping is not necessary on `DefaultObjectBuilder` and `ParameterGroupObjectBuilder`.

I could even move this inside `MethodArgumentResolverDelegate`.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/objectbuilder/DefaultObjectBuilder.java
asanguinetti: That's great! I can also confirm that if I only do the wrapping on the `MethodArgumentResolverDelegate`, the test cases pass, but wasn't sure if it was a matter of lack of coverage.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/MethodArgumentResolverDelegate.java
ndinu: 👍

core/src/main/java/org/mule/runtime/core/internal/util/message/MessageUtils.java
marianogonzalez: exactly my point, this method should only be invoked when we know that the value may be a stream.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/objectbuilder/DefaultObjectBuilder.java
marianogonzalez: @ndinu can you point to/add that fills the gap between `i think` and `i'm sure`? :)

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/objectbuilder/ParameterGroupObjectBuilder.java
marianogonzalez: same reply

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/MethodArgumentResolverDelegate.java
marianogonzalez: This ties back to a previous comment in terms of being smarter about when to call this method. In any case, please remove the lambda. Lambdas are cool for tasks that execute sporadically. For code that executes constantly and concurrently, it adds unnecessary pressure on the GC and adds a penalty in stack jumps

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ResolverUtils.java
```diff
@@ -121,7 +122,14 @@ private ResolverUtils() {
    * @return the given {@code value} but converting a {@link CursorProvider} to a {@link Cursor} if any is present.
    */
   public static Object resolveCursor(Object value) {
-    return resolveCursor(value, identity());
+    // For performance reasons, we avoid delegating to the other version with the valueMapper parameter.
```
asanguinetti, line 125: Actually with this implementation I can remove the other versions that receive the `valueMapper` parameter, because they are no longer used. Do you think I should do it? or does it make sense to keep them in this utilitary class?
(My vote would be in favor of the removal)

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ResolverUtils.java
```diff
@@ -164,8 +172,21 @@ public static Object mapTypeValue(Object value, UnaryOperator valueMapper) {
...
 
-  public static Object resolveCursor(TypedValue<?> typedValue) {
-    return resolveCursor(typedValue, identity());
+  /**
+   * Obtains a {@link TypedValue} of {@link Cursor} based on the given {@code typedValue}, if one is available.
+   *
+   * @return the given {@code typedValue} but converting a {@link CursorProvider} to a {@link Cursor} if any is present.
+   */
+  public static TypedValue<?> resolveCursor(TypedValue<?> typedValue) {
+    // For performance reasons, we avoid delegating to the other version with the valueMapper parameter.
```
asanguinetti, line 181: same

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ResolverUtils.java
```diff
@@ -185,13 +206,68 @@ public static Object resolveCursor(TypedValue<?> typedValue, UnaryOperator value
...
+  private static TypedValue<?> typedValue(TypedValue<?> typedValue, UnaryOperator valueMapper, Object value) {
     return new TypedValue<>(valueMapper.apply(value), DataType.builder()
         .type(value != null ? value.getClass() : Object.class)
         .mediaType(typedValue.getDataType().getMediaType())
         .build(), typedValue.getByteLength());
   }
 
+  private static TypedValue<?> typedValue(TypedValue<?> typedValue, Object value) {
+    // Even though we could delegate to the other method that receives a valueMapper passing an identity, for performance
+    // reasons we prefer to have this other version
```
asanguinetti, line 264: same

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/objectbuilder/DefaultObjectBuilder.java
```diff
@@ -120,8 +119,7 @@ public T build(ValueResolvingContext context) throws MuleException {
       final Object resolvedValue = resolveValue(entry.getValue(), context);
 
       entry.getKey().set(object,
-                         context == null || context.resolveCursors() ? resolveCursor(resolvedValue, MessageUtils::decorateInput)
-                             : resolvedValue);
+                         context == null || context.resolveCursors() ? resolveCursor(resolvedValue) : resolvedValue);
```
asanguinetti, line 122: We still need to create a test that would demonstrate that decorating with `UnclosableCursorStream` would be necessary here too. Until then, I'm leaving it like this. But I'm working with @ndinu on this.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/objectbuilder/ParameterGroupObjectBuilder.java
```diff
@@ -92,8 +91,7 @@ private T doBuild(Predicate<String> hasParameter, Function<String, Object> param
       String name = field.getName();
       if (hasParameter.test(name)) {
         Object resolvedValue = resolveValue(new StaticValueResolver<>(parameters.apply(name)), context);
-        Object value = context == null || context.resolveCursors() ? resolveCursor(resolvedValue, MessageUtils::decorateInput)
-            : resolvedValue;
+        Object value = context == null || context.resolveCursors() ? resolveCursor(resolvedValue) : resolvedValue;
```
asanguinetti, line 94: same

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ResolverUtils.java
asanguinetti: The unused functions were removed

core/src/main/java/org/mule/runtime/core/internal/util/message/MessageUtils.java
asanguinetti: This method no longer exists

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/objectbuilder/DefaultObjectBuilder.java
asanguinetti: This issue has now been addressed in master by #11333. We added a few missing test cases and made sure the `UnclosableCursorStream` wrapping was done regardless of the parameter being annotated as `@Content` or not.

I rebased this PR on top of master and adjusted accordingly.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/objectbuilder/ParameterGroupObjectBuilder.java
asanguinetti: same as above.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/MethodArgumentResolverDelegate.java
asanguinetti: I changed it to no longer use a lambda for this. I created special versions of the cursor resolving methods that have logic specifically for this particular decoration. These should perform better due to inlining and also because, by knowing the exact type of mapping we are doing and the conditions in which it is done, we can be smarter about when and how to create new `TypedValue` wrappings.
The downside is we are mixing decoration logic with cursor resolution. Also, the code is a little bit longer.
I removed the versions of `resolveCursor` that took a mapping function as a parameter (which were created for the ABP support), to avoid the temptation of using them with lambdas, with the performance penalty that could mean.
I will be conducting a few performance tests today.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/objectbuilder/DefaultObjectBuilder.java
asanguinetti: Already solved on master by #11333. Rebased and adjusted accordingly.

elrodro83 (approved):

marianogonzalez (approved):

PR #11312, opened 2022-02-11
fix/W-10674245: Implement a runtime processing strategy for global er…

marianorean:
…ror handling

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/exception/GlobalErrorHandler.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/exception/GlobalErrorHandler.java
@@ -6,28 +6,42 @@
  */
 package org.mule.runtime.core.internal.exception;
 
-import org.mule.runtime.api.component.location.Location;
+import org.mule.runtime.api.lifecycle.InitialisationException;
 import org.mule.runtime.core.api.event.CoreEvent;
-import org.mule.runtime.core.privileged.exception.MessagingExceptionHandlerAcceptor;
 import org.mule.runtime.core.privileged.exception.TemplateOnErrorHandler;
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/mule/runtime/core/internal/exception/GlobalErrorHandler.java
```diff
@@ -6,28 +6,42 @@
...
-  public ErrorHandler createLocalErrorHandler(Location flowLocation) {
-    ErrorHandler local = new ErrorHandler();
-    local.setName(this.name);
-    local.setExceptionListeners(this.getExceptionListeners());
-    local.setExceptionListenersLocation(flowLocation);
-    return local;
+  @Override
+  public Consumer<Exception> router(Function<Publisher<CoreEvent>, Publisher<CoreEvent>> publisherPostProcessor,
+                                    Consumer<CoreEvent> continueCallback, Consumer<Throwable> propagateCallback) {
+    if (consumer == null) {
```
fsgonz, line 29: This has to be refactored. We are working on this.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/factories/ErrorHandlerFactoryBean.java
```diff
@@ -20,20 +22,30 @@
  */
 public class ErrorHandlerFactoryBean extends AbstractComponentFactory<ErrorHandler> {
 
+  private static Map<String, ErrorHandler> globalErrorHandlers = new HashMap<String, ErrorHandler>();
```
fsgonz, line 25: This has to be refactored. We are working on this.

marianorean:
see W-10888504

PR #11316, opened 2022-02-15
MULE-19513: Support new SDK API connectivity package.

SebaElizalde:


modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaConnectionProviderModelParserTestCase.java
```diff
@@ -157,6 +163,42 @@ public void noOauthCallbackValues() {
...
+  @Test
+  public void isPoolingConnectionProvider() {
+    mockConnectionProviderWithClass(PoolingTransactionalConnectionProvider.class);
+    assertThat(JavaConnectionProviderModelParserUtils.isPoolingConnectionProvider(connectionProviderElement), is(true));
+  }
+
+  @Test
+  public void isSdkPoolingConnectionProvider() {
+    mockConnectionProviderWithClass(SdkPoolingTransactionalConnectionProvider.class);
+    assertThat(JavaConnectionProviderModelParserUtils.isPoolingConnectionProvider(connectionProviderElement), is(true));
```
ndinu, line 175: I would test that the JavaConnectionProviderModelParser returns the correct management type for the CP, instead of testing the util method. (this also applies to the other tests)

The test is correct but as a unit test of the utils class.

ndinu (approved):

PR #11327, opened 2022-02-23
W-10780458: Allow execution of Mule operation

marianogonzalez:


marianogonzalez:
--validate

core/src/main/java/org/mule/runtime/core/api/artifact/ArtifactAstUtils.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.Set;
+
+import org.slf4j.Logger;
+
+/**
+ * Utilities for parsing and handling {@link ArtifactAst}
+ *
+ * @since 4.5.0
+ */
+public final class ArtifactAstUtils {
```
elrodro83, line 48: this should be moved to the proper module that requires this (maybe the name is too generic also).

elrodro83, line 48: why would this be api? Who needs this outside of the runtime?

core/src/main/java/org/mule/runtime/core/api/artifact/AstXmlParserSupplier.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.util.Set;
+
+/**
+ * A component capable of supplying {@link AstXmlParser} instances
+ *
+ * @since 4.5.0
+ */
+@FunctionalInterface
+public interface AstXmlParserSupplier {
```
elrodro83, line 20: why would this be api? Who needs this outside of the runtime?

core/src/main/java/org/mule/runtime/core/api/construct/Operation.java
```diff
@@ -0,0 +1,80 @@
...
+import org.mule.runtime.core.api.processor.Processor;
+
+import java.util.List;
+
+/**
+ * A strongly typed operation defined through the Mule language.
+ *
+ * @since 4.5.0
+ */
+public interface Operation extends ExecutableComponent, Lifecycle {
```
elrodro83, line 22: why would this be api? Who needs this outside of the runtime?

core/src/main/java/org/mule/runtime/core/internal/el/function/LookupFunction.java
```diff
@@ -82,11 +83,13 @@ public Object call(Object[] parameters, BindingContext context) {
     if (component instanceof Flow) {
       Message incomingMessage = lookupValue(context, MESSAGE, Message.builder().nullValue().build());
       Map<String, ?> incomingVariables = lookupValue(context, VARS, EMPTY_MAP);
+      Map<String, ?> incomingParams = lookupValue(context, PARAMS, EMPTY_MAP);
       Error incomingError = lookupValue(context, ERROR, null);
 
       Message message = Message.builder(incomingMessage).value(payload).mediaType(APPLICATION_JAVA).build();
       CoreEvent event = CoreEvent.builder(PrivilegedEvent.getCurrentEvent().getContext())
           .variables(incomingVariables)
+          .parameters(incomingParams)
```
elrodro83, line 92: this is similiar to a flow-ref. Are we sure we want to forward the params as well? This could generate coupling on the target flow to its caller.

modules/extensions-mule-support/src/test/java/org/mule/runtime/module/extension/mule/internal/operation/MuleOperationExecutionTestCase.java
```diff
@@ -0,0 +1,110 @@
...
+import org.mule.runtime.api.metadata.TypedValue;
+import org.mule.runtime.core.api.event.CoreEvent;
+
+import java.util.List;
+import java.util.Map;
+
+import io.qameta.allure.Description;
+import org.junit.Test;
+
+public class MuleOperationExecutionTestCase extends MuleArtifactFunctionalTestCase {
```
elrodro83, line 27: allure

tests/component-plugin/src/main/java/org/mule/functional/api/component/DumpInterceptedParametersProcessor.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.List;
+
+import javax.inject.Inject;
+
+/**
+ * Returns and clears all the parameters captured in the {@link ParameterInterceptorConfig}
+ *
+ * @since 4.5.0
+ */
+public class DumpInterceptedParametersProcessor extends AbstractComponent implements Processor {
```
elrodro83, line 24: is this being used? We should aim to use the test-extension instead of adding more stuff in here.

tests/component-plugin/src/main/java/org/mule/functional/api/extension/TestComponentExtensionLoadingDelegate.java
```diff
@@ -78,6 +74,8 @@ public void accept(ExtensionDeclarer extensionDeclarer, ExtensionLoadingContext
     declareProcessor(extensionDeclarer);
     declareInvocationCounter(extensionDeclarer);
     declareAssert(extensionDeclarer);
+    declareParameterInterceptor(extensionDeclarer);
+    declareDumpInterceptedParameters(extensionDeclarer);
```
elrodro83, line 78: is this being used? We should aim to use the test-extension instead of adding more stuff in here.

elrodro83 (approved):

core/src/main/java/org/mule/runtime/core/api/artifact/ArtifactAstUtils.java
marianogonzalez: this is needed in `spring-config` but also in `mule-test-functional` which is why I put it here

core/pom.xml
```diff
@@ -103,6 +103,10 @@
             <groupId>org.mule.runtime</groupId>
             <artifactId>mule-module-http-policy-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.mule.runtime</groupId>
+            <artifactId>mule-artifact-ast-xml-parser</artifactId>
```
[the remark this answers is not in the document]
marianogonzalez: where exactly?

core/src/main/java/org/mule/runtime/core/api/artifact/ArtifactAstUtils.java
marianogonzalez: because otherwise I get `NCD` when running tests

core/src/main/java/org/mule/runtime/core/api/construct/Operation.java
marianogonzalez: For consistency with `Flow` mainly...

core/src/main/java/org/mule/runtime/core/internal/el/function/LookupFunction.java
marianogonzalez: you're right. We should actually add a validator to disalllow flow-ref inside operations

tests/src/main/resources/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,31 @@
+Manifest-Version: 1.0
```
elrodro83, line 1: what is this file for?

elrodro83 (changes requested):

elrodro83 (approved):

asanguinetti (approved):

PR #11333, opened 2022-02-25
Decorating CursorStreams inside ParameterGroups as unclosable.

asanguinetti:


ndinu (approved):

asanguinetti:
Conducted performance validations using the HTTP default proxy scenario with 10 executions, it can be seen here:
https://stats.perf.anypoint.mulesoft.com/goto/2a055df8db9480e9a613928bced8e9e3

There are no performance degradations, actually the results were better from the code of the PR but it may be due to the intrinsic variability of the test.

PR #11339, opened 2022-03-03
W-10619610: Migrate LoggingSupportabilityContextTestCase from system-…

sofiamorseletto:
…tests

### DIFF
--- a/modules/deployment/src/test/java/org/mule/runtime/module/deployment/logging/LoggingSupportabilityContextTestCase.java
+++ b/modules/deployment/src/test/java/org/mule/runtime/module/deployment/logging/LoggingSupportabilityContextTestCase.java
@@ -7,8 +7,11 @@
 package org.mule.runtime.module.deployment.logging;
 
 import static java.util.Arrays.asList;
+
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasItem;
 import static uk.org.lidalia.slf4jtest.LoggingEvent.error;
+import static uk.org.lidalia.slf4jtest.TestLoggerFactory.getTestLogger;
 
 import org.mule.runtime.core.internal.processor.LoggerMessageProcessor;
 import org.mule.runtime.module.deployment.impl.internal.builder.ApplicationFileBuilder;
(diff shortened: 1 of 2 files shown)


modules/deployment/src/test/java/org/mule/runtime/module/deployment/logging/LoggingSupportabilityContextTestCase.java
```diff
@@ -0,0 +1,60 @@
...
+import io.qameta.allure.Description;
+import static org.hamcrest.Matchers.is;
+import org.junit.Test;
+import org.junit.runners.Parameterized;
+import uk.org.lidalia.slf4jtest.TestLogger;
+import uk.org.lidalia.slf4jtest.TestLoggerFactory;
+
+public class LoggingSupportabilityContextTestCase extends AbstractApplicationDeploymentTestCase {
+
+  TestLogger logger = TestLoggerFactory.getTestLogger(LoggerMessageProcessor.class);
```
asanguinetti, line 29: Static import

modules/deployment/src/test/java/org/mule/runtime/module/deployment/logging/LoggingSupportabilityContextTestCase.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Parameterized.Parameters(name = "Parallel: {0}")
+  public static List<Boolean> params() {
+    return asList(false);
+  }
```
asanguinetti, line 34: It wouldn't work if you removed this and made the constructor receive no arguments, and just passed false to the constructor of the super class?

asanguinetti (changes requested):

modules/deployment/src/test/java/org/mule/runtime/module/deployment/logging/LoggingSupportabilityContextTestCase.java
```diff
@@ -0,0 +1,60 @@
...
+        appFileBuilder("logging-app").definedBy("logging-supportability-app.xml");
+
+    addExplodedAppFromBuilder(loggingAppFileBuilder);
+
+    startDeployment();
+    assertApplicationDeploymentSuccess(applicationDeploymentListener, loggingAppFileBuilder.getId());
+
+    executeApplicationFlow("logging", correlationId);
+
+    assertThat(logger.getAllLoggingEvents(), is(asList(error(mdcMap, "I'm a logger"))));
```
asanguinetti, line 58: Wouldn't it be better to do something like:
`assertThat(logger.getAllLoggingEvents(), hasItem(error(mdcMap, "I'm a logger")))`?
This should make the test more stable against features/changes that could add some more logging on the same class.
After all, the test is only trying to check that the supportability information is present, not that it is the only thing there.

modules/deployment/src/test/java/org/mule/runtime/module/deployment/logging/LoggingSupportabilityContextTestCase.java
asanguinetti: Just checked and no, it wouldn't work.

asanguinetti (approved):

PR #11345, opened 2022-03-09
W-10618278: Create a way to trigger for memory manager events

fsgonz:


modules/memory-management/src/main/java/org/mule/runtime/internal/memory/bytebuffer/ThreadPoolBasedByteBufferProvider.java
```diff
@@ -70,6 +89,8 @@ private boolean isPowerOfTwo(int valueToCheck) {
 
   @Override
   public ByteBuffer allocate(int size) {
+    allocationDataProducer
+        .triggerProfilingEvent(new DefaultByteBufferProviderEventContext(name, System.currentTimeMillis(), size));
```
eze210, line 93: static import

eze210 (changes requested):

eze210 (dismissed):

eze210 (approved):

PR #11346, opened 2022-03-10
W-10619583: Redefine Mule DSL syntax for optional parameter.

asanguinetti:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/delegate/ParameterModelsLoaderDelegate.java
```diff
@@ -49,11 +49,11 @@ public ParameterModelsLoaderDelegate(Supplier<StereotypeModelLoaderDelegate> ste
         group.getDisplayModel().ifPresent(groupDeclarer::withDisplayModel);
         group.getLayoutModel().ifPresent(groupDeclarer::withLayout);
         group.getAdditionalModelProperties().forEach(groupDeclarer::withModelProperty);
-        group.getExclusiveOptionals().ifPresent(descriptor -> groupDeclarer
-            .withExclusiveOptionals(descriptor.getExclusiveOptionals(), descriptor.isOneRequired()));
         groupDeclarer.getDeclaration().setDescription(group.getDescription());
         group.getLayoutModel().ifPresent(groupDeclarer::withLayout);
       }
+      group.getExclusiveOptionals().ifPresent(descriptor -> groupDeclarer
+          .withExclusiveOptionals(descriptor.getExclusiveOptionals(), descriptor.isOneRequired()));
```
asanguinetti, line 56: This part needs to be checked. I had to change it because otherwise, the exclusive optionals wouldn't be added to the extension model for parameters that belong to the default group. But in the new Mule SDK all parameters got to the default group.
I'm not sure whether this could have an unexpected impact when using other SDK implementations. Also, it smells that I'm changing this only for the exclusive optionals... what about the other things?

asanguinetti:
--validate

modules/extensions-mule-support/src/test/java/org/mule/runtime/module/extension/mule/internal/validation/InvalidOperationInvocationTestCase.java
```diff
@@ -0,0 +1,75 @@
...
+    expectedException
+        .expectMessage("[validation/salutation-exclusive-parameters-operation.xml:11]: "
+            + "Element <helloPlace>, the following parameters cannot be set at the same time: [city, country].");
+    parseConfig("validation/salutation-exclusive-parameters-operation.xml");
+  }
+
+  @Test
+  public void missingRequiredExclusiveParametersOperation() throws Exception {
+    // FIXME W-10831629: there is an inconsistency between the element referred to by this error message and the one where there
+    // are conflicting exclusive optionals
```
asanguinetti, line 48: Notice that a new ticket has been created for this issue found: [W-10831629](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000sWyAZYA0/view)

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/validation/InvalidExtensionConfigTestCase.java
```diff
@@ -142,6 +142,8 @@ public void petStoreExclusiveGroupInsidePojoOneRequired() throws Exception {
 
   @Test
   public void petStoreExclusiveParameterRequired() throws Exception {
+    // FIXME W-10831629: there is an inconsistency between the element referred to by this error message and the one where there
+    // are conflicting exclusive optionals
```
asanguinetti, line 146: The fix for [W-10831629](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000sWyAZYA0/view) should also affect this case.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkParameterGroupModelParser.java
```diff
@@ -20,26 +23,49 @@
...
-    return unmodifiableList(parameters.stream()
-        .map(p -> new MuleSdkParameterModelParserSdk(p, typeLoader))
-        .collect(toList()));
+  private List<ParameterModelParser> doParserParameters(ComponentAst parametersComponent, TypeLoader typeLoader) {
+    Stream<ParameterModelParser> parameterParsers = getChildren(parametersComponent, "parameter")
+        .map(p -> new MuleSdkParameterModelParserSdk(p, typeLoader));
+    Stream<ParameterModelParser> optionalParameterParsers = getChildren(parametersComponent, "optional-parameter")
+        .map(p -> new MuleSdkOptionalParameterModelParserSdk(p, typeLoader));
+
+    return unmodifiableList(concat(parameterParsers, optionalParameterParsers).collect(toList()));
```
marianogonzalez, line 53: remove the unmodifiableList

modules/extensions-mule-support/src/test/resources/salutation-operations-def.xml
```diff
@@ -104,4 +98,24 @@
...
+            <operation:optional-parameter name="city" type="string" summary="The name of a city to say hello to"/>
+            <operation:optional-parameter name="country" type="string" summary="The name of a country to say hello to"/>
+            <operation:exclusive-optionals exclusiveOptionals="city,country" oneRequired="true"/>
+        </operation:parameters>
+        <operation:output>
+            <operation:payload-type type="string"/>
+            <operation:attributes-type type="string"/>
+        </operation:output>
+        <operation:body>
+            <this:hello-world greeting="#[params.greeting]"
```
marianogonzalez, line 113: there should be a test which calls this operation correctly and it executes well. The params passed to hello-world are incorrect and that will never happen

modules/extensions-mule-support/src/test/resources/salutation-operations-def.xml
```diff
@@ -104,4 +98,24 @@
...
+                              receiver="#[params.receiver]"
+                              target="helloWorld"/>
+
+            <set-payload value="#[vars.helloWorld]"/>
```
marianogonzalez, line 117: remove

modules/extensions-mule-support/src/test/resources/salutation-operations-def.xml
```diff
@@ -104,4 +98,24 @@
...
+            <operation:exclusive-optionals exclusiveOptionals="city,country" oneRequired="true"/>
+        </operation:parameters>
+        <operation:output>
+            <operation:payload-type type="string"/>
+            <operation:attributes-type type="string"/>
+        </operation:output>
+        <operation:body>
+            <this:hello-world greeting="#[params.greeting]"
+                              receiver="#[params.receiver]"
+                              target="helloWorld"/>
```
marianogonzalez, line 115: remove

modules/extensions-mule-support/src/test/resources/validation/salutation-exclusive-parameters-operation.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<mule xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xmlns:this="http://www.mulesoft.org/schema/mule/this"
+      xmlns="http://www.mulesoft.org/schema/mule/core"
+      xsi:schemaLocation="http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd
+        http://www.mulesoft.org/schema/mule/this http://www.mulesoft.org/schema/mule/this/current/mule-this.xsd">
+
+    <import file="salutation-operations-def.xml"/>
+
+    <flow name="composedOperation">
```
marianogonzalez, line 10: use miningful name

modules/extensions-mule-support/src/test/java/org/mule/runtime/module/extension/mule/internal/validation/InvalidOperationInvocationTestCase.java
```diff
@@ -0,0 +1,75 @@
...
+
+import io.qameta.allure.Feature;
+import io.qameta.allure.Story;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+@Feature(REUSE)
+@Story(OPERATIONS)
+public class InvalidOperationInvocationTestCase extends MuleArtifactFunctionalTestCase {
```
marianogonzalez, line 24: rename to something that refers to exclusive optionals. You won't be able to test additional invalid operations in the same class

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/delegate/ParameterModelsLoaderDelegate.java
marianogonzalez: this change looks good to me. Java SDK doesn't offer a syntax for defining EO in the default group and XML SDK doesn't support the concept at all

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/MuleOperationExtensionModelDeclarer.java
```diff
@@ -183,86 +183,94 @@ private DisplayModel display(String displayName, String summary, String example)
   }
 
   private void addParametersDeclaration(ConstructDeclarer def) {
-    final NestedComponentDeclarer parametersDef = def.withOptionalComponent("parameters")
+    final NestedComponentDeclarer<?, ?> parametersDef = def.withOptionalComponent("parameters")
```
marianogonzalez, line 186: those generics add no value

marianogonzalez (changes requested):

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkParameterGroupModelParser.java
asanguinetti: Note that the method [getParameterParsers](https://github.com/mulesoft/mule/blob/c89b364e48e2f806ce6374109fbefd94e67f327a/modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkParameterGroupModelParser.java#L83) will be returning the reference to the `parameters` field unwrapped.

I assumed this wrapping was done here originally to protect it from external modifications (because the toList collector does not provide any immutability guarantees). Since we don't need to modify it internally either, we could have it wrapped as unmodifiable at this point instead of doing it before returning it from the `getParametersParsers` method.

If you still think the wrapping is not necessary I would remove it. Or I could move it to the getter.

modules/extensions-mule-support/src/test/resources/salutation-operations-def.xml
asanguinetti: fixed and added test.

asanguinetti: fixed and added test.

modules/extensions-mule-support/src/test/resources/salutation-operations-def.xml
asanguinetti: removed

asanguinetti: removed

modules/extensions-mule-support/src/test/java/org/mule/runtime/module/extension/mule/internal/validation/InvalidOperationInvocationTestCase.java
asanguinetti: renamed to `InvalidExclusiveOptionalsOperationInvocationTestCase`.

modules/extensions-mule-support/src/test/resources/mule-salutation-operations-config.xml
```diff
@@ -23,4 +23,8 @@
         <test-components:dump-intercepted-values />
     </flow>
 
+    <flow name="exclusiveOptionalsFlow">
+        <this:hello-place greeting="good morning" city="London"/>
```
marianogonzalez, line 27: do you use the parameters

marianogonzalez (changes requested):

marianogonzalez (approved):

PR #11354, opened 2022-03-15
W-10619741: Fix flaky test ExtensionMessageSourceTestCase

eze210:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -263,17 +263,19 @@ private void stopSource() throws MuleException {
   }
 
   private RestartContext stopSource(boolean restarting) throws MuleException {
-    if (sourceAdapter != null) {
-      final String sourceName = sourceAdapter.getName();
+    // Using an auxiliary variable so that it isn't nullified by another thread in the middle of the "if" body.
+    final SourceAdapter auxAdapter = sourceAdapter;
```
marianogonzalez, line 267: aren't you just hiding the symptom instead of the root cause? How come another thread can nullify this while a stop is in progress?

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
marianogonzalez: that doesn't answer the question

marianogonzalez: also.. if there's a race condition that nullifies the variable, this change doesn't prevent that

eze210: The change wasn't trying to avoid the race condition, it just was ensuring that the auxAdapter is null xor not-null during the whole method. I was just trying to avoid a NPE

The threads that are calling this method are
1. The thread that it's executing the test: https://github.com/mulesoft/mule/blob/ccda76a4ac6e3922c91126f051228c84a0d470e2/modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSourceTestCase.java#L416
2. When the startSource fails, it stops, disposes, and nullifies the source, and the failing tests are testing the situation of an async start that fails

So a better fix for this is to remove the messageSource.stop() call in the tests that are testing this situation...

I'll commit that change instead

PR #11355, opened 2022-03-15
W-10634129: SDK connector without connector configuration fail (it is non-mandatory)

Bardammu:
First fix attempt

### DIFF
--- a/modules/extensions-xml-support/src/test/resources/flows/flow-using-module-implicit-configuration.xml
+++ b/modules/extensions-xml-support/src/test/resources/flows/flow-using-module-implicit-configuration.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<mule xmlns="http://www.mulesoft.org/schema/mule/core"
+      xmlns:module-implicit-configuration="http://www.mulesoft.org/schema/mule/module-implicit-configuration"
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xsi:schemaLocation="
+       http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd
+       http://www.mulesoft.org/schema/mule/module-implicit-configuration http://www.mulesoft.org/schema/mule/module-implicit-configuration/current/mule-module-implicit-configuration.xsd">
+
+<!--    <module-implicit-configuration:config name="alpingoconfig" />-->
+
+    <flow name="testSetPayloadConfigDefaultParamFlow">
+<!--    <module-implicit-configuration:set-payload-config-default-param-value config-ref="alpingoconfig"/>-->
+        <module-implicit-configuration:set-payload-config-default-param-value />
+    </flow>
+
+</mule>
--- a/modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/extension/xml/MacroExpansionModuleModel.java
+++ b/modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/extension/xml/MacroExpansionModuleModel.java
@@ -118,8 +118,9 @@ public class MacroExpansionModuleModel {
     final List<ComponentModel> moduleGlobalElements = getModuleGlobalElements();
     final Set<String> moduleGlobalElementsNames =
         moduleGlobalElements.stream().map(ComponentModel::getNameAttribute).collect(toSet());
-    expandOperations(moduleGlobalElementsNames);
+
     expandGlobalElements(moduleGlobalElements, moduleGlobalElementsNames);
+    expandOperations(moduleGlobalElementsNames);
   }
 
   private void expandOperations(Set<String> moduleGlobalElementsNames) {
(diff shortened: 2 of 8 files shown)


modules/extensions-xml-support/src/test/resources/flows/flow-using-module-implicit-configuration.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dabreu, line 1: I think that this file is not being used and can be removed, right?

modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/extension/xml/MacroExpansionModuleModel.java
```diff
@@ -312,11 +339,15 @@ private ComponentModel createModuleOperationChain(ComponentModel operationRefMod
 
     operationRefModel.getMetadata().getSourceCode().ifPresent(processorChainBuilder::setSourceCode);
 
+    final Optional<String> implicitConfigRef =
```
dabreu, line 342: The var name implicitConfigRef might be confusing here, given that not always an implicit config will be used. It might be changed to a more generic name, i.e. configRef

modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/extension/xml/MacroExpansionModuleModel.java
```diff
@@ -213,6 +214,24 @@ private void addDefaultGlobalElements(Set<String> moduleGlobalElementsNames) {
   }
 
   private void macroExpandGlobalElements(List<ComponentModel> moduleComponentModels, Set<String> moduleGlobalElementsNames) {
+    if (!hasExplicitConfiguration(applicationModel, extensionModel)
+        && extensionModel.getConfigurationModel(MODULE_CONFIG_GLOBAL_ELEMENT_NAME).isPresent()) {
+      ComponentModel configModel = new ComponentModel.Builder()
```
dabreu, line 219: Please correct me if my understanding is not correct, but with this check, the implicit configuration would be created even if there are no operations on the application requiring it, right? I was wondering whether it might be possible to refine the check to create the implicit config only when required, when the following conditions are met
- !hasExplicitConfiguration AND
- extensionModel.getConfigurationModel(MODULE_CONFIG_GLOBAL_ELEMENT_NAME).isPresent() 
- there is at least one operation within the app which requires a configuration (by inspecting its ext model?) but no config-ref is defined for it (implicit config)
Not sure if the last check could be added at this point, but I think that I would avoid creating unnecessary implicit configs which might affect existing apps.

ndinu: +1

Bardammu: > with this check, the implicit configuration would be created even if there are no operations on the application requiring it, right?

If the extension model has a configuration with name 'config' it means the xml conector has properties. This config is being added by the declarer.

I understand properties are properties of a configuration. In fact, if the operation does not have a 'config-ref' parameter then properties are not accesible.

So, my reasoning is the following:  if the app has a configuration with name 'config' and it does not have an explicit configuration we already know an implicit config is needed.

Maybe we can discus this again on meet.

dabreu: Thanks @Bardammu for the explanation, I see your point. 
My point is whether it is worth to create the implicit config even if there are no operations within the application that require a config (it might be the case that some operations on the extension require a config and some others don't). We can meet to discuss whether it is worth to consider/handle this scenario.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/extension/xml/MacroExpansionModuleModel.java
```diff
@@ -240,6 +259,17 @@ private void macroExpandGlobalElements(List<ComponentModel> moduleComponentModel
     });
   }
 
+  private boolean existOperationThatUsesImplicitConfiguration() {
+    return applicationModel.getRootComponentModel().getInnerComponents().stream()
+        .filter(componentModel -> componentModel.getIdentifier().getName().equals("flow")
```
ndinu, line 264: I would extract the flow and sub-flow constants to a static list.

modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/extension/xml/MacroExpansionModuleModel.java
```diff
@@ -240,6 +263,17 @@ private void macroExpandGlobalElements(List<ComponentModel> moduleComponentModel
     });
   }
 
+  private boolean existOperationThatUsesImplicitConfiguration() {
+    return applicationModel.getRootComponentModel().getInnerComponents().stream()
+        .filter(componentModel -> componentModel.getIdentifier().getName().equals(FLOW_ELEMENT_IDENTIFIER)
+            || componentModel.getIdentifier().getName().equals(SUBFLOW_ELEMENT_IDENTIFIER))
+        .flatMap(componentModel -> componentModel.getInnerComponents().stream())
```
ndinu, line 270: Won't this miss the cases where the use of the xml sdk operation is inside a scope, like an async or try scope?

modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/extension/xml/MacroExpansionModuleModel.java
```diff
@@ -268,10 +268,23 @@ private boolean existOperationThatUsesImplicitConfiguration() {
...
+
+  private boolean existOperationThatUsesImplicitConfiguration(ComponentModel componentModel) {
+    if (componentModel.getIdentifier().getNamespace().equals(extensionModel.getXmlDslModel().getPrefix()) &&
+        componentModel.getModel(OperationModel.class).isPresent() &&
+        !componentModel.getRawParameters().keySet().contains(MODULE_OPERATION_CONFIG_REF)) {
+      return true;
+    }
+
+    for (ComponentModel childComponent : componentModel.getInnerComponents()) {
+      if (existOperationThatUsesImplicitConfiguration(childComponent)) {
```
ndinu, line 282: Is there anyway of this recursion to have a cycle?

ndinu (approved):

PR #11357, opened 2022-03-16
W-10619908: Runtime should dispatch ExceptionNotification when ExtensionMessageSource fails to start

Bardammu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -370,13 +372,21 @@ private void onReconnectionSuccessful() {
...
     LOGGER.error(format("Message source '%s' on flow '%s' could not be reconnected. Will be shutdown. %s",
                         sourceModel.getName(), getLocation().getRootContainerName(), exception.getMessage()),
                  exception);
     shutdown();
     reconnecting.set(false);
   }
 
+  private void dispatchExceptionNotification(Throwable exception) {
+    SystemExceptionHandler systemExceptionHandler = muleContext.getExceptionListener();
+    RetryPolicyExhaustedException retryPolicyExhaustedException =
```
marianogonzalez, line 385: check for the exception not already being of this type

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -370,13 +372,21 @@ private void onReconnectionSuccessful() {
...
   private void onReconnectionFailed(Throwable exception) {
+    dispatchExceptionNotification(exception);
     LOGGER.error(format("Message source '%s' on flow '%s' could not be reconnected. Will be shutdown. %s",
                         sourceModel.getName(), getLocation().getRootContainerName(), exception.getMessage()),
                  exception);
     shutdown();
     reconnecting.set(false);
   }
 
+  private void dispatchExceptionNotification(Throwable exception) {
```
marianogonzalez, line 383: firing notifications is something that the exception handler might do. Rename this method to describe what it's actually doing which is invoking the system handler

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -212,12 +212,14 @@ private void startSource(boolean restarting, RestartContext restartContext) thro
     Consumer<Throwable> onFailure;
     if (retryPolicyTemplate.isAsync()) {
       onSuccess = this::onReconnectionSuccessful;
-      onFailure = this::onReconnectionFailed;
+      onFailure = (t) -> {
+        dispatchExceptionNotification(t);
```
ndinu, line 216: in this case, is the dispatch method being called twice? once here, and another time in the first line of the onReconnectionFailed method.

Bardammu: Yes, it is called twice on the async case. I'll fix it

Bardammu:
--validate

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSourceTestCase.java
```diff
@@ -193,6 +200,33 @@ public void failToStart() throws Exception {
...
+    registerNotificationListener(listener);
+
+    ConnectionException connectionException = new ConnectionException(ERROR_MESSAGE);
+    MuleException e = new DefaultMuleException(connectionException);
+    doThrow(e).when(source).onStart(any());
+
+    messageSource.initialise();
+    try {
+      messageSource.start();
+      latch.await();
```
marianogonzalez, line 220: set timeout

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSourceTestCase.java
```diff
@@ -193,6 +200,33 @@ public void failToStart() throws Exception {
...
+      assertThat(notifications.get(0), is(instanceOf(ExceptionNotification.class)));
+      assertThat(notifications.get(0).getException(), instanceOf(RetryPolicyExhaustedException.class));
+    } catch (Exception ex) {
+      latch.await();
```
marianogonzalez, line 224: who will release the latch if  start() failed?  If I'm wrong about this, add timeout

marianogonzalez (changes requested):

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSourceTestCase.java
Bardammu: If everything works the exception will be thrown after the the call to the system handler, so the latch will be released. I'm adding the timeout in case a future bug makes the exception be thrown before the call to the system handler, so the tests fails quickly

marianogonzalez (approved):

PR #11363, opened 2022-03-17
W-10622707: destroy() is not being called for MuleArtifactContext

elrodro83:


eze210:
Sounds good, but it's called from here: https://github.com/mulesoft/mule-spring-module/blob/master/module/src/main/java/org/mule/extension/spring/api/SpringConfig.java#L165
Have you reviewed that?

elrodro83:
that's just a delegate to close in latest Spring versions, but since the spring module has to work with previous versions of spring, there is no problem with that.

fsgonz (approved):

PR #11375, opened 2022-03-21
W-10617943: Fix namespace read while picking up connector error type as seen for SAP S4hana Odata connector

mule-pachanady:
We notice some test failures in  Sap - S4hana Odata Mule 4. 
```
+ Running test: config-null-base-url-test - Cause a CONNECTIVITY Error. Thrown +
+ when the base url is null.                                                   +
java.lang.AssertionError: The error ID thrown does not match the expected one.  expected:<[S4HANA]:CONNECTIVITY> but was:<[MULE]:CONNECTIVITY>
```
This is due to recent changes to `ExceptionHandlerManager -> resolveConnectionErrorType()` 11/2/21, where a namespace name S4 HANA does not match expected type S4HANA

` "S4HANA:CONNECTIVITY"`

### DIFF
--- a/modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/exception/ExceptionHandlerManagerTestCase.java
+++ b/modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/exception/ExceptionHandlerManagerTestCase.java
@@ -15,6 +15,8 @@ import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.when;
 import static org.mule.test.module.extension.internal.util.ExtensionsTestUtils.mockExceptionEnricher;
 
+import io.qameta.allure.Description;
+import io.qameta.allure.Issue;
 import org.mule.runtime.api.component.ComponentIdentifier;
 import org.mule.runtime.api.connection.ConnectionException;
 import org.mule.runtime.api.exception.ErrorTypeRepository;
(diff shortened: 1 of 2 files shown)


mule-pachanady:
@dabreu I think you worked on related code here, doing this change to fix some regression issue in Sap4 HANA connector. Please review when you get a chance.

dabreu (approved):

asanguinetti (changes requested):
Please add a test covering this case so we make sure it doesn't break in the future.

mule-pachanady:
@asanguinetti Please review

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/exception/ExceptionHandlerManagerTestCase.java
```diff
@@ -154,6 +159,28 @@ public void handleConnectionExceptionExtensionWithCustomConnectivityError() {
     assertThat(((ConnectionException) throwable).getErrorType().get().getNamespace(), is(EXTENSION_NAMESPACE));
   }
 
+  @Test
```
asanguinetti, line 162: Please add an Allure annotation relating the new test to the issue that required its creation.

Like this:
`@Issue("W-10617943")`

You can also add a more human friendly description to the test with:
`@Description("Your description here")`

mule-pachanady:
--validate

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/exception/ExceptionHandlerManagerTestCase.java
mule-pachanady: Done Thanks

asanguinetti (approved):

PR #11386, opened 2022-03-23
W-10682121 - Poll frequency of 60000 Milliseconds not working as expected

dabreu:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/spring/ComponentConfigurationBuilder.java
```diff
@@ -481,6 +479,12 @@ public void onMultipleValues(KeyAttributeDefinitionPair[] definitions) {
     }
   }
 
+  private boolean shouldSetDefaultFrequencyForFixedFrequencyStrategy(String parameterName, ComponentParameterAst parameter) {
+    return "frequency".equals(parameterName)
+        && ownerComponent.getIdentifier().equals(FIXED_FREQUENCY_STRATEGY_IDENTIFIER)
+        && parameter.isDefaultValue() && parameter.getResolvedRawValue() == null;
```
Bardammu, line 485: Shouldn't parameter.getResolvedRawValue() be different from `null`?

Bardammu (approved):

ndinu (approved):

modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/spring/ComponentConfigurationBuilder.java
dabreu: It returns `null` when no frequency value is explicitly set on the` fixed-frequency `component

PR #11394, opened 2022-03-30
W-10818447: Verify that ErrorHandlerFactoryBean does not generate dup…

marianorean:
…licate bean when defined as prototype

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/exception/GlobalErrorHandler.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/exception/GlobalErrorHandler.java
@@ -7,27 +7,35 @@
 package org.mule.runtime.core.internal.exception;
 
 import org.mule.runtime.api.component.location.Location;
+import org.mule.runtime.api.exception.MuleException;
 import org.mule.runtime.core.api.event.CoreEvent;
-import org.mule.runtime.core.privileged.exception.MessagingExceptionHandlerAcceptor;
-import org.mule.runtime.core.privileged.exception.TemplateOnErrorHandler;
 import org.reactivestreams.Publisher;
 
-import java.util.List;
-
 import static java.lang.Boolean.getBoolean;
-import static java.util.stream.Collectors.toList;
 import static org.mule.runtime.api.util.MuleSystemProperties.REVERT_SIGLETON_ERROR_HANDLER_PROPERTY;
 
 public class GlobalErrorHandler extends ErrorHandler {
+
+  private static final boolean IS_PROTOTYPE = getBoolean(REVERT_SIGLETON_ERROR_HANDLER_PROPERTY);
+
+  // We need to keep a reference to one of the local error handlers to be able to stop its inner processors.
+  ErrorHandler local;
 
   @Override
   public Publisher<CoreEvent> apply(Exception exception) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/mule/runtime/core/internal/exception/LocalErrorHandler.java
```diff
@@ -0,0 +1,26 @@
...
+package org.mule.runtime.core.internal.exception;
+
+import org.mule.runtime.api.exception.MuleException;
+
+public class LocalErrorHandler extends ErrorHandler {
+
+  @Override
+  public void start() throws MuleException {
+    if (logger.isDebugEnabled()) {
+      logger.debug("In start of local error handler: {}", this);
```
fsgonz, line 16: Maybe indicate that the exceptionListeners will not be started/stopped because they refer to a global error handler.
Would "Starting local error handler. Nothing will be done as this depends on a global error handler."

fsgonz (changes requested):

core/src/main/java/org/mule/runtime/core/internal/exception/GlobalErrorHandler.java
```diff
@@ -24,7 +24,7 @@
   }
 
   public ErrorHandler createLocalErrorHandler(Location flowLocation) {
-    ErrorHandler local = new ErrorHandler();
+    ErrorHandler local = new LocalErrorHandler();
```
fsgonz, line 27: Notice that you also have to use the system property here because with the former behavior the exceptionListeners should be stopped.

fsgonz (changes requested):

core/src/main/java/org/mule/runtime/core/internal/exception/GlobalErrorHandler.java
```diff
@@ -7,27 +7,45 @@
...
+import org.mule.runtime.api.exception.MuleException;
 import org.mule.runtime.core.api.event.CoreEvent;
-import org.mule.runtime.core.privileged.exception.MessagingExceptionHandlerAcceptor;
-import org.mule.runtime.core.privileged.exception.TemplateOnErrorHandler;
 import org.reactivestreams.Publisher;
 
-import java.util.List;
-
-import static java.util.stream.Collectors.toList;
+import static java.lang.Boolean.getBoolean;
```
fsgonz, line 14: order of imports.

core/src/main/java/org/mule/runtime/core/internal/exception/GlobalErrorHandler.java
```diff
@@ -7,27 +7,45 @@
...
+  // We need to keep a reference to one of the local error handlers to be able to stop its inner processors.
+  ErrorHandler local;
+
   @Override
   public Publisher<CoreEvent> apply(Exception exception) {
     throw new IllegalStateException("GlobalErrorHandlers should be used only as template for local ErrorHandlers");
   }
 
+  @Override
+  public void stop() throws MuleException {
```
fsgonz, line 30: Can't we just let the GlobalErrorHandler close the exceptionListeners itself?

fsgonz (changes requested):

fsgonz (approved):

fsgonz (approved):

marianorean:
--validate

PR #11405, opened 2022-04-04
W-10881459: Replace slf4j-test library with new version with TCCL check

sofiamorseletto:
This PR also includes the following changes:
- `MemoryAppender` and `MemoryAppenderResource` classes were deleted since they are no longer used
- The thread reference clearance done on LogCleanup is no longer needed due to changes on slf4j-test, so it was removed


modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/ClassloadingTroubleshootingTestCase.java
```diff
@@ -359,7 +360,7 @@ public String describeFailure() {
         });
   }
 
-  private List<String> toMessages(List<LoggingEvent> loggingEvents) {
+  private List<String> toMessages(ImmutableList<LoggingEvent> loggingEvents) {
```
asanguinetti, line 363: I don't see what we gain by doing this change. Yes, we are not going to modify the list passed as a parameter, but we don't **need** it to be immutable either.
By changing it like this we are forcing the callers to take care of wrapping their list if they don't have it already immutable, when in fact the current implementation will return a new list, so the original one will remain safe from modifications.
I'd say, let's keep it as abstract as possible.

modules/deployment/src/test/java/org/mule/runtime/module/deployment/logging/LoggingSeparationTestCase.java
```diff
@@ -55,13 +63,13 @@ public void whenLoggingEachAppShouldLogOnItsOwnThreadSuccessfully() throws Excep
     executeApplicationFlow("logging");
 
     assertThat(logger.getAllLoggingEvents().size(), is(1));
-    // TODO: Change thread name check to TCCL check when available on slf4j-test
-    assertThat("", logger.getAllLoggingEvents().get(0).getThreadName(), stringContainsInOrder("logging-app-1"));
+    assertThat("", ((MuleArtifactClassLoader) logger.getAllLoggingEvents().get(0).getThreadContextClassLoader()).getArtifactId(),
+               stringContainsInOrder("logging-app-1"));
```
asanguinetti, line 67: I'm just noticing that we were using the `assertThat` version that takes a reason as the first parameter but we are passing it an empty string. Why is that?
If we are not interested in putting a custom reason string, let's use the version that takes only the actual and the matcher.

modules/deployment/src/test/java/org/mule/runtime/module/deployment/logging/LoggingSeparationTestCase.java
```diff
@@ -55,13 +63,13 @@ public void whenLoggingEachAppShouldLogOnItsOwnThreadSuccessfully() throws Excep
...
 
     executeApplicationFlow("logging2", null, 1);
 
     assertThat(logger.getAllLoggingEvents().size(), is(2));
-    // TODO: Change thread name check to TCCL check when available on slf4j-test
-    assertThat("", logger.getAllLoggingEvents().get(1).getThreadName(), stringContainsInOrder("logging-app-2"));
+    assertThat("", ((MuleArtifactClassLoader) logger.getAllLoggingEvents().get(1).getThreadContextClassLoader()).getArtifactId(),
+               stringContainsInOrder("logging-app-2"));
```
asanguinetti, line 73: Same.

modules/deployment/src/test/java/org/mule/runtime/module/deployment/logging/LoggingSeparationTestCase.java
```diff
@@ -6,24 +6,32 @@
...
+import com.github.valfirst.slf4jtest.TestLogger;
 import io.qameta.allure.Description;
+import io.qameta.allure.Feature;
+import io.qameta.allure.Story;
 import org.junit.Test;
 import org.junit.runners.Parameterized;
-import uk.org.lidalia.slf4jtest.TestLogger;
 
+@Feature(INTEGRATIONS_TESTS)
+@Story(LOGGER)
```
asanguinetti, line 34: I'm not sure we are supposed to do this. The `LOGGER` story is part of the `CORE_COMPONENTS` feature. I don't know what happens if we mix these. Besides, I'm not sure what kind of feature `INTEGRATION_TESTS` would be. I'd prefer to use `CORE_COMPONENTS` as the feature here.

modules/deployment/src/test/java/org/mule/runtime/module/deployment/logging/LoggingSupportabilityContextTestCase.java
```diff
@@ -20,11 +23,15 @@
...
+import com.github.valfirst.slf4jtest.TestLogger;
 import io.qameta.allure.Description;
+import io.qameta.allure.Feature;
+import io.qameta.allure.Story;
 import org.junit.Test;
 import org.junit.runners.Parameterized;
-import uk.org.lidalia.slf4jtest.TestLogger;
 
+@Feature(INTEGRATIONS_TESTS)
+@Story(LOGGER)
```
asanguinetti, line 34: Same.

tests/unit/src/main/java/org/mule/tck/junit4/rule/LogCleanup.java
```diff
@@ -32,20 +27,20 @@ protected void before() throws Throwable {
...
   @Override
   protected void after() {
-    clearLogsAndMDCThreadReferences();
+    clearAllLogs();
   }
 
   /**
    * Logs that are stored for later assert need to be cleared before every test
-   * 
+   * <p>
```
asanguinetti, line 35: This opening tag without matching closure looks suspicious. Please check.

modules/core-mvel/src/main/java/org/mule/runtime/core/internal/el/mvel/MVELExpressionExecutor.java
```diff
@@ -48,7 +54,16 @@ public MVELExpressionExecutor(final ParserConfiguration parserConfiguration) {
...
+    // Since the DynamicOptimizer class grabs the TCCL, it could grab the app's classloader and that could lead to memory leakages
+    Thread currentThread = currentThread();
+    ClassLoader originalClassLoader = currentThread.getContextClassLoader();
+    ClassLoader newClassLoader = MVELExpressionLanguage.class.getClassLoader();
+    setContextClassLoader(currentThread, originalClassLoader, newClassLoader);
+    try {
+      OptimizerFactory.setDefaultOptimizer(OptimizerFactory.SAFE_REFLECTIVE);
+    } finally {
+      setContextClassLoader(currentThread, newClassLoader, originalClassLoader);
+    }
```
asanguinetti, line 66: I would move this to another PR. The reason being the following: this is a change in productive code, everything else affects only the tests. If there happens to be a bug in production with this change, it may require a revert followed by an alternative fix. By tying everything together into a single commit you are making it harder to revert cleanly.
Additionally, I would suggest adding a test that breaks without this fix that would show how the CL is potentially leaked. I know we already have a test that fails in such case but the problem is that test only fails if it is executed first in the test case, which is currently not guaranteed.
Finally, consider using `ClassUtils.withContextClassLoader` for readability and to reduce duplication.

asanguinetti (changes requested):

tests/unit/src/main/java/org/mule/tck/junit4/rule/LogCleanup.java
sofiamorseletto: This tag can be used on its own, without a closing tag according to the Java Doc [documentation](https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#format), to separate paragraphs

modules/core-mvel/src/main/java/org/mule/runtime/core/internal/el/mvel/MVELExpressionExecutor.java
sofiamorseletto: Got it, I'll move this change to another PR. I didn't use `ClassUtils.withContextClassLoader` since it's deprecated and using `setContextClassloader` instead is [recommended](https://github.com/mulesoft/mule/blob/a604d3e17ffab48397bb19ffbb4de27d714807d0/core/src/main/java/org/mule/runtime/core/api/util/ClassUtils.java#L967).

tests/unit/src/main/java/org/mule/tck/junit4/rule/LogCleanup.java
asanguinetti: Sorry, I didn't know that.

asanguinetti (approved):

PR #11411, opened 2022-04-06
W-10719167: XML SDK modules cannot access configs defined at the domain level.

SebaElizalde:


SebaElizalde:
--validate

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpansionModuleModel.java
```diff
@@ -478,9 +478,18 @@ private Map<String, String> extractProperties(Optional<String> configRefName) {
               .filter(componentModel -> componentModel.getModel(ConfigurationModel.class).isPresent()
                   && configParameter.equals(componentModel.getComponentId().orElse(null)))
               .findFirst()
-              .orElseThrow(() -> new IllegalArgumentException(format("There's no <%s:config> named [%s] in the current mule app",
-                                                                     extensionModel.getXmlDslModel().getPrefix(),
-                                                                     configParameter)));
+              .orElseGet(() -> { // Else look for the element in the parent ast (e.g. the domain of the app)
```
ndinu, line 481: Should this be recursive? Or just check 1 level ?

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpansionModuleModel.java
```diff
@@ -478,9 +478,18 @@ private Map<String, String> extractProperties(Optional<String> configRefName) {
...
+                Optional<ArtifactAst> parent = applicationModel.getParent();
+                if (parent.isPresent()) {
+                  return parent.get()
```
ndinu, line 484: I think that the whole condition/processing should be modularized so that the code to check the app and the domain is not duplicated.

ndinu (changes requested):

ndinu (approved):

PR #11413, opened 2022-04-07
W-10618624: Pass config-ref's as parameters.

asanguinetti:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DeferredConfigurationProvider.java
```diff
@@ -0,0 +1,120 @@
...
+   * @param operationModel the model for the component that requires the configuration.
+   * @return A {@link ConfigurationModel} suitable for the given {@link ComponentModel}.
+   */
+  private static ConfigurationModel getConfigurationModelForComponent(ExtensionModel extensionModel,
+                                                                      ComponentModel operationModel) {
+    return extensionModel.getConfigurationModels().stream()
+        .filter(cm -> cm.getOperationModels().stream().anyMatch(om -> om.getName().equals(operationModel.getName())))
+        .findAny()
+        .orElse(null);
+  }
```
asanguinetti, line 44: Is this really valid?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessorBuilder.java
```diff
@@ -153,9 +157,36 @@ protected ResolverSet getArgumentsResolverSet() throws ConfigurationException {
   }
 
   protected ConfigurationProvider getConfigurationProvider() {
-    return parameters.values().stream()
-        .filter(v -> v instanceof ConfigurationProvider)
-        .map(v -> ((ConfigurationProvider) v)).findAny()
-        .orElse(configurationProvider);
+    ConfigurationProvider explicitConfigProvider = getConfigurationProvider(parameters.get("config-ref"));
```
asanguinetti, line 160: Still need to find a better way of finding out for which parameters do we need to do all the `ValueResolver` considerations.

modules/extensions-spring-support/src/test/resources/petstore-expression-config-ref.xml
```diff
@@ -0,0 +1,34 @@
...
+            <petstore:pet value="Skye"/>
+        </petstore:pets>
+    </petstore:config>
+
+    <flow name="getPetsWithReference">
+        <petstore:get-pets ownerName="john" config-ref="paw-patrol-store"/>
+    </flow>
+
+    <flow name="getPetsWithExpression">
+        <petstore:get-pets ownerName="john" config-ref="#[app.registry[vars.storeName ++ '-store']]"/>
```
asanguinetti, line 31: I think it should not be necessary to access the registry, right? Probably the result of the expression should be automatically used as a bean reference. What do you think?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DeferredConfigurationProvider.java
marianogonzalez: i don't get the purpose of this. The same operation can be compatible with N different configs. Why is this relevant at all since at the end of the day it's the user the one that should be resolving the config through the expression?

marianogonzalez (changes requested):

asanguinetti:
--validate

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DeferredConfigurationProvider.java
asanguinetti: I did this because the `ConfigurationProvider` has a method for getting the `ConfigurationModel`. That method is being called during initialization for validation purposes (more specifically [here](https://github.com/mulesoft/mule/blob/aa340afc2ec146c7e1f5706bf2285bf76f9d4269/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/OperationMessageProcessor.java#L134)).
I wasn't sure whether a config-ref for an operation could be compatible with configs of different `ConfigurationModels`.
But now I should see if I can do something else about this. Probably that validation may have to be skipped at that stage, if an expression was used.

marianogonzalez: THat's a good point, but you won't be able to do that validation anyway right? Also, the `getConfigurationModel()` implementation that you have in this class is not accurate as you have no way of being sure if that's the configuraitonModel that you'll get. What it's clear though is that the validation that you pointed out should only apply for static config-refs.

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
```diff
@@ -588,12 +622,18 @@ private MetadataContext getMetadataContext(MetadataCache cache)
...
                                                               INVALID_CONFIGURATION));
 
-        if (configurationProvider instanceof DynamicConfigurationProvider) {
+        if (configurationProviderResolver.dependsOnEvent()) {
+          throw new MetadataResolvingException("Configuration used for Metadata fetch cannot depend on Event context",
+                                               INVALID_CONFIGURATION);
+        }
+
+        Optional<ConfigurationProvider> configurationProvider = resolveConfigurationProvider(configurationProviderResolver);
+        if (!configurationProvider.isPresent() || configurationProvider.get() instanceof DynamicConfigurationProvider) {
```
asanguinetti, line 636: Shouldn't this `instanceof` check be replaced by `ConfigurationProvider#isDynamic`?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessorBuilder.java
```diff
@@ -152,10 +155,21 @@ public ComponentMessageProcessorBuilder<M, P> setTerminationTimeout(long termina
...
+  private ConfigurationProviderResolverWrapper getConfigurationProviderResolver(Object configRefParameter) {
+    if (configRefParameter instanceof ValueResolver) {
+      return new ConfigurationProviderResolverWrapper((ValueResolver<ConfigurationProvider>) configRefParameter);
+    }
+
+    if (configRefParameter instanceof ConfigurationProvider) {
+      return new ConfigurationProviderResolverWrapper((ConfigurationProvider) configRefParameter);
+    }
+
+    return null;
```
asanguinetti, line 173: I'm seeing if I can actually move this to an earlier stage so we already receive a ConfigProviderResolver and we can revert back to the original logic (only checking for a `ConfigurationProviderResolver` instead of `ConfigurationProvider`).

Maybe I can do that at the `ComponentConfigurationBuilder` which is the one that seems to be creating the resolvers.

asanguinetti (commented):
Some comments for the reviewers

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
asanguinetti: Changed to `ValueResolver#isDynamic`.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DeferredConfigurationProvider.java
```diff
@@ -0,0 +1,120 @@
...
+
+  /**
+   * {@inheritDoc}
+   *
+   * @return {@code false}
+   */
+  @Override
+  public boolean isDynamic() {
+    return true;
```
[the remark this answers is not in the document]
asanguinetti: Changed to `ValueResolver` instead of `ConfigurationProvider`.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DeferredConfigurationProvider.java
```diff
@@ -0,0 +1,120 @@
...
+   * Creates a new instance
+   *
+   * @param extensionModel        the model that owns the {@link ConfigurationModel} for the instances that are going to be
+   *                              provided.
+   * @param componentModel        the model for the component that requires the configuration, used to determine the right
+   *                              {@link ConfigurationModel}.
+   * @param configurationResolver the {@link ValueResolver} that should resolve to a {@link ConfigurationProvider} instance.
+   * @param expressionManager     the {@link ExpressionManager} used to create a session used to evaluate the attributes.
+   */
+  public DeferredConfigurationProvider(ExtensionModel extensionModel,
```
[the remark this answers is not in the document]
asanguinetti: The whole implementation has changed for a `ValueResolver` instead of a `ConfigurationProvider`, so I'm resolving this.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessorBuilder.java
asanguinetti: There is a more up to date [comment](https://github.com/mulesoft/mule/pull/11413#discussion_r863039104) which is related to this issue. I'm marking this one as resolved so we can track it in the other one.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DeferredConfigurationProvider.java
asanguinetti: The implementation has changed. We are using a `ValueResolver` now, which does not require implementing `getConfigurationModel`.
The validation mentioned in the discussion can now be skipped if the configuration is dynamic.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/TypeSafeValueResolverWrapper.java
```diff
@@ -70,6 +71,8 @@ public void initialise() throws InitialisationException {
     if (!valueResolverDelegate.isDynamic() && !InputStream.class.isAssignableFrom(expectedType)) {
       resolver = new CachedResolver(resolver);
     }
+
+    initialiseIfNeeded(valueResolverDelegate, true, muleContext);
```
asanguinetti, line 75: Should we be worried about adding this here?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -646,6 +652,7 @@ protected ParameterValueResolver getParameterValueResolver() {
 
   @Override
   protected void doInitialise() throws InitialisationException {
+    validateConfigurationProviderIsNotExpression();
```
asanguinetti, line 655: Since I removed this validation from the base class (`ExtensionComponent`) I had to move it here.

asanguinetti (commented):
Some new comments for the reviewers.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/TypeSafeValueResolverWrapper.java
ndinu: Is this because the RegistryLookupValueResolverWrapper is not getting initialized/injected ? Maybe you can do it when instantiating it.

ndinu (dismissed):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
```diff
@@ -252,23 +283,27 @@ private void initConfigurationResolver() {
...
-      };
+      // the optional should be always present at this point because we already checked dependsOnEvent
+      configurationResolver = event -> resolveConfigFromProvider(getConfigurationProvider().get(), event);
     } else {
       // obtain implicit instance
       configurationResolver = event -> extensionManager.getConfiguration(extensionModel, componentModel, event);
     }
   }
 
+  private Optional<ConfigurationInstance> resolveConfigFromProvider(ConfigurationProvider configurationProvider,
```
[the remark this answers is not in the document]
asanguinetti: This one was born as a refactor to avoid repeating the logic for the resolution of the `ConfigurationInstance` from a `ConfigurationProvider`.
We used to have something like this:
```
configurationResolver = event -> {
        ConfigurationInstance instance = configurationProvider.get().get(event);
        if (instance == null) {
          throw new IllegalModelDefinitionException(format(
                                                           "Root component '%s' contains a reference to config '%s' but it doesn't exists",
                                                           getLocation().getRootContainerName(),
                                                           configurationProvider));
        }

        return of(instance);
      };
```
Where `configurationResolver` is of type `Function<CoreEvent, Optional<ConfigurationInstance>>`.
Now we just do this:
```
configurationResolver = event -> resolveConfigFromProvider(getConfigurationProvider().get(), event);
```
It seems the `configurationResolver` (which existed before this PR) needs to return an `Optional<ConfigurationInstance>` mostly because the configuration may be optional for the `ExtensionComponent`.
That being said, you are right in that we could have this new method return a ConfigurationInstance and just wrap on the lambda, like this:
```
configurationResolver = event -> of(resolveConfigFromProvider(getConfigurationProvider().get(), event));
```
But we would have to add a similar wrapping a few lines above that one, in the lambda for the dynamic `ValueResolver<ConfigurationProvider>` case. We can definitely do that if you think it will be clearer. I'm not completely decided for one or the other.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/resolver/ValueResolverFactory.java
```diff
@@ -71,7 +75,7 @@ public <T> ValueResolver<T> of(String parameterName, MetadataType expectedType,
 
     if (isExpression(value)) {
       final String expression = (String) value;
-      resolver = getExpressionBasedValueResolver(expectedType, expression, modelProperties, expectedClass);
+      resolver = getExpressionBasedValueResolver(expectedType, expression, modelProperties, acceptsReferences, expectedClass);
```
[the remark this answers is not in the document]
asanguinetti: Exactly, initially I was considering making the change for any parameter that would accept references and support expressions, but then we decided to just make it for the config-refs to avoid messing with too many things at once. 
So, if you check inside `getExpressionBasedValueResolver` you'll see I also added a condition to make it only valid for `ValueResolvers` with `ConfigurationProvider` as `expectedClass`.
```
} else if (acceptsReferences && expectedClass.equals(ConfigurationProvider.class)) {
```
With this in mind, I don't know if the `acceptsReferences` check is redundant now, because config-ref parameters [always accept references](https://github.com/mulesoft/mule-extensions-api/blob/03d928d87c88a46e02a089df2db5bf0ab7fe4bb2/mule-extensions-api/src/main/java/org/mule/runtime/extension/internal/loader/enricher/ConfigRefDeclarationEnricher.java#L111). But I thought it wouldn't hurt either and might protect us from issues down the line if we ever decide to have some other parameter that would accept a `ConfigurationProvider` but not necessarily a reference. For example: would it be possible to write a Java SDK operation that takes a parameter (not config-ref) with `ConfigurationProvider` as expected class, which supports expressions but does not accept references?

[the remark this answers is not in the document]
ndinu: I think it is not possible to declare such parameter, so it is fine 👌

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/TypeSafeValueResolverWrapper.java
ndinu: I think that the `initialiseIfNeeded` does not take into account if it was already initialized.

asanguinetti:
Had to rebase to fix conflict with the serialized Extension Model JSON of the MetadataExtension which is new.

ndinu (approved):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
marianogonzalez, line 294: this method name is not clear. It's not just resolving config from provider, there's an underlying assumption that this method is only to be used for dynamic configs.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
```diff
@@ -252,23 +283,27 @@ private void initConfigurationResolver() {
...
+                                                                    CoreEvent event) {
+    ConfigurationInstance instance = configurationProvider.get(event);
+    if (instance == null) {
```
marianogonzalez, line 297: A configProvider for a static config will never return null. If the configProvider points to an expression, then the error message doesn't make sense

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessorBuilder.java
```diff
@@ -152,10 +155,21 @@ public ComponentMessageProcessorBuilder<M, P> setTerminationTimeout(long termina
     return this;
   }
 
-  protected ConfigurationProvider getConfigurationProvider() {
-    return parameters.values().stream()
-        .filter(v -> v instanceof ConfigurationProvider)
-        .map(v -> ((ConfigurationProvider) v)).findAny()
-        .orElse(configurationProvider);
+  protected ValueResolver<ConfigurationProvider> getConfigurationProviderResolver() {
+    // Uses the configurationProvider given to the builder if any, otherwise evaluates the parameters.
```
marianogonzalez, line 159: how many times is this method called on any given instance? Shouldn't this be cached?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/RegistryLookupValueResolverWrapper.java
```diff
@@ -0,0 +1,87 @@
...
+   * @return the registry value associated with the key resolved from {@link #delegate}
+   * @throws MuleException          if an error occurred fetching the value
+   * @throws ConfigurationException if no object is registered under the key resolved from {@link #delegate}
+   */
+  @Override
+  public T resolve(ValueResolvingContext context) throws MuleException {
+    String key = delegate.resolve(context);
+    checkArgument(!isBlank(key), "A null or empty key was provided. Registry lookup cannot be performed with a blank key");
+
+    return registry.<T>lookupByName(key)
```
marianogonzalez, line 63: this should be cached. For any given key, the registry should be accessed only once

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/RegistryLookupValueResolverWrapper.java
```diff
@@ -0,0 +1,87 @@
...
+  private final ValueResolver<String> delegate;
+  private MuleContext muleContext;
+  private Registry registry;
+
+  /**
+   * Construct a new instance and set the {@link ValueResolver} of the key that will be used to access the registry
+   *
+   * @param delegate a resolver for a not blank {@link String}
+   */
+  public RegistryLookupValueResolverWrapper(ValueResolver<String> delegate) {
```
marianogonzalez, line 46: `delegate` -> `keysResolver`

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
asanguinetti: Actually it is currently used both for static and dynamic configs, because when the expression is resolved and we have the `ConfigurationProvider`, it can be any of those.
I'm not sure if it would be helpful to create different versions for static and dynamic configs only to avoid the null check for static ones.
The only cases in which we don't go here is when there is no expression in the config-ref **and** the ConfigurationProvider is static, or when using implicit configs.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessorBuilder.java
asanguinetti: This one is called once per call to `createMessageProcessor` on the builders. So I guess it would only make sense to cache it if we are using the same builder to create multiple processors from the same parameters.
Yet it doesn't seem to be a very costly operation. It is a lookup on a map to get the config-ref parameter, then a few `instanceof` checks and at most one construction of a `StaticValueResolver` which seems pretty cheap.
I could easily store the result in the builder instance and reuse it, if you think it makes sense.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
asanguinetti: I added some comments to make it clearer that the method may be used for both static and dynamic config providers. I still think the name of the method is representative of what it does. Maybe we can change it to `getConfigFromProvider` if that sounds better. Basically that's all it does, but with some added error handling (which I don't even know if it is actually needed, but I kept just in case).

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/RegistryLookupValueResolverWrapper.java
asanguinetti: I removed this class in the latest version.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/RegistryLookupValueResolverWrapper.java
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * Implementation of {@link ValueResolver} which accesses the mule registry and returns the value associated with the result of
+ * another {@link ValueResolver}.
+ * <p/>
+ * Although the registry is mutable, the value associated to a given key is not meant to change. As a consequence,
+ * {@link #isDynamic()} will delegate to the wrapped {@link ValueResolver}.
+ *
+ * @since 4.5.0
+ */
+public class RegistryLookupValueResolverWrapper<T> implements ValueResolver<T>, Initialisable, MuleContextAware {
```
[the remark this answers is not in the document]
asanguinetti: I removed this class in the latest version.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/RegistryLookupValueResolverWrapper.java
asanguinetti: In the latest version, I added caching in the case where the reference is static.

But for the case of a dynamic reference (expression in config-ref) I think the caching shouldn't be done at the resolver level and would be better placed inside the registry itself. My reasoning is the following, please correct me if I'm wrong:

1. If we put the cache at the ValueResolver **class**, we might have problems with providers of the same name across different apps. Would classloader isolation protect us from this? And also, what about lifecycle on redeployments, I'm worried the cache would retain disposed instances that are no longer valid.
2. If we put the cache at the ValueResolver **instance**, we would probably be wasting extra memory for cases in which the caches for different resolvers have the same contents. And we wouldn't benefit from previous hits on the same key by different resolver instances.
3. If the registry access is expensive, we should place the cache in a common place so that everyone accessing the registry in runtime goes through it, right?
4. Are we sure that accessing the registry is more expensive than accessing a cache map (concurrent hash map most likely)?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/ExtensionRegistry.java
```diff
@@ -128,7 +128,7 @@ Collection<ConfigurationProvider> getConfigurationProviders(ExtensionModel exten
    * @return a {@link ConfigurationProvider}
    */
   Optional<ConfigurationProvider> getConfigurationProvider(String key) {
-    return registry.lookupByName(key);
+    return registry.lookupByName(key).filter(cp -> cp instanceof ConfigurationProvider).map(cp -> (ConfigurationProvider) cp);
```
asanguinetti, line 131: I added this so we get an empty optional if the object corresponding to the key is not a ConfigurationProvider.
Otherwise we would be getting class cast exceptions.
Do you think this could cause any issue?
I could also change it to avoid the lambdas if you think performance could be an issue here. But I wasn't sure if it would be premature optimization.

asanguinetti (commented):
Comment for the reviewers.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/ExtensionRegistry.java
marianogonzalez: If the key doesn't correspond to a COnfigurationProvider or doesn't exist then an error should be thrown. Each scenario should have a different error, with descriptive messages

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/RegistryLookupValueResolverWrapper.java
marianogonzalez: Caching at the registry level is not that straight forward:

A registry level cache should:

1) Be agnostic of the type of object being handled (no special treatment for configuration providers)
2) consider prototype objects, which are specific to spring registry but don't exist in the SimpleRegistry (used by MuleFramework)
3) Consider that objects can be registered/unregistered at any time

Although there are some merits to your point, caching at the registry level is very complicated yet never been a problem since lookups usually happen at startup time. All the problems you mention disappear if you use the ExtensionManager instead.

As for lifecycle, that's not an issue as configs lifecycle is tied to that of the mule context.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ConfigurationProviderValueResolver.java
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * Construct a new instance.
+   *
+   * @param configurationProviderReference a not blank {@link String}
+   */
+  public ConfigurationProviderValueResolver(String configurationProviderReference) {
+    if (isExpression(configurationProviderReference)) {
+      configurationProviderNameResolver =
+          new TypeSafeExpressionValueResolver<>(configurationProviderReference, String.class, fromType(String.class));
```
marianogonzalez, line 48: `fromType(String)` -> `org.mule.runtime.api.metadata.DataType#STRING`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ConfigurationProviderValueResolver.java
```diff
@@ -0,0 +1,100 @@
...
+   * @return {@code false}
+   */
+  @Override
+  public boolean isDynamic() {
+    return configurationProviderNameResolver.isDynamic();
+  }
+
+  @Override
+  public void initialise() throws InitialisationException {
+    initialiseIfNeeded(configurationProviderNameResolver, muleContext);
```
marianogonzalez, line 93: inject = true

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
marianogonzalez: what I meant to say is that configurationProvider.get(event) will never return null. If it does, is because a regression was introduced at some point.

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/ExtensionRegistry.java
asanguinetti: Yes, I have the check [here](https://github.com/mulesoft/mule/pull/11413/files#diff-99911b51df158c9f4dfff10192fc7e131e802732880b9a1e45cab45cf9f5db68R79). And the test [here](https://github.com/mulesoft/mule/pull/11413/files#diff-4495b18327d12c8ca6d408332ac61d600826fc0bb550b2c645acdf7d11c3a6c4R67).
But if I didn't change this line in particular, I would have had to catch class cast exceptions instead of just checking for presence.

marianogonzalez (approved):

PR #11429, opened 2022-04-18
Cosmetic changes

eze210:
- Renaming some variables (and a method) in order to avoid shadowing
- Adding a missing parameter in a String.format to avoid undefined behaviour
- Adding a missing final modifier in a constant

Also added a TO-DO, so don't merge this PR until:
- The PR has a corresponding ticket
- The TO-DO has a corresponding ticket or a proper resolution

### DIFF
--- a/modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkOperationModelParserSdk.java
+++ b/modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkOperationModelParserSdk.java
@@ -57,21 +57,21 @@ import java.util.stream.Stream;
  */
 class MuleSdkOperationModelParserSdk extends BaseMuleSdkExtensionModelParser implements OperationModelParser {
 
-  private final ComponentAst operation;
+  private final ComponentAst operationAst;
   private final TypeLoader typeLoader;
   private Boolean isBlocking;
 
   private String name;
 
   public MuleSdkOperationModelParserSdk(ComponentAst operation, TypeLoader typeLoader) {
-    this.operation = operation;
+    this.operationAst = operation;
     this.typeLoader = typeLoader;
 
     parseStructure();
   }
 
   private void parseStructure() {
-    name = getParameter(operation, "name");
+    name = getParameter(operationAst, "name");
   }
 
   private OutputModelParser asOutputModelParser(ComponentAst outputTypeElement) {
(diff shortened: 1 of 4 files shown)


modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkOperationModelParserSdk.java
```diff
@@ -267,7 +267,8 @@ private Optional<ComponentAst> getOutputElement(String elementName) {
   }
 
   private ComponentAst getBody() {
-    return getSingleChild(operation, "body").get();
+    // TODO: What happens if the operation AST doesn't have a body? Is this code reachable in that case?
```
marianogonzalez, line 270: validation will fail in that case. At this point, we assume that the syntax is valid.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkOptionalParameterModelParserSdk.java
```diff
@@ -22,10 +22,10 @@ public class MuleSdkOptionalParameterModelParserSdk extends MuleSdkParameterMode
   public MuleSdkOptionalParameterModelParserSdk(ComponentAst parameter, TypeLoader typeLoader) {
     super(parameter, typeLoader);
 
-    parseStructure();
+    parseOptionalStructure();
```
marianogonzalez, line 25: bad rename. Not necessarily optional.

marianogonzalez (changes requested):

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkOptionalParameterModelParserSdk.java
eze210: What I want to do is to avoid the name collision between this method and the parent's one, in order to avoid shadowing and some future confusion (for the moment it's obvious, but who knows...)

It's parsing the part of the structure that it's specific for optional parameters (the default value), so I would rename it to parseOptionalSpecificStructure(), do you have a better name?

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkOperationModelParserSdk.java
eze210: Changed the TO-DO with a comment explaining why we aren't using isPresent there

eze210:
--validate

PR #11434, opened 2022-04-20
Avoid duplicating error handlers when referencing a global error handler

marianorean:


marianorean:
--validate

core/src/main/java/org/mule/runtime/core/privileged/exception/OnRuntimeProcessingStrategy.java
```diff
@@ -35,11 +42,27 @@ public Sink createSink(FlowConstruct flowConstruct, ReactiveProcessor pipeline)
...
         .flatMap(e -> {
-          String location = e.getContext().getOriginatingLocation().getRootContainerName();
-          return Mono.just(e).transform(getProcessingStrategy(location).get().onProcessor(processor));
+          String location = getFlowName(e);
+          return Mono.just(e).transform(getProcessingStrategy(location).map(ps -> ps.onProcessor(processor))
+              .orElse(getProcessor(processor, location)));
         });
   }
 
+  private String getFlowName(CoreEvent e) {
```
fsgonz, line 51: getFlowName -> getParentFlowNameForErrorHandler

core/src/main/java/org/mule/runtime/core/privileged/exception/OnRuntimeProcessingStrategy.java
```diff
@@ -9,17 +9,24 @@
...
 import org.mule.runtime.core.privileged.processor.MessageProcessors;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 import java.util.Optional;
 
 public class OnRuntimeProcessingStrategy implements ProcessingStrategy {
```
fsgonz, line 26: This should have been present in the previous review:

add a javadoc for this, maybe like this.
/**
 * A {@link ProcessingStrategy} used for {@link org.mule.runtime.core.internal.exception.GlobalErrorHandler}.
 * The resolution of the processing strategy is done in runtime depending on the flow location where the
 * error that is being handled was thrown.
 */

core/src/main/java/org/mule/runtime/core/privileged/exception/OnRuntimeProcessingStrategy.java
```diff
@@ -35,11 +42,27 @@ public Sink createSink(FlowConstruct flowConstruct, ReactiveProcessor pipeline)
...
+  private String getFlowName(CoreEvent e) {
+    FlowStackElement element = e.getFlowCallStack().peek();
+    if (element == null) {
+      return "null";
+    }
+    return element.getFlowName();
+  }
+
+  private ReactiveProcessor getProcessor(ReactiveProcessor processor, String location) {
+    if (logger.isDebugEnabled()) {
```
fsgonz, line 60: I don't understand this. Is this always logged?

fsgonz (changes requested):

fsgonz (approved):
Verify if the injection of the feature flag works ok. I think this should fail in some test anyway.

PR #11438, opened 2022-04-22
W-11032583: Refactor - Make it easy to add new characteristics to be discovered by traversing the inner operations of a reusable operation

asanguinetti:
Refactor to generate a single stream expanding processors from operation calls.

This should scale better to other characteristics that may depend on the order of the processors and should also make it easier to compute multiple characteristics on a single pass.


asanguinetti:
Merged in #11439.

PR #11441, opened 2022-04-27
W-10781615 - SDK forward compatibility for Object Store - Support injection of the SDK API Object Store

dabreu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/store/ObjectStoreSettingsConverter.java
```diff
@@ -0,0 +1,63 @@
...
+
+import org.mule.runtime.api.store.ObjectStoreSettings;
+
+/**
+ * Converts Object Store settings between the SDK api {@link org.mule.sdk.api.store.ObjectStoreSettings} and the Mule opi
+ * {@link ObjectStoreSettings}
+ *
+ * @since 4.5.0
+ */
+public class ObjectStoreSettingsConverter {
```
marianogonzalez, line 20: feels like this class and `ObjectStoreExceptionConverter` should be collapsed into some kind of `SdkObjectStoreUtils` or something like that.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/store/SdkObjectStoreAdapter.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.Map;
+
+/**
+ * Adapts the Mule api for object store {@link ObjectStore} into the SDK api {@link org.mule.sdk.api.store.ObjectStore}
+ *
+ * @since 4.5.0
+ */
+public class SdkObjectStoreAdapter<T extends Serializable> implements org.mule.sdk.api.store.ObjectStore<T> {
+
+  private ObjectStore delegate;
```
marianogonzalez, line 26: final

marianogonzalez (changes requested):

tests/test-extensions/heisenberg-extension/pom.xml
```diff
@@ -26,5 +26,11 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>org.mule.sdk</groupId>
+            <artifactId>mule-sdk-api</artifactId>
+            <version>1.0.0-SNAPSHOT</version>
```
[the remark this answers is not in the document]
dabreu: my bad, I will remove this

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/store/SdkObjectStoreUtils.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  /**
+   * Converts the SDK api OS settings {@link org.mule.sdk.api.store.ObjectStoreSettings} into the corresponding Mule api OS
+   * settings {@code ObjectStoreSettings}
+   *
+   * @param sdkObjectStoreSettings the SDK api OS settings to be converted
+   * @return the converted Mule api OS settings
+   */
+  public static ObjectStoreSettings from(org.mule.sdk.api.store.ObjectStoreSettings sdkObjectStoreSettings) {
```
marianogonzalez, line 43: using `from` is not very intuitive in this case, specially since the method that transforms A into B has the same name than the one that transforms B into A. Can you rename it to something more explicit?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/store/SdkObjectStoreAdapter.java
```diff
@@ -0,0 +1,140 @@
...
+import java.io.Serializable;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Adapts the Mule api for object store {@link ObjectStore} into the SDK api {@link org.mule.sdk.api.store.ObjectStore}
+ *
+ * @since 4.5.0
+ */
+public class SdkObjectStoreAdapter<T extends Serializable> implements org.mule.sdk.api.store.ObjectStore<T> {
```
marianogonzalez, line 24: this class transforms a real objectstore into an SDK one, which means that thrown exceptions are going to be SDK ones, while the rest of the Mule code is going to be expecting Mule exceptions. Where is that being handled?

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/store/SdkObjectStoreUtils.java
dabreu: sure, I will rename it

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/store/SdkObjectStoreAdapter.java
dabreu: Yes, in fact, the exceptions thrown by the Mule Runtime Object Store will be converted into the SDK OS exceptions. 
I see your point and I remember discussing the topic of exceptions within the team. We discussed making the SDK Object Store exceptions to extend the Mule API Object Store exceptions, but then we discarded the idea because that created a dependency between the SDK and Mule APIs, which is something that should be avoided as part of the forward compatibility feature, right? Do you see any other way to handle this?

marianogonzalez (approved):

PR #11444, opened 2022-04-28
W-10848628: File attributes are removed when using Parallel For Each

sofiamorseletto:


modules/core-components/src/main/java/org/mule/runtime/core/internal/routing/ParallelForEach.java
```diff
@@ -127,4 +125,15 @@ protected ForkJoinStrategyFactory getDefaultForkJoinStrategyFactory() {
   public void setCollectionExpression(String collectionExpression) {
     this.collectionExpression = collectionExpression;
   }
+
+  private Message createMessage(TypedValue<?> partTypedValue, CoreEvent event) {
+    if (featureFlaggingService.isEnabled(PARALLEL_FOREACH_FLATTEN_MESSAGE) && partTypedValue.getValue() instanceof Message) {
```
sofiamorseletto, line 130: Added the feature flag here again instead of joining the case with the one on manageTypedValuePayload to consider cases in which the flag is on but the partTypedValue is not a Message

elrodro83 (approved):

PR #11445, opened 2022-04-28
W-10927591: add ArtifactCoordinates to the ExtensionModel

jcgrethe:


core/src/main/java/org/mule/runtime/core/api/artifact/ArtifactCoordinates.java
```diff
@@ -11,20 +11,6 @@
  *
  * @since 4.5.0
```
elrodro83, line 12: add `@deprecated` with explanation in the javadoc as well

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/extension/discovery/RepositoryLookupExtensionModelGenerator.java
```diff
@@ -73,23 +77,24 @@ public ExtensionModel obtainExtensionModel(ExtensionDiscoveryRequest discoveryRe
    * @param artifactName                   the name of the artifact being loaded.
    * @param additionalAttributes           custom parameters for the
    *                                       {@link org.mule.runtime.extension.api.loader.ExtensionModelLoadingRequest}.
+   * @param artifactPluginDescriptor
```
elrodro83, line 80: add description

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/extension/discovery/RepositoryLookupExtensionModelGenerator.java
```diff
@@ -73,23 +77,24 @@ public ExtensionModel obtainExtensionModel(ExtensionDiscoveryRequest discoveryRe
...
+                                                               ArtifactPluginDescriptor artifactPluginDescriptor) {
     ExtensionModelLoader loader = extensionModelLoaderRepository.getExtensionModelLoader(loaderDescriber)
         .orElseThrow(() -> new IllegalArgumentException(format("The identifier '%s' does not match with the describers available "
             + "to generate an ExtensionModel (working with the plugin '%s')", loaderDescriber.getId(), artifactName)));
     Map<String, Object> attributes = new HashMap<>(loaderDescriber.getAttributes());
     attributes.putAll(additionalAttributes);
 
     return loader.loadExtensionModel(builder(artifactClassloader.get(), getDefault(dependencies))
-        .addParameters(attributes)
+        .addParameters(attributes).addEnricher(new ArtifactCoordinatesEnricher(artifactPluginDescriptor.getBundleDescriptor()))
```
elrodro83, line 97: shouldn't the enricher be added in the same place where the rest of the enrichers are declared? @marianogonzalez

elrodro83 (changes requested):

core/src/main/java/org/mule/runtime/core/api/artifact/ArtifactCoordinates.java
eze210: Does it make sense to deprecate something that was added "since 4.5"?

jcgrethe:
--validate

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/extension/discovery/RepositoryLookupExtensionModelGenerator.java
marianogonzalez: it depends. If this is something that must only be added when the EM is created through this code path, then it belongs here.

If it's something we want enforced at a broader scope then it should be added at either the loader or ExtensionModelFactory levels. It depends on the use case.

elrodro83 (approved):

PR #11447, opened 2022-04-28
W-10619588: Allow deprecation of Mule Operation

eze210:


eze210:
--validate

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/BaseMuleSdkExtensionModelParser.java
```diff
@@ -44,7 +46,8 @@ protected <T> T getParameter(ComponentAst ast, String paramName) {
    * @return an {@link Optional} for the parameter's value
    */
   protected <T> Optional<T> getOptionalParameter(ComponentAst ast, String paramName) {
-    return ofNullable(getParameter(ast, paramName));
+    return ofNullable(ast.getParameter(DEFAULT_GROUP_NAME, paramName)).map(ComponentParameterAst::<T>getValue)
```
marianogonzalez, line 49: how does this change the behavior of all the other parsers? IIRC, I originally wrote this method similar to this and then changed it to its current form, don't remember exactly why. Anyway, if this does make sense, then collapse the two map() into one.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkOperationModelParserSdk.java
```diff
@@ -226,8 +227,14 @@ public Optional<ExceptionHandlerModelProperty> getExceptionHandlerModelProperty(
 
   @Override
   public Optional<DeprecationModel> getDeprecationModel() {
-    // TODO: MULE-20083
-    return empty();
+    return getSingleChild(operation, "deprecated").map(this::buildDeprecationModel);
+  }
+
+  private DeprecationModel buildDeprecationModel(ComponentAst deprecationAst) {
+    String message = getParameter(deprecationAst, "message");
```
marianogonzalez, line 234: use constants for the attribute names. buchwald did that on his last PR

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkOperationModelParserSdk.java
```diff
@@ -226,8 +227,14 @@ public Optional<ExceptionHandlerModelProperty> getExceptionHandlerModelProperty(
 
   @Override
   public Optional<DeprecationModel> getDeprecationModel() {
-    // TODO: MULE-20083
-    return empty();
+    return getSingleChild(operation, "deprecated").map(this::buildDeprecationModel);
+  }
+
+  private DeprecationModel buildDeprecationModel(ComponentAst deprecationAst) {
```
marianogonzalez, line 233: move this to an utils class since sources and parameters will need the same thing

marianogonzalez (changes requested):
missing the ability to deprecate a parameter

marianogonzalez (approved):

PR #11449, opened 2022-04-29
W-10848628: File attributes are removed when using Parallel For Each …

sofiamorseletto:
…(#11444)

* W-10848628: File attributes are removed when using Parallel For Each

* Removing unused import

* Reformatting

* Adding feature flag

(cherry picked from commit c9efcab70cec877d5ca00ab408490a8b9eaa1f7a)


sofiamorseletto:
--validate

elrodro83 (approved):

PR #11452, opened 2022-04-29
W-10634121: ValueProvider parameters validations are not working when combining with TypeResolver annotations

SebaElizalde:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/validation/JavaValueProviderModelValidator.java
```diff
@@ -161,9 +168,24 @@ private void validateOptionsResolver(ParameterModel param, boolean mustBeStringT
       } else {
         if (parameterInfo.getExtractionExpression().equals(parameterInfo.getParameterName())) {
           MetadataType metadataType = allParameters.get(parameterInfo.getParameterName());
-          Class<?> expectedType = getType(metadataType)
-              .orElseThrow(() -> new IllegalStateException(format("Unable to get Class for parameter: %s",
-                                                                  parameterInfo.getParameterName())));
+          Set<TypeAnnotation> annotations = metadataType.getAnnotations();
+          Class<?> expectedType;
+          if (annotations.stream().anyMatch(ann -> ann instanceof EnumAnnotation)
```
marianogonzalez, line 173: very uneficient assertion. If you want to test for both both annotations being present, simply fetch them with MetadataType.getAnnotation(Class)

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/validation/JavaValueProviderModelValidator.java
```diff
@@ -161,9 +168,24 @@ private void validateOptionsResolver(ParameterModel param, boolean mustBeStringT
...
+              && annotations.stream().anyMatch(ann -> ann instanceof CustomDefinedStaticTypeAnnotation)) {
+            TypeAnnotation classInformation = annotations.stream().filter(ann -> ann instanceof ClassInformationAnnotation)
+                .findFirst().orElseThrow(() -> new IllegalStateException(format("Unable to get Class for parameter: %s",
+                                                                                parameterInfo.getParameterName())));
+            String className = ((ClassInformationAnnotation) classInformation).getClassname();
```
marianogonzalez, line 178: this seems like a very ad hoc fix which doesn't fix the underlying problem which is assuming that StringTypes are necessarily represented as a java String, when it could be represented in other ways or wrapped in another compoents.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/validation/JavaValueProviderModelValidator.java
```diff
@@ -161,9 +168,24 @@ private void validateOptionsResolver(ParameterModel param, boolean mustBeStringT
       } else {
         if (parameterInfo.getExtractionExpression().equals(parameterInfo.getParameterName())) {
           MetadataType metadataType = allParameters.get(parameterInfo.getParameterName());
-          Class<?> expectedType = getType(metadataType)
-              .orElseThrow(() -> new IllegalStateException(format("Unable to get Class for parameter: %s",
-                                                                  parameterInfo.getParameterName())));
+          Set<TypeAnnotation> annotations = metadataType.getAnnotations();
+          Class<?> expectedType;
+          if (annotations.stream().anyMatch(ann -> ann instanceof EnumAnnotation)
+              && annotations.stream().anyMatch(ann -> ann instanceof CustomDefinedStaticTypeAnnotation)) {
```
marianogonzalez, line 174: why is the CustomDefinedStatypeTypeAnnotation relevant? Can't this same scenario happen in other situations?

marianogonzalez (changes requested):

ndinu (approved):

PR #11462, opened 2022-05-10
Draft for Fix/W-11125519

acurci-at-mulesoft:
WARNING: This draft doesn't have any feature flag to preserve the previous behavior. This solution must be implemented with a feature flag (or system property).


mbuchwald:
There was no further development about this for almost a year, closing it

PR #11470, opened 2022-05-12
W-11070209:  Enable FIPS support for JDK11 and beyond

mule-pachanady:
Adding FIPs support for JDK 11 and hopefully for 17. This is still a work in progress. 
Tested JDK 11 locally but need to test on docker containers in cloud hub.

I would raise a separate PR for wrapper.conf changes

### DIFF
--- a/modules/launcher/pom.xml
+++ b/modules/launcher/pom.xml
@@ -53,6 +53,11 @@
             <version>${project.version}</version>
         </dependency>
         <dependency>
+            <groupId>com.mulesoft.mule.runtime.modules</groupId>
+            <artifactId>mule-module-cluster-ee</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
             <groupId>it.unimi.dsi</groupId>
             <artifactId>fastutil</artifactId>
         </dependency>
--- a/modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
+++ b/modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
@@ -9,8 +9,7 @@ package org.mule.runtime.module.launcher;
 import static org.mule.runtime.api.exception.ExceptionHelper.getRootException;
 import static org.mule.runtime.api.exception.ExceptionHelper.getRootMuleException;
 import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
-import static org.mule.runtime.api.util.MuleSystemProperties.DEPLOYMENT_APPLICATION_PROPERTY;
-import static org.mule.runtime.api.util.MuleSystemProperties.MULE_SIMPLE_LOG;
+import static org.mule.runtime.api.util.MuleSystemProperties.*;
 import static org.mule.runtime.container.api.MuleFoldersUtil.getExecutionFolder;
 import static org.mule.runtime.core.api.config.i18n.CoreMessages.fatalErrorInShutdown;
 import static org.mule.runtime.core.api.config.i18n.CoreMessages.fatalErrorWhileRunning;
@@ -21,7 +20,10 @@ import static org.mule.runtime.core.api.util.StringMessageUtils.getBoilerPlate;
 import static org.mule.runtime.core.internal.logging.LogUtil.log;
 import static org.mule.runtime.module.deployment.internal.MuleDeploymentService.findSchedulerService;
 import static org.mule.runtime.module.deployment.internal.processor.SerializedAstArtifactConfigurationProcessor.serializedAstWithFallbackArtifactConfigurationProcessor;
+import static com.mulesoft.mule.runtime.module.cluster.internal.security.NodeSecurityManager.PROPERTY_SECURITY_MODEL;
+import static com.mulesoft.mule.runtime.module.cluster.internal.security.NodeSecurityManager.FIPS_SECURITY_MODEL;
 
+import java.security.Security;
 import static java.lang.ClassLoader.getSystemClassLoader;
 import static java.lang.Runtime.getRuntime;
 import static java.lang.String.format;
@@ -60,6 +62,9 @@ import org.mule.runtime.module.troubleshooting.api.TroubleshootingService;
 import org.mule.runtime.module.troubleshooting.internal.DefaultTroubleshootingService;
 
 import java.io.File;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.security.Provider;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -70,6 +75,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class MuleContainer {
+
+  public static final String PROPERTY_FIPS_PROVIDER = SYSTEM_PROPERTY_PREFIX + "fips.provider";
+  public static final String PROPERTY_JSSE_PROVIDER = SYSTEM_PROPERTY_PREFIX + "jsse.provider";
+  public static final String SUN_JSSE_PROVIDER = "SunJSSE";
 
   public static final String CLI_OPTIONS[][] =
       {{"builder", "true", "Configuration Builder Type"}, {"config", "true", "Configuration File"},
@@ -217,6 +226,67 @@ public class MuleContainer {
     artifactResourcesRegistry.getMemoryManagementService().initialise();
     artifactResourcesRegistry.inject(artifactResourcesRegistry.getContainerProfilingService());
     initialiseIfNeeded(artifactResourcesRegistry.getContainerProfilingService());
+
+    if (isFipsEnabled()) {
+      System.out.println("Detected FIPs mode");
+      configureSecurityManager();
+      setSecurityAlgorithm();
+    }
+  }
+
+
+  private void configureSecurityManager() {
+    System.out.println("Entered configureSecurityManager()");
+    try {
+      Class<?> classDef =
+          Class.forName(System.getProperty(PROPERTY_FIPS_PROVIDER, "org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider"));
+      Constructor<?> constructor = classDef.getConstructor();
+      Provider fipsProvider = (Provider) constructor.newInstance();
+      java.security.Security.insertProviderAt(fipsProvider, 1);
+
+      classDef =
+          Class.forName(System.getProperty(PROPERTY_JSSE_PROVIDER, "org.bouncycastle.jsse.provider.BouncyCastleJsseProvider"));
+      constructor = classDef.getConstructor();
+      Provider jsseProvider = (Provider) constructor.newInstance();
+      jsseProvider.setProperty("fips", "BCFIPS");
+      java.security.Security.insertProviderAt(jsseProvider, 2);
+
+      Provider sunJsseProvider = Security.getProvider("SunJSSE");
+      if (sunJsseProvider != null) {
+        System.out.println("Remove SunJSSE");
+        Security.removeProvider(SUN_JSSE_PROVIDER);
+      }
+      System.out.println("Exit configureSecurityManager()");
+      printDetails();
+    } catch (ClassNotFoundException | InvocationTargetException | NoSuchMethodException | InstantiationException
+        | IllegalAccessException e) {
+      System.out.println("Exception in configureSecurityManager");
+      e.printStackTrace(System.out);
+    }
+    /*
+     * System.setSecurityManager(new MySecurityManager());
+     * 
+     * try { java.security.Security.addProvider(new Provider("SomeName", 1d, "Some info") {});
+     * System.out.println("SECOND PROVIDER ADDED"); } catch (Exception e) { System.out.println("SECOND PROVIDER CANNOT BE ADDED");
+     * }
+     */
+
+  }
+
+  private void printDetails() {
+    for (Provider provider : Security.getProviders()) {
+      System.out.println(provider.getName());
+      if (provider.getProperty("fips") != null) {
+        System.out.println("fips:" + provider.getProperty("fips"));
+      }
+    }
+  }
+
+  private void setSecurityAlgorithm() {
+    System.out.println("Setting algorrithm and keystore type");
+    Security.setProperty("ssl.KeyManagerFactory.algorithm", "PKIX");
+    Security.setProperty("ssl.TrustManagerFactory.algorithm", "PKIX");
+    Security.setProperty("keystore.type", "PKCS12");
   }
 
   /**
@@ -465,6 +535,11 @@ public class MuleContainer {
    */
   public ArtifactClassLoader getContainerClassLoader() {
     return artifactResourcesRegistry.getContainerClassLoader();
+  }
+
+
+  private boolean isFipsEnabled() {
+    return FIPS_SECURITY_MODEL.equals(getProperty(PROPERTY_SECURITY_MODEL));
   }
 }
 
--- a/modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
+++ b/modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
@@ -9,7 +9,9 @@ package org.mule.runtime.module.launcher;
 import static org.mule.runtime.api.exception.ExceptionHelper.getRootException;
 import static org.mule.runtime.api.exception.ExceptionHelper.getRootMuleException;
 import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
-import static org.mule.runtime.api.util.MuleSystemProperties.*;
+import static org.mule.runtime.api.util.MuleSystemProperties.DEPLOYMENT_APPLICATION_PROPERTY;
+import static org.mule.runtime.api.util.MuleSystemProperties.MULE_SIMPLE_LOG;
+import static org.mule.runtime.api.util.MuleSystemProperties.SYSTEM_PROPERTY_PREFIX;
 import static org.mule.runtime.container.api.MuleFoldersUtil.getExecutionFolder;
 import static org.mule.runtime.core.api.config.i18n.CoreMessages.fatalErrorInShutdown;
 import static org.mule.runtime.core.api.config.i18n.CoreMessages.fatalErrorWhileRunning;
@@ -74,12 +76,13 @@ import org.slf4j.LoggerFactory;
 
 public class MuleContainer {
 
+  private static Logger LOGGER = LoggerFactory.getLogger(MuleContainer.class.getName());
   public static final String PROPERTY_FIPS_PROVIDER = SYSTEM_PROPERTY_PREFIX + "fips.provider";
   public static final String PROPERTY_JSSE_PROVIDER = SYSTEM_PROPERTY_PREFIX + "jsse.provider";
   public static final String SUN_JSSE_PROVIDER = "SunJSSE";
-  public static final String KEY_FACTORY_ALGORITHM = "ssl.KeyManagerFactory.algorithm";
-  public static final String TRUST_MANAGER_FACTORY_ALGORITHM = "ssl.TrustManagerFactory.algorithm";
-  public static final String KEYSTORE_TYPE = "keystore.type";
+  public static final String PROPERTY_KEY_FACTORY_ALGORITHM = "ssl.KeyManagerFactory.algorithm";
+  public static final String PROPERTY_TRUST_MANAGER_FACTORY_ALGORITHM = "ssl.TrustManagerFactory.algorithm";
+  public static final String PROPERTY_KEYSTORE_TYPE = "keystore.type";
 
   public static final String CLI_OPTIONS[][] =
       {{"builder", "true", "Configuration Builder Type"}, {"config", "true", "Configuration File"},
@@ -229,16 +232,12 @@ public class MuleContainer {
     initialiseIfNeeded(artifactResourcesRegistry.getContainerProfilingService());
 
     if (isFipsEnabled()) {
-      System.out.println("Detected FIPs mode");
       configureSecurityManager();
       setSecurityAlgorithm();
     }
   }
 
-  private void configureSecurityManager() {
-    System.out.println("Entered configureSecurityManager()");
-    printDetails();
-    System.out.println("Finished Initial Print details");
+  private void configureSecurityManager() throws InitialisationException {
     try {
       Class<?> classDef =
           Class.forName(System.getProperty(PROPERTY_FIPS_PROVIDER, "org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider"));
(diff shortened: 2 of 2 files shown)


modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -217,6 +230,52 @@ protected void init(String[] args) throws IllegalArgumentException, Initialisati
...
+      java.security.Security.insertProviderAt(jsseProvider, 2);
+
+      Provider sunJsseProvider = Security.getProvider("SunJSSE");
+      // for java 8
+      Provider sunJsseProviderLegacy = Security.getProvider("com.sun.net.ssl.internal.ssl.Provider");
+
+      if (sunJsseProvider != null) {
+        Security.removeProvider(SUN_JSSE_PROVIDER);
+      }
+
```
mule-pachanady, line 262: Trying to mimic manual setting `security.provider.6=com.sun.net.ssl.internal.ssl.Provider BCFIPS`

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -217,6 +230,52 @@ protected void init(String[] args) throws IllegalArgumentException, Initialisati
...
+      if (sunJsseProviderLegacy != null) {
+        sunJsseProviderLegacy.setProperty("BCFIPS", "");
```
mule-pachanady, line 264: Trying to mimic manual setting security.provider.6=com.sun.net.ssl.internal.ssl.Provider BCFIPS

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -217,6 +230,52 @@ protected void init(String[] args) throws IllegalArgumentException, Initialisati
...
+
+      Provider sunJsseProvider = Security.getProvider("SunJSSE");
+      // for java 8
+      Provider sunJsseProviderLegacy = Security.getProvider("com.sun.net.ssl.internal.ssl.Provider");
+
+      if (sunJsseProvider != null) {
+        Security.removeProvider(SUN_JSSE_PROVIDER);
+      }
+
+      if (sunJsseProviderLegacy != null) {
```
mule-pachanady, line 263: Trying to mimic manual setting security.provider.6=com.sun.net.ssl.internal.ssl.Provider BCFIPS

modules/launcher/pom.xml
```diff
@@ -52,6 +52,11 @@
             <artifactId>mule-module-reboot</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.mulesoft.mule.runtime.modules</groupId>
+            <artifactId>mule-module-cluster-ee</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
asanguinetti, line 59: Careful with this, because this repository (mule) is part of the Mule CE (Community Edition), and here you would be adding a dependency on something that belongs to the Mule EE (Enterprise Edition) product.

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -217,6 +230,52 @@ protected void init(String[] args) throws IllegalArgumentException, Initialisati
...
+      java.security.Security.insertProviderAt(fipsProvider, 1);
+
+      classDef =
+          Class.forName(System.getProperty(PROPERTY_JSSE_PROVIDER, "org.bouncycastle.jsse.provider.BouncyCastleJsseProvider"));
+      constructor = classDef.getConstructor();
+      Provider jsseProvider = (Provider) constructor.newInstance();
+      jsseProvider.setProperty("fips", "BCFIPS");
+      java.security.Security.insertProviderAt(jsseProvider, 2);
+
+      Provider sunJsseProvider = Security.getProvider("SunJSSE");
```
asanguinetti, line 255: This one should be:
```suggestion
      Provider sunJsseProvider = Security.getProvider(SUN_JSSE_PROVIDER);
```
right?

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -21,7 +22,10 @@
 import static org.mule.runtime.core.internal.logging.LogUtil.log;
 import static org.mule.runtime.module.deployment.internal.MuleDeploymentService.findSchedulerService;
 import static org.mule.runtime.module.deployment.internal.processor.SerializedAstArtifactConfigurationProcessor.serializedAstWithFallbackArtifactConfigurationProcessor;
+import static com.mulesoft.mule.runtime.module.cluster.internal.security.NodeSecurityManager.PROPERTY_SECURITY_MODEL;
+import static com.mulesoft.mule.runtime.module.cluster.internal.security.NodeSecurityManager.FIPS_SECURITY_MODEL;
```
asanguinetti, line 26: We need to decide if FIPS support needs to be added to Mule CE or just Mule EE.

If it is Mule CE, we would have to define this constants somewhere in this project, because we can't create a dependency on Mule EE.

If it is only Mule EE then, check if all of this can be done in the [MuleContainerBootstrap.java](https://github.com/mulesoft/mule-ee/blob/master/modules/boot/src/main/java/org/mule/runtime/module/reboot/MuleContainerBootstrap.java) from `mule-ee`.

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -60,9 +64,10 @@
 import org.mule.runtime.module.troubleshooting.internal.DefaultTroubleshootingService;
 
 import java.io.File;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.security.Provider;
+import java.util.*;
```
asanguinetti, line 70: Code convention: avoid the wildcard imports.

asanguinetti (changes requested):
I marked a few issues that I'm seeing, but I'm lacking the context to verify if what we are doing is correct/enough to enable FIPS in the right way. Would you like to share the resources you are using as a reference?

modules/launcher/pom.xml
mule-pachanady: Ok I dont really need this, need some constants, I will redefine those

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -217,6 +235,52 @@ protected void init(String[] args) throws IllegalArgumentException, Initialisati
...
+    if (isFipsEnabled()) {
+      configureSecurityManager();
+      setSecurityAlgorithm();
+    }
+  }
+
+  private void configureSecurityManager() throws InitialisationException {
+    try {
+      Class<?> classDef =
+          Class.forName(System.getProperty(PROPERTY_FIPS_PROVIDER, "org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider"));
```
asanguinetti, line 248: Are we ok with the fact that if the get a `ClassNotFoundException` here we would be skipping everything that follows bellow? there is no fallback for this right?

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -217,6 +235,52 @@ protected void init(String[] args) throws IllegalArgumentException, Initialisati
...
+      Constructor<?> constructor = classDef.getConstructor();
+      Provider fipsProvider = (Provider) constructor.newInstance();
+      java.security.Security.insertProviderAt(fipsProvider, 1);
+
+      classDef =
+          Class.forName(System.getProperty(PROPERTY_JSSE_PROVIDER, "org.bouncycastle.jsse.provider.BouncyCastleJsseProvider"));
```
asanguinetti, line 254: Same here

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -217,6 +235,52 @@ protected void init(String[] args) throws IllegalArgumentException, Initialisati
...
+    }
+  }
+
+  private void configureSecurityManager() throws InitialisationException {
+    try {
+      Class<?> classDef =
+          Class.forName(System.getProperty(PROPERTY_FIPS_PROVIDER, "org.bouncycastle.jcajce.provider.BouncyCastleFipsProvider"));
+      Constructor<?> constructor = classDef.getConstructor();
+      Provider fipsProvider = (Provider) constructor.newInstance();
+      java.security.Security.insertProviderAt(fipsProvider, 1);
```
asanguinetti, line 251: do we really need to use the fully qualified name for the `Security` class here? what is it colliding with? Actually we should be using a static import here to follow code conventions.

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -217,6 +235,52 @@ protected void init(String[] args) throws IllegalArgumentException, Initialisati
...
+
+      classDef =
+          Class.forName(System.getProperty(PROPERTY_JSSE_PROVIDER, "org.bouncycastle.jsse.provider.BouncyCastleJsseProvider"));
+      constructor = classDef.getConstructor();
+      Provider jsseProvider = (Provider) constructor.newInstance();
+      jsseProvider.setProperty(FIPS_KEY, FIPS_VALUE);
+      java.security.Security.insertProviderAt(jsseProvider, 2);
```
asanguinetti, line 258: Same here

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -217,6 +235,52 @@ protected void init(String[] args) throws IllegalArgumentException, Initialisati
...
+      if (sunJsseProvider != null) {
+        Security.removeProvider(SUN_JSSE_PROVIDER);
+      }
+
+      if (sunJsseProviderLegacy != null) {
+        sunJsseProviderLegacy.setProperty(FIPS_VALUE, "");
+      }
+    } catch (ClassNotFoundException | InvocationTargetException | NoSuchMethodException | InstantiationException
+        | IllegalAccessException e) {
+      LOGGER.error("Critical error while enabling FIPS:" + e.getMessage());
```
asanguinetti, line 273: I think it is better to include the cause for the logged error message instead of concatenating just the `e.getMessage()`.
```suggestion
      LOGGER.error("Critical error while enabling FIPS", e);
```

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -217,6 +235,52 @@ protected void init(String[] args) throws IllegalArgumentException, Initialisati
...
+      throw new InitialisationException(createStaticMessage("Critical error while enabling FIPS:"), null);
```
asanguinetti, line 274: Similarly here, I think we should include the original cause for the new exception that is being thrown. It should help when troubleshooting:
```suggestion
      throw new InitialisationException(createStaticMessage("Critical error while enabling FIPS"), e, null);
```
Also please check if it is ok for the `Initialisable` component to be `null`.

asanguinetti (changes requested):

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
mule-pachanady: We should fail thee container start as I mentioned above

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
mule-pachanady: Its safer to specify full name here, It will be instantiated later by JVM as needed.  From docs
```
public static [Class](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html)<?> forName([String](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html) className)
                        throws [ClassNotFoundException] 

Parameters:
className - the fully qualified name of the desired class.
```

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
mule-pachanady: Please see above

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
mule-pachanady: I guess I don't have handle to Initialisable here, all I want is to fail startup of container.

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
asanguinetti: sorry, the lines marked for my comment are confusing, but I was referring to:

```suggestion
      Security.insertProviderAt(fipsProvider, 1);
```
or even:
```suggestion
      insertProviderAt(fipsProvider, 1);
```

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
mule-pachanady: Sure Will do

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
mule-pachanady: I think Security.insertProviderAt is bit better as that way we know its in a different class

asanguinetti: Personally, I also prefer `Security.insertProviderAt`, but our [code conventions](https://docs.google.com/document/d/1N2Axq3ABJjYmj9BjfYQkvJfF0odHQnlb8MAdVQydOWI/edit#heading=h.ujxk1qbe2hxj) say this:

> Static imports
> Always use static imports unless it’s not clear what the method is doing without using the class name when referencing the method/variable.

This one is kind of greyish, because `insertProviderAt` does not indicate what is the provider for.

Each reviewer may have their own opinion about when is it clear and when is it not.

For this particular case, I will approve either option.

mule-pachanady: Im not in favor of that as you can imagine you could have static imports from say 5 classes and can get very confusing. But I guess following agreed upon conventions is important.

mule-pachanady:
@asanguinetti Please review, removed logic to read properties from wrapper.conf

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -71,6 +76,22 @@
 
 public class MuleContainer {
 
+  private static Logger LOGGER = LoggerFactory.getLogger(MuleContainer.class.getName());
+
+  private static final String PROPERTY_SECURITY_MODEL = SYSTEM_PROPERTY_PREFIX + "security.model";
```
asanguinetti, line 81: Remove this. Use `MULE_SECURITY_SYSTEM_PROPERTY` already defined in `MuleSystemProperties`.

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -466,5 +532,9 @@ public DeploymentService getDeploymentService() {
   public ArtifactClassLoader getContainerClassLoader() {
     return artifactResourcesRegistry.getContainerClassLoader();
   }
+
+  private boolean isFipsEnabled() {
+    return FIPS_SECURITY_MODEL.equals(getProperty(PROPERTY_SECURITY_MODEL));
```
asanguinetti, line 537:
```suggestion
    return FIPS_SECURITY_MODEL.equals(getProperty(MULE_SECURITY_SYSTEM_PROPERTY));
```

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -11,6 +11,7 @@
 import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
 import static org.mule.runtime.api.util.MuleSystemProperties.DEPLOYMENT_APPLICATION_PROPERTY;
 import static org.mule.runtime.api.util.MuleSystemProperties.MULE_SIMPLE_LOG;
+import static org.mule.runtime.api.util.MuleSystemProperties.SYSTEM_PROPERTY_PREFIX;
```
asanguinetti, line 14:
```suggestion
import static org.mule.runtime.api.util.MuleSystemProperties.MULE_SECURITY_SYSTEM_PROPERTY;
```

asanguinetti (changes requested):

asanguinetti (commented):
LGTM so far. But I don't feel comfortable approving without at least @elrodro83 and/or @fsgonz taking a look at it.

There are a few questions that still need to be answered like:
1- How would the customer be able to choose a different security provider other than BouncyCastle?
2- Do we want to include this in Mule CE?

mule-pachanady:
1) There was a separate PR to allow customers to edit FIPS Provider but Axel after talking with you we decided to not merge it to avoid customers making accidental errors. But we can bring it in if desired

https://github.com/mulesoft/mule-ee-distributions/pull/2161

2) Asking @facundopoblete about support in Mule CE

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -22,6 +23,7 @@
 import static org.mule.runtime.module.deployment.internal.MuleDeploymentService.findSchedulerService;
 import static org.mule.runtime.module.deployment.internal.processor.SerializedAstArtifactConfigurationProcessor.serializedAstWithFallbackArtifactConfigurationProcessor;
 
+import java.security.Security;
```
fsgonz, line 26: order of the imports. It should be after the static imports after org mule.

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -217,6 +237,51 @@ protected void init(String[] args) throws IllegalArgumentException, Initialisati
...
     artifactResourcesRegistry.inject(artifactResourcesRegistry.getContainerProfilingService());
     initialiseIfNeeded(artifactResourcesRegistry.getContainerProfilingService());
+
+    if (isFipsEnabled()) {
+      configureSecurityManager();
+      setSecurityAlgorithm();
+    }
+  }
+
+  private void configureSecurityManager() throws InitialisationException {
```
fsgonz, line 247: cant this method be something like configureSecurityManagerWithFipsPriority or something more specific if it is only used for fips?

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -71,6 +76,21 @@
 
 public class MuleContainer {
```
fsgonz, line 77: How can this be tested? Can we create some system tests for this? 
Even if they are not run everytime? I don't know if this can be tested with munit? Probably a cluster hast to be created.

asanguinetti:
> 1. There was a separate PR to allow customers to edit FIPS Provider but Axel after talking with you we decided to not merge it to avoid customers making accidental errors. But we can bring it in if desired

Yes, but at that time I was missing some information which is that users were able to configure that before as stated here:
https://docs.mulesoft.com/mule-runtime/4.3/fips-140-2-compliance-support

I didn't want to introduce new configurations if we can avoid it, but if we were allowing customers to configure the provider before, then I suppose we need to keep that or we would be breaking FIPS support for some users (even if they stay in JDK8).

Maybe we need to at least allow them to configure the provider. For the other configurations, since I can't find any way for them to configure them before, I would say we can hard-code them initially.

That being said, we should verify if a reasonable error will be thrown if the user enables FIPS but selects a security provider that is not FIPS compliant.

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -466,5 +531,9 @@ public DeploymentService getDeploymentService() {
   public ArtifactClassLoader getContainerClassLoader() {
     return artifactResourcesRegistry.getContainerClassLoader();
   }
+
+  private boolean isFipsEnabled() {
+    return FIPS_SECURITY_MODEL.equals(getProperty(MULE_SECURITY_SYSTEM_PROPERTY));
```
fsgonz, line 536: Are we planning to have other security models apart from this? If we are not probably this should be only a boolean. If we are planning to do that, we should add something to retrieve what we have to do according to the string (though in this case it seems a bit of overengineering to do that). 
Are we sure we want the providers to be hardcoded? Probably yes, as we will only support BouncyCastle.

fsgonz (changes requested):

fsgonz:
Are we planning to add some verification that no other component can change this as in Pablo's document?

fsgonz:
> LGTM so far. But I don't feel comfortable approving without at least @elrodro83 and/or @fsgonz taking a look at it.
> 
> There are a few questions that still need to be answered like: 1- How would the customer be able to choose a different security provider other than BouncyCastle? 2- Do we want to include this in Mule CE?

Regarding 1) that appeared in my PR. If we are suppose to support it. Regarding 2) according to @elrodro83 we shouldn't.

fsgonz:
If this is a manual process, how will this be solved in RTF? Will unpackage the distribution and add the classes manually?

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
```diff
@@ -217,6 +237,51 @@ protected void init(String[] args) throws IllegalArgumentException, Initialisati
     artifactResourcesRegistry.getMemoryManagementService().initialise();
     artifactResourcesRegistry.inject(artifactResourcesRegistry.getContainerProfilingService());
     initialiseIfNeeded(artifactResourcesRegistry.getContainerProfilingService());
+
+    if (isFipsEnabled()) {
+      configureSecurityManager();
+      setSecurityAlgorithm();
+    }
```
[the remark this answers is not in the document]
mule-pachanady: Sure Doing it

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
mule-pachanady: For now we are going to support only BouncyCastle, but we had other  thoughts like exposing as properties in wrapper.conf

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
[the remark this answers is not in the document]
mule-pachanady: sure

modules/launcher/src/main/java/org/mule/runtime/module/launcher/MuleContainer.java
mule-pachanady: Removed

mule-pachanady:
We wont support FIPS on Mule CE

mule-pachanady:
> If this is a manual process, how will this be solved in RTF? Will unpackage the distribution and add the classes manually?

I don't know much about RFT. May be some extra work is needed

PR #11476, opened 2022-05-16
W-10923083: PropertyNotFoundException error when defining configuration elements name from application properties

alopezbizon:


alopezbizon:
I'll make the change to the master and cherry pick this change here later.

PR #11486, opened 2022-05-20
W-11086587: Add logging to offer method in AbstractQueueStoreDelegate…

sofiamorseletto:
… so that we know why a record was not successfully queued

A test for this use case already exists and it's located in https://github.com/mulesoft/mule/blob/f3c3da51aeebe9232d28a6a6e65f1411fafabcb9/tests/unit/src/main/java/org/mule/tck/core/util/queue/QueueStoreTestCase.java#L63


sofiamorseletto:
--validate

fsgonz (dismissed):

fsgonz (approved):

PR #11491, opened 2022-05-24
W-10619591: Implement isTransactional() method for MuleOperations

mbuchwald:
This PR also includes the refactor for Characteristics to be in a separate file


runtime-extension-model/src/main/resources/META-INF/mule-operation.xsd
```diff
@@ -296,7 +296,6 @@
             <xsd:enumeration value="datetime"/>
             <xsd:enumeration value="time"/>
             <xsd:enumeration value="boolean"/>
-            <xsd:enumeration value="binary"/>
```
mbuchwald, line 299: In this tiny view it's not visible but actually `binary` is repeated here

mbuchwald:
--validate

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -1258,23 +1259,26 @@ private boolean supportsTransactions(T componentModel) {
     return componentModel instanceof ConnectableComponentModel && ((ConnectableComponentModel) componentModel).isTransactional();
   }
 
+  private boolean requiresTransactionalActionConfiguration(T componentModel) {
+    return !componentModel.getModelProperty(NoTransactionalActionModelProperty.class).isPresent();
+  }
```
asanguinetti, line 1264: I'm noticing we are checking the model property to determine if a CMP requires transactional configuration, which for me makes perfect sense.
For a moment I considered the possibility of checking for the existence of the parameter, but I guess we wouldn't want any config to be applied if, for some reason, a "transactionalAction" parameter is explicitly is added to a Reusable Operation.
I think what we are doing is better, but I also think we should add a test that describes what should happen in such case, right? Should we reject such operation definition? or should we just ignore it for transactional purposes? I'm more inclined towards the former, but I think right now we are doing the later.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/utils/Characteristic.java
```diff
@@ -0,0 +1,159 @@
...
+
+    private BooleanCharacteristic(Predicate<OperationModel> predicate, Boolean defaultValue, Boolean stopValue) {
+      super(((operationModel,
+              curValue) -> (curValue != null && curValue == stopValue) ? curValue : predicate.test(operationModel)),
+            defaultValue, stopValue);
+    }
+
+  }
+
+  public static class AnyMatchCharacteristic extends BooleanCharacteristic {
```
asanguinetti, line 94: Javadoc

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/utils/Characteristic.java
```diff
@@ -0,0 +1,159 @@
...
+
+    public AnyMatchCharacteristic(Predicate<OperationModel> predicate) {
+      super(predicate, false, true);
+    }
+
+  }
+
+  public static class AggregatedNotificationsCharacteristic extends Characteristic<List<NotificationModel>> {
```
asanguinetti, line 102: Javadoc

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/utils/Characteristic.java
```diff
@@ -0,0 +1,159 @@
...
+    public boolean filterComponent(ComponentAst componentAst) {
+      return filterExpression.test(componentAst);
+    }
+
+    public boolean ignoreComponent(ComponentAst componentAst) {
+      return ignoreExpression.test(componentAst);
+    }
+  }
+
+  public static class BooleanFilteringCharacteristic extends FilteringCharacteristic<Boolean> {
```
asanguinetti, line 142: Javadoc

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/utils/Characteristic.java
```diff
@@ -0,0 +1,159 @@
...
+
+    private BooleanFilteringCharacteristic(Predicate<OperationModel> predicate, Boolean defaultValue, Boolean stopValue,
+                                           Predicate<ComponentAst> filter, Predicate<ComponentAst> ignore) {
+      super(((operationModel,
+              curValue) -> (curValue != null && curValue == stopValue) ? curValue : predicate.test(operationModel)),
+            defaultValue, stopValue, filter, ignore);
+    }
+  }
+
+  public static class AnyMatchFilteringCharacteristic extends BooleanFilteringCharacteristic {
```
asanguinetti, line 152: Javadoc

asanguinetti (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
mbuchwald: You mean to check if it's okey to have a parameter that is transactionalAction? I guess it would be confusing as a parameter name, but if that's what Daiana defines, it's on her. I can add this test to make sure it works!

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkOperationModelParserSdk.java
```diff
@@ -115,7 +122,7 @@ public String getDescription() {
 
   @Override
   public List<ModelProperty> getAdditionalModelProperties() {
-    return emptyList();
+    return singletonList(new NoTransactionalActionModelProperty());
```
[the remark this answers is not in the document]
mbuchwald: As far as I checked, it's only called once per operation. However, I don't see any harm in caching this/ having it as an attribute... or maybe for having it on memory? I will ask Rodro/MG about it to be sure

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkOperationModelParserSdk.java
```diff
@@ -292,134 +298,99 @@ private ComponentAst getBody() {
...
-      return operationParser.getOperationModelsRecursiveStream(operationModelParsersByName, visitedOperations);
+      return operationParser.getOperationModelsRecursiveStream(operationModelParsersByName, visitedOperations, filterCondition,
+                                                               ignoreCondition);
     } else {
       return Stream.empty();
     }
   }
 
   private Stream<OperationModel> getOperationModelsRecursiveStream(Map<String, MuleSdkOperationModelParserSdk> operationModelParsersByName) {
+    return getOperationModelsRecursiveStream(operationModelParsersByName, componentAst -> false, componentAst -> false);
```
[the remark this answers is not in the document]
mbuchwald: When the filterExpression returns true, it means this component and everything inside it has to be filtered. That's why the default one returns false always, because it will never filter

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
asanguinetti: Yes, I think it will work right now. But the importance of this test is to document what would we like to happen in such case (and that we actually considered it). Because there are other alternatives, for example, forbidding it.

To be honest I don't know which approach is best. If we go for consistency with other SDKs I suppose we should allow it. But if we want to apply the "put the gorra now, you can always take it off later" philosophy, maybe we shouldn't.

We can discuss it in the spec if you want. And we can create another task if you want to defer it.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkOperationModelParserSdk.java
```diff
@@ -292,134 +298,99 @@ private ComponentAst getBody() {
...
+                                                                   Predicate<ComponentAst> filterCondition,
+                                                                   Predicate<ComponentAst> ignoreCondition) {
     if (!visitedOperations.add(this.getName())) {
       return Stream.empty();
     }
 
+    Set<ComponentAst> filtered = new HashSet<>();
+
     return getBody().recursiveStream()
         .flatMap(componentAst -> {
```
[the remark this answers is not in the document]
mbuchwald: This is internal so I don't see a unit test for this per se. Each characteristic should have enough coverage to show that this works properly (which in the end is the goal for this, so the test should remain as black-box test). 
I do agree on the following comment that there should be some unit tests for the characteristics

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/utils/Characteristic.java
```diff
@@ -0,0 +1,159 @@
+/*
```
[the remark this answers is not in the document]
mbuchwald: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000y9Tu8YAE/view
Added this for next sprint

modules/extensions-mule-support/src/test/java/org/mule/runtime/module/extension/mule/internal/operation/MuleOperationIsTransactionalTestCase.java
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  @Test
+  @Description("Checks that operation joining within async is not transactional")
+  public void asyncWithJoinIfPossible() {
+    assertForOperation("asyncWithJoinIfPossible", false);
+  }
+
+  @Test
+  @Description("Checks that operation not joining within async within is not transactional")
```
asanguinetti, line 131:
```suggestion
  @Description("Checks that operation not joining within async within try is not transactional")
```

modules/extensions-mule-support/src/test/java/org/mule/runtime/module/extension/mule/internal/operation/MuleOperationIsTransactionalTestCase.java
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  @Test
+  @Description("Checks that operation joining within async within is not transactional")
+  public void tryAndAsyncJoin() {
+    assertForOperation("tryAndAsyncJoin", false);
+  }
+
+  @Test
+  @Description("Checks that operation not joining within async within, followed by not supported is not transactional")
```
asanguinetti, line 143:
```suggestion
  @Description("Checks that operation not joining within async within try, followed by not supported is not transactional")
```

modules/extensions-mule-support/src/test/java/org/mule/runtime/module/extension/mule/internal/operation/MuleOperationIsTransactionalTestCase.java
```diff
@@ -0,0 +1,207 @@
...
+  public void tryAndAsyncJoinFollowedByNotSupported() {
+    assertForOperation("tryAndAsyncJoinFollowedByNotSupported", false);
+  }
+
+  @Test
+  @Description("Checks that operation not joining within async within, followed by join is transactional")
```
asanguinetti, line 149:
```suggestion
  @Description("Checks that operation not joining within async within try, followed by join is transactional")
```

modules/extensions-mule-support/src/test/java/org/mule/runtime/module/extension/mule/internal/operation/MuleOperationIsTransactionalTestCase.java
```diff
@@ -0,0 +1,207 @@
...
+import io.qameta.allure.Description;
+import io.qameta.allure.Feature;
+import io.qameta.allure.Story;
+import org.mule.functional.junit4.MuleArtifactFunctionalTestCase;
+import org.mule.runtime.api.meta.model.operation.OperationModel;
+import org.mule.runtime.core.api.extension.ExtensionManager;
+
+import org.junit.Test;
+import javax.inject.Inject;
+import java.util.List;
```
asanguinetti, line 24: Order of imports? ([code conventions](https://docs.google.com/document/d/1N2Axq3ABJjYmj9BjfYQkvJfF0odHQnlb8MAdVQydOWI/edit#heading=h.ujxk1qbe2hxj))

asanguinetti (changes requested):
Minor stuff

asanguinetti (dismissed):

eze210 (approved):

PR #11508, opened 2022-06-02
Fixing isDynamic method for some processors that had it wrong.

asanguinetti:


asanguinetti:
Abandoned, too risky without a business need.

PR #11509, opened 2022-06-03
W-11177824: flow ref validations fails for xml sdk components that declare inner flows when using lazy init

SebaElizalde:


modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpansionModulesModel.java
```diff
@@ -86,6 +93,33 @@ public ArtifactAst expand() {
...
 
     if (hasMacroExpansionExtension) {
+      // The macro expansion of xml sdk components causes inner flow components to be generated as a child of another component;
+      // so when generating the minimal app during lazy init, those flow components are not registered as top level components
+      // causing certain validations to fail. This code looks for those flow components and adds them as top level.
+      Stream<ComponentAst> flowComponents =
+          applicationModel.recursiveStream()
+              .filter(comp -> comp.getComponentType() != null && comp.getComponentType().equals(FLOW));
+      List<ComponentAst> componentsToAdd =
+          flowComponents.filter(comp -> !applicationModel.topLevelComponents().contains(comp)).collect(Collectors.toList());
```
elrodro83, line 103: inline `flowComponents`

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpansionModulesModel.java
```diff
@@ -86,6 +93,33 @@ public ArtifactAst expand() {
     }
 
     if (hasMacroExpansionExtension) {
+      // The macro expansion of xml sdk components causes inner flow components to be generated as a child of another component;
+      // so when generating the minimal app during lazy init, those flow components are not registered as top level components
+      // causing certain validations to fail. This code looks for those flow components and adds them as top level.
+      Stream<ComponentAst> flowComponents =
+          applicationModel.recursiveStream()
+              .filter(comp -> comp.getComponentType() != null && comp.getComponentType().equals(FLOW));
```
elrodro83, line 101: be careful to only move flows from the extension that is being macroexpanded

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpansionModulesModel.java
elrodro83, line 103: `!applicationModel.topLevelComponents().contains(comp)` shouldn't be needed if you are only navigating the extension being macroexpanded

elrodro83, line 103: static import: `Collectors.toList()`

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpansionModulesModel.java
```diff
@@ -86,6 +93,33 @@ public ArtifactAst expand() {
...
+      List<ComponentAst> componentsToAdd =
+          flowComponents.filter(comp -> !applicationModel.topLevelComponents().contains(comp)).collect(Collectors.toList());
+      final UnaryOperator<ComponentAst> function = comp -> {
+        List<ComponentAst> childrenToKeep = comp.directChildrenStream()
+            .filter(child -> !(child.getComponentType() != null && child.getComponentType().equals(FLOW)))
+            .collect(Collectors.toList());
```
elrodro83, line 107: static import

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpansionModulesModel.java
```diff
@@ -86,6 +93,33 @@ public ArtifactAst expand() {
...
+      // so when generating the minimal app during lazy init, those flow components are not registered as top level components
+      // causing certain validations to fail. This code looks for those flow components and adds them as top level.
+      Stream<ComponentAst> flowComponents =
+          applicationModel.recursiveStream()
+              .filter(comp -> comp.getComponentType() != null && comp.getComponentType().equals(FLOW));
+      List<ComponentAst> componentsToAdd =
+          flowComponents.filter(comp -> !applicationModel.topLevelComponents().contains(comp)).collect(Collectors.toList());
+      final UnaryOperator<ComponentAst> function = comp -> {
+        List<ComponentAst> childrenToKeep = comp.directChildrenStream()
+            .filter(child -> !(child.getComponentType() != null && child.getComponentType().equals(FLOW)))
```
elrodro83, line 106: `!(child.getComponentType() != null && child.getComponentType().equals(FLOW))` -> `!FLOW.equals(child.getComponentType())`

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpansionModulesModel.java
```diff
@@ -86,6 +93,33 @@ public ArtifactAst expand() {
...
+            return childrenToKeep;
+          }
+
+          @Override
+          public Stream<ComponentAst> directChildrenStream() {
+            return childrenToKeep.stream();
+          }
+        };
+      };
+      applicationModel = copyRecursively(applicationModel, function, () -> componentsToAdd, c -> false);
```
elrodro83, line 121: inline `function`

elrodro83 (changes requested):

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpansionModulesModel.java
SebaElizalde: all extensions have been macroexpanded and added to the application model by this point

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpansionModulesModel.java
SebaElizalde: Even if dealing with only 1 extension, there is the case that the extension has 2 or more operations pointing to the same flow, so while travelling the ast you can find the flow twice, but you only want to add it once to top level. So this check should remain imo.

elrodro83 (approved):

SebaElizalde:
--validate

PR #11519, opened 2022-06-08
W-11102733: Allow deprecation of Mule Operation Parameters

eze210:


modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/BaseMuleSdkExtensionModelParser.java
```diff
@@ -24,6 +24,7 @@
  */
 abstract class BaseMuleSdkExtensionModelParser {
 
+  static final String DEPRECATED_CONSTRUCT_NAME = "deprecated";
```
asanguinetti, line 27: Shouldn't this one be protected?

runtime-extension-model/src/main/resources/META-INF/mule-operation.xsd
```diff
@@ -39,6 +39,13 @@
                             </xsd:documentation>
                         </xsd:annotation>
                     </xsd:element>
+                    <xsd:element name="deprecated" type="deprecatedType" minOccurs="0">
```
asanguinetti, line 42: Shouldn't this have `maxOccurs="1"`? And I think the same would go for the deprecation construct inside operations, right?

Which also makes me wonder... how is all of this working without having to modify `MuleOperationExtensionModelDeclarer`? I mean, the operations have the [deprecation construct as a NestedComponent](https://github.com/mulesoft/mule/blob/c588c59ace9408923aa66193d6b1af5a130f1a2f/runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/MuleOperationExtensionModelDeclarer.java#L163), but the parameters don't.

asanguinetti (changes requested):

runtime-extension-model/src/main/resources/META-INF/mule-operation.xsd
eze210: maxOccurs=1 -> yeah, I just copied from other place, I'll fix it

declarer thing -> THANK YOU, I completely forgot it

asanguinetti (approved):

eze210:
--validate

asanguinetti (approved):

rbourbonnavarro:
--validate

PR #11522, opened 2022-06-08
W-11267062: Cherry pick W-11177824 to 4.4.x (flow ref validations fails for xml sdk components that declare inner flows when using lazy init)

SebaElizalde:


SebaElizalde:
--validate

PR #11524, opened 2022-06-09
W-11063960: Exception deserializing complex parameters sent by OCS

ndinu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ResolverSetUtils.java
```diff
@@ -109,107 +127,170 @@ private static Map<String, ValueResolver> getParameterGroupValueResolvers(Parame
...
-                                                     muleContext));
-      }
+    Optional<Class<Object>> mapClassOptional = getType(type);
+    Class mapClass = mapClassOptional.get();
 
-      return MapValueResolver.of(expectedType, keyResolvers, valueResolvers, reflectionCache, muleContext);
-    } else {
-      return new StaticValueResolver(map);
+    MetadataType valueType = type.getOpenRestriction().orElse(null);
+    Function<Object, ValueResolver<Object>> valueValueResolverFunction = valueType != null
```
marianogonzalez, line 275: rewrite this code for clarity. The ternary + the lambda makes this hard to read and debug

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/resolver/ValueResolverFactoryTypeVisitor.java
```diff
@@ -69,6 +70,15 @@ public ValueResolverFactoryTypeVisitor(DslSyntaxResolver dslSyntaxResolver, Stri
     this.dslSyntaxResolver = dslSyntaxResolver;
     this.defaultValue = defaultValue;
     this.acceptsReferences = acceptsReferences;
+    initializeValueResolverDelegates();
+  }
+
+  private void initializeValueResolverDelegates() {
+    valueResolverDelegates.put(metadataType -> MediaType.class.equals(getType(metadataType)),
```
marianogonzalez, line 77: why is this implemented like this. Sounds super overkill. Specially the DataType part.

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/resolver/ValueResolverFactoryTypeVisitor.java
ndinu: Before, this class was in the extensions-spring-support module, and relied on ValueResolverParsingDelegate like this one : https://github.com/mulesoft/mule/blob/master/modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/object/CharsetValueResolverParsingDelegate.java . 

The ValueResolverFactoryTypeVisitor only used 2, which only made use of the MetadataType and the value given, so I turned those 2 into the valueResolverDelegates map. The predicates are the implementation the accept method had, and the given value resolver is the same as the parse method would returned.

marianogonzalez: I understand that. but this implementation is still criptic and inneficient. For instance, in the original implementation the map had several entries, but no real need to keep using a map if it's only two now.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ResolverSetUtils.java
```diff
@@ -109,107 +127,170 @@ private static Map<String, ValueResolver> getParameterGroupValueResolvers(Parame
...
 
-    valueResolver = new TypeSafeValueResolverWrapper(valueResolver, expectedType);
 
-    muleContext.getInjector().inject(valueResolver);
-    initialiseIfNeeded(valueResolver);
+      return new ValueResolver() {
+
+        @Override
+        public Object resolve(ValueResolvingContext context) throws MuleException {
+          Object result = objectBuilder.build(context);
```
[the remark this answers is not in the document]
ndinu: 👌

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ResolverSetUtils.java
```diff
@@ -109,107 +127,170 @@ private static Map<String, ValueResolver> getParameterGroupValueResolvers(Parame
...
+          muleContext.getInjector().inject(result);
```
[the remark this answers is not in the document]
ndinu: You are right, this was not being done before. The need of using the injector in the output value comes from the fact that when the beans/pojos are parsed from the dsl and created as a spring bean they have their dependencies injected. In this case we need to maintain that behavior.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ResolverSetUtils.java
ndinu: 👌

marianogonzalez (approved):

PR #11538, opened 2022-06-14
W-10964385: Make ExtensionModelDiscoverer#discoverRuntimeExtensionModels static

elrodro83:
* Fix to make the discovery return runtime extension models as well


elrodro83:
--validate

PR #11554, opened 2022-06-21
W-11302803: Log a warning when trying to dispose a ByteBufferProvider…

TomoDC:
… not present

### DIFF
--- a/modules/memory-management/tests/src/test/java/org/mule/runtime/internal/memory/management/DefaultMemoryManagementServiceTestCase.java
+++ b/modules/memory-management/tests/src/test/java/org/mule/runtime/internal/memory/management/DefaultMemoryManagementServiceTestCase.java
@@ -85,6 +85,8 @@ public class DefaultMemoryManagementServiceTestCase extends AbstractMuleTestCase
   @Test
   @Description("When disposing a not present bytebuffer provider, no exceptions should be raised.")
   public void disposingByteBufferProviderNotPresent() {
-    memoryManagementService.disposeByteBufferProvider("notPresentByteBuffer");
+    String notPresentByteBuffer = "notPresentByteBuffer";
+    memoryManagementService.disposeByteBufferProvider(notPresentByteBuffer);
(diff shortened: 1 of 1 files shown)


modules/memory-management/src/main/java/org/mule/runtime/internal/memory/management/DefaultMemoryManagementService.java
```diff
@@ -79,7 +83,12 @@ public synchronized ByteBufferProvider<ByteBuffer> getByteBufferProvider(String
 
   @Override
   public synchronized void disposeByteBufferProvider(String name) {
-    byteBufferProviders.remove(name).dispose();
+    ByteBufferProvider<ByteBuffer> bufferProvider = byteBufferProviders.remove(name);
+    if (bufferProvider != null) {
+      bufferProvider.dispose();
+    } else {
+      LOGGER.warn("Unable to dispose not present ByteBufferProvider " + name);
```
eze210, line 90:
```suggestion
      LOGGER.warn("Unable to dispose not present ByteBufferProvider '{}'", name);
```

modules/memory-management/src/main/java/org/mule/runtime/internal/memory/management/DefaultMemoryManagementService.java
```diff
@@ -79,7 +83,12 @@ public synchronized ByteBufferProvider<ByteBuffer> getByteBufferProvider(String
 
   @Override
   public synchronized void disposeByteBufferProvider(String name) {
-    byteBufferProviders.remove(name).dispose();
+    ByteBufferProvider<ByteBuffer> bufferProvider = byteBufferProviders.remove(name);
+    if (bufferProvider != null) {
+      bufferProvider.dispose();
```
eze210, line 88: Add an unit test

eze210 (changes requested):

eze210 (approved):

PR #11563, opened 2022-06-24
W-11336719: Domain classloader creation is regenerating the container

elrodro83:
classloader instead of using a previously generated one


modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/classloader/ArtifactClassLoaderResolverConstants.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Provides constants to be used internally for the generation of classloaders
+ *
+ * @since 4.5
+ */
+public class ArtifactClassLoaderResolverConstants {
+
+  /**
+   * A {@link ModuleRepository} for the modules available in the same classpath as this class.
```
rbourbonnavarro, line 25: Shouldn't the class loader of `ArtifactClassLoaderResolverConstants` be used instead of `ArtifactClassLoaderResolver` for this to be strictly true?

rbourbonnavarro (changes requested):

rbourbonnavarro (approved):

PR #11565, opened 2022-06-24
W-10877267: A.11.2 Implement the automatic propagation of trace context and baggage

fsgonz:


core/src/main/java/org/mule/runtime/core/api/event/EventContextFactory.java
```diff
@@ -81,6 +85,24 @@ static EventContext create(String id, String serverId, ComponentLocation locatio
...
+   * @param location                            the location of the component that received the first message for this context.
+   * @param correlationId                       See {@link EventContext#getCorrelationId()}.
+   * @param externalCompletion                  future that completes when source completes enabling termination of
+   *                                            {@link BaseEventContext} to depend on completion of source.
+   * @param sdkDistributedTraceContextMapGetter the {@link SdkDistributedTraceContextMapGetter} used to retrieve the distributed
+   *                                            trace context.
+   */
+  static EventContext create(FlowConstruct flow, ComponentLocation location, String correlationId,
+                             Optional<CompletableFuture<Void>> externalCompletion,
+                             SdkDistributedTraceContextMapGetter sdkDistributedTraceContextMapGetter) {
```
marianogonzalez, line 101: creation of the mule event should not be coupled to an SDK concept. Whatever tracing information is neeeded at the event level should be modeled as part of the event. The SDK must do nothing but provide a facade to that.

core/src/main/java/org/mule/runtime/core/internal/event/trace/extractor/ComposedTraceContextFieldExtractor.java
```diff
@@ -0,0 +1,36 @@
...
+  private final TraceContextFieldExtractor[] traceContextFieldExtractors;
+
+  public ComposedTraceContextFieldExtractor(TraceContextFieldExtractor... traceContextFieldExtractors) {
+    this.traceContextFieldExtractors = traceContextFieldExtractors;
+  }
+
+  @Override
+  public Map<String, String> extract(SdkDistributedTraceContextMapGetter sdkDistributedTraceContextMapGetter) {
+    Map<String, String> result = new HashMap<>();
+    Arrays.stream(traceContextFieldExtractors).forEach(traceContextFieldExtractor -> result
```
marianogonzalez, line 32: don't use java streams on anything that's meant to run several times a second at runtime.

marianogonzalez (changes requested):

PR #11571, opened 2022-06-28
W-11299968: Make ResolverSetUtils adopt the ComponentParameterization API

ndinu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ComponentParameterizationUtils.java
```diff
@@ -0,0 +1,40 @@
...
+  /**
+   * Creates a {@link ComponentParameterization} of a {@link ParameterizedModel} based on static values of its parameters. Keep in
+   * mind that this method will not work if the given {@link ParameterizedModel} contains parameter that share the same name,
+   * since the map representation falls short in that case.
+   * 
+   * @param parameterizedModel parameterizedModel to be described by the result
+   * @param parameters         static parameter values.
+   * @return a component parameterization that describes the given value.
+   */
+  public static ComponentParameterization createComponentParameterization(ParameterizedModel parameterizedModel,
```
marianogonzalez, line 31: isn't this better suited as a static method in ComponentParameterization or builder itself?

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/ComponentParameterizationUtils.java
ndinu: I didn't want to expose this in the API, since :
1- this creating will only work if the parameterized model does not have parameters with the same name 
2- will should have a way to define the parameter values that express if it is an array/pojo/simple value, etc. I don't think we should encourage this way of defining it.

Do you think it would be better to have it in the ComponentParameterization API? I could add a method to the builder that receives the map.

marianogonzalez (approved):

PR #11574, opened 2022-06-29
W-10998630: Remove warning: "An expression value was given for parameter 'variableName' but it doesn't support expressions" from logging

sofiamorseletto:


runtime-extension-model/src/test/java/org/mule/runtime/config/internal/validation/SetVariableVariableNameValidationTestCase.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.config.internal.validation;
+
+import static org.hamcrest.CoreMatchers.is;
```
fsgonz, line 9: order of static imports.

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -481,7 +481,7 @@ private void declareSetVariable(ExtensionDeclarer extensionDeclarer) {
     setVariable.onDefaultParameterGroup()
         .withOptionalParameter("variableName")
         .ofType(STRING_TYPE)
-        .withExpressionSupport(NOT_SUPPORTED)
+        .withExpressionSupport(SUPPORTED)
```
fsgonz, line 484: Add a fail safe mechanism for this.

fsgonz (changes requested):

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
fsgonz: Add the corresponding test when you add the fail safe mechanism.

sofiamorseletto:
--validate

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -1394,4 +1408,9 @@ private void declareSecurityFilter(ExtensionDeclarer extensionDeclarer) {
         .describedAs("The name of the encryption strategy to use. This should be configured using the "
             + "'password-encryption-strategy' element, inside a 'security-manager' element at the top level.");
   }
+
+  public static void refreshSystemProperties() {
+    SUPPORT_EXPRESSIONS_IN_VARIABLE_NAME_IN_SET_VARIABLE =
+        parseBoolean(getProperty(SUPPORT_EXPRESSIONS_IN_VARIABLE_NAME_IN_SET_VARIABLE_PROPERTY, "true"));
```
fsgonz, line 1414: use getBoolean with default false.  and rename the variable to indicate that this is a kill switch. REVERT_SUPPORT_EXPRESSIONS_N_SET_VARIABLE_PROPERTY.
Boolean.getBoolean(REVERT_SUPPORT_EXPRESSIONS_N_SET_VARIABLE_PROPERTY)

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -478,11 +484,19 @@ private void declareSetVariable(ExtensionDeclarer extensionDeclarer) {
     setVariable.withOutput().ofType(VOID_TYPE);
     setVariable.withOutputAttributes().ofType(VOID_TYPE);
 
-    setVariable.onDefaultParameterGroup()
-        .withOptionalParameter("variableName")
-        .ofType(STRING_TYPE)
-        .withExpressionSupport(NOT_SUPPORTED)
-        .describedAs("The name of the variable.");
+    if (SUPPORT_EXPRESSIONS_IN_VARIABLE_NAME_IN_SET_VARIABLE) {
```
fsgonz, line 487: revert this to make it more clear that this is a kill switch.

fsgonz (changes requested):

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -1394,4 +1408,9 @@ private void declareSecurityFilter(ExtensionDeclarer extensionDeclarer) {
         .describedAs("The name of the encryption strategy to use. This should be configured using the "
             + "'password-encryption-strategy' element, inside a 'security-manager' element at the top level.");
   }
+
+  public static void refreshSystemProperties() {
```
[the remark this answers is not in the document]
fsgonz: I think an exception in mule-api would be enough in practical terms, but just to not break api formally.

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
sofiamorseletto: As far as I know getBoolean does not seem to support adding a default value, so I need to keep `parseBoolean(getProperty...`

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -160,10 +161,11 @@ class MuleExtensionModelDeclarer {
   final ErrorModel validationError = newError(VALIDATION).withParent(anyError).build();
   final ErrorModel duplicateMessageError = newError(DUPLICATE_MESSAGE).withParent(validationError).build();
 
-  private static boolean SUPPORT_EXPRESSIONS_IN_VARIABLE_NAME_IN_SET_VARIABLE =
-      parseBoolean(getProperty(SUPPORT_EXPRESSIONS_IN_VARIABLE_NAME_IN_SET_VARIABLE_PROPERTY, "true"));
+  private boolean REVERT_SUPPORT_EXPRESSIONS_IN_VARIABLE_NAME_IN_SET_VARIABLE;
 
   ExtensionDeclarer createExtensionModel() {
+    REVERT_SUPPORT_EXPRESSIONS_IN_VARIABLE_NAME_IN_SET_VARIABLE =
```
fsgonz, line 167: Use Boolean.getBoolean(REVERT_SUPPORT_EXPRESSIONS_IN_VARIABLE_NAME_IN_SET_VARIABLE_PROPERTY). That directly verifies if the system property is set to true. In any other case (no system property) it assumes is false. Use that directly in the if.

fsgonz (changes requested):

PR #11581, opened 2022-07-04
W-11147961: Error Mapping not working while deploying Header Injection custom policy

sofiamorseletto:


tests/performance/src/main/java/org/mule/runtime/core/policy/CompositeOperationPolicyBenchmark.java
```diff
@@ -57,7 +57,7 @@ public Publisher<CoreEvent> apply(Publisher<CoreEvent> publisher) {
       }
     }, "")), empty(), (policy, nextProcessor) -> nextProcessor,
                                            muleContext.getConfiguration().getShutdownTimeout(),
-                                           fluxCompleteScheduler);
+                                           fluxCompleteScheduler, feature -> false);
```
sofiamorseletto, line 60: Changing to false on this version since the feature flag is not enabled by default in 4.3.x

sofiamorseletto:
--validate

PR #11583, opened 2022-07-05
W-10877267: A.11.2 Implement the automatic propagation of trace context

fsgonz:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/parameter/PropagateAllDistributedTraceContextManager.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  @Override
+  public void setRemoteTraceContextMap(Map<String, String> contextMap) {
+    this.contextMap = contextMap;
+  }
+
+  @Override
+  public Map<String, String> getRemoteTraceContextMap() {
+    return contextMap;
```
marianogonzalez, line 41: should use unmodifiableMap() here?

marianogonzalez (changes requested):

core/src/main/java/org/mule/runtime/core/internal/event/AbstractEventContext.java
```diff
@@ -48,7 +49,7 @@
  *
  * @since 4.0
  */
-abstract class AbstractEventContext implements BaseEventContext {
+abstract class AbstractEventContext implements DistributedTraceContextAware, BaseEventContext {
```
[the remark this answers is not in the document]
fsgonz: BaseContext is a privileged interface. I would rather not change anything related to api (even privileged apis), as this is only used internally.

core/src/main/java/org/mule/runtime/core/internal/event/DefaultEventContext.java
```diff
@@ -317,15 +322,23 @@ protected String basicToString() {
         + getOriginatingLocation().getRootContainerName() + " }";
   }
 
+  public DistributedTraceContext getDistributedTraceContext() {
```
marianogonzalez, line 325: are these two missing an `@Override`?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/parameter/PropagateAllDistributedTraceContextManager.java
```diff
@@ -27,8 +28,7 @@ public PropagateAllDistributedTraceContextManager(DistributedTraceContext distri
   }
 
   private void resolveContextMap(DistributedTraceContext distributedTraceContext) {
-    contextMap = new HashMap<>(distributedTraceContext.tracingFieldsAsMap());
-    contextMap.putAll(distributedTraceContext.baggageItemsAsMap());
+    contextMap = unmodifiableMap(distributedTraceContext.baggageItemsAsMap());
```
marianogonzalez, line 31: is this change correct? It seems like previously you were also including the tracing fields but not anymore. Which version is boggus?

marianogonzalez (changes requested):

marianogonzalez (approved):

PR #11590, opened 2022-07-06
W-11272392: support parameters that reference configurations

asanguinetti:


modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkParameterModelParserSdk.java
```diff
@@ -44,23 +52,83 @@ public class MuleSdkParameterModelParserSdk extends BaseMuleSdkExtensionModelPar
 
   protected final ComponentAst parameterAst;
   private final TypeLoader typeLoader;
+  private final ExtensionModelHelper extensionModelHelper;
 
   private String name;
+  private MetadataType type;
+  private List<StereotypeModel> allowedStereotypes = emptyList();
+  private Optional<ParameterDslConfiguration> parameterDslConfiguration = empty();
```
asanguinetti, line 60: Note that now we are going to be precalculating all these when instantiating the parser. This is similar to what we have been doing with other parsers so far. Also, consider that `ModelParsers` don't have a `parse` method.
This will make it easier to avoid repeated computations for the same value without needing to do a lazy value approach. The downside (which can actually be an upside depending how you look at it) is that parsing errors will be raised earlier (for example: even if the `type` is never queried). But I think that is actually a good thing in this context.

modules/extensions-mule-support/src/test/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkParameterModelParserSdkBuilder.java
```diff
@@ -0,0 +1,158 @@
...
+    return new MuleSdkParameterModelParserSdk(componentAst, typeLoader, extensionModelHelper);
+  }
+
+  private Set<ExtensionModel> loadExtensionModels() {
+    if (extensionDeclarers.isEmpty()) {
+      return emptySet();
+    }
+
+    return loadExtensionModels(createLoadingRequest(), new ExtensionModelFactory());
+  }
```
asanguinetti, line 140: For the `ExtensionModel` I actually found out that it was easier to load the real thing from the `ExtensionDeclarer` rather than mock everything. Mocking would have us messing too much with the internal hierarchy of models. Also we would have been missing possible tricks played by the enrichers.
Yes, the tests will be less "unitary", but I think that ship sailed anyway.

modules/extensions-mule-support/src/test/resources/models/app-as-mule-extension.json
```diff
@@ -3917,6 +3917,320 @@
...
+              "isConfigOverride": false,
+              "isComponentId": false,
+              "fieldValueProviderModels": [],
+              "expressionSupport": "SUPPORTED",
+              "role": "BEHAVIOUR",
+              "dslConfiguration": {
+                "allowsInlineDefinition": false,
+                "allowsReferences": false,
+                "allowTopLevelDefinition": false
+              },
```
asanguinetti, line 3981: This is curious. Note that the [parser is generating the right thing](https://github.com/mulesoft/mule/pull/11590/files#diff-d78636ed7912dc60ef62df5315b9e7bad8e0ea3fb3636681b9964ae20ce7c65aR125) (being tested [here](https://github.com/mulesoft/mule/pull/11590/files#diff-dd22cfd0f8a8adfb635ab40b8b7dbf928823defe304a8d1c5a52b84bb270a361R95)).
Still, the `ExtensionModel` seems not to be having the right DSL Configuration. I tried to track it down and it seems it is because of [this](https://github.com/mulesoft/mule-extensions-api/blob/83c0b01fc3822b0a194a0e8f48bc625d7cf3fc21/mule-extensions-api/src/main/java/org/mule/runtime/extension/internal/loader/enricher/ParameterDslDeclarationEnricher.java#L96). Looks like the `ParameterDslDeclarationEnricher` is overriding the `ParameterDslConfiguration` based on the `MetadataType` and, at that point, we are declaring it as String.
Any thoughts on this @ndinu ?

asanguinetti: I just changed it so that it does not attempt to indicate that the parameter accepts references.
I created the task [W-11402315](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000010zRGIYA2/view) to review that decision.
But if you all agree, we can still move forward with this one as it adds value on top of what we currently have and, since it is a completely new feature, we don't risk breaking anything.

mbuchwald (approved):

PR #11596, opened 2022-07-08
W-11403067: Wrong "Element DSL" in log when "Expose Headers" is enabl…

IvanAndresFritzler:
…ed on OAuth 2.0 access token enforcement policy


core-tests/src/test/java/org/mule/runtime/core/internal/policy/DefaultPolicyManagerTestCase.java
```diff
@@ -608,17 +635,17 @@ public void cachesEvictedDoesntIncreaseActivePoliciesCount() throws InterruptedE
 
   /**
    * Stubs a Policy, instead of mocking it. Mocking a policy would imply invocation interceptions of the form:
-   * 
+   *
```
fsgonz, line 638: undo change.

fsgonz (changes requested):

PR #11605, opened 2022-07-12
W-11348869: Mule SDK doesn't support List of java.math.BigDecimal/BigInteger parameters correctly

SebaElizalde:


modules/extensions-spring-support/src/test/resources/values/some-extension-big-numbers-config.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<mule xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xmlns:some="http://www.mulesoft.org/schema/mule/some"
+      xmlns="http://www.mulesoft.org/schema/mule/core"
+      xsi:schemaLocation="http://www.mulesoft.org/schema/mule/core http://www.mulesoft.org/schema/mule/core/current/mule.xsd
+               http://www.mulesoft.org/schema/mule/some http://www.mulesoft.org/schema/mule/some/current/mule-some.xsd">
+
```
ndinu, line 7: Add cases where the list is defined inline, it doesn't need to cover each combination, but at lease us all the cases. (the inline definition is what was causing troubles in the issue raised that is what I think we should also add those cases)

For this cases I suggest not use DW to declare the list items.

ndinu (changes requested):

ndinu (approved):

PR #11610, opened 2022-07-15
W-11358987: Add thrown errors to the Mule Operation extension model (APP local phase)

eze210:


modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkErrorModelParser.java
```diff
@@ -0,0 +1,48 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.module.extension.mule.internal.loader.parser;
+
+import org.mule.runtime.api.meta.model.error.ErrorModel;
+import org.mule.runtime.module.extension.internal.loader.parser.ErrorModelParser;
+
+import java.util.Optional;
+
+public class MuleSdkErrorModelParser implements ErrorModelParser {
```
marianogonzalez, line 14: jdoc

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkExtensionModelParser.java
```diff
@@ -179,6 +178,11 @@ public List<NotificationModel> getNotificationModels() {
     return emptyList();
   }
 
+  @Override
+  public String getNamespace() {
+    return "THIS";
```
marianogonzalez, line 183: extract to constant. Any other parsers we should be doing the same for?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/ErrorModelParser.java
```diff
@@ -25,9 +25,9 @@ public interface ErrorModelParser {
   String getType();
 
   /**
-   * @return {@code true} if it represents to a core Mule language error.
+   * @return the error's namespace
    */
-  boolean isMuleError();
```
marianogonzalez, line 30: I. undersetand why adding a getNamespace(), but why removing this method?

marianogonzalez (changes requested):

marianogonzalez (approved):

PR #11611, opened 2022-07-15
W-11425552: Add the ByteBufferProvider with a pool based on DW

TomoDC:


modules/memory-management/src/main/java/org/mule/runtime/internal/memory/bytebuffer/WeavePoolBasedByteBufferProvider.java
```diff
@@ -0,0 +1,180 @@
...
+
+  @Override
+  public void release(ByteBuffer buffer) {
+    deallocationDataProducer
+        .triggerProfilingEvent(new DefaultByteBufferProviderEventContext(name, currentTimeMillis(), buffer.limit()));
+    if (buffer.isDirect() && buffer.capacity() == capacity) {
+      // Release direct memory.
+      pool.release(buffer);
+    } else {
+      // Release heap memory.
```
TomoDC, line 75: Should I create a task about ByteBufferProviderEventContext to handle the type of memory allocated / released? (Direct or heap) So we can differentiate what kind of memory is being used.

modules/memory-management/tests/src/test/java/org/mule/runtime/internal/memory/bytebuffer/WeavePoolBasedByteBufferProviderTestCase.java
```diff
@@ -0,0 +1,102 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.internal.memory.bytebuffer;
+
+import io.qameta.allure.Feature;
```
fsgonz, line 9: static import order.

modules/memory-management/src/main/java/org/mule/runtime/internal/memory/bytebuffer/WeavePoolBasedByteBufferProvider.java
fsgonz: Yes, create one to add that info for profiling.

fsgonz (changes requested):

fsgonz (dismissed):

fsgonz (approved):

PR #11623, opened 2022-07-21
W-11472505: Spans export test infrastructure

fsgonz:


IvanAndresFritzler (changes requested):

core/src/main/java/org/mule/runtime/core/internal/profiling/DefaultProfilingService.java
```diff
@@ -193,4 +197,19 @@ private String getScope() {
...
     }
   }
+
+  @Override
+  public ExportedSpanCapturer getExportedSpanCapturer() {
+    return eventTracer.getSpanCapturer();
+  }
+
+  @Override
+  public void startComponentSpan(CoreEvent event, Component component) {
```
IvanAndresFritzler, line 207: Let's add a TODO here in order to tackle the FlowRunner limitations and be able to remove this "testing specific" logic.

core/src/main/java/org/mule/runtime/core/internal/profiling/DefaultProfilingService.java
```diff
@@ -193,4 +197,19 @@ private String getScope() {
       return getArtifactId(muleContext);
     }
   }
+
+  @Override
+  public ExportedSpanCapturer getExportedSpanCapturer() {
```
IvanAndresFritzler, line 202: I think we can improve the design here by adding an interface to the export of the traces.

IvanAndresFritzler (changes requested):

core/src/main/java/org/mule/runtime/core/internal/event/DefaultEventContext.java
```diff
@@ -325,7 +327,7 @@ protected String basicToString() {
   @Override
   public DistributedTraceContext getDistributedTraceContext() {
     if (distributedTraceContext == null) {
-      distributedTraceContext = emptyDistributedEventContext();
+      distributedTraceContext = builder().withGetter(emptyTraceContextMapGetter()).build();
```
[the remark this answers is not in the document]
fsgonz: not a constant because this will be used to carry the current span. But I can avoid resolving the empty trace context.

tests/component-plugin/src/main/java/org/mule/functional/api/flow/FlowRunner.java
```diff
@@ -221,9 +230,15 @@ private ExecutionCallback<CoreEvent> getFlowRunCallback() {
...
-          fluxSink.next(getOrBuildEvent());
+          CoreEvent event = getOrBuildEvent();
+          fluxSink.next(event);
           sinkReference.set(fluxSink);
-        }).transform(flow::apply).blockFirst();
+        }).doOnNext(event -> privilegedProfilingService
+            .ifPresent(privilegedProfilingService -> privilegedProfilingService.startComponentSpan(event, flow)))
+            .transform(flow::apply)
+            .doOnNext(event -> privilegedProfilingService
+                .ifPresent(privilegedProfilingService -> privilegedProfilingService.endComponentSpan(event)))
```
sofiamorseletto, line 240: If the flow fails, the span is never ended, maybe a doOnError should be added?

fsgonz: This is for capturing the spans in a test environment only. I think it will be correct for the span not to end.

fsgonz: I've change this to end the span in case an error happens.

IvanAndresFritzler (approved):

PR #11627, opened 2022-07-22
W-11476664: Global Error Handler's On-Error-Propagate is not rolling back tx since fix for W-10992817

marianorean:


marianorean:
--validate

mbuchwald (approved):

PR #11631, opened 2022-07-25
W-10620138: Support new SDK API metadata package (keys - input metadata)

Bardammu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/enricher/DynamicMetadataDeclarationEnricher.java
```diff
@@ -281,6 +289,10 @@ public org.mule.sdk.api.metadata.resolving.AttributesTypeResolver getAttributesR
           return new org.mule.sdk.api.metadata.NullMetadataResolver();
         }
       };
+      addQueryModelProperties(declaration, query);
+      declareDynamicType(declaration.getOutput());
+      declareMetadataKeyId(declaration, null);
```
ndinu, line 294: Why did these lines need to change the placement?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/KeyIdResolverModelParser.java
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Parses the syntactic definition of the metadata keys id so that the semantics reflected in it can be extracted in a uniform
+ * way, regardless of the actual syntax used by the extension developer.
+ *
+ * @see ExtensionModelParser
+ * @since 4.5.0
+ */
+public interface KeyIdResolverModelParser {
+
+  boolean hasKeyIdResolver();
```
ndinu, line 20: Suppose that this method gives false, what would the other 2 return?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/KeyIdResolverModelParser.java
```diff
@@ -0,0 +1,26 @@
...
+
+  boolean isPartialKeyResolver();
```
ndinu, line 22: Add jdocs to the methods

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaInputResolverModelParser.java
```diff
@@ -0,0 +1,51 @@
...
+
+import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
+import static org.mule.runtime.core.api.util.ClassUtils.getClassName;
+
+import org.mule.runtime.api.exception.MuleRuntimeException;
+import org.mule.runtime.core.api.util.ClassUtils;
+import org.mule.runtime.module.extension.internal.metadata.SdkInputTypeResolverAdapter;
+import org.mule.sdk.api.metadata.resolving.InputTypeResolver;
+
+public class JavaInputResolverModelParser {
```
ndinu, line 18: Is there no interface for this class to implement?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaInputResolverModelParser.java
```diff
@@ -0,0 +1,51 @@
...
+    this.parameterName = parameterName;
+    this.inputTypeResolverDeclarationClass = inputTypeResolverDeclarationClass;
+    this.muleResolver = muleResolver;
+  }
+
+  public String getParameterName() {
+    return parameterName;
+  }
+
+  public InputTypeResolver getInputResolver() {
```
ndinu, line 34: I am debating whether is would be better to use the Mule interface here.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaInputResolverModelParser.java
```diff
@@ -0,0 +1,51 @@
...
+import org.mule.runtime.module.extension.internal.metadata.SdkInputTypeResolverAdapter;
+import org.mule.sdk.api.metadata.resolving.InputTypeResolver;
+
+public class JavaInputResolverModelParser {
+
+  private final String parameterName;
+  private final Class<?> inputTypeResolverDeclarationClass;
+  private final boolean muleResolver;
+
+  public JavaInputResolverModelParser(String parameterName, Class<?> inputTypeResolverDeclarationClass, boolean muleResolver) {
```
ndinu, line 24: Receiving a boolean whether it is a mule resolver or not seems to be a smell, why not delegate that responsability to what creates the resolver instance?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaInputResolverModelParser.java
```diff
@@ -0,0 +1,51 @@
...
+
+  public InputTypeResolver getInputResolver() {
+    return instantiateResolver(inputTypeResolverDeclarationClass);
+  }
+
+  private InputTypeResolver instantiateResolver(Class<?> factoryType) {
+    try {
+      Object resolver = ClassUtils.instantiateClass(factoryType);
+      if (muleResolver) {
+        return new SdkInputTypeResolverAdapter((org.mule.runtime.api.metadata.resolving.InputTypeResolver) resolver);
```
ndinu, line 42: What if the factoryType is the sdk-api NullTypeResolver?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaKeyIdResolverModelParser.java
```diff
@@ -0,0 +1,93 @@
...
+                                      boolean muleResolver) {
+    this.metadataType = metadataType;
+    this.categoryName = categoryName;
+    this.keyIdResolverDeclarationClass = keyIdResolverDeclarationClass;
+    this.muleResolver = muleResolver;
+  }
+
+  @Override
+  public boolean hasKeyIdResolver() {
+    return (!org.mule.sdk.api.metadata.NullMetadataResolver.class.isAssignableFrom(keyIdResolverDeclarationClass) &&
```
ndinu, line 47: There should be a utils to check if something is a null resolver, I think we have this logic in other places also.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaKeyIdResolverModelParser.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  @Override
+  public boolean isPartialKeyResolver() {
+    return PartialTypeKeysResolver.class.isAssignableFrom(keyIdResolverDeclarationClass) ||
+        org.mule.runtime.api.metadata.resolving.PartialTypeKeysResolver.class.isAssignableFrom(keyIdResolverDeclarationClass);
+  }
+
+  @Override
+  public org.mule.sdk.api.metadata.resolving.TypeKeysResolver getKeyResolver() {
```
ndinu, line 59: Why is there a need to use the FQCN of TypeKeysResolver here?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaInputResolverModelParser.java
```diff
@@ -0,0 +1,51 @@
...
+  private final boolean muleResolver;
+
+  public JavaInputResolverModelParser(String parameterName, Class<?> inputTypeResolverDeclarationClass, boolean muleResolver) {
+    this.parameterName = parameterName;
+    this.inputTypeResolverDeclarationClass = inputTypeResolverDeclarationClass;
+    this.muleResolver = muleResolver;
+  }
+
+  public String getParameterName() {
+    return parameterName;
```
ndinu, line 31: If the parser is attached to a parameter model parser, then this would not be necessary. Maybe when doing the decouple task we can delete this method.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataScopeAdapter.java
```diff
@@ -26,15 +24,19 @@ default boolean isCustomScope() {
     return hasOutputResolver() || hasInputResolvers();
   }
 
+  boolean isPartialKeyResolver();
```
ndinu, line 27: what if the is no key resolver? what does this method return?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MuleInputTypeResolverAdapter.java
```diff
@@ -0,0 +1,61 @@
...
+
+public class MuleInputTypeResolverAdapter implements InputTypeResolver {
+
+  private final org.mule.sdk.api.metadata.resolving.InputTypeResolver delegate;
+
+  MuleInputTypeResolverAdapter(org.mule.sdk.api.metadata.resolving.InputTypeResolver delegate) {
+    this.delegate = delegate;
+  }
+
+  public static InputTypeResolver from(Object resolver) {
```
ndinu, line 27: I feel like this logic if repeated (although for the other types of resolvers) in a few places. And the null resolver part is duplicated.

Maybe have some adapter factory class that resolves this. See that for example. the static method may not even create a MuleInputTypeResolverAdapter.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MuleTypeKeysResolverAdapter.java
```diff
@@ -0,0 +1,72 @@
...
+  private final org.mule.sdk.api.metadata.resolving.TypeKeysResolver delegate;
+
+  MuleTypeKeysResolverAdapter(org.mule.sdk.api.metadata.resolving.TypeKeysResolver delegate) {
+    this.delegate = delegate;
+  }
+
+  public static TypeKeysResolver from(Object resolver) {
+    checkArgument(resolver != null, "Cannot adapt null resolver");
+
+    if (resolver instanceof TypeKeysResolver) {
```
ndinu, line 38: Same as comment above

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/OutputResolverFactory.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Factory of {@link OutputTypeResolver}
+ */
+public class OutputResolverFactory {
+
+  public static OutputTypeResolver of(Object resolver) {
+    checkArgument(resolver != null, "Cannot create null resolver");
+
+    if (resolver instanceof OutputTypeResolver) {
```
ndinu, line 24: Same as above

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/OutputResolverFactory.java
```diff
@@ -0,0 +1,37 @@
...
+
+import static java.lang.String.format;
+import static org.mule.runtime.api.util.Preconditions.checkArgument;
+
+import org.mule.runtime.api.metadata.resolving.OutputTypeResolver;
+import org.mule.sdk.api.metadata.NullMetadataResolver;
+
+/**
+ * Factory of {@link OutputTypeResolver}
+ */
```
ndinu, line 18: add since version

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/KeyIdResolverModelParser.java
Bardammu: `isPartialKeyResolver` will give false, `getKeyResolver` will give the null metadata resolver

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaInputResolverModelParser.java
Bardammu: To avoid the use of instanceof. I'll remove the boolean if it's clearer

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaInputResolverModelParser.java
Bardammu: It'll return an instance of `NullTypeResolver`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaInputResolverModelParser.java
Bardammu: Yep

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/OutputResolverFactory.java
Bardammu: Actually I shouldn't have included this class

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/enricher/DynamicMetadataDeclarationEnricher.java
Bardammu: To use the metadataScope defined before and avoid using instanceOf on parseMetadataKeyAnnotations

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataScopeAdapter.java
Bardammu: It should return false

Bardammu:
--validate

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/DefaultMetadataScopeAdapter.java
```diff
@@ -60,32 +48,32 @@
  */
 public final class DefaultMetadataScopeAdapter implements MetadataScopeAdapter {
 
-  private Supplier<NullMetadataResolver> nullMetadataResolverSupplier = new NullMetadataResolverSupplier();
-  private Supplier<? extends TypeKeysResolver> keysResolver = nullMetadataResolverSupplier;
-  private Map<String, Supplier<? extends InputTypeResolver>> inputResolvers = new HashMap<>();
+  private JavaKeyIdResolverModelParser javaKeyIdResolverModelParser =
+      new JavaKeyIdResolverModelParser(null, null, org.mule.sdk.api.metadata.NullMetadataResolver.class);
```
ndinu, line 52: Create a NullKeyIdResolverModelParser, which receives no arguments in the constructor and directly resolves to the correct values. This also goes for the other model parsers.

ndinu (approved):

Bardammu:
--validate

PR #11634, opened 2022-07-25
W-10875483: A.2.2 Implementation of the minimum set of attributes

fsgonz:


core/src/main/java/org/mule/runtime/core/internal/profiling/tracing/event/span/CoreEventSpanCustomizer.java
```diff
@@ -22,8 +28,21 @@ public interface CoreEventSpanCustomizer {
    *
    * @param coreEvent the {@link CoreEvent} to resolve the span name from.
    * @param component the {@link Component} to resolve the span name from.
-   * @return
+   * @return the name
```
IvanAndresFritzler, line 31: ... of the span?

core/src/main/java/org/mule/runtime/core/internal/profiling/tracing/export/OpentelemetrySpanExporter.java
```diff
@@ -59,6 +59,10 @@ public OpentelemetrySpanExporter(Tracer tracer, EventContext eventContext,
 
   @Override
   public void export(InternalSpan internalSpan) {
+    for (Map.Entry<String, String> entry : internalSpan.attributesAsMap().entrySet()) {
```
IvanAndresFritzler, line 62: Is this ok to be here or should we add the attributes to the OTEL span as soon as they are added to our internal span?

IvanAndresFritzler (changes requested):

core/src/main/java/org/mule/runtime/core/internal/profiling/tracing/export/OpentelemetrySpanExporter.java
fsgonz: In practical terms, it is the same as the export in opentelemetry happens when you set the end, but I've changed it for future developments.

IvanAndresFritzler (dismissed):

fsgonz:
--validate

fsgonz:
--validate

fsgonz:
--validate

IvanAndresFritzler (dismissed):

IvanAndresFritzler (approved):

PR #11641, opened 2022-07-26
W-11493854: Produce trigger, private and active flows metrics

marianogonzalez:


core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
```diff
@@ -522,22 +529,27 @@ protected void doStart() throws MuleException {
...
-        throw ce;
-      } catch (Exception e) {
-        // If the source couldn't be started we would need to stop the pipeline (if possible) in order to leave
-        // its LifecycleManager also as initialise phase so the flow can be disposed later
-        stopOnFailure(e);
       }
+      allStatistics.addActiveTriggerFlow();
+    } else {
+      allStatistics.addActivePrivateFlow();
     }
```
mbuchwald, line 553:
```java
if (source == null) {
    allStatistics.addActivePrivateFlow();
    return;
}
// main code
```

PR #11650, opened 2022-08-02
W-11503016: Reuse reactor chains for global error handlers to use the…

marianorean:
… default processing strategy


marianorean:
--validate

marianorean:
--validate

marianorean:
--validate

core/src/main/java/org/mule/runtime/core/internal/exception/GlobalErrorHandler.java
```diff
@@ -24,11 +29,23 @@ public class GlobalErrorHandler extends ErrorHandler {
   private final AtomicBoolean initialised = new AtomicBoolean(false);
   private final AtomicInteger started = new AtomicInteger(0);
 
+  private Map<MessageProcessorChain, Consumer<Exception>> routers = new HashMap<>();
```
fsgonz, line 32: Can we add this to a different component so that we can have a simple unit test for this and also inject it as a dependency so that we can test that when. Or is there a way to have unit tests that verifies that when you stop start the application errors or invoke this callbaks here this is correctly cleared?

https://github.com/mulesoft/mule/pull/11650/files#diff-b20d355a087f0f0386461d0c91479f50813b97bff87d37b597ac45f4f53405c5R229

fsgonz (approved):

PR #11652, opened 2022-08-05
W-11464483: Raise Error usage within a reusable Mule Operation

eze210:


eze210:
--validate

eze210:
--validate

eze210:
--validate

eze210:
--validate

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/api/processor/MuleSdkRaiseErrorProcessor.java
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Processor capable of raising errors within a Mule Operation's body on demand, given a type and optionally a message.
+ *
+ * @since 4.5
+ */
+public final class MuleSdkRaiseErrorProcessor extends AbstractRaiseErrorProcessor {
+
+  @Override
+  protected ComponentIdentifier calculateErrorIdentifier(String typeId) {
+    return builder().namespace("THIS").name(typeId).build();
```
asanguinetti, line 23: Have we checked what would happen if someone tries to do something like this?
`<operation:raise-error type="THIS:CUSTOM" />`
Or even:
`<operation:raise-error type="OTHER:CUSTOM" />`

Shouldn't we have a test for those cases showing the expected outcome (maybe error in both, or maybe ignoring the namespace in the first one)?

asanguinetti (changes requested):

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/api/processor/MuleSdkRaiseErrorProcessor.java
asanguinetti: Also, we should probably extract the string "THIS" to a shared constant so can track its usages more easily.
Maybe making this one public or moving it to a constants class: https://github.com/mulesoft/mule/blob/eb7c7bb88896929865674e4f36aed563311ff565/modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkExtensionModelParser.java#L55

eze210:
--validate

eze210:
--validate

eze210:
--validate

eze210:
--validate

asanguinetti (approved):

PR #11666, opened 2022-08-10
W-11492717: Polling Source notifications Improvements

SebaElizalde:


SebaElizalde:
--validate

core/src/main/java/org/mule/runtime/core/internal/execution/FlowProcessMediator.java
```diff
@@ -217,6 +219,10 @@ public void process(FlowProcessTemplate template,
       final CoreEvent event = createEvent(template, messageSource,
                                           responseCompletion, flowConstruct);
 
+      template.getSourceMessage().getNotification().ifPresent(n -> {
```
ndinu, line 222: The SourceResultAdapter class says that the notification is an Object, if the notification is not a PollingSourceItemNotification, it will fail. Also, if in the future we need to do the same with another kind of annotation, the getNotification methods falls short.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -284,16 +311,20 @@ private DefaultPollContext(SourceCallback<T, A> sourceCallback, Serializable cur
       this.updatedWatermark = updatedWatermark;
       this.currentPollItems = 0;
       this.minimumRejectedByLimitPassingWatermark = null;
+      this.timestamp = ZonedDateTime.now();
+    }
+
+    public String getPollId() {
+      return componentLocation.getLocation() + " | " + timestamp;
```
ndinu, line 318: We should discuss if we want the id make this id to have this information included. A random UID is also a possibility

SebaElizalde: We talked about this in the spec stage, but we can reopen the discussion. As Composer will display this ID we thought it would be best for it to be as user friendly as possible. What are the advantages of a random UID that you think of?

ndinu: I think that since the composer user does not directly create the mule app, maybe the names that composer uses for flows may not be the most descriptive. This will also add a /source to the flow name. 

I am not against this option, since the flow name will give useful information, just bringing this to the table.

core/src/main/java/org/mule/runtime/core/internal/execution/SourceResultAdapter.java
```diff
@@ -31,6 +33,7 @@ public class SourceResultAdapter {
   private final MediaType mediaType;
   private final Optional<String> correlationId;
   private final PayloadMediaTypeResolver payloadMediaTypeResolver;
+  private final Optional<Object> notification;
```
marianogonzalez, line 36: Bad impl:

* Imposes artificial limitation of a single notification per instance
* Sources already have mechanism for firing notifications
* Leads to blind cast called out above

core/src/main/java/org/mule/runtime/core/internal/execution/FlowProcessMediator.java
```diff
@@ -217,6 +219,10 @@ public void process(FlowProcessTemplate template,
...
+        ((PollingSourceItemNotification) n).setEventId(event.getContext().getId());
```
marianogonzalez, line 223: bad practice blind cast

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/source/PollingSourceNotificationsTestCase.java
```diff
@@ -0,0 +1,252 @@
...
+public class PollingSourceNotificationsTestCase extends AbstractExtensionFunctionalTestCase {
+
+  private static final List<CoreEvent> ADOPTION_EVENTS = new LinkedList<>();
+
+  @Rule
+  public SystemProperty emitNotifications = new SystemProperty("mule.emit.polling.source.notifications", "true");
+
+  @Override
+  protected boolean mustRegenerateExtensionModels() {
+    return true;
```
marianogonzalez, line 62: why?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
marianogonzalez: @ndinu: the feature needs to be generic and not tailored to how composer may or may not name flows
@SebaElizalde: The full component location doesn't make sense as these are sources and they'll always be at the same coordinate. What you should use is the container flow name. Also, `@` is a more clear separator than `|`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -754,6 +792,29 @@ private void shutdownScheduler() {
     }
   }
 
+  private int matchAction(PollContext.PollItemStatus status, boolean currentPollItemLimitApplied) {
```
marianogonzalez, line 795: improve method name. matchAction to what? `statusToNotificationType` is probably a better name

marianogonzalez (changes requested):

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/source/PollingSourceNotificationsTestCase.java
SebaElizalde: this is needed to access the poll item limit feature we added

marianogonzalez: not the answer i was looking for. Why do you need to regenerate the entire extension model just for this?

core/src/main/java/org/mule/runtime/core/internal/execution/SourceResultAdapter.java
```diff
@@ -48,7 +50,7 @@ public SourceResultAdapter(Result<?, ?> result,
                              MediaType mediaType,
                              boolean isCollection,
                              Optional<String> correlationId) {
-    this(result, cursorProviderFactory, mediaType, isCollection, correlationId, null);
+    this(result, cursorProviderFactory, mediaType, isCollection, correlationId, null, null);
```
marianogonzalez, line 53: why would this be optional? Avoid nulls

core/src/main/java/org/mule/runtime/core/internal/execution/ItemInformation.java
```diff
@@ -0,0 +1,41 @@
...
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.internal.execution;
+
+import java.io.Serializable;
+import java.util.Optional;
+
+public class ItemInformation {
```
marianogonzalez, line 12: this is not item information, this is poll information

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -307,6 +339,15 @@ public PollItemStatus accept(Consumer<PollItem<T, A>> consumer) {
           status = FILTERED_BY_WATERMARK;
         } else if (currentPollItems < maxItemsPerPoll) {
           currentPollItems++;
+          if (EMIT_NOTIFICATIONS) {
+            callbackContext
+                .addVariable(ACCEPTED_ITEM_INFORMATION,
+                             new ItemInformation(getPollId(), itemId, pollItem.getWatermark(), componentLocation.getLocation()));
+            /*
```
marianogonzalez, line 346: remove

core/src/main/java/org/mule/runtime/core/internal/execution/SourceResultAdapter.java
```diff
@@ -127,6 +132,15 @@ public PayloadMediaTypeResolver getPayloadMediaTypeResolver() {
     return payloadMediaTypeResolver;
   }
 
+  /**
+   * @return information about the polled item that originated the message
```
marianogonzalez, line 136: bad javadoc. It doesn't explain that this would only be available when the message has been generated by a PollingSource

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -97,13 +116,12 @@
...
   private static final String UPDATE_PROCESSED_LOCK = "OSClearing";
   private static final String INFLIGHT_IDS_OS_NAME_SUFFIX = "inflight-ids";
-  private static final String POLLING_SOURCE_EXECUTOR_KEY = "Polling source executor";
-  private static final String RUNNABLE_KEY = "Runnable";
 
   private final PollingSource<T, A> delegate;
   private final SchedulingStrategy scheduler;
   private final int maxItemsPerPoll;
   private final SystemExceptionHandler systemExceptionHandler;
+  private final boolean EMIT_NOTIFICATIONS = parseBoolean(System.getProperty(EMIT_POLLING_SOURCE_NOTIFICATIONS));
```
marianogonzalez, line 124: this should be static. Also, static imports.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -81,6 +98,8 @@
  */
 public class PollingSourceWrapper<T, A> extends SourceWrapper<T, A> implements Restartable {
 
+  public static final String ACCEPTED_ITEM_INFORMATION = "mule-polling-source-accepted-item-information";
```
marianogonzalez, line 101: undescriptive name

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -754,6 +800,29 @@ private void shutdownScheduler() {
...
+        return ITEM_REJECTED_WATERMARK;
+      case ALREADY_IN_PROCESS:
+        return ITEM_REJECTED_IDEMPOTENCY;
+      case SOURCE_STOPPING:
+        return ITEM_REJECTED_SOURCE_STOPPING;
+    }
+    return NO_ACTION_ID;
+  }
+
+  private void dispatchPollingSourceNotification(Notification notification) {
```
marianogonzalez, line 820: remove this method. It's only purpose is to perform the if, but you already instantiated the notification, loosing cycles and GC pressure on collecting instances that will be most of the time discarded.

marianogonzalez (changes requested):

core/src/main/java/org/mule/runtime/core/internal/execution/SourceResultAdapter.java
SebaElizalde: because it will only be present when the message originated from a polling source

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
SebaElizalde: this is needed to propagate the polling information until the event is created and the notification can be sent. unless we find another mechanism/way, this can't be removed

core/src/main/java/org/mule/runtime/core/internal/execution/PollItemInformation.java
```diff
@@ -0,0 +1,41 @@
...
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.internal.execution;
+
+import java.io.Serializable;
+import java.util.Optional;
+
+public class PollItemInformation {
```
marianogonzalez, line 12: jdoc

core/src/main/java/org/mule/runtime/core/internal/execution/SourceResultAdapter.java
marianogonzalez: then receive it as optional

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/source/PollingSourceNotificationsTestCase.java
marianogonzalez: still unresolved

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/source/PollingSourceNotificationsTestCase.java
```diff
@@ -0,0 +1,260 @@
...
+      }
+    };
+    notificationListenerRegistry.registerListener(sourceListener);
+
+    try {
+      startFlow("oneItemPoll");
+      boolean timeout = !latch.await(TIMEOUT, MILLISECONDS);
+      assertThat(timeout, is(false));
+      assertThat(sourceNotifications.size(), is(2));
+      assertThat(sourceNotifications.get(0).getActionName(), is("poll started"));
```
marianogonzalez, line 114: use constants

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/source/PollingSourceNotificationsTestCase.java
```diff
@@ -0,0 +1,260 @@
...
+        latch.release();
+      }
+    };
+    notificationListenerRegistry.registerListener(sourceListener);
+
+    try {
+      startFlow("idempotent");
+      boolean timeout = !latch.await(TIMEOUT, MILLISECONDS);
+      assertThat(timeout, is(false));
+      assertThat(sourceNotifications.get(0).getActionName(), is("item dispatched to flow"));
```
marianogonzalez, line 192: constants all along

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -754,6 +797,29 @@ private void shutdownScheduler() {
...
+        return ITEM_REJECTED_WATERMARK;
+      case ALREADY_IN_PROCESS:
+        return ITEM_REJECTED_IDEMPOTENCY;
+      case SOURCE_STOPPING:
+        return ITEM_REJECTED_SOURCE_STOPPING;
+    }
+    return NO_ACTION_ID;
+  }
+
+  private void dispatchPollingSourceNotification(int action, String componentLocation, String pollId) {
```
marianogonzalez, line 817: there was a comment about deleting this method. Where did it go?

tests/functional/src/main/java/org/mule/functional/junit4/FunctionalTestCase.java
```diff
@@ -81,6 +82,9 @@ public FunctionalTestCase() {
   public SystemProperty muleDisableAsyncRetryPolicyOnSourcesProperty =
       new SystemProperty(DISABLE_ASYNC_RETRY_POLICY_ON_SOURCES, "true");
 
+  @Rule
+  public SystemProperty emitNotifications = new SystemProperty(EMIT_POLLING_SOURCE_NOTIFICATIONS, "true");
```
marianogonzalez, line 86: only enable this for the test case that requires it.

marianogonzalez (changes requested):

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/source/PollingSourceNotificationsTestCase.java
SebaElizalde: I need to use the item limit feature and I copied this code from https://github.com/mulesoft/mule/blob/44faefefef1d33316342f0122e64aa22f3680566/modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/source/PollingSourceLimitingTestCase.java#L68
Look into the PR that introduced the feature to know why it was needed:
https://github.com/mulesoft/mule/commit/83d46ead6204e4e0849d4f9fb62a089332d134ef

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
SebaElizalde: I reworked the method to avoid the points you raised, maybe to the refactor the comment went away? Anyways please look at it again

tests/functional/src/main/java/org/mule/functional/junit4/FunctionalTestCase.java
SebaElizalde: this is a problem because the field in the PollingSourceWrapper is static:
`  private static final boolean EMIT_NOTIFICATIONS = parseBoolean(getProperty(EMIT_POLLING_SOURCE_NOTIFICATIONS));
`
so when other test suites run (that don't have the @Rule set) they initialize the PollingSourceWrapper class and set the static field to false. I tried to access the PollingSourceWrapper class through reflection to change it before the tests, but was unable to do so because of class loading isolation. This was the way I found to make the tests possible.

marianogonzalez: then simply don't make it static and use a Rule in the tests that you want this enabled on

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/source/PollingSourceNotificationsTestCase.java
marianogonzalez: why do you need that feature to test this? You should be able to do this with a simple CountDownLatch

tests/functional/src/main/java/org/mule/functional/junit4/FunctionalTestCase.java
SebaElizalde: was not static, changed it because of this https://github.com/mulesoft/mule/pull/11666#discussion_r951466616 :)

marianogonzalez: I understand that, but if the consequence is having to to enable this for the entire suite then it's a no-go. Another alternative is to have it as static but mutable through a setter. You can then enable it in your test and disable it afterwards

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/poll/PollingSourceWrapper.java
```diff
@@ -97,13 +115,12 @@
...
   private static final String UPDATE_PROCESSED_LOCK = "OSClearing";
   private static final String INFLIGHT_IDS_OS_NAME_SUFFIX = "inflight-ids";
-  private static final String POLLING_SOURCE_EXECUTOR_KEY = "Polling source executor";
-  private static final String RUNNABLE_KEY = "Runnable";
 
   private final PollingSource<T, A> delegate;
   private final SchedulingStrategy scheduler;
   private final int maxItemsPerPoll;
   private final SystemExceptionHandler systemExceptionHandler;
+  private final boolean EMIT_NOTIFICATIONS = parseBoolean(getProperty(EMIT_POLLING_SOURCE_NOTIFICATIONS));
```
marianogonzalez, line 123: camelcase

marianogonzalez (changes requested):

marianogonzalez (approved):

PR #11668, opened 2022-08-10
W-11544941: Exception Notification being thrown even for retryable/handled exception.

lbarrios:


modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSourceAccessTokenExpiredExceptionHandlingTestCase.java
```diff
@@ -0,0 +1,133 @@
...
+
+@RunWith(Parameterized.class)
+public class ExtensionMessageSourceAccessTokenExpiredExceptionHandlingTestCase extends AbstractExtensionMessageSourceTestCase {
+
+  protected static final int TEST_TIMEOUT = 3000;
+  protected static final int TEST_POLL_DELAY = 1000;
+
+  @Parameterized.Parameters
+  public static List<Object[]> data() {
+    return asList(new Object[][] {{"true"}, {"false"}});
```
ndinu, line 48: Put a name to the parameterizations

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -115,6 +118,9 @@
 public class ExtensionMessageSource extends ExtensionComponent<SourceModel> implements MessageSource,
     ExceptionCallback<ConnectionException>, ParameterizedSource, ConfiguredComponent, LifecycleStateEnabled {
 
+  private static final Boolean SHOULD_HANDLE_ACCESS_TOKEN_EXPIRED_EXCEPTIONS_ON_SOURCES =
```
ndinu, line 121: Maybe it would be better to have its name saying IGNORE_ACCESS_TOKEN_EXPIRED_EXCEPTION_NOTIFICATIONS_ON_SOURCES , and if the property is present, ignore the handling.

ndinu (changes requested):
Also, before going with this change, explore the option of made the client filter unwanted notifications.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/ExtensionMessageSource.java
```diff
@@ -349,10 +355,14 @@ public void onException(ConnectionException exception) {
...
-
-    refreshTokenIfNecessary(getConfigurationInstance()
-        .flatMap(configurationInstance -> configurationInstance.getConnectionProvider()).orElse(null), exception);
+    boolean accessTokenExpired = refreshTokenIfNecessary(
+                                                         getConfigurationInstance()
+                                                             .flatMap(ConfigurationInstance::getConnectionProvider).orElse(null),
+                                                         exception);
+    boolean shouldHandleException = !accessTokenExpired || SHOULD_HANDLE_ACCESS_TOKEN_EXPIRED_EXCEPTIONS_ON_SOURCES;
+    if (shouldHandleException) {
+      muleContext.getExceptionListener().handleException(exception, getLocation());
```
marianogonzalez, line 364: this is the wrong place to place this if. This is not just preventing the notification from being fired, this is skipping the exception handler alltogether, which is doing way more things than simply firing the notification

marianogonzalez (changes requested):

PR #11682, opened 2022-08-18
W-11617156: Feature flag for disabling explicit GC call when disposing RegionClassLoader.

asanguinetti:


modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/classloader/DefaultArtifactClassLoaderResolver.java
```diff
@@ -74,13 +76,22 @@ public class DefaultArtifactClassLoaderResolver implements ArtifactClassLoaderRe
   private final ModuleRepository moduleRepository;
   private final NativeLibraryFinderFactory nativeLibraryFinderFactory;
   private final MuleDeployableArtifactClassLoader defaultDomainClassloader;
+  private final FeatureFlaggingService featureFlaggingService;
 
   public DefaultArtifactClassLoaderResolver(ArtifactClassLoader containerClassLoader,
                                             ModuleRepository moduleRepository,
                                             NativeLibraryFinderFactory nativeLibraryFinderFactory) {
+    this(containerClassLoader, moduleRepository, nativeLibraryFinderFactory, null);
```
asanguinetti, line 84: I don't like the fact that constructing this instance with a `null` `FeatureFlaggingService` is still allowed, that could become a headache down the line.
Changing this would require changing the [mule-framework api](https://github.com/mulesoft/mule/blob/255a790e2a565d71113c181b30dce2b3aff25aff/modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/classloader/ArtifactClassLoaderResolver.java#L58). That wouldn't be too much of a problem because it is still not GA, but I don't know if we want `mule-framework` users to have to be aware about feature flagging.
Any thoughts?

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/classloader/DefaultArtifactClassLoaderResolver.java
```diff
@@ -131,6 +142,26 @@ public MuleDeployableArtifactClassLoader createDomainClassLoader(DomainDescripto
...
+                                                    ArtifactDescriptor descriptor,
+                                                    ClassLoader parentClassloader,
+                                                    ClassLoaderLookupPolicy parentLookupPolicy) {
+    RegionClassLoader regionClassLoader;
+    if (featureFlaggingService != null) {
+      regionClassLoader = new FeatureFlaggingAwareRegionClassLoader(artifactId,
+                                                                    descriptor,
+                                                                    parentClassloader,
+                                                                    parentLookupPolicy,
+                                                                    featureFlaggingService);
```
asanguinetti, line 155: Considering the feature flagging service would not always be available for the RegionClassLoader, I preferred to create a sub-class (which will not work with a `null` FF service).
This way we can more easily track in which places we are using it with feature flagging capabilities and in which places we are using it without it.

modules/artifact/src/main/java/org/mule/runtime/module/artifact/internal/classloader/FeatureFlaggingAwareRegionClassLoader.java
```diff
@@ -0,0 +1,67 @@
...
+    if (!regionClassloaderFeaturesSet.getAndSet(true)) {
+      FeatureFlaggingRegistry featureFlaggingRegistry = FeatureFlaggingRegistry.getInstance();
+      featureFlaggingRegistry.registerFeatureFlag(DISABLE_EXPLICIT_GC_WHEN_DISPOSING_ARTIFACT,
+                                                  featureContext -> featureContext.getArtifactMinMuleVersion()
+                                                      .filter(muleVersion -> muleVersion
+                                                          .atLeast(DISABLE_EXPLICIT_GC_WHEN_DISPOSING_ARTIFACT
+                                                              .getEnabledByDefaultSince()))
+                                                      .isPresent());
+    }
+  }
```
asanguinetti, line 42: I stole this fragment from a similar snippet [here](https://github.com/mulesoft/mule/blob/255a790e2a565d71113c181b30dce2b3aff25aff/core/src/main/java/org/mule/runtime/core/internal/profiling/AbstractProfilingService.java#L142).

I couldn't find any ticket tracking this TODO, but it would be nice to know what are the plans. I'm not sure I understand why the registration needs to be manually done when all the metadata to do it automatically is available in the enum. Is it because some FF are being registered with a predicate different than the `minMuleVersion`?

asanguinetti (commented):
Leaving some comments for the reviewers.

Also note that this PR is missing tests. I'm still working on that

modules/artifact/src/main/java/org/mule/runtime/module/artifact/internal/classloader/FeatureFlaggingAwareRegionClassLoader.java
fsgonz: I'll create one. Yes, I think when ff flag was created, the idea is to have different predicates to be evaluated. So you have to register what predicate will activate the f (which is not always the mule version).

fsgonz: I'll create a ticket for the TODO.

fsgonz (approved):
I approve (taking into account anyway that the other option is being considered)

mbuchwald:
closing :( https://docs.google.com/document/d/1uSyR6GbKRMQD8knvSCRRGfzmHGCep0Vsv79mGym5HIA/edit#heading=h.q6c7zxyhept4

PR #11686, opened 2022-08-19
W-11617156: Feature flag for disabling explicit GC call when disposing RegionClassLoader.

asanguinetti:


core/src/main/java/org/mule/runtime/core/internal/context/DefaultMuleContext.java
```diff
@@ -1473,6 +1476,11 @@ private static void configureHonourErrorMappingsWhenPolicyAppliedOnOperation() {
                                                 minMuleVersion("4.5.0"));
   }
 
+  private static void registerFeatureOnMinMuleVersion(Feature feature) {
+    FeatureFlaggingRegistry featureFlaggingRegistry = FeatureFlaggingRegistry.getInstance();
+    featureFlaggingRegistry.registerFeatureFlag(feature, minMuleVersion(feature.getEnabledByDefaultSince()));
+  }
```
asanguinetti, line 1482: I think we should eventually refactor this class to use this function instead of all the many individual feature registration functions that are so repetitive.
Additionally, there is no reason to hard-code the `minMuleVersion` during registration because we actually have it already defined in the `Feature` itself as its `enabledByDefaultSince` property. This could lead to mismatch issues.
But I don't want to mix all those changes in this same PR.

modules/artifact/src/main/java/org/mule/runtime/module/artifact/internal/descriptor/FeatureFlaggingUtils.java
```diff
@@ -4,7 +4,7 @@
  * license, a copy of which has been included with this distribution in the
  * LICENSE.txt file.
  */
-package org.mule.runtime.module.deployment.impl.internal.config;
+package org.mule.runtime.module.artifact.internal.descriptor;
```
asanguinetti, line 7: I had to move it from `deployment-impl` to `artifact` because we now would need to use it in the later. The `deployment-impl` module already has the `artifact` module as a dependency.
This one still sounds like a reasonable place to have this util class, because it deals mostly with `ArtifactDescriptor` which is an `artifact` class. For this same reason, putting this in `core` is not possible.

modules/artifact/src/test/java/org/mule/runtime/module/artifact/internal/descriptor/ArtifactTestingFeatures.java
```diff
@@ -0,0 +1,79 @@
...
+
+
+/**
+ * Features meant to be used in {@link Feature} flag related tests.
+ * 
+ * @see FeatureFlaggingUtils
+ * @see org.mule.runtime.api.config.FeatureFlaggingService
+ * @since 4.5.0
+ */
+public enum ArtifactTestingFeatures implements Feature {
```
asanguinetti, line 25: This class has been copied from [DeploymentTestingFeature](https://github.com/mulesoft/mule/pull/11686/files#diff-79c4c2e18c754d872d9a9f12dc8cfce5eb8db436e7e7f198678f2944626ee990).
This is because when we moved the `FeatureFlaggingUtilsTestCase` to this package we also needed to bring these features. But some of them (not all) were still needed for the tests in `deployment-impl`, this is why it wasn't just a move.

modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/config/DeploymentTestingFeatures.java
```diff
@@ -24,18 +25,15 @@
  */
 public enum DeploymentTestingFeatures implements Feature {
 
-  ALWAYS_ON_FEATURE("Testing feature", "MULE-123", "4.4.0"), OVERRIDEABLE_FEATURE("Testing feature", "MULE-123", "4.4.0",
-      "overrideable.feature.override");
+  ALWAYS_ON_FEATURE("Testing feature", "MULE-123", "4.4.0");
```
asanguinetti, line 28: Removed the features that were not being used by tests in `deployment-impl`.

asanguinetti (commented):
Some comments for the reviewers.

mbuchwald:
Closing :( https://docs.google.com/document/d/1uSyR6GbKRMQD8knvSCRRGfzmHGCep0Vsv79mGym5HIA/edit#heading=h.q6c7zxyhept4

PR #11687, opened 2022-08-22
W-11573484: pass the correct memory management service

fsgonz:


rbourbonnavarro (dismissed):

eze210 (dismissed):

IvanAndresFritzler (approved):

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/MuleArtifactResourcesRegistry.java
```diff
@@ -297,7 +297,7 @@ private MuleArtifactResourcesRegistry(ArtifactClassLoader containerClassLoader,
                                                        artifactPluginDescriptorLoader,
                                                        licenseValidator,
                                                        runtimeLockFactory,
-                                                       memoryManagementService,
+                                                       this.memoryManagementService,
```
asanguinetti, line 300: You sure this one doesn't need to be `artifactMemoryManagementService`, right?

Just mentioning it because, `memoryManagementService` and `this.memoryManagementService` seem to be referring to the same object at this point. But maybe I'm confused.

asanguinetti: Well actually they are not the same if `memoryManagementService` is `null`. My bad.

fsgonz: The idea is that [here](https://github.com/mulesoft/mule/blob/master/modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/MuleArtifactResourcesRegistry.java#L218)

that I can pass a non default memory management service that is not default (for example when creating FakeMuleServer for testing gw in the same classloader). I am messing up the reference to the parameter and the one that is finally set. 

The artifactMemoryManagement service is used for injecting the memory management in artifacts and there is a memory management service at container level.

asanguinetti (approved):

fsgonz:
--validate

PR #11691, opened 2022-08-22
W-11300015: ComponentParameterization API must expose the different values it can handle

ndinu:


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ExtensionDefinitionParser.java
```diff
@@ -162,7 +162,7 @@ protected ExtensionDefinitionParser(Builder definitionBuilder, DslSyntaxResolver
     this.definitionBuilder = definitionBuilder;
     this.dslResolver = dslResolver;
     this.parsingContext = ctx;
-    this.valueResolverFactory = new ValueResolverFactory(dslResolver);
```
marianogonzalez, line 165: love to see the `DslSyntaxResolver` die, but I don't get why it's not needed anymore? Where's the code that uses the declarer instead?

marianogonzalez (changes requested):

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ExtensionDefinitionParser.java
ndinu: This is where it was used:

https://github.com/mulesoft/mule/blob/support/4.3.x/modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/resolver/ValueResolverFactoryTypeVisitor.java#L96

It used it to get a DslElementSyntax to use a ValueResolverParsingDelegate. But the ValueResolverParsingDelegate was not needed for its processing. 

The code removed from the [ValueResolverFactoryTypeVisitor.java](https://github.com/mulesoft/mule/pull/11691/files#diff-61f0c1c4f994f6e7c26f8261031e7458641b66dacc977450adda8da8d22d1646) is the only usage it remained of it, and the DslElementSyntax was only checked for its existence.

marianogonzalez (approved):

PR #11692, opened 2022-08-22
W-11495137: Mule Operations shouldn't declare errors caught by an on-error-continue

eze210:


modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/utils/AggregatedErrorsCharacteristic.java
```diff
@@ -0,0 +1,296 @@
...
+        } else if (container.getIdentifier().equals(ERROR_HANDLER_IDENTIFIER)) {
+          --containerIndex;
+        }
+        --containerIndex;
+      }
+      return suppressedErrors;
+    }
+
+    private void addSuppressedErrors(ComponentAst tryScopeAst, List<ErrorModelParserMatcher> suppressedErrors) {
+      Optional<ComponentAst> errorHandler = tryScopeAst.directChildrenStreamByIdentifier(null, ERROR_HANDLER).findFirst();
```
asanguinetti, line 127: Shouldn't we set `CORE_PREFIX` for the namespace matcher?
According to the Javadoc, `null` acts as a wildcard, meaning that will match anything.
What if there ever exists some component (from an extension most likely) named `<some-extension:error-handler>`? I guess we don't want to take any action on that one, because it wouldn't be related with the `try` scope.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/utils/AggregatedErrorsCharacteristic.java
```diff
@@ -0,0 +1,296 @@
...
+      if (!errorHandler.isPresent()) {
+        return;
+      }
+
+      Optional<ComponentAst> onErrorContinue =
+          errorHandler.get().directChildrenStreamByIdentifier(null, ON_ERROR_CONTINUE).findFirst();
```
asanguinetti, line 133: Same here

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/utils/AggregatedErrorsCharacteristic.java
```diff
@@ -0,0 +1,296 @@
...
+      if (!onErrorContinue.isPresent()) {
+        return;
+      }
+
+      String typeAsString =
+          (String) onErrorContinue.get().getParameter(DEFAULT_GROUP_NAME, ERROR_TYPE_PARAM).getValue().getValue().orElse(null);
```
asanguinetti, line 139: I don't understand how this doesn't blow up when the `ERROR_TYPE_PARAM` is not present like in your [test](https://github.com/mulesoft/mule/pull/11692/files#diff-ee6a64fa8e3c21dfaf9faac6e8e5ca0a76a59679b9d6eb944f900f3c6c8751efR45).

Initially I would think that `getParameter` would return `null` (as explained in the Javadoc). And immediately, the first `getValue` should throw an NPE.

Do you know what's going on?

modules/extensions-mule-support/src/test/resources/mule-error-handling-with-try-operations-config.xml
```diff
@@ -0,0 +1,94 @@
...
+
+    <operation:def name="operationSilencingOneSpecificErrorAndRaisingAnother" >
+        <operation:output>
+            <operation:payload-type type="void"/>
+        </operation:output>
+        <operation:body>
+            <try>
+                <heisenberg:cure-cancer />
+                <error-handler>
+                    <on-error-continue type="HEISENBERG:HEALTH"> <!-- catching HEALTH but not OAUTH2 -->
```
asanguinetti, line 25: I think it is not possible to combine both, but should we check just in case what happens if there is a `when` and a `type` parameter?
Or actually, in the case of no `type`, should we be careful if we have something like `when="#[false]"`. That would certainly complicate things.

And now I'm remembering it is also possible to have references to global error handlers... *cries*

asanguinetti (commented):
Submitting partial review feedback. I still haven't finished reviewing all of it.

modules/extensions-mule-support/src/test/resources/mule-error-handling-with-try-operations-config.xml
eze210: `when`: Maybe if it has a when we shouldn't consider nothing as suppressed, because in parse time we don't know the result...
`global error handler`: You're right

Do you agree with creating 2 new work items?

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/utils/AggregatedErrorsCharacteristic.java
asanguinetti: that's tricky but is probably something to be clarified on the Javadoc of `getParameter`, which I wouldn't do as part of this PR.

modules/extensions-mule-support/src/test/resources/mule-error-handling-with-try-operations-config.xml
asanguinetti: that would work for me

eze210: Created W-11640594 to address these cases and some others that I found

eze210: Also added a TODO in the corresponding class

asanguinetti (approved):

PR #11696, opened 2022-08-23
W-10876269: Add exceptions records to spans and error data to on error handlers

IvanAndresFritzler:


core/src/main/java/org/mule/runtime/core/internal/profiling/tracing/event/span/InternalSpan.java
```diff
@@ -33,6 +35,12 @@ public interface InternalSpan extends Span {
    */
   void end();
 
+
+  /**
+   * Ends the span, previously recording the provided error.
```
fsgonz, line 40: Why we should end the span when an error is registered? I would think we should

fsgonz: I think this is only a jdoc comment.

core/src/main/java/org/mule/runtime/core/internal/profiling/tracing/export/OpenTelemetrySpanExporter.java
```diff
@@ -67,17 +80,31 @@ public <T> T visit(InternalSpanExporterVisitor<T> internalSpanExporterVisitor) {
...
+  }
+
+  private void recordSpanException(SpanError spanError) {
+    Attributes errorAttributes = Attributes.of(
+                                               EXCEPTION_TYPE_KEY, spanError.getError().getErrorType().toString(),
+                                               EXCEPTION_MESSAGE_KEY, spanError.getError().getDescription(),
+                                               EXCEPTION_STACK_TRACE_KEY,
+                                               InternalSpanError.getInternalSpanError(spanError).getErrorStacktrace().toString(),
+                                               EXCEPTION_ESCAPED_KEY, spanError.isEscapingSpan());
+    openTelemetrySpan.addEvent("exception", errorAttributes);
```
fsgonz, line 94: add this to the constants too.
Add a TODO to this ticket: W-11610439: Track if the semconv dependency (alpha) should be added.
To remember to verify if it convenient to add these constants from the semconv project

core/src/main/java/org/mule/runtime/core/internal/profiling/tracing/event/span/InternalSpan.java
IvanAndresFritzler: It is not ending the Span, it was just the JDoc.

fsgonz (approved):

PR #11697, opened 2022-08-23
W-11425680: Implement ForwardCompatibilityHelper.getDistributedTraceContextManager()

SebaElizalde:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/runtime/compatibility/DefaultForwardCompatibilityHelper.java
```diff
@@ -6,7 +6,41 @@
...
+import org.mule.runtime.module.extension.internal.runtime.source.trace.DefaultDistributedSourceTraceContext;
+import org.mule.sdk.api.runtime.source.DistributedTraceContextManager;
+import org.mule.sdk.compatibility.api.utils.ForwardCompatibilityHelper;
+
+import javax.inject.Inject;
 
 public class DefaultForwardCompatibilityHelper implements ForwardCompatibilityHelper {
+
+  @Inject
+  private InternalProfilingService profilingService;
```
fsgonz, line 23: DefaultForwardCompatibilityHelper is loaded as a service so it is not managed by the registry and nothing will be injected. For this to be injected, you can inject the dependencies programmatically for example muleContext.getInjector().inject(helper) or you can define define the helper as a bean (I don't know if this is what is intended)
Possibly as this is a helper you can pass the coreEventTracer to the method as a parameter and the method can be static (so nothing is injected through @Inject). The places where you invoke those methods should be in a place that is managed by the mule runtime registry so that you can pass the profiling server or the coreEvenTracer that you retrieved from that injected profiling server.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/runtime/compatibility/DefaultForwardCompatibilityHelper.java
```diff
@@ -6,7 +6,41 @@
...
+          new DistributedTraceContextManagerResolver(profilingService.getCoreEventTracer());
+      return argumentResolver.resolve(event);
+    } else {
+      return new DefaultDistributedSourceTraceContext();
+    }
+  }
+
+  public DistributedTraceContextManager getDistributedTraceContextManager(org.mule.sdk.api.runtime.parameter.CorrelationInfo correlationInfo) {
+    if (correlationInfo instanceof ImmutableCorrelationInfo) {
+      CoreEvent event = ((ImmutableCorrelationInfo) correlationInfo).getEvent();
```
fsgonz, line 38: As this is implemented, the only way for the resolvers that have something injected from mule context container is to pass it through the correlationInfo. Not sure if this is the expected approach.

fsgonz (dismissed):

SebaElizalde:
--validate

SebaElizalde:
--validate

marianogonzalez (approved):

SebaElizalde:
--validate

SebaElizalde:
--validate

PR #11699, opened 2022-08-24
W-11644164: Adapt metadata SDK api to mule api

Bardammu:
Use mule-api


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/utils/JavaMetadataTypeResolverUtils.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class JavaMetadataTypeResolverUtils {
+
+  public static boolean isNullResolver(Class<?> typeResolverClass) {
+    return NullMetadataResolver.class.isAssignableFrom(typeResolverClass) ||
+        org.mule.runtime.extension.api.metadata.NullMetadataResolver.class.isAssignableFrom(typeResolverClass);
+  }
+
+  public static Class<?> getEnclosingClass(NamedTypeResolver namedTypeResolver) {
+    if (namedTypeResolver instanceof MuleOutputTypeResolverAdapter) {
```
ndinu, line 24: Do we always only want the class of the delegate? Maybe we can make them implement an interface that has a `getDelegateResolverClass` of similar, and just check for that one.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/DefaultMetadataScopeAdapter.java
```diff
@@ -111,13 +112,15 @@ private String getCategoryName(JavaOutputResolverModelParser javaOutputResolverM
                                  JavaAttributesResolverModelParser javaAttributesResolverModelParser,
                                  List<JavaInputResolverModelParser> javaInputResolverModelParsers) {
 
-    NamedTypeResolver namedTypeResolver = javaOutputResolverModelParser.getOutputResolver();
-    if (!(namedTypeResolver instanceof org.mule.sdk.api.metadata.NullMetadataResolver)) {
+    org.mule.runtime.api.metadata.resolving.NamedTypeResolver namedTypeResolver =
```
ndinu, line 115: I think we can avoid the FQCN here

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/DefaultMetadataScopeAdapter.java
```diff
@@ -111,13 +112,15 @@ private String getCategoryName(JavaOutputResolverModelParser javaOutputResolverM
...
+        javaOutputResolverModelParser.getOutputResolver();
+    if (javaOutputResolverModelParser.hasOutputResolver()) {
       return namedTypeResolver.getCategoryName();
     }
 
-    NamedTypeResolver namedTypeAttributesResolver = javaAttributesResolverModelParser.getAttributesResolver();
-    if (!(namedTypeAttributesResolver instanceof org.mule.sdk.api.metadata.NullMetadataResolver)) {
+    org.mule.runtime.api.metadata.resolving.NamedTypeResolver namedTypeAttributesResolver =
```
ndinu, line 121: same here

ndinu (approved):

PR #11709, opened 2022-08-26
W-11643926: An error affecting the tracing should not affect the application execution

fsgonz:


core-tests/src/test/java/org/mule/runtime/core/internal/profiling/consumer/tracing/DefaultCoreEventTracerTestCase.java
```diff
@@ -129,6 +132,39 @@ public void endCurrentExecutionSpan() {
     verify(distributedTraceContext).endCurrentContextSpan();
   }
 
+  @Test
```
IvanAndresFritzler, line 135: Can we parametrize the tests you added so we can check that the system property that propagates the errors also works?

IvanAndresFritzler (changes requested):

IvanAndresFritzler (approved):

IvanAndresFritzler (approved):

PR #11721, opened 2022-09-01
W-11500514: A.3.2 Implement the production of units for Batch according to the spec A.3.1.

sofiamorseletto:


core/src/main/java/org/mule/runtime/core/internal/profiling/tracing/event/span/BatchChainCustomizationInfo.java
```diff
@@ -0,0 +1,44 @@
...
+import org.mule.runtime.api.component.location.ComponentLocation;
+import org.mule.runtime.core.api.event.CoreEvent;
+import org.mule.runtime.core.privileged.profiling.tracing.ChildSpanCustomizationInfo;
+
+/**
+ * Batch Chain Customization info. Includes batch-step-record, batch-aggregator and batch-on-complete spans.
+ *
+ * @since 4.5.0
+ */
+public class BatchChainCustomizationInfo extends AbstractDefaultAttributesResolvingSpanCustomizationInfo {
```
fsgonz, line 20: Pass this to mule-ee. as this corresponds to batch. Maybe I was the one that set it here in the working branch. Maybe it can be generalized so that this is not related to batch. Possibly set a descriptive name and it can be reuse for other components. It simply is basing everything in the location and the name.

core/src/main/java/org/mule/runtime/core/privileged/processor/MessageProcessors.java
```diff
@@ -99,6 +104,35 @@ public static MessageProcessorChain newChain(Optional<ProcessingStrategy> proces
     }
   }
 
+  public static MessageProcessorChain newChain(Optional<ProcessingStrategy> processingStrategy, List<Processor> processors,
+                                               String name) {
+    if (processors.size() == 1 && processors.get(0) instanceof MessageProcessorChain) {
+      return (MessageProcessorChain) processors.get(0);
+    } else {
+      return buildNewChainWithListOfProcessors(processingStrategy, processors, new SpanCustomizationInfo() {
```
fsgonz, line 112: Verify if this can be set in a different class that implements SpanCustomizationInfo.

fsgonz (changes requested):

fsgonz (approved):

fsgonz (approved):

PR #11725, opened 2022-09-06
W-11468826: Mule SDK ExtensionModelDeclarer.

asanguinetti:
- Also adding XSD definition.
- Also adding DslNamespaceInfoProvider.
- Also adding ExtensionModelProvider.
- Also adding RuntimeExtensionModelProvider.


modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/api/extension/MuleSdkExtensionExtensionModelDeclarer.java
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * An {@link ExtensionDeclarer} for Mule SDK Extensions
+ *
+ * @since 4.5
+ */
+public class MuleSdkExtensionExtensionModelDeclarer {
+
+  public ExtensionDeclarer declareExtensionModel() {
+    final BaseTypeBuilder typeBuilder = BaseTypeBuilder.create(JavaTypeLoader.JAVA);
```
eze210, line 40: static imports

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/api/extension/MuleSdkExtensionRuntimeExtensionModelProvider.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.module.extension.mule.api.extension;
+
```
eze210, line 8: blank

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/api/extension/MuleSdkExtensionRuntimeExtensionModelProvider.java
```diff
@@ -0,0 +1,30 @@
...
+ * @since 4.5
+ */
+@NoInstantiate
+public class MuleSdkExtensionRuntimeExtensionModelProvider implements RuntimeExtensionModelProvider {
+
+  @Override
+  public ExtensionModel createExtensionModel() {
+    return getExtensionModel();
+  }
+
```
eze210, line 29: blank

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/dsl/processor/xml/MuleSdkExtensionDslNamespaceInfoProvider.java
```diff
@@ -0,0 +1,49 @@
...
+  public static final String MULE_EXTENSION_DSL_NAMESPACE = "extension";
+  public static final String MULE_EXTENSION_DSL_NAMESPACE_URI = format(DEFAULT_NAMESPACE_URI_MASK, "mule-extension");
+  public static final String MULE_EXTENSION_DSL_XSD_FILE_NAME = "mule-extension.xsd";
+  public static final String MULE_EXTENSION_DSL_SCHEMA_LOCATION =
+      buildSchemaLocation(MULE_EXTENSION_DSL_NAMESPACE, MULE_EXTENSION_DSL_XSD_FILE_NAME);
+
+  @Override
+  public Collection<XmlNamespaceInfo> getXmlNamespacesInfo() {
+
+    return asList(new XmlNamespaceInfo() {
```
eze210, line 36: Maybe singletonList since it's only one element (and remove blank)

modules/extensions-mule-support/src/test/java/org/mule/runtime/module/extension/mule/api/extension/MuleExtensionExtensionModelTestCase.java
```diff
@@ -0,0 +1,151 @@
...
+  @Test
+  public void parametersAreParsedAndHaveTheRightValueWhenMinimallyDefined() {
+    ArtifactAst extensionAst = getArtifactAst("extensions/extension-minimally-parameterized.xml");
+
+    ComponentAst extensionComponentAst = getRootComponent(extensionAst);
+
+    Map<String, Matcher<Object>> expectedParameters = new LinkedHashMap<>();
+    expectedParameters.put("name", is("Minimally Parameterized Extension"));
+    expectedParameters.put("category", is(COMMUNITY.name()));
+    expectedParameters.put("vendor", is(Extension.MULESOFT));
```
eze210, line 109: static import

modules/extensions-mule-support/src/test/java/org/mule/runtime/module/extension/mule/api/extension/MuleExtensionExtensionModelTestCase.java
```diff
@@ -0,0 +1,151 @@
...
+    assertThat(extensionIdentifier.getName(), is("extension"));
+    assertThat(extensionIdentifier.getNamespace(), is("extension"));
+    assertThat(extensionIdentifier.getNamespaceUri(), is("http://www.mulesoft.org/schema/mule/mule-extension"));
+
+    return onlyTopLevelComponent;
+  }
+
+  private void assertThatParametersAreExactly(ComponentAst componentAst, Map<String, Matcher<Object>> expectedParameters) {
+    assertThat(componentAst.getParameters(), hasSize(expectedParameters.size()));
+    for (Map.Entry<String, Matcher<Object>> entry : expectedParameters.entrySet()) {
```
eze210, line 143: statically import Entry

eze210 (changes requested):

eze210 (approved):

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/api/extension/MuleSdkExtensionExtensionModelProvider.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+@NoInstantiate
+public class MuleSdkExtensionExtensionModelProvider {
+
+  private static final LazyValue<ExtensionModel> EXTENSION_MODEL = new LazyValue<>(() -> new ExtensionModelFactory()
+      .create(contextFor(new MuleSdkExtensionExtensionModelDeclarer().declareExtensionModel())));
+
+  private static ExtensionLoadingContext contextFor(ExtensionDeclarer declarer) {
+    ExtensionLoadingContext context = new DefaultExtensionLoadingContext(declarer, loadingRequest());
+    context.addParameter(FORCE_EXTENSION_VALIDATION_PROPERTY_NAME, true);
```
marianogonzalez, line 34: why?

marianogonzalez (changes requested):

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/api/extension/MuleSdkExtensionExtensionModelProvider.java
asanguinetti: Honestly I forgot to ask about this one.
I took it trying to mimick what we did for the [OperationDslExtensionModelProvider](https://github.com/mulesoft/mule/blob/9c92460a7a3bb39d6eccbc42a6bb65d3662d8785/core/src/main/java/org/mule/runtime/core/api/extension/OperationDslExtensionModelProvider.java#L22), which in turn delegates to [this](https://github.com/mulesoft/mule/blob/9c92460a7a3bb39d6eccbc42a6bb65d3662d8785/runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelProvider.java#L94).
I assumed this one should be similar, but I had to check what that property does and I forgot. Let me see if I can find out quickly.

asanguinetti: I was tracking it and seems it is only used for deciding if [this code](https://github.com/mulesoft/mule-extensions-api/blob/84712b1dbc09ed0df586fc6f0dd43088a7985b82/mule-extensions-api/src/main/java/org/mule/runtime/extension/internal/loader/ExtensionModelFactory.java#L225) should execute.
Which is basically running the validators defined [here](https://github.com/mulesoft/mule-extensions-api/blob/84712b1dbc09ed0df586fc6f0dd43088a7985b82/mule-extensions-api/src/main/java/org/mule/runtime/extension/internal/loader/ExtensionModelFactory.java#L190) (and also any custom ones which for this case there are none) against the ExtensionModel that is being created.
Apparently, the default is to not run the validations.
It was added to the `MuleExtensionModelProvider` as part of [this change](https://github.com/mulesoft/mule/commit/1723401c53a5400b7547fc8da0e79faf06779cba#diff-92c3b537c785fb24eeb4f03e894071b91c10fce569a81072629c0440fcca8dbfR94) which was part of the PoC for the Application ExtensionModel.
I suppose the `ExtensionModel` for the `extension` namespace should be affected by the same criteria we used for the `operation` namespace.
Intuitively it makes sense to validate even our own hand-crafted models, just to make sure they are consistent with all the rules. If we are concerned about performance we could validate them as part of the `mvn validate` goal of the corresponding Maven module instead of doing it in runtime during deployment. But that change would be out of scope for this PR.

marianogonzalez (approved):
conditional approve

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/api/extension/MuleSdkExtensionExtensionModelProvider.java
asanguinetti: After following up in Slack, we agreed that this property should be removed from all places in which we are creating our own extension models.
The validation should already be happening during testing because of [this](https://github.com/mulesoft/mule-extensions-api/blob/84712b1dbc09ed0df586fc6f0dd43088a7985b82/mule-extensions-api/src/main/java/org/mule/runtime/extension/internal/loader/ExtensionModelFactory.java#L207). We should not do it in runtime then.
I created another work item ([W-11719146](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000015uqwdYAA/view)) to track the removal of the property and verifying that the validations are run when testing. Hence, we will be leaving it out of the scope of this particular PR.

PR #11733, opened 2022-09-07
W-11719146: Remove forceExtensionValidation property when creating our own ExtensionModels

asanguinetti:


asanguinetti:
Some comments for the reviewers:

There are another two remaining usages of this property.

One is in an annotation processor [here](https://github.com/mulesoft/mule/blob/0c98ddc72502c19b4ef50c88a6d8c9f0f0487601/modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/resources/BaseExtensionResourcesGeneratorAnnotationProcessor.java#L92). But that one should remain I believe. That code is for the extensions developed using the SDKs.
If we wanted, we could change it to set the property on the `ExtensionLoadingContext` instead of setting it as a system property. But since this is an annotation processor, it shouldn't cause any problem as it is. Any thoughts?

The other one is in the packager [here](https://github.com/mulesoft/mule-extensions-maven-plugin/blob/d0b7c0d887df53ec5f2b166d0964de3abb2365fd/mule-extensions-maven-plugin/src/main/java/org/mule/extension/maven/ExtensionModelMojo.java#L97). Again, this one should remain because it is used to verify extensions developed using the SDKs.

Regarding the validations during testing: they will indeed be run for any TestCase that extends from [AbstractMuleTestCase](https://github.com/mulesoft/mule/blob/0c98ddc72502c19b4ef50c88a6d8c9f0f0487601/tests/unit/src/main/java/org/mule/tck/junit4/AbstractMuleTestCase.java#L97) (which should actually involve all TestCases according to our [coding conventions](https://docs.google.com/document/d/1N2Axq3ABJjYmj9BjfYQkvJfF0odHQnlb8MAdVQydOWI/edit#heading=h.ujxk1qbe2hxj)). Also any test extending from [BaseExtensionModelFactoryTestCase](https://github.com/mulesoft/mule-extensions-api/blob/84712b1dbc09ed0df586fc6f0dd43088a7985b82/mule-extensions-api/src/test/java/org/mule/runtime/extension/internal/loader/BaseExtensionModelFactoryTestCase.java#L46) will do the same.
Even though, in practice, not all our TestCases extend from one of these, it is enough with having at least one test doing so, provided it is executed as part of our CI/CD pipelines. We currently have many.

Just in case, it was also verified by debugging, that the result of [_isTestingMode_](https://github.com/mulesoft/mule-api/blob/99bac571e0fbfc13c162db832cdc8037f807746d/src/main/java/org/mule/runtime/api/util/MuleSystemProperties.java#L609) was true for some of our unit tests.

PR #11735, opened 2022-09-07
W-11719589: MuleArtifactClassLoader leak when the Object class is not…

alopezbizon:
… remove from the classValueMap of Scala


eze210 (changes requested):
For the time being I wouldn't merge this PR since it hides the root cause of the leak, which is the Scala bug. We could include this mitigation in the future, but not before upgrading Scala

PR #11743, opened 2022-09-09
W-11208135: AST Validation forbidding expressions for parameter defaults on reusable operations

eze210:


eze210:
--validate

eze210:
--validate

mbuchwald (approved):

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/OperationParameterDefaultValueDoesntSupportExpressions.java
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @since 4.5
+ */
+public class OperationParameterDefaultValueDoesntSupportExpressions implements Validation {
+
+  private static final String DEFAULT_EXPRESSION_PREFIX = "#[";
+  private static final String DEFAULT_EXPRESSION_SUFFIX = "]";
+
+  private static final String OPERATION = "operation";
+  private static final String OPTIONAL_PARAMETER = "optional-parameter";
```
asanguinetti, line 39: Shouldn't we take this constants from a common place?
I know it's very unlikely we will ever change them because we would be breaking backwards, but still, it would make it easier to track its usages if we take them from a common place.
Maybe the `MuleOperationExtensionModelDeclarer` could expose them. Any thoughts?

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/OperationParameterDefaultValueDoesntSupportExpressions.java
```diff
@@ -0,0 +1,93 @@
...
+      return validationFailed(component);
+    } else {
+      return validationOk();
+    }
+  }
+
+  private Optional<ValidationResultItem> validationFailed(ComponentAst component) {
+    return component.getParameter(DEFAULT_GROUP_NAME, "name").getValue().getValue()
+        .map(name -> create(component, this,
+                            format("An expression was given for 'defaultValue' of the optional parameter '%s'", name)));
```
asanguinetti, line 77: The fact that this validator will return Optional.empty() if there is no "name" on the "optional-parameter" worries me a little bit.
Just a little bit, because I know the name is mandatory from the XSD validations already, but...
Do you think we should handle it just in case?

asanguinetti (changes requested):
The request for change is only for the isExpression part. The other ones are just suggestions.

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/OperationParameterDefaultValueDoesntSupportExpressions.java
eze210: Ok, but I would like it more to have it at `OperationDslNamespaceInfoProvider`. Do you agree?

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/OperationParameterDefaultValueDoesntSupportExpressions.java
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  private static boolean isExpression(Optional<String> defaultValue) {
+    if (!defaultValue.isPresent()) {
+      return false;
+    }
+
+    String stringValue = defaultValue.get();
+    return stringValue.startsWith(DEFAULT_EXPRESSION_PREFIX)
+        && stringValue.endsWith(DEFAULT_EXPRESSION_SUFFIX);
```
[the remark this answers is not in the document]
eze210: 👍

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/OperationParameterDefaultValueDoesntSupportExpressions.java
asanguinetti: Not entirely sure if that's the right place, because it is OperationDsl**Namespace**InfoProvider. I think those providers are for XML-specific stuff.

I still think that these constants belong to the declaration of the ExtensionModel. They could be then consumed from the parsers, validators, etc.

We could involve someone from the SDK team for their opinion.

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/OperationParameterDefaultValueDoesntSupportExpressions.java
asanguinetti: Yeah, either that or create a `ValidationResultItem` specifying the issue without stating the name...
I don't know what happens if an `IllegalStateException` is thrown from those methods.
Normally I would say whatever is best for the developer to identify the issue, but in this case, if this happens it means there is an error in our end when wiring up the validations (because a non-existent name should trigger earlier). So it would be us who need to catch it on a test and understand what happened.

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/OperationParameterDefaultValueDoesntSupportExpressions.java
eze210: Added to the declarer. Also made a little refactor there to expose some other constants

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/OperationParameterDefaultValueDoesntSupportExpressions.java
eze210: Added the IllegalStateException

PR #11744, opened 2022-09-09
W-11733064: DataWeaveExpressionLanguageAdaptor spring bean is not being disposed in the runtime

fsgonz:
This is a way to guarantee that singleton created by the FactoryBEan is disposed.

`The container is only responsible for managing the lifecycle of the FactoryBean instance, not the lifecycle of the objects created by the FactoryBean. Therefore, a destroy method on an exposed bean object (such as [Closeable.close()](https://docs.oracle.com/javase/8/docs/api/java/io/Closeable.html?is-external=true#close--)) will not be called automatically. Instead, a FactoryBean should implement [DisposableBean](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/DisposableBean.html) and delegate any such close call to the underlying object.`

[Source](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/FactoryBean.html) and tested.


fsgonz:
--validate

fsgonz:
--validate

modules/spring-config/src/main/java/org/mule/runtime/config/internal/el/DataWeaveExtendedExpressionLanguageAdaptorFactoryBean.java
```diff
@@ -15,9 +17,15 @@
 
 import javax.inject.Inject;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.DisposableBean;
 import org.springframework.beans.factory.FactoryBean;
 
-public class DataWeaveExtendedExpressionLanguageAdaptorFactoryBean implements FactoryBean<ExtendedExpressionLanguageAdaptor> {
+public class DataWeaveExtendedExpressionLanguageAdaptorFactoryBean implements FactoryBean<ExtendedExpressionLanguageAdaptor>,
```
marianogonzalez, line 25: This solution relies heavily in assuming that a Spring registry which is not the case in Mule Framework. I think that the actual issue is that the DataWeaveExpressionLanguageAdaptor class does implement `Disposable` and lifecycle should be applied to it regardless of how it got there. Do you have more details of how and why this happens? The GUS issue doesn't provide many details and if this also happens in MuleFramework this solution won't cut it.

modules/spring-config/src/test/java/org/mule/runtime/config/internal/el/DataWeaveExtendedExpressionLanguageAdaptorFactoryBeanTestCase.java
```diff
@@ -0,0 +1,63 @@
...
+import io.qameta.allure.Issue;
+import org.junit.Test;
+import org.springframework.beans.factory.DisposableBean;
+
+@Issue("W-10672687")
+public class DataWeaveExtendedExpressionLanguageAdaptorFactoryBeanTestCase {
+
+  @Test
+  public void whenDataWeaveExtendedExpressionLanguageAdaptorFactoryBeanIsDistroyedTheAdapatorIsDisposed() throws Exception {
+    TestDataWeaveExtendedExpressionLanguageAdaptorFactoryBean factoryBean =
```
marianogonzalez, line 29: this is testing the implementation rather than the actual behavior. Can't we instead hace a test that disposes a MuleContext and verifies that the adaptor has been disposed?

marianogonzalez (changes requested):

modules/spring-config/src/main/java/org/mule/runtime/config/internal/el/DataWeaveExtendedExpressionLanguageAdaptorFactoryBean.java
fsgonz: FactoryBean is a spring class (so in that sense at this level it is assuming it is a Spring Registry otherwise) and DataWeaveExpressionLanguageAdaptor is Disposable.

fsgonz: The gus ticket has a parent which is related to munit failing tests.

modules/spring-config/src/test/java/org/mule/runtime/config/internal/el/DataWeaveExtendedExpressionLanguageAdaptorFactoryBeanTestCase.java
fsgonz: It tests that when the destroy method of the factory is invoked it also disposes the singleton bean created which is the way to do achieve the disposal of the singleton according to spring java docs.
I'll try anyway to create an integration test.

fsgonz:
I'll close and open another PR because the root cause is a problem with lazy init.

PR #11746, opened 2022-09-12
W-11640594: Contemplate multiple on-error-continues and ignore handlers with when parameter for reusable operation models

eze210:


eze210:
--validate

mbuchwald (approved):

PR #11770, opened 2022-09-19
W-11779154: Reuse error handling logic in reusability operations

eze210:


modules/extensions-mule-support/src/test/java/org/mule/runtime/module/extension/mule/internal/loader/ErrorHandlingExtensionModelTestCase.java
```diff
@@ -66,7 +66,7 @@ public static void setupExpectedErrors() {
     expectedErrors.put("operationRaisingUniqueErrorAndCatchingIt", emptyList());
 
     expectedErrors.put("operationWithMultipleOnErrorContinues", emptyList());
-    expectedErrors.put("operationCatchingAllButWithWhen", asList("HEISENBERG:OAUTH2", "HEISENBERG:OAUTH2"));
+    expectedErrors.put("operationCatchingAllButWithWhen", asList("HEISENBERG:HEALTH", "HEISENBERG:OAUTH2"));
```
asanguinetti, line 69: This one is a little bit disturbing.

I realized the problem was actually [here](https://github.com/mulesoft/mule/blob/e2b230a329a7e41be5e408ce74667dd9dd399761/modules/extensions-mule-support/src/test/java/org/mule/runtime/module/extension/mule/internal/loader/ErrorHandlingExtensionModelTestCase.java#L109), right?

We are just checking that each item in the list of expected is present in the list of actual (and that the sizes of both lists match). This doesn't account for having repeated items.

However, the errors [are modeled as a `Set` in the `ExtensionModel`](https://github.com/mulesoft/mule-api/blob/5e04cfe654e1d28918ea832a657a27f97a88b8e7/src/main/java/org/mule/runtime/api/meta/model/ExtensionModel.java#L199). Couldn't we compare `Set` against `Set` in a similar fashion? That way we will make sure we catch any subtle duplication mistake like this one.

In any case, I'm not sure if that should be fixed in this PR because it is completely unrelated. If we ever decide we need to revert this PR we will be reverting that part as well.

asanguinetti (changes requested):

asanguinetti (approved):

PR #11774, opened 2022-09-20
W-11785114: Shared library takes precedence over additional plugin dependency

rbourbonnavarro:


rbourbonnavarro:
--validate

elrodro83 (changes requested):
add test case

elrodro83 (approved):

PR #11776, opened 2022-09-20
W-11465452: Capability to load Mule SDK extensions from resource file.

asanguinetti:
The goal of this PR is to be able to load the `ExtensionModel` from a dedicated artifact (a.k.a. Mule SDK Extension XML).

Our current implementation of [MuleSdkExtensionModelParserFactory](https://github.com/mulesoft/mule/blob/98ec323a74acadb54ba24a24ef1844e9017bc7a5/modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkExtensionModelParserFactory.java#L30) relies on having some properties available in the `ExtensionLoadingContext` such as the `ArtifactAst`, the `TypeLoader` and the extension's name.

This works perfectly well when creating a parser in the context of [generating the Extension Model of an application](https://github.com/mulesoft/mule/blob/a99200a5d633bed1263d203343030667333140ce/modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/ast/ArtifactAstUtils.java#L126) with reusable assets.

However, it does not work in the context of the [discovery of Plugins Extension Models](https://github.com/mulesoft/mule-extensions-maven-plugin/blob/d0b7c0d887df53ec5f2b166d0964de3abb2365fd/mule-extensions-maven-plugin/src/main/java/org/mule/extension/maven/loader/MavenProjectExtensionModelLoader.java#L142), where the properties for the `ExtensionLoadingContext` [are extracted from the attributes](https://github.com/mulesoft/mule/blob/dd27aba2fc742687a409d14c648028fe9b3adc21/modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/extension/discovery/RepositoryLookupExtensionModelGenerator.java#L90) section of the `mule-artifact.json` file.

In this PR we are giving the parser factory the capability to receive the resource file location in the loading request context and generate the AST from there. One important thing to note here is that this AST needs to be generated with a similar two pass strategy that we use for applications in order to support operations that reference other operations from the same artifact using the namespace `this`.

Finally, the `ApplicationTypeLoader` can be generated inside the `ParserFactory` because it only requires the `ExtensionModels` in context.

There are some TODO notes for things that could be simplified if we were to change the DSL for the Mule SDK Extensions.


modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/ast/ArtifactAstUtils.java
```diff
@@ -107,43 +146,106 @@ public static ArtifactAst parseAndBuildAppExtensionModel(String[] configResource
...
+      throws ConfigurationException {
+    return parserSupplier.getParser(extensions, disableValidations).parse(loadConfigResources(configResources));
+  }
+
+  private static boolean definesOwnExtensionModel(ArtifactAst artifactAst) {
+    ArtifactType artifactType = artifactAst.getArtifactType();
+    switch (artifactType) {
+      case APPLICATION:
+        return artifactAst.topLevelComponentsStream()
+            .anyMatch(component -> APPLICATION_COMPONENT_TYPES.contains(component.getComponentType()));
```
elrodro83, line 241: rename `APPLICATION_COMPONENT_TYPES` to something more like `MULE_EXTENSION_COMPONENT_TYPES`

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/ast/ArtifactAstUtils.java
```diff
@@ -107,43 +146,106 @@ public static ArtifactAst parseAndBuildAppExtensionModel(String[] configResource
...
-      throw new MuleRuntimeException(e);
+                                                    boolean disableValidations)
+      throws ConfigurationException {
+    return parserSupplier.getParser(extensions, disableValidations).parse(loadConfigResources(configResources));
+  }
+
+  private static boolean definesOwnExtensionModel(ArtifactAst artifactAst) {
+    ArtifactType artifactType = artifactAst.getArtifactType();
+    switch (artifactType) {
+      case APPLICATION:
```
elrodro83, line 239: haven't we dropped this already?

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/MuleSdkExtensionModelLoader.java
```diff
@@ -28,6 +29,10 @@ public String getId() {
 
   @Override
   protected ExtensionModelParserFactory getExtensionModelParserFactory(ExtensionLoadingContext context) {
-    return new MuleSdkExtensionModelParserFactory();
+    if (MuleSdkExtensionExtensionModelParserFactory.handles(context)) {
+      return new MuleSdkExtensionExtensionModelParserFactory();
+    } else {
+      return new MuleSdkApplicationExtensionModelParserFactory();
```
elrodro83, line 35: do we still need this?

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkExtensionExtensionModelParserFactory.java
```diff
@@ -0,0 +1,131 @@
...
+    return context.<ArtifactAst>getParameter(MULE_SDK_ARTIFACT_AST_PROPERTY_NAME)
+        .map(ast -> ast.getArtifactType().equals(MULE_EXTENSION))
+        .orElse(true);
+  }
+
+  @Override
+  protected MuleSdkExtensionModelMetadataParser createMetadataParser(ExtensionLoadingContext context) {
+    Optional<String> extensionName = context.getParameter(MULE_SDK_EXTENSION_NAME_PROPERTY_NAME);
+    if (extensionName.isPresent()) {
+      return new MuleSdkApplicationExtensionModelMetadataParser(extensionName.get());
```
elrodro83, line 69: ?

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/metadata/MuleSdkApplicationExtensionModelMetadataParser.java
```diff
@@ -0,0 +1,67 @@
...
+import org.mule.runtime.module.extension.internal.loader.parser.XmlDslConfiguration;
+
+import java.util.Optional;
+
+/**
+ * {@link MuleSdkExtensionModelMetadataParser} implementation for Mule SDK extensions defined as part of an application.
+ *
+ * @since 4.5.0
+ */
+public class MuleSdkApplicationExtensionModelMetadataParser implements MuleSdkExtensionModelMetadataParser {
```
elrodro83, line 30: i don't think we need this anymore

elrodro83 (changes requested):

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/ast/ArtifactAstUtils.java
asanguinetti: renamed to `REUSABLE_COMPONENT_TYPES` and added a TODO to move it somewhere else.

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/ast/ArtifactAstUtils.java
```diff
@@ -107,43 +147,106 @@ public static ArtifactAst parseAndBuildAppExtensionModel(String[] configResource
...
+        return artifactAst.topLevelComponentsStream()
+            .anyMatch(component -> REUSABLE_COMPONENT_TYPES.contains(component.getComponentType()));
+      case MULE_EXTENSION:
+        // TODO: try to unify this one with the case above by changing the Mule Extensions DSL
```
elrodro83, line 244: add ticket id

elrodro83 (changes requested):

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/ast/ArtifactAstUtils.java
```diff
@@ -107,43 +147,106 @@ public static ArtifactAst parseAndBuildAppExtensionModel(String[] configResource
...
+  }
+
+  private static boolean definesOwnExtensionModel(ArtifactAst artifactAst) {
+    ArtifactType artifactType = artifactAst.getArtifactType();
+    switch (artifactType) {
+      case APPLICATION:
+        return artifactAst.topLevelComponentsStream()
+            .anyMatch(component -> REUSABLE_COMPONENT_TYPES.contains(component.getComponentType()));
+      case MULE_EXTENSION:
+        // TODO W-11796932: Try to unify this one with the case above by changing the Mule Extensions DSL
```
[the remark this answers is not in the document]
asanguinetti: The problem is that, at this point, we know the artifact is being parsed as a `MULE_EXTENSION`, but we actually don't know yet what it has inside.
It may be something that is not really an extension (an app's XML for example), because someone got really confused and gave it the wrong XML.
If we remove this check here we will assume it is an extension and will fail horribly before being able to reach the proper validation.
The proper validation is performed over the complete AST, not the partial one. So the artifact-local `ExtensionModel` generation is not protected against malformed artifacts, that is why we need to skip it.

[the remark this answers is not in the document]
asanguinetti: I just added a new tests that fails if we change this to always true. This way we are covered for when we do the TODO.

eze210 (approved):

elrodro83 (dismissed):

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/ast/ArtifactAstUtils.java
marianogonzalez, line 244: I don't see the need for this unification as they're conceptually different things that can (and most likely will) evolve separetly.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/api/extension/MuleSdkExtensionExtensionModelDeclarer.java
```diff
@@ -38,6 +38,14 @@
 public class MuleSdkExtensionExtensionModelDeclarer {
 
   public static final String EXTENSION_CONSTRUCT_NAME = "extension";
+  public static final String NAME_PARAMETER_NAME = "name";
```
marianogonzalez, line 41: shouldn't all these be private?

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/MuleSdkExtensionModelLoader.java
```diff
@@ -28,6 +29,10 @@ public String getId() {
 
   @Override
   protected ExtensionModelParserFactory getExtensionModelParserFactory(ExtensionLoadingContext context) {
-    return new MuleSdkExtensionModelParserFactory();
+    if (MuleSdkExtensionExtensionModelParserFactory.handles(context)) {
```
marianogonzalez, line 32: I don't think we should solve this with an if but with inheritance

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/BaseMuleSdkExtensionModelParserFactory.java
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * Gets a parameter from the given {@link ExtensionLoadingContext} assuming it needs to be present.
+   *
+   * @param context       The {@link ExtensionLoadingContext}.
+   * @param parameterName The parameter name, typically a constant from {@link ExtensionConstants}.
+   * @param <T>           The expected type of the parameter.
+   *
+   * @return The parameter value.
+   */
+  protected <T> T getMandatoryParameter(ExtensionLoadingContext context, String parameterName) {
```
marianogonzalez, line 57: Mandatory -> required. Also use a name which makes it clear that you're refering to loader parameter and not extension parameters

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkExtensionExtensionModelParserFactory.java
```diff
@@ -0,0 +1,135 @@
...
+  private ArtifactAst parseAst(ExtensionLoadingContext context) throws ConfigurationException {
+    Set<ExtensionModel> dependencies = context.getDslResolvingContext().getExtensions();
+
+    String[] resources = {getMandatoryParameter(context, MULE_SDK_RESOURCE_PROPERTY_NAME)};
+    ArtifactAst artifactAst = parseArtifact(resources,
+                                            this::createAstParser,
+                                            dependencies,
+                                            false,
+                                            context.getExtensionClassLoader(),
+                                            "this", // should be irrelevant as long as it doesn't collide
```
marianogonzalez, line 108: how wouldn't it collide if the value is hardcoded?

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/metadata/MuleSdkArtifactLocalExtensionModelMetadataParser.java
```diff
@@ -0,0 +1,66 @@
...
+import static java.util.Optional.of;
+
+import org.mule.runtime.api.meta.Category;
+import org.mule.runtime.module.extension.internal.loader.java.property.LicenseModelProperty;
+import org.mule.runtime.module.extension.internal.loader.parser.XmlDslConfiguration;
+
+import java.util.Optional;
+
+/**
+ * {@link MuleSdkExtensionModelMetadataParser} implementation for Mule SDK extensions defined as part of the same artifact.
```
marianogonzalez, line 25: this jdco and classname are not clear not clearly reflect the domain. Be clear about this being for operations defined as part of an app

marianogonzalez (changes requested):

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/api/extension/MuleSdkExtensionExtensionModelDeclarer.java
asanguinetti: I'm exposing them so that I can use them from the parser, to make sure the same identifiers are used.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/MuleSdkExtensionModelLoader.java
asanguinetti: I don't think I understand exactly what you mean.
I'm actually using inheritance (together with polymorphism) to deal with the Application vs Extension cases, however, at some point I need to figure out which concrete class to instantiate. This is were I'm making the decision.
I'm using this class method (`handles`) to figure out which concrete class is capable of handling the current loading request before instantiating it.

Another option I've been contemplating would be to actually use different ExtensionModelLoaders, one for the application and one for the extension. In that case there would be no need for this if.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/BaseMuleSdkExtensionModelParserFactory.java
asanguinetti: done, changed to `getRequiredLoadingParameter`.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkExtensionExtensionModelParserFactory.java
asanguinetti: It would be weird because there should be an extension in context named "this". However, I just changed it so that it uses a name which includes also the resource file name (which would be more helpful also for troubleshooting if there is any error during parsing).

In any case, keep in mind that this is only the name of the temporary Extension that gets build under the namespace "this" so that the full AST can get generated before doing the final loading of the real ExtensionModel. I explained a little bit more about this in the comments now.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/metadata/MuleSdkArtifactLocalExtensionModelMetadataParser.java
asanguinetti: Originally I had named it `MuleSdkApplicationExtensionModelMetadataParser`. However, I realized that I also needed to generate an `ExtensionModel` for the `this` namespace when loading an extension from its own artifact.

This is why I ended up reusing the same thing, but I renamed it to MuleSdk**ArtifactLocal**ExtensionModelMetadataParser.

So I think the Javadoc and the classname are reflecting what it is actually being used for: parsing the ExtensionModel for the cases in which the ExtensionModel is going to be referenced from the same artifact (either application or extension) in which it is being defined.

I got the concept from [this constant](https://github.com/mulesoft/mule/blob/7bc5e0580c328297c4e453cfc3f9fb33634b73fd/modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkExtensionModelParser.java#L56) which would now be renamed to `ARTIFACT_LOCAL_EXTENSION_NAMESPACE`.

Do you think it is too convoluted?

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/ast/ArtifactAstUtils.java
asanguinetti: I agree, but I'm suggesting to do it not because we **need to** but because it could make our lives easier if we can.
If there comes the time in which we can't do that anymore, we can add complexity then, so long as the API supports it.
I've also created this other task: ([W-11796759](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE0000170sjtYAA/view)) to move this logic into the `mule-extensions-support` module. Also added the TODO [here](https://github.com/mulesoft/mule/pull/11776/files#diff-bdcfd9b812bf50e66a6a7978ac2c90666ce05ab08ae2a57328b7cbfeb168b78cR60).

marianogonzalez: My point is that I don't see how would that make your life any easier, and if it does, it will be short lived. Thus, the TODO only enlarges an already large backlog with a task of little to none value.

As for the second TODO, this shouldn't move to `mule-module-extension-support` but to `mule-module-extensions-mule-support`

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/api/extension/MuleSdkExtensionExtensionModelDeclarer.java
marianogonzalez: parser and declarer should not be coupled

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/MuleSdkExtensionModelLoader.java
marianogonzalez: what you said in the last paragraph is exactly what I meant. Use different loaders with a common parent class

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/ast/ArtifactAstUtils.java
asanguinetti: Yes, my bad, I meant to say `mule-module-extensions-mule-support`.

I just removed the `TODO` with the rationale that the logic is fine as it is. The task [W-11796932](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE0000170jmbYAA/view) may or may not be done, but if it is done, it shouldn't be because of this `TODO`. In case we do it, we may want to come back and simplify this part, but not necessarily.

We can also revisit it when we do [W-11796759](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE0000170sjtYAA/view).

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/api/extension/MuleSdkExtensionExtensionModelDeclarer.java
asanguinetti: You are right in that the parser and declarer should not necessarily be coupled. However, in this case, the parser relies on an AST the components of which should match with the declaration of the "extension" extension. We very much want the component identifiers to match the names of the components and constructs of the model. Similarly, we want the component parameter names to match the parameter names in the model. If they don't match, the AST parsing will fail.

For this reason I think having these defined only once makes more sense.

That being said, I like your idea of moving them to a separate constants class. I just pushed that change. Thanks.

Also note that this new class (an interface actually) I put it under `api`, because I suppose these constants will not be allowed to change without breaking backwards. However, we could prefer to leave them under `internal` to be allowed to rename the constants themselves without changing their values. What do you think?

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/MuleSdkExtensionModelLoader.java
asanguinetti: Done. We have two separate loaders now, each with its ID. The `handles` method was no longer necessary, so I removed it. Each parser is now a little bit simpler I think. Thanks for the suggestion.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/metadata/MuleSdkArtifactLocalExtensionModelMetadataParser.java
asanguinetti: Renamed `ArtifactLocal` to `This`, for example: `MuleSdkArtifactLoaclExtensionModelMetadataParser` to `MuleSdkThisExtensionModelMetadataParser`.
I couldn't use `Application` because it would also be misleading. But I suppose the extension `this` is a clearer concept for someone working on this module. In the comments I explain that these classes are used for extensions that are consumed from the same artifact in which they are defined.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/api/extension/MuleSdkExtensionExtensionModelDeclarer.java
marianogonzalez: should be internal. The class is only a holder for values which reflect an schema. The schema is what's api, not the class holding them

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/ast/ArtifactAstUtils.java
```diff
@@ -107,43 +147,106 @@ public static ArtifactAst parseAndBuildAppExtensionModel(String[] configResource
...
+                                                    boolean disableValidations)
+      throws ConfigurationException {
+    return parserSupplier.getParser(extensions, disableValidations).parse(loadConfigResources(configResources));
+  }
+
+  private static boolean definesOwnExtensionModel(ArtifactAst artifactAst) {
+    ArtifactType artifactType = artifactAst.getArtifactType();
+    switch (artifactType) {
+      case APPLICATION:
+        return artifactAst.topLevelComponentsStream()
```
marianogonzalez, line 242: this is conceptually wrong. If the application doesn't define any reusable components it doesn't mean it doesn't have an extension model. It just means it has one with no components. There's still value in having an extension model that simply exposes types, or to at least being able to assume that the extension model will always be there.

Not a fan of the method name either.

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/ast/ArtifactAstUtils.java
```diff
@@ -107,43 +147,106 @@ public static ArtifactAst parseAndBuildAppExtensionModel(String[] configResource
...
+   * @param dependencies        the dependencies in context.
+   * @return an optional {@link ExtensionModel}
+   */
+  public static Optional<ExtensionModel> parseArtifactExtensionModel(ArtifactAst ast,
+                                                                     ClassLoader artifactClassLoader,
+                                                                     String artifactId,
+                                                                     Optional<String> artifactVersion,
+                                                                     Set<ExtensionModel> dependencies) {
+    if (!artifactVersion.isPresent()) {
+      logModelNotGenerated("No version specified", artifactId, ast.getArtifactType());
```
marianogonzalez, line 179: shouldn't it fail rather than returning empty?

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/ast/ArtifactAstUtils.java
```diff
@@ -107,43 +147,106 @@ public static ArtifactAst parseAndBuildAppExtensionModel(String[] configResource
...
     }
   }
 
+  /**
+   * Handles a {@link ValidationResult} uniformly. Warnings are logged, errors raise exceptions.
+   *
+   * @param validationResult The {@link ValidationResult} obtained using an {@link ArtifactAstValidator}.
+   * @throws ConfigurationException If there was any result item with {@link ERROR} level.
+   */
+  public static void handleValidationResult(ValidationResult validationResult)
```
marianogonzalez, line 204: You're putting this method as API but is only used once in an internal class. This should simply be a private method of the factory in which you use it, or at most, part of an internal utils. Don't make things API unnecesarily.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/api/dsl/MuleSdkDslConstants.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.module.extension.mule.api.dsl;
```
marianogonzalez, line 7: internal

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/MuleSdkThisExtensionModelLoader.java
```diff
@@ -0,0 +1,33 @@
...
+import org.mule.runtime.module.extension.internal.loader.AbstractExtensionModelLoader;
+import org.mule.runtime.module.extension.internal.loader.parser.ExtensionModelParserFactory;
+import org.mule.runtime.module.extension.mule.internal.loader.parser.MuleSdkThisExtensionModelParserFactory;
+
+/**
+ * {@link ExtensionModelLoader} implementation for Mule SDK Extensions defined as part of the same artifact.
+ *
+ * @since 4.5.0
+ */
+public class MuleSdkThisExtensionModelLoader extends AbstractExtensionModelLoader {
```
marianogonzalez, line 22: This name is worst than the original. Why not simply `ApplicationExtensionModelLoader` ?

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/BaseMuleSdkExtensionModelParser.java
```diff
@@ -22,7 +22,7 @@
  *
  * @since 4.5.0
  */
-abstract class BaseMuleSdkExtensionModelParser {
+public abstract class BaseMuleSdkExtensionModelParser {
```
marianogonzalez, line 25: why? Looks like all implementations are in the same package

marianogonzalez (changes requested):

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/ast/ArtifactAstUtils.java
asanguinetti: Similarly, I didn't want to change the [previous behavior](https://github.com/mulesoft/mule/blob/f3ad65b953a6af055a0aeee06e7c9bd19929c5b9/modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/ast/ArtifactAstUtils.java#L120) to avoid breaking other unrelated things with this PR.
We could review that separately.

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/ast/ArtifactAstUtils.java
asanguinetti: You are right. I just moved it to its own utility class on an internal package. Thanks

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/BaseMuleSdkExtensionModelParser.java
asanguinetti: I have some new ones that are in the sub-package parser.**metadata**, Java doesn't treat it as the same package, hence the visibility issue.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/api/extension/MuleSdkExtensionExtensionModelDeclarer.java
asanguinetti: Moved to internal now.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/MuleSdkThisExtensionModelLoader.java
asanguinetti: I just changed this as discussed yesterday. Now we have one `MuleSdkApplicationExtensionModelLoader` and one `MuleSdkExtensionExtensionModelLoader`.
The `MuleSdkApplicationExtensionModelLoader` is the easiest one because it keeps doing what it has been doing before.
The `MuleSdkExtensionExtensionModelLoader` has a decorator for providing the temporary local `ExtensionModel` used for AST validations (in the context of the chicken and egg problem). This way we avoid the complexity of having to deal with different parsers depending on the context and it prepares the ground for the next move which would be generating the `ExtensionModel` only once.
The undecorated one is readily available so it would make no sense to generate it again, the problem now is only that we have it inside the parser, which has no way of letting the loader know it doesn't need to call the `LoaderDelegate` and then pass it to the `ExtensionModelFactory`. I'll experiment with a solution for that now.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkExtensionModelParser.java
```diff
@@ -77,19 +77,17 @@ public List<ModelProperty> getAdditionalModelProperties() {
 
   @Override
   public String getName() {
-    return extensionName;
+    return metadataParser.getName();
```
marianogonzalez, line 80: this feels unnecesary complex. Why introducing this concept of metadataParser (which by the way is a very confusing name since things like `name` are not metadata) when you can simply apply the template pattern to the MuleSdkModelParser?

marianogonzalez (changes requested):

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/ast/ApplicationExtensionModelsEnricher.java
```diff
@@ -0,0 +1,113 @@
...
+  private final String artifactId;
+  private final Optional<String> artifactVersion;
+
+  public ApplicationExtensionModelsEnricher(String artifactId, Optional<String> artifactVersion) {
+    this.artifactId = artifactId;
+    this.artifactVersion = artifactVersion;
+  }
+
+  @Override
+  public boolean applicable(ArtifactAst ast) {
```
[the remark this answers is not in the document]
asanguinetti: This method is part of the `ArtifactExtensionModelsEnricher` (I'm in the process of choosing a better name for it).
Even though the user of this specific concrete class will know that it is used in the context of an application, [the code that consumes the instance](https://github.com/mulesoft/mule/pull/11776/files#diff-8a588bb1806dd53f441fcbda4dfe927efa22ad5e5deeabe5126e0016b41ab4e2R52) will not know if it is happening in the context of an application or an extension. It only sees the interface and not the concrete class.
Still, I'm considering refactoring this class so that it returns an `Optional<ExtensionModel>` instead of an "enriched" `Set<ExtensionModel>`. If I do that, I think I could make this a private method and call it directly from the parsing/loading method.
For now I renamed it to `isApplicable` as suggested.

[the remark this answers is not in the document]
asanguinetti: Removed.

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/ast/ApplicationExtensionModelsEnricher.java
```diff
@@ -0,0 +1,113 @@
...
+        .anyMatch(component -> REUSABLE_COMPONENT_TYPES.contains(component.getComponentType()));
+  }
+
+  @Override
+  public Set<ExtensionModel> getEnrichedExtensionModels(ArtifactAst ast, ClassLoader classLoader,
+                                                        Set<ExtensionModel> extensions) {
+    Optional<ExtensionModel> artifactExtensionModel =
+        parseApplicationExtensionModel(ast, classLoader, extensions);
+
+    if (artifactExtensionModel.isPresent()) {
```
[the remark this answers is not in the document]
asanguinetti: This piece of code is no longer present in the most recent version.
That being said, I similar fragment now appears [here](https://github.com/mulesoft/mule/pull/11776/files#diff-8a588bb1806dd53f441fcbda4dfe927efa22ad5e5deeabe5126e0016b41ab4e2R55), but in this case I think it will be a lot easier to read without using the functional style.
Let me know if you still would like me to change that new one.

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/ast/ApplicationExtensionModelsEnricher.java
```diff
@@ -0,0 +1,113 @@
...
+   *
+   * @param ast                 the application's AST
+   * @param artifactClassLoader the application's classloader
+   * @param dependencies        the dependencies in context.
+   * @return an optional {@link ExtensionModel}
+   */
+  public Optional<ExtensionModel> parseApplicationExtensionModel(ArtifactAst ast,
+                                                                 ClassLoader artifactClassLoader,
+                                                                 Set<ExtensionModel> dependencies) {
+    if (!artifactVersion.isPresent()) {
```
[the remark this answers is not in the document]
asanguinetti: Done, now it throws `ConfigurationException` with the same message that was previously used for logging.

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/ast/ArtifactAstUtils.java
```diff
@@ -0,0 +1,90 @@
...
+
+import java.io.IOException;
+import java.util.Set;
+
+/**
+ * Utilities for parsing and handling {@link ArtifactAst} (for internal usage only).
+ *
+ * @since 4.5.0
+ */
+@NoInstantiate
```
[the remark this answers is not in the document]
asanguinetti: Removed the annotation, as pointed out, this class already had a private constructor to prevent instantiation, so the annotation was completely redundant.

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/ast/ArtifactExtensionModelsEnricher.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package org.mule.runtime.module.artifact.activation.internal.ast;
+
+import org.mule.runtime.api.meta.model.ExtensionModel;
+import org.mule.runtime.ast.api.ArtifactAst;
+
+import java.util.Set;
+
+/**
+ * Allows for enriching the {@link ExtensionModel}s available in the context of the {@link ArtifactAst} being parsed.
```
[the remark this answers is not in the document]
asanguinetti: Done. Renamed to `ArtifactExtensionModelParser`. Also renamed subclasses accordingly of course.
Also, as suggested, it no longer returns an "enriched" `Set<ExtensionModel>`, it now returns an `Optional<ExtensionModel>` and it is the caller's responsibility to do what needs to be done with it.

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/ast/validation/ValidationUtils.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Collection;
+
+import org.slf4j.Logger;
+
+/**
+ * Utilities for handling {@link ValidationResult}s in a uniform way.
+ *
+ * @since 4.5.0
+ */
+public class ValidationUtils {
```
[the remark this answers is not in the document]
asanguinetti: Done, renamed to `AstValidationUtils`.

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/ast/validation/ValidationUtils.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  /**
+   * Handles a {@link ValidationResult} uniformly. Warnings are logged, errors raise exceptions.
+   *
+   * @param validationResult The {@link ValidationResult} obtained using an {@link ArtifactAstValidator}.
+   * @param logger           The {@link Logger} to use for logging.
+   * @throws ConfigurationException If there was any result item with {@link ERROR} level.
+   */
+  public static void handleValidationResult(ValidationResult validationResult, Logger logger)
```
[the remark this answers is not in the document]
asanguinetti: Here I have a couple of options.

1. What this method is doing is quite generic, so I could split it in two: `logWarnings` and `throwIfContainsErrors`.
Since it is only used in two places, it sounds manageable, but my concern would be that most often than not, we would like to do both things together, so we would be repeating the pattern of calling `logWarnings` followed by `throwIfContainsErrors`. 
2. We could keep it as one method but rename it to something more explicit like `logWarningsAndThrowIfContainsErrors`. It is borderline verbose, but could work.
3. I'm not so convinced that this method "isn't handling anything". It takes the results of the validation and performs an action appropriate for each of the items (or in aggregation). If we go too over-specific on the method name we would be more limited in how we can change what we do with the items in the future without breaking the semantics.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/BaseMuleSdkExtensionModelParserFactory.java
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * Implementations need to specify how to extract some parameters needed by the parser from the extension loading context.
+ *
+ * @since 4.5.0
+ */
+public abstract class BaseMuleSdkExtensionModelParserFactory implements ExtensionModelParserFactory {
+
+  @Override
+  public ExtensionModelParser createParser(ExtensionLoadingContext context) {
+    return new MuleSdkExtensionModelParser(createMetadataParser(context),
```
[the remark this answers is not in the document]
asanguinetti: This is because there is only one implementation of `MuleSdkExtensionModelParser`, I preferred to go with composition over inheritance in this case. That is why I'm providing the strategy for parsing the metadata and for getting the top level elements which are the only differences so far between applications and extensions.

But maybe you are right in that it will be simpler using inheritance with templating. Specially considering that I'm already using templating at the factory level. I can try and see how it looks.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkExtensionExtensionModelParserFactory.java
```diff
@@ -0,0 +1,151 @@
...
+import org.slf4j.Logger;
+
+/**
+ * {@link ExtensionModelParserFactory} implementation for Mule SDK in the context of extensions.
+ *
+ * Takes a resource file as parameter of the {@link ExtensionLoadingContext} using {@link MULE_SDK_RESOURCE_PROPERTY_NAME}.
+ *
+ * @since 4.5.0
+ */
+public class MuleSdkExtensionExtensionModelParserFactory extends BaseMuleSdkExtensionModelParserFactory
```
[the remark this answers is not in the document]
asanguinetti: This was intentional to respect the naming pattern of `<Whatever>ExtensionModelParserFactory`, in this case <Whatever> would be replaced by `MuleSdkExtension` or `MuleSdkApplication`.
You will notice this in many classes of this PR. I can rename them if you think this is confusing.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkExtensionExtensionModelParserFactory.java
```diff
@@ -0,0 +1,151 @@
...
+    Set<ExtensionModel> dependencies = context.getDslResolvingContext().getExtensions();
+
+    String version = getRequiredLoadingParameter(context, VERSION_PROPERTY_NAME);
+    String[] resources = {getRequiredLoadingParameter(context, MULE_SDK_RESOURCE_PROPERTY_NAME)};
+    ArtifactAst artifactAst = parseArtifactWithExtensionsEnricher(resources,
+                                                                  this::createAstParser,
+                                                                  dependencies,
+                                                                  false,
+                                                                  context.getExtensionClassLoader(),
+                                                                  new MuleSdkLocalExtensionModelsEnricher(version,
```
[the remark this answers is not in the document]
asanguinetti: Renamed to `MuleSdkExtensionExtensionModelParser`. Could have a similar issue as discussed in [this comment](https://github.com/mulesoft/mule/pull/11776#discussion_r988106970). If you don't agree with my answer for that one, we can think of a different name for this one.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkExtensionModelParser.java
asanguinetti: Again, this was just me having picked composition over inheritance. See [this other answer](https://github.com/mulesoft/mule/pull/11776#discussion_r988374631) for more details.
I'll try using the template pattern and see how it looks.

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/ast/validation/ValidationUtils.java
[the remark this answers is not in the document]
asanguinetti: Renamed to `logWarningsAndThrowIfContainsErrors`.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkExtensionExtensionModelParserFactory.java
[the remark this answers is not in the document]
asanguinetti: Renamed these to `MuleSdkPluginExtensionModelParserFactory`, and so on.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkExtensionModelParser.java
asanguinetti: This has been changed to use templating. It looks simpler now.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/MuleSdkPluginExtensionModelLoader.java
```diff
@@ -0,0 +1,111 @@
...
+    return MULE_SDK_EXTENSION_LOADER_ID;
+  }
+
+  @Override
+  protected ExtensionModelParserFactory getExtensionModelParserFactory(ExtensionLoadingContext context) {
+    return new MuleSdkPluginExtensionModelParserFactory();
+  }
+
+  @Override
+  public ExtensionModel loadExtensionModel(ExtensionModelLoadingRequest request) {
```
marianogonzalez, line 60: by overriding the entire method you're skipping non-negoatible pieces of logic. Need to extract and override with  finer granularity

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/MuleSdkPluginExtensionModelLoader.java
```diff
@@ -0,0 +1,111 @@
...
+      ArtifactAst artifactAst = parseArtifact(resources,
+                                              this::createAstParser,
+                                              dependencies,
+                                              false,
+                                              request.getExtensionClassLoader(),
+                                              loadingHelper);
+
+      // Applies the AST validators and throws if there was any error
+      logWarningsAndThrowIfContainsErrors(validatorBuilder().build().validate(artifactAst), LOGGER);
+      return loadingHelper.getExtensionModel();
```
marianogonzalez, line 105: won't this end up looking up for another instance of this same loader in order to create the extension model?

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/AbstractMuleSdkExtensionModelParser.java
```diff
@@ -6,64 +6,56 @@
...
 /**
  * {@link ExtensionModelParser} implementation for Mule SDK extensions
  *
  * @since 4.5.0
  */
-public class MuleSdkExtensionModelParser implements ExtensionModelParser {
-
-  // The namespace of the extension when it's defined within an application rather than in a separate artifact.
-  public static final String APP_LOCAL_EXTENSION_NAMESPACE = "THIS";
+public abstract class AbstractMuleSdkExtensionModelParser extends BaseMuleSdkExtensionModelParser
```
marianogonzalez, line 48: Abstract extends Base? Rename one of the two

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkPluginExtensionModelParser.java
```diff
@@ -0,0 +1,126 @@
...
+    return getExtensionComponentAst(ast).directChildrenStream();
+  }
+
+  private ComponentAst getExtensionComponentAst(ArtifactAst ast) {
+    // At this point we can assume there is only one top level component which is the extension:extension component
+    // We don't need to check for this because it should be guaranteed by previous validations
+    return ast.topLevelComponents().get(0);
+  }
+
+  private void parseMetadata(ComponentAst extensionComponentAst) {
```
marianogonzalez, line 94: rename to `parseStructure`, not only for consistency but because this isn't parsing metadata.

modules/extensions-mule-support/src/test/java/org/mule/runtime/module/extension/mule/internal/loader/ast/AbstractMuleSdkAstTestCase.java
```diff
@@ -45,13 +45,13 @@ public abstract class AbstractMuleSdkAstTestCase extends AbstractMuleTestCase {
...
 
-  private static List<ExtensionModel> runtimeExtensionModels;
+  protected static Set<ExtensionModel> runtimeExtensionModels;
 
   protected abstract String getConfigFile();
 
   @BeforeClass
   public static void beforeClass() throws Exception {
-    runtimeExtensionModels = new ArrayList<>();
+    runtimeExtensionModels = new HashSet<>();
```
marianogonzalez, line 54: why? this might turn the test flaky. At most, it should be a LinkedHashSet to maintain behavior

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/ast/ArtifactAstUtils.java
```diff
@@ -0,0 +1,100 @@
...
+   * @param extensionModelParser a parser capable of providing the {@link ExtensionModel} which models the artifact itself.
+   * @return an {@link ArtifactAst}
+   * @throws ConfigurationException it the artifact couldn't be parsed
+   */
+  public static ArtifactAst parseArtifact(String[] configResources,
+                                          AstXmlParserSupplier parserSupplier,
+                                          Set<ExtensionModel> extensions,
+                                          boolean disableValidations,
+                                          ClassLoader artifactClassLoader,
+                                          MuleSdkExtensionModelLoadingHelper extensionModelParser)
```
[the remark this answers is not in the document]
asanguinetti: Fixed.

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/ast/MuleSdkExtensionModelLoadingHelper.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.mule.runtime.api.meta.model.ExtensionModel;
+import org.mule.runtime.ast.api.ArtifactAst;
+import org.mule.runtime.core.api.config.ConfigurationException;
+
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * Allows for extracting an {@link ExtensionModel} which represents the {@link ArtifactAst} being parsed.
```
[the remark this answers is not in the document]
asanguinetti: No longer needed, reverted back to the immutable version to avoid these kind of issues. The method `getExtensionModel` has been removed.

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/ast/MuleSdkExtensionModelLoadingHelper.java
```diff
@@ -0,0 +1,38 @@
...
+   * @throws ConfigurationException if the artifact couldn't be parsed.
+   */
+  Optional<ExtensionModel> loadExtensionModel(ArtifactAst ast, ClassLoader classLoader, Set<ExtensionModel> extensions)
+      throws ConfigurationException;
+
+  /**
+   * @return an {@link ExtensionModel} that represents the {@code ast}. Previously loaded using
+   *         {@link #loadExtensionModel(ArtifactAst, ClassLoader, Set)}.
+   */
+  Optional<ExtensionModel> getExtensionModel();
```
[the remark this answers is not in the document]
asanguinetti: Removed.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/MuleSdkPluginExtensionModelLoader.java
asanguinetti: Overriding only the `doCreate` now, which has a reduced scope.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/MuleSdkPluginExtensionModelLoader.java
asanguinetti: The `getExtensionModel` method wouldn't, the `loadExtensionModel` would. That one gets called as part of the parsing. The `getExtensionModel` method doesn't exist anymore now.

The `loadExtensionModel` still works as it was. We don't need it to be the same instance of the loader for it to work (because it is stateless), but we could change it to use the same instance if we think it makes more sense, or if it we just want to avoid creating another instance (with the associated enrichers and validators) for performance reasons.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/AbstractMuleSdkExtensionModelParser.java
asanguinetti: Renamed the abstract one to just `MuleSdkExtensionModelParser`.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkPluginExtensionModelParser.java
asanguinetti: Renamed

modules/extensions-mule-support/src/test/java/org/mule/runtime/module/extension/mule/internal/loader/ast/AbstractMuleSdkAstTestCase.java
asanguinetti: The reason I changed it to a `Set` impl was because in all places we were using it, we either required a `Collection` or a `Set` ([here](https://github.com/mulesoft/mule/blob/6b0eee046bc371e81193208c788a287be85d9203/modules/extensions-mule-support/src/test/java/org/mule/runtime/module/extension/mule/internal/loader/MuleSdkPluginExtensionModelLoaderTestCase.java#L169)), and for the later, we would have to be creating the `Set` from the `List` anyway everytime.

Changed to `LinkedHashSet`.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/MuleSdkPluginExtensionModelLoader.java
asanguinetti: Done, reusing the same loader instance now.

marianogonzalez (approved):

PR #11781, opened 2022-09-21
W-1169886: Adding Transaction timeout to transaction created in TryScope (cherrypick from W-11741912)

mbuchwald:


mbuchwald:
--validate

jcgrethe (approved):

PR #11784, opened 2022-09-21
W-11680184: Implement changes to show / not show policies

fsgonz:


core/src/main/java/org/mule/runtime/core/internal/profiling/tracing/event/span/NamedSpanBasedOnParentSpanChildSpanCustomizationInfo.java
```diff
@@ -42,4 +48,10 @@ private Optional<InternalSpan> getSpan(CoreEvent coreEvent) {
   public String getLocationAsString(CoreEvent coreEvent) {
     return getSpan(coreEvent).map(internalSpan -> internalSpan.getAttribute(LOCATION_KEY).orElse("")).orElse("");
   }
+
+  @Override
+  public boolean isExportable(CoreEvent coreEvent) {
```
sofiamorseletto, line 53: Shouldn't this method implementation go on the child class (`NoChildrenExportableUntilExecuteNextSpanCustomizationInfo`)?

fsgonz: No, because this is done to avoid exporting a double execute-next:route as the execute next is a processor. But I will refactor this to add it to a different spanCustomizerInfo.

fsgonz: I've addressed this comment [in this commit.](https://github.com/mulesoft/mule/pull/11784/commits/987c454578f89317f0bb9df02ade4a2b91ec5403)

IvanAndresFritzler (approved):

PR #11786, opened 2022-09-21
W-11794071: Adding new artifact type for Mule SDK Extensions.

asanguinetti:


modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/validation/ExtensionStructureValidations.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  public boolean applicable(ArtifactAst artifactAst) {
+    return artifactAst.getArtifactType().equals(MULE_EXTENSION);
+  }
+
+  private Optional<ValidationResultItem> validateContainsExtension(ArtifactAst artifactAst) {
+    List<ComponentAst> topLevelComponents = artifactAst.topLevelComponents();
+    if (topLevelComponents.size() != 1) {
+      return of(create(emptyList(), this, "Expected exactly one top level component"));
```
elrodro83, line 60: mention what is the expected top level component in the message

elrodro83 (changes requested):

PR #11795, opened 2022-09-22
W-11791554: add resources paths in DeployableProjectModel

jcgrethe:


jcgrethe:
--validate

jcgrethe:
--validate

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/deployable/DeployableProjectModel.java
```diff
@@ -52,11 +53,14 @@ public final class DeployableProjectModel {
   private final Set<BundleDescriptor> sharedLibraries;
   private final Map<BundleDescriptor, List<BundleDependency>> additionalPluginDependencies;
 
+  private final List<Path> resourcesPath;
```
elrodro83, line 56: isn't this just something intermediate for calculating the value for `resources`?

elrodro83 (changes requested):

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/deployable/DeployableProjectModel.java
jcgrethe: I will need it in the mule-framework to create the classloaders pointing to those resources folders

elrodro83 (approved):

PR #11809, opened 2022-09-26
W-11784411: Apply error mappings when calculating what errors may be thrown by a reusable operation

eze210:


modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/utils/AggregatedErrorsCharacteristic.java
```diff
@@ -78,10 +84,46 @@ public List<ErrorModelParser> apply(ComponentAstWithHierarchy componentAstWithHi
...
+          .filter(errorMapping -> doesMappingSourceMatch(errorMapping, errorModel))
+          .map(Aggregator::buildParserFromTarget)
+          .findFirst()
+          .orElseGet(() -> new MuleSdkErrorModelParser(errorModel));
+    }
+
+    private static MuleSdkErrorModelParser buildParserFromTarget(ErrorMapping errorMapping) {
+      String mappingTargetAsString = errorMapping.getTarget();
+      ComponentIdentifier targetErrorId = parseErrorType(mappingTargetAsString, CORE_ERROR_NAMESPACE);
+      return new MuleSdkErrorModelParser(targetErrorId.getNamespace(), targetErrorId.getName(), null);
```
asanguinetti, line 118: If the error was from `CORE` (or eventually from the same extension), shouldn't we try to retain the parent information? here it seems we are always setting it to `null`.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/utils/AggregatedErrorsCharacteristic.java
```diff
@@ -78,10 +84,46 @@ public List<ErrorModelParser> apply(ComponentAstWithHierarchy componentAstWithHi
...
+      List<ErrorMapping> errorMappings = errorMappingsOpt.get();
+      return applyMappingIsSomeMatches(errorModel, errorMappings);
+    }
+
+    private MuleSdkErrorModelParser applyMappingIsSomeMatches(ErrorModel errorModel, List<ErrorMapping> errorMappings) {
+      return errorMappings.stream()
+          .filter(errorMapping -> doesMappingSourceMatch(errorMapping, errorModel))
+          .map(Aggregator::buildParserFromTarget)
+          .findFirst()
+          .orElseGet(() -> new MuleSdkErrorModelParser(errorModel));
```
asanguinetti, line 112: Have we checked that the `findFirst` here represents how the error mapping logic will actually be applied?
I understand it doesn't really make much sense to have multiple mappings with the same `sourceType`, but it could happen more often with parent relationships, where there is a mapping for a base error type and another mapping for a more specific one.
Should the first one be used? or the more specific one?

asanguinetti (commented):
A few questions

asanguinetti (changes requested):
A small typo.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/utils/AggregatedErrorsCharacteristic.java
eze210: Yes, it can be seen in this test: https://github.com/mulesoft/mule-ee/blob/master/modules/http-policy/src/test/resources/http-proxy-operation-policy-with-transitive-error-mappings.xml
Where the assertion in the test is for the message "Policy custom error mapping was applied".
I will add a test to cover this case

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/utils/AggregatedErrorsCharacteristic.java
```diff
@@ -78,10 +84,46 @@ public List<ErrorModelParser> apply(ComponentAstWithHierarchy componentAstWithHi
...
+      Optional<List<ErrorMapping>> errorMappingsOpt = errorMappingsParam.<List<ErrorMapping>>getValue().getValue();
+      if (!errorMappingsOpt.isPresent()) {
+        return new MuleSdkErrorModelParser(errorModel);
+      }
+
+      List<ErrorMapping> errorMappings = errorMappingsOpt.get();
+      return applyMappingIsSomeMatches(errorModel, errorMappings);
+    }
+
+    private MuleSdkErrorModelParser applyMappingIsSomeMatches(ErrorModel errorModel, List<ErrorMapping> errorMappings) {
```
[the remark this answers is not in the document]
eze210: Fixed

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/utils/AggregatedErrorsCharacteristic.java
eze210: Test added

asanguinetti: Mmm.. but I wasn't referring to transitive mappings, I meant something like this:

```
<error-mapping sourceType="HEISENBERG:CONNECTIVITY" targetType="MY:MAPPEDCONNECTIVITY"/>
<error-mapping sourceType="HEISENBERG:HEALTH" targetType="MY:MAPPEDHEALTH"/>
```
What would happen if the operation threw `HEISENBERG:HEALTH` which is a child of `HEISENBERG:CONNECTIVITY`?
With the `findFirst` logic, it will be assumed to be mapped to `MY:MAPPEDCONNECTIVITY`, but is that the same thing that will actually happen in runtime? Which makes me wonder if there isn't any logic that we should be reusing.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/utils/AggregatedErrorsCharacteristic.java
eze210: The target namespace in error-mapping might be:
1. The namespace "THIS": It's not allowed by a AST validation that prevents it to use namespaces from some extension. If you want to raise an error from such namespace, you need to use the operation:raise-error component
2. The namespace of some existing extension: It's not allowed by the same AST validation
3. The core namespace ("MULE"): It is allowed, but is there some core error that doesn't inherit from MULE:ANY directly? I can't find one
4. Some original super-cool namespace, as "SUPERCOOL": In this case, there isn't any syntax to make that error not-a-child of MULE:ANY

eze210: 3. I answer myself: Actually there are several core errors with a parent other than MULE:ANY. Working on that...

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/utils/AggregatedErrorsCharacteristic.java
eze210: Added a test for this case too

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/utils/AggregatedErrorsCharacteristic.java
eze210: Fixed

asanguinetti (approved):

PR #11822, opened 2022-09-29
W-11573985: Make tracing tests more declarative

sofiamorseletto:


sofiamorseletto:
--validate

tests/infrastructure/src/main/java/org/mule/test/infrastructure/profiling/SpanTestHierarchy.java
```diff
@@ -0,0 +1,111 @@
...
+    CapturedExportedSpan parentSpan = spanHashMap.get(expectedSpan.getParentSpanId());
+    if (expectedParentName != null && parentSpan == null) {
+      return false;
+    } else if (expectedParentName == null && parentSpan == null) {
+      return expectedSpan.getParentSpanId().equals(NO_PARENT_SPAN);
+    }
+    return parentSpan.getName().equals(expectedParentName);
+  }
+
+  public class SpanNode {
```
fsgonz, line 97: add jdoc as this is public.

tests/infrastructure/src/main/java/org/mule/test/infrastructure/profiling/SpanTestHierarchy.java
```diff
@@ -0,0 +1,111 @@
...
+      currentNode = currentNode.parent.parent;
+    }
+    return this;
+  }
+
+  public SpanNode getRoot() {
+    return root;
+  }
+
+  public void assertSpanTree(SpanNode node, CapturedExportedSpan parent) {
```
fsgonz, line 68: add jdoc to public methods.

tests/infrastructure/src/main/java/org/mule/test/infrastructure/profiling/SpanTestHierarchy.java
```diff
@@ -0,0 +1,111 @@
...
+    CapturedExportedSpan capturedExportedSpan;
```
fsgonz, line 69: declaration and assignation can be in the same line.

tests/infrastructure/src/main/java/org/mule/test/infrastructure/profiling/SpanTestHierarchy.java
fsgonz: Do I really need the parent parameter in the public method or is it only needed to perform the recursion?I would any way create a public method where you don't pass the second parameter so that you don't need to pass null (which is the most common use).

tests/infrastructure/src/main/java/org/mule/test/infrastructure/profiling/SpanTestHierarchy.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.List;
+
+public class SpanTestHierarchy {
+
+  private SpanNode root;
+  private SpanNode currentNode;
+  private SpanNode lastChild;
+  private final String NO_PARENT_SPAN = "0000000000000000";
+  private HashSet<String> visitedSpans = new HashSet();
+  private HashMap<String, CapturedExportedSpan> spanHashMap = new HashMap<>();
```
fsgonz, line 28: This can be final.

tests/infrastructure/src/main/java/org/mule/test/infrastructure/profiling/SpanTestHierarchy.java
fsgonz, line 97: inner class should be static.

tests/infrastructure/src/main/java/org/mule/test/infrastructure/profiling/SpanTestHierarchy.java
```diff
@@ -0,0 +1,111 @@
...
+      return false;
+    } else if (expectedParentName == null && parentSpan == null) {
+      return expectedSpan.getParentSpanId().equals(NO_PARENT_SPAN);
+    }
+    return parentSpan.getName().equals(expectedParentName);
+  }
+
+  public class SpanNode {
+
+    private String spanName;
```
fsgonz, line 99: final parameters.

tests/infrastructure/src/main/java/org/mule/test/infrastructure/profiling/SpanTestHierarchy.java
sofiamorseletto: Actually, I'm not using it outside of this class so I'll make it private

sofiamorseletto:
--validate

fsgonz (approved):

sofiamorseletto:
--validate

fsgonz (approved):

PR #11824, opened 2022-09-30
W-11835623: Add cause parameter to the operation:raise-error component

eze210:


core/src/main/java/org/mule/runtime/core/api/processor/AbstractRaiseErrorProcessor.java
```diff
@@ -61,10 +60,12 @@ public void initialise() throws InitialisationException {
 
   protected abstract ComponentIdentifier calculateErrorIdentifier(String typeId);
 
+  protected abstract MuleRuntimeException getException(ErrorType type, String message, CoreEvent event);
```
asanguinetti, line 63: Are we sure we want to allow this one to return a `MuleRuntimeException`? shouldn't we require it to return a `TypedException`?
Maybe this is me being too pedantic because we only have two implementations and both return a `TypedException`. I'm just thinking "if we ever had a third one (and I hope we don't), would we want it to be able to return a `MuleRuntimeException`?"

asanguinetti (approved):
Just added a small comment/suggestion.
Otherwise LGTM.

core/src/main/java/org/mule/runtime/core/api/processor/AbstractRaiseErrorProcessor.java
eze210: Done

asanguinetti (approved):

PR #11834, opened 2022-10-03
W-11793048: Scopes should ignore the export of the routes

sofiamorseletto:


modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ComponentMessageProcessorObjectFactory.java
```diff
@@ -66,7 +68,7 @@ public P doGetObject() {
     final MessageProcessorChain nestedChain;
 
     if (nestedProcessors != null) {
-      nestedChain = newChain(empty(), nestedProcessors);
```
sofiamorseletto, line 69: I found this is the class that handles the custom scopes, is that right?

fsgonz (approved):

PR #11837, opened 2022-10-05
W-10883933: Decoupling syntax from dynamic metadata feature

Bardammu:


modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/java/JavaDeclarationDelegateTestCase.java
```diff
@@ -373,7 +373,7 @@ public void flowListeningOperationWithoutAttributes() {
 
     OperationDeclaration operation = getOperation(getConfiguration(declaration, BANANA), "getLunch");
     assertThat(operation, is(notNullValue()));
-    assertOutputType(operation.getOutput(), toMetadataType(Fruit.class), false);
+    assertOutputType(operation.getOutput(), toMetadataType(Fruit.class), true);
```
SebaElizalde, line 376: Why did this change? Are we sure we are not breaking anything?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/utils/JavaOutputResolverModelParserUtils.java
```diff
@@ -27,87 +28,182 @@
...
+                                  });
+
+    if (!attributesResolverModelParser.isPresent()) {
+      attributesResolverModelParser = mapReduceSingleAnnotation(extensionType, "source", "",
+                                                                org.mule.runtime.extension.api.annotation.metadata.MetadataScope.class,
+                                                                MetadataScope.class,
+                                                                valueFetcher -> {
+                                                                  Type type = valueFetcher
+                                                                      .getClassValue(org.mule.runtime.extension.api.annotation.metadata.MetadataScope::outputResolver);
+                                                                  Class<?> declaringClass = type.getDeclaringClass().orElse(null);
```
SebaElizalde, line 177: Can we use .map() to simplify this code and avoid null checking?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/delegate/OperationModelLoaderDelegate.java
```diff
@@ -141,4 +191,81 @@ private void declareErrorModels(OperationDeclarer operation, OperationModelParse
...
   }
+
+  private void declareMetadataKeyIdModelProperty(OperationDeclarer operation,
+                                                 Optional<OutputResolverModelParser> outputResolverModelParser,
+                                                 List<InputResolverModelParser> inputResolverModelParsers,
+                                                 Optional<KeyIdResolverModelParser> keyIdResolverModelParser) {
+    if (keyIdResolverModelParser.isPresent()) {
+      String parameterName = keyIdResolverModelParser.get().getParameterName();
+      MetadataType metadataType = keyIdResolverModelParser.get().getMetadataType();
+      String categoryName = JavaModelLoaderUtils.getCategoryName(keyIdResolverModelParser.orElse(null), inputResolverModelParsers,
```
SebaElizalde, line 202: Static import for utils methods

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/utils/JavaModelLoaderUtils.java
```diff
@@ -105,11 +128,140 @@ public static boolean isRoute(ExtensionParameter parameter) {
...
+   * @param keyIdResolverModelParser  parser with the key id metadata
+   *
+   * @return the category name of the resolvers if present, null otherwise
+   *
+   * @since 4.5.0
+   */
+  public static String getCategoryName(KeyIdResolverModelParser keyIdResolverModelParser,
+                                       List<InputResolverModelParser> inputResolverModelParsers,
+                                       OutputResolverModelParser outputResolverModelParser) {
+    if (keyIdResolverModelParser != null && keyIdResolverModelParser.hasKeyIdResolver()) {
```
SebaElizalde, line 257: anyway we can refactor to avoid null checks?

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/java/JavaDeclarationDelegateTestCase.java
Bardammu: The test was wrong. Because the metadata enricher is run after the declaration this wasn't noticed. Once I changed the enricher logic was moved to the loaders the test started to fail

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/delegate/OperationModelLoaderDelegate.java
Bardammu: There are two methods with the same name, I can't use static import here

SebaElizalde (approved):

PR #11838, opened 2022-10-05
W-11793077: Add First Successful number of attempt in each attempt span and refactor Until Successful attempt span addition

sofiamorseletto:


modules/core-components/src/main/java/org/mule/runtime/core/internal/routing/ProcessorRoute.java
```diff
@@ -69,11 +73,18 @@ public void setMessagingExceptionHandler(FlowExceptionHandler messagingException
...
+  public void setSpanCustomizationInfo(SpanCustomizationInfo spanCustomizationInfo) {
+    this.spanCustomizationInfo = spanCustomizationInfo;
+  }
+
   @Override
   public void initialise() throws InitialisationException {
     if (processor instanceof MessagingExceptionHandlerAware) {
       ((MessagingExceptionHandlerAware) processor).setMessagingExceptionHandler(messagingExceptionHandler);
     }
+    if (processor instanceof SpanCustomizationInfoAware) {
```
sofiamorseletto, line 85: Not a fan of having to create an interface to add the SpanCustomizationInfo to the processor, but couldn't find another way, since at that point `processor` is an instance of `LazyProcessorChainBuilder` which is a private class, child of `AbstractMessageProcessorChain` which is a package private class so that was the only way I could think of that would allow me to call `setSpanCustomizationInfo`. I accept other suggestions if they come to mind

sofiamorseletto:
--validate

modules/core-components/src/main/java/org/mule/runtime/core/internal/routing/ProcessorRoute.java
fsgonz: Neither do I. I see why it is needed.

core/src/main/java/org/mule/runtime/core/privileged/processor/chain/DefaultMessageProcessorChainBuilder.java
```diff
@@ -304,14 +305,19 @@ public interface MessagingExceptionHandlerAware {
     void setMessagingExceptionHandler(FlowExceptionHandler messagingExceptionHandler);
   }
 
+  public interface SpanCustomizationInfoAware {
```
fsgonz, line 308: I would set this as a not inner interface. Add a jdoc when it is in its own file.

fsgonz (changes requested):
Rename ticket to involve all the attempts connectors (and not just first successful)

fsgonz (approved):

PR #11853, opened 2022-10-11
W-11887741: StackOverFlow error during TypeWrapper initialization for Enum Type

SebaElizalde:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/TypeWrapper.java
```diff
@@ -76,7 +76,7 @@ public TypeWrapper(ResolvableType resolvableType, ClassTypeLoader typeLoader) {
       resolvableTypeGenerics = resolvableType.getGenerics();
       for (ResolvableType type : resolvableTypeGenerics) {
         TypeWrapper concreteType;
-        if (type == resolvableType) {
+        if (type == resolvableType || (type.isAssignableFrom(resolvableType) && resolvableType.isAssignableFrom(type))) {
```
marianogonzalez, line 79: are enums the only use case for this? if so, why not replacng the second condition with one that checks for an enum?

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/TypeWrapper.java
SebaElizalde: it affects more uses than just enums, added another test case to show this

marianogonzalez:
approved

SebaElizalde:
--validate

PR #11861, opened 2022-10-13
W-11861058: Allow obtaining the failing component from a PipelineMessageNotification

marianorean:


core-tests/src/test/java/org/mule/runtime/core/internal/processor/strategy/processor/PipelineMessageNotificationTestCase.java
```diff
@@ -287,13 +287,19 @@ public boolean matches(Notification argument) {
       EnrichedServerNotification notification = (EnrichedServerNotification) argument;
       Exception exception = notification.getException();
 
-      boolean result =
+      boolean result = true;
+      if (notification instanceof PipelineMessageNotification && exception instanceof MessagingException) {
```
IvanAndresFritzler, line 291: Why do you need to check for the instances?

IvanAndresFritzler (changes requested):

core-tests/src/test/java/org/mule/runtime/core/internal/processor/strategy/processor/PipelineMessageNotificationTestCase.java
marianorean: The one for the exception is not needed, but notification can be an ErrorHandlerNotification

IvanAndresFritzler (approved):

PR #11862, opened 2022-10-13
W-11625237: Application deployment failure with Couldn't find resourc…

alopezbizon:
…e: .dwl neither on classpath or in file system

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
@@ -202,7 +202,7 @@ public abstract class ComponentMessageProcessor<T extends ComponentModel> extend
   protected CompletableComponentExecutor componentExecutor;
   protected ReturnDelegate returnDelegate;
   protected PolicyManager policyManager;
-  protected ClassLoader classLoader;
+  protected ClassLoader nestedChainClassLoader;
   private Optional<TransactionConfig> transactionConfig;
 
   @Inject
@@ -266,7 +266,7 @@ public abstract class ComponentMessageProcessor<T extends ComponentModel> extend
     this.policyManager = policyManager;
     this.retryPolicyTemplate = retryPolicyTemplate;
     this.nestedChain = nestedChain;
-    this.classLoader = classLoader;
+    this.nestedChainClassLoader = classLoader;
     this.reflectionCache = reflectionCache;
     this.resultTransformer = resultTransformer;
     this.hasNestedChain = hasNestedChain(componentModel);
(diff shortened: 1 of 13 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -200,6 +202,7 @@
   protected CompletableComponentExecutor componentExecutor;
   protected ReturnDelegate returnDelegate;
   protected PolicyManager policyManager;
+  protected ClassLoader classLoader;
```
fsgonz, line 205: can we add a more declarative name for this classloader (to avoid aliasing too). For example classloaderToNestedChain.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
```diff
@@ -898,14 +903,25 @@ protected boolean isAsyncExecutableBasedOn(Optional<ConfigurationInstance> stati
     return retryPolicyResolver.apply(staticConfig).isEnabled();
   }
 
+  protected void startIfNeededNestedChain() throws MuleException {
```
fsgonz, line 906: Add a comment indicated why this classloader swap is needed.

fsgonz: and jdoc as this is protected.

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/ComponentMessageProcessorClassLoadingTestCase.java
```diff
@@ -0,0 +1,126 @@
...
+import org.mule.tck.junit4.AbstractMuleContextTestCase;
+
+import io.qameta.allure.Description;
+import io.qameta.allure.Issue;
+import org.junit.After;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ComponentMessageProcessorClassLoadingTestCase extends AbstractMuleContextTestCase {
```
fsgonz, line 48: allure annotations.

fsgonz (changes requested):

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/ComponentMessageProcessorClassLoadingTestCase.java
alopezbizon: Which allure annotations should I add here? I added the `Description` and the `Issue` annotations to the test method.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
alopezbizon: Done.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessor.java
alopezbizon: Javadoc and clarification added.

fsgonz (approved):

PR #11895, opened 2022-10-24
W-11929632: Fixing race conditions in InstanceLockGroup

eze210:


eze210:
--validate

core-tests/src/test/java/org/mule/runtime/core/internal/lock/InstanceLockGroupTestCase.java
```diff
@@ -57,6 +66,65 @@ public void testWhenSeveralLockOneUnlockThenDestroy() throws Exception {
     lockUnlockThenDestroy(5);
   }
 
+  @Test
+  @Issue("W-11929632")
+  public void disposeDoesNotLossReferenceToTakenLocks() throws InterruptedException {
```
fsgonz, line 71: disposeDoesNotLossReferenceToTakenLocks -> disposeDoesNotLoseReferenceToTakenLocks(

fsgonz (changes requested):

core/src/main/java/org/mule/runtime/core/internal/lock/MuleLockFactory.java
```diff
@@ -41,12 +46,25 @@ public void dispose() {
...
   }
 
   @Inject
   @Named(OBJECT_LOCK_PROVIDER)
   public void setLockProvider(LockProvider lockProvider) {
     this.lockProvider = lockProvider;
   }
+
+  private LockGroup createLockGroup() {
+    if (muleContext == null) {
```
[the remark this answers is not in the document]
eze210: This class is used in batch module, and it doesn't inject the dependencies, but only calls the setter for the lock provider (see BatchLockFactory). We would need a change in that module before removing these null checks.

Added the task to investigate the dependency injection problem.

core/src/main/java/org/mule/runtime/core/internal/lock/InstanceLockGroup.java
```diff
@@ -6,39 +6,45 @@
  */
 package org.mule.runtime.core.internal.lock;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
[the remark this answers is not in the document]
eze210: Removed, and also reordered the imports

core-tests/src/test/java/org/mule/runtime/core/internal/lock/TestLockProviderWrapper.java
```diff
@@ -0,0 +1,116 @@
...
+    LockWrapper spiedLock = spy(wrappedLock);
+    providedLocks.put(lockId, spiedLock);
+    return spiedLock;
+  }
+
+  public Lock getSpiedLock(String lockId) {
+    return providedLocks.get(lockId);
+  }
+
+  public void makeLocksToRaiseExceptions() {
```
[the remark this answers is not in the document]
eze210: thanks

core-tests/src/test/java/org/mule/runtime/core/internal/lock/InstanceLockGroupTestCase.java
eze210: thanks

core-tests/src/test/java/org/mule/runtime/core/internal/lock/InstanceLockGroupTestCase.java
```diff
@@ -57,6 +66,65 @@ public void testWhenSeveralLockOneUnlockThenDestroy() throws Exception {
...
+  @Test
+  @Issue("W-11929632")
+  public void disposeDoesNotLoseReferenceToTakenLocks() throws InterruptedException {
+    String testLockId = "TestLockId";
+    instanceLockGroup.lock(testLockId);
+    verify(lockProvider.getSpiedLock(testLockId)).lock();
+
+    // Spawn a task that disposes the lock group and wait some time to ensure that it actually started.
+    executor.submit(instanceLockGroup::dispose);
+    sleep(100);
```
asanguinetti, line 78: Honestly, the potential for flakiness from this worries me a little bit, but I'm not sure how we could avoid it without adding stuff to the `InstanceLockGroup` that would otherwise be unnecessary for productive code.
I'm worried because the delay needs to fall exactly in the sweet spot between two events: the call to `wait` inside the `dispose` method (because otherwise you wouldn't be testing anything) and the timeout expiration (because otherwise the assertion would fail).
What if we grabbed the `Thread` reference from the `Runnable` and then, on the main thread, add a probe which polls for the `Thread`'s state until it is waiting.
Then if it never gets to a waiting state after a reasonable period of time, which can be considerably bigger than the default timeout, we fail. The rationale would be that, if there are pending locks, the thread trying to run through `dispose` should wait (until the timeout at least).
It wouldn't be entirely bullet-proof but should be more robust, what do you think?

core/src/main/java/org/mule/runtime/core/internal/lock/InstanceLockGroup.java
```diff
@@ -151,8 +128,55 @@ public boolean hasPendingLocks() {
 
   @Override
   public void dispose() {
+    waitForLocksToBeUnlocked();
+  }
```
asanguinetti, line 132: We are no longer clearing up the Map (not even after the timeout expires), is this intentional?

asanguinetti (commented):
I'm just adding a few suggestions/concerns, not really requesting changes yet, but can't approve yet either. I would like to hear your opinion on some of the comments.

Also, I'm leaving a general comment with an alternative for consideration:
What if instead of having a `LockEntry` with `incrementLockCount` and a `decrementLockCount` method we had it implementing `Lock` as a wrapper, which would increment/decrement the counter alongside the corresponding call to the delegate, that way we could avoid the potential race condition in all the cases in which we are incrementing/decrementing -> releasing the lockAccessMonitor -> doing the actual lock/unlock. We could maybe even get rid of the lockAccessMonitor entirely and use a `ConcurrentHashMap`, but we would still need a way to notify the disposer when the map becomes empty.

Another option which I think is even better, and we already discussed, would be to not use the `InstanceLockGroup` at all. Just by making the `LockFactory` delegate on the `LockProvider` directly and that's it.

core/src/main/java/org/mule/runtime/core/internal/lock/InstanceLockGroup.java
eze210: Yes, clearing the map would generate the same bug after the graceful shutdown timeout

core/src/main/java/org/mule/runtime/core/internal/lock/InstanceLockGroup.java
```diff
@@ -151,8 +129,55 @@ public boolean hasPendingLocks() {
...
+            isTimeout = true;
+          }
+        }
+      } catch (InterruptedException e) {
+        currentThread().interrupt();
+      }
+      if (!locks.isEmpty()) {
+        LOGGER.warn("These locks weren't unlocked before disposing its lock group: {}", locks.keySet());
+      }
     }
```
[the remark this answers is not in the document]
eze210: Refactored to something similar (notice that the pseudocode uses the remainingMillis)

core-tests/src/test/java/org/mule/runtime/core/internal/lock/InstanceLockGroupTestCase.java
eze210: Good idea, done

fsgonz (approved):

asanguinetti (approved):

PR #11896, opened 2022-10-24
W-11412423: Automatically calculate components min mule version depending on the features it uses

SebaElizalde:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/SourceModelParser.java
```diff
@@ -152,6 +150,8 @@ public interface SourceModelParser extends SemanticTermsParser, StereotypeModelP
 
   Optional<SinceMuleVersionModelProperty> getSinceMuleVersionModelProperty();
 
+  Optional<String> getMinMuleVersion();
```
ndinu, line 153: With these changes, the parser will now be able to parse the min mule version of the components. Is there any issue created to add this notion to the model and make the model loader populate it?

SebaElizalde: This issue https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000lfXIeYAM/view handles comparing the parsed min mule version against the pom (this would be the model loader responsibility).
While https://gus.lightning.force.com/lightning/r/a07EE00000lfQ0mYAE/view will explore how to add this information to the model (if via model property, or as a new element, etc.)

ndinu: Got it 👌  , I think that the last part should be covered by https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000011Dq4hYAC/view this issue, which was closed based on the assumption that we would use the SinceMuleVersionModelProperty. Don't know if we should re open that one, or create a follow-up issue.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/MethodWrapper.java
```diff
@@ -143,6 +145,12 @@ public <A extends Annotation> Optional<AnnotationValueFetcher<A>> getValueFromAn
         : empty();
   }
 
+  @Override
+  public List<Type> getAnnotations() {
+    return Arrays.stream(method.getAnnotations()).map(ann -> new TypeWrapper(ann.annotationType(), typeLoader))
```
ndinu, line 150: Can we add some unit tests for these new methods ?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/ConfigurationModelParser.java
```diff
@@ -96,4 +95,6 @@ public interface ConfigurationModelParser extends StereotypeModelParser, Additio
   Optional<DeprecationModel> getDeprecationModel();
 
   Optional<SinceMuleVersionModelProperty> getSinceMuleVersionModelProperty();
+
+  Optional<String> getMinMuleVersion();
```
ndinu, line 99: Add jdoc explaining that does the version represents.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/SourceModelParser.java
SebaElizalde: I think we should create a new one, after designing the solution in the spec.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/ConfigurationModelParser.java
ndinu, line 99: What do you think of returning Optional<MuleVersion> ?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,149 @@
...
+import static java.lang.String.format;
+import static java.util.Arrays.asList;
+import static java.util.Optional.of;
+import static org.mule.sdk.api.util.SdkApiConstants.NON_ENFORCE_MIN_MULE_VERSION_CLASSES;
+import static org.slf4j.LoggerFactory.getLogger;
+
+public class JavaParserUtils {
+
+  private static final Logger LOGGER = getLogger(JavaParserUtils.class);
+  private static final String firstMuleVersion = "4.1.0";
```
ndinu, line 35: I think first mule version is 4.1.1

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,149 @@
...
+  private static String calculateFromClass(Type clazz, Set<String> seenTypes) {
+    // TODO W-11938731 and W-11955483
+    if (clazz.getTypeName().startsWith("org.mule.extension.http") || clazz.getTypeName().startsWith("com.google.common")
+        || clazz.getTypeName().startsWith("java.")
+        || primitiveTypeNames.contains(clazz.getTypeName()) || seenTypes.contains(clazz.getTypeName())) {
+      return firstMuleVersion;
+    } else {
+      seenTypes.add(clazz.getTypeName());
+    }
+    // TODO: get MMV from pom
```
ndinu, line 62: Point to an actual gus issue

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,149 @@
...
+    return calculateFromMethod(method, new HashSet<>());
+  }
+
+  public static String calculateFromParameter(ExtensionParameter parameter) {
+    return calculateFromField(parameter, new HashSet<>());
+  }
+
+  private static String calculateFromClass(Type clazz, Set<String> seenTypes) {
+    // TODO W-11938731 and W-11955483
+    if (clazz.getTypeName().startsWith("org.mule.extension.http") || clazz.getTypeName().startsWith("com.google.common")
```
ndinu, line 55: why do we need all these base cases? google and http ones seems odd.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,149 @@
...
+    // TODO: get MMV from pom
+    String calculatedMMV = firstMuleVersion;
+    // Look for the annotation at the class level
+    Optional<String> classLevelMMV = getMinMuleVersion(clazz);
+    if (classLevelMMV.isPresent()) { // We cut the algorithm short (only compare later against generics)
+      calculatedMMV = maxMMV(calculatedMMV, classLevelMMV.get());
+    } else { // We calculate from class inspection
+      // Parse the superClass
+      calculatedMMV =
+          maxMMV(calculatedMMV, clazz.getSuperType().map(type -> calculateFromClass(type, seenTypes)).orElse("4.1.0"));
```
ndinu, line 71: Use the constant defined above

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaParserUtilsTestCase.java
```diff
@@ -0,0 +1,231 @@
...
+
+public class JavaParserUtilsTestCase {
+
+  @Rule
+  public ExpectedException expectedException = none();
+
+  @Test
+  public void javaLangClassReturns410() {
+    Type clazz = new TypeWrapper(String.class, TYPE_LOADER);
+    assertThat(calculateFromClass(clazz), is("4.1.0"));
```
ndinu, line 43: Use constants

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaOperationModelParser.java
```diff
@@ -406,6 +408,11 @@ public Optional<SinceMuleVersionModelProperty> getSinceMuleVersionModelProperty(
     return JavaExtensionModelParserUtils.getSinceMuleVersionModelProperty(operationElement);
   }
 
+  @Override
+  public Optional<String> getMinMuleVersion() {
+    return of(calculateFromMethod(operationElement));
```
ndinu, line 413: Checking the implementation, I think that this does not take into account if the configuration/extension is annotationed with the sdk-api @Operation (which may raise the minMuleVersion)

ndinu, line 413: Is checking the method enough here? Maybe there is something at class level / field level that needs to be checked?

ndinu (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/SourceModelParser.java
ndinu: I feel that the "where to put that info in the model" is out of the scope of the pruning spec.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/FieldWrapper.java
```diff
@@ -104,6 +108,15 @@ public <A extends Annotation> Optional<AnnotationValueFetcher<A>> getValueFromAn
         : empty();
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public List<Type> getAnnotations() {
+    return Arrays.stream(field.getAnnotations()).map(ann -> new TypeWrapper(ann.annotationType(), typeLoader))
```
marianogonzalez, line 116: static import

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/MethodWrapper.java
marianogonzalez: static import

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/TypeWrapper.java
```diff
@@ -305,4 +313,14 @@ public Optional<MethodElement> getMethod(String name, Class<?>... objects) {
...
   }
+
+  @Override
+  public List<MethodElement> getEnclosingMethods() {
+    return Arrays.stream(aClass.getDeclaredMethods()).map(method -> new MethodWrapper(method, typeLoader)).collect(toList());
+  }
+
+  @Override
+  public List<Type> getImplementingInterfaces() {
+    return Arrays.stream(aClass.getInterfaces()).map(clazz -> new TypeWrapper(clazz, typeLoader)).collect(toList());
```
marianogonzalez, line 324: static import

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/ConnectionProviderModelParser.java
```diff
@@ -95,4 +95,6 @@ public interface ConnectionProviderModelParser extends SemanticTermsParser, Ster
   Optional<DisplayModel> getDisplayModel();
 
   Optional<SinceMuleVersionModelProperty> getSinceMuleVersionModelProperty();
+
+  Optional<String> getMinMuleVersion();
```
marianogonzalez, line 99: jdoc on every one of these added methods

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,149 @@
...
+import static org.slf4j.LoggerFactory.getLogger;
+
+public class JavaParserUtils {
+
+  private static final Logger LOGGER = getLogger(JavaParserUtils.class);
+  private static final String firstMuleVersion = "4.1.0";
+
+  private static final Set<String> primitiveTypeNames =
+      new HashSet<>(asList("byte", "short", "int", "long", "float", "double", "char", "boolean", "void"));
+  private static final List<Class<?>> denyList = NON_ENFORCE_MIN_MULE_VERSION_CLASSES;
```
marianogonzalez, line 39: naming convention

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
marianogonzalez, line 35: naming convention

marianogonzalez: correct

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
marianogonzalez, line 55: why would this ever be called for a google package?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
marianogonzalez, line 62: add issue

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,149 @@
...
+      // Parse the class fields
+      for (FieldElement field : clazz.getFields()) {
+        calculatedMMV = maxMMV(calculatedMMV, calculateFromField(field, seenTypes));
+      }
+      // Parse the class methods
+      for (MethodElement<?> method : clazz.getEnclosingMethods()) {
+        calculatedMMV = maxMMV(calculatedMMV, calculateFromMethod(method, seenTypes));
+      }
+    }
+    // Finally we always parse the class generics
```
marianogonzalez, line 89: why?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
marianogonzalez, line 55: I think you're treating this as separate bugs when in reality these might lead to the fact that you're considering things that shouldn't be.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,149 @@
...
+    } else {
+      seenTypes.add(clazz.getTypeName());
+    }
+    // TODO: get MMV from pom
+    String calculatedMMV = firstMuleVersion;
+    // Look for the annotation at the class level
+    Optional<String> classLevelMMV = getMinMuleVersion(clazz);
+    if (classLevelMMV.isPresent()) { // We cut the algorithm short (only compare later against generics)
+      calculatedMMV = maxMMV(calculatedMMV, classLevelMMV.get());
+    } else { // We calculate from class inspection
```
marianogonzalez, line 68: i don't think this is correct. If the user annotated the class with `@MinMuleVersion`, then for sure we should never resolve to a lower version. However, you still need to introspect fields, interfaces, etc. to make sure that the user is not forcing this to a LOWER version than it's really needed (in which case, this should be reported as a Problem)

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,149 @@
...
+    for (TypeGeneric typeGeneric : clazz.getGenerics()) {
+      calculatedMMV = maxMMV(calculatedMMV, calculateFromClass(typeGeneric.getConcreteType(), seenTypes));
+    }
+    return calculatedMMV;
+  }
+
+  private static String calculateFromMethod(MethodElement<?> method, Set<String> seenTypes) {
+    Optional<String> minMuleVersionAnnotation = getMinMuleVersionFromAnnotation(method);
+    if (minMuleVersionAnnotation.isPresent()) {
+      return minMuleVersionAnnotation.get();
```
marianogonzalez, line 99: same as above. Respect the hint but also validate it

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+    String MaxMMV = firstMuleVersion;
```
marianogonzalez, line 101: camel case

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/loader/java/type/Type.java
```diff
@@ -169,4 +169,16 @@ default List<Type> getInterfaceGenerics(Class interfaceType) {
    * @since 4.2
    */
   Optional<MethodElement> getMethod(String name, Class<?>... parameterTypes);
+
+  /**
+   * @return All the {@link MethodElement}s present in the class
+   * @since 4.5
+   */
+  List<MethodElement> getEnclosingMethods();
```
marianogonzalez, line 177: per the usage it seems like these are always either iterated or transformed. Better to just return the `Stream<T>` to avoid the unnecesary `collect(toList())`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,149 @@
...
+import static java.util.Optional.of;
+import static org.mule.sdk.api.util.SdkApiConstants.NON_ENFORCE_MIN_MULE_VERSION_CLASSES;
+import static org.slf4j.LoggerFactory.getLogger;
+
+public class JavaParserUtils {
+
+  private static final Logger LOGGER = getLogger(JavaParserUtils.class);
+  private static final String firstMuleVersion = "4.1.0";
+
+  private static final Set<String> primitiveTypeNames =
```
marianogonzalez, line 37: naming

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,149 @@
...
+      new HashSet<>(asList("byte", "short", "int", "long", "float", "double", "char", "boolean", "void"));
+  private static final List<Class<?>> denyList = NON_ENFORCE_MIN_MULE_VERSION_CLASSES;
+
+  public static String calculateFromClass(Type clazz) {
```
marianogonzalez, line 41: bad method signature:
* Name: Calculate what exactly?
* Return type: Why a String and not a MuleVersion?

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/api/loader/java/type/WithAliasTestCase.java
```diff
@@ -67,6 +71,12 @@ public <A extends Annotation> Optional<AnnotationValueFetcher<A>> getValueFromAn
           : empty();
     }
 
+    @Override
+    public List<Type> getAnnotations() {
+      return Arrays.stream(getClass().getAnnotations()).map(ann -> new TypeWrapper(ann.annotationType(), TYPE_LOADER))
```
marianogonzalez, line 76: static import

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/api/loader/java/type/WithAliasTestCase.java
```diff
@@ -110,6 +126,12 @@ public <A extends Annotation> Optional<AnnotationValueFetcher<A>> getValueFromAn
...
```
marianogonzalez, line 131: static import

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
marianogonzalez, line 41: I don't think neither of these are calculating from, they're calculating **for**.

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,385 @@
...
+    }
+    return getMinMuleVersionFromAnnotations(type).orElse(FIRST_MULE_VERSION);
+  }
+
+  private static Optional<MuleVersion> getMinMuleVersionFromAnnotations(WithAnnotations type) {
+    return type.getValueFromAnnotation(MinMuleVersion.class)
+        .map(fetcher -> new MuleVersion(fetcher.getStringValue(MinMuleVersion::value)));
+  }
+
+  public static MuleVersion maxMMV(MuleVersion currentMax, MuleVersion candidate) {
```
marianogonzalez, line 367: rename to simply `max`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,385 @@
...
+    }
+    return candidate;
+  }
+
+  private static boolean belongsToSdkPackages(String fullyQualifiedName) {
+    return belongsToExtensionsApiPackages(fullyQualifiedName) || belongsToSdkApiPackages(fullyQualifiedName);
+  }
+
+  private static boolean belongsToExtensionsApiPackages(String fullyQualifiedName) {
+    return fullyQualifiedName.startsWith("org.mule.runtime.extension");
```
marianogonzalez, line 379: shouldn't it also include the `.api` ?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,385 @@
...
+  }
+
+  private static boolean belongsToSdkApiPackages(String fullyQualifiedName) {
+    return fullyQualifiedName.startsWith("org.mule.sdk");
```
marianogonzalez, line 383: shouldn't it also include the `.api` ?

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,385 @@
...
+    for (Type annotation : containerType.getAnnotations()) {
+      calculatedMMV = maxMMV(calculatedMMV, getEnforcedMinMuleVersion(annotation));
+    }
+    for (FieldElement field : containerType.getFields()) {
+      calculatedMMV = maxMMV(calculatedMMV, calculateFieldMinMuleVersion(field));
+    }
+    return calculatedMMV;
+  }
+
+  private static List<MethodElement> getAnnotatedSourceMethods(Type source) {
```
marianogonzalez, line 339: why is this method necessary? Why not just analysing all the methods in the source, which would also caught methods from other interfaces. Another con of this method is that it needs changes each time we add a new feature or annotation

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/loader/java/type/WithAnnotations.java
```diff
@@ -48,4 +49,11 @@ public interface WithAnnotations {
   default boolean isAnnotatedWith(Class<? extends Annotation> annotation) {
     return getAnnotation(annotation).isPresent();
   }
+
+  /**
+   * Returns all the annotations present within the component as a {@link Type}
+   *
+   * @return A {@code List} of the annotations as a {@link Type}
+   */
+  List<Type> getAnnotations();
```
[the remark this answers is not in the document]
SebaElizalde: Why? Is it a performance issue? Because the other interfaces "WithXXX" all return Lists so I went that way to keep it consistent too.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaConfigurationModelParser.java
```diff
@@ -185,4 +194,15 @@ public Optional<DeprecationModel> getDeprecationModel() {
   public Optional<SinceMuleVersionModelProperty> getSinceMuleVersionModelProperty() {
     return JavaExtensionModelParserUtils.getSinceMuleVersionModelProperty(configElement);
   }
+
+  @Override
+  public Optional<MuleVersion> getMinMuleVersion() {
+    MuleVersion propagatedMinMuleVersion = FIRST_MULE_VERSION;
+    List<Type> sdkConfigurations = extensionElement.getValueFromAnnotation(Configurations.class)
+        .map(vf -> vf.getClassArrayValue(Configurations::value)).orElse(emptyList());
+    if (sdkConfigurations.stream().anyMatch(s -> s.getTypeName().equals(configElement.getTypeName()))) {
```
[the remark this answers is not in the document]
SebaElizalde: Rather be names?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,385 @@
...
+    MuleVersion calculatedMMV = FIRST_MULE_VERSION;
+    if (extension.isAnnotatedWith(Extension.class)) {
+      calculatedMMV = maxMMV(calculatedMMV, new MuleVersion(Extension.class.getAnnotation(MinMuleVersion.class).value()));
+    }
+    if (!(extension.isAnnotatedWith(Configurations.class)
+        || extension.isAnnotatedWith(org.mule.sdk.api.annotation.Configurations.class))) {
+      for (FieldElement field : extension.getFields()) {
+        calculatedMMV = maxMMV(calculatedMMV, calculateFieldMinMuleVersion(field));
+      }
+      for (MethodElement<?> method : extension.getEnclosingMethods()) {
```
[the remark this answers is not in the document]
SebaElizalde: No

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,385 @@
...
+                         finalCalculatedMMV, mmv));
+        return finalCalculatedMMV;
+      } else {
+        return mmv;
+      }
+    }).orElse(calculatedMMV);
+  }
+
+  public static MuleVersion calculateConfigMinMuleVersion(Type config, MuleVersion propagatedMinMuleVersion) {
+    MuleVersion calculatedMMV = maxMMV(propagatedMinMuleVersion, config.getSuperType()
```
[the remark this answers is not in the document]
SebaElizalde: I don't understand, the processing of the current type continues below this line.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/loader/java/type/WithAnnotations.java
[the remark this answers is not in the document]
marianogonzalez: performance and convenience. If you look at the implementations, you always generate a stream and then collect it to a list, only to either iterate, filter, reduce or a combination of the above. Returning the Stream makes much more sense

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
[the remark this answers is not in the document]
marianogonzalez: I believe it has to

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaConfigurationModelParser.java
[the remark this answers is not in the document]
marianogonzalez: is this about checking for the case in which the `@Extension` class is also acting as a configuration? In that case, I think that you wouldn't find that in the `@Configurations` annotation as that is done implicitly.

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,385 @@
...
+                         finalCalculatedMMV, mmv));
+        return finalCalculatedMMV;
+      } else {
+        return mmv;
+      }
+    }).orElse(calculatedMMV);
+  }
+
+  public static MuleVersion calculateSourceMinMuleVersion(Type source) {
+    MuleVersion calculatedMMV = source.getSuperType().map(type -> {
```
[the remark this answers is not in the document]
SebaElizalde: You mean looking at the superType is unnecessary? Or you mean the next line for the isSameType?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,385 @@
...
+      calculatedMMV = getEnforcedMinMuleVersion(outputType);
+    }
+    return calculatedMMV;
+  }
+
+  private static MuleVersion calculateFieldMinMuleVersion(ExtensionParameter field) {
+    Type parameterType = field.getType();
+    MuleVersion calculatedMMV = FIRST_MULE_VERSION;
+    for (Type annotation : field.getAnnotations()) {
+      if (annotation.isSameType(MinMuleVersion.class)) {
```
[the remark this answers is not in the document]
SebaElizalde: Suppose the following example:
```
@Parameter
@MinMuleVersion(4.6)
String someField;
```
When moving from 4.5 to 4.6 the app would break as now there is a mandatory parameter.
So why would we allow it? What use case do we want to support?

I do agree that the validation should happen elsewhere, so removing it for now.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaConfigurationModelParser.java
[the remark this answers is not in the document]
SebaElizalde: No, this is checking for a given configuration if it was included inside a @extension.api.Configuration({...}) or sdk.api.Configurations({...}), cos if it is the second case then the configuration will only appear in mule 4.5+ thus it must inherit 4.5 mmv.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
SebaElizalde: Was trying to be more surgical, but agreed that better parse all methods.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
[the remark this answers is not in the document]
SebaElizalde: Do we support that case where an Extension extends another Extension? If yes, better call this method recursively with the super class (like the other methods) rather than get the inherited methods.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/loader/java/type/WithAnnotations.java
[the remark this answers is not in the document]
SebaElizalde: okay

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/loader/java/type/Type.java
marianogonzalez, line 177: Stream

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/loader/java/type/Type.java
```diff
@@ -169,4 +169,16 @@ default List<Type> getInterfaceGenerics(Class interfaceType) {
...
+
+  /**
+   * @return All the Interfaces this class implements as {@link Type}s
+   * @since 4.5
+   */
+  List<Type> getImplementingInterfaces();
```
marianogonzalez, line 183: Stream

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -0,0 +1,436 @@
...
+import static java.util.Collections.emptyList;
+import static java.util.stream.Collectors.toList;
+import static org.slf4j.LoggerFactory.getLogger;
+
+/**
+ * Utils class to calculate {@link MinMuleVersion} for SDK Components.
+ *
+ * @since 4.5
+ */
+public class JavaParserUtils {
```
marianogonzalez, line 54: make final , add private constructor

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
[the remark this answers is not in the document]
marianogonzalez: We do support that case, but even if we didn't we always need to  recursively evaluate parent classes. These extensions are written in Java and we need to assume that the author made full use of the language capabilities. Sources, operations, configurations and even the extension classes can have parent classes with common stuff. The cloud connectors team does that for once.

marianogonzalez (changes requested):

marianogonzalez (changes requested):

SebaElizalde:
--validate

marianogonzalez (dismissed):

SebaElizalde:
--validate

SebaElizalde:
--validate

ndinu (approved):

PR #11908, opened 2022-10-27
W-11969914: Spans attributes and name are not overridden in sources

fsgonz:


core/src/main/java/org/mule/runtime/core/internal/execution/SourceResultAdapter.java
```diff
@@ -151,4 +158,22 @@ public Optional<PollItemInformation> getPollItemInformation() {
   public DistributedTraceContextGetter getDistributedTraceContextGetter() {
     return distributedTraceContextGetter;
   }
+
+  /**
+   * @return the root name. This will override the name of the flow associated to the source.
```
sofiamorseletto, line 163: The name of the flow or the name of the span associated to the source?

core/src/main/java/org/mule/runtime/core/internal/trace/DistributedTraceContext.java
```diff
@@ -65,6 +65,22 @@ public interface DistributedTraceContext extends CurrentSpanAware {
...
 
+  /**
+   * Set the initial overridden span name
+   * 
+   * @param name the name to set.
+   */
+  void setRootSpanName(String name);
+
+  /**
+   * @return set the initial overridden span nam.
```
sofiamorseletto, line 76: name*

sofiamorseletto (approved):

PR #11937, opened 2022-11-09
W-11858790: Add attribute assertions and exception propagation assert…

sofiamorseletto:
…ions to SpanTestHierarchy


tests/infrastructure/src/main/java/org/mule/test/infrastructure/profiling/tracing/SpanTestHierarchy.java
```diff
@@ -25,6 +35,20 @@
  */
 public class SpanTestHierarchy {
 
+  public static final String ARTIFACT_TYPE_KEY = "artifactType";
```
fsgonz, line 38: I would move the original keys to some public location in the case of the errors.

tests/infrastructure/src/main/java/org/mule/test/infrastructure/profiling/tracing/SpanTestHierarchy.java
```diff
@@ -107,11 +154,52 @@ private boolean hasCorrectParent(CapturedExportedSpan span, String expectedParen
     return parentSpan.getName().equals(expectedParentName);
   }
 
+  private boolean hasCorrectLocation(CapturedExportedSpan span, String expectedLocation) {
+    return expectedLocation == null || span.getAttributes().get(LOCATION_KEY).equals(expectedLocation);
+  }
+
+  private void assertAttributes(CapturedExportedSpan actualSpan, SpanNode expectedNode) {
```
fsgonz, line 161: I think the specific attributes to assert should be defined in the hierarchy yso ou can set some attributes to verify and then verify them according to the expected values. I would do it in a generic way. In the case of error that is different because it is part of the optel standard. But for the other attributes they should be set in the test that uses this hierarchy.

fsgonz (changes requested):

fsgonz (approved):

PR #11948, opened 2022-11-11
W-12036134: Isolated Properties resolution, without using System and Environment Properties

mbuchwald:


modules/properties-config/src/main/java/org/mule/runtime/config/internal/model/properties/PropertiesResolverUtils.java
```diff
@@ -102,6 +102,32 @@ public static PropertiesResolverConfigurationProperties createConfigurationAttri
 
   }
 
+  public static PropertiesResolverConfigurationProperties createIsolatedConfigurationAttributeResolver(ArtifactAst artifactAst,
```
elrodro83, line 105: javadoc

elrodro83 (changes requested):

PR #11958, opened 2022-11-14
W-11949160: return property name when not found to avoid NullPointerE…

jcgrethe:
…xception

### DIFF
--- a/modules/properties-config/src/test/java/org/mule/runtime/config/internal/dsl/model/config/DefaultConfigurationPropertiesResolverTestCase.java
+++ b/modules/properties-config/src/test/java/org/mule/runtime/config/internal/dsl/model/config/DefaultConfigurationPropertiesResolverTestCase.java
@@ -6,6 +6,9 @@
  */
 package org.mule.runtime.config.internal.dsl.model.config;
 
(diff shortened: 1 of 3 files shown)


elrodro83 (changes requested):
add a test case

modules/properties-config/src/test/java/org/mule/runtime/config/internal/dsl/model/config/DefaultConfigurationPropertiesResolverTestCase.java
```diff
@@ -0,0 +1,37 @@
...
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.mockito.Mockito.mock;
+
+import org.mule.runtime.properties.api.ConfigurationPropertiesProvider;
+
+import org.junit.Test;
+
+public class DefaultConfigurationPropertiesResolverTestCase {
```
elrodro83, line 19: allure

elrodro83 (changes requested):

elrodro83 (approved):

PR #11961, opened 2022-11-15
W-12073186: Enable the possibility of parallel extension model loading when deploying applications

marianogonzalez:


core/src/main/java/org/mule/runtime/core/api/util/boot/ExtensionLoaderUtils.java
```diff
@@ -80,5 +82,10 @@ public static Optional<ExtensionModelLoader> getOptionalLoaderById(ClassLoader c
         .findFirst();
   }
 
+  public static boolean isParallelExtensionModelLoadingEnabled() {
+    String property = getProperty(PARALLEL_EXTENSION_MODEL_LOADING_PROPERTY);
+    return property != null ? Boolean.valueOf(property) : false;
```
elrodro83, line 87: `Boolean.getBoolean()`

modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/artifact/ArtifactExtensionManagerFactoryTestCase.java
```diff
@@ -0,0 +1,103 @@
...
+import io.qameta.allure.Story;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@SmallTest
+@RunWith(MockitoJUnitRunner.class)
```
elrodro83, line 41: use the rule instead of the runner

elrodro83 (changes requested):

PR #11964, opened 2022-11-16
W-12065739: Adding system property to disable expressions in config-ref

eze210:


asanguinetti (commented):
We need to also check these two tests:
- [ExpressionsOnConfigRefTestCase](https://github.com/mulesoft/mule/blob/master/modules/extensions-xml-support/src/test/java/org/mule/test/functional/ExpressionsOnConfigRefTestCase.java#L59). Probably working because the XML SDK works differently in many ways. I don't remember but it may be that it doesn't use the enricher, so might need to disable something else.
- [DomainDeploymentWithELServiceTestCase](https://github.com/mulesoft/mule/blob/master/modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/DomainDeploymentWithELServiceTestCase.java#L137). This one is just not running in hawk, that may be the reason you didn't see it fail.

eze210:
--validate

asanguinetti (approved):

PR #11976, opened 2022-11-22
W-10620845: Fix for reserved properties

mbuchwald:


modules/properties-config/src/main/java/org/mule/runtime/config/api/properties/ConfigurationPropertiesHierarchyBuilder.java
```diff
@@ -188,4 +187,26 @@ public ConfigurationPropertiesResolver build() {
     lastResolver.setAsRootResolver();
     return lastResolver;
   }
+
+  @Deprecated
```
eze210, line 191: We usually put the annotations after the javadoc.
Also, add the javadoc for this deprecation

eze210 (changes requested):

eze210 (approved):

mbuchwald:
--validate

PR #11978, opened 2022-11-22
W-12074281: Change ApplicationTypeLoader

ndinu:


core/src/main/java/org/mule/runtime/core/internal/type/catalog/DefaultArtifactTypeLoader.java
```diff
@@ -0,0 +1,134 @@
...
+  private Map<String, Optional<MetadataType>> loadedTypes;
+
+  @Inject
+  private ExtensionManager extensionManager;
+
+  private Collection<ExtensionModel> extensionModels;
+
+  public DefaultArtifactTypeLoader() {
+
+  }
```
asanguinetti, line 61: Just a thought:
How do we feel about using constructor injection instead? Wouldn't it be helpful to prevent instances of this class from being created without either an `extensionManager` or `extensionModels`.

core/src/main/java/org/mule/runtime/core/internal/type/catalog/DefaultArtifactTypeLoader.java
```diff
@@ -0,0 +1,134 @@
...
+
+  public DefaultArtifactTypeLoader(Collection<ExtensionModel> extensionModels) {
+    this.extensionModels = extensionModels;
+  }
+
+  @Override
+  public void initialise() throws InitialisationException {
+    if (extensionModels == null) {
+      extensionModels = extensionManager.getExtensions();
+    }
```
asanguinetti, line 71: Shouldn't we throw an `InitializationException` if `extensionManager` is `null`? It could happen if the initialization is done without injection (and the no-args constructor was used).

asanguinetti (changes requested):
Just a couple of observations and a small request for change due to a typo.

core/src/main/java/org/mule/runtime/core/internal/type/catalog/DefaultArtifactTypeLoader.java
```diff
@@ -0,0 +1,134 @@
...
+    typesByExtension = new ConcurrentHashMap<>();
+    loadedTypes = new ConcurrentHashMap<>();
+    for (ExtensionModel extensionModel : extensionModels) {
+      String extensionPrefix = extensionModel.getXmlDslModel().getPrefix();
+      typesByExtension.put(extensionPrefix, extensionModel.getTypes());
+    }
```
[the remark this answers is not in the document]
ndinu: You are right, I will do this.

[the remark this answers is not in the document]
ndinu: I think you meant the `typesByExtension` map.

core/src/main/java/org/mule/runtime/core/internal/type/catalog/DefaultArtifactTypeLoader.java
ndinu: I like the idea. I will do this 👍

core/src/main/java/org/mule/runtime/core/internal/type/catalog/DefaultArtifactTypeLoader.java
ndinu: I will take the suggestion from above and add a nullCheck in the constructor.

core/src/main/java/org/mule/runtime/core/internal/type/catalog/DefaultArtifactTypeLoader.java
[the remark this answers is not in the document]
asanguinetti: yes, sorry, the `typesByExtension` map.

ndinu:
--validate

core/src/main/java/org/mule/runtime/core/internal/type/catalog/DefaultArtifactTypeLoader.java
```diff
@@ -0,0 +1,147 @@
...
+    if (primitive.isPresent()) {
+      return primitive;
+    }
+
+    Optional<MetadataType> special = specialTypesLoader.load(typeIdentifier);
+    if (special.isPresent()) {
+      return special;
+    }
+
+    return ofNullable(lookupType(typeIdentifier));
```
elrodro83, line 112: Use the optional funtional api

core/src/main/java/org/mule/runtime/core/internal/type/catalog/DefaultArtifactTypeLoader.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  private MetadataType lookupType(String typeIdentifier) {
+    if (!typeIdentifier.contains(":")) {
+      return null;
+    }
+    int separatorIndex = typeIdentifier.indexOf(":");
+    String extensionIdentifier = typeIdentifier.substring(0, separatorIndex);
+    String typeIdOrAlias = typeIdentifier.substring(separatorIndex + 1);
```
elrodro83, line 121: use `String.split` instead of substring

elrodro83 (changes requested):

core/src/main/java/org/mule/runtime/core/internal/type/catalog/DefaultArtifactTypeLoader.java
```diff
@@ -99,23 +99,17 @@ public Optional<MetadataType> load(String typeIdentifier) {
...
-    }
-
-    Optional<MetadataType> special = specialTypesLoader.load(typeIdentifier);
-    if (special.isPresent()) {
-      return special;
-    }
-
-    return ofNullable(lookupType(typeIdentifier));
+    return ofNullable(primitivesTypeLoader.load(typeIdentifier)
+        .orElse(specialTypesLoader.load(typeIdentifier)
```
elrodro83, line 103: `orElseGet`

core/src/main/java/org/mule/runtime/core/internal/type/catalog/DefaultArtifactTypeLoader.java
```diff
@@ -99,23 +99,17 @@ public Optional<MetadataType> load(String typeIdentifier) {
...
+            .orElse(lookupType(typeIdentifier))));
```
elrodro83, line 104: `orElseGet`

elrodro83 (changes requested):

elrodro83 (approved):

asanguinetti (approved):

PR #11983, opened 2022-11-24
W-12118812: Allow dependencies with same group ID and artifact ID but different classifier in DeployableProjectModel

rbourbonnavarro:


modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/api/deployable/DeployableProjectModel.java
```diff
@@ -131,7 +131,8 @@ private Map<String, List<BundleDescriptor>> getRepeatedDependencies(List<BundleD
 
     for (BundleDependency dependency : dependencies) {
       BundleDescriptor descriptor = dependency.getDescriptor();
-      String pluginKey = descriptor.getGroupId() + ":" + descriptor.getArtifactId();
+      String pluginKey =
+          descriptor.getGroupId() + ":" + descriptor.getArtifactId() + ":" + descriptor.getClassifier().orElse("empty");
```
elrodro83, line 135: avoid the `:empty` altogether if there is no classifier

elrodro83 (changes requested):

elrodro83 (approved):

PR #11988, opened 2022-11-25
Change XML SDK implicit config creation system property to true

Bardammu:


modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/extension/xml/MacroExpansionModuleModel.java
```diff
@@ -125,7 +125,7 @@ public class MacroExpansionModuleModel {
    * If true avoid the creation of an implicit configuration
    */
   private final boolean disable_xml_sdk_implicit_configuration_creation =
-      valueOf(getProperty(MULE_DISABLE_XML_SDK_IMPLICIT_CONFIGURATION_CREATION, "false"));
+      valueOf(getProperty(MULE_DISABLE_XML_SDK_IMPLICIT_CONFIGURATION_CREATION, "true"));
```
ndinu, line 128: Would this only be done en 4.4.x? What is the behavior in 4.3.x? Will master still default to false?

modules/extensions-xml-support/src/test/java/org/mule/test/functional/ModuleWithImplicitConfigurationUsingScopesTestCase.java
```diff
@@ -13,9 +13,11 @@
 import org.mule.test.petstore.extension.PetStoreClient;
 
 import io.qameta.allure.Feature;
+import org.junit.Ignore;
 import org.junit.Test;
 
 @Feature(XML_SDK)
+@Ignore("W-12121045: Ignore until we can set MULE_DISABLE_XML_SDK_IMPLICIT_CONFIGURATION_CREATION to false")
```
ndinu, line 20: Shouldn't this tests set the property to false, so that they test the behavior when the new feature is enabled?

modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/extension/xml/MacroExpansionModuleModel.java
Bardammu: Yes, won't be on 4.3 and 4.5. Someone should fix this problem in the next release and set the property to false again

modules/extensions-xml-support/src/test/java/org/mule/test/functional/ModuleWithImplicitConfigurationUsingScopesTestCase.java
Bardammu: The problem is that the extension can be compiled

modules/spring-config/src/main/java/org/mule/runtime/config/internal/dsl/model/extension/xml/MacroExpansionModuleModel.java
ndinu: I think that the issue with this would be that we would be breaking compatibility for someone who is using implicit configs without setting the property.I will talk with the team to see if we can take the issue and fix it before the next cutoff.

PR #11999, opened 2022-11-28
W-12050227: Adding cluster components extension model

eze210:


runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -80,13 +80,13 @@
 import static org.mule.runtime.internal.dsl.DslConstants.CORE_PREFIX;
 import static org.mule.runtime.internal.dsl.DslConstants.CORE_SCHEMA_LOCATION;
 import static org.mule.runtime.internal.dsl.DslConstants.FLOW_ELEMENT_IDENTIFIER;
+import static org.mule.sdk.api.stereotype.MuleStereotypes.CONFIGURATION_ELEMENT;
 
+import static com.google.common.collect.ImmutableSet.of;
```
elrodro83, line 85: java imports go first

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -1122,6 +1123,12 @@ private void declareConfiguration(ExtensionDeclarer extensionDeclarer) {
         .describedAs("The default correlation id generation expression for every source. This must be DataWeave expression.");
   }
 
+  private void declareExtraAppConfigs(ConstructDeclarer configuration) {
```
elrodro83, line 1126: rename

elrodro83 (changes requested):

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
eze210: done

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
eze210: done, also added a comment to be clearer on what does it represent

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/MuleExtensionModelDeclarer.java
```diff
@@ -1122,6 +1124,13 @@ private void declareConfiguration(ExtensionDeclarer extensionDeclarer) {
         .describedAs("The default correlation id generation expression for every source. This must be DataWeave expression.");
   }
 
+  // See the "abstract-configuration-element" in the mule-core-common.xsd file. It represents the same placeholder.
+  private void declareConfigurationElementsPlaceholder(ConstructDeclarer configuration) {
+    configuration.withOptionalComponent("abstractConfigurationElement")
```
elrodro83, line 1129: add the value of the xsd (https://github.com/mulesoft/mule/blob/master/runtime-extension-model/src/main/resources/META-INF/mule-core-common.xsd#L650) as the description, so the source of truth is the extModel

elrodro83 (changes requested):

elrodro83 (approved):

PR #12023, opened 2022-12-05
W-12143338: Properly identify import tag validation failures because of unresolved properties

elrodro83:


runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/AbstractErrorTypesValidation.java
```diff
@@ -73,17 +63,18 @@ public Level getLevel() {
...
     }
-    final ComponentParameterAst errorMappingsAst =
-        operationComponent.getParameter(ERROR_MAPPINGS, ERROR_MAPPINGS_PARAMETER_NAME);
-    return errorMappingsAst != null && errorMappingsAst.getValue().getValue().isPresent();
-  }
 
-  protected static List<ErrorMapping> getErrorMappings(ComponentAst component) {
-    return (List<ErrorMapping>) component.getParameter(ERROR_MAPPINGS, ERROR_MAPPINGS_PARAMETER_NAME).getValue().getRight();
+    if (waiveUnresolvedPropertiesOnParams) {
+      if (getErrorTypeParam(component).getRawValue().contains("${")) {
```
eze210, line 72: This logic should belong to some api, and be more correct. Please add it or create a ticket and add a TODO

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/AbstractErrorValidation.java
```diff
@@ -0,0 +1,89 @@
...
+      builder().namespace(CORE_PREFIX).name(ON_ERROR).build();
+  protected static final ComponentIdentifier ON_ERROR_PROPAGATE_IDENTIFIER =
+      builder().namespace(CORE_PREFIX).name(ON_ERROR_PROPAGATE).build();
+  protected static final ComponentIdentifier ON_ERROR_CONTINUE_IDENTIFIER =
+      builder().namespace(CORE_PREFIX).name(ON_ERROR_CONTINUE).build();
+
+  protected boolean isErrorTypePresentAndPropertyDependant(ComponentAst component) {
+    String errorTypeString = getErrorTypeParam(component).getRawValue();
+    return !isEmpty(errorTypeString)
+        && errorTypeString.contains("${");
```
eze210, line 42: Same as the comment in AbstractErrorTypesValidation

eze210 (changes requested):

PR #12036, opened 2022-12-07
W-11681056: Flow ref validations fails for xml sdk components that declare inner SUB flows when using lazy init

SebaElizalde:


SebaElizalde:
--validate

ndinu (approved):

SebaElizalde:
--validate

PR #12066, opened 2022-12-16
W-12207110: Wrap an IllegalArgumentException in a MessagingException at ForeachRouter in order to fix an event stuck

eze210:


eze210:
Summary of the problem:
- There is a validation in ForeachRouter that prevents the router to process collections if the expression is `#[payload]` and the payload content is a Map.
- This check was accidentally removed in mule runtime 4.3, and re-added in 4.4.
- The 4.4 version raises an IllegalArgumentException and loses the reference to the event context. When that exception reaches the continue strategy defined in AbstractMessageProcessorChain, a ClassCastException is generated, breaking the Flux. Fixing only the ClassCastException isn't enough, since we need access to the EventContext in order to complete it exceptionally (with the .error() method).

Summary of this change:
- Now the ForeachRouter wraps the IllegalArgumentException in a MessagingException. This makes it possible to complete the EventContext from the continue strategy defined in the AbstractMessageProcessorChain. Otherwise, we haven't access to that EventContext and it causes the event getting stuck.
- Added a feature flag, since the concerning validation wasn't correctly done in 4.3, and then adding the check without a FF would break backwards compatibility. minMuleVersion of the FF will be 4.5, and the validation will be just removed in 4.4
- Extracted the methods not annotated with `@Test` from ForeachTestCase to AbstractForeachTestCase, and created a new test class to test the new feature flag.

modules/core-components/src/main/java/org/mule/runtime/core/internal/routing/Foreach.java
```diff
@@ -88,7 +93,16 @@ public CoreEvent process(CoreEvent event) throws MuleException {
...
   @Override
   public Publisher<CoreEvent> apply(Publisher<CoreEvent> publisher) {
-    return new ForeachRouter(this, streamingManager, publisher, expression, batchSize, nestedChain).getDownstreamPublisher();
+    return new ForeachRouter(this, streamingManager, publisher, expression, batchSize, nestedChain, shouldRejectMapExpressions())
+        .getDownstreamPublisher();
+  }
+
+  private boolean shouldRejectMapExpressions() {
+    if (featureFlaggingService == null) {
+      return false;
```
elrodro83, line 102: maybe its better to assume the wanted behaviour in this case?

elrodro83 (changes requested):

elrodro83 (approved):

eze210:
--validate

elrodro83 (approved):

PR #12069, opened 2022-12-16
W-11412449: Test min mule version algorithm with AST and test extensions.

SebaElizalde:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -401,7 +401,8 @@ private static MuleVersion calculateParameterContainerMinMuleVersion(Type contai
   }
 
   private static MuleVersion getEnforcedMinMuleVersion(Type type) {
-    if (type.isAnnotatedWith(DoNotEnforceMinMuleVersion.class)) {
+    if (isPrimitiveType(type) || type.getTypeName().contains("java.lang")
```
ndinu, line 404: Why is the "java.lang" needed?

Also, use a constant

SebaElizalde: When wrapping java primitives and classes in AST mode, the MuleTypeVisitor sets the concreteType as null in those cases. So when calling type.isAnnotatedWith(DoNotEnforceMinMuleVersion.class) in this code, it throws NullPointerException. Checking for primitive or java lang avoids this problem.

ndinu: Which are the types that are not covered by the `isPrimitiveType` method that cause the NullPointer? 

Also, it sounds more like a bug in the `isAnnotatedWith` method that should be solved there.

SebaElizalde: I think you are right, better to fix the isAnnotatedWith method

ndinu (approved):

PR #12097, opened 2022-12-27
Fix/W-12206167: avoid race condition when using loggingVariables

hyerans:


core/src/main/java/org/mule/runtime/core/internal/event/DefaultEventBuilder.java
```diff
@@ -107,9 +107,13 @@ public DefaultEventBuilder(InternalEvent event) {
     this.session = event.getSession();
     this.error = event.getError().orElse(null);
     this.notificationsEnabled = event.isNotificationsEnabled();
-
     this.originalVars = (CaseInsensitiveHashMap<String, TypedValue<?>>) event.getVariables();
     this.loggingVariables = (CaseInsensitiveHashMap<String, String>) event.getLoggingVariables().orElse(null);
+    CaseInsensitiveHashMap<String, String> loggingVariablesCopyMap = new CaseInsensitiveHashMap<>();
```
fsgonz, line 112: this instantiation is only needed in case it will be set (when loggingVariable is null)

fsgonz (changes requested):

core/src/main/java/org/mule/runtime/core/internal/event/DefaultEventBuilder.java
hyerans: done!

fsgonz (approved):

PR #12098, opened 2022-12-27
W-10619594: Properly computing requiresConnection for Mule SDK operations.

asanguinetti:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/OperationModelParser.java
```diff
@@ -111,6 +111,13 @@ public interface OperationModelParser extends SemanticTermsParser, AdditionalPro
    */
   boolean isConnected();
 
+  /**
+   * @return whether this operation requires a connection provided to it from an associated connection provider
```
marianogonzalez, line 115: `@since`

marianogonzalez, line 115: rephrase: `whether this operation requires connections obtained through a {@link ConnectionProvider} defined in this same extension`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/ReconnectionUtils.java
```diff
@@ -133,15 +134,21 @@ public static InterceptorChain createReconnectionInterceptorsChain(ExtensionMode
                                                                      ReflectionCache reflectionCache) {
     InterceptorChain.Builder chainBuilder = InterceptorChain.builder();
 
-    if (componentModel instanceof ConnectableComponentModel) {
-      if (((ConnectableComponentModel) componentModel).requiresConnection()) {
-        addConnectionInterceptors(chainBuilder, extensionModel, componentModel, connectionSupplier, reflectionCache);
-      }
+    if (requiresConnectionInterceptors(componentModel)) {
+      addConnectionInterceptors(chainBuilder, extensionModel, componentModel, connectionSupplier, reflectionCache);
```
marianogonzalez, line 138: this doesn't look correct. This method adds two interceptors. `ConnectionInterceptor` and `CursorResetInterceptor`. 

The first one you will still need in Mule SDK. At most, you could optimize by removing it when the extension flavour is `in app reuse`

The need for the second should not be determined by the connection provisioning but the presence of `NoReconnectionStrategyModelProperty`. 

The concept of `ConnectionProvisioningModelProperty` makes less and less sense to me

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/OperationModelParser.java
asanguinetti: done

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/ReconnectionUtils.java
asanguinetti: Changed to use a flavor-based model property so we can specifically check for the `in app reuse`. Still I kept it encapsulated in a utils method because we need a similar check in more than one place. This way, if we decide to change how we implement the differentiation (flavor or whatever) we don't need to change everywhere (hopefully). Let me know if that makes sense.

Also separated the condition for the `CursorResetInterceptor` so we can check for the `NoReconnectionStrategyModelProperty`.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkApplicationExtensionModelParser.java
```diff
@@ -47,6 +52,11 @@ public MuleSdkApplicationExtensionModelParser(String extensionName,
     this.extensionName = extensionName;
   }
 
+  @Override
+  public List<ModelProperty> getAdditionalModelProperties() {
+    return singletonList(new SdkFlavorModelProperty(SDK_FLAVOR_MULE_IN_APP));
```
marianogonzalez, line 57: this should not come as an additional property. `ExtensionModelParser` should be given an explicit `ExtensionDevelopmentFramework getExtensionDevelopmentFramework()` method. Adding the model property should be done on the loader delegate.

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkOperationModelParser.java
```diff
@@ -213,7 +217,11 @@ public boolean isRouter() {
 
   @Override
   public boolean isConnected() {
-    // TODO: MULE-20077
+    return isConnected.getValue();
+  }
+
+  @Override
+  public boolean requiresOwnConnection() {
```
marianogonzalez, line 224: this method should no longer be needed since everything now depends on the flavor/reconnection strategy

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkPluginExtensionModelParser.java
```diff
@@ -65,6 +69,11 @@ protected void init(ArtifactAst ast) {
     super.init(ast);
   }
 
+  @Override
+  public List<ModelProperty> getAdditionalModelProperties() {
+    return singletonList(new SdkFlavorModelProperty(SDK_FLAVOR_MULE));
```
marianogonzalez, line 74: remove per previous comment

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaExtensionModelParser.java
```diff
@@ -128,6 +130,7 @@ private void parseStructure(ExtensionElement extensionElement) {
     additionalModelProperties.add(new ExtensionTypeDescriptorModelProperty(extensionElement));
     extensionElement.getDeclaringClass()
         .ifPresent(extensionClass -> additionalModelProperties.add(new ImplementingTypeModelProperty(extensionClass)));
+    additionalModelProperties.add(new SdkFlavorModelProperty(SDK_FLAVOR_JAVA));
```
marianogonzalez, line 133: remove per previous comment

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/ReconnectionUtils.java
```diff
@@ -134,21 +136,33 @@ public static InterceptorChain createReconnectionInterceptorsChain(ExtensionMode
...
+  private static boolean requiresConnectionInterceptors(ExtensionModel extensionModel, ConnectableComponentModel componentModel) {
+    // Only connectable components that require a connection to be provided beforehand should add the connection interceptors
+    return requiresConnectionProvisioning(extensionModel, componentModel);
+  }
+
+  private static boolean requiresCursorResetInterceptors(ConnectableComponentModel componentModel) {
+    // Connectable components that don't allow for controlling the reconnection strategy should not add the cursor reset
+    // interceptors
+    return componentModel.requiresConnection()
+        && !componentModel.getModelProperty(NoReconnectionStrategyModelProperty.class).isPresent();
```
marianogonzalez, line 160: already seen this quite a few times. Find all the places in which the presence of `NoReconnectionStrategyModelProperty` is checked and extract to some utils.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/ReconnectionUtils.java
```diff
@@ -134,21 +136,33 @@ public static InterceptorChain createReconnectionInterceptorsChain(ExtensionMode
     InterceptorChain.Builder chainBuilder = InterceptorChain.builder();
 
     if (componentModel instanceof ConnectableComponentModel) {
-      if (((ConnectableComponentModel) componentModel).requiresConnection()) {
-        addConnectionInterceptors(chainBuilder, extensionModel, componentModel, connectionSupplier, reflectionCache);
+      if (requiresConnectionInterceptors(extensionModel, (ConnectableComponentModel) componentModel)) {
```
marianogonzalez, line 139: not strictly related to your change, but in the spirit of the boyscout rule let's take the opportunity to refactor this. This class is for reconnection, but The CursorReset is the only interceptor that is actually about reconnecting. This method should receive the chainBuilder and only add the reconnection interceptors only. The ConnectionInterceptor should be added someplace else.

marianogonzalez (changes requested):

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkApplicationExtensionModelParser.java
asanguinetti: Done

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkOperationModelParser.java
asanguinetti: Removed

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkPluginExtensionModelParser.java
asanguinetti: Done

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaExtensionModelParser.java
asanguinetti: Done

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/ReconnectionUtils.java
asanguinetti: Done

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/ReconnectionUtils.java
asanguinetti: Done.
I created an `InterceptorChainUtils` with a `createConnectionInterceptorsChain` which adds the `ConnectionInterceptor` and then delegates to the `ReconnectionUtils` class to add the `CursorResetInterceptor` (if necessary) to the builder as you described.

PR #12099, opened 2022-12-28
W-12220317: [Tracer Performance] Remove commands for coreEventTracer or do not generate instances for each request

fsgonz:


fsgonz:
--validate

modules/tracing/mule-tracer-internal-impl/src/main/java/org/mule/runtime/tracer/impl/span/command/EventContextAddAttributesCommand.java
```diff
@@ -7,47 +7,40 @@
...
+import org.slf4j.Logger;
 
 import java.util.Map;
-
-import static org.mule.runtime.tracer.impl.span.command.spancontext.SpanContextFromEventContextGetter.getSpanContextFromEventContextGetter;
+import java.util.function.BiConsumer;
 
 /**
- * A {@link VoidCommand} that ads span attributes. The carrier is the {@link EventContext}
+ * A {@link AbstractFailSafeVoidBiCommand} that ads span attributes. The carrier is the {@link EventContext}
```
sofiamorseletto, line 20: adds*

modules/tracing/mule-tracer-internal-impl/src/main/java/org/mule/runtime/tracer/impl/span/command/EventContextEndSpanCommand.java
```diff
@@ -11,55 +11,38 @@
...
-    this.eventContext = eventContext;
-    this.assertion = assertion;
+  private EventContextEndSpanCommand(Logger logger, String errorMessage, boolean propagateExceptions) {
+    super(logger, errorMessage, propagateExceptions);
   }
 
-  protected Runnable getRunnable() {
-    return () -> {
+  @Override
+  BiConsumer<EventContext, Assertion> getConsumer() {
```
sofiamorseletto, line 39: To understand a little bit more, on the "getFunction" classes such as AbstractFailsafeUnaryCommand, the command is created on the constructor, saved on a variable and then returned on the getter. Why are we not doing the same for the "getConsumer" classes such as this one? Why is the consumer created on the getter and not on the constructor?

fsgonz: Yes, you're right. My mistake. It should be created in the constructor.

modules/tracing/mule-tracer-internal-impl/src/main/java/org/mule/runtime/tracer/impl/span/command/EventContextInjectDistributedTraceContextCommand.java
```diff
@@ -13,32 +13,33 @@
...
-    this.eventContext = eventContext;
-    this.getter = getter;
+  private EventContextInjectDistributedTraceContextCommand(Logger logger, String errorMessage, boolean propagateExceptions) {
+    super(logger, errorMessage, propagateExceptions);
   }
 
-  protected Runnable getRunnable() {
-    return () -> {
+  @Override
+  BiConsumer<EventContext, DistributedTraceContextGetter> getConsumer() {
```
sofiamorseletto, line 41: According to what we discussed before, the creation of the consumer should be moved to the constructor of the class. This goes for all classes that use the getConsumer method, I assume are the ones that extend `AbstractFailSafeVoidBiCommand`

modules/tracing/mule-tracer-internal-impl/src/main/java/org/mule/runtime/tracer/impl/span/command/EventContextAddAttributesCommand.java
sofiamorseletto: @fsgonz if you can please change this typo too

sofiamorseletto (changes requested):

sofiamorseletto (approved):

PR #12103, opened 2022-12-30
W-12284401: OpenTelemetry exporter configuration improvements

IvanAndresFritzler:


fsgonz (approved):

modules/tracing/mule-tracer-exporter-configuration-api/src/main/java/org/mule/runtime/tracer/exporter/api/config/OpenTelemetrySpanExporterConfigurationProperties.java
```diff
@@ -7,6 +7,7 @@
 
 package org.mule.runtime.tracer.exporter.api.config;
 
+// TODO: Move the OPTEL config properties away from API.
```
fsgonz, line 10: I'm not sure if this should be removed from API as is part of the definitions so in part it is api (notice that it is not exposed as API in the sense of mule)

fsgonz (approved):

fsgonz:
--validate

fsgonz (approved):

PR #12106, opened 2023-01-02
W-12233066: [Tracing performance] Improve ExecutionSpan errors tracing

fsgonz:


modules/tracing/mule-tracer-internal-impl/src/main/java/org/mule/runtime/tracer/impl/span/ExecutionSpan.java
```diff
@@ -47,7 +47,7 @@ public static ExecutionSpanBuilder getExecutionSpanBuilder() {
   private final Long startTime;
   private Long endTime;
   private final Map<String, String> attributes = new HashMap<>();
-  private final Set<SpanError> errors = new HashSet<>();
+  private final List<SpanError> errors = new ArrayList<>(1);
```
sofiamorseletto, line 50: If we know that we only have 1 error, wouldn't it be less error-prone to have a SpanError variable instead of a list?

sofiamorseletto (approved):

modules/tracing/mule-tracer-internal-impl/src/main/java/org/mule/runtime/tracer/impl/span/ExecutionSpan.java
fsgonz: I will change that in the following PR. This is a cherrypick.

fsgonz:
--validate

fsgonz:
--validate

fsgonz:
--validate

fsgonz:
--validate

fsgonz:
--validate

PR #12116, opened 2023-01-03
W-12285074: Internal span attribute span.kind.override is showing up …

sofiamorseletto:
…on final exported span when it shouldn't

### DIFF
--- a/modules/tracing/mule-tracer-internal-impl/src/main/java/org/mule/runtime/tracer/impl/span/ExecutionSpan.java
+++ b/modules/tracing/mule-tracer-internal-impl/src/main/java/org/mule/runtime/tracer/impl/span/ExecutionSpan.java
@@ -35,6 +35,9 @@ import java.util.function.BiConsumer;
  */
 public class ExecutionSpan implements InternalSpan {
 
+  private static final String SPAN_KIND = "span.kind.override";
+  private static final String STATUS = "status.override";
+
   private final InitialSpanInfo initialSpanInfo;
   private SpanExporter spanExporter = NOOP_EXPORTER;
   private SpanError lastError;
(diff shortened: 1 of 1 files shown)


modules/tracing/mule-tracer-internal-impl/src/main/java/org/mule/runtime/tracer/impl/span/ExecutionSpan.java
```diff
@@ -165,7 +168,9 @@ public Long getEnd() {
 
   @Override
   public void addAttribute(String key, String value) {
-    additionalAttributes.put(key, value);
+    if (!key.equals(SPAN_KIND) && !key.equals(STATUS)) {
```
fsgonz, line 171: Put a TODO here indicating: This should be moved to the OpentelemetrySpanExporter foreach (it only makes sense in optel) after performance assessment. 
https://github.com/mulesoft/mule/blob/master/modules/tracing/mule-tracer-exporter-impl/src/main/java/org/mule/runtime/tracer/impl/exporter/OpenTelemetrySpanExporter.java#L134

I've created a ticket for the TODO W-12297079

fsgonz (changes requested):

sofiamorseletto:
--validate

fsgonz (approved):
Change the integration test for JMS so that the number of attributes is tested to be ok. The mule ticket is ok.

fsgonz (approved):

modules/tracing/mule-tracer-internal-impl/src/main/java/org/mule/runtime/tracer/impl/span/ExecutionSpan.java
sofiamorseletto: I added this suggested change on this PR

PR #12144, opened 2023-01-12
W-12129696: WSC connector with HTTP request fails when using a domain project

marianorean:


tests/allure/src/main/java/org/mule/test/allure/AllureConstants.java
```diff
@@ -482,6 +482,8 @@ interface WscFeature {
 
     interface WscStory {
 
+      String HTTP_REQUEST = "HTTP Request";
```
elrodro83, line 485: remote WSDL loading better?

elrodro83 (changes requested):

elrodro83 (approved):

PR #12156, opened 2023-01-13
W-12389157: Support MessageSources in ExtensionsClient

marianogonzalez:


marianogonzalez:
--validate

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/EventedExtensionsClientDecorator.java
```diff
@@ -58,4 +61,11 @@ public <T, A> CompletableFuture<Result<T, A>> execute(String extension,
       }
     });
   }
+
+  @Override
+  public <T, A> SourceHandler createSource(String extension, String sourceName,
+                                           Consumer<SourceResultCallback<T, A>> callback,
+                                           Consumer<SourceParameterizer> parameters) {
+    return extensionsClient.createSource(extension, sourceName, callback, parameters);
```
ndinu, line 69: The context of the event will not be used to create the source.

What about logging a Warning or a throwing a not supported exception here?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConfigurer.java
```diff
@@ -126,39 +127,31 @@ public Object build(ValueResolvingContext context) throws MuleException {
...
-      if (sdkSource instanceof PollingSource) {
-        ValueResolver<?> valueResolver = resolverSet.getResolvers().get(SCHEDULING_STRATEGY_PARAMETER_NAME);
-        if (valueResolver != null) {
-          context = ValueResolvingContext.builder(initialiserEvent, expressionManager).build();
-          SchedulingStrategy scheduler = (SchedulingStrategy) valueResolver.resolve(context);
+        if (sdkSource instanceof PollingSource) {
+          SchedulingStrategy scheduler;
+          ValueResolver<?> valueResolver = resolverSet.getResolvers().get(SCHEDULING_STRATEGY_PARAMETER_NAME);
+          if (valueResolver == null) {
+            scheduler = new FixedFrequencyScheduler(60000, 0, MILLISECONDS);
```
ndinu, line 140: What about putting this default in the SourceParameterizer, if it is a polling source and no scheduling strategy is given?

Putting it here makes it a hidden default that may cover an issue in future sources.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/EventedExtensionsClientDecorator.java
marianogonzalez: that's an internal implementation detail about a class that's not even API and only used for operations. What value would such log add?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceConfigurer.java
marianogonzalez: I tried originally but introduces a lot of other issues. This doesn't hide a default, since the default is already present on the XSD and the EM. Quite the contrary, this fixes a bug in the original code in which if a polling source is not provided with a scheduler, then it never starts and dies silently

ndinu (dismissed):

ndinu (approved):

PR #12164, opened 2023-01-17
W-11151215: Create Logging Information about the resulting MinMuleVersion.

SebaElizalde:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/SdkComponentsMinMuleVersionUtils.java
```diff
@@ -0,0 +1,655 @@
...
+      }
+      if (annotation.isSameType(Config.class) || annotation.isSameType(org.mule.sdk.api.annotation.param.Config.class)) {
+        methodParameterComponent =
+            max(methodParameterComponent,
+                getConfigurationComponent(methodParameter.getType(),
+                                          getEnforcedMinMuleVersion(annotation).getMinMuleVersion()));
+      }
+      SdkComponent annotationComponent = getEnforcedMinMuleVersion(annotation);
+      methodParameterComponent.updateIfHigherMMV(annotationComponent,
+                                                 format("Parameter %s has min mule version %s because it is annotated with %s.",
```
marianogonzalez, line 577: not a helpful message for the user

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/SdkComponentsMinMuleVersionUtils.java
```diff
@@ -0,0 +1,655 @@
...
+      sdkComponent = max(sdkComponent, getEnforcedMinMuleVersion(outputType));
+      sdkComponent =
+          max(sdkComponent, getEnforcedMinMuleVersion(outputType.getGenerics().get(1).getConcreteType()));
+    } else {
+      sdkComponent = getEnforcedMinMuleVersion(outputType);
+    }
+    return sdkComponent;
+  }
+
+  private static SdkComponent calculateFieldMinMuleVersion(ExtensionParameter field) {
```
marianogonzalez, line 494: none of the error messages generated by this method are helpful to the user. Saying `version V was resolved because X is of type Y` or `X is annotated with Y` is not helpful for the customer.

Messages should be clear, prescriptive and not requiere internal Mule knowledge to interpret. Instead, it should say `version V was resolved because X is annotated with Y, which vas introduced in Mule 4.7.`

Apply this comment to every single error message generated in this class

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/SdkComponent.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.mule.runtime.api.meta.MuleVersion;
+import org.mule.sdk.api.annotation.MinMuleVersion;
+
+/**
+ * Holds information about an SdkComponent, its {@link MinMuleVersion} and the reason for that value.
+ *
+ * @since 4.6
+ */
+public class SdkComponent {
```
marianogonzalez, line 17: This class is very confusing:

* First I think the name is wrong. This is not modeling an `SdkComponent`, but the result of a version calculation.
* Also because reason seems to be something that is always assumed to be non-null, does this actually represent a result in which component MMV > extension MMV? 

I would rethink this completely

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaSourceModelParser.java
```diff
@@ -309,7 +309,14 @@ public SourceClusterSupportModelProperty getSourceClusterSupportModelProperty()
...
   public Optional<MuleVersion> getMinMuleVersion() {
-    return of(calculateSourceMinMuleVersion(sourceElement));
+    // TODO W-12392052
+    return of(getSourceComponent(sourceElement).getMinMuleVersion());
+  }
+
+  @Override
+  public Optional<String> getMinMuleVersionReason() {
+    // TODO W-12392052
+    return of(getSourceComponent(sourceElement).getReason());
```
marianogonzalez, line 319: why `of` ? If there's ALWAYS a reason, then what's the value for a case in which the resolved version is 4.1.1? Also, why having `getMinMuleVersion()` and `getMinMuleversionReason()` instead of a single method that returns an object with all the version related data at once?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/SdkComponentsMinMuleVersionUtils.java
```diff
@@ -0,0 +1,655 @@
...
+import static java.lang.String.format;
+import static java.util.Collections.emptyList;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Utils class to create {@link SdkComponent}s from {@link Type}s.
+ *
+ * @since 4.5
+ */
+public final class SdkComponentsMinMuleVersionUtils {
```
marianogonzalez, line 52: Why the `SDkComponents` prefix? This looks more as `JavaMinMuleVersionUtils`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/SdkComponentsMinMuleVersionUtils.java
```diff
@@ -0,0 +1,655 @@
...
+
+  private static final MuleVersion SDK_EXTENSION_ANNOTATION_MIN_MULE_VERSION =
+      new MuleVersion(Extension.class.getAnnotation(MinMuleVersion.class).value());
+  public static final MuleVersion FIRST_MULE_VERSION = new MuleVersion("4.1.1");
```
marianogonzalez, line 56: This actually looks like a constant that belongs someplace else as this is a fact not exclusive to this class domain

marianogonzalez (changes requested):

modules/extensions-spring-support/src/test/resources/models/heisenberg.json
```diff
@@ -32410,7 +32410,7 @@
           "namespace": "MULE"
         }
       },
-      "minMuleVersion": "4.5.0",
+      "minMuleVersion": "4.6.0",
```
SebaElizalde, line 32413: while doing the refactor I fixed a bug in the min mule version calculation for methods with parameter groups, where annotations inside parameter groups were not taken into account. I've checked models changes are they are correct. I'm copying here the reason for their new version:
`Method getInlinePersonalInfo has min mule version 4.6.0 because of its parameter info. Parameter info has min mule version 4.6.0 because it is of type PersonalInfo. Type PersonalInfo has min mule version 4.6.0 because of its field description. Field description has min mule version 4.6.0 because it is annotated with Text. Text was introduced in Mule 4.6.0.`

modules/extensions-spring-support/src/test/resources/models/heisenberg.json
```diff
@@ -33351,7 +33351,7 @@
           "namespace": "MULE"
         }
       },
-      "minMuleVersion": "4.1.1",
+      "minMuleVersion": "4.5.0",
```
SebaElizalde, line 33354: Method approve has min mule version 4.5.0 because of its parameter recursivePojo. Parameter recursivePojo has min mule version 4.5.0 because it is of type RecursivePojo. Type RecursivePojo has min mule version 4.5.0 because of its field next. Field next has min mule version 4.5.0 because it is a parameter of type RecursivePojo. Type RecursivePojo has min mule version 4.5.0 because it is annotated with TypeDsl. TypeDsl was introduced in Mule 4.5.0.

modules/extensions-spring-support/src/test/resources/models/heisenberg.json
```diff
@@ -41882,7 +41882,7 @@
           "namespace": "MULE"
         }
       },
-      "minMuleVersion": "4.5.0",
+      "minMuleVersion": "4.6.0",
```
SebaElizalde, line 41885: Method getInlinePersonalInfo has min mule version 4.6.0 because of its parameter info. Parameter info has min mule version 4.6.0 because it is of type PersonalInfo. Type PersonalInfo has min mule version 4.6.0 because of its field description. Field description has min mule version 4.6.0 because it is annotated with Text. Text was introduced in Mule 4.6.0.

modules/extensions-spring-support/src/test/resources/models/values.json
```diff
@@ -11034,7 +11034,7 @@
           }
         }
       },
-      "minMuleVersion": "4.1.1",
+      "minMuleVersion": "4.5.0",
```
SebaElizalde, line 11037: Method multiLevelValue has min mule version 4.5 because of its parameter optionsParameter. Parameter optionsParameter has min mule version 4.5 because it is of type GroupAsMultiLevelValue. Type GroupAsMultiLevelValue has min mule version 4.5 because of its field country. Field country has min mule version 4.5 because it is annotated with ValuePart. ValuePart was introduced in Mule 4.5.

modules/extensions-spring-support/src/test/resources/models/typed-value.json
```diff
@@ -2759,7 +2759,7 @@
           }
         }
       },
-      "minMuleVersion": "4.1.1",
+      "minMuleVersion": "4.5.0",
```
SebaElizalde, line 2762: Method getAquarium has min mule version 4.5.0 because of its parameter aquarium. Parameter aquarium has min mule version 4.5.0 because it is of type Aquarium. Type Aquarium has min mule version 4.5.0 because of its field pond. Field pond has min mule version 4.5.0 because it is annotated with ParameterGroup. ParameterGroup was introduced in Mule 4.5.0.

modules/extensions-spring-support/src/test/resources/models/petstore.json
```diff
@@ -21558,7 +21558,7 @@
           }
         }
       },
-      "minMuleVersion": "4.1.1",
+      "minMuleVersion": "4.5.0",
```
SebaElizalde, line 21561: Method getAquarium has min mule version 4.5.0 because of its parameter aquarium. Parameter aquarium has min mule version 4.5.0 because it is of type Aquarium. Type Aquarium has min mule version 4.5.0 because of its field pond. Field pond has min mule version 4.5.0 because it is annotated with ParameterGroup. ParameterGroup was introduced in Mule 4.5.0.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/SourceModelParser.java
```diff
@@ -150,9 +150,10 @@ public interface SourceModelParser extends SemanticTermsParser, StereotypeModelP
   SourceClusterSupportModelProperty getSourceClusterSupportModelProperty();
 
   /**
-   * @return a {@link MuleVersion} representing the minimum mule version this component can run on
+   * @return a {@link MinMuleVersionResult} that contains the minimum mule version this component can run on and the reason why
+   *         that version was assigned.
```
ndinu, line 154: Shouldn't this still be Optional<MinMuleVersionResult> ? Why is the possibility of empty value eliminated.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/MinMuleVersionResult.java
```diff
@@ -0,0 +1,51 @@
...
+    this.componentName = componentName;
+    this.minMuleVersion = minMuleVersion;
+    this.reason = reason;
+  }
+
+  public MuleVersion getMinMuleVersion() {
+    return minMuleVersion;
+  }
+
+  public String getReason() {
```
ndinu, line 33: Is the reason being logged somewhere? I haven't find it and don't know if maybe this is due to a discussion that already happened.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/SourceModelParser.java
SebaElizalde: We discussed it above: it always returned a value, so it didn't make sense to make it an optional.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/MinMuleVersionResult.java
SebaElizalde: You are right. It is not logged anywhere at the moment. We should discuss if we want to just log it in the ModelLoaderDelegates or do we extend the min mule version property and add the reason to it? What do you think?

ndinu: I don't think that information is relevant for the model, since it speak more of the implementation than the behavior or contract of the extension itself. If we choose to log it, yes, do it at model loader delegates level, so that all sdks get that logging.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/SourceModelParser.java
ndinu: Will ParameterParsers go through the same change? I see in this PR that they are unchanged. 

In my opinion, the parameter parser should at least return an Optional<MinMuleVersionResult>. What do you think?

SebaElizalde: I think I would either remove the methods or make them return the default min mule version. Seems inconsistent to make them Optional while in others they aren't optional.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/ConfigurationModelParser.java
```diff
@@ -101,7 +102,8 @@ public interface ConfigurationModelParser extends StereotypeModelParser, Additio
   Optional<DisplayModel> getDisplayModel();
 
   /**
-   * @return a {@link MuleVersion} representing the minimum mule version this component can run on
+   * @return a {@link MinMuleVersionResult} that contains the minimum mule version this component can run on and the reason why
+   *         that version was assigned.
    */
-  Optional<MuleVersion> getMinMuleVersion();
```
marianogonzalez, line 106: the reasoning for this to be optional is decoupling the parser from the default. For instance, if the parser doesn't determine any specific version, the delegate can default to the extension's MMV without the need to pass it all around. What's the justification for this change? Why is it being done in a PR that's supposed to be about logging?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaConfigurationModelParser.java
```diff
@@ -70,11 +69,11 @@ public JavaConfigurationModelParser(JavaExtensionModelParser extensionModelParse
     this.configElement = configElement;
 
     parseStructure();
-    this.minMuleVersion = calculateConfigMinMuleVersion(configElement,
-                                                        getContainerAnnotationMinMuleVersion(extensionElement,
-                                                                                             Configurations.class,
-                                                                                             Configurations::value,
-                                                                                             configElement));
+    this.minMuleVersionResult = getConfigurationResult(configElement,
```
marianogonzalez, line 72: previous method name was way more decriptive

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaConnectionProviderModelParser.java
```diff
@@ -96,7 +96,7 @@ public JavaConnectionProviderModelParser(JavaExtensionModelParser extensionModel
         .orElse(ExtensionModel.class.getClassLoader());
 
     collectAdditionalModelProperties();
-    this.minMuleVersion = calculateConnectionProviderMinMuleVersion(element);
+    this.minMuleVersionResult = getConnectionProviderResult(element);
```
marianogonzalez, line 99: same as before. These are not good renames.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/MinMuleVersionResult.java
```diff
@@ -0,0 +1,51 @@
...
+package org.mule.runtime.module.extension.internal.loader.parser.java.utils;
+
+import org.mule.runtime.api.meta.MuleVersion;
+
+/**
+ * Holds information about the minimum {@link MuleVersion} that a component can run on and the reason for that value.
+ *
+ * @since 4.6
+ */
+public class MinMuleVersionResult {
```
marianogonzalez, line 16: This should be called `ResolvedMinMuleVersion` or simply `MinMuleVersion`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/MinMuleVersionResult.java
marianogonzalez: if not logged, why have it at all?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/MinMuleVersionUtils.java
```diff
@@ -0,0 +1,709 @@
...
+   * @return the minimum mule version of the given extension
+   */
+  public static MinMuleVersionResult getExtensionResult(Type extension) {
+    if (belongsToJavaPackages(extension.getTypeName())) {
+      return createResultWithDefaultMMV("Java type", extension.getTypeName());
+    }
+    MinMuleVersionResult extensionResult;
+    if (extension.isAnnotatedWith(Extension.class)) {
+      extensionResult = new MinMuleVersionResult(extension.getName(), SDK_EXTENSION_ANNOTATION_MIN_MULE_VERSION,
+                                                 format("Extension %s has min mule version %s because it if annotated with the new sdk api @Extension.",
```
marianogonzalez, line 85: do not user String.format() here. THat method is VERY slow and will affect Mule startup time.

marianogonzalez (changes requested):
Where are the logs?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/ConfigurationModelParser.java
SebaElizalde: Good point. I was thinking that the parser would always return a result and would handle the default case too. But maybe what you say makes more sense. I'm gonna return it to an optional then.

marianogonzalez (approved):

PR #12169, opened 2023-01-18
W-12271802: Add Min Mule Version information to the Extension Model

ndinu:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -430,4 +433,15 @@ private static boolean belongsToExtensionsApiPackages(String fullyQualifiedName)
   private static boolean belongsToSdkApiPackages(String fullyQualifiedName) {
     return fullyQualifiedName.startsWith("org.mule.sdk.api");
   }
+
+  /**
+   * ADD EXPLANAITION
```
marianogonzalez, line 438: do, or do not. There's no try

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/MinMuleVersionLoaderDelegatesTestCase.java
```diff
@@ -0,0 +1,346 @@
...
+    DslResolvingContext resolvingContext = getDefault(emptySet());
+    DefaultExtensionLoadingContext loadingContext = new DefaultExtensionLoadingContext(classLoader, resolvingContext);
+    loadingContext.addParameter(VERSION, getProductVersion());
+    loadingContext.addParameter(EXTENSION_TYPE, DummyExtension.class);
+    return loadingContext;
+  }
+
+  private JavaExtensionModelParserFactory createExtensionParserFactoryForMinMuleVersion(DefaultExtensionLoadingContext loadingContext,
+                                                                                        Optional<MuleVersion> muleVersion) {
+    DefaultAnyType defaultAnyType = new DefaultAnyType(MetadataFormat.JAVA, ImmutableMap.of());
```
marianogonzalez, line 80: static import and use Collections.emptyMap() instead

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/MinMuleVersionLoaderDelegatesTestCase.java
```diff
@@ -0,0 +1,346 @@
...
+import org.mule.runtime.module.extension.internal.loader.parser.SourceModelParser;
+import org.mule.runtime.module.extension.internal.loader.parser.java.JavaExtensionModelParser;
+import org.mule.runtime.module.extension.internal.loader.parser.java.JavaExtensionModelParserFactory;
+import org.mule.sdk.api.annotation.Extension;
+import org.mule.sdk.api.annotation.source.SourceClusterSupport;
+
+import java.util.List;
+import java.util.Optional;
+
+public class MinMuleVersionLoaderDelegatesTestCase {
```
marianogonzalez, line 67: tests need to at least extend `AbstractMuleTestCase` at minimum. Also, it'd be great to add allure annotations

marianogonzalez, line 67: not getting the point of this test. It's way too hardwired. Most of the code is actually manually building models which then of course will match the assertions.

I would rather have a functional test which takes test extensions and validates that each component has the right versions.

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaFunctionModelParserTestCase.java
```diff
@@ -33,6 +34,7 @@ public class JavaFunctionModelParserTestCase {
   ClassTypeLoader typeLoader =
       ExtensionsTypeLoaderFactory.getDefault().createTypeLoader(Thread.currentThread().getContextClassLoader());
 
+  @Ignore("W-12398760")
```
marianogonzalez, line 37: this looks weird. The issue only talks about `calculation issues`, but the tests are already existing but you're only ignoring them in this PR. Why weren't they failing before?

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
ndinu: Removing this

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/MinMuleVersionLoaderDelegatesTestCase.java
ndinu: This tests make sure that the loader delegates use the parsers to select the min mule version of the models. For this, the parsers used are mocks, this way it does not depend on the java parsers implementation.

I want to know that for all models that can have the minMuleVersion configured, the loaders will ask the parser for the minMuleVersion and configure the declaration with that version.

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaFunctionModelParserTestCase.java
marianogonzalez: but why ignore them in this PR?. If they're failing at master right now, then they should be ignored in their own PR. If they only fail in this PR, then it's this PR the one breaking them

ndinu: Before this PR, the min mule version calculates was "dead code", meaning, it was only used for its tests. Now that it is being called, the problems for running that code against the test extensions came to float.

I changed it so that for now the calculation just returns empty() and that is why these tests break.

We may have to wait for the bug to get fixed in order to merge this PR.

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/MinMuleVersionLoaderDelegatesTestCase.java
ndinu: I think that we already have tests for the parsers calculation. This test isolates the parser behavior by creating the mocks and tests the usage of the parsers.

If this test breaks, we know that the issue is not in the parsers logic, and we know that it will affect all SDKs.

What you are proposing is have this test use an already existing implementation of the model parsers?

marianogonzalez: what I'm saying is that tests should focus on the system behavior, resilience performance, etc.. Not in the exact implementation that leads to those things.

If tomorrow we decide to refactor the parsers completely, or getting rid of the loadingDelegate all together, then this test is forced to change alongside it. By doing that, it's not really verifying that no regressions are introduced or compatibility is changed. This is because you're **testing the implementation**. 

On the other side, if you write tests that takes extensions and verify that the versions are correclty resolved, you're testing the behavior. You can completely change how we resolve this, your test is still valid as it's untouched. That's because it's **testing the behaviour.**

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaFunctionModelParserTestCase.java
marianogonzalez: Ok, I understand now. It's fine then, but only as long a retrospective is done with the team to analyze how come this dead yet not working code got merged and how to avoid that in the future. This is a sign of not great coding practices.

@afelisatti ☝🏻

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/MinMuleVersionLoaderDelegatesTestCase.java
ndinu: Got it, I will add examples in the test-extensions and make sure that when the whole model is created the min mule version is the correct one.

What would happen for the case of parameters? we cannot make such test for parameters since there is no way to set a min mule version for them.

ndinu: Ok.

For those cases, I think that the decision is that the operation mule version should be higher instead of the parameters, at least for what I see in the code.

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaFunctionModelParserTestCase.java
afelisatti: We've agreed to not ignore anything. The code needs to be fixed and then Nico's change can go in without the ignores. As for the "dead code", it sounds like some reviews failed.

marianogonzalez (approved):

PR #12179, opened 2023-01-20
W-12398760: Calculations of minMuleVersion for test extensions fail.

SebaElizalde:


marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaFunctionModelParser.java
```diff
@@ -58,6 +60,7 @@ public JavaFunctionModelParser(ExtensionElement extensionElement,
 
     parseStructure();
     collectAdditionalModelProperties();
+    this.minMuleVersion = getMinMuleVersion().orElse(FIRST_MULE_VERSION);
```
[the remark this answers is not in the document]
SebaElizalde: Why? If the extension/config has a higher MMV they will be pruned accordingly. Also what you mean by baseline, the one set at the extension level (via @mmv), the one calculated or the one declared in the pom? In any case, I think this is a separate issue to this PR. The idea of this line is to call the getMinMuleVersion when running the test extensions.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -310,7 +314,8 @@ private static MuleVersion calculateSourceGenericsMinMuleVersion(List<Type> sour
 
   private static MuleVersion calculateOutputMinMuleVersion(Type outputType) {
     MuleVersion calculatedMMV = FIRST_MULE_VERSION;
-    if (outputType.asMetadataType() instanceof ArrayType) {
+    if (outputType.isAssignableTo(Collection.class) || outputType.isAssignableTo(Iterable.class)
+        || outputType.isAssignableTo(Iterator.class)) {
```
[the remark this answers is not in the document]
SebaElizalde: Do you know where can I find the translation from Java classes to ArrayType?

SebaElizalde:
--validate

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaConfigurationModelParser.java
```diff
@@ -69,6 +71,7 @@ public JavaConfigurationModelParser(JavaExtensionModelParser extensionModelParse
     this.configElement = configElement;
 
     parseStructure();
+    this.minMuleVersion = getMinMuleVersion().orElse(FIRST_MULE_VERSION);
```
ndinu, line 74: I would also add this to the Extension parser, in order to check there is no issue with those too.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaConnectionProviderModelParser.java
```diff
@@ -95,6 +97,7 @@ public JavaConnectionProviderModelParser(JavaExtensionModelParser extensionModel
         .orElse(ExtensionModel.class.getClassLoader());
 
     collectAdditionalModelProperties();
+    this.minMuleVersion = getMinMuleVersion().orElse(FIRST_MULE_VERSION);
```
ndinu, line 100: What would happen to this code later? My understanding is that this is added so that the test extensions actually calculate this, but in the final code this line would not exist.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -206,29 +208,24 @@ public static MuleVersion calculateConnectionProviderMinMuleVersion(Type connect
    * @param source the connection provider, as a {@link Type}, to calculate its min mule version
    * @return the minimum mule version of the given source
    */
-  public static MuleVersion calculateSourceMinMuleVersion(Type source) {
+  public static MuleVersion calculateSourceMinMuleVersion(SourceElement source) {
+    MuleVersion sourceGenericsMMV = calculateSourceGenericsMinMuleVersion(source.getSuperClassGenerics());
```
ndinu, line 212: Have you checked that this issue does not happen with other cases, like connection providers?

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaConfigurationModelParser.java
SebaElizalde: done

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaConnectionProviderModelParser.java
SebaElizalde: yes, this is not final code.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
SebaElizalde: Yes. right now for connection providers we are not analyzing its generic.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaConnectionProviderModelParser.java
ndinu: Add a TODO pointing to a gus issue (you can use the one I am currently tackling)

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaConfigurationModelParser.java
marianogonzalez, line 74: the agreement on the call was that all instances of this would have a TODO pointing to the issue that tracks using the MMV as a baseline

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -329,21 +326,26 @@ private static MuleVersion calculateOutputMinMuleVersion(Type outputType) {
     return calculatedMMV;
   }
 
-  private static MuleVersion calculateFieldMinMuleVersion(ExtensionParameter field) {
-    MuleVersion calculatedMMV = getMinMuleVersionFromAnnotations(field).orElse(FIRST_MULE_VERSION);
+  private static MuleVersion calculateFieldMinMuleVersion(ExtensionParameter field, Set<String> seenTypes) {
     Type parameterType = field.getType();
+    if (seenTypes.contains(parameterType.getTypeName())) {
+      return FIRST_MULE_VERSION;
```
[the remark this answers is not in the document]
SebaElizalde: the current type has not finished resolving (think of the recursive stack) so we can't return that; thus we return the first version to cut the recursion. The first call will analyse all other fields (the ones that are not recursive), so no information is missed by doing this.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
```diff
@@ -111,7 +113,7 @@ public static MuleVersion calculateConfigMinMuleVersion(Type config, MuleVersion
     calculatedMMV = max(calculatedMMV, config.getAnnotations().map(JavaParserUtils::getEnforcedMinMuleVersion)
         .reduce(FIRST_MULE_VERSION, JavaParserUtils::max));
     for (FieldElement field : config.getFields()) {
-      calculatedMMV = max(calculatedMMV, calculateFieldMinMuleVersion(field));
+      calculatedMMV = max(calculatedMMV, calculateFieldMinMuleVersion(field, new HashSet<>()));
```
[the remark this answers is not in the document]
SebaElizalde: Ending the recursion is not the same as having a cache. Read the above comment. Think of the recursive stack, this is a way to end that recursion. As we haven't finished calculating the mule version for the current component so it won't be present in the cache (it is added to the cache once it is calculated).
I agree with adding a cache for performance on a separate issue.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaConfigurationModelParser.java
SebaElizalde: adding it

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java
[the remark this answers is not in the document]
marianogonzalez: and what if they are all recursive? or a single one that is recursive? Also, if the set is just for recursion control, then rename the arg to reflect that, otherwise future devs  could see this as just a cache.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaSourceModelParser.java
```diff
@@ -100,6 +102,7 @@ public JavaSourceModelParser(ExtensionElement extensionElement,
 
     parseStructure();
     collectAdditionalModelProperties();
+    this.minMuleVersion = getMinMuleVersion().orElse(FIRST_MULE_VERSION);
```
marianogonzalez, line 105: I don't understand this assignment. You're calling getMinMuleVersion() which is implemetned as:

```
@Override
  public Optional<MuleVersion> getMinMuleVersion() {
    return of(calculateSourceMinMuleVersion(sourceElement));
  }
  ```

Clients of the parser will call that getter, and the calculation will be performed everytime. Besides the efficiency concerns around that, why calculate this field if it isn't going to be returned ?

**Review by marianogonzalez — CHANGES_REQUESTED** [review]

File: modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaSourceModelParser.java (hunk 8 above)
Comment (SebaElizalde) [reply modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaSourceModelParser.java]: You are right, fixing this.

File: modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java (hunk 6 above)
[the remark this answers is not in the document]
Comment (SebaElizalde) [reply modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/utils/JavaParserUtils.java:332]: works for both cases. My point was that the first call will analyse the class completely, so the second call (the recursive one) returning the first mule version does not cause loss of information.
Ok, renaming the arg.

File: modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaConnectionProviderModelParser.java [hunk 9]
```diff
@@ -95,6 +97,8 @@ public JavaConnectionProviderModelParser(JavaExtensionModelParser extensionModel
         .orElse(ExtensionModel.class.getClassLoader());
 
     collectAdditionalModelProperties();
+    // TODO W-10621050
+    this.minMuleVersion = calculateConnectionProviderMinMuleVersion(element);
```
ndinu, line 101: You can already remove this code and uncomment the new lines that have the W-12398760 mentioned

SebaElizalde: Done

ndinu: You probably will need to update the serialized extension models after this.

marianogonzalez (approved):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/loader/parser/java/JavaConfigurationModelParser.java
```diff
@@ -193,8 +199,6 @@ public Optional<DisplayModel> getDisplayModel() {
 
   @Override
   public Optional<MuleVersion> getMinMuleVersion() {
-    return of(calculateConfigMinMuleVersion(configElement,
-                                            getContainerAnnotationMinMuleVersion(extensionElement, Configurations.class,
-                                                                                 Configurations::value, configElement)));
+    return of(this.minMuleVersion);
```
ndinu, line 202: This trusts that the calculation will always retrieve a value, what about using ofNullable?

marianogonzalez (approved):

marianogonzalez (approved):

PR #12188, opened 2023-01-24
W-12128703: Fix deadlock waiting for previous flow ref to be completed

hyerans:


hyerans:
--validate

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/FluxSinkWrapper.java
```diff
@@ -0,0 +1,83 @@
...
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+
+package org.mule.runtime.core.internal.processor.strategy;
+
+import org.mule.runtime.core.api.event.CoreEvent;
+import reactor.core.Disposable;
```
fsgonz, line 11: import order.

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/FluxSinkWrapper.java
```diff
@@ -0,0 +1,83 @@
...
+import reactor.core.publisher.FluxSink;
+import reactor.util.context.Context;
+
+import java.util.function.LongConsumer;
+
+public class FluxSinkWrapper implements FluxSink<CoreEvent> {
```
fsgonz, line 17: Add a jdoc indicating that this is wrapper for verifying that the sink is being used. and a @see to the class where it is used.

fsgonz (changes requested):

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/FluxSinkWrapper.java
hyerans: re-arranged the import order based on the convention.

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/FluxSinkWrapper.java
hyerans: added java doc with @see

core/src/main/java/org/mule/runtime/core/internal/context/DefaultMuleContext.java
```diff
@@ -1537,6 +1539,16 @@ private static void configureForeachRouterRejectsMapExpressionFeatureFlag() {
...
   }
 
+  /**
+   * Configures the {@link MuleRuntimeFeature#USE_TRANSACTION_SINK_INDEX} feature flag.
+   *
+   * @since 4.0.0
+   */
+  private static void configureUseTransactionSinkIndex() {
+    FeatureFlaggingRegistry featureFlaggingRegistry = FeatureFlaggingRegistry.getInstance();
+    featureFlaggingRegistry.registerFeatureFlag(USE_TRANSACTION_SINK_INDEX, minMuleVersion("4.0.0"));
```
fsgonz, line 1549: change this to alsways return -> true, featureContext -> true instead of minMuleVersion("4.0.0) (this is the best way to do it, my mistake)

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/AbstractCachedThreadReactorSinkProvider.java
```diff
@@ -25,18 +25,39 @@ public abstract class AbstractCachedThreadReactorSinkProvider implements Reactor
 
   private static final int THREAD_CACHE_TIME_LIMIT_IN_MINUTES = 60;
   private static final int TRANSACTION_CACHE_TIME_LIMIT_IN_MINUTES = 10;
+  private boolean sinkIndexEnabled;
 
-  private final Cache<Thread, FluxSink<CoreEvent>> sinks =
+  private final Cache<String, FluxSinkWrapper> sinks =
```
fsgonz, line 30: If possible, we can add a way to make the keys weak reference so that not to add pressure on memory.

fsgonz (changes requested):

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/AbstractCachedThreadReactorSinkProvider.java
```diff
@@ -25,18 +25,39 @@ public abstract class AbstractCachedThreadReactorSinkProvider implements Reactor
...
+
+  private final Cache<Transaction, FluxSink<CoreEvent>> legacySinksNestedTx =
       Caffeine.newBuilder().weakKeys()
           .removalListener((RemovalListener<Transaction, FluxSink<CoreEvent>>) (transaction, coreEventFluxSink,
                                                                                 removalCause) -> coreEventFluxSink.complete())
           .expireAfterAccess(TRANSACTION_CACHE_TIME_LIMIT_IN_MINUTES, MINUTES).build();
 
+  public AbstractCachedThreadReactorSinkProvider() {}
+
+  public AbstractCachedThreadReactorSinkProvider(boolean enabled) {
```
elrodro83, line 57: use a strategy intead of having this flag queried everywhere

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/AbstractCachedThreadReactorSinkProvider.java
```diff
@@ -51,9 +72,43 @@ protected void invalidateAll() {
...
-      return sinks.get(currentThread(), t -> createSink());
+      if (sinkIndexEnabled) {
+        return getSimpleFluxSinkWrapper();
+      } else {
+        return legacySinks.get(currentThread(), t -> new FluxSinkWrapper(createSink()));
+      }
+    }
+  }
+
+  private FluxSink<CoreEvent> getNestedFluxSinkWrapper(TransactionCoordination txCoord) {
```
elrodro83, line 89: `getNestedTxFluxSinkWrapper`

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/AbstractCachedThreadReactorSinkProvider.java
```diff
@@ -51,9 +72,43 @@ protected void invalidateAll() {
...
+    int index = 0;
+    while (true) {
+      FluxSinkWrapper fluxSinkWrapper =
+          sinksNestedTx.get(txCoord.getTransaction().getId() + "-" + index, t -> new FluxSinkWrapper(createSink()));
```
elrodro83, line 93: can the transacion id be repeated? form what i see in the code it is a uuid

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/AbstractCachedThreadReactorSinkProvider.java
```diff
@@ -51,9 +72,43 @@ protected void invalidateAll() {
...
+        index++;
+        continue;
+      }
+      return fluxSinkWrapper;
+    }
+  }
+
+  private FluxSinkWrapper getSimpleFluxSinkWrapper() {
+    int index = 0;
+    while (true) {
```
elrodro83, line 104: can we put a reasonable limit to the iterations here?

elrodro83 (changes requested):

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/AbstractCachedThreadReactorSinkProvider.java
asanguinetti: I don't think it would be an issue if it eventually repeats, as long as we can find a sink that is not being used we can go ahead and use it I guess.
If I'm not getting it wrong, the transaction ID (as well as the thread ID in the other case) are only used for bucketing to make it less likely to find sinks that are in use, hence reducing the number of cache lookups.
I'm even going to throw into the board the possibility to just use the Thread ID here as well... just for simplifying things. It may also increase the chance of being able to reuse an already existing sink and avoid creating a new one.
It looks like it comes down to a trade-off between the likelihood of being able to reuse an existing Sink and the time it may take to find an available one.
That being said, it is a good idea to make sure we don't steal a Sink that might be chosen from a `getSink` being executed in a different thread, because the `isBeingUsed` check is not synchronized with effectively marking the Sink as used before handing it over.

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/AbstractCachedThreadReactorSinkProvider.java
```diff
@@ -51,9 +72,43 @@ protected void invalidateAll() {
...
+          sinksNestedTx.get(txCoord.getTransaction().getId() + "-" + index, t -> new FluxSinkWrapper(createSink()));
+      if (fluxSinkWrapper.isBeingUsed()) {
+        index++;
+        continue;
+      }
+      return fluxSinkWrapper;
+    }
+  }
+
+  private FluxSinkWrapper getSimpleFluxSinkWrapper() {
```
asanguinetti, line 102: I see a lot of common code between this method and `getNestedFluxSinkWrapper`.
Couldn't we parameterize it to receive the cache on which to lookup and the key prefix?

asanguinetti (changes requested):

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/AbstractCachedThreadReactorSinkProvider.java
fsgonz: In this case, we always use the sink from the same thread (even for nested transactions) because in the case of transactions no matter if they are nested or not everything executes in the same thread. That's why the isBeingUsed is not synchronized.

asanguinetti: Yes, I know, I was just pointing it out in case we decided to do something different in term of bucketing

fsgonz: The transactionId is used to guarantee that no work stealing is done inside reactor. If there is work stealing, the transactions that are handled through thread locals can be mixed up.

If the same sink is used and the next is executed from two different threads, after one thread executes everything, it can go on executing the chain for the event that the other thread enqueued (and it will not have the correct transaction information in the thread local). I think that was the original problem tackled by this class.

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/AbstractCachedThreadReactorSinkProvider.java
fsgonz: Probably we can use the MULE_FLOW_STACK_MAX_DEPTH property.

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/AbstractCachedThreadReactorSinkProvider.java
fsgonz: I don’t think it makes sense to extract an strategy for something related to internal use in this PR taking into account the difference in used is only due to a kill switch (that should be removed eventually)

asanguinetti (approved):

fsgonz:
--validate

eze210:
--validate

fsgonz (approved):

PR #12212, opened 2023-01-30
W-12336322: Span Status does not get propagated to exported source span

sofiamorseletto:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/source/SourceAdapter.java
```diff
@@ -328,12 +334,20 @@ public void onCompletion(CoreEvent event, Map<String, Object> parameters, Comple
       if (context.getTransactionHandle().isTransacted()) {
         callback = callback.before(v -> commit());
       }
```
sofiamorseletto, line 336: Adding these lines on onCompletion too because according to OTEL semantic [conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#status), the span status depends on the HTTP status code and since users can modify that code regardless of error, it should be checked also on the successful case

fsgonz:
--validate

PR #12231, opened 2023-02-07
W-12128703: Fix deadlock waiting for previous flow ref to be completed (#330)

fsgonz:


core/src/main/java/org/mule/runtime/core/internal/processor/strategy/StreamPerThreadSink.java
```diff
@@ -115,8 +175,10 @@ public BackPressureReason emit(CoreEvent event) {
   @Override
   public void dispose() {
     disposing = true;
-    sinks.asMap().values().forEach(sink -> sink.complete());
-    sinksNestedTx.asMap().values().forEach(sink -> sink.complete());
+    sinks.asMap().values().forEach(sinkList -> sinkList.forEach(FluxSinkWrapper::complete));
+    legacySinks.asMap().values().forEach(FluxSink::complete);
+    sinksNestedTx.asMap().values().forEach(sinkList -> sinkList.forEach(FluxSinkWrapper::complete));
+    legacySinksNestedTx.asMap().values().forEach(FluxSink::complete);
```
asanguinetti, line 181: I think we are missing the invalidateAll in the legacy ones down below on lines 200-201 and 211-212 (it doesn't let me put the commend directly there)

asanguinetti (changes requested):

core/src/main/java/org/mule/runtime/core/internal/processor/strategy/StreamPerThreadSink.java
fsgonz: Yes. Sorry. Missed that.

asanguinetti (approved):

PR #12232, opened 2023-02-07
W-12128703: Fix deadlock waiting for previous flow ref to be completed (#330)

fsgonz:


fsgonz:
Not cherry-picked to 4.3.x

PR #12233, opened 2023-02-09
W-12520448: Create profile for java 17 to configure surefire plugin

marianorean:


pom.xml
```diff
@@ -1185,5 +1185,39 @@
                 </pluginManagement>
             </build>
         </profile>
+        <profile>
+            <id>jdk-12+</id>
+            <activation>
+                <jdk>[12,)</jdk>
```
mule-pachanady, line 1191: I guess we can just focus on 17+ on only LTS versions

mule-pachanady (dismissed):

pom.xml
marianogonzalez: This should be a fixed 17

marianogonzalez (changes requested):

pom.xml
marianorean: done

marianogonzalez (approved):

PR #12250, opened 2023-02-22
Fix/w 12460123 4.4.x

acurci-at-mulesoft:


eze210 (changes requested):
Check the title before merge

eze210 (approved):

PR #12271, opened 2023-03-01
W-12450302: Mask PassPhrase in Logs

hyerans:


core/src/main/java/org/mule/transport/AbstractMessageDispatcher.java
```diff
@@ -116,6 +116,9 @@ public MuleEvent process(final MuleEvent event) throws MuleException
         }
         catch (Exception e)
         {
+            if (getEndpoint().getProperties().get("passphrase") != null) {
+                getEndpoint().getProperties().put("passphrase", "<<passphrase>>");
```
fsgonz, line 120: Doesn't this change directly the value of the endpoint? What happens if after failing once you send another request in test?

fsgonz (commented):
I would add a test for this?

PR #12273, opened 2023-03-02
W-12450302: Mask PassPhrase in Logs

hyerans:


core/src/main/java/org/mule/api/MuleException.java
```diff
@@ -199,9 +203,14 @@ public String getSummaryMessage()
         }
         StringBuilder buf = new StringBuilder(1024);
         buf.append(SystemUtils.LINE_SEPARATOR).append(EXCEPTION_MESSAGE_DELIMITER);
+        Pattern p = compile("passphrase=([^,])+");
```
fsgonz, line 206: Pass this logic to a method method in CredentialsMaskUtils.
Use a constant for the regex passphrase=([^,])+

fsgonz (changes requested):
Add a test for this.

fsgonz (approved):

PR #12275, opened 2023-03-02
W-12637122: Dispose region's class loader as part of its owner class loader's disposal'

rbourbonnavarro:


modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/RegionClassLoader.java
```diff
@@ -424,8 +424,21 @@ public final Enumeration<URL> findResources(final String name) throws IOExceptio
...
+   * @deprecated since 4.5
+   */
   @Override
+  @Deprecated
   public void dispose() {
+    disposeClassLoader(ownerClassLoader);
+    doDispose();
+  }
+
+  public void disposeFromOwnerClassLoader() {
```
elrodro83, line 437: package visibility, not public

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/MuleDeployableArtifactClassLoader.java
```diff
@@ -69,4 +69,20 @@ public List<ArtifactClassLoader> getArtifactPluginClassLoaders() {
             : fcl)
         .collect(toList());
   }
+
+  @Override
+  public void dispose() {
+    super.dispose();
+    if (isRegionClassLoaderMember(this)) {
```
elrodro83, line 76: when would this not be the case?

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/RegionClassLoader.java
```diff
@@ -424,8 +424,21 @@ public final Enumeration<URL> findResources(final String name) throws IOExceptio
     return new CompoundEnumeration<>(enumerations.toArray(new Enumeration[0]));
   }
 
+  /**
+   * @deprecated since 4.5
```
elrodro83, line 428: document the replacement

elrodro83 (changes requested):

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/RegionClassLoader.java
```diff
@@ -424,8 +425,25 @@ public final Enumeration<URL> findResources(final String name) throws IOExceptio
     return new CompoundEnumeration<>(enumerations.toArray(new Enumeration[0]));
   }
 
+  /**
+   * @deprecated since 4.5, use {@link #disposeFromOwnerClassLoader()} instead.
```
elrodro83, line 429: the idea would be for client code to jaut call `#dispose` on the app/domain classlaoder, right?

elrodro83 (changes requested):

modules/artifact/src/main/java/org/mule/runtime/module/artifact/api/classloader/RegionClassLoader.java
rbourbonnavarro: Yes, in the `disposeFromOwnerClassLoader()` Java doc I mention that the method is intended to be used by the owner class loader while being disposed.

elrodro83: i would mention that in this deprecation note instead of having that indirection through docs

elrodro83 (changes requested):

elrodro83 (approved):

PR #12290, opened 2023-03-08
W-12556497: Parallel-for-each gets stuck with global error handler op…

marianorean:
…timizations

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/privileged/processor/chain/AbstractMessageProcessorChain.java
+++ b/core/src/main/java/org/mule/runtime/core/privileged/processor/chain/AbstractMessageProcessorChain.java
@@ -24,6 +24,7 @@ import static org.mule.runtime.core.api.util.StreamingUtils.updateEventForStream
 import static org.mule.runtime.core.api.util.StringUtils.isBlank;
 
 import static org.mule.runtime.core.internal.event.NullEventFactory.getNullEvent;
+import static org.mule.runtime.core.internal.util.rx.RxUtils.REACTOR_RECREATE_ROUTER;
 import static org.mule.runtime.core.privileged.event.PrivilegedEvent.setCurrentEvent;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/mule/runtime/core/privileged/processor/chain/AbstractMessageProcessorChain.java
```diff
@@ -134,6 +135,7 @@ abstract class AbstractMessageProcessorChain extends AbstractExecutableComponent
       "Unexpected state. Error handler should be invoked with either an Event instance or a MessagingException. " +
           "This may lead to an event getting stuck, or even a processor may stop responding.";
   public static final String UNKNOWN = "unknown";
+  public static final String REACTOR_RECREATE_ROUTER = "recreateRouter";
```
fsgonz, line 138: add a comment to indicate the rationale behind this.

fsgonz (changes requested):

modules/core-components/src/main/java/org/mule/runtime/core/internal/routing/forkjoin/AbstractForkJoinStrategyFactory.java
```diff
@@ -76,6 +76,7 @@ public abstract class AbstractForkJoinStrategyFactory implements ForkJoinStrateg
 
   public static final String TIMEOUT_EXCEPTION_DESCRIPTION = "Route Timeout";
   public static final String TIMEOUT_EXCEPTION_DETAILED_DESCRIPTION_PREFIX = "Timeout while processing route/part:";
+  public static final String REACTOR_RECREATE_ROUTER = "recreateRouter";
```
fsgonz, line 79: Can't we pass this to RxUtils and use the same always?

fsgonz (changes requested):

fsgonz (dismissed):

fsgonz (approved):

PR #12292, opened 2023-03-09
W-12336322: Span Http Status does not get propagated to exported source span

fsgonz:


modules/tracing/mule-tracer-internal-impl/src/main/java/org/mule/runtime/tracer/impl/context/EventSpanContext.java
```diff
@@ -69,6 +70,9 @@ private InternalSpan resolveParentAsInternalSpan() {
   @Override
   public void setSpan(InternalSpan span, Assertion assertion) throws AssertionFailedException {
     assertion.assertOnSpan(currentSpan);
+    if (currentSpan.getName().equals(FLOW_DISPATCHING_HANDLING_SPAN_NAME)) {
```
IvanAndresFritzler, line 73: This is something that will be evaluated every time we add a span to the context. In order to avoid this (plus having to know about the dispatching span name) we could invert this and add a "Dependant Span"  to the current span (that in this case will be the dispatching span)? We can get rid of the listeners also. But this is the place where we clearly see that we are coupling things a bit.

IvanAndresFritzler (changes requested):

modules/tracing/mule-tracer-internal-impl/src/main/java/org/mule/runtime/tracer/impl/context/NoExportOnEndInternalSpan.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  @Override
+  public void addAttribute(String key, String value) {
+    delegate.addAttribute(key, value);
+  }
+
+  @Override
+  public void export() {
+    this.delegate.end(endTime);
```
IvanAndresFritzler, line 131: We should call to export here, right?

modules/tracing/mule-tracer-internal-impl/src/main/java/org/mule/runtime/tracer/impl/context/NoExportOnEndInternalSpan.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.function.BiConsumer;
+
+/**
+ * A {@link InternalSpan} wrapper to prevent exporting on end.
+ *
+ * @since 4.5.1, 4.6.0
+ */
+public class NoExportOnEndInternalSpan implements ExportableInternalSpan {
```
IvanAndresFritzler, line 30: We are not using inheritance but the ExecutionSpan is practically the same than this one, but also exporting at the end() call, right? Can we name them more closely?

IvanAndresFritzler (changes requested):

modules/tracing/mule-tracer-internal-impl/src/main/java/org/mule/runtime/tracer/impl/context/NoExportOnEndInternalSpan.java
fsgonz: I've removed the ExportableInternalSpan interface because taking into account this request for change I found that it can work without adding this.

modules/tracing/mule-tracer-internal-impl/src/main/java/org/mule/runtime/tracer/impl/context/NoExportOnEndInternalSpan.java
fsgonz: I've renamed ExecutionSpan to ExportOnEndExecutionSpan

IvanAndresFritzler (approved):

fsgonz:
--validate

PR #12315, opened 2023-03-14
W-12271999: Improve performance of ExtensionComponent by caching some static values

asanguinetti:
### Motivation
We found a performance regression introduced by [this other change](https://github.com/mulesoft/mule/commit/7b04c1c43de0131250925a7524cb080cabf70fd1) (config-refs with expressions).

The problem came from what we thought would be a relatively inexpensive operation (resolving a `StaticValueResolver`) was actually not. To be fair, the [resolution](https://github.com/mulesoft/mule/blob/2a87abccb12f634037000cc782b1e46318fa9b2f/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/resolver/StaticValueResolver.java#L49) itself was not expensive of course. The problem came from the fact that we were creating a dummy `ValueResolvingContext` from a Null Event every time ([here](https://github.com/mulesoft/mule/blob/2a87abccb12f634037000cc782b1e46318fa9b2f/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java#L456)).

cpu-trace-resolveConfigurationProviderStatically

Having noticed that, we realized that we were also doing [the same](https://github.com/mulesoft/mule/blob/2a87abccb12f634037000cc782b1e46318fa9b2f/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java#L495) for retrieving the configuration instance from a static configuration provider. This was previous to the regression. In this PR we also want to take the opportunity to improve that one.

### Proposal

The Null Event concept was coined to resolve things during initialization, where a real event was not available. Using a Null Event in runtime is a smell.
In this particular case, we are using it to comply with the `ValueResolver` interface, knowing that the resolved value actually does not depend on the Events (we know that because `StaticValueResolver#isDynamic` returns false).
With that in mind (subsequent invocations of the resolve method will return the same value) we can very well cache the result.
Similarly, if the resolved `ConfigurationProvider` is static, we can also cache the `ConfigurationInstance` provided.

### Results

The preliminary results in the Scatter Gather scenario are showing about 15% improvement. See this [Kibana dashboard](https://stats.perf.anypoint.mulesoft.com/goto/36a819e1becf99989d25ebdcf5878514).
It is probably augmented in this scenario vs others because it has many operations with configs for each request that goes through.

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
@@ -128,6 +128,7 @@ public abstract class ExtensionComponent<T extends ComponentModel> extends Abstr
   private final MetadataMediator<T> metadataMediator;
   private final ClassTypeLoader typeLoader;
   private final LazyValue<Boolean> requiresConfig = new LazyValue<>(this::computeRequiresConfig);
+  private final LazyValue<Boolean> usesDynamicConfiguration = new LazyValue<>(this::computeUsesDynamicConfiguration);
   private final LazyValue<Optional<ConfigurationProvider>> staticallyResolvedConfigurationProvider =
       new LazyValue<>(this::doResolveConfigurationProviderStatically);
 
@@ -484,6 +485,10 @@ public abstract class ExtensionComponent<T extends ComponentModel> extends Abstr
   }
 
   protected boolean usesDynamicConfiguration() {
+    return usesDynamicConfiguration.get();
+  }
+
+  private boolean computeUsesDynamicConfiguration() {
     // TODO W-11267571: if not being able to resolve the ValueResolver<ConfigurationProvider> at this point, hence
     // treating it as dynamic, ends up causing performance issues.
     return isConfigurationSpecified()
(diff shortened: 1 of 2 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
```diff
@@ -258,12 +263,6 @@ private void initConfigurationResolver() {
     // check for implicit provider
     findConfigurationProviderResolver().ifPresent(configurationProviderResolver::set);
 
-    Optional<ConfigurationInstance> staticConfiguration = getStaticConfiguration();
-    if (staticConfiguration.isPresent()) {
-      configurationResolver = event -> staticConfiguration;
-      return;
-    }
-
```
asanguinetti, line 266: This part is being removed because it had actually no effect. The method [getStaticConfiguration](https://github.com/mulesoft/mule/blob/2a87abccb12f634037000cc782b1e46318fa9b2f/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java#L489) will always return empty at this point because the `configurationResolver` is still `null`.

While removing it could be considered out of scope for this PR, leaving it would complicate things, because we are trying to cache the result, and we don't want it to be always empty. I could use the internal, uncached, `doGetStaticConfiguration` but that will just add more to the confusion.

We could consider trying to actually fix this so that this optimization attempt actually works. But in the end, it is probably not a big win vs calling resolveConfigFromProvider which will just do a get on the provider. Considering the optimization was only meant for static ConfigurationProviders, that should be a direct return of a cached value. Notice that here we don't deal with the Null Event cost because the real event is given to the resolver.

asanguinetti (commented):
Leaving a comment for the reviewers.

PR #12316, opened 2023-03-15
W-12658023: User Story A - API Module for Tracing Levels

fsgonz:


modules/tracing/mule-tracer-customization-api/src/main/java/org/mule/runtime/tracer/configuration/api/InitialSpanInfoBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+   */
+  InitialSpanInfoBuilder withForceNotExportUntil(String componentName);
+
+  /**
+   * @return the resulting {@link InitialSpanInfo}.
+   */
+  InitialSpanInfo build();
+
+  /**
+   * Indicates that the export should not be exported.
```
sofiamorseletto, line 52:
```suggestion
   * Indicates that the span should not be exported.
```

modules/tracing/mule-tracer-customization-api/src/main/java/org/mule/runtime/tracer/configuration/api/InitialSpanInfoBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+  InitialSpanInfoBuilder withSuffix(String suffix);
+
+  /**
+   * Indicates that no children will be exported till a span is found with the component name passed as parameter.
+   *
+   * @param componentName the component name when this will start exporting again
+   *
+   * @return he {@link InitialSpanInfoBuilder}
+   */
+  InitialSpanInfoBuilder withForceNotExportUntil(String componentName);
```
sofiamorseletto, line 44: I think it sounds better with No instead of Not
```suggestion
  InitialSpanInfoBuilder withForceNoExportUntil(String componentName);
```

modules/tracing/mule-tracer-customization-impl/src/test/java/org/mule/runtime/tracer/configuration/internal/builder/GenericInitialSpanInfoBuilderTestCase.java
```diff
@@ -0,0 +1,24 @@
...
+import io.qameta.allure.Feature;
+import io.qameta.allure.Story;
+import org.mule.runtime.tracer.configuration.api.InitialSpanInfoBuilder;
+
+@Feature(PROFILING)
+@Story(DEFAULT_CORE_EVENT_TRACER)
+public class GenericInitialSpanInfoBuilderTestCase extends InitialSpanInfoBuilderTestCase {
+
+  @Override
+  InitialSpanInfoBuilder getComponentInitialSpanInfoBuilder() {
```
sofiamorseletto, line 21: In this case, it wouldn't be a ComponentInitialSpanInfoBuilder, right? Maybe the name can be changed to getInitialSpanInfoBuilder

sofiamorseletto (dismissed):

core/src/main/java/org/mule/runtime/core/internal/policy/PolicyNextActionMessageProcessor.java
```diff
@@ -117,7 +117,7 @@ public Publisher<CoreEvent> apply(Publisher<CoreEvent> eventPub) {
                 : policyEventMapper.onOperationPolicyNext(event))
             .transform((ReactiveProcessor) ((Reference) ctx.get(POLICY_NEXT_OPERATION)).get()));
       }
-    }), policyNextErrorHandler(), NO_EXPORT_CHILD_NAMED_SPAN_BASED_ON_PARENT_SPAN_CHILD_SPAN_CUSTOMIZATION_INFO);
+    }), policyNextErrorHandler(), initialSpanInfoBuilderProvider.getComponentInitialSpanInfoBuilder(this).withNoExport().build());
```
IvanAndresFritzler, line 120: This is where I think that a break of encapsulation is seen. If we are going to base the export levels feature on the existing InitialExportInfo. a withNoExport method should not be  present on the builders of the InitialSpanInfo but be part of the different export strategies logic. If this is meant to be changed in next PRs lets add a TODO with the task ID and move on, it's ok to me, but I think that it would be better to have an getInitialExportInfoProvider than a getInitialSpanInfoBuilerProvider, because we are not changing anything from the InitialSpanInfo but it's composed InitialExportInfo.

IvanAndresFritzler (changes requested):

core/src/main/java/org/mule/runtime/core/internal/policy/PolicyNextActionMessageProcessor.java
fsgonz: I've changed this. In the "execution engine" you directly retrieve the InitialSpanInfo from an injectable (managed) component without indicating anything regarding the exportability.

Within the implementation for the InitialSpanInfoProvider you have a component ExportInitialSpanInfoProvider that is responsibility for indicating if it is exportable and how.

For the moment an implementation for the default level is implemented. As for the roadmap the other implementations will be handled in [User Story B](https://github.com/mulesoft/mule/pull/12316/files#diff-0ea6de91b7f846fa6f2ce6c659bdd5f2c8a406f06144f1e83a2783a91bb5540cR24) (next sprint).

And eventually this is component will be inyected according to the retrievable configuration (verify the user stories for that)

IvanAndresFritzler (changes requested):

core/src/main/java/org/mule/runtime/core/internal/construct/AbstractPipeline.java
```diff
@@ -148,6 +150,8 @@ public AbstractPipeline(String name, MuleContext muleContext, MessageSource sour
     try {
       interceptorManager = ((MuleContextWithRegistry) muleContext).getRegistry().lookupObject(InterceptorManager.class);
       notificationFirer = ((MuleContextWithRegistry) muleContext).getRegistry().lookupObject(NotificationDispatcher.class);
+      initialSpanInfoBuilderProvider =
```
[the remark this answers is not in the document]
fsgonz: It's probable that in some cases the lifecycle methos are used programmatically. I'll add a TODO to investigate this three lines.

IvanAndresFritzler (approved):

IvanAndresFritzler (approved):

PR #12336, opened 2023-03-24
W-12743226: Opentelemetry marshaller does not group spans by Resource and Instrumentation Scope

fsgonz:


sofiamorseletto:
Is it possible to add tests to verify that the spans are being grouped correctly?

fsgonz:
--validate

PR #12341, opened 2023-03-28
W-12694901: encapsulate `tests-runner` container and services class loader generation

rbourbonnavarro:


modules/container/src/main/java/org/mule/runtime/container/api/TestContainerClassLoaderAssembler.java
```diff
@@ -0,0 +1,32 @@
...
+public interface TestContainerClassLoaderAssembler extends AutoCloseable {
+
+  static TestContainerClassLoaderAssembler create(List<String> extraBootPackages, Set<String> extraPrivilegedArtifacts,
+                                                  List<URL> urls) {
+    return new DefaultTestContainerClassLoaderAssembler(extraBootPackages, extraPrivilegedArtifacts, urls);
+  }
+
+  ArtifactClassLoader createContainerClassLoader();
+
+  ArtifactClassLoader getContainerClassLoader();
```
elrodro83, line 26: what would be the difference between these 2?

modules/container/src/main/java/org/mule/runtime/container/internal/TestContainerClassLoaderFactory.java
```diff
@@ -0,0 +1,142 @@
...
+import com.google.common.collect.ImmutableSet;
+
+/**
+ * Extends the default {@link ContainerClassLoaderFactory} for testing in order to add boot packages and build a
+ * {@link ClassLoader} for the container that do resolves classes instead of delegating to its parent and also allows to create
+ * the container {@link ClassLoaderLookupPolicy} based on a {@link ClassLoader}.
+ *
+ * @since 4.6
+ */
+public class TestContainerClassLoaderFactory extends ContainerClassLoaderFactory implements AutoCloseable {
```
elrodro83, line 35: can we configure the behaviour on the ContainerClassLoaderFactory instead of subclassing it?

modules/container/src/main/java/org/mule/runtime/container/internal/TestContainerModuleDiscoverer.java
```diff
@@ -0,0 +1,33 @@
...
+package org.mule.runtime.container.internal;
+
+import java.util.List;
+
+/**
+ * Discovers container modules including both productive and test modules
+ *
+ * @since 4.6
+ */
+public class TestContainerModuleDiscoverer extends ContainerModuleDiscoverer {
```
elrodro83, line 16: we really need the Modulediscoverer to be ecapsulated, since we won't be using it with JPMS

elrodro83 (changes requested):

modules/container/src/main/java/org/mule/runtime/container/api/TestContainerClassLoaderAssembler.java
rbourbonnavarro: I ended up removing this in favour of encapsulating the usage of the `ContainerOnlyLookupStrategy`.

modules/container/src/main/java/org/mule/runtime/container/internal/TestContainerModuleDiscoverer.java
rbourbonnavarro: Already encapsulated.

modules/container/src/main/java/org/mule/runtime/container/internal/TestContainerClassLoaderFactory.java
rbourbonnavarro: This child class modifies basically everything about the creation of the container class loader. It's not possible to parameterize it and embed it in the parent class.

elrodro83 (changes requested):

PR #12342, opened 2023-03-29
W-12622230:  avoid mock/spy to InputStream implementations

mule-pachanady:
spy verify will not work in Java 17 without explicit add-opens

USing approach similar to [this](https://github.com/mulesoft/mule-artifact-ast/pull/399)


tests/unit/src/main/java/org/mule/tck/core/internal/serialization/CustomByteArrayInputStream.java
```diff
@@ -0,0 +1,36 @@
...
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.tck.core.internal.serialization;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class CustomByteArrayInputStream extends InputStream {
```
fsgonz, line 13: Use a more declarative name for the class. This is not necessarily a ByteInputStream that is decorated, it is a wrapper/decorator for a InputStream.

Something like CloseCounterInputStreamWrapper can be an option.

Add a jdoc indicating the driver behind this (avoid spy for java 17)

tests/unit/src/main/java/org/mule/tck/core/internal/serialization/CustomByteArrayInputStream.java
```diff
@@ -0,0 +1,36 @@
...
+
+  private final InputStream inputStream;
+
+  private AtomicInteger closeCount = new AtomicInteger();
```
fsgonz, line 17: this can be final.

tests/unit/src/main/java/org/mule/tck/core/internal/serialization/CustomByteArrayInputStream.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright (c) MuleSoft, Inc.  All rights reserved.  http://www.mulesoft.com
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.tck.core.internal.serialization;
+
+import java.io.IOException;
```
fsgonz, line 9: This is just a comment (not sure about this): this is not a complete decorator. It can be used for testing purposes only in the case that the this specific read method is invoked (this will failed for not overridden methods). If it will not be generally used, I would declare it as an inner class in the test.

fsgonz (changes requested):

mule-pachanady:
We dont need strategy, I will update this to use mockito-inline

fsgonz (approved):

mule-pachanady:
Thanks Fabi!

mule-pachanady:
--validate

mule-pachanady:
--validate

mule-pachanady:
--validate

PR #12344, opened 2023-03-29
W-12665132: Remove OpenTelemetryResources.getTracer method

sofiamorseletto:


modules/tracing/mule-tracer-exporter-impl/src/test/java/org/mule/runtime/tracer/impl/exporter/OpenTelemetryExporterConfigTestCase.java
```diff
@@ -307,4 +318,20 @@ public List<ExportTraceServiceRequest> getTraceRequests() {
       return traceRequests;
     }
   }
+
+  private static Tracer getTracer(SpanExporterConfiguration spanExporterConfiguration, String serviceName)
```
fsgonz, line 322: I think this code does not effectly test anything now from our code.
What needs to be done is to change the code so that instead of a tracer [you get a SpanProcessor through this method](https://github.com/mulesoft/mule/blob/master/modules/tracing/mule-tracer-exporter-impl/src/main/java/org/mule/runtime/tracer/impl/exporter/optel/resources/OpenTelemetryResources.java#L107)
and make them export according to GRPC and HTTP.

fsgonz (changes requested):

PR #12346, opened 2023-03-29
W-12771370: Handle null values during masking

hyerans:


fsgonz (changes requested):

core/src/main/java/org/mule/api/util/CredentialsMaskUtil.java
```diff
@@ -115,6 +115,7 @@ public static String maskPasswordAttribute(String password)
      * @return input with passphrase masked
      */
     public static String maskPassPhrase(String input) {
```
fsgonz, line 117: Use formatter for mule 3.

fsgonz (changes requested):

fsgonz (approved):

fsgonz (changes requested):
Could you add a test?

fsgonz (approved):

PR #12365, opened 2023-04-10
W-12761115: `xml-apis` shadows classes already present in Java 11/17 `java.xml` module

elrodro83:
...effectively causing a split package error when starting up as module

### DIFF
--- a/modules/reboot/src/main/java/org/mule/runtime/module/reboot/internal/DefaultMuleClassPathConfig.java
+++ b/modules/reboot/src/main/java/org/mule/runtime/module/reboot/internal/DefaultMuleClassPathConfig.java
@@ -7,6 +7,8 @@
 package org.mule.runtime.module.reboot.internal;
 
 import static java.lang.System.getProperty;
+import static java.util.Arrays.asList;
+import static java.util.Collections.emptyList;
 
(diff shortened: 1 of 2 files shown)


modules/reboot/src/main/java/org/mule/runtime/module/reboot/internal/DefaultMuleClassPathConfig.java
```diff
@@ -102,14 +112,11 @@ public void addFile(File jar) {
...
-        try {
-          return pathname.getCanonicalPath().endsWith(".jar");
-        } catch (IOException e) {
-          throw new RuntimeException(e.getMessage());
-        }
+    File[] jars = path.listFiles((FileFilter) pathname -> {
+      try {
+        return pathname.getCanonicalPath().endsWith(".jar");
+      } catch (IOException e) {
+        throw new RuntimeException(e.getMessage());
```
eze210, line 119: Isn't it better to create the RuntimeException with the IOException as a cause instead?

eze210 (changes requested):

eze210 (approved):

eze210 (approved):

PR #12370, opened 2023-04-10
W-13005445: Remove ignore override once we support lazy initialization of configurations referenced by expressions.

alopezbizon:


modules/extensions-xml-support/src/test/java/org/mule/test/functional/LazyInitExpressionsOnConfigRefTestCase.java
```diff
@@ -34,20 +34,6 @@ public boolean enableLazyInit() {
...
-    super.setPayloadWithExpression();
-  }
-
-  @Ignore("W-11525405: remove ignore override once we support lazy initialization of configurations referenced by expressions")
-  @Test
-  @Override
-  public void setPayloadWithExpressionResolvingToIncompatibleConfigFails() throws Exception {
-    super.setPayloadWithExpressionResolvingToIncompatibleConfigFails();
-  }
-
```
alopezbizon, line 50: It looks like we are removing tests but we are actually unignoring inherited tests.

PR #12385, opened 2023-04-17
W-12732035: replace `jsonassert` with our forked version

rbourbonnavarro:


modules/extensions-mule-support/pom.xml
```diff
@@ -16,6 +16,7 @@
 
     <properties>
         <formatterConfigPath>../../formatter.xml</formatterConfigPath>
+        <jsonassert.version>1.5.2</jsonassert.version>
```
elrodro83, line 19: move the property to the base pom

modules/pom.xml
```diff
@@ -14,6 +14,7 @@
 
     <properties>
         <formatterConfigPath>../formatter.xml</formatterConfigPath>
+        <jsonassert.version>1.5.2</jsonassert.version>
```
elrodro83, line 17: root root

elrodro83 (changes requested):

modules/pom.xml
rbourbonnavarro: Done.

elrodro83 (approved):

PR #12386, opened 2023-04-17
W-13049081: Remove test DefaultTlsContextFactoryTestCase#defaultIncludesTls12Ciphers

eze210:


modules/tls/src/test/java/org/mule/test/module/tls/DefaultTlsContextFactoryTestCase.java
```diff
@@ -202,13 +202,6 @@ public void cannotMutateEnabledCipherSuites() throws InitialisationException {
...
-  @Test
-  public void defaultIncludesTls12Ciphers() throws Exception {
-    assumeThat(IS_JAVA_1_8, is(true));
-
-    defaultIncludesDEfaultTlsVersionCiphers("TLSv1.2");
-  }
-
   @Test
   public void defaultIncludesTls13Ciphers() throws Exception {
     // For versions greater than 8, the default is TLS 1.3
```
asanguinetti, line 207:
```suggestion
    // For versions greater than 8u352, the default is TLS 1.3
```
Also, we need to remove the next assume so this is tested in Java 8 also. A TODO may also be valid if we want to do the "if >= 8u352" thing.

PR #12398, opened 2023-04-19
W-12980682: Update groovy used at build time to 4.x

elrodro83:


elrodro83:
--validate

elrodro83:
--validate

rbourbonnavarro (approved):

rbourbonnavarro (changes requested):

rbourbonnavarro (dismissed):

pom.xml
```diff
@@ -237,7 +237,7 @@
         <surefire.args>${surefire.args.base}</surefire.args>
 
         <commonsCollectionsVersion>3.2.2</commonsCollectionsVersion>
-        <groovyVersion>3.0.7</groovyVersion>
+        <groovyVersion>4.0.11</groovyVersion>
```
asanguinetti, line 240: Be aware that this property is also defined in the BOM:
https://github.com/mulesoft/mule-runtime-bom/blob/master/pom.xml#L55

I'm not sure if this change will end up taking precedence and potentially affecting the version of [this](https://github.com/mulesoft/mule-runtime-bom/blob/master/pom.xml#L153).

pom.xml
```diff
@@ -237,7 +237,8 @@
         <surefire.args>${surefire.args.base}</surefire.args>
 
         <commonsCollectionsVersion>3.2.2</commonsCollectionsVersion>
-        <groovyVersion>3.0.7</groovyVersion>
```
asanguinetti, line 240: Should we keep this one just in case it was overriding the BOM one? or have you verified that the one from the BOM (3.0.13) was always being used?

asanguinetti (approved):

PR #12399, opened 2023-04-20
W-12586811: Add a log to indicate that the spans are exported when op…

alopezbizon:
…tel spans export is enabled

### DIFF
--- a/modules/tracing/mule-tracer-exporter-impl/src/main/java/org/mule/runtime/tracer/exporter/impl/OpenTelemetrySpanExporterFactory.java
+++ b/modules/tracing/mule-tracer-exporter-impl/src/main/java/org/mule/runtime/tracer/exporter/impl/OpenTelemetrySpanExporterFactory.java
@@ -95,7 +95,7 @@ public class OpenTelemetrySpanExporterFactory implements SpanExporterFactory, Di
 
   protected SpanProcessor resolveOpenTelemetrySpanProcessor() {
     if (isExportEnabled()) {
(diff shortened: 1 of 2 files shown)


alopezbizon:
--validate

alopezbizon:
--validate

fsgonz (changes requested):
Add a log (in debug level for all the properties in OpentelemetryResources not only for batch size and exporter batch queue size. Basically log everything that is retrieved from the configuration layer.

modules/tracing/mule-tracer-exporter-impl/src/main/java/org/mule/runtime/tracer/exporter/impl/OpenTelemetrySpanExporterFactory.java
```diff
@@ -91,6 +94,7 @@ public SpanExporter getSpanExporter(InternalSpan internalSpan, InitialSpanInfo i
 
   protected SpanProcessor resolveOpenTelemetrySpanProcessor() {
     if (isExportEnabled()) {
+      LOGGER.info("{} is enabled.", MULE_OPEN_TELEMETRY_EXPORTER_ENABLED);
```
sofiamorseletto, line 97: Maybe we could log a message instead of the name of the property for this case, like "Mule OpenTelemetry Exporter is enabled"

fsgonz (commented):
Apart from Sofi's comment for the debug logging only add the endpoint and the protocol used.

fsgonz (approved):

fsgonz:
--validate

PR #12416, opened 2023-04-28
W-12625688: Avoid errors caused by using mock/spy to ClassLoader

hyerans:


eze210:
--validate

eze210:
There is a test (at least) that failed for me in spring-config module for the same reason (`LazyComponentInitializerAdapterTestCase`). Can you add the dependency there too?
I added it in another PR (#12420), but I think it makes more sense to add it here.

hyerans:
> There is a test (at least) that failed for me in spring-config module for the same reason (`LazyComponentInitializerAdapterTestCase`). Can you add the dependency there too? I added it in another PR (#12420), but I think it makes more sense to add it here.

Working on it now. Thanks Eze!

hyerans:
--validate

modules/extensions-support/pom.xml
```diff
@@ -249,6 +249,12 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-inline</artifactId>
+            <version>4.11.0</version>
```
elrodro83, line 255: use the `${mockitoVersion}` property

modules/launcher/pom.xml
```diff
@@ -77,5 +77,11 @@
             <scope>test</scope>
             <type>test-jar</type>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-inline</artifactId>
+            <version>4.11.0</version>
```
elrodro83, line 83: use the `${mockitoVersion}` property

elrodro83 (changes requested):

elrodro83 (approved):

PR #12418, opened 2023-05-02
W-13136310: NPE on scheduler due to propagation of distributed trace context

fsgonz:


fsgonz:
--validate

fsgonz:
--validate

sofiamorseletto (approved):

IvanAndresFritzler (commented):
We should also check if we can leverage this NoOp also with the sources that are not a scheduler but also do not receive a remote context.

IvanAndresFritzler (changes requested):
Requesting changes (see previous comments).

IvanAndresFritzler (approved):

fsgonz:
We've verify that for the cases of internal scheduler the callback context is created in a different way than in the sdk. I'll create a task to verify if that can be optimized.

PR #12431, opened 2023-05-08
W-13065364: User Story A - API Module for Metrics

fsgonz:


modules/metrics/mule-metrics-api/src/main/java/org/mule/runtime/metrics/api/instrument/InstrumentType.java
```diff
@@ -0,0 +1,29 @@
...
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.metrics.api.instrument;
+
+/**
+ * Enum for the differnt {@link InstrumentType}
+ */
+public enum InstrumentType {
```
IvanAndresFritzler, line 12: It's not being used yet, but is this specificity necessary here? LONG_COUNTER and LONG_UP_DOWN_COUNTER instead of just COUNTER? The other aspects seems to be implementation related at this point.

IvanAndresFritzler (changes requested):

modules/metrics/mule-metrics-api/src/main/java/org/mule/runtime/metrics/api/instrument/InstrumentType.java
fsgonz: Not necessary now. I'll remove it. LONG_COUNTER and LONG_UP_DOWN_COUNTER is a distinction of Optel. Probably if we don't make that distinction in our api then we will have problems if we open the api. But I'll remove the type enum.

IvanAndresFritzler (approved):

PR #12454, opened 2023-05-11
W-13065380: User Story C and D - Metric Export API and Implementation following OTel

sofiamorseletto:


modules/metrics/mule-metrics-exporter-api/src/main/java/module-info.java
```diff
@@ -0,0 +1,20 @@
...
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+
+/**
+ * Definitions for Mule Metric exporter.
+ *
+ * @moduleGraph
+ * @since 4.5
```
sofiamorseletto, line 12: I set 4.5 to everything since I believe this will be the version we'll end up releasing this in, right?

fsgonz: Yes, we have to change all those.

modules/metrics/mule-metrics-api/src/main/java/org/mule/runtime/metrics/api/instrument/builder/LongCounterBuilder.java
```diff
@@ -31,6 +31,15 @@ public interface LongCounterBuilder {
...
   LongCounterBuilder withUnit(String unit);
 
+  /**
+   * Sets the meter name for the instrument to build.
+   *
+   * @param meterName the meter name.
+   *
+   * @return the {@link LongCounterBuilder}.
+   */
+  LongCounterBuilder withMeterName(String meterName);
```
fsgonz, line 41: I would not set the meter name as part of the instrument builder. The meter will provide the appropriate builder associated to itself. In anycase it can internally add its name if it is useful for an instrument to know what its meter is.

modules/metrics/mule-metrics-api/src/main/java/org/mule/runtime/metrics/api/instrument/builder/LongUpDownCounterBuilder.java
```diff
@@ -43,6 +43,15 @@ public interface LongUpDownCounterBuilder {
...
   LongUpDownCounterBuilder withInitialValue(long initialValue);
 
+  /**
+   * Sets the meter name for the instrument to build.
+   *
+   * @param meterName the meter name.
+   *
+   * @return the {@link LongUpDownCounterBuilder}.
+   */
+  LongUpDownCounterBuilder withMeterName(String meterName);
```
fsgonz, line 53: I would not set the meter name as part of the instrument builder. The meter will provide the appropriate builder associated to itself. In anycase it can internally add its name if it is useful for an instrument to know what its meter is.

modules/metrics/mule-metrics-exporter-api/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+    <artifactId>mule-metrics-exporter-api</artifactId>
+    <version>4.6.0-SNAPSHOT</version>
+    <name>Mule Metrics Exporter Api</name>
+    <description>Definitions for Mule Metrics Exporter Api</description>
+    <dependencies>
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-api</artifactId>
+            <version>${opentelemetryVersion}</version>
+            <scope>compile</scope>
```
fsgonz, line 21: remove compile as it is the default scope. (seeking where we did this in other parts of the code, not in this PR).

modules/metrics/mule-metrics-exporter-api/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.mule.runtime</groupId>
+            <artifactId>mule-metrics-api</artifactId>
+            <version>${project.version}</version>
+            <scope>compile</scope>
```
fsgonz, line 27: remove compile as it is the default scope. (seeking where we did this in other parts of the code, not in this PR).

modules/metrics/mule-metrics-exporter-api/pom.xml
```diff
@@ -0,0 +1,39 @@
...
+        </dependency>
+    </dependencies>
+    
+
+    <properties>
```
fsgonz, line 32: I would move the properties before the dependencies.

modules/metrics/mule-metrics-exporter-impl/src/test/java/org/mule/runtime/metrics/exporter/impl/OpenTelemetryMeterExporterTestCase.java
```diff
@@ -0,0 +1,138 @@
...
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+import io.opentelemetry.sdk.metrics.data.MetricData;
+import io.opentelemetry.sdk.testing.exporter.InMemoryMetricExporter;
+import org.junit.Before;
+import org.junit.Test;
+
+public class OpenTelemetryMeterExporterTestCase {
```
fsgonz, line 32: add allure annotations

modules/metrics/mule-metrics-api/src/main/java/org/mule/runtime/metrics/api/instrument/builder/LongUpDownCounterBuilder.java
fsgonz: So in the implementation implementation of LongUpDownCounterBuilder that the meter provides you also has a reference to the meter to have in the instrument created.

fsgonz (changes requested):

fsgonz (approved):

PR #12459, opened 2023-05-15
W-12594415: HTTP Connector POST 303 Redirect passing body into the GET Request

marianorean:


modules/http-api/src/main/java/org/mule/runtime/http/api/client/HttpRequestOptions.java
```diff
@@ -51,6 +51,11 @@ static HttpRequestOptionsBuilder builder(HttpRequestOptions options) {
    */
   Optional<HttpAuthentication> getAuthentication();
 
+  /**
+   * @return if the request should contain a body even for methods without body semantics.
```
eze210, line 55: Please specify which are the HTTP methods without body semantics and a link to the RFC

eze210 (changes requested):

modules/http-api/src/main/java/org/mule/runtime/http/api/client/HttpRequestOptions.java
marianorean: done

PR #12465, opened 2023-05-16
Remove usages of ClassUtils#getFieldValue and ClassUtils#setFieldValue from tests.

asanguinetti:


core/src/main/java/org/mule/runtime/core/internal/retry/policies/SimpleRetryPolicy.java
```diff
@@ -170,6 +170,20 @@ public PolicyStatus applyPolicy(Throwable cause) {
...
+    return count;
+  }
+
+  /**
+   * @return how long to wait between retries.
+   */
+  public Duration getFrequency() {
+    return frequency;
+  }
+
```
asanguinetti, line 186: Let me know if you are ok with exposing these. I figure it is not only harmless but also kind of makes sense in the domain of the class to be able to provide the current values for these parameters. Also this impl class is internal.

Can't just make them package-private because it is in a different package and actually a different module than the test ([ReconnectionTestCase](https://github.com/mulesoft/mule/pull/12465/files#diff-1c456ccef41ce92baac50cdd0b886b5300dc4fd5d32d163b865bb215164e0046)). The test is not unitary but functional.

Ideally we would like to test that these parameters are respected when applying the policy but that would require a lot more complexity by using some fake schedulers, etc. We could do that but I feel like it is outside of this task's scope, and I'm not sure it is completely worth the effort.

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/scopes/ScopeExecutionTestCase.java
```diff
@@ -56,13 +60,15 @@ public void verifyProcessorInitialise() throws Exception {
...
     assertThat(getChainFirst, is(not(sameInstance(getChainSecond))));
 
-    Object firstChain = ClassUtils.getFieldValue(getChainFirst, "chain", false);
-    Object secondChain = ClassUtils.getFieldValue(getChainSecond, "chain", false);
-    assertThat(firstChain, is(sameInstance(secondChain)));
+    assertThat("Multiple executions of the same scope should yield the same processor instances",
+               getChainFirst.getMessageProcessors(),
+               contains(getChainSecond.getMessageProcessors().stream()
+                   .map(IsSame::sameInstance)
+                   .collect(toList())));
```
asanguinetti, line 71: Note that this is not exactly the same assertion as before.
We were comparing the MessageProcessorChain instances inside each SDK Chain.
Now we are just comparing the instances of each Processor inside each MPC.

I suppose this should be enough for our actual functional requirements. If you think it is not the case, then I guess I can add a getter for the MPC and move on. After all, the SDK Chain impls are internal.

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/source/PollingSourceWrapperTestCase.java
```diff
@@ -103,7 +101,6 @@ public class PollingSourceWrapperTestCase {
   @Before
   public void setUp() throws Exception {
     when(componentLocationMock.getRootContainerName()).thenReturn(TEST_FLOW_NAME);
-    setComponentLocationMock();
```
asanguinetti, line 106: Not needed because of the [InjectMocks](https://github.com/mulesoft/mule/pull/12465/files#diff-865978990efd2ce2ca126489962d054c8b98dac5ae4377f79ae6d7e1b94a0cbdR97) annotation which will introspect the instance's fields and set any that is compatible with any mock in this test class. Confirmed by debugging and actually running the tests (there would be an NPE if the componentLocation is null on the wrapper when calling onStart).

This is done by Mockito using reflection, it currently works even on JDK 17. I think even when mule-module-extensions-support is modularized this will continue working because the setAccessible will come from within the same module (even though it is not the same package).

asanguinetti (commented):
Leaving comments for the reviewer

modules/extensions-spring-support/src/test/java/org/mule/test/module/extension/scopes/ScopeExecutionTestCase.java
elrodro83: ok

elrodro83 (dismissed):

PR #12466, opened 2023-05-16
W-13152861: Fix split internal packages in mule modules

mbuchwald:


rbourbonnavarro (approved):

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/dsl/processor/xml/provider/CoreXmlNamespaceInfoProvider.java
```diff
@@ -4,7 +4,7 @@
  * license, a copy of which has been included with this distribution in the
  * LICENSE.txt file.
  */
-package org.mule.runtime.config.internal.dsl.processor.xml;
+package org.mule.runtime.config.internal.dsl.processor.xml.provider;
```
rbourbonnavarro, line 7: This should change only in one of the modules, otherwise we end with the same split package issue.

rbourbonnavarro (changes requested):

rbourbonnavarro (approved):

mbuchwald:
--validate

PR #12493, opened 2023-05-25
W-2354025 In Mule 3 Update snakeyaml to 2.0

mule-pachanady:
This is still not verified as I dont know the full implication of upgrading jrubyVersion

Jruby version update has many changes as well

Still bit WIP  

Also waiting for Apikit [PR](https://github.com/mulesoft/apikit/pull/653)

### DIFF
--- a/distributions/standalone/assembly-allowlist.txt
+++ b/distributions/standalone/assembly-allowlist.txt
@@ -120,8 +120,8 @@
 /mule-standalone-${productVersion}/lib/mule/mule-extensions-annotations-1.0.0-alpha-1.jar
 /mule-standalone-${productVersion}/lib/mule/mule-extensions-api-1.0.0-alpha-1.jar
 /mule-standalone-${productVersion}/lib/mule/mule-module-annotations-${productVersion}.jar
-/mule-standalone-${productVersion}/lib/mule/mule-module-apikit-3.9.5-SNAPSHOT.jar
-/mule-standalone-${productVersion}/lib/mule/mule-module-apikit-spi-3.9.5-SNAPSHOT.jar
+/mule-standalone-${productVersion}/lib/mule/mule-module-apikit-3.9.6-SNAPSHOT.jar
+/mule-standalone-${productVersion}/lib/mule/mule-module-apikit-spi-3.9.6-SNAPSHOT.jar
 /mule-standalone-${productVersion}/lib/mule/mule-module-atom-${productVersion}.jar
 /mule-standalone-${productVersion}/lib/mule/mule-module-bpm-${productVersion}.jar
 /mule-standalone-${productVersion}/lib/mule/mule-module-builders-${productVersion}.jar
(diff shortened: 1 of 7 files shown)


distributions/standalone/assembly-allowlist.txt
```diff
@@ -120,8 +120,8 @@
 /mule-standalone-${productVersion}/lib/mule/mule-extensions-annotations-1.0.0-alpha-1.jar
 /mule-standalone-${productVersion}/lib/mule/mule-extensions-api-1.0.0-alpha-1.jar
 /mule-standalone-${productVersion}/lib/mule/mule-module-annotations-${productVersion}.jar
-/mule-standalone-${productVersion}/lib/mule/mule-module-apikit-3.9.6-SNAPSHOT.jar
-/mule-standalone-${productVersion}/lib/mule/mule-module-apikit-spi-3.9.6-SNAPSHOT.jar
+/mule-standalone-${productVersion}/lib/mule/mule-module-apikit-3.10.0-SNAPSHOT.jar
```
fsgonz, line 123: Why is this version from apikit boost to 3.10? If there is a change needed probably thay have to cherry-pick it to their 3.9.x support branch.

modules/scripting-jruby/pom.xml
```diff
@@ -136,22 +136,22 @@
         <dependency>
             <groupId>org.yaml</groupId>
             <artifactId>snakeyaml</artifactId>
-            <version>1.31</version>
+            <version>2.0</version>
         </dependency>
         <dependency>
             <groupId>org.jruby.joni</groupId>
```
fsgonz, line 142: why are these updates needed?

pom.xml
```diff
@@ -116,7 +116,7 @@
         <skipPerformanceTests>true</skipPerformanceTests>
         <skipArchetypeTests>true</skipArchetypeTests>
 
-        <skipDistributions>true</skipDistributions>
+        <skipDistributions>false</skipDistributions>
```
fsgonz, line 119: why skipDistributions false? This should be default in mule 3 and enabled by a command line property when needed. I don't think we should change the behavior as this could disrupt several pipelines.

pom.xml
```diff
@@ -219,7 +219,7 @@
         <jsonSchemaValidatorVersion>2.2.14</jsonSchemaValidatorVersion>
         <junitVersion>4.13.1</junitVersion>
         <jythonVersion>2.7.2b2</jythonVersion>
-        <jrubyVersion>9.2.14.0</jrubyVersion>
+        <jrubyVersion>9.4.2.0</jrubyVersion>
```
fsgonz, line 222: why is this also needed in the context of a snakeyaml update?

fsgonz (changes requested):

distributions/standalone/assembly-allowlist.txt
mule-pachanady: I have requested them doe updating that beanch

pom.xml
mule-pachanady: Jruby 9.2.14 uses snakesyaml 1.3.x I think and would bring in that transitively, hence upgrading that eventhough Im bit concerned about all the other changes since then in it

pom.xml
mule-pachanady: I was testing locallly, will revert

modules/scripting-jruby/pom.xml
mule-pachanady: We are updating jrubyversion, hence updating , I think this will also update what snakeyaml  we bundle in Mule 3

pom.xml
fsgonz: Verify if the jenkins jobs for this branch are green. The dependency tree says that Jruby has snakeyaml as dependency?

mule-pachanady: @fsgonz This is what I found, mule-module-scripting-jruby has a dependency on jruby-core and jruby-lib. Ran a dependency:tree on jruby lib 9.2.14.0 that we were usibg, it b rings in snakeyaml 1.26  due to
 org.jruby:jruby-stdlib:jar:9.2.14.0
    +- rubygems:psych:gem:3.2.0:provided
      \- org.yaml:snakeyaml:jar:1.26:provided

So we had to switch to newer snakeyaml 2.0 in mule-module-scripting-jruby . interestingly jruby has switched to snakeyaml engine to bypass vulenrability, so they dint inclkude snakeyaml.jar anymore but the engine in jruby-stdlib:jar:9.4.3.0-SNAPSHOT or 9.4.2.0 that I changed

+- rubygems:psych:gem:5.1.0:provided
[INFO] |  \- org.snakeyaml:snakeyaml-engine:jar:2.6:provided

modules/scripting-jruby/pom.xml
mule-pachanady: Moreover snakeyaml 1.31 is vulneranle , we are switching to 2.0

mule-pachanady:
@fsgonz Updated raml-parser-1 Please review again

fsgonz (approved):

fsgonz (dismissed):

mule-pachanady:
@fsgonz We came back to 3.9.6-SNAPSHOT for apikit after discussion with Ale and Juan, as they updated it. need another apoproval

PR #12497, opened 2023-05-30
W-13500699: Split package between spring and extensions-support shaded classes from spring

elrodro83:
* Apply a relocation for spring shaded classes in extensions-support


elrodro83:
--validate

PR #12509, opened 2023-06-01
W-13509911: Policies' Variables are being overriden in parallel scenario (race condition)

mbuchwald:


modules/core-components/src/main/java/org/mule/runtime/core/internal/routing/RoutingUtils.java
```diff
@@ -0,0 +1,28 @@
...
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.internal.routing;
+
+import org.mule.runtime.api.config.FeatureFlaggingService;
+import org.mule.runtime.core.internal.message.EventInternalContext;
+import org.mule.runtime.core.internal.message.InternalEvent;
+import org.mule.runtime.core.internal.policy.SourcePolicyContext;
+
+public final class RoutingUtils {
```
eze210, line 14: Add Javadoc

modules/core-components/src/main/java/org/mule/runtime/core/internal/routing/RoutingUtils.java
```diff
@@ -0,0 +1,28 @@
...
+
+  private RoutingUtils() {
+
+  }
+
+  public static void setSourcePolicyChildContext(InternalEvent event, FeatureFlaggingService featureFlaggingService) {
+    EventInternalContext<?> context = event.getSourcePolicyContext();
+    SourcePolicyContext policyContext = (SourcePolicyContext) context;
```
eze210, line 22: Add an instanceof with a WARN just in case, to avoid a ClassCastException

modules/core-components/src/main/java/org/mule/runtime/core/internal/routing/RoutingUtils.java
```diff
@@ -0,0 +1,28 @@
...
+import org.mule.runtime.core.internal.message.InternalEvent;
+import org.mule.runtime.core.internal.policy.SourcePolicyContext;
+
+public final class RoutingUtils {
+
+  private RoutingUtils() {
+
+  }
+
+  public static void setSourcePolicyChildContext(InternalEvent event, FeatureFlaggingService featureFlaggingService) {
```
eze210, line 20: You forgot to query the feature flag here

eze210 (changes requested):

PR #12510, opened 2023-06-02
W-13497626: Add supported java version concept in SDK

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
```diff
@@ -835,4 +842,76 @@ public static Optional<ImplementingTypeModelProperty> extractImplementingTypePro
...
+            isCommon = false;
+            break;
+          }
+        }
+
+        if (isCommon) {
+          commonVersions.add(version);
+        }
+      }
+    }
```
elrodro83, line 884: why not just use `Stream.distinct()`?

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/DefaultExtensionModelFactoryTestCase.java
```diff
@@ -341,6 +346,17 @@ public void objectStoreParameters() {
     assertThat(typeImport.isPresent(), is(true));
   }
 
+  @Test
```
elrodro83, line 349: allure

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/DefaultExtensionModelFactoryTestCase.java
```diff
@@ -341,6 +346,17 @@ public void objectStoreParameters() {
...
+  public void defaultJavaVersionSupport() {
+    assertThat(heisenbergExtension.getSupportedJavaVersions(), equalTo(DEFAULT_SUPPORTED_JAVA_VERSIONS));
+  }
+
+  @Test
```
elrodro83, line 354: allure

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtilsTestCase.java
```diff
@@ -255,6 +267,63 @@ public void parameterGroupExclusivenessForDslFalseWithMultipleParametersWithAlia
                                      getAliasedParameters());
   }
 
+  @Test
```
elrodro83, line 270: allure

elrodro83 (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/util/MuleExtensionUtils.java
marianogonzalez: because this is not the same. What this is doing is looking the java versions that are common to all extensions. distinct() on its down doesn't fix this

PR #12528, opened 2023-06-06
W-12975498: java.lang.NullPointerException is thrown from org.mule.runtime.module.extension.internal.runtime.connectivity.oauth.ExtensionsOAuthUtils.refreshTokenIfNecessary

marianorean:
(ExtensionsOAuthUtils.java:290)


fsgonz (approved):
Can we create a ticket for an integration or tita test for a connector with a source with oauth that results in a connection exception?

PR #12540, opened 2023-06-07
W-13465559: Delete mule-maven-client properties that are not used as that versions are resolved from the mule-runtime-bom repo

anosenzo:


anosenzo:
--validate

helfgott (approved):

PR #12541, opened 2023-06-07
W-13563214: Fixing issue with composite operation policies redeployment

asanguinetti:
The original reported issue with steps to reproduce can be found here: [W-13520499](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001SkcXRYAZ/view).
The ticket for the root cause issue is this one: [W-13563214](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001TIlliYAD/view)

### DIFF
--- a/modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/AbstractDeploymentTestCase.java
+++ b/modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/AbstractDeploymentTestCase.java
@@ -58,6 +58,7 @@ import static java.lang.System.getProperty;
 import static java.lang.System.setProperty;
 import static java.lang.Thread.currentThread;
 import static java.util.Arrays.asList;
+import static java.util.Arrays.stream;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.emptyMap;
 import static java.util.Collections.emptySet;
@@ -81,7 +82,7 @@ import static org.hamcrest.Matchers.arrayWithSize;
 import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.nullValue;
 import static org.hamcrest.Matchers.sameInstance;
-import static org.hamcrest.collection.IsArrayContainingInAnyOrder.arrayContainingInAnyOrder;
+import static org.hamcrest.collection.ArrayMatching.arrayContainingInAnyOrder;
 import static org.hamcrest.collection.IsCollectionWithSize.hasSize;
 import static org.hamcrest.core.Is.is;
 import static org.hamcrest.core.IsNull.notNullValue;
@@ -106,7 +107,6 @@ import org.mule.runtime.api.artifact.Registry;
 import org.mule.runtime.api.config.custom.CustomizationService;
 import org.mule.runtime.api.deployment.meta.MuleArtifactLoaderDescriptor;
 import org.mule.runtime.api.deployment.meta.MulePolicyModel.MulePolicyModelBuilder;
-import org.mule.runtime.api.deployment.meta.Product;
 import org.mule.runtime.api.exception.MuleException;
 import org.mule.runtime.api.lifecycle.Initialisable;
 import org.mule.runtime.api.lifecycle.InitialisationException;
@@ -130,6 +130,7 @@ import org.mule.runtime.module.artifact.api.classloader.RegionClassLoader;
 import org.mule.runtime.module.artifact.api.descriptor.ApplicationDescriptor;
 import org.mule.runtime.module.artifact.api.descriptor.ArtifactDescriptorValidatorBuilder;
 import org.mule.runtime.module.artifact.api.descriptor.DomainDescriptor;
+import org.mule.runtime.module.artifact.builder.AbstractDependencyFileBuilder;
 import org.mule.runtime.module.artifact.builder.TestArtifactDescriptor;
 import org.mule.runtime.module.artifact.internal.util.ServiceRegistryDescriptorLoaderRepository;
 import org.mule.runtime.module.deployment.api.DeploymentListener;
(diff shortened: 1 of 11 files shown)


modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/AbstractDeploymentTestCase.java
```diff
@@ -310,30 +308,8 @@ public static void afterClass() throws IllegalAccessException {
...
-          .setRequiredProduct(MULE)
-          .withBundleDescriptorLoader(
-                                      createBundleDescriptorLoader(BAR_POLICY_NAME,
-                                                                   MULE_POLICY_CLASSIFIER,
-                                                                   PROPERTIES_BUNDLE_DESCRIPTOR_LOADER_ID))
-          .withClassLoaderModelDescriptorLoader(
-                                                new MuleArtifactLoaderDescriptor(MULE_LOADER_ID, emptyMap()))
-          .build());
+  protected final PolicyFileBuilder barPolicyFileBuilder = createPolicyFileBuilder(BAR_POLICY_NAME);
+  protected final PolicyFileBuilder policyUsingAppPluginFileBuilder = createPolicyFileBuilder(APP_PLUGIN_POLICY_NAME);
```
asanguinetti, line 312: The changes in this file are mostly a refactor because I created the [createPolicyFileBuilder](https://github.com/mulesoft/mule/pull/12541/files#diff-10352cf492c370178de8d34ca443355f732a28ca6eed9dcf8b98764887cc5a0dR1203) method to avoid so much repetition.

Additionally, in this particular highlighted fragment you might spot that the `artifactId` of the `policyUsingAppPluginFileBuilder` was fixed, to actually be named according to the resource file that is going to be used for it [here](https://github.com/mulesoft/mule/blob/8a03b5b3f13427d548ba644a96408939539577c8/modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/ApplicationPolicyDeploymentTestCase.java#L540). Otherwise it made no sense to have a different PolicyFileBuilder for it, because it was exactly equal to `barPolicyFileBuilder`.

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/ApplicationPolicyRedeploymentTestCase.java
```diff
@@ -66,37 +71,68 @@ public void objectStoreSamePartitionWorksAfterRedeployingPolicy() throws Excepti
...
+    policyManager.addPolicy(applicationFileBuilder.getId(), policyToRedeployFileBuilder.getArtifactId(),
+                            policyToRedeployParametrization);
+    policyManager.addPolicy(applicationFileBuilder.getId(), policyToKeepFileBuilder.getArtifactId(), policyToKeepParametrization);
+
+    assertManualExecutionsCount(2);
+
+    policyManager.removePolicy(applicationFileBuilder.getId(), policyToRedeployFileBuilder.getArtifactId());
+
+    // TODO: actually wait until the stale policy is disposed to avoid a source of test flakiness
+    sleep(2000);
```
asanguinetti, line 118: Will try to spend a little bit more time trying to solve this before merging the PR.

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessorTestCase.java
```diff
@@ -56,61 +56,46 @@
...
   public void before() throws MuleException {
-    CoreEvent response = testEvent();
-
     extensionModel = mock(ExtensionModel.class);
     when(extensionModel.getXmlDslModel()).thenReturn(XmlDslModel.builder().setPrefix("mock").build());
 
     componentModel = mock(ComponentModel.class, withSettings().extraInterfaces(EnrichableModel.class));
     when((componentModel).getModelProperty(CompletableComponentExecutorModelProperty.class))
-        .thenReturn(of(new CompletableComponentExecutorModelProperty((cp, p) -> (ctx, callback) -> callback.complete(response))));
+        .thenReturn(of(new CompletableComponentExecutorModelProperty(IdentityExecutor::create)));
```
asanguinetti, line 91: This can no longer respond with a cached `Event` instance (the result of `testEvent()` [is cached](https://github.com/mulesoft/mule/blob/8a03b5b3f13427d548ba644a96408939539577c8/tests/unit/src/main/java/org/mule/tck/junit4/AbstractMuleTestCase.java#L331)).

This is because the response is modified (for example by adding/removing the SDK internal context) and because now some tests will be feeding events concurrently through the `ComponentMessageProcessor`, it will be causing concurrent modifications of the same instance (instance that is not thread-safe).

We will be now using different Event instances as the input for all tests in this TestCase. And for this particular processor, we will be returning the input event untouched. It didn't seem to increase the total duration of this TestCase by any noticeable amount.

asanguinetti (commented):
Leaving some comments for the reviewers

elrodro83 (changes requested):

modules/deployment/src/test/java/org/mule/runtime/module/deployment/internal/ApplicationPolicyRedeploymentTestCase.java
```diff
@@ -66,37 +88,102 @@ public void objectStoreSamePartitionWorksAfterRedeployingPolicy() throws Excepti
...
+    // Control test to make sure there is caching
+    assertThat("Operation policies are not cached, so the test can give false positives",
+               operationPolicy,
+               is(sameInstance(operationPolicy2)));
+
+    return new PhantomReference<>(operationPolicy, new ReferenceQueue<>());
+  }
+
+  private void waitEnqueued(Reference<?> reference) {
+    new PollingProber(5000, 100).check(new JUnitLambdaProbe(() -> {
```
elrodro83, line 172: use the constants form the superclasses

elrodro83 (changes requested):

elrodro83 (approved):

PR #12560, opened 2023-06-14
W-13563214: Fixing issue with composite operation policies redeployment (#12541)

asanguinetti:
- ComponentMessageProcessor's inner flux supplier is now re-created (if needed) on first subscription

(cherry picked from commit 9f475849f3d3cf7b03eb10df9c99a647aeba5432)


asanguinetti:
At the end it was not required for us to cherry-pick this back to 4.3.x.

PR #12575, opened 2023-06-16
Reducing repetitions of Wiremock version and exclusions

asanguinetti:


asanguinetti:
Not needed, we are going to leave the dependencyManagement for test deps on a per-repo basis.

PR #12577, opened 2023-06-20
W-13625620: get file.separator in platform independent way so that ndows build works, this failed tests ParameterAstEETestCase

mule-pachanady:
Fix file.separator so that windows build tests work Eg  ParameterAstEETestCase


mule-pachanady:
Not needed

PR #12588, opened 2023-06-22
W-13544791: Support BODY as a parameter placement in @OAuthParameter

marianogonzalez:


tests/test-extensions/test-oauth-extension/src/main/java/org/mule/test/oauth/TestWithOAuthParamsConnectionProvider.java
```diff
@@ -39,6 +41,10 @@ public class TestWithOAuthParamsConnectionProvider extends LegacyTestOAuthConnec
   @OAuthParameter(placement = HttpParameterPlacement.HEADERS)
   String header;
 
+  @org.mule.sdk.api.annotation.connectivity.oauth.OAuthParameter(placement = BODY)
```
ndinu, line 44: maybe use the mule-extensions-api in one example to get more coverage?

ndinu (approved):

tests/test-extensions/test-oauth-extension/src/main/java/org/mule/test/oauth/TestWithOAuthParamsConnectionProvider.java
marianogonzalez: This feature will not be supported for those using extensions-api direclty, even if we can't enforce it

marianogonzalez:
--validate

PR #12594, opened 2023-06-22
W-13648907: SDK must not process synthetic fields from pojos

elrodro83:


modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/TypeWrapperTestCase.java
```diff
@@ -63,6 +65,16 @@ public void filterInternalJdkFields() {
     assertThat(fields, empty());
   }
 
+  @Test
+  @Issue("W-13648907")
+  @Description("Ensure synthetic fields are filtered out as they might be added to classes we can't use reflection on in Java 17 and above")
+  public void filterSyntheticFields() {
+    TypeWrapper type = new TypeWrapper(SomeClass.class, new DefaultExtensionsTypeLoaderFactory()
```
marianogonzalez, line 72: SomeClass doesn't look like it would have synthetic fields in the first place. I think the test should start by asserting that SomeClass actually does have synthetic fields

marianogonzalez (changes requested):

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/TypeWrapperTestCase.java
elrodro83: it's a non-static inner class, it has the `this$0` field. will add it to the test

mbuchwald (dismissed):

elrodro83:
--validate

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/loader/java/type/runtime/TypeWrapperTestCase.java
```diff
@@ -63,6 +70,25 @@ public void filterInternalJdkFields() {
...
+  @Issue("W-13648907")
+  @Description("Ensure synthetic fields are filtered out as they might be added to classes we can't use reflection on in Java 17 and above")
+  public void filterSyntheticFields() {
+    // precondition check
+    final Collection<String> allFiledNames = IntrospectionUtils.getFields(SomeClass.class)
+        .stream()
+        .map(Field::getName)
+        .collect(toSet());
+    // being a non-static inner class makes it have a synthetic field pointing to its owner instance.
+    assertThat(allFiledNames, hasItem("this$0"));
```
marianogonzalez, line 83: you should check by the same condition that the productive code is checking. Also this assertion would break is future java versions change this impl.

marianogonzalez (changes requested):

marianogonzalez (approved):

PR #12600, opened 2023-06-26
W-13586373: Duplicating hack to make cached spring references WEAK instead of SOFT

eze210:
For reference, see https://github.com/spring-projects/spring-framework/issues/25028


fsgonz (approved):

eze210:
--validate

PR #12602, opened 2023-06-27
W-13594025: Performance investigation (HTTP Proxy drop)

sofiamorseletto:


sofiamorseletto:
--validate

sofiamorseletto:
--validate

sofiamorseletto:
--validate

sofiamorseletto:
--validate

modules/tracing/mule-tracer-customization-impl/src/main/java/org/mule/runtime/tracer/customization/impl/export/TracingLevelExportInfo.java
```diff
@@ -36,11 +37,17 @@ public boolean isOverride() {
   }
 
   public boolean isExportable() {
-    return getInitialExportInfo().isExportable();
+    if (initialExportInfo == null) {
+      initialExportInfo = getInitialExportInfo();
```
alopezbizon, line 41: I think we also have to set initialExportInfo when propagateExportInfo() is called because due to its parent's level propagation, for example, the initialExportInfo might have changed to be exportable or vice versa.

alopezbizon (approved):

sofiamorseletto:
--validate

PR #12611, opened 2023-06-30
W-13689895: Tracer configuration not reading boolean value for the "m…

IvanAndresFritzler:
…ule.openTelemetry.tracer.exporter.enabled" configuration value.


modules/tracing/mule-tracer-exporter-configuration-impl/src/test/java/org/mule/runtime/tracer/exporter/config/impl/FileSpanExporterConfigurationTestCase.java
```diff
@@ -143,6 +146,26 @@ public void whenFileIsNotFoundNoPropertyIsFound() {
     assertThat(testNoFileFoundSpanExporterConfiguration.getStringValue(KEY_PROPERTY_NON_SYSTEM_PROPERTY), is(nullValue()));
   }
 
+  @Test
+  public void readsStringValue() {
```
sofiamorseletto, line 150: Could we add a similar check to FileTracingLevelConfigurationTestCase too?

IvanAndresFritzler: We cannot add that coverage because the interface is different and it only exposes tracing levels (see TracingLevelConfiguration). Since all the possible levels are strings, we cannot try to read a boolean as a string.

fsgonz (approved):

sofiamorseletto (approved):

PR #12613, opened 2023-07-03
W-13682699: Parse Template Component doesn't support absolute paths

eze210:
The problem was that the validation method was calling `ClassLoader#getResourceAsStrem`, and then only the resources accessible by the classloader were able to be loaded.
Now we use the utility method [`IOUtils#getResourceAsUrl`](https://github.com/mulesoft/mule-api/blob/3858279cda736349dcc688c7be506c77ecf10cc6/src/main/java/org/mule/runtime/api/util/IOUtils.java#L54), since it also tries to use the File api to search the resource.


fsgonz (approved):

sofiamorseletto (approved):

eze210:
--validate

eze210:
--validate

eze210:
--validate

PR #12614, opened 2023-07-03
W-13709325: Separate containerClassLoader mule and 3rd parties

elrodro83:
* move test specific code to test modules
* deprecation of `CompositeClassLoader` in favor of ByteBuddy's `MultiParentClassLoader`
* additioal minor code cleanups

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/util/CompositeClassLoader.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/util/CompositeClassLoader.java
@@ -6,16 +6,16 @@
  */
 package org.mule.runtime.core.internal.util;
 
-import static com.github.benmanes.caffeine.cache.Caffeine.newBuilder;
 import static java.lang.System.identityHashCode;
 import static java.util.Arrays.asList;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
 import static java.util.Collections.unmodifiableList;
 import static java.util.concurrent.TimeUnit.MINUTES;
-import org.mule.runtime.core.api.util.CompoundEnumeration;
 
-import com.github.benmanes.caffeine.cache.Cache;
+import static com.github.benmanes.caffeine.cache.Caffeine.newBuilder;
+
+import org.mule.runtime.core.api.util.CompoundEnumeration;
 
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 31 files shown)


elrodro83:
--validate

core/src/main/java/org/mule/runtime/core/internal/util/CompositeClassLoader.java
```diff
@@ -24,14 +24,19 @@
...
+
 /**
  * Classloader implementation that, given a set of classloaders, will first search for a resource/class in the first one. If it is
  * not found, it will try the next and so on until the resource/class is found or all classloaders have been tried.
  * <p>
  * For {@link #getResources(String)}, all the classloaders will be queried to get the union of all found resources.
  *
  * @since 1.0
+ * 
+ * @deprecated Use {@link MultiParentClassLoaderUtils} instead.
```
eze210, line 37: Regarding the "*Utils": Is it intentional because you're recommending to use the utility methods for creation or it's a typo and you meant MultiParentClassLoader?

core/src/main/java/org/mule/runtime/core/privileged/component/AnnotatedObjectInvocationHandler.java
```diff
@@ -101,13 +102,7 @@ && asList(clazz.getMethods()).stream().anyMatch(m -> "getAnnotations".equals(m.g
             .intercept(to(annotatedObjectInvocationHandler))));
     builder.set(builder.get().method(isToString().and(isDeclaredBy(Object.class))).intercept(to(new ToStringInterceptor())));
 
-    ClassLoader classLoader;
-    if (ByteBuddy.class.getClassLoader() != clazz.getClassLoader()) {
```
eze210, line 105: We're removing this conditional, is it intentional?

eze210 (changes requested):

core/src/main/java/org/mule/runtime/core/internal/util/CompositeClassLoader.java
elrodro83: it is intentional, since the Utils handlings the caching as the `from` method from `CompositeClassLoader` does

core/src/main/java/org/mule/runtime/core/privileged/component/AnnotatedObjectInvocationHandler.java
elrodro83: yes, that condition is already handled by BB `MultiParentClassLoader`'s builder

eze210 (approved):

PR #12620, opened 2023-07-07
W-12082402: Startup script gives higher priority to java in PATH instead

elrodro83:
* Fix how system tests determine the JAVA_HOME when configuring tests to use a different jvm

### DIFF
--- a/tests/infrastructure/pom.xml
+++ b/tests/infrastructure/pom.xml
@@ -58,6 +58,7 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-exec</artifactId>
+            <version>1.3</version>
         </dependency>
         <dependency>
             <groupId>commons-net</groupId>
--- a/tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/AbstractOSController.java
+++ b/tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/AbstractOSController.java
@@ -21,6 +21,12 @@ import static org.slf4j.LoggerFactory.getLogger;
 import org.mule.tck.probe.PollingProber;
 import org.mule.tck.probe.Probe;
 
+import java.io.BufferedReader;
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.FileReader;
+import java.io.IOException;
+import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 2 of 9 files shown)


tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/AbstractOSController.java
```diff
@@ -119,7 +126,7 @@ public boolean isSatisfied() {
       @Override
       public String describeFailure() {
         runSync(DUMP_CMD);
-        return format("The mule instance didn't start on time: %s", status);
+        throw new MuleControllerException(format("The mule instance didn't start on time: %s", status));
```
eze210, line 129: Why would we do this?

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/Controller.java
```diff
@@ -76,7 +85,35 @@ public String getMuleBin() {
...
+      final StringBuilder muleLogBuilder = new StringBuilder();
+      muleLogBuilder.append(muleLogFile.getName() + ":" + lineSeparator());
+      muleLogBuilder.append("============" + lineSeparator());
+      try (BufferedReader br = new BufferedReader(new FileReader(muleLogFile))) {
+        String line;
+        while ((line = br.readLine()) != null) {
+          muleLogBuilder.append(" > " + line + lineSeparator());
+        }
+      }
+      LOGGER.info(muleLogBuilder.toString());
```
eze210, line 113: To avoid the toString when info is disabled:
```suggestion
      LOGGER.info("{}", muleLogBuilder);
```

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/Controller.java
```diff
@@ -76,7 +85,35 @@ public String getMuleBin() {
...
+    } else {
+      LOGGER.warn("Log (" + muleLogFile.toString() + ") not available.");
```
eze210, line 115: Same as previous comment:
```suggestion
      LOGGER.warn("Log ({}) not available.", muleLogFile);
```

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/MuleProcessController.java
```diff
@@ -7,17 +7,25 @@
...
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 public class MuleProcessController {
 
+  private static final Logger LOGGER = LoggerFactory.getLogger(MuleProcessController.class);
```
eze210, line 26: Unused

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/MuleProcessController.java
```diff
@@ -7,17 +7,25 @@
...
+  private static final Logger LOGGER_WRAPPER = LoggerFactory.getLogger(MuleProcessController.class.getName() + ".wrapper");
```
eze210, line 27: static import

tests/infrastructure/pom.xml
```diff
@@ -58,6 +58,7 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-exec</artifactId>
+            <version>1.3</version>
```
eze210, line 61: Would be better to update it in the bom

eze210 (changes requested):

tests/infrastructure/pom.xml
elrodro83: i will actually remove it from the bom, since this is used only in these tests

tests/infrastructure/src/main/java/org/mule/test/infrastructure/process/AbstractOSController.java
elrodro83: i needed to force an exception to catch in and print the log. Refactored this

eze210 (approved):

PR #12634, opened 2023-07-13
W-13753869: Validate deployed extensions support running Java version

marianogonzalez:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/DefaultExtensionManager.java
```diff
@@ -134,6 +148,8 @@ public void registerExtension(ExtensionModel extensionModel) {
       LOGGER.debug("Registering extension {} (version: {} vendor: {} )", extensionName, extensionVersion, extensionVendor);
     }
 
+    extensionJdkValidator.validateJdkSupport(extensionModel);
```
elrodro83, line 151: instead of failing on the first extension found, we should log a single failure with all unsupported extensions

elrodro83 (approved):

elrodro83 (changes requested):
See comments

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/manager/DefaultExtensionManager.java
marianogonzalez: --validate

facunlk (approved):

PR #12644, opened 2023-07-24
W-12412016: [NoTanuki] Move logging initialization to MuleContainerBootstrap

eze210:


modules/log4j-boot-configurator/src/main/java/module-info.java
```diff
@@ -0,0 +1,19 @@
...
+ */
+/**
+ * Mule Log4j Configurator Module.
+ * 
+ * @moduleGraph
+ * @since 4.5
+ */
+module org.mule.runtime.boot.log4j {
+  // Exported. Add "to"?
+  exports org.mule.runtime.module.log4j.boot.api;
```
asanguinetti, line 12: I think if we are going to call the package `*.api.*` then the convention is to export it to everybody. So I think we can remove the comment.

modules/reboot/src/main/java/org/mule/runtime/module/reboot/internal/MuleLog4jConfigurer.java
```diff
@@ -0,0 +1,38 @@
...
+  @Override
+  public boolean configure() throws BootstrapConfigurationException {
+    if (getProperty(MULE_SIMPLE_LOG) == null) {
+      // We need to force the creation of a logger before we can change the manager factory.
+      // This is because if not, any logger that will be acquired by MuleLog4jContextFactory code
+      // will fail since it will try to use a null factory.
+      getLogger("triggerDefaultFactoryCreation");
+      // We need to set this property so log4j uses the same context factory everywhere
+      setProperty("log4j2.loggerContextFactory", MuleLog4jContextFactory.class.getName());
+      MuleLog4jContextFactory log4jContextFactory = new MuleLog4jContextFactory();
```
asanguinetti, line 32: Is it intentional that we are not going through the `MuleLog4jConfiguratorUtils`?

modules/log4j-configurator/src/main/java/org/mule/runtime/module/log4j/internal/MuleLog4jConfiguratorUtils.java
```diff
@@ -0,0 +1,44 @@
...
+    configureSelector(contextFactory, getProperty(MULE_LOG_SEPARATION_DISABLED) == null);
+  }
+
+  public static void configureSelector(MuleLog4jContextFactory contextFactory, boolean logSeparationEnabled) {
+    if (logSeparationEnabled) {
+      contextFactory.setContextSelector(new ArtifactAwareContextSelector(), MuleLog4jConfiguratorUtils::disposeIfDisposable);
+    } else {
+      contextFactory.setContextSelector(new SimpleContextSelector(), MuleLog4jConfiguratorUtils::disposeIfDisposable);
+    }
+  }
```
asanguinetti, line 37: Is there any reason why these are not methods of `MuleLog4jContextFactory`?
It looks like they are all creating or working on `MuleLog4jContextFactory` instances.
Wouldn't they be easier to find if they belong to the same class?

modules/log4j-configurator/src/main/java/module-info.java
```diff
@@ -13,12 +13,15 @@
...
   requires org.mule.runtime.artifact.activation;
+  requires org.mule.runtime.boot.log4j;
   requires org.mule.runtime.core;
   requires org.mule.runtime.deployment.model;
 
   // Third party modules
+  requires com.github.benmanes.caffeine;
   requires com.lmax.disruptor;
   requires it.unimi.dsi.fastutil;
+  requires org.apache.commons.lang3;
```
asanguinetti, line 24: I'm not seeing why these new `requires` have been added, I'm guessing they were missing already. Just mentioning it in case you added them by mistake.

modules/log4j-boot-configurator/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+    <packaging>jar</packaging>
+    <name>Mule Log4j Boot Configurator Module</name>
+    <description>Mule Log4j Boot Configurator Module</description>
+
+    <properties>
+        <!-- Add visibility required by the tests exclusively -->
+        <surefire.args>
+            -XX:+IgnoreUnrecognizedVMOptions
+            --add-modules=net.bytebuddy.agent
+        </surefire.args>
```
asanguinetti, line 21: Which tests? I'm not seeing any in this module.

modules/log4j-boot-configurator/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+        <formatterConfigPath>../../formatter.xml</formatterConfigPath>
+    </properties>
+
+    <dependencies>
+        <!-- Mule dependencies -->
+        <dependency>
+            <groupId>org.mule.runtime</groupId>
+            <artifactId>mule-module-logging</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
asanguinetti, line 34: This sounds fishy, wouldn't this create a dependency from a lib/boot module to a lib/mule module?

modules/launcher/src/main/java/org/mule/runtime/module/launcher/DefaultMuleContainer.java
```diff
@@ -103,14 +104,10 @@ public class DefaultMuleContainer implements MuleContainer {
...
-      // will fail since it will try to use a null factory.
-      LoggerFactory.getLogger("triggerDefaultFactoryCreation");
-      // We need to set this property so log4j uses the same context factory everywhere
-      setProperty("log4j2.loggerContextFactory", MuleLog4jContextFactory.class.getName());
-      log4jContextFactory = new MuleLog4jContextFactory();
-      LogManager.setFactory(log4jContextFactory);
+      log4jContextFactory = (MuleLog4jContextFactory) LogManager.getFactory();
+      System.out.println("Setting selector");
+      configureSelector(log4jContextFactory);
+      System.out.println("Selector was set");
```
asanguinetti, line 110: Do we want to keep these prints?

asanguinetti (changes requested):

modules/reboot/src/main/java/org/mule/runtime/module/reboot/internal/MuleLog4jConfigurer.java
eze210: That utility class is not available in lib/boot. I added javadoc in that class clarifying that, and I removed some methods only used from tests

modules/log4j-configurator/src/main/java/org/mule/runtime/module/log4j/internal/MuleLog4jConfiguratorUtils.java
eze210: Yes... Added javadoc clarifying

modules/log4j-boot-configurator/src/main/java/module-info.java
eze210: I left only the `// Exported` just as a "section"

modules/launcher/src/main/java/org/mule/runtime/module/launcher/DefaultMuleContainer.java
eze210: Removed

modules/log4j-boot-configurator/pom.xml
eze210: mule-module-logging is in lib/boot

modules/log4j-configurator/src/main/java/module-info.java
eze210: Removed

modules/log4j-boot-configurator/pom.xml
eze210: Sorry, bad copy-paste

modules/log4j-boot-configurator/pom.xml
asanguinetti: My bad

asanguinetti (dismissed):

asanguinetti (approved):

PR #12650, opened 2023-07-25
W-13742405: Create a module layer instead of a classloader for the container - Prework

elrodro83:
* Move `MethodInvoker` to service module
* Make stack trace checker aware of modules
* Generate dynamic classes in the module where it has the required visibility to required modules
* Avoid split packages between deployment-model-impl and its test-jar


elrodro83:
--validate

elrodro83:
--validate

eze210 (approved):

PR #12674, opened 2023-08-04
W-13762853: Some interfaces should return a Span instead of an Intern…

IvanAndresFritzler:
…alSpan.


fsgonz (approved):

IvanAndresFritzler:
--validate

PR #12698, opened 2023-08-11
W-12421187: [StartupPerf] ArtifactAst validations: Avoid generating a…

fsgonz:
… dependency graph for every validation that needs it


runtime-extension-model/src/test/java/org/mule/runtime/config/internal/validation/CachingArtifactAstGraphDependencyProviderTestCase.java
```diff
@@ -0,0 +1,37 @@
...
+import org.mule.runtime.ast.api.ArtifactAst;
+import org.mule.runtime.config.internal.validation.ast.CachingArtifactAstGraphDependencyProvider;
+
+import io.qameta.allure.Feature;
+import io.qameta.allure.Story;
+import org.junit.Assert;
+
+@Feature(MULE_DSL)
+@Story(DSL_VALIDATION_STORY)
+public class CachingArtifactAstGraphDependencyProviderTestCase {
```
elrodro83, line 26: `@Issue`

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/CoreValidationsProvider.java
```diff
@@ -34,6 +36,8 @@ public class CoreValidationsProvider implements ValidationsProvider {
 
   private boolean ignoreParamsWithProperties;
 
+  private ArtifactAstDependencyGraphProvider artifactAstDependencyGraphProvider = new CachingArtifactAstGraphDependencyProvider();
```
elrodro83, line 39: this would still be creating the dependencyGraph for each validation provider requiring it.

elrodro83 (changes requested):

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/ast/CachingArtifactAstDependencyGraphProvider.java
```diff
@@ -0,0 +1,32 @@
...
+import org.mule.runtime.ast.graph.api.ArtifactAstDependencyGraphFactory;
+import org.mule.runtime.ast.graph.api.ArtifactAstDependencyGraphProvider;
+
+import com.github.benmanes.caffeine.cache.Cache;
+
+
+/**
+ * An {@link ArtifactAstDependencyGraphProvider} that caches the {@link ArtifactAstDependencyGraphProvider}.
+ *
+ * TODO W-13931931: verify if it makes sense to have a bean in the registry to be used apart from the use case for AST validations
```
elrodro83, line 19: registry won't be good for this case since it will cause the depGraph and the AST to be held in memory for as long as the app is tunning

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/ast/CachingArtifactAstDependencyGraphProvider.java
```diff
@@ -0,0 +1,32 @@
...
+ * in W-12421187
+ *
+ * @since 4.6.0
+ */
+public class CachingArtifactAstDependencyGraphProvider implements ArtifactAstDependencyGraphProvider {
+
+  private Cache<ArtifactAst, ArtifactAstDependencyGraph> dependencyGraphCache = newBuilder().build();
```
elrodro83, line 26: a deployment only has one AST, so keeping a map here is not needed

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/CoreValidationsProvider.java
```diff
@@ -28,12 +31,14 @@
...
 import javax.inject.Named;
 
-public class CoreValidationsProvider implements ValidationsProvider {
+public class CoreValidationsProvider implements ValidationsProvider, ArtifactAstGraphDependencyProviderAware {
 
   private ClassLoader artifactRegionClassLoader;
 
   private boolean ignoreParamsWithProperties;
 
+  private ArtifactAstDependencyGraphProvider artifactAstDependencyGraphProvider = new CachingArtifactAstDependencyGraphProvider();
```
elrodro83, line 40: Let's discuss about declaring this as `@Inject private Supplier<ArtifactAstDependencyGraph>`

elrodro83 (changes requested):

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/ast/CachingArtifactAstDependencyGraphProvider.java
fsgonz: It also applies to the other cases where we talk where it is already been held.
I've changed the TODO in the javadoc.

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/ast/CachingArtifactAstDependencyGraphProvider.java
fsgonz: A deployment has one AST but this is an implementation of a validator. There is nothing in the contract that even is related to the "deployment". I've added a task to create a context associated to the validations.

runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/CoreValidationsProvider.java
fsgonz: Already discussed added comment and what is needed to not fail before the creation of the validation context.

elrodro83 (approved):

fsgonz:
--validate

fsgonz:
--validate

fsgonz:
--validate

PR #12701, opened 2023-08-14
W-13945672: Services: provide support to include libraries to be used with a minimal java version

elrodro83:


elrodro83:
--validate

modules/service/src/main/java/org/mule/runtime/module/service/internal/artifact/LibFolderClassLoaderConfigurationLoader.java
```diff
@@ -81,12 +84,27 @@ private List<URL> getServiceUrls(File rootFolder) {
...
+  private List<URL> getServiceUrlsForJavaVersion(File rootFolder) {
+    List<URL> urls = new LinkedList<>();
+
+    final File libJavaVersionsDir = new File(rootFolder, LIB_DIR + "/java-versions");
+    if (libJavaVersionsDir.exists()) {
+      for (File file : libJavaVersionsDir.listFiles()) {
+        if (JVM_SPECIFICATION_VERSION >= parseInt(file.getName())) {
+          loadJarsFromFolder(urls, file);
+        }
+      }
```
rbourbonnavarro, line 96: We should consider the issue that every version of a lib that is present for older JVMs than the one we are running with will be loaded. Suppose you are running with Java 17, and there are different versions of a lib for Java 11 and Java 17, both of them would be loaded.

elrodro83:
--validate

PR #12711, opened 2023-08-17
W-13916544: Upgrade Xstream

marianorean:


modules/xml/src/main/java/org/mule/module/xml/transformer/wire/XStreamWireFormat.java
```diff
@@ -20,7 +20,7 @@
  */
 public class XStreamWireFormat extends TransformerPairWireFormat
 {
-    
+
```
fsgonz, line 23: extra space.

fsgonz (changes requested):

fsgonz (dismissed):
Approved. Let's see if all the tests are passing.

fsgonz (approved):

PR #12642, opened 2023-07-21
W-12276645: Performance improvements on DefaultExtensionsClient

asanguinetti:

PR #12716, opened 2023-08-22
W-12276645: Performance improvements on DefaultExtensionsClient

asanguinetti:
### Context

This is a continuation of #12642.

In this version we have focused on moving some of the `ResolverSet` creation/initialization out from the execution phase and to the `OperationClient` initialization phase.
In order to do that we separated the logic that depends on the actual values of the parameters from the logic that depends only on the model.
We added a level of indirection that will take care of creating the specific ValueResolvers once the actual values are available (during the execution). But we don't need to recreate all the wrapping ValueResolver structure.
There are couple of `TODOs` with things to check, mostly regarding implementation details and naming of things.

From the other proposed improvements in the previous candidate PR we removed everything except for the part of fixing the caching of the configuration provider by using the right OperationKey instance.

The proposal involving caching the Null Event made no sense in the end, because in reality, in a delegation scenario such as the WSC, the OperationClient will be using the Event that is going through the operation that is using the ExtensionsClient. Other scenarios such as MuleFramework are outside of the scope of this task. They can continue working as they are right now, or we can introduce caching of the event later on.

The part about skipping the resolution of `target` and `targetValue` parameters has been left out because the performance gains were not big enough to justify the added complexity in the code. We might still want to come back to that if we come up with a simpler implementation.

### Results so far
```
4.5.0-20230812.011250-3427
Requests      [total, rate, throughput]         227653, 1896.82, 1896.08
Duration      [total, attack, wait]             2m0s, 2m0s, 46.922ms
Latencies     [min, mean, 50, 90, 95, 99, max]  1.248ms, 104.715ms, 98.616ms, 172.124ms, 198.278ms, 255.563ms, 537.088ms
Bytes In      [total, mean]                     37335092, 164.00
Bytes Out     [total, mean]                     14569792, 64.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:227653
Error Set:

4.5.0-20230812.011250-3427-poc/EC-perf-improvements-b
Requests      [total, rate, throughput]         268471, 2237.19, 2236.47
Duration      [total, attack, wait]             2m0s, 2m0s, 38.544ms
Latencies     [min, mean, 50, 90, 95, 99, max]  1.143ms, 88.73ms, 82.818ms, 148.303ms, 171.123ms, 223.429ms, 435.307ms
Bytes In      [total, mean]                     44029244, 164.00
Bytes Out     [total, mean]                     17182144, 64.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:268471
Error Set:

4.4.0-20221212
Requests      [total, rate, throughput]         284638, 2371.98, 2370.97
Duration      [total, attack, wait]             2m0s, 2m0s, 51.027ms
Latencies     [min, mean, 50, 90, 95, 99, max]  1.158ms, 83.745ms, 78.681ms, 141.657ms, 162.149ms, 208.498ms, 1.055s
Bytes In      [total, mean]                     46680632, 164.00
Bytes Out     [total, mean]                     18216832, 64.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:284638
Error Set:
```


marianogonzalez (changes requested):

asanguinetti:
Superseded by #12749, #12752, #12754, #12761.

PR #12717, opened 2023-08-22
W-12276645: Performance improvements on DefaultExtensionsClient - With added filtering of targetValue

asanguinetti:
Same as #12716 but skipping resolution of `targetValue`.


asanguinetti:
Superseded by #12749, #12752, #12754, #12761.

PR #12732, opened 2023-08-28
W-12276645: Reverting back to ComponentMessageProcessor for legacy ExtensionsClient API.

asanguinetti:
PoC for temporarily reverting back to the previous implementation of the ExtensionsClient API using cached ComponentMessageProcessor instances.

The new implementation is much better in terms of memory footprint but is less performant in terms of throughput. While we work on fixing the root cause for the performance issue, we can revert back to the previous impl.


asanguinetti:
Discarded the revert and went for a full fix.

PR #12741, opened 2023-08-29
W-12276645: Reverting back to the legacy parameter resolution impl for legacy ExtensionsClient API.

asanguinetti:


asanguinetti:
Discarded the revert and went for a full fix.

PR #12759, opened 2023-09-01
W-14067425: Avoid excesive creations of test flows in ComponentMessageProcessorTestCase.

asanguinetti:
- Also making sure InfiniteEmitters are stopped even when the test fails.

### DIFF
--- a/modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessorTestCase.java
+++ b/modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessorTestCase.java
@@ -7,6 +7,7 @@
 package org.mule.runtime.module.extension.internal.runtime.operation;
 
 import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
+import static org.mule.runtime.core.api.event.EventContextFactory.create;
 import static org.mule.runtime.core.api.lifecycle.LifecycleUtils.disposeIfNeeded;
 import static org.mule.runtime.core.api.lifecycle.LifecycleUtils.initialiseIfNeeded;
 import static org.mule.runtime.core.api.lifecycle.LifecycleUtils.startIfNeeded;
@@ -39,10 +40,12 @@ import org.mule.runtime.api.meta.model.ComponentModel;
 import org.mule.runtime.api.meta.model.EnrichableModel;
 import org.mule.runtime.api.meta.model.ExtensionModel;
 import org.mule.runtime.api.meta.model.XmlDslModel;
+import org.mule.runtime.core.api.construct.Flow;
 import org.mule.runtime.core.api.event.CoreEvent;
 import org.mule.runtime.core.api.expression.ExpressionRuntimeException;
 import org.mule.runtime.core.api.extension.ExtensionManager;
 import org.mule.runtime.core.internal.exception.MessagingException;
+import org.mule.runtime.core.internal.message.InternalEvent;
 import org.mule.runtime.core.internal.policy.PolicyManager;
 import org.mule.runtime.extension.api.runtime.config.ConfigurationProvider;
 import org.mule.runtime.metadata.api.cache.MetadataCacheIdGeneratorFactory;
@@ -87,6 +90,10 @@ public class ComponentMessageProcessorTestCase extends AbstractMuleContextTestCa
   protected ExtensionManager extensionManager;
   protected PolicyManager mockPolicyManager;
 
+  // A cached flow for creating test events. It doesn't even have to contain the processor we are going to test, because we will
+  // be sending the events directly through the processor, without using the flow.
+  private Flow testFlow;
+
   @Before
   public void before() throws MuleException {
     extensionModel = mock(ExtensionModel.class);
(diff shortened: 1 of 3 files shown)


modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessorTestCase.java
```diff
@@ -227,11 +246,16 @@ public void multipleUpstreamPublishers() throws MuleException, InterruptedExcept
...
 
     eventsEmitter.start();
-    eventsEmitter2.start();
-    eventsConsumer.await();
-    eventsEmitter.stop();
-    eventsConsumer2.await();
-    eventsEmitter2.stop();
+    try {
+      eventsEmitter2.start();
+      eventsConsumer.await();
```
elrodro83, line 251: in the spirit of reducing flakiness, let's also pass a timeout to `await`

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessorTestCase.java
```diff
@@ -227,11 +246,16 @@ public void multipleUpstreamPublishers() throws MuleException, InterruptedExcept
...
+      eventsEmitter.stop();
+      eventsConsumer2.await();
+      eventsEmitter2.stop();
+    } catch (Throwable t) {
+      eventsEmitter.stop();
+      eventsEmitter2.stop();
```
elrodro83, line 257: could we just do these `stop`s in a `finally` block?

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessorTestCase.java
```diff
@@ -138,6 +151,12 @@ public ProcessingType getInnerProcessingType() {
     };
   }
 
+  @Override
+  protected CoreEvent.Builder getEventBuilder() {
```
elrodro83, line 155: instead of overriding here, could you make this change to apply to all calls to `MuleTestUtils#getTestFlow()`? That would benefit the whole tests execution i think.

elrodro83 (changes requested):

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessorTestCase.java
asanguinetti: Don't we have the overall test timeout for that?
or maybe you are thinking of a shorter timeout for this?

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessorTestCase.java
asanguinetti: Yes, I think we could. I was trying to make it messy by completing the first publisher before awaiting for the second subscriber to have received all items. But in reality it makes no sense, because it could have received all the events already anyway, because I already did both starts.
So yes, let's simplify it.

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessorTestCase.java
asanguinetti: I guess we could. But I would prefer to make that into a separate PR. Because it may have hidden side effects that we are not considering. For example, I'm seeing that at some point we [added something to keep track of test flows](https://github.com/mulesoft/mule/blob/master/tests/unit/src/main/java/org/mule/tck/junit4/AbstractMuleTestCase.java#L308), which implies that it is not expected for `MuleTestUtils#getTestFlow` to return always the same cached instance.

On the other hand, we do have things like [this one](https://github.com/mulesoft/mule/blob/master/tests/unit/src/main/java/org/mule/tck/util/MuleContextUtils.java#L369), where we create a new test flow and we register it in the registry. But since the test flow always has the same name, calling that method multiple times will override the previous ones in the registry. Only the last one will undergo the lifecycle transitions of the registry.

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessorTestCase.java
elrodro83: shorter for this one

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessorTestCase.java
elrodro83: mm, that would mean that only the last one to be registered will be stopped/disposed?

asanguinetti: Seems like it, though I haven't actually verified it. However, we added [this](https://github.com/mulesoft/mule/commit/1109cc25db6a7bebe3534bdbcd582f73a0aa389c) some time ago, I'm guessing it is related. And it will probably still be needed regardless, for tests that do not use any MuleContext.

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessorTestCase.java
asanguinetti: using `RECEIVE_TIMEOUT`

modules/extensions-support/src/test/java/org/mule/runtime/module/extension/internal/runtime/operation/ComponentMessageProcessorTestCase.java
asanguinetti: Done

elrodro83 (approved):

PR #12793, opened 2023-09-12
W-14077882: Handle concurrency with Maven provided features

rbourbonnavarro:


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/artifact/MavenClassLoaderConfigurationLoader.java
```diff
@@ -68,10 +75,11 @@ private void refresh() {
   }
 
   private void createClassLoaderConfigurationLoaders() {
-    Optional<MavenClient> mavenClient = ofNullable(createMavenClient(mavenRuntimeConfig));
+    closeMavenClient();
```
elrodro83, line 78: is this something that may happen?

elrodro83 (changes requested):

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/artifact/MavenClassLoaderConfigurationLoader.java
rbourbonnavarro: It could happen if the maven configuration changed between invocations to the `load` method.

elrodro83 (approved):

PR #12794, opened 2023-09-12
W-13681772: XML SDK fails to replace a complex parameter with a prope…

alopezbizon:
…rty value

### DIFF
--- a/modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
+++ b/modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
@@ -147,8 +147,8 @@ class MacroExpandedComponentAst extends BaseComponentAstDecorator {
       private Object copyComplexComponentTreeRecursively() {
         ComponentAst component = (ComponentAst) getDecorated().getValue().getRight();
         MacroExpandedComponentAst macroExpandedComponentAst =
-            new MacroExpandedComponentAst(component, location, moduleGlobalElementsNames, defaultGlobalElementSuffix,
-                                          literalsParameters, macroExpandedChildren);
+            new MacroExpandedComponentAst(component, component.getLocation(), moduleGlobalElementsNames,
+                                          defaultGlobalElementSuffix, literalsParameters, macroExpandedChildren);
         return copyComponentTreeRecursively(macroExpandedComponentAst, identity());
       }
 
--- a/modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
+++ b/modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
@@ -115,17 +115,17 @@ class MacroExpandedComponentAst extends BaseComponentAstDecorator {
       }
 
       private <T> Object mapComponent(T rawValue) {
-        if (isAnExpression(rawValue)) {
+        if (mustMacroExpandRawValue(rawValue)) {
           return macroExpandedRawValue((String) rawValue);
         } else if (isAnErrorMappings()) {
           return mapErrorMappings(getDecorated());
-        } else if (supportsChildDeclaration() && !isAKnownType(rawValue)) {
+        } else if (supportsChildDeclaration() && !isResolvedByMetadataVisitor(rawValue)) {
           return copyMacroExpandedComponentTreeRecursively();
         }
         return rawValue;
       }
 
-      private <T> boolean isAnExpression(T rawValue) {
+      private <T> boolean mustMacroExpandRawValue(T rawValue) {
         // Evaluates if the parameter is not declared as a child and can be an expression so that it is resolved.
         // In that case, the raw value is always a String.
         return rawValue instanceof String;
@@ -135,7 +135,7 @@ class MacroExpandedComponentAst extends BaseComponentAstDecorator {
         return getDecorated().getModel().getName().equals(ERROR_MAPPINGS_PARAMETER_NAME);
       }
 
-      private <T> boolean isAKnownType(T rawValue) {
+      private <T> boolean isResolvedByMetadataVisitor(T rawValue) {
         // These cases are resolved by a MetadataTypeVisitor in MuleAstUtils.
         // See the visitArrayType and visitObject methods.
         return rawValue instanceof List || rawValue instanceof Map;
--- a/modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
+++ b/modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
@@ -9,8 +9,6 @@ package org.mule.runtime.extension.internal.ast;
 import static org.mule.runtime.api.functional.Either.right;
 import static org.mule.runtime.ast.api.util.MuleArtifactAstCopyUtils.copyComponentTreeRecursively;
 import static org.mule.runtime.extension.api.ExtensionConstants.ERROR_MAPPINGS_PARAMETER_NAME;
-import static org.mule.runtime.extension.api.ExtensionConstants.RECONNECTION_CONFIG_PARAMETER_NAME;
-import static org.mule.runtime.extension.api.ExtensionConstants.RECONNECTION_STRATEGY_PARAMETER_NAME;
 
 import static java.util.Collections.emptyMap;
 import static java.util.Objects.requireNonNull;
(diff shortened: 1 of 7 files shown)


modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
```diff
@@ -128,6 +136,19 @@ private Object mapErrorMappings(final ComponentParameterAst originalParameter) {
...
+        MacroExpandedComponentAst macroExpandedComponentAst = (MacroExpandedComponentAst) originalParameter.getValue().getRight();
+        return MuleArtifactAstCopyUtils.copyComponentTreeRecursively(macroExpandedComponentAst, UnaryOperator.identity());
+      }
+
+      private Object mapReconnectionStrategy(final ComponentParameterAst originalParameter) {
+        MacroExpandedComponentAst macroExpandedComponentAst =
+            new MacroExpandedComponentAst((ComponentAst) originalParameter.getValue().getRight(), location,
+                                          moduleGlobalElementsNames, defaultGlobalElementSuffix, literalsParameters,
+                                          macroExpandedChildren);
+        return MuleArtifactAstCopyUtils.copyComponentTreeRecursively(macroExpandedComponentAst, UnaryOperator.identity());
```
fsgonz, line 149: static import. MuleArtifactAstCopyUtils and UnityOperator if possible.

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
```diff
@@ -128,6 +136,19 @@ private Object mapErrorMappings(final ComponentParameterAst originalParameter) {
             .getRight();
       }
 
+      private Object mapReconnection(final ComponentParameterAst originalParameter) {
+        MacroExpandedComponentAst macroExpandedComponentAst = (MacroExpandedComponentAst) originalParameter.getValue().getRight();
+        return MuleArtifactAstCopyUtils.copyComponentTreeRecursively(macroExpandedComponentAst, UnaryOperator.identity());
```
fsgonz, line 141: UnaryOperator.identity() static import.

fsgonz (changes requested):

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
```diff
@@ -107,14 +112,20 @@ public <T> Either<String, T> getValue() {
...
-                return mapErrorMappings(getDecorated());
-              }
-              return v;
-            });
+            .mapRight(this::mapComponent);
+      }
+
+      private <T> Object mapComponent(T rawValue) {
+        if (rawValue instanceof ArrayList) {
+          return rawValue;
```
elrodro83, line 120: what is this case needed for?

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
```diff
@@ -128,6 +139,19 @@ private Object mapErrorMappings(final ComponentParameterAst originalParameter) {
             .getRight();
       }
 
+      private boolean isComplexComponent() {
+        Optional<DslElementSyntax> dslElementSyntax = getDecorated().getGenerationInformation().getSyntax();
+        return dslElementSyntax.map(DslElementSyntax::supportsChildDeclaration).orElse(false);
```
fsgonz, line 144: Not only verify that this supports a Child declaration but also that is actually is declared as a child.

The difference between this:

    <petstore:config name="petstore" cashierName="jim">
        <petstore:pooled-connection username="#[vars.username]" password="#[vars.password]" />
        <petstore:pets>
            <petstore:pet value="la tota" />
            <petstore:pet value="la porota" />
        </petstore:pets>
    </petstore:config>

and this:

    <petstore:config name="petstore" cashierName="jim">
        <petstore:pooled-connection username="#[vars.username]" password="#[vars.password]" pets="#[vars.pets"]/>
    </petstore:config>

fsgonz (changes requested):

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
```diff
@@ -107,14 +112,20 @@ public <T> Either<String, T> getValue() {
...
-              return v;
-            });
+            .mapRight(this::mapComponent);
+      }
+
+      private <T> Object mapComponent(T rawValue) {
+        if (rawValue instanceof ArrayList) {
+          return rawValue;
+        } else if (rawValue instanceof String) {
+          return macroExpandedRawValue((String) rawValue);
```
fsgonz, line 122: Add a comment that can be something like this: "This is done to evaluate if the parameter is not declared as a child and can be an expression, so that it is resolved"

alopezbizon: Done. Please, review it.

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
alopezbizon: [Arrays](https://github.com/mulesoft/mule-artifact-ast/blob/8f3a72a3698b447c0503b0d8f4751bef0043852a/mule-ast/src/main/java/org/mule/runtime/ast/api/util/MuleAstUtils.java#L220) and [Maps](https://github.com/mulesoft/mule-artifact-ast/blob/8f3a72a3698b447c0503b0d8f4751bef0043852a/mule-ast/src/main/java/org/mule/runtime/ast/api/util/MuleAstUtils.java#L230) are contemplated by the code, so when it comes one of them, the rawValue is returned as it is. But with the new change as an Array supports Child Declaration, and doesn't need to be treated as other components, we have to exclude it and keep the same behaviour.

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
```diff
@@ -107,14 +111,39 @@ public <T> Either<String, T> getValue() {
...
+        // Evaluates if the parameter is not declared as a child and can be an expression so that it is resolved.
+        // In that case, the raw value is always a String.
+        return rawValue instanceof String;
+      }
+
+      private boolean isAnErrorMappings() {
+        return getDecorated().getModel().getName().equals(ERROR_MAPPINGS_PARAMETER_NAME);
+      }
+
+      private <T> boolean isAKnownType(T rawValue) {
```
fsgonz, line 138: I would change the isAKnownType ->  isResolvedByMetadataVisitor.
I would add the comments to the invocation of the methods which is where the reader of the code benefits from the comment.

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
```diff
@@ -107,14 +111,39 @@ public <T> Either<String, T> getValue() {
...
+          return macroExpandedRawValue((String) rawValue);
+        } else if (isAnErrorMappings()) {
+          return mapErrorMappings(getDecorated());
+        } else if (supportsChildDeclaration() && !isAKnownType(rawValue)) {
+          return copyMacroExpandedComponentTreeRecursively();
+        }
+        return rawValue;
+      }
+
+      private <T> boolean isAnExpression(T rawValue) {
```
fsgonz, line 128: I would change isAnExpression -> mayBeAnExpression.
I would add the comments to the invocation of the methods which is where the reader of the code benefits from the comment.

fsgonz (changes requested):

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
elrodro83: mmm, but from how thi is used, it looks that the value MUST be an expression, so the check for the markers would be missing here

elrodro83 (changes requested):

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
fsgonz: Rereading the code (Aye only added a method, didn't change what was originally there) I would name it must mustMacroExpandRawValue and not perform that verification. There is also some logic regarding adding a suffix if it is global which may not applie to the #[ ].

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
alopezbizon: I didn't change the location of the comment because I think it won't be clear where it belongs due to the formatter code and the ifs/else ifs.

fsgonz: ok. I think it is clear by the rename.

fsgonz (approved):
It's ok for me.

elrodro83 (approved):

PR #12800, opened 2023-09-14
W-14122647: modularize global-config module

mbuchwald:


modules/global-config/src/main/java/module-info.java
```diff
@@ -0,0 +1,26 @@
...
+ * @since 4.5
+ */
+module org.mule.runtime.global.config {
+
+  requires org.mule.runtime.api;
+  requires org.mule.runtime.api.annotations;
+  requires org.mule.runtime.container;
+  requires org.mule.runtime.core;
+  requires org.mule.runtime.maven.client.api;
+  requires config;
```
elrodro83, line 20: what is this?

elrodro83 (changes requested):

modules/global-config/src/main/java/module-info.java
mbuchwald: It's from [typesafe](https://github.com/mulesoft/mule/blob/master/modules/global-config/pom.xml#L26)

PR #12801, opened 2023-09-14
W-14124345: Modularize TLS module

mbuchwald:


modules/extensions-spring-support/pom.xml
```diff
@@ -17,6 +17,11 @@
 
     <properties>
         <surefire.args>${surefire.test.unit.open.args}</surefire.args>
+        <surefire.args>
```
elrodro83, line 20: remove the entry above, otherwise you are defining the same property twice

elrodro83 (changes requested):

elrodro83 (approved):

PR #12806, opened 2023-09-15
W-13994259: Single thread subscription of flow related Fluxes causes StackOverflow errors.

IvanAndresFritzler:


core/src/main/java/org/mule/runtime/core/internal/util/rx/RxUtils.java
```diff
@@ -406,4 +483,74 @@ public static <T> FluxSinkSupplier<T> createRoundRobinFluxSupplier(Function<Flux
...
+    @Override
+    public Context currentContext() {
+      return CoreSubscriber.super.currentContext().put(MULTI_FLUX_SUBSCRIBER, this);
+    }
+
+    @Override
+    public void onSubscribe(Subscription s) {
+      // Protect the subscriptionPhaser from potential extra calls.
+      synchronized (this) {
+        if (!subscribed) {
```
fsgonz, line 509: Can we add a logger with a warning a case extra calls happen, because a far as I understand it would be an unexpected behavior.

fsgonz (changes requested):

core/src/main/java/org/mule/runtime/core/internal/util/rx/RxUtils.java
```diff
@@ -406,4 +483,74 @@ public static <T> FluxSinkSupplier<T> createRoundRobinFluxSupplier(Function<Flux
     }, identity());
   }
 
+  public static class MultiFluxSubscriber<T> implements CoreSubscriber<T> {
```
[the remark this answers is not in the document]
fsgonz: Ok, this will be refactor later. It's clear for now as we talked.

core/src/main/java/org/mule/runtime/core/internal/util/rx/RxUtils.java
fsgonz: possible internal reactor behavior as discussed.

fsgonz (dismissed):

IvanAndresFritzler:
--validate

PR #12809, opened 2023-09-18
W-13145685: Modularize `mule-ee` modules - kryo and batch

elrodro83:


core/src/main/java/module-info.java
```diff
@@ -504,7 +520,8 @@
   opens org.mule.runtime.core.internal.profiling to
       spring.core;
   opens org.mule.runtime.core.internal.streaming to
-      spring.core;
+      spring.core,
+      kryo.shaded;
```
mbuchwald, line 524: Sort

mbuchwald (changes requested):

mbuchwald (approved):

PR #12839, opened 2023-09-26
W-14185880: Make objects used by DataWeave part of the api

elrodro83:


core/src/main/java/org/mule/runtime/core/api/event/DefaultFlowCallStack.java
```diff
@@ -0,0 +1,149 @@
+/*
+ * Copyright 2023 Salesforce, Inc. All rights reserved.
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.mule.runtime.core.api.event;
```
marianogonzalez, line 7: can't we expose an interface instead? or at most, make this priviledged

core/src/main/java/org/mule/runtime/core/api/el/dataweave/DataWeaveArtifactContext.java
```diff
@@ -25,14 +24,33 @@
  *
  * @since 4.0
  */
-public class DataWeaveArtifactContext extends AbstractArtifactContext {
+public class DataWeaveArtifactContext {
```
marianogonzalez, line 27: can't we expose an interface instead? or at most, make it priviledged

marianogonzalez (changes requested):

core/src/main/java/org/mule/runtime/core/api/event/DefaultFlowCallStack.java
elrodro83: DW doesn't handle interfaces, only concrete classes

core/src/main/java/org/mule/runtime/core/api/el/dataweave/DataWeaveArtifactContext.java
elrodro83: DW doesn't handle interfaces, only concrete classes

core/src/main/java/org/mule/runtime/core/privileged/event/DefaultFlowCallStack.java
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  @Override
+  public String toString() {
+    return doToString(FlowStackElement::toString);
+  }
+
+  public String toStringWithElapsedTime() {
+    return doToString(FlowStackElement::toStringWithElapsedTime);
+  }
```
mbuchwald, line 132: Doc in this one?

mbuchwald (changes requested):

mbuchwald (approved):

PR #12872, opened 2023-10-02
W-13038823 - Groovy Resource Releaser (#12571)

juanfloresMule:


modules/artifact/src/main/java/org/mule/module/artifact/classloader/GroovyResourceReleaser.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+
+import org.slf4j.Logger;
+
+/**
+ * A utility class to release all resources associated to Groovy Dependency on un-deployment to prevent classloader leaks.
+ *
+ * @since 4.4.0
```
asanguinetti, line 29: We'll need to change this if we backport it to 4.3.

asanguinetti (changes requested):
Just a small JDoc change.

But you will also need to add the 4.3.x suffix to the branch name so that it runs the validations against the proper branches from the other repositories.

Additionally, I have the understanding that we are not allowed to merge changes to versions that [are outside standard support](https://www.mulesoft.com/legal/versioning-back-support-policy#support-matrix) without VP approval.

PR #12875, opened 2023-10-02
W-13881167: Error handler with logException=false is still logging th…

alopezbizon:
…e exception when using Scheduler


modules/core-components/src/main/java/org/mule/runtime/core/internal/source/scheduler/SchedulerFlowProcessingTemplate.java
```diff
@@ -46,6 +48,10 @@ public void sendResponseToClient(CoreEvent response, Map<String, Object> paramet
   @Override
   public void sendFailureResponseToClient(MessagingException exception, Map<String, Object> parameters,
                                           CompletableCallback<Void> callback) {
+    if (defaultSchedulerMessageSource.featureFlaggingService != null
```
fsgonz, line 51: pass the flow as parameter to the constructor instead of setting the visibility to protected.

fsgonz (changes requested):

fsgonz (approved):

PR #12894, opened 2023-10-06
W-14209023: Support in mule-artifact.json of a property that specifies

elrodro83:
which java version the app works for

### DIFF
--- a/modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/deployable/AbstractDeployableArtifactDescriptorFactory.java
+++ b/modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/deployable/AbstractDeployableArtifactDescriptorFactory.java
@@ -6,20 +6,20 @@
  */
 package org.mule.runtime.module.artifact.activation.internal.deployable;
 
-import static org.mule.runtime.api.i18n.I18nMessageFactory.createStaticMessage;
 import static org.mule.runtime.container.api.MuleFoldersUtil.getMuleHomeFolder;
 import static org.mule.runtime.module.artifact.activation.internal.ExecutionEnvironment.isMuleFramework;
-import static org.mule.runtime.module.artifact.api.descriptor.BundleDescriptor.MULE_PLUGIN_CLASSIFIER;
 
 import static java.lang.String.format;
 import static java.util.Collections.emptySet;
+import static java.util.Collections.singleton;
 import static java.util.Optional.empty;
 import static java.util.Optional.of;
+
+import static org.apache.commons.lang3.SystemUtils.JAVA_SPECIFICATION_VERSION;
 
 import org.mule.runtime.api.deployment.meta.MuleArtifactLoaderDescriptor;
 import org.mule.runtime.api.deployment.meta.MuleDeployableModel;
 import org.mule.runtime.api.deployment.meta.MulePluginModel;
-import org.mule.runtime.module.artifact.activation.api.ArtifactActivationException;
 import org.mule.runtime.module.artifact.activation.api.deployable.DeployableProjectModel;
 import org.mule.runtime.module.artifact.activation.api.plugin.PluginDescriptorResolver;
 import org.mule.runtime.module.artifact.activation.api.plugin.PluginModelResolver;
@@ -44,7 +44,6 @@ import java.util.Set;
 
 import com.google.common.collect.ImmutableSet;
 
-import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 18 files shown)


modules/artifact/src/main/java/org/mule/runtime/module/artifact/internal/descriptor/SupportedJvmArtifactDescriptorValidator.java
```diff
@@ -0,0 +1,103 @@
...
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(SupportedJvmArtifactDescriptorValidator.class);
+
+  /**
+   * System property to set the enforcement policy. Defined here as a decision was made not to expose it as an API yet. For now,
+   * it will be for internal use only.
+   *
+   * @since 4.5.0
+   */
+  static final String DEPLOYABLE_ARTIFACT_JVM_ENFORCEMENT_PROPERTY = SYSTEM_PROPERTY_PREFIX + "jvm.version.extension.enforcement";
```
asanguinetti, line 45: Should this property be the same as the one for Extensions? if that is the case, shouldn't we change the name to something more generic that does not include the word `extension` in it.

tests/allure/src/main/java/org/mule/test/allure/AllureConstants.java
```diff
@@ -95,6 +95,16 @@ interface UndeploymentFailureStory {
...
 
     }
+
+    interface SupportedJavaVersions {
+
+      String JAVA_VERSIONS_IN_DEPLOYABLE_ARTIFACT =
+          "Supported Java Versions are reflected in the ApplicationModel or DomainModel";
+
+      String ENFORCE_DEPLOYABLE_ARTIFACT_JAVA_VERSION =
+          "Validate that all registered extensions support the Java version Mule is running on";
```
asanguinetti, line 105: Similarly, it looks like this description should be adapted to deployable artifacts and not extensions.

modules/artifact/src/main/java/org/mule/runtime/module/artifact/internal/descriptor/SupportedJvmArtifactDescriptorValidator.java
```diff
@@ -0,0 +1,103 @@
...
+ */
+public class SupportedJvmArtifactDescriptorValidator implements ArtifactDescriptorValidator {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(SupportedJvmArtifactDescriptorValidator.class);
+
+  /**
+   * System property to set the enforcement policy. Defined here as a decision was made not to expose it as an API yet. For now,
+   * it will be for internal use only.
+   *
+   * @since 4.5.0
```
asanguinetti, line 43: 4.6?

modules/artifact-activation/src/test/resources/apps/basic-with-supported-java-versions/src/main/java/org/test/TestClass.java
```diff
@@ -0,0 +1,11 @@
...
+ * Copyright 2023 Salesforce, Inc. All rights reserved.
+ * The software in this package is published under the terms of the CPAL v1.0
+ * license, a copy of which has been included with this distribution in the
+ * LICENSE.txt file.
+ */
+package org.test;
+
+public class TestClass {
+
+}
```
asanguinetti, line 11: Shouldn't we simplify the test artifact with just the basics needed for the test case?
It looks like just an empty Mule app with the required mule-artifact.json and the most basic pom.xml with no dependencies should be enough.
We could remove the test-script.dwl, test-resource.yaml and TestClass.java files as well as trimming the pom.xml a little.

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/deployable/AbstractDeployableArtifactDescriptorFactory.java
```diff
@@ -129,6 +128,14 @@ protected void doDescriptorConfig(T descriptor) {
     if (!isMuleFramework()) {
       descriptor.setLogConfigFile(getLogConfigFile(getArtifactModel()));
     }
+
+    if (getArtifactModel().getSupportedJavaVersions() != null
+        && !getArtifactModel().getSupportedJavaVersions().isEmpty()) {
+      descriptor.setSupportedJavaVersions(getArtifactModel().getSupportedJavaVersions());
+    } else {
+      // If the model doesn't have this data assume the current jvm is good for this deployment.
+      descriptor.setSupportedJavaVersions(singleton(JAVA_SPECIFICATION_VERSION));
```
asanguinetti, line 137: We have this logic in a couple of places for setting the descriptor with a default value when the model doesn't have any version specified.
Wouldn't it make more sense to leave the descriptor without any version specified just like the model and then have the default logic in the validator?

asanguinetti (changes requested):

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/policy/PolicyTemplateDescriptorFactory.java
```diff
@@ -75,6 +78,9 @@ public PolicyTemplateDescriptorFactory(ArtifactPluginDescriptorLoader artifactPl
   protected void doDescriptorConfig(MulePolicyModel artifactModel, PolicyTemplateDescriptor descriptor, File artifactLocation) {
     descriptor.setRootFolder(artifactLocation);
     descriptor.setPlugins(parseArtifactPluginDescriptors(descriptor));
+
+    // Policy model doesn't have this data, assume the current jvm is good for this deployment.
+    descriptor.setSupportedJavaVersions(singleton(JAVA_SPECIFICATION_VERSION));
```
asanguinetti, line 83: Do we still want to do this? or should we leave it empty?

asanguinetti (changes requested):

asanguinetti (approved):

PR #12905, opened 2023-10-10
W-14108905: Pass a timeout to the connectivity tester tryLock

eze210:
This gives some time to distributed locks to reach the quorum (otherwise they would return immediately false when the quorum isn't reached, making the connectivity tester assume that the lock is hold by other thread)

### DIFF
--- a/core/src/main/java/org/mule/runtime/core/internal/connection/DefaultConnectivityTesterFactory.java
+++ b/core/src/main/java/org/mule/runtime/core/internal/connection/DefaultConnectivityTesterFactory.java
@@ -57,16 +57,28 @@ public class DefaultConnectivityTesterFactory implements ConnectivityTesterFacto
 
   private static final Logger LOGGER = getLogger(DefaultConnectivityTesterFactory.class);
 
-  @Inject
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/mule/runtime/core/internal/connection/DefaultConnectivityTesterFactory.java
```diff
@@ -88,7 +90,7 @@ public void testConnectivity(ConnectionProvider provider, ConfigurationInstance
           public void doWork(RetryContext context) throws Exception {
             try {
               if (testConnectivityLock != null) {
-                final boolean lockAcquired = testConnectivityLock.tryLock();
+                final boolean lockAcquired = testConnectivityLock.tryLock(10, SECONDS);
```
elrodro83, line 93: add unit tests for this behaviour

elrodro83: make the time configurable

elrodro83 (changes requested):

PR #12914, opened 2023-10-12
W-14137848: Upgrade OTEL bom to the latest version

sofiamorseletto:


modules/metrics/mule-metrics-exporter-impl/pom.xml
```diff
@@ -54,6 +54,20 @@
...
+                    <groupId>com.squareup.okio</groupId>
+                    <artifactId>okio</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <!-- Added back because this was brought transitively by okio -->
+        <dependency>
+            <groupId>com.squareup.okio</groupId>
+            <artifactId>okio-jvm</artifactId>
+            <version>3.2.0</version>
```
fsgonz, line 69: Have you verified that this is the last version corresponding to the update in the otel pom?

modules/tracing/mule-tracer-exporter-impl/src/main/java/org/mule/runtime/tracer/exporter/impl/config/SpanExporterConfigurationUtils.java
```diff
@@ -15,14 +15,14 @@
 import static java.lang.Long.parseLong;
 import static java.time.Duration.ofSeconds;
 
+import io.opentelemetry.sdk.common.export.RetryPolicy;
```
fsgonz, line 18: Order of import.

modules/tracing/mule-tracer-exporter-impl/pom.xml
```diff
@@ -89,7 +85,11 @@
                 </exclusion>
             </exclusions>
         </dependency>
-        <!-- Added back because this was brought transitively by okio -->
+        <dependency>
```
fsgonz, line 88: why is this needed?

fsgonz (changes requested):

modules/tracing/mule-tracer-exporter-impl/pom.xml
sofiamorseletto: Which dependency do you mean?

fsgonz: Now you don't have the exclusion opentelemetry-exporter-otlp-common but in opentelemetry-exporter-otlp. Is that correct?

modules/metrics/mule-metrics-exporter-impl/pom.xml
sofiamorseletto: Yes it's 3.2.0

modules/tracing/mule-tracer-exporter-impl/pom.xml
sofiamorseletto: Yes, on the new opentelemetry version, okio comes with `opentelemetry-exporter-otlp` instead of `opentelemetry-exporter-otlp-common` so I moved the exclusion from `opentelemetry-exporter-otlp-common` to `opentelemetry-exporter-otlp`

fsgonz (approved):

sofiamorseletto:
--validate

PR #12933, opened 2023-10-19
W-14316889: Review failing tests in deployment module in Java 17

mbuchwald:


modules/container/src/main/java/org/mule/runtime/container/internal/DefaultPreFilteredContainerClassLoaderCreator.java
```diff
@@ -63,6 +66,11 @@ public Set<String> getBootPackages() {
     return finalBootPackages;
   }
 
+  @Override
+  public Set<String> getAdditionallyExportedResourceDirectories() {
```
rbourbonnavarro, line 70: Here is used `additionally` instead of `additional` as in the variable names, it'd be better to unify it.

rbourbonnavarro (approved):

PR #12936, opened 2023-10-23
W-13038823 - Groovy Resource Releaser (#12571)

acurci-at-mulesoft:


modules/artifact/src/main/java/org/mule/module/artifact/classloader/GroovyResourceReleaser.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+
+import org.slf4j.Logger;
+
+/**
+ * A utility class to release all resources associated to Groovy Dependency on un-deployment to prevent classloader leaks.
+ *
+ * @since 4.4.0
```
asanguinetti, line 29:
```suggestion
 * @since 4.3.0
```
We'll need to change this if we backport it to 4.3.

asanguinetti (changes requested):
Just a small JDoc change.

Additionally, I have the understanding that we are not allowed to merge changes to versions that [are outside standard support](https://www.mulesoft.com/legal/versioning-back-support-policy#support-matrix) without VP approval.

facunlk:
@asanguinetti this change is approved, if the review is ok, we can move forward with it.

asanguinetti (approved):

PR #12938, opened 2023-10-23
W-14154476: Add a fallback to use a classloader for service dependencies that cannot be loaded as modules

elrodro83:
* Revert as this change is no longer needed


elrodro83:
--validate

rbourbonnavarro (approved):

PR #12943, opened 2023-10-23
[fix/W-12388528] clean JMSCCThreadPoolMaster worker thread for ibm mq connector

cristianchan-mulesoft:
…

### DIFF
--- a/modules/artifact/src/main/java/org/mule/module/artifact/classloader/IBMMQResourceReleaser.java
+++ b/modules/artifact/src/main/java/org/mule/module/artifact/classloader/IBMMQResourceReleaser.java
@@ -20,6 +20,8 @@ import org.mule.sdk.api.artifact.lifecycle.ArtifactLifecycleListener;
 
 import java.lang.ref.Reference;
 import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.HashSet;
 import java.util.Hashtable;
 import java.util.Iterator;
@@ -27,11 +29,14 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.logging.Level;
+import java.util.stream.Collectors;
 import javax.management.MBeanRegistrationException;
 import javax.management.MBeanServer;
 import javax.management.MalformedObjectNameException;
 import javax.management.ObjectInstance;
 import javax.management.ObjectName;
+
+import org.apache.commons.lang3.ThreadUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


modules/artifact/src/main/java/org/mule/module/artifact/classloader/IBMMQResourceReleaser.java
```diff
@@ -388,4 +398,50 @@ private void processThreadLocalMap(Field threadLocalMapTableField, Object thread
...
+      Class<? extends Thread> threadClass = thread.getClass();
+      Method closeMethod = threadClass.getDeclaredMethod("close");
+      closeMethod.setAccessible(true);
+      closeMethod.invoke(thread);
+      LOGGER.info("Thread name : " + thread.getName());
+      thread.interrupt();
+    } catch (Throwable e) {
+      LOGGER.debug("An error occurred trying to close the '" + JMSCC_THREAD_POOL_MAIN_NAME + "' Thread", e);
+    }
+  }
```
asanguinetti, line 436: I'm not entirely sure that calling `close` on the `WorkQueueManagerThread` is necessary because it seems the `WorkQueueManager#close` already does that. Unless you tell me that in a previous version of the driver it wasn't like that. If that's the case we can probably encourage the customer to use a more recent version.

Avoiding this would be very important for Java 17 support because that inner class class is private and it wouldn't work without opening the module.

Finally, letting these threads finish doesn't really remove the static reference to them via `WorkQueueManager` -> `WorkQueueManagerImplementation manager` -> `WorkQueueManagerThread workManagerThread`. So there will still be a leak of the classloader of the artifact the connection resides in, unless it is the same as the one that loaded the `WorkQueueManager` class (which is usually the CL that has the driver as dep).

modules/artifact/src/test/java/org/mule/runtime/module/artifact/classloader/IBMMQResourceReleaserTestCase.java
```diff
@@ -179,6 +183,12 @@ public void mBeansTest() throws Exception {
     assertThat(countMBeans(artifactClassLoader), is(0));
   }
 
+  @Test
```
asanguinetti, line 186: Please also add a test for the scenario reported by the customer (with the driver declared in a shared domain classloader).

And also a test in which there are different applications, each with its driver and only one is undeployed and the other must not be affected.

modules/artifact/src/main/java/org/mule/module/artifact/classloader/IBMMQResourceReleaser.java
```diff
@@ -93,6 +100,9 @@ public void release() {
     LOGGER.debug("Releasing IBM MQ resources - Removes the static references held by the TraceController Class.");
     cleanPrivateStaticFieldForClass(IBM_MQ_TRACE_CLASS, "traceController");
 
+    LOGGER.debug("Releasing IBM MQ resources - Removes JMSCCThreadPoolMaster threads.");
+    disposeMQThreads();
```
asanguinetti, line 104: This disposal which was originally in the `mule-ibm-mq-connector` had [a property to disable it](https://github.com/mulesoft/mule-ibm-mq-connector/blob/f550d3a865f9375a8885d75560183b9d193047bf/src/main/java/com/mulesoft/connectors/ibmmq/internal/connection/IBMMQConnectionProvider.java#L281). We should probably keep that property to prevent breaking backwards.

modules/artifact/src/test/java/org/mule/runtime/module/artifact/classloader/IBMMQResourceReleaserTestCase.java
```diff
@@ -306,4 +316,10 @@ private int countJULKnownLevels(ClassLoader artifactClassLoader) throws Exceptio
     return counter;
   }
 
+  private int countWorkerThreads(ClassLoader artifactClassLoader) {
+    List<Thread> threads = ThreadUtils.getAllThreads().stream()
```
asanguinetti, line 320: static import

asanguinetti (changes requested):
Overall I'm not really convinced with this approach.
We are going to have a call to discuss it in real time.
I'm more inclined to think that this problem should be solved by making sure the TCCL during the call to `WorkQueueManager#initialize` is the same one that owns the driver.

PR #12944, opened 2023-10-24
W-14321097: XML SDK support for xmlns:tlsEnabled marker

asanguinetti:


asanguinetti:
Included as part of #12952

PR #12948, opened 2023-10-24
W-14357577: Provide an API for the GW to obtain a PolicyTemplateDescriptorFactory

elrodro83:


modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/MuleArtifactResourcesRegistry.java
```diff
@@ -278,13 +278,9 @@ private MuleArtifactResourcesRegistry(ArtifactClassLoader containerClassLoader,
 
     domainDescriptorFactory = new DomainDescriptorFactory(artifactPluginDescriptorLoader, descriptorLoaderRepository,
                                                           artifactDescriptorValidatorBuilder);
-    ApplicationDescriptorFactory applicationDescriptorFactory =
```
eze210, line 281: Why this part?

eze210 (changes requested):

modules/deployment-model-impl/src/main/java/org/mule/runtime/module/deployment/impl/internal/MuleArtifactResourcesRegistry.java
elrodro83: it was not being used

eze210: 🤦 ok then, LGTM

eze210 (approved):

PR #12952, opened 2023-10-27
W-14321226: Macro expanding the tlsContext passed in the configuration to the target element.

asanguinetti:
This PR is currently rebased on top of #12944. This is why, when looking at the changes against the main branch, you will see more changes than the subject implies.
The idea is to review and merge that one first. But we could also review and merge all at once in this one.


modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpansionModuleModel.java
```diff
@@ -363,15 +363,15 @@ public Optional<String> getComponentId() {
 
   private ComponentAst expandGlobalElement(List<ComponentAst> moduleComponentModels, Set<String> moduleGlobalElementsNames,
                                            ComponentAst comp, ConfigurationModel configurationModel) {
-    Map<String, String> propertiesMap = comp.getParameters()
+    Map<String, Object> propertiesMap = comp.getParameters()
         .stream()
         .filter(paramAst -> paramAst.getValue().isRight())
         .collect(toMap(paramAst -> paramAst.getModel().getName(),
-                       paramAst -> paramAst.getValue().getRight().toString()));
+                       paramAst -> paramAst.getValue().getRight()));
```
asanguinetti, line 370: This is necessary for preserving complex values defined inline.

asanguinetti (commented):
Just a comment for the reviewer

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
```diff
@@ -150,9 +171,11 @@ private Object mapErrorMappings(final ComponentParameterAst originalParameter) {
         return originalParameter.getValue()
             .mapRight(mappings -> ((List<ErrorMapping>) mappings)
                 .stream()
-                .map(mapping -> new ErrorMapping(literalsParameters.getOrDefault(mapping.getSource(), mapping.getSource()),
+                .map(mapping -> new ErrorMapping(literalsParameters.getOrDefault(mapping.getSource(), mapping.getSource())
```
marianogonzalez, line 174: i don't understand this change. It seems to be altering the logic quite a bit

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpansionModuleModel.java
```diff
@@ -545,6 +545,10 @@ private String getReplaceableExpression(String name, String prefix) {
     return "#[" + prefix + "." + name + "]";
   }
 
+  private boolean isExpression(Object value) {
```
marianogonzalez, line 548: use `org.mule.runtime.module.extension.internal.util.MuleExtensionUtils#isExpression` instead

modules/extensions-xml-support/src/test/resources/modules/module-tls-config-with-default.xml
```diff
@@ -0,0 +1,50 @@
...
+           http://www.mulesoft.org/schema/mule/module http://www.mulesoft.org/schema/mule/module/current/mule-module.xsd
+           http://www.mulesoft.org/schema/mule/tls http://www.mulesoft.org/schema/mule/tls/current/mule-tls.xsd
+           http://www.mulesoft.org/schema/mule/petstore http://www.mulesoft.org/schema/mule/petstore/current/mule-petstore.xsd ">
+
+    <property name="someUserConfig" type="string"/>
+    <property name="somePassConfig" type="string"/>
+
+    <petstore:config name="petstore-config" cashierName="cashierName">
+        <petstore:connection username="#[vars.someUserConfig]" password="#[vars.somePassConfig]" xmlns:tlsEnabled="true">
+            <tls:context>
```
marianogonzalez, line 17: so the semantics is that if no TLS is provided this one is used by default? Does this mean the tls parameter will always be optional?

marianogonzalez (changes requested):

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpandedComponentAst.java
asanguinetti: This is necessary to compensate for [this other change](https://github.com/mulesoft/mule/pull/12952/files/701011683b2cff91c81866072b15673267810d8b#r1374944032).
Before, it was stringified earlier, when generating the `propertiesMap` and, immediately, the `literalsParameters` map.
Now we are preserving the `Object` that is computed as the `getValue().getRight()` from the `ComponentParameterAst`. We have a Map<String, Object> instead of a Map<String, String>.
So when we get to this point, I'm doing the `toString` to preserve the previous behavior regarding the output value from this method.

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpansionModuleModel.java
asanguinetti: Will do.

modules/extensions-xml-support/src/test/resources/modules/module-tls-config-with-default.xml
asanguinetti: Yes, we will always be declaring the new synthetic parameter as optional (see [here](https://github.com/mulesoft/mule/pull/12952/files/701011683b2cff91c81866072b15673267810d8b#diff-2171aea98cf85c8e6bc80674bdeb462f3d559c2d8e931e48c7c1ae0fb79f293aR93)).

So, at least from the point of view of the XML module, it is always optional.

Now, I'm thinking there might be a case in which the target element didn't have the `TlsContextFactory` parameter as optional. Right now, such situation will force the developer of the XML module to provide a default TLS context, otherwise it will fail when validating the module's AST because it is missing a required parameter (previous macro-expansion). I need to check that.

marianogonzalez: make sure to reflect this in the ADR

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/loader/xml/TlsEnabledComponentUtils.java
```diff
@@ -111,6 +120,29 @@ private static boolean hasTlsContextFactoryParameter(ParameterizedModel model) {
     return model.getAllParameterModels().stream().anyMatch(TlsEnabledComponentUtils::isTlsContextFactoryParameter);
   }
 
+  private static boolean requiresTlsContextFactoryParameter(ParameterizedModel model) {
+    return model.getAllParameterModels().stream()
+        .filter(ParameterModel::isRequired)
+        .anyMatch(TlsEnabledComponentUtils::isTlsContextFactoryParameter);
+  }
+
+  private static boolean isTlsConfigurationRequired(ComponentAst componentAst) {
```
marianogonzalez, line 129: this method name is confusing. If this is to check if the paramater is requried, why mentioning the `Configuration` term? When should I use  this method instead of `requiresTlsContextFactoryParameter`

marianogonzalez (changes requested):

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/loader/xml/TlsEnabledComponentUtils.java
asanguinetti: Yes, you are right, we use the Configuration term for other things already and it may be confusing. I was thinking about the settings for the TLS context, not a Configuration in the sense of an Extension's Configuration.

In any case, I think this method could very well be an overload of `requiresTlsContextFactoryParameter`. One operates on a `ComponentAst` and the other one operates on a `ParameterizedModel`. Also one already delegates on the other. How does it sound?

modules/extensions-xml-support/src/main/java/org/mule/runtime/extension/internal/ast/MacroExpansionModuleModel.java
```diff
@@ -545,10 +547,6 @@ private String getReplaceableExpression(String name, String prefix) {
     return "#[" + prefix + "." + name + "]";
   }
 
-  private boolean isExpression(Object value) {
-    return value instanceof String && isExpression((String) value);
-  }
-
   private boolean isExpression(String value) {
```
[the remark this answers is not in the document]
asanguinetti: Sorry, that was sloppy on my part. Will do.

marianogonzalez (approved):

PR #12979, opened 2023-11-06
W-14350781: Enable GroovyResourceReleaser for each JDK. Removing incorrect cleaning method.

eze210:
1. Groovy can be used without mule-scripting-module, so the releaser should be part of the Mule Runtime instead of being part of the extension.
2. Before this PR, the GroovyResourceReleaser has two parts:
  a. The "unregisterAllClassesFromInvokerHelper" only uses reflection to access public methods, so there is no problem with JDK 17. This cleaning should be moved to the runtime code, and be enabled for all JDKs (today the whole GroovyResourceReleaser is deprecated and enabled for JDK version at most 11).
  b. The "cleanSpisEngines" part uses setAccessible for not-opened fields in order to clean references to the Groovy script engine that log4j discovered via SPI. This part of the releaser can't be used in JDK 17.
3. The cleaning done by "cleanSpisEngines" was removed. Why?
  a. It was cutting the references to groovy script engines from some objects present in this map: https://github.com/apache/logging-log4j2/blob/dcd30cd6942fd285766360bb458b6f7e8dd2001f/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AbstractManager.java#L82, but it wasn't cleaning the map itself, so it would be just a mitigation of the leak and not a fix.
  b. However, those objects are removed from the map when a LoggerContext is stopped. See this line: https://github.com/apache/logging-log4j2/blob/dcd30cd6942fd285766360bb458b6f7e8dd2001f/log4j-core/src/main/java/org/apache/logging/log4j/core/appender/AbstractManager.java#L115. Calling the stop method is enough to fix the leak.
  c. Mule Runtime already calls the LoggerContext#stop method, but after a timeout: it keeps a LoggerContextCache that is composed by two internal caches (one for "active" contexts, and another for "disposed" contexts). When a context is passed to "LoggerContextCache#remove", it's moved from the "active" to the "disposed" cache. The "disposed" cache has an expiration timeout (default is 15 seconds), and when the timeout is elapsed, it calls the actual "LoggerContext#stop()". It removes the entry from the log4j's map mentioned above.
  This timeout was added to give async logging some grace time to log messages after an artifact is disposed. See this class: https://github.com/mulesoft/mule/blob/fb275627d738ad1c49183bbdc054980900bcdcd1/modules/log4j-configurator/src/main/java/org/mule/runtime/module/log4j/internal/LoggerContextCache.java#L73
  d. The timeout can be changed in order to repeat the tests without that cache noise, by setting the property "-M-Dmule.log.context.dispose.delay.millis=10". When you test these scenarios with that property, the OOM can't be reproduced, even after commenting the "cleanSpisEngines" call.

Therefore, the solution here is:
1. Remove the lifecycle listener from the scripting-module code.
2. Remove the cleanSpisEngines() method.
3. Un-deprecate the GroovyResourceReleaser and make it run with all JDKs.

NOTE (about the PR):
- It also removes the aspectj weaver agent, because it holds a reference to the classloaders created by the test (and the test asserts that there aren't references to such classloader)
- The regression test for this change is enabling the same existing test for JDK17.

### DIFF
--- a/modules/artifact/pom.xml
+++ b/modules/artifact/pom.xml
@@ -15,6 +15,7 @@
         <surefire.args.base>
             -XX:+TieredCompilation
             -Dfile.encoding=UTF-8
+            -javaagent:${settings.localRepository}/org/aspectj/aspectjweaver/${aspectjVersion}/aspectjweaver-${aspectjVersion}.jar
             -javaagent:${settings.localRepository}/org/jacoco/org.jacoco.agent/${jacoco.version}/org.jacoco.agent-${jacoco.version}-runtime.jar=destfile='${session.executionRootDirectory}/target/jacoco.exec',excludes=mypackage.MyClass
         </surefire.args.base>
         <!-- open internal required stuff from org.mule.runtime.core to the test infrastructure -->
@@ -120,6 +121,12 @@
         <dependency>
             <groupId>org.mule</groupId>
             <artifactId>mule-maven-client-impl</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.aspectj</groupId>
+            <artifactId>aspectjweaver</artifactId>
+            <version>${aspectjVersion}</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/modules/artifact/pom.xml
+++ b/modules/artifact/pom.xml
@@ -15,7 +15,6 @@
         <surefire.args.base>
             -XX:+TieredCompilation
             -Dfile.encoding=UTF-8
-            -javaagent:${settings.localRepository}/org/aspectj/aspectjweaver/${aspectjVersion}/aspectjweaver-${aspectjVersion}.jar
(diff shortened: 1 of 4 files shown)


asanguinetti (approved):

asanguinetti (approved):

eze210:
--validate

eze210:
--validate

eze210:
--validate

eze210:
--validate

eze210:
--validate

eze210:
--validate

eze210:
--validate

eze210:
--validate

eze210:
--validate

eze210:
--validate

eze210:
--validate

eze210:
--validate

eze210:
--validate

eze210:
--validate

eze210:
--validate

PR #12958, opened 2023-10-30
W-14288864: Adapt JpmsUtils validations for supporting its use in an

elrodro83:
EmbeddedContainer client

PR #12985, opened 2023-11-07
W-14150316: Relaxing add-modules tweaking check to accomodate for wrapperless mode.

asanguinetti:


asanguinetti:
This should no longer be needed after #12958.

PR #13009, opened 2023-11-15
W-14417598: XML2Object transform is failing when it update to latest Runtime 3.9.3 or 3.9.5

marianorean:


fsgonz (changes requested):
We should add a test.

fsgonz (approved):

PR #13021, opened 2023-11-22
W-14488411: Provide an API for building/querying the registry

rbourbonnavarro:


elrodro83 (approved):

rbourbonnavarro:
--validate

rbourbonnavarro:
--validate

elrodro83 (dismissed):

eze210 (approved):

core/src/main/java/org/mule/runtime/core/api/config/ConfigurationBuilderRegistryFilter.java
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * Filters every object added to the target {@link MuleContext}'s registry, allowing customization when needed.
+   *
+   * @param serviceId           the ID of the object being added to the registry.
+   * @param serviceImplSupplier a {@link Supplier} for the object being added to the registry.
+   * @return the object to add to the registry, or {@link Optional#empty()} if no implementation should be added for the given
+   *         {@code serviceId}.
+   */
+  Optional<Object> filterRegisterObject(String serviceId, Supplier<Object> serviceImplSupplier);
```
marianogonzalez, line 31: shouldn't this be called filtered object instead? Also, shouldn't it return Optional<Supplier<Object>> instead? There's no point in feeding it with a supplier otherwise

core/src/main/java/org/mule/runtime/core/api/config/ConfigurationBuilderRegistryFilter.java
```diff
@@ -0,0 +1,40 @@
...
+
+  /**
+   * @param muleContext the target {@link MuleContext}.
+   * @return a {@link Map} with additional objects to add to the target {@link MuleContext}'s registry and their corresponding
+   *         IDs.
+   */
+  Map<String, Object> getAdditionalObjectsToRegister(MuleContext muleContext);
```
marianogonzalez, line 38: this shouldn't be the responsibility of a filter

core/src/main/java/org/mule/runtime/core/api/context/MuleContextBuilder.java
```diff
@@ -80,4 +83,17 @@ static MuleContextBuilder builder(ArtifactType artifactType) {
...
+
+  /**
+   * @return the {@link Registry} of the built {@link MuleContext}.
+   *         <p>
+   *         This method must be called after {@link #buildMuleContext()}, otherwise it will always return
+   *         {@link Optional#empty()}.
+   *
+   * @since 4.6.0
+   */
+  default Optional<Registry> getRegistry() {
```
marianogonzalez, line 95: the builder should most definitively not be capable of returning the registry, and there should be no expectation for the registry to exist before the build() mthod is invoked

marianogonzalez (changes requested):

core/src/main/java/org/mule/runtime/core/api/config/ConfigurationBuilderRegistryFilter.java
rbourbonnavarro: You mean the method to be called something like `filteredObject` or the javadoc?
The idea of receiving a supplier is to be able to prevent the creation of objects that we won't use, or defer it. For example, in the `mule-framework` we make the Expression Manager lazy and this way we can create it only when it's needed. However, we need to put an actual object in the registry, so returning it right away seemed appropriate.

core-tests/src/test/java/org/mule/runtime/core/internal/el/DefaultExpressionManagerDefaultTestCase.java
```diff
@@ -8,6 +8,7 @@
 
 import static org.mule.runtime.api.el.BindingContextUtils.NULL_BINDING_CONTEXT;
 import static org.mule.runtime.api.message.Message.of;
+import static org.mule.runtime.core.api.lifecycle.LifecycleUtils.initialiseIfNeeded;
```
elrodro83, line 11: ?

core/src/main/java/org/mule/runtime/core/api/lifecycle/LifecycleUtils.java
```diff
@@ -124,7 +124,9 @@ public static void initialiseIfNeeded(Object object, boolean inject, MuleContext
    * @throws InitialisationException
    */
   public static void initialiseIfNeeded(Collection<? extends Object> objects) throws InitialisationException {
-    initialiseIfNeeded(objects, null);
```
elrodro83, line 127: ?

core/src/main/java/org/mule/runtime/core/internal/el/DefaultExpressionManager.java
```diff
@@ -400,4 +403,10 @@ public void setMelDefault(boolean melDefault) {
   public String toString() {
     return this.getClass().getName() + "[" + (expressionLanguage != null ? expressionLanguage.toString() : "null") + "]";
   }
+
+  @Override
+  public void initialise() throws InitialisationException {
```
elrodro83, line 408: should this have dispose as well?

core/src/main/java/org/mule/runtime/core/internal/el/ExpressionLanguageAdaptorHandler.java
```diff
@@ -271,4 +274,10 @@ private String resolveLanguagePrefix(String expression) {
   public String toString() {
     return this.getClass().getName() + "[" + expressionLanguages.toString() + "]";
   }
+
+  @Override
+  public void initialise() throws InitialisationException {
```
elrodro83, line 279: dispose too

elrodro83 (changes requested):

core/src/main/java/org/mule/runtime/core/internal/el/DefaultExpressionManager.java
rbourbonnavarro: Done.

core/src/main/java/org/mule/runtime/core/internal/el/ExpressionLanguageAdaptorHandler.java
rbourbonnavarro: Done.

marianogonzalez (changes requested):

core/src/main/java/org/mule/runtime/core/internal/config/CustomServiceRegistry.java
```diff
@@ -27,4 +28,15 @@ public interface CustomServiceRegistry extends CustomizationService {
...
+
+  /**
+   * Overrides the given service implementation if a {@link ServiceOverrider} has been registered for the given {@code serviceId}.
+   *
+   * @param serviceId   identifier of the service.
+   * @param serviceImpl the service implementation to be overridden.
+   * @return the overridden service implementation, or {@link Optional#empty()} if it should be ignored.
+   * @since 4.6
+   */
+  Optional<Object> overrideDefaultService(String serviceId, Object serviceImpl);
```
marianogonzalez, line 40: why does this CustomServiecRegistry needs to exist? WHy does something callled registry extend CustomizationService?

marianogonzalez (changes requested):

core/src/main/java/org/mule/runtime/core/internal/config/DefaultCustomizationService.java
```diff
@@ -67,6 +80,61 @@ public Map<String, CustomService> getCustomServices() {
     return unmodifiableMap(customServices);
   }
 
+  @Override
+  public Optional<Object> overrideDefaultService(String serviceId, Object serviceImpl) {
```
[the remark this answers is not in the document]
rbourbonnavarro: It's API, we could deprecate it.

core-tests/src/test/java/org/mule/runtime/core/internal/el/DefaultExpressionManagerDefaultTestCase.java
rbourbonnavarro: Removed.

rbourbonnavarro:
--validate

rbourbonnavarro:
--validate

core-tests/src/test/java/org/mule/runtime/core/api/context/DefaultMuleContextFactoryTestCase.java
```diff
@@ -218,6 +229,46 @@ public void notifiesMuleContextEvents() throws InitialisationException, Configur
     inOrder.verify(listener).onInitialization(eq(context), any(Registry.class));
   }
 
+  @Test
```
elrodro83, line 232: allure, add at the class level

core/src/main/java/org/mule/runtime/core/internal/config/CustomService.java
```diff
@@ -54,7 +68,61 @@ public Optional<Class> getServiceClass() {
...
+
+    @Override
+    public void newServiceImpl(Object newServiceImpl) {
+      this.newServiceImpl = newServiceImpl;
+    }
+
+    @Override
+    public void skip() {
+      if (newServiceImpl != null) {
+        throw new IllegalStateException("A 'newServiceImpl' is already present");
```
elrodro83, line 108: improve the error message, give context tothe developer of what went wrong (add object key, object string representation, ...)

core/src/main/java/org/mule/runtime/core/internal/config/CustomService.java
```diff
@@ -54,7 +68,61 @@ public Optional<Class> getServiceClass() {
...
+      }
+
+      remove = true;
+    }
+
+    public Object getNewServiceImpl() {
+      if (remove) {
+        throw new IllegalStateException("Service set to be removed, can't be overridden");
```
elrodro83, line 116: improve the error message, give context tothe developer of what went wrong (add object key, object string representation, ...)

elrodro83 (changes requested):

elrodro83 (approved):

rbourbonnavarro:
--validate

rbourbonnavarro:
--validate

rbourbonnavarro:
--validate

rbourbonnavarro:
--validate

rbourbonnavarro:
--validate

rbourbonnavarro:
--validate

core-tests/src/test/java/org/mule/runtime/core/api/context/DefaultMuleContextFactoryTestCase.java
```diff
@@ -218,6 +235,46 @@ public void notifiesMuleContextEvents() throws InitialisationException, Configur
...
+  public void testCreateMuleContextWithCustomServices() throws InitialisationException, ConfigurationException {
+    MinimalConfigurationBuilder minimalConfigurationBuilder = new MinimalConfigurationBuilder();
+
+    Object testCustomServiceImpl = new Object();
+    Object customQueueManagerImpl = new Object();
+    Object customSecurityManagerImpl = new Object();
+    Map<String, Consumer<ServiceInterceptor<Object>>> interceptors = new HashMap<>();
+    // replace the default Security Manager
+    interceptors.put(OBJECT_SECURITY_MANAGER, serviceInterceptor -> {
+      assertThat(serviceInterceptor.getDefaultServiceImpl().isPresent(), is(true));
```
marianogonzalez, line 248: isn't it better to assert that the default is actually the expected instance?

core/src/main/java/org/mule/runtime/core/internal/config/CustomService.java
```diff
@@ -18,28 +23,31 @@
  *
  * @since 4.0
  */
-public class CustomService {
+public class CustomService<T> {
 
-  private Optional<Class> serviceClass;
-  private Optional<Object> serviceImpl;
+  private final String serviceId;
+  private final Optional<Class> serviceClass;
```
marianogonzalez, line 29: unneficient use of optional. Except very few cases, it's generally not a good practice to have Optional fields. Just keep the class and consumer and handle null internally.

core/src/main/java/org/mule/runtime/core/internal/config/CustomServiceRegistry.java
```diff
@@ -10,6 +10,7 @@
 
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.Function;
 
 public interface CustomServiceRegistry extends CustomizationService {
```
marianogonzalez, line 15: can we take the opportunity to rename this to something better?

marianogonzalez (changes requested):

core-tests/src/test/java/org/mule/runtime/core/api/context/DefaultMuleContextFactoryTestCase.java
rbourbonnavarro: I don't have access to the default value, but I could subclass `MinimalConfigurationBuilder` to have control over it.

rbourbonnavarro:
--validate

rbourbonnavarro:
--validate

marianogonzalez (approved):

PR #13045, opened 2023-11-28
Nested transactions / calling flows needs to be informed about the rollback

ca-stefan-cordes:
The transaction may be rolledback tue to timeout and then the calling flow must not continue when the child transaction has been rolled back.
The parent needs to be "informed" via Exception that something went wrong.


ca-stefan-cordes:
ping

PR #13143, opened 2024-01-04
W-14616115: Enable default events tracking element is missing in all

elrodro83:
core extension model elements that should support it

* Add support for runtime modules to add parameters on components from another module


runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/provider/enricher/RuntimeExtensionModelEnrichersProvider.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+
+/**
+ * Allows Mule modules to contribute enrichers for Runtime Extension Models.
+ */
+@NoImplement
+public interface RuntimeExtensionModelEnrichersProvider {
```
marianogonzalez, line 25: `@since`

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/provider/enricher/RuntimeExtensionModelEnrichersProvider.java
```diff
@@ -0,0 +1,48 @@
...
+                                                                                 RuntimeExtensionModelEnrichersProvider.class
+                                                                                     .getClassLoader())
+                                                                                         .iterator()).spliterator(),
+                  false)
+                      .filter(Objects::nonNull)
+                      .collect(toSet());
+  }
+
+  /**
+   *
```
marianogonzalez, line 43: empty line

marianogonzalez (changes requested):

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/provider/MuleExtensionModelDeclarer.java
```diff
@@ -691,6 +701,9 @@ private void declareFlow(ExtensionDeclarer extensionDeclarer) {
         .ofType(BASE_TYPE_BUILDER.numberType().integer().range(1, null).build())
         .withExpressionSupport(NOT_SUPPORTED);
 
+    addBusinessEventParameters(flow,
```
marianogonzalez, line 704: I will approve this anyway, but my suggestion would be to rename this method to something referencing the tracking module. Although the parameter is called business events, that's a user facing name. From the POV of a developer onboarding into the code base, this is much clearer if it refers to the actual module name and namespace that it represents

marianogonzalez (approved):

marianogonzalez (dismissed):

marianogonzalez (approved):

PR #13193, opened 2024-01-23
W-14875634: Start the polling of the apps folder in single app mode ad stop it once an app is deployed

fsgonz:


fsgonz:
--revalidate

fsgonz:
--validate

modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/DeploymentDirectoryWatcher.java
```diff
@@ -198,12 +211,14 @@ public void stop() {
...
-      stopArtifacts(domains);
-    } finally {
-      deploymentLock.unlock();
+    if (disposeArtifactsOnStop) {
+      try {
+        notifyStopListeners();
+        stopArtifacts(applications);
+        stopArtifacts(domains);
+      } finally {
+        deploymentLock.unlock();
```
marianogonzalez, line 220: how is this lock released if `disposeArtifactsOnStop == false`

marianogonzalez (changes requested):

modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/DeploymentDirectoryWatcher.java
fsgonz: A mistake. It should be inside the condition. 
Minor modification: I've changed the method notifyStopListeners so that it is clear what it does (nothing related to listeners but avoiding to persist the status of the app for not starting those apps in case of mule restart).

marianogonzalez (approved):

PR #13259, opened 2024-02-29
W-14954428: MetadataService and MetadataContext scope and router's propagation context

mbuchwald:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/DefaultMetadataContext.java
```diff
@@ -40,9 +47,50 @@ public class DefaultMetadataContext extends ExtensionResolvingContext implements
...
+   * @param connectionManager     {@link ConnectionManager} which is able to find a connection for the component using the
+   *                              {@param configInstance}
+   * @param cache                 instance of the {@link MetadataCache} for this context
+   * @param typeLoader            instance of a {@link ClassTypeLoader} in the context of this extension
+   * @param innerChainOutputType  Supplier of the {@link MetadataType} of the inner chain of a scope
+   */
+  public DefaultMetadataContext(Supplier<Optional<ConfigurationInstance>> configurationSupplier,
+                                ConnectionManager connectionManager, MetadataCache cache, ClassTypeLoader typeLoader,
+                                Supplier<MetadataType> innerChainOutputType) {
+    this(configurationSupplier, connectionManager, cache, typeLoader, of(innerChainOutputType), emptyMap());
```
marianogonzalez, line 66: of() -> ofNullable()

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/DefaultMetadataContext.java
```diff
@@ -40,9 +47,50 @@ public class DefaultMetadataContext extends ExtensionResolvingContext implements
...
+
+  private DefaultMetadataContext(Supplier<Optional<ConfigurationInstance>> configurationSupplier,
+                                 ConnectionManager connectionManager, MetadataCache cache, ClassTypeLoader typeLoader,
+                                 Optional<Supplier<MetadataType>> innerChainOutputType,
+                                 Map<String, Supplier<MetadataType>> innerRoutesOutputType) {
     super(configurationSupplier, connectionManager);
     this.cache = cache;
     this.classTypeLoader = typeLoader;
+    this.innerRoutesOutputType = innerRoutesOutputType;
+    this.innerChainOutputType = innerChainOutputType;
```
marianogonzalez, line 93: umodifiableMap()

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
```diff
@@ -785,11 +787,17 @@ public Message getSampleData() throws SampleDataException {
 
   protected <R> MetadataResult<R> runWithMetadataContext(Function<MetadataContext, MetadataResult<R>> contextConsumer)
       throws MetadataResolvingException, ConnectionException {
+    return runWithMetadataContext(contextConsumer, metadataCache -> () -> getMetadataContext(metadataCache));
+  }
+
+  protected <R> MetadataResult<R> runWithMetadataContext(Function<MetadataContext, MetadataResult<R>> contextConsumer,
+                                                         Function<MetadataCache, Callable<MetadataContext>> metadataContextSupplier)
```
marianogonzalez, line 794: why does this need to be a function and why is the cache the input?That doesn't sound very natural. Also, the IDEs get to supply their own Cache implementation so this is probably not super robust

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/ExtensionComponent.java
mbuchwald: This was just to avoid having to duplicate the code or fill with optional contexts here... If you think this last one sounds more natural, I have no bigger concern

marianogonzalez: not following. How does these functions provide what you mention? It's the modeling of the signatures what's weird to me. Can you clarify further?

marianogonzalez (changes requested):

marianogonzalez (approved):

PR #13280, opened 2024-03-04
W-14988896: discover `MuleContainerModule`s from given class loader/module layer

rbourbonnavarro:


modules/container/src/main/java11/org/mule/runtime/container/internal/ContainerModulesDiscovererProvider.java
```diff
@@ -34,4 +34,14 @@ public static ModuleDiscoverer containerModulesDiscoverer() {
     }
   }
 
+  public static ModuleDiscoverer containerModulesDiscoverer(Class<?> clazz) {
```
elrodro83, line 37: classloader instead of class

elrodro83 (changes requested):

PR #13307, opened 2024-03-11
W-15158118: Make OutputTypeResolver work for scopes and routers

mbuchwald:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ImmutableProcessorChildContextChainExecutor.java
```diff
@@ -151,4 +155,18 @@ public List<Processor> getMessageProcessors() {
   public CoreEvent getOriginalEvent() {
     return originalEvent;
   }
+
+  private ComponentLocation calculateComponentLocation(ComponentLocation chainLocation) {
+    if (!(chainLocation instanceof DefaultComponentLocation)) {
+      return chainLocation;
+    }
+    // As per W-15158118, we need to remove the 'processors' added to the location of the scope, that identifies the chain
```
marianogonzalez, line 163: why? Read that ticket and I don't find anything that justifies this. Or is it because of the tracing thing? really feels like you're hiding a bug here.

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/operation/ImmutableProcessorChildContextChainExecutor.java
mbuchwald: The problem is that in this object we need the scope's location, which is an annotation. But adding the exact location to the component creates the re-registration of the component in the component location, when the chain is not the operation itself, but the processors within this. Unless we decide to create a different annotation for this, I don't see a way of having the location but not getting the other behavior

mbuchwald: (I am in fact proposing to create another annotation)

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/runtime/privileged/ChildContextChain.java
```diff
@@ -19,6 +21,8 @@
  */
 public interface ChildContextChain extends Chain {
 
+  QName CHAIN_LOCATION = new QName("http://www.mulesoft.org/schema/mule/parser-metadata", "CHAIN_LOCATION");
```
marianogonzalez, line 24: add jdoc on what this represents and the corresponding `@since`

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ComponentMessageProcessorObjectFactory.java
```diff
@@ -82,7 +88,13 @@ public P doGetObject() {
                                                                              nestedChain)));
 
       // For MULE-18771 we need access to the chain's location to create a new event and sdk context
-      nestedChain.setAnnotations(this.getAnnotations());
+      // Update: For W-15158118, to avoid issues with the registration of the chain in the component locator, we are not
+      // adding the location as the location key but as another annotation.
+      Map<QName, Object> annotations = new HashMap<>(this.getAnnotations());
+      ComponentLocation chainLocation = ((DefaultComponentLocation) annotations.get(LOCATION_KEY));
```
marianogonzalez, line 94: Why cast? Also, isn't this the same as just calling this.getLocation()?

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ComponentMessageProcessorObjectFactory.java
```diff
@@ -82,7 +88,13 @@ public P doGetObject() {
                                                                              nestedChain)));
 
       // For MULE-18771 we need access to the chain's location to create a new event and sdk context
-      nestedChain.setAnnotations(this.getAnnotations());
+      // Update: For W-15158118, to avoid issues with the registration of the chain in the component locator, we are not
```
marianogonzalez, line 91: explain that the issue is that because both the chain and scope were sharing the same location, the chain overrides the scope in the Locator class

marianogonzalez (changes requested):

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ComponentMessageProcessorObjectFactory.java
```diff
@@ -82,7 +88,13 @@ public P doGetObject() {
...
+      // adding the location as the location key but as another annotation.
+      Map<QName, Object> annotations = new HashMap<>(this.getAnnotations());
+      ComponentLocation chainLocation = ((DefaultComponentLocation) annotations.get(LOCATION_KEY));
+      annotations.remove(LOCATION_KEY);
```
marianogonzalez, line 95: I think it's best for the chain to not return null when asked for its location. This can generate problesm down the line. I think the second annotation approach is fine, but the LOCATION_KEY should still be present and point to `{parentLocation}/processors` (as you had it before)

marianogonzalez (changes requested):

modules/extensions-spring-support/src/main/java/org/mule/runtime/module/extension/internal/config/dsl/ComponentMessageProcessorObjectFactory.java
mbuchwald: No, because we removed the location from the annotations previously (that was the point of this)

marianogonzalez (approved):

PR #13309, opened 2024-03-12
W-15226975: Core scopes and routers must be Operations and not Constructs

mbuchwald:


mbuchwald:
--validate

mbuchwald:
--validate

elrodro83 (dismissed):

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkOperationModelParser.java
```diff
@@ -403,14 +403,14 @@ private Stream<ComponentAstWithHierarchy> getOperationsAstRecursiveStream(Map<St
 
           if (filtered.contains(componentAst) || ignoreCondition.test(componentAstWithHierarchy)) {
             return Stream.empty();
+          } else if (filterCondition.test(componentAstWithHierarchy)) {
```
marianogonzalez, line 406: why this change? isn't this altering behavior? With the prior logic, it'd be possible for a component to be returned even though the filtered condition would also be true

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/provider/MuleExtensionModelDeclarer.java
```diff
@@ -798,14 +817,18 @@ private void declareScatterGather(ExtensionDeclarer extensionDeclarer, ClassType
         .withLayout(LayoutModel.builder().tabName(ADVANCED_TAB).build())
         .withModelProperty(new TargetModelProperty());
 
+    // Comment to remember for the PR: how to set array of any?
```
marianogonzalez, line 820: ArrayType with AnyType as the value

marianogonzalez (changes requested):

modules/extensions-mule-support/src/main/java/org/mule/runtime/module/extension/mule/internal/loader/parser/MuleSdkOperationModelParser.java
mbuchwald: The problem is that before a scope/router that should have been ignored won't be ignored. Since in the next line we check if there is an operation model, and then we proceed... but now the scopes and routers are operations, so we find an operation model for them, and we don't ignore them. 
The example is for checking if an operation is transactional, and it has an async which inside has a transactional try. That operation is not transactional, and we have this into consideration in the ignore corresponding to isTransactional.. but since now there is an operation model, we don't check if it has to be ignored

mbuchwald: And that logic was added in particular for cases like this one :/

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/provider/MuleExtensionModelDeclarer.java
```diff
@@ -849,13 +873,16 @@ private void declareParallelForEach(ExtensionDeclarer extensionDeclarer, ClassTy
         .withDisplayModel(DisplayModel.builder().displayName(TARGET_VALUE_PARAMETER_DISPLAY_NAME).build())
         .withLayout(LayoutModel.builder().tabName(ADVANCED_TAB).build())
         .withModelProperty(new TargetModelProperty());
+
+    parallelForeach.withOutput().ofDynamicType(BaseTypeBuilder.create(MetadataFormat.JAVA).arrayType().of(ANY_TYPE).build());
+    parallelForeach.withOutputAttributes().ofDynamicType(ANY_TYPE);
```
marianogonzalez, line 878: is it any or is it void? does it retain whatever attributes entered the parallel-foreach or does it nullify it?

mbuchwald: As far as I know, it keeps the attributes and combines them in case there is more than one

mbuchwald:
--validate

marianogonzalez (approved):

PR #13310, opened 2024-03-12
W-15173253: [Windows] Fix tests in org.mule.runtime.module.deployment.impl.internal.application.ApplicationDescriptorFactoryTestCase

marianorean:


modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/artifact/DeployableArtifactDescriptorFactoryTestCase.java
```diff
@@ -371,11 +377,80 @@ public void classLoaderConfigurationWithPluginDependencyAndAdditionalDependencie
   @Test
   public void classLoaderConfigurationWithPluginDependencyAndAdditionalDependenciesLightweightUseLocalRepository()
       throws Exception {
+    replacePlaceholderInClassloaderModel();
     final String location = "/plugin-dependency-with-additional-dependencies-lightweight-local-repository";
     populateRepositoryDependencies(new File(getArtifact(getArtifactRootFolder() + location), "local-repository"));
     assertClassLoaderConfigurationWithPluginDependencyAndAdditionalDependencies(location);
   }
 
+  private void replacePlaceholderInClassloaderModel() throws IOException {
```
fsgonz, line 386: can't be pass this to the non abstract class. For example creating a protected method that returns "apps" by default in ApplicationDescriptorFactoryTestCase and that is overridden in DomainDescriptorFactoryTestCase with "domains" and that is set in the root.

modules/deployment-model-impl/src/test/java/org/mule/runtime/module/deployment/impl/internal/artifact/DeployableArtifactDescriptorFactoryTestCase.java
```diff
@@ -371,11 +377,80 @@ public void classLoaderConfigurationWithPluginDependencyAndAdditionalDependencie
...
+
+  private static String replacePlaceholderWithSystemProperty(String content, String placeholder) {
+    // Replace ${placeholder} with the system property value
+    String placeholderString = "${" + placeholder + "}";
+    String replacement = System.getProperty(placeholder);
+
+    if (replacement != null) {
+      content = content.replace(placeholderString, replacement);
+    } else {
+      System.err.println("System property not set for " + placeholderString);
```
fsgonz, line 448: We don't use System.err. use a logging.

fsgonz (changes requested):

fsgonz (approved):

fsgonz (approved):

PR #13352, opened 2024-03-25
W-14739189: Simplify default tracing level handling for the tracing feature

marianorean:


modules/tracing/mule-tracer-configuration-impl/src/main/java/org/mule/runtime/tracing/level/impl/config/OpenTelemetryAutoConfigurableTracingLevelConfiguration.java
```diff
@@ -0,0 +1,84 @@
...
+import org.mule.runtime.tracing.level.api.config.TracingLevel;
+import org.mule.runtime.tracing.level.api.config.TracingLevelConfiguration;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Consumer;
+
+import javax.inject.Inject;
+
+public class OpenTelemetryAutoConfigurableTracingLevelConfiguration implements TracingLevelConfiguration {
```
fsgonz, line 28: add java doc.

fsgonz (changes requested):

modules/tracing/mule-tracer-configuration-api/src/main/java/org/mule/runtime/tracing/level/api/config/TracingLevelConfiguration.java
```diff
@@ -16,10 +16,15 @@
 public interface TracingLevelConfiguration {
 
   /**
-   * @return the default tracing level, MONITORING, if no other tracing level is specified from a configuration.
+   * @return the default tracing level, if no other tracing level is specified from a configuration.
```
IvanAndresFritzler, line 19: Lets put "@return The configured tracing level." 

The reason for this is that the default is also a configuration. Default configs must belong to the implementations, not the API.

modules/tracing/mule-tracer-configuration-api/src/main/java/org/mule/runtime/tracing/level/api/config/TracingLevelConfiguration.java
```diff
@@ -16,10 +16,15 @@
...
    */
   TracingLevel getTracingLevel();
 
+  default TracingLevel getTracingLevel(TracingLevel defaultLevel) {
```
IvanAndresFritzler, line 23: You should not need this. Defaults must be implementation only.

modules/tracing/mule-tracer-configuration-api/src/main/java/org/mule/runtime/tracing/level/api/config/TracingLevelConfiguration.java
```diff
@@ -29,6 +34,11 @@ public interface TracingLevelConfiguration {
    */
   TracingLevel getTracingLevelOverride(String location);
 
+  default TracingLevel getTracingLevelOverride(String location, TracingLevel defaultLevel) {
```
IvanAndresFritzler, line 37: You should not need this. Defaults must be implementation only.

modules/tracing/mule-tracer-configuration-impl/src/main/java/org/mule/runtime/tracing/level/impl/config/FileTracingLevelConfiguration.java
```diff
@@ -203,31 +194,17 @@ private JsonNode getTracingLevelConfiguration() {
       if (parseBoolean(spanExporterConfiguration.getStringValue(MULE_OPEN_TELEMETRY_EXPORTER_ENABLED, "false"))) {
         if (LOGGER.isDebugEnabled()) {
           LOGGER
-              .debug(format("Non existent or non parseable tracing level config file found. The tracing level will be set to the default: %s",
-                            defaultLevel),
+              .debug(format("Non existent or non parseable tracing level config file found. The tracing level will be set to the default."),
```
IvanAndresFritzler, line 197: This is the part where you need to move the default to the AutoConfigurable class only. This one (The FileConfiguration) will no longer know about a default. It's the AutoConfigurable one the one that handles the default when the File one has failed (because of a typo in the config) or returned null. This is the class where you need to do a bit of refactor. Also, the message must not change, try not to loose the default value.

Also, take a look at the getTracingLevel() method because you will need to start checking the raw config value instead of the parsed value there (you no longer have the default value so the raw, file configured value is the only one you can strore).

modules/tracing/mule-tracer-configuration-impl/src/main/java/org/mule/runtime/tracing/level/impl/config/FileTracingLevelConfiguration.java
```diff
@@ -203,31 +194,17 @@ private JsonNode getTracingLevelConfiguration() {
...
                      e);
         } else {
           LOGGER
-              .info(format("Non existent or non parseable tracing level config file found. The tracing level will be set to the default: %s. Enable DEBUG log level to see the exception",
-                           defaultLevel));
+              .info(format("Non existent or non parseable tracing level config file found. The tracing level will be set to the default. Enable DEBUG log level to see the exception"));
```
IvanAndresFritzler, line 201: Same as before (try to refactor this logic into the AutoConfigurable one)

modules/tracing/mule-tracer-configuration-impl/src/main/java/org/mule/runtime/tracing/level/impl/config/OpenTelemetryAutoConfigurableTracingLevelConfiguration.java
IvanAndresFritzler, line 28: I think the tracing levels are not tied to OpenTelemetry. It's not like the SpanExporter config or the MetricsExporter config where you have OTEL specifics, this is a general, Mule Runtime feature so the name should be just AutoConfigurableTracingLevelConfiguration.

modules/tracing/mule-tracer-configuration-impl/src/main/java/org/mule/runtime/tracing/level/impl/config/OpenTelemetryAutoConfigurableTracingLevelConfiguration.java
```diff
@@ -0,0 +1,84 @@
...
+
+  @Inject
+  private MuleContext muleContext;
+  private SpanExporterConfiguration spanExporterConfiguration;
+  private TracingLevelConfiguration delegate;
+  private TracingLevel defaultLevel =
+      valueOf(getProperty(MULE_OPEN_TELEMETRY_EXPORTER_DEFAULT_TRACING_LEVEL, MONITORING.toString()).toUpperCase());
```
IvanAndresFritzler, line 35: We cannot change the name of the system property because of backwards compatibility (yikes) but it should be named "EXPORTERS_DEFAULT_TRACING_LEVEL". Maybe we can clarify that in the JavaDoc of the System property.

modules/tracing/mule-tracer-configuration-impl/src/test/java/org/mule/runtime/tracing/level/impl/config/OpenTelemetryTracingLevelConfigTestCase.java
```diff
@@ -0,0 +1,153 @@
...
+import org.mule.runtime.tracing.level.api.config.TracingLevel;
+import org.mule.runtime.tracing.level.api.config.TracingLevelConfiguration;
+
+import io.qameta.allure.Feature;
+import io.qameta.allure.Story;
+import org.junit.Test;
+
+@Feature(PROFILING)
+@Story(TRACING_CONFIGURATION)
+public class OpenTelemetryTracingLevelConfigTestCase {
```
IvanAndresFritzler, line 32: This is also not OTEL dependant, so name should be AutoConfigurableTracingLevelConfigurationTestCase, I think.

IvanAndresFritzler (changes requested):

modules/tracing/mule-tracer-configuration-impl/src/main/java/org/mule/runtime/tracing/level/impl/config/AutoConfigurableTracingLevelConfiguration.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  @Inject
+  public void setSpanExporterConfiguration(SpanExporterConfiguration spanExporterConfiguration) {
+    this.spanExporterConfiguration = spanExporterConfiguration;
+  }
+
+  @Override
+  public TracingLevel getTracingLevel() {
+    return getTracingLevel(false, null);
```
IvanAndresFritzler, line 67: Why not delegate.getTracingLevel?

modules/tracing/mule-tracer-configuration-impl/src/main/java/org/mule/runtime/tracing/level/impl/config/AutoConfigurableTracingLevelConfiguration.java
```diff
@@ -0,0 +1,106 @@
...
+
+  public TracingLevel getTracingLevel(boolean isOverride, String location) {
+    try {
+      if (delegate == null) {
+        this.delegate = new FileTracingLevelConfiguration(muleContext);
+        consumersOnChange.forEach(consumer -> this.delegate.onConfigurationChange(consumer));
+      }
+      TracingLevel level = isOverride ? delegate.getTracingLevelOverride(location) : delegate.getTracingLevel();
+      return level != null ? level : defaultLevel;
+    } catch (IllegalArgumentException e) {
```
IvanAndresFritzler, line 83: Why is this needed?

IvanAndresFritzler (changes requested):

modules/tracing/mule-tracer-configuration-impl/src/main/java/org/mule/runtime/tracing/level/impl/config/FileTracingLevelConfiguration.java
```diff
@@ -111,16 +91,14 @@ private void setTracingLevels() {
...
+    String configuredTracingLevel = getStringValue(LEVEL_PROPERTY_NAME);
     if (configuredTracingLevel != null) {
       try {
-        configuredTracingLevel = propertyResolver.apply(configuredTracingLevel);
         tracingLevel = valueOf(configuredTracingLevel.toUpperCase(Locale.ROOT));
       } catch (IllegalArgumentException e) {
-        LOGGER
-            .error(format("Wrong tracing level found in configuration file: %s. The tracing level will be set to the default level: %s",
-                          configuredTracingLevel, defaultLevel));
+        throw new IllegalArgumentException(
```
IvanAndresFritzler, line 99: This must be moved to the AutoConfigurable class.

modules/tracing/mule-tracer-configuration-impl/src/main/java/org/mule/runtime/tracing/level/impl/config/AutoConfigurableTracingLevelConfiguration.java
marianorean: that expeption is thrown [here](https://github.com/mulesoft/mule/pull/13352/files#diff-2c719d8516c51af6c354e87bbd91995b620bf23cb6cea2b497e973b0dff19d6eR99)

IvanAndresFritzler (changes requested):

modules/tracing/mule-tracer-configuration-impl/src/main/java/org/mule/runtime/tracing/level/impl/config/AutoConfigurableTracingLevelConfiguration.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Consumer;
+
+import javax.inject.Inject;
+
+import org.slf4j.Logger;
+
+/**
+ * Autoconfiguration of tracing level. Will use MONITORING as the default level if there is not one configured in a config file.
```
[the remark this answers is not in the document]
marianorean: but it is the default in case there is no default configured, right?

modules/tracing/mule-tracer-configuration-impl/src/main/java/module-info.java
```diff
@@ -30,4 +32,6 @@
       org.mule.runtime.spring.config,
       spring.beans;
 
+  opens org.mule.runtime.tracing.level.impl.config to spring.core;
```
IvanAndresFritzler, line 35: Why is this needed?

modules/tracing/mule-tracer-configuration-impl/src/main/java/org/mule/runtime/tracing/level/impl/config/AutoConfigurableTracingLevelConfiguration.java
marianorean: Not only the error handling but also the creation of the delegate if necessary. I'll use the supplier.

IvanAndresFritzler (changes requested):

modules/tracing/mule-tracer-configuration-impl/src/main/java/module-info.java
marianorean: This seems to be needed for the tests in java 17.

`Error creating bean with name '_muleCoreComponentTracerFactory': Unsatisfied dependency expressed through field 'initialSpanInfoProvider'; nested exception is org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name '_muleTracerInitialSpanInfoProvider': Unsatisfied dependency expressed through method 'setTracingLevelConfiguration' parameter 0; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name '_muleTracingLevelConfiguration': Injection of autowired dependencies failed; nested exception is java.lang.reflect.InaccessibleObjectException: Unable to make field private org.mule.runtime.core.api.MuleContext org.mule.runtime.tracing.level.impl.config.AutoConfigurableTracingLevelConfiguration.muleContext accessible: module org.mule.runtime.tracer.configuration.impl does not "opens org.mule.runtime.tracing.level.impl.config" to module spring.core`

modules/tracing/mule-tracer-configuration-impl/src/test/java/org/mule/runtime/tracing/level/impl/config/AutoConfigurableTracingLevelConfigurationTestCase.java
```diff
@@ -0,0 +1,158 @@
...
+        new TestNoFileTracingLevelConfiguration(muleContext);
+    AutoConfigurableTracingLevelConfiguration tracingLevelConfiguration =
+        new AutoConfigurableTracingLevelConfiguration(muleContext);
+    tracingLevelConfiguration.setDelegate(fileTracingLevelConfiguration);
+    tracingLevelConfiguration.setSpanExporterConfiguration(spanExporterConfiguration);
+    assertThat(tracingLevelConfiguration.getTracingLevel(), is(DEFAULT_LEVEL));
+  }
+
+  @Test
+  public void whenLevelIsWrongInFileDefaultLevelIsReturned() {
```
IvanAndresFritzler, line 66: This tests (and the rest) need to go back to the FIleConfigurationTestCase and test for the expected result there (error, empty response or whatever should the FileConfiguration return) Here we need to mock the delegate and test the conditions where we return the default or not.

modules/tracing/mule-tracer-configuration-impl/src/main/java/org/mule/runtime/tracing/level/impl/config/AutoConfigurableTracingLevelConfiguration.java
```diff
@@ -0,0 +1,117 @@
...
+  private final List<Consumer<TracingLevelConfiguration>> consumersOnChange = synchronizedList(new ArrayList<>());
+
+  @Inject
+  public AutoConfigurableTracingLevelConfiguration(MuleContext muleContext) {
+    this(muleContext, new FileTracingLevelConfiguration(muleContext));
+  }
+
+  protected AutoConfigurableTracingLevelConfiguration(MuleContext muleContext, TracingLevelConfiguration delegate) {
+    this.muleContext = muleContext;
+    this.delegate = delegate;
```
IvanAndresFritzler, line 63: Validate that the delegate cannot be null.

modules/tracing/mule-tracer-configuration-impl/src/main/java/org/mule/runtime/tracing/level/impl/config/AutoConfigurableTracingLevelConfiguration.java
```diff
@@ -0,0 +1,117 @@
...
+    if (delegate != null) {
+      delegate.onConfigurationChange(onConfigurationChangeConsumer);
+    } else {
+      consumersOnChange.add(onConfigurationChangeConsumer);
+    }
+  }
+
+  @Override
+  public void dispose() {
+    if (delegate != null) {
```
IvanAndresFritzler, line 113: You no longer need this since the delegate is guaranteed to be not null.

modules/tracing/mule-tracer-configuration-impl/src/test/java/org/mule/runtime/tracing/level/impl/config/AutoConfigurableTracingLevelConfigurationTestCase.java
```diff
@@ -0,0 +1,75 @@
...
+    AutoConfigurableTracingLevelConfiguration tracingLevelConfiguration =
+        new AutoConfigurableTracingLevelConfiguration(muleContext, fileTracingLevelConfiguration);
+    tracingLevelConfiguration.setSpanExporterConfiguration(mock(SpanExporterConfiguration.class));
+    assertThat(tracingLevelConfiguration.getTracingLevel(), is(OVERVIEW));
+  }
+
+  @Test
+  public void whenDelegateReturnsNullDefaultLevelIsReturned() {
+    MuleContext muleContext = mock(MuleContext.class);
+    FileTracingLevelConfiguration fileTracingLevelConfiguration =
```
IvanAndresFritzler, line 52: It's better to directly mock TracingLevelConfiguration here and not FileTracingLevelConfiguration

IvanAndresFritzler (changes requested):

modules/tracing/mule-tracer-configuration-impl/src/test/java/org/mule/runtime/tracing/level/impl/config/AutoConfigurableTracingLevelConfigurationTestCase.java
```diff
@@ -0,0 +1,76 @@
...
+@Story(TRACING_CONFIGURATION)
+public class AutoConfigurableTracingLevelConfigurationTestCase {
+
+  private static final TracingLevel DEFAULT_LEVEL = MONITORING;
+
+  @Test
+  public void returnValueFromDelegate() {
+    MuleContext muleContext = mock(MuleContext.class);
+    TracingLevelConfiguration delegate =
+        mock(FileTracingLevelConfiguration.class);
```
IvanAndresFritzler, line 42: mock(TracingLevelConfiguration.class)

IvanAndresFritzler (changes requested):

IvanAndresFritzler (approved):

fsgonz (approved):

PR #13381, opened 2024-04-10
W-15343349: Verify initialization of configuration beans for configur…

alopezbizon:
…ation in tracking


fsgonz (changes requested):

fsgonz (dismissed):

fsgonz (approved):

modules/tracing/mule-tracer-configuration-impl/src/main/java/org/mule/runtime/tracing/level/impl/config/FileTracingLevelConfiguration.java
```diff
@@ -51,41 +52,53 @@
...
+  private final LazyValue<TracingLevel> tracingLevelInitializer = new LazyValue<>(() -> {
+    initialise();
+    return null;
+  });
   private TracingLevel tracingLevel = null;
-  private JsonNode configuration;
 
-  private List<Runnable> onConfigurationChangeRunnables = synchronizedList(new ArrayList<>());
+  private final MuleContext muleContext;
+  private JsonNode configuration;
```
fsgonz, line 73: the json part has to be decoupled from this. Possibly the json retrieval should not be an abstract method.

modules/tracing/mule-tracer-configuration-impl/src/main/java/org/mule/runtime/tracing/level/impl/config/FileTracingLevelConfiguration.java
```diff
@@ -51,41 +52,53 @@
...
+  private final String CONFIGURATION_FILE_PATH =
+      getProperty(MULE_OPEN_TELEMETRY_TRACING_CONFIGURATION_FILE_PATH,
+                  getConfFolder() + FileSystems.getDefault().getSeparator() + getPropertiesFileName());
+
   private final HashMap<String, TracingLevel> tracingLevelOverrides = new HashMap<>();
-  private boolean tracingConfigurationFileWatcherInitialised;
+  private final List<Runnable> onConfigurationChangeRunnables = synchronizedList(new ArrayList<>());
+  private final LazyValue<TracingLevel> tracingLevelInitializer = new LazyValue<>(() -> {
+    initialise();
+    return null;
```
fsgonz, line 68: why is this is a return null. make it return the corresponding tracinglevel.

fsgonz (commented):
In general try to encapsulate the fact that this is a file managed by a json file and this file will only now that it is managing the tracing level properties no matter what kind of file the configuration is stored.

modules/tracing/mule-tracer-configuration-impl/src/main/java/org/mule/runtime/tracing/level/impl/config/FileTracingLevelConfiguration.java
```diff
@@ -51,41 +52,53 @@
...
-  private JsonNode configuration;
 
-  private List<Runnable> onConfigurationChangeRunnables = synchronizedList(new ArrayList<>());
+  private final MuleContext muleContext;
+  private JsonNode configuration;
   private URL configurationUrl;
-  private TracingConfigurationFileWatcher tracingConfigurationFileWatcher;
   private ConfigurationPropertiesResolver propertyResolver;
-  private boolean initialised;
+  private TracingConfigurationFileWatcher tracingConfigurationFileWatcher;
```
fsgonz, line 76: The tracing level configuration watcher should not be here either. this should be decoupled to the other file as in the previous configuration.

modules/tracing/mule-tracer-configuration-impl/src/main/java/org/mule/runtime/tracing/level/impl/config/FileTracingLevelConfiguration.java
```diff
@@ -51,41 +52,53 @@
...
   private Runnable getOnConfigurationChanged() {
     return () -> onConfigurationChangeRunnables.forEach(Runnable::run);
   }
 
   private void setTracingLevels() {
     configuration = getTracingLevelConfiguration();
-    propertyResolver =
-        new DefaultConfigurationPropertiesResolver(empty(),
-                                                   new SystemPropertiesConfigurationProvider());
+    propertyResolver = new DefaultConfigurationPropertiesResolver(empty(), new SystemPropertiesConfigurationProvider());
```
fsgonz, line 101: same this. This file should only know of tracing levels no matter where it takes the values for the keys.

fsgonz (changes requested):
Let's go back to the original idea in this sprint.

modules/tracing/mule-tracer-configuration-impl/src/main/java/org/mule/runtime/tracing/level/impl/config/FileTracingLevelConfiguration.java
alopezbizon: This LazyValue variable has the only task of running the `initialise()` method. There is no point in returning the tracing level because, as it will be executed only one time, the value could become outdated at some point. On the other hand, the tracing level is already managed in another variable and settled up by the `setTracingLevel()` method.

fsgonz (approved):

PR #13399, opened 2024-04-16
W-15261626: Set OutputTypeResolvers to Core's Scope and Routers

mbuchwald:


modules/extensions-soap-support/src/main/java/org/mule/runtime/module/extension/soap/internal/loader/SoapInvokeOperationDeclarer.java
```diff
@@ -109,7 +109,7 @@ void declare(ConfigurationDeclarer configDeclarer, ClassTypeLoader loader, Set<E
   }
 
   private void declareMetadata(OperationDeclarer operation, ClassTypeLoader loader) {
-    new ComponentMetadataConfigurer()
+    new DefaultComponentMetadataConfigurer()
```
marianogonzalez, line 112: why not using your new and shiny factory?

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/provider/MuleExtensionModelDeclarer.java
```diff
@@ -164,7 +167,13 @@ class MuleExtensionModelDeclarer {
   private static final String TRACKING_NAMESPACE = "tracking";
   private static final String TRACKING_NAMESPACE_URI = "http://www.mulesoft.org/schema/mule/ee/tracking";
 
-  ExtensionDeclarer createExtensionModel() {
+  private final Optional<ComponentMetadataConfigurer.ComponentMetadataConfigurerFactory> configurerFactory;
+
+  public MuleExtensionModelDeclarer(ComponentMetadataConfigurer.ComponentMetadataConfigurerFactory configurerFactory) {
+    this.configurerFactory = ofNullable(configurerFactory);
+  }
+
```
marianogonzalez, line 175: this shouldn't be optional. If null, it's a bug and it should fail. Also, using an Optional as an if(null) substitute is a documented anti-pattern. Optional is for communicating nullability, not for handling it internally

runtime-extension-model/src/test/java/org/mule/runtime/core/extension/CoreExtensionDeclarerTestCase.java
```diff
@@ -0,0 +1,149 @@
...
+    assertThat(scopesPassThrough.stream().map(this::getNameOfDeclarer).collect(toList()),
+               containsInAnyOrder("untilSuccessful", "parallelForeach", "try"));
+  }
+
+  private String getNameOfDeclarer(ParameterizedDeclarer declarer) {
+    return ((NamedDeclaration) declarer.getDeclaration()).getName();
+  }
+
+
+  private class MockConfigurer implements ComponentMetadataConfigurer {
```
marianogonzalez, line 60: why not just use Mockito? you can use a captor to assert over the arguments of the configure() method

marianogonzalez (changes requested):

runtime-extension-model/src/test/java/org/mule/runtime/core/extension/CoreExtensionDeclarerTestCase.java
mbuchwald: The problem is that the `configure` method is separated from how it is configured, that's why I first store "which type it is" to then add it to the corresponding list

marianogonzalez: you can still do that with mockito, but whatever is easier for you

marianogonzalez (approved):

marianogonzalez (approved):

PR #13410, opened 2024-04-23
W-15228973: Provide expected pom version when artifact version contains variable

hyerans:


hyerans:
--validate

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/maven/AbstractMavenDeployableProjectModelBuilder.java
```diff
@@ -107,6 +109,14 @@ protected AbstractMavenDeployableProjectModelBuilder(MavenConfiguration mavenCon
   public final DeployableProjectModel build() {
     File pom = getPomFromFolder(projectFolder);
 
+    Properties pomProperties;
+    String version = "";
```
elrodro83, line 113: Use an `Optional` to handle empty values rather than an empty string

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/maven/MavenUtilsForArtifact.java
```diff
@@ -0,0 +1,79 @@
...
+        break;
+      }
+      File[] directories = lookupFolder.listFiles((FileFilter) DIRECTORY);
+      checkState(directories != null || directories.length == 0,
+                 format("No directories under %s so pom.properties file for artifact in folder %s could not be found",
+                        lookupFolder.getAbsolutePath(), artifactFolder.getAbsolutePath()));
+      checkState(directories.length == 1,
+                 format("More than one directory under %s so pom.properties file for artifact in folder %s could not be found",
+                        lookupFolder.getAbsolutePath(), artifactFolder.getAbsolutePath()));
+      lookupFolder = directories[0];
```
elrodro83, line 69: why is this and the while needed? In what cases can the properties be nested in a lower level?

modules/artifact-activation/src/test/java/org/mule/runtime/module/artifact/activation/internal/deployable/LightweightDeployableProjectModelBuilderTestCase.java
```diff
@@ -89,6 +91,26 @@ public void createDeployableProjectModelWithAdditionalDependenciesInAPlugin() th
     assertThat(descriptor.getArtifactId(), is("derby"));
   }
 
+  @Test
+  @Issue("W-15228973")
+  @Description("When a version is provided using pom properties, include the information when building DeployableProjectModel.")
+  public void buildDeployableProjectModelWithPomProperty() throws Exception {
+    installArtifact(getResourceFolder("apps/lightweight/application-using-additional-libraries"),
```
elrodro83, line 98: what is you set `versionToParse` as a property to a value different than `0`? It should take that instead. Add a test case for that scenario.

elrodro83 (changes requested):

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/maven/AbstractMavenDeployableProjectModelBuilder.java
```diff
@@ -107,6 +112,17 @@ protected AbstractMavenDeployableProjectModelBuilder(MavenConfiguration mavenCon
...
     File pom = getPomFromFolder(projectFolder);
 
+    Properties pomProperties;
+    Optional<String> version = empty();
+
+    if (projectFolder.equals(APP) || projectFolder.equals(DOMAIN) || projectFolder.isDirectory()) {
+      try {
+        pomProperties = getPomPropertiesFolder(projectFolder);
+        version = ofNullable(pomProperties.getProperty("version"));
+      } catch (Exception e) {
```
elrodro83, line 122: handle this exception properly

modules/artifact-activation/src/test/java/org/mule/runtime/module/artifact/activation/internal/deployable/LightweightDeployableProjectModelBuilderTestCase.java
```diff
@@ -89,6 +92,29 @@ public void createDeployableProjectModelWithAdditionalDependenciesInAPlugin() th
...
```
elrodro83, line 99: add a test passing system properties that are reflected in the pom.properties

elrodro83 (changes requested):

modules/artifact-activation/src/test/java/org/mule/runtime/module/artifact/activation/internal/deployable/LightweightDeployableProjectModelBuilderTestCase.java
hyerans: added `DeployableProjectModelVersionParseTestCase` for the cases:
1) case that uses version from pom.properties
2) case that uses system property that was passed

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/maven/AbstractMavenDeployableProjectModelBuilder.java
hyerans: added a debug logger - there's a case where pom.properties doesn't exist and getPomPropertiesFolder would throw an exception for that. We are fine with not having the file as this is the first approach to find the artifact's version. just added a log.

jesica-fera: Shouldn't we catch FileNotFoundException or something similar here (just more specific than Exception)? just to avoid swallowing the exception in other cases

hyerans: replaced it with ArtifactDescriptorCreateException to specify the case

jesica-fera (dismissed):
LGTM

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/maven/AbstractMavenDeployableProjectModelBuilder.java
```diff
@@ -107,9 +122,32 @@ protected AbstractMavenDeployableProjectModelBuilder(MavenConfiguration mavenCon
...
+        logger.debug("unable to get version info from pom.properties:" + e.getMessage());
+      }
+    }
+
     List<String> activeProfiles = mavenConfiguration.getActiveProfiles().orElse(emptyList());
     MavenPomParser parser = MavenPomParserProvider.discoverProvider().createMavenPomParserClient(pom.toPath(), activeProfiles);
 
+    // 2) if a version is passed using system properties, use the version instead
+    String originalPomVersion = parser.getModel().getVersion();
+    Pattern pattern = compile("\\$\\{\\s*([^}]*)\\s*\\}");
```
elrodro83, line 143: make the pattern a static field. document what the pattern is looking for

modules/artifact-activation/src/main/java/org/mule/runtime/module/artifact/activation/internal/maven/AbstractMavenDeployableProjectModelBuilder.java
```diff
@@ -107,9 +122,32 @@ protected AbstractMavenDeployableProjectModelBuilder(MavenConfiguration mavenCon
...
+    Matcher matcher = pattern.matcher(originalPomVersion);
+    if (matcher.find()) {
+      String potentialProperty = matcher.group(1);
+      if (getProperty(potentialProperty) != null) {
+        version = ofNullable(originalPomVersion.replaceAll("\\$\\{[^}]*\\}", getProperty(potentialProperty)));
```
elrodro83, line 148: use the capturing group from the previos regex instead of creating a new one for replaceAll

elrodro83 (changes requested):

elrodro83 (approved):

hyerans:
--validate

PR #13421, opened 2024-04-24
W-15599112: secure-properties:config with unresolved key param causes introspection to fail

elrodro83:


modules/properties-config/src/main/java/org/mule/runtime/config/api/properties/PropertiesResolverUtils.java
```diff
@@ -112,26 +114,54 @@ public static Supplier<Map<String, ConfigurationProperty>> createGlobalPropertie
...
+  }
+
+  /**
+   * @param artifactAst                    the {@link ArtifactAst} to get the {@link ConfigurationPropertiesProvider} from.
+   * @param externalResourceProvider       a {@link ResourceProvider} to use to read files when needed for properties resolution.
+   * @param localResolver                  A resolver that retrieves properties that are used when resolving parameters of a
+   *                                       {@link ConfigurationPropertiesProvider}.
+   * @param ignoreCreateProviderExceptions if {@code true}, exceptions that occur when calling
+   *                                       {@link ConfigurationPropertiesProviderFactory#createProvider(org.mule.runtime.ast.api.ComponentAst, java.util.function.UnaryOperator, ResourceProvider)}
+   *                                       are just logged instead of rethrown.
```
asanguinetti, line 127: Javadoc says the exceptions will be logged but I'm not seeing the logging code here. If it is someone else doing the logging maybe we shouldn't mention it in the contract.

PR #13431, opened 2024-04-29
W-15398905: Fixing types in Scope and Router's type resolvers 

mbuchwald:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultExtensionDesignTimeResolversFactory.java
```diff
@@ -265,6 +267,16 @@ public MetadataContext createMetadataContext(Supplier<Optional<ConfigurationInst
     return new DefaultMetadataContext(configurationSupplier, connectionManager, cache, typeLoader);
   }
 
+  @Override
+  public MetadataContext createMetadataContext(Supplier<Optional<ConfigurationInstance>> configurationSupplier,
```
marianogonzalez, line 271: approving for now but need to revisit this class and the commit that introduced it (see conversation in mule-framework channel)

marianogonzalez (approved):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/config/DefaultExtensionDesignTimeResolversFactory.java
mbuchwald: Checked, and agreed. Previous to my rebase I didn't have this :/

PR #13451, opened 2024-05-07
W-15707567: Disable deprecated methods in eExtensionsClient

elrodro83:


rbourbonnavarro (approved):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/DefaultExtensionsClient.java
```diff
@@ -281,6 +286,10 @@ private ExtensionModel findExtension(String extensionName) {
   public <T, A> CompletableFuture<Result<T, A>> executeAsync(String extensionName,
                                                              String operationName,
                                                              OperationParameters parameters) {
+    // Prevent using code that relies on org.mule.runtime.extension.internal.client.ComplexParameter
+    if (featureFlaggingService.isEnabled(UNSUPPORTED_EXTENSIONS_CLIENT_RUN_ASYNC)) {
```
marianogonzalez, line 290: why this? what if you try to use a connector that uses this such as wsc?

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/client/DefaultExtensionsClient.java
elrodro83: it will need to stop using the deprecated method for supproting java 21

PR #13452, opened 2024-05-08
W 15708770 - Adding logger statements to increase traceability for sdk context

ManiS84:
###  Description
Adding logger statements to SdkInternalContext is modified in the execution path.
### Reason For Change
As part of the parent investigation [W-15618324](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001q7MTlYAM/view) we encountered NPE sporadically. To understand what is causing the issue, it would be beneficial to add logger statements while creating, removing and configuring the SdkInternalContext to an event

### DIFF
--- a/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/SdkInternalContext.java
+++ b/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/SdkInternalContext.java
@@ -36,7 +36,9 @@ import static org.slf4j.LoggerFactory.getLogger;
  * @since 4.3
  */
 public class SdkInternalContext implements EventInternalContext<SdkInternalContext> {
+
   private static final Logger LOGGER = getLogger(SdkInternalContext.class);
+
   /**
    * Extracts an instance stored in the given {@code event}
    *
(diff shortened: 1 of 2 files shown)


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/SdkInternalContext.java
```diff
@@ -56,10 +60,12 @@ public static SdkInternalContext from(CoreEvent event) {
   private Function<Context, Context> innerChainSubscriberContextMapping = identity();
 
   public void removeContext(ComponentLocation location, String eventId) {
+    LOGGER.debug("Removing context at location - {} for event - {}", location.getLocation(), eventId);
```
asanguinetti, line 63: I'm worried that in some corner case here `location` could be `null`. Hence, we would be causing an NPE that wouldn't otherwise had happened.

We could use the version of this with the supplier which would make this theoretical problem never happen unless the logging level is DEBUG.

But maybe we also want to check if `location` is `null` when logging.

Maybe something like this:
```suggestion
    LOGGER.debug("Removing context at location - {} for event - {}", () -> {location != null ? location.getLocation() : "null"}, eventId);
```

asanguinetti: Of course this applies to all the other new log messages.

asanguinetti (changes requested):

asanguinetti (approved):

asanguinetti (approved):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/runtime/execution/SdkInternalContext.java
ManiS84: Completed the changes suggested

PR #13504, opened 2024-05-22
W-15832989: single app mode doesn't correctly dispose deployed artifa…

fsgonz:
…ct on graceful shutdown


alopezbizon (approved):

modules/deployment/src/main/java/org/mule/runtime/module/deployment/internal/singleapp/SingleAppDeploymentService.java
```diff
@@ -321,7 +322,7 @@ public void onDeploymentFailure(String artifactName, Throwable cause) {
     this.deploymentDirectoryWatcher.start();
   }
 
-  private DeploymentDirectoryWatcher resolveDeploymentDirectoryWatcher() {
+  protected DeploymentDirectoryWatcher resolveDeploymentDirectoryWatcher() {
```
chris-ebert, line 325: Another way to do this would be to replace this with a constructor parameter that was a Supplier<DeploymentDirectoryWatcher> - the default implementation would more or less do the same thing as the current code and you could pass in an alternate implementation for tests. That way you wouldn't have to have a whole extra class just to get the test behavior you wanted...

PR #13598, opened 2024-07-01
W-16124318: Batch module fails with "Could not serialize object" when client id enforcement policy is enabled

elrodro83:


eze210 (approved):

elrodro83:
--validate

PR #13602, opened 2024-07-01
W-15676027: Chain input type resolution - pass-through and collection-based - updates on core EM

asanguinetti:


modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/metadata/DefaultComponentMetadataConfigurer.java
```diff
@@ -164,6 +173,9 @@ public DefaultComponentMetadataConfigurer setConnected(boolean connected) {
 
   @Override
   public DefaultComponentMetadataConfigurer asOneOfRouter() {
+    if (chainInputTypeResolver == null) {
```
marianogonzalez, line 176: not a good default

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/metadata/DefaultComponentMetadataConfigurer.java
```diff
@@ -191,6 +209,10 @@ public <T extends ParameterizedDeclarer, D extends ParameterizedDeclaration> voi
 
   @Override
   public <T extends ComponentDeclaration> void configure(ParameterizedDeclaration<T> declaration) {
+    if (routesChainInputTypesResolvers.isEmpty() && chainInputTypeResolver != null
+        && declaration instanceof WithNestedComponentsDeclaration) {
+      applyChainInputTypeResolverToAllRoutes((WithNestedComponentsDeclaration<?>) declaration);
```
marianogonzalez, line 214: what is this for? Let's avoid auto magical defaults. Also remember that the base line here is the SDK, not the core langauge

marianogonzalez (changes requested):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/metadata/DefaultComponentMetadataConfigurer.java
asanguinetti: I know, we talked about this some time ago. I'm going to change it. Was just uploading the draft PR. Thanks for the reminder though.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/metadata/DefaultComponentMetadataConfigurer.java
marianogonzalez: SDK users won't use the configurer directly, but the SDK implementation should. Otherwise we would have repeated code on how to configure these things.

In either case, yes, this seems way too automagical to me

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/metadata/DefaultComponentMetadataConfigurer.java
asanguinetti: done

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/api/metadata/DefaultComponentMetadataConfigurer.java
asanguinetti: Ok forget about the last paragraph, the annotation goes to the parameter, so they will need to add it to each route anyway.

asanguinetti: Resolved in https://github.com/mulesoft/mule/pull/13602/commits/43051b4b364026be6a1e9b5396f6648252260ec9

mbuchwald (approved):

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/DefaultMetadataContext.java
```diff
@@ -67,11 +74,36 @@ public DefaultMetadataContext(Supplier<Optional<ConfigurationInstance>> configur
...
+   * @param connectionManager        {@link ConnectionManager} which is able to find a connection for the component using the
+   *                                 {@param configInstance}
+   * @param cache                    instance of the {@link MetadataCache} for this context
+   * @param typeLoader               instance of a {@link ClassTypeLoader} in the context of this extension
+   * @param scopePropagationContext  an optional {@link ScopeOutputMetadataContext} for the inner chain. Value is only present for
+   *                                 scope components
+   * @param routerPropagationContext an optional {@link RouterOutputMetadataContext} with routes information. Value is only
+   *                                 present for router components
+   */
+  public DefaultMetadataContext(Supplier<Optional<ConfigurationInstance>> configurationSupplier,
```
marianogonzalez, line 94: `@since`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/DefaultMetadataContext.java
```diff
@@ -107,4 +139,14 @@ public ClassTypeLoader getTypeLoader() {
   public BaseTypeBuilder getTypeBuilder() {
     return BaseTypeBuilder.create(JAVA);
   }
+
+  @Override
+  public Optional<ExpressionLanguageMetadataService> getExpressionLanguageMetadataService() {
```
marianogonzalez, line 144: `@since`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/DefaultMetadataContext.java
```diff
@@ -107,4 +139,14 @@ public ClassTypeLoader getTypeLoader() {
...
+    return expressionLanguageMetadataService;
+  }
+
+  @Override
+  public Optional<TypeBindings> getTypeBindings() {
```
marianogonzalez, line 149: `@since`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataInputDelegate.java
```diff
@@ -117,7 +120,8 @@ MetadataResult<MessageMetadataType> getScopeChainInputType(MetadataContext conte
     } catch (ConnectionException e) {
       return connectivityFailure(e);
     } catch (Exception e) {
-      return failure(newFailure(e).withMessage("Failed to resolve input types for scope inner chain").onComponent());
+      return failure(newFailure(e).withMessage(format("Failed to resolve input types for scope inner chain: %s", e.getMessage()))
```
marianogonzalez, line 123: concatenate instead of `format`

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataInputDelegate.java
```diff
@@ -155,6 +163,12 @@ MetadataResult<Map<String, MessageMetadataType>> getRoutesChainInputType(Metadat
     }
   }
 
+  private List<String> getRouteNames() {
+    return ((ComposableModel) model).getNestedComponents().stream()
+        .filter(nested -> nested instanceof NestedRouteModel)
+        .map(NamedObject::getName)
+        .collect(toList());
```
marianogonzalez, line 170: don't collect, return the Stream

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/provider/MuleExtensionModelDeclarer.java
```diff
@@ -645,6 +655,11 @@ private void declareForEach(ExtensionDeclarer extensionDeclarer, ClassTypeLoader
     forEach.withOutput().ofType(VOID_TYPE);
     forEach.withOutputAttributes().ofType(VOID_TYPE);
 
+    configurerFactory.create()
+        .setKeysResolver(new NullMetadataResolver(), "collection", STRING_TYPE, false)
```
marianogonzalez, line 659: not a string

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/provider/MuleExtensionModelDeclarer.java
```diff
@@ -697,7 +712,11 @@ private void declareChoice(ExtensionDeclarer extensionDeclarer) {
         .setExecutionOccurrence(ONCE_OR_NONE);
     choice.withOutput().ofDynamicType(ANY_TYPE);
     choice.withOutputAttributes().ofDynamicType(ANY_TYPE);
-    configurerFactory.create().asOneOfRouter().configure(choice);
+    configurerFactory.create()
```
marianogonzalez, line 715: 👏🏻

runtime-extension-model/src/main/java/org/mule/runtime/core/internal/extension/ForEachChainInputTypeResolver.java
```diff
@@ -0,0 +1,56 @@
...
+import org.mule.sdk.api.metadata.resolving.ChainInputTypeResolver;
+
+/**
+ * {@link ChainInputTypeResolver} for the ForEach and ParallelForEach scopes.
+ *
+ * @since 4.8.0
+ */
+public class ForEachChainInputTypeResolver implements ChainInputTypeResolver {
+
+  private static final String PARAMETER_NAME = "collection";
```
marianogonzalez, line 25: overkill

runtime-extension-model/src/main/java/org/mule/runtime/core/internal/extension/ForEachCollectionTypeResolver.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Override
+  public String getResolverName() {
+    return "FOREACH_INPUT_COLLECTION";
+  }
+
+  @Override
+  public MetadataType getInputMetadata(MetadataContext context, String key)
+      throws MetadataResolvingException, ConnectionException {
+    if (context instanceof InternalMetadataContext) {
```
marianogonzalez, line 47: it should fail if not true. Otherwise you're hiding a bug

runtime-extension-model/src/main/java/org/mule/runtime/core/internal/extension/ForEachCollectionTypeResolver.java
```diff
@@ -0,0 +1,110 @@
...
+      InternalMetadataContext internalContext = (InternalMetadataContext) context;
+      if (internalContext.getExpressionLanguageMetadataService().isPresent() && internalContext.getTypeBindings().isPresent()) {
```
marianogonzalez, line 49: same.. fail if not

runtime-extension-model/src/main/java/org/mule/runtime/core/internal/extension/ForEachCollectionTypeResolver.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+        return ((ArrayType) resolvedType).getType();
+      }
+    }
+
+    // Fallback to any
+    return ANY_TYPE;
+  }
+
+  private String sanitizeExpression(String expression) {
```
marianogonzalez, line 65: isn't this exact logic already present someplace else. This definitively rings a bell

marianogonzalez (changes requested):

runtime-extension-model/src/main/java/org/mule/runtime/core/internal/extension/ForEachChainInputTypeResolver.java
asanguinetti: Do you mean the constant? but if I don't define it, the same string would be repeated in two places. I'd like to make sure they are always in sync.

runtime-extension-model/src/main/java/org/mule/runtime/core/internal/extension/ForEachCollectionTypeResolver.java
asanguinetti: Yes, but I couldn't find any implementation that has been apified that I could reuse.
One is in [core](https://github.com/mulesoft/mule/blob/master/core/src/main/java/org/mule/runtime/core/internal/el/ExpressionLanguageUtils.java#L65).
Another one in [mule-framework](https://github.com/mulesoft/mule-framework/blob/master/mule-framework/src/main/java/org/mule/framework/internal/tooling/type/propagation/cfg/TypePropagationCFGVisitor.java#L567).
Maybe we can apify this on [MuleExpressionLanguage](https://github.com/mulesoft/mule-api/blob/master/src/main/java/org/mule/runtime/api/el/MuleExpressionLanguage.java#L20), what do you think?

runtime-extension-model/src/main/java/org/mule/runtime/core/internal/extension/ForEachCollectionTypeResolver.java
asanguinetti: The `TypeBindings` will only be available when doing propagation. If this is executed as part of plain type resolution, it is very possible that there are no `TypeBindings` provided (they can be provided, but they probably won't). I think falling back to ANY makes sense there.

runtime-extension-model/src/main/java/org/mule/runtime/core/internal/extension/ForEachCollectionTypeResolver.java
asanguinetti: We will only have this in the context of DietTooling right now. So, I'm not sure failing is a good idea, unless we are sure MTS is not used for a ForEach.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/DefaultMetadataContext.java
asanguinetti: The `@since` is in the `InternalMetadataContext` interface.

modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/DefaultMetadataContext.java
asanguinetti: The `@since` is in the `InternalMetadataContext` interface.

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/provider/MuleExtensionModelDeclarer.java
asanguinetti: Unfortunately we can't make it be the same type of the parameter in this case, for a couple of reasons:
1- The value is expected to be an expression. We can't resolve expressions when doing type resolution because we don't have an execution context. So the only value we can provide is the unresolved expression. The closest metadata type we have to represent that is `String`. If we were to use the parameter type here (`Iterable<Object>`), there would be a mismatch. Fortunately, `String` is also the key type that the `InpuTypeResolver` needs in this case.
2- There is [a check](https://github.com/mulesoft/mule/blob/master/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataKeyIdObjectResolver.java#L303) in the `MetadataKeyIdObjectResolver` which explicitly says `'Iterable' type is invalid for MetadataKeyId parameters, use String type instead`. It seems only Boolean, String or Object metadata types are allowed.

marianogonzalez: please explain this is a comment

runtime-extension-model/src/main/java/org/mule/runtime/core/internal/extension/ForEachChainInputTypeResolver.java
marianogonzalez: but you don't reuse it. This string will be kept in memory forever, even at runtime.

runtime-extension-model/src/main/java/org/mule/runtime/core/internal/extension/ForEachCollectionTypeResolver.java
marianogonzalez: have mixed feelings about the expression language being in charge of sanitizing its own expressions, although I see its merits... @elrodro83 what do you think?

As for the impl, I would apify the core one. THe framework one doesn't seem to be doing the exact same thing

marianogonzalez (changes requested):

runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/provider/MuleExtensionModelDeclarer.java
asanguinetti: The type there is used as the expected type of the metadata key object (see [MetadataKeyIdObjectResolver](https://github.com/mulesoft/mule/blob/master/modules/extensions-support/src/main/java/org/mule/runtime/module/extension/internal/metadata/MetadataKeyIdObjectResolver.java)). Up to this point it has always been the case that the type of the metadata key object is the same of the parameter annotated with `MetadataKeyId`.

This case is different, the [parameter is of ArrayType](https://github.com/mulesoft/mule/blob/master/runtime-extension-model/src/main/java/org/mule/runtime/core/api/extension/provider/MuleExtensionModelDeclarer.java#L617) but our metadata key (which is derived from it) is **the expression** itself. 

If we were to say that the type of the metadata key needs to be an `ArrayType` just like the parameter, not only we don't have a way to extract an array from the expression (remember we can't evaluate it at design time), but also it wouldn't be useful for our `InputTypeResolver` that depends on that key.

Our `InputTypeResolver` needs the expression so that it can give it to the `ExpressionLanguageMetadataService` so that it can tell it what would be the output type of the expression considering other types bound in context (payload, attributes, vars).

Because (AFAIK) we don't have a special MetadataType to represent an expression, I'm using StringType.

runtime-extension-model/src/main/java/org/mule/runtime/core/internal/extension/ForEachChainInputTypeResolver.java
asanguinetti: I do use it, [here](https://github.com/mulesoft/mule/pull/13602/files/788389de81e8038c5ed2d1759b8de6aae5ff4b19#diff-d357c4ff11c93c077669d118a359fb1830ade8166c9234536ea40ce7a02646e2R35) and [here](https://github.com/mulesoft/mule/pull/13602/files/788389de81e8038c5ed2d1759b8de6aae5ff4b19#diff-d357c4ff11c93c077669d118a359fb1830ade8166c9234536ea40ce7a02646e2R52).
I haven't checked the details of how the JVM does this, but I would assume that even if I leave the string literals inlined in the code where they are used, those literals would be in memory as long as the class is in memory, same as the static variable. But it could be that they go to different memory sections.

But I don't think it is worth extending the debate for such a small thing, so I will just change it.

runtime-extension-model/src/main/java/org/mule/runtime/core/internal/extension/ForEachCollectionTypeResolver.java
asanguinetti: I forgot to mention it yesterday, but I already changed this one. I made sure all our current `MetadataContext` implementations (which were not that many) implement `InternalMetadataContext` too.

runtime-extension-model/src/main/java/org/mule/runtime/core/internal/extension/ForEachCollectionTypeResolver.java
asanguinetti: I was suggesting the `MuleExpressionLanguage` in particular, not the more generic `ExpressionLanguage`. At least the former already has the `isExpression` logic which checks the delimiters.

On the other hand, one downside of doing it through a non-static object is that it may be hard to get from some context without DI. For example, from the new resolvers we are creating in these PRs, we would need to provide it in the `InternalMetadataContext` too. If we wan't people to stop reimplementing the sanitize logic we should make it easy to reuse, specially considering it is just a few lines of code.

Maybe it is more convenient to just put another `ExpressionLanguageUtils` in `mule-api` and just move it there. The only problem is that it won't be able to access the [same constants](https://github.com/mulesoft/mule/blob/master/core/src/main/java/org/mule/runtime/core/api/el/ExpressionManager.java#L32) defined in `mule-core`, but we can invert that dependency.

I also just found [this other candidate](https://github.com/mulesoft/mule/blob/master/runtime-extension-model/src/main/java/org/mule/runtime/core/internal/expression/util/ExpressionUtils.java#L15). But it would be the same, we would need to repeat the impl, because there is no visibility between `runtime-extensions-model` and `mule-core`.

runtime-extension-model/src/main/java/org/mule/runtime/core/internal/extension/ForEachChainInputTypeResolver.java
asanguinetti: We agreed to keep it as is. Marking as resolved.

runtime-extension-model/src/main/java/org/mule/runtime/core/internal/extension/ForEachCollectionTypeResolver.java
elrodro83: let's put this in a static utility method in mule-api

marianogonzalez (approved):

PR #13604, opened 2024-07-02
W-15509819: Validate `import` target has the same top level element as the importing file

elrodro83:


runtime-extension-model/src/main/java/org/mule/runtime/config/internal/validation/ImportValidTarget.java
```diff
@@ -64,97 +35,8 @@ public Level getLevel() {
...
-    public Map<String, Object> getAnnotations() {
-      return emptyMap();
-    }
-
-    @Override
-    public List<ComponentAst> directChildren() {
-      return emptyList();
-    }
+  protected boolean appliesValidationFor(String resolutionFailureMessage) {
+    return !IMPORT_ROOT_ELEMENT_MISMATCH_MESSAGE_PATTERN.matcher(resolutionFailureMessage).matches();
```
rbourbonnavarro, line 39: Could this condition be clarified?

rbourbonnavarro (approved):

rbourbonnavarro (approved):

